PR #1243, opened 2024-06-13
chore: really really test circular refs

pauldambra:
we want to be really really sure size estimation is safe


pauldambra:
we found a way around this

PR #1247, opened 2024-06-17
feat(surveys): add branching logic

jurajmajerik:
## Changes
^^

## Checklist


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -671,3 +678,105 @@ export const renderChildrenAsTextOrHtml = ({ component, children, renderAsHtml,
...
+        return SurveyQuestionBranchingType.ConfirmationMessage
+    } else if (question.branching.type === SurveyQuestionBranchingType.SpecificQuestion) {
+        if (Number.isInteger(question.branching.index)) {
+            return question.branching.index
+        }
+    } else if (question.branching.type === SurveyQuestionBranchingType.ResponseBased) {
+        // Single choice
+        if (question.type === SurveyQuestionType.SingleChoice) {
+            // :KLUDGE: for now, look up the choiceIndex based on the response
+            // TODO: once QuestionTypes.MultipleChoiceQuestion is refactored, pass the selected choiceIndex into this method
```
jurajmajerik, line 708: Will address this kludge later to not interfere with @dmarticus refactor

src/extensions/surveys/surveys-utils.tsx
```diff
@@ -671,3 +678,105 @@ export const renderChildrenAsTextOrHtml = ({ component, children, renderAsHtml,
...
+            const selectedChoiceIndex = question.choices.indexOf(`${response}`)
+
+            if (question.branching?.responseValues?.hasOwnProperty(selectedChoiceIndex)) {
```
neilkakkar, line 711: We should make this part of the SDK like `getActiveMatchingSurveys`, to `getNextQuestion(survey, currentQuestionIndex, response) -> nextQuestionIndex` for better API mode handling.

Plus, this means we can much more flexibly change things here later, which is bound to happen imo (new branching options, changing brackets, etc. etc.) - all of which would be very hard to handle for API mode if we don't have it as an SDK level primitive.

jurajmajerik: makes sense and done - named it `getNextSurveyStep` since it returns either `index` or `confirmation_message`

src/extensions/surveys.tsx
```diff
@@ -356,16 +363,20 @@ export function Questions({
         originalQuestionIndex: number
         displayQuestionIndex: number
     }) => {
-        const isLastDisplayedQuestion = displayQuestionIndex === survey.questions.length - 1
+        if (!posthog) {
```
neilkakkar, line 366: when does this happen?

src/extensions/surveys.tsx
```diff
@@ -356,16 +363,20 @@ export function Questions({
...
+        }
 
         const responseKey =
             originalQuestionIndex === 0 ? `$survey_response` : `$survey_response_${originalQuestionIndex}`
 
-        if (isLastDisplayedQuestion) {
-            return sendSurveyEvent({ ...questionsResponses, [responseKey]: res }, survey, posthog)
+        setQuestionsResponses({ ...questionsResponses, [responseKey]: res })
+
+        const nextStep = posthog?.getNextSurveyStep(survey, displayQuestionIndex, res)
```
neilkakkar, line 375: Here, `nextStep` should always exist right, so 
```suggestion
        const nextStep = posthog.getNextSurveyStep(survey, displayQuestionIndex, res)
```

I'd expect the above to work^ if it doesn't there's potentially a bug? (or ts being dumb)

src/posthog-surveys.ts
```diff
@@ -155,4 +161,101 @@ export class PostHogSurveys {
...
+            return SurveyQuestionBranchingType.ConfirmationMessage
+        } else if (question.branching.type === SurveyQuestionBranchingType.SpecificQuestion) {
+            if (Number.isInteger(question.branching.index)) {
+                return question.branching.index
+            }
+        } else if (question.branching.type === SurveyQuestionBranchingType.ResponseBased) {
+            // Single choice
+            if (question.type === SurveyQuestionType.SingleChoice) {
+                // :KLUDGE: for now, look up the choiceIndex based on the response
+                // TODO: once QuestionTypes.MultipleChoiceQuestion is refactored, pass the selected choiceIndex into this method
```
neilkakkar, line 187: I think now this is independent of any refactors?

Reminder that choices can be shuffled too, would be nice to see some tests for the single choice case + where the choices are shuffled

src/posthog-surveys.ts
```diff
@@ -155,4 +161,101 @@ export class PostHogSurveys {
...
+                        5: 'detractors',
+                        6: 'detractors',
+                        7: 'passives',
+                        8: 'passives',
+                        9: 'promoters',
+                        10: 'promoters',
+                    }
+                }
+
+                const resNumber = Number(response)
```
neilkakkar, line 237: this needs to defend against invalid input, since users can mistakenly input `choiceX` in response - we should throw a more appropriate error.

src/posthog-surveys.ts
```diff
@@ -155,4 +161,101 @@ export class PostHogSurveys {
...
+                const ratingBucket = responseValueToBucket[resNumber]
```
neilkakkar, line 238: minor nit: The object is hard to parse, specially because of the repeated strings. Since it's only used once, I think works better as a function `getBucketForResponseValue` that just compares the no. to return the bucket.

return resNumber <= 6 ? 'detractors' : resNumber <= 8 ? 'passives' : 'promoters'

src/posthog-surveys.ts
```diff
@@ -155,4 +161,101 @@ export class PostHogSurveys {
...
+
+                    if (nextStep === SurveyQuestionBranchingType.ConfirmationMessage) {
+                        return SurveyQuestionBranchingType.ConfirmationMessage
+                    }
+
+                    return nextQuestionIndex
+                }
+            }
+
+            return nextQuestionIndex
```
neilkakkar, line 256: I think `console.warn`s whenever we fallback because of invalid input would be useful - there's no case I can see where if things are setup correctly we reach here

neilkakkar (commented):
Nicely done!

src/extensions/surveys.tsx
jurajmajerik: `posthog` is an optional argument to `SurveyPopup` (we're not passing it in the preview mode), so this needs to be checked

src/posthog-surveys.ts
jurajmajerik: Should've been clearer - I want to refactor `MultipleChoiceQuestion ` into `SingleSelectQuestion` and `MultipleSelectQuestion`. When that's done, I'll pass `choiceIndex` into this method directly rather than looking it up with the choice value. Want to do it later to not interfere with Dylan (who's also working on this file).

Added the test for single choice w/ shuffling 👍

jurajmajerik:
Changes:
- added a test for single-choice + shuffling
- defend against non-integer input when looking up response bucket
- added getRatingBucketForResponseValue method
- added console.warn when unexpected branching type

src/posthog-surveys.ts
neilkakkar: Looks good! Are there any other cases where it's an unexpected return, or are they all for the implicit next question case?

neilkakkar (approved):

src/__tests__/surveys.test.ts
```diff
@@ -782,4 +789,301 @@ describe('surveys', () => {
...
+        const survey: Survey = {
+            name: 'My survey',
+            description: '',
+            type: SurveyType.Popover,
+            questions: [] as SurveyQuestion[],
+            start_date: new Date().toISOString(),
+            end_date: null,
+        } as Survey
+
+        // Simple branching
```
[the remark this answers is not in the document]
jurajmajerik: tests added :) https://github.com/PostHog/posthog-js/pull/1247/commits/3d0e32b79ee75e15cf068de72bc3b1837bc81909

PR #1119, opened 2024-04-08
feat: Update support for segment analytics

benjackwhite:
## Changes

Fixes https://github.com/PostHog/posthog-js/issues/1106

Annoyingly Segment's documentation does not match reality. They claim that the Analytics JS will always have non-promise responses once it is ready but this isn't the case.

* Changes the Segment setup to all happen in one function
* Accounts for the User object being a promise or not
* Handles both distinct_id and device_id

## Checklist
- [x] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [ ] Accounted for the impact of any changes across different browsers
- [ ] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

PR #1248, opened 2024-06-18
fix: match Segment types with @segment/analytics-next package

George-Payne:
## Problem

Since #1119 typescript complains about passing segment imported from `@segment/analytics-next` due to minor differences in the types. 

image

## Changes

This PR updates the Segment related types to be compatible with `@segment/analytics-next`, without any code changes.

## Checklist


benjackwhite (approved):
Looks fine to me!

PR #1250, opened 2024-06-18
fix: Allow no capture of $opt_in event

benjackwhite:
## Changes

[Ticket](https://posthoghelp.zendesk.com/agent/tickets/14282)

The consent changes I made removed the ability to ignore the opt in event which wasn't intended.

Adds an option to skip using the existing properties. This is slightly breaking but pretty certain it wasn't used this way before anyways

## Checklist

### DIFF
--- a/src/__tests__/consent.test.ts
+++ b/src/__tests__/consent.test.ts
@@ -78,6 +78,34 @@ describe('consentManager', () => {
         expect(posthog.persistence?.disabled).toBe(true)
     })
 
+    it('should send opt in event if not disabled', () => {
+        const onCapture = jest.fn()
+        posthog = createPostHog({ opt_out_capturing_by_default: true, _onCapture: onCapture })
+        posthog.opt_in_capturing()
+        expect(onCapture).toHaveBeenCalledWith('$opt_in', expect.objectContaining({}))
+        onCapture.mockClear()
+
+        posthog.opt_in_capturing({
+            captureEventName: 'override-opt-in',
+            captureProperties: {
+                foo: 'bar',
+            },
+        })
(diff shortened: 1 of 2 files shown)


src/__tests__/consent.test.ts
```diff
@@ -78,6 +78,34 @@ describe('consentManager', () => {
         expect(posthog.persistence?.disabled).toBe(true)
     })
 
+    it('should send opt in event if not disabled', () => {
```
pauldambra, line 81: should be more than one test really 🙈

src/posthog-core.ts
```diff
@@ -1911,17 +1911,20 @@ export class PostHog {
...
     opt_in_capturing(options?: {
-        captureEventName?: string /** event name to be used for capturing the opt-in action */
+        captureEventName?: string | null | false /** event name to be used for capturing the opt-in action */
         captureProperties?: Properties /** set of properties to be captured along with the opt-in action */
     }): void {
         this.consent.optInOut(true)
         this._sync_opt_out_with_persistence()
 
-        // TODO: Do we need it to be sent instantly?
+        if (!isUndefined(options?.captureEventName) && !options.captureEventName) {
```
pauldambra, line 1924:
```suggestion
        if (isNull(options?.captureEventName) || (isBoolean(options?.captureEventName) && options?.captureEventName === false)) {
```

or something like that... it's longer but a little easier to parse IMHO

src/__tests__/consent.test.ts
benjackwhite: FINE

pauldambra (approved):
needs a little massaging if only to make TS happy but otherwise 👍

PR #1251, opened 2024-06-18
feat: Allow bootstrapping session id

robbie-c:
See https://posthog.slack.com/archives/C05LJK1N3CP/p1718670024960049?thread_ts=1718248453.056799&cid=C05LJK1N3CP for motivation

## Changes
* Add an option to bootstrap the session id
* Added a function to convert a v7 uuid to a timestamp

## Checklist

### DIFF
--- a/src/__tests__/sessionid.ts
+++ b/src/__tests__/sessionid.ts
@@ -1,8 +1,8 @@
 import { SessionIdManager } from '../sessionid'
 import { SESSION_ID } from '../constants'
 import { sessionStore } from '../storage'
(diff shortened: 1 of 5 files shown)


benjackwhite (approved):
I think this is generally fine. It should be enough to stop people from doing something weird with sessionIDs.

Blobby won't be happy if we end up with weird IDs but that should be validated on the backend anyways instead

robbie-c:
Session table v2 will straight up ignore sessions with invalid ids, so we should probably add a general ingestion warning too.

PR #1254, opened 2024-06-19
fix: no scheduled snapshots while idle

pauldambra:
see: https://posthoghelp.zendesk.com/agent/tickets/14905 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/17005)

a user complained of long inactive sessions, when looking at them a number of them have a snapshot timer running even though idle

this

* makes sure we don't schedule full snapshots while idle
* reduces the amount of custom events we gather while idle

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -570,11 +570,6 @@ export class SessionRecording {
             // We check if the lastActivityTimestamp is old enough to go idle
             if (event.timestamp - this._lastActivityTimestamp > RECORDING_IDLE_ACTIVITY_TIMEOUT_MS) {
                 this.isIdle = true
-                this._tryAddCustomEvent('sessionIdle', {
-                    reason: 'user inactivity',
-                    timeSinceLastActive: event.timestamp - this._lastActivityTimestamp,
-                    threshold: RECORDING_IDLE_ACTIVITY_TIMEOUT_MS,
-                })
                 // don't take full snapshots while idle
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -223,6 +223,8 @@ export interface SessionRecordingOptions {
     // our settings here only support a subset of those proposed for rrweb's network capture plugin
     recordHeaders?: boolean
     recordBody?: boolean
+    // ADVANCED: while a user is active we take a full snapshot of the browser every interval. For a very few sites playback performance might be better with different interval. Set to 0 to disable
```
daibhin, line 226:
```suggestion
    // ADVANCED: while a user is active we take a full snapshot of the browser every interval. For very few sites playback performance might be better with different interval. Set to 0 to disable
```

src/extensions/replay/sessionrecording.ts
```diff
@@ -586,12 +585,8 @@ export class SessionRecording {
         if (isUserInteraction) {
             this._lastActivityTimestamp = event.timestamp
             if (this.isIdle) {
-                // Remove the idle state if set and trigger a full snapshot as we will have ignored previous mutations
+                // Remove the idle state
                 this.isIdle = false
-                this._tryAddCustomEvent('sessionNoLongerIdle', {
```
daibhin, line 591: What's the reason behind removing events like this? I'd have imagined they were somewhat useful for users in the inspector

daibhin (approved):
One question but looks good

PR #1255, opened 2024-06-20
chore: test prettier

pauldambra:
my IDE and command line prettier are fighting over a line of code in the sentry integration file

i want CI to be the deciding vote on which is wrong


pauldambra:
well that doesn't make any sense

PR #1259, opened 2024-06-20
feat(web-analytics): Add property for autocapture link click

robbie-c:
## Changes

This PR creates a new property, which is the href when an external link is clicked. External is defined as a different host, and the intention here is for people to measure things like which landing pages lead to which external link clicks.

(I was basically nerd-sniped into adding this by some customer feedback)

We already capture this, but it's very buried inside the $element array, and not always at the same place (not always $elements.0.attr_href), so its hard to add to the sessions MV.

Tested this manually by adding an external link to the nextjs playground, and confirming that I could see the event and property.

## Checklist

### DIFF
--- a/src/__tests__/autocapture.test.ts
+++ b/src/__tests__/autocapture.test.ts
@@ -557,6 +557,7 @@ describe('Autocapture system', () => {
             expect(props['$elements'][1]).toHaveProperty('tag_name', 'span')
             expect(props['$elements'][2]).toHaveProperty('tag_name', 'div')
             expect(props['$elements'][props['$elements'].length - 1]).toHaveProperty('tag_name', 'body')
+            expect(props['$click_external_href']).toEqual('https://test.com')
         })
 
         it('truncate any element property value to 1024 bytes', () => {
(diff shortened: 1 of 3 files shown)


src/__tests__/autocapture.test.ts
```diff
@@ -594,7 +595,27 @@ describe('Autocapture system', () => {
...
+            elGrandparent.setAttribute('href', 'https://www.example.com/link')
+            elGrandparent.appendChild(elParent)
+            autocapture['_captureEvent'](
+                makeMouseEvent({
+                    target: elTarget,
+                })
+            )
+            const props = captureMock.mock.calls[0][1]
+            expect(props['$elements'][0]).toHaveProperty('attr__href', 'https://www.example.com/link')
+            expect(props['external_click_url']).toBeUndefined()
```
mariusandra, line 618:
```suggestion
            expect(props['$external_click_url']).toBeUndefined()
```

mariusandra (approved):
Found a typo, otherwise looks good. Not 100% convinced of the property name, but got nothing better.

PR #1261, opened 2024-06-20
fix: exception autocapture

pauldambra:
get setup for exception autocapture in js
2% up is a bit much and might be possible to get that down
but this works again


marandaneto (approved):

pauldambra:
closes #18548

src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -19,27 +20,6 @@ import { isNumber, isString, isUndefined } from '../../utils/type-utils'
...
-    $exception_message: string
-    $exception_source?: string
-    $exception_lineno?: number
-    $exception_colno?: number
-    $exception_DOMException_code?: string
-    $exception_is_synthetic?: boolean
-    $exception_stack_trace_raw?: string
-    $exception_handled?: boolean
-    $exception_personURL?: string
-}
```
daibhin, line 41: Do we know if it's possible to align the Sentry integration properties with what we collect in our own SDK?

daibhin (approved):

src/extensions/exception-autocapture/error-conversion.ts
pauldambra: they should be very similar!
what differences do you see?
(when we did autocapture we changed the sentry integration to send some duplicate properties because we'd have $sentry_exception_type for example - they should all be present!)

marandaneto: There's a bit more info needed here such as the severity, the source maps id for symbolication, environment (staging, prod...), and overridden grouping fingerprint, I'd make the `handled` boolean actually something that allows more states (not a boolean), `ErrorProperties` actually should have a children list of `ErrorProperties`, because an error may have nested (inner) causes

marandaneto: Not a blocker for this PR though.

daibhin: Looks like the Sentry properties are defined in https://github.com/PostHog/posthog-js/blob/3017d39afb80da6b4de895c2d974128b9bde931e/src/extensions/sentry-integration.ts#L100-L119

Missing from there is all the optional params (which might be expected). The Sentry integration includes a `$level` property that we do not. That feels like the most beneficial thing that is missing

daibhin (approved):
Exception? More like EXCITING!!

PR #1265, opened 2024-06-21
feat: Improved various bundles

benjackwhite:
## Changes

* Refactor entrypoints so that they are clearer and more automatable
* Sadly couldn't just user the file extensions `.mts` as tsc doesn't like it so used a sub extension to indicate target
* Add more stuff to the `full` file
* Adds a config option for explicitly stating that you never want to load external scripts

## Checklist


src/entrypoints/exception-autocapture.ts
pauldambra, line 1: want me to update this? it's changed in recent PRs

src/entrypoints/array.full.ts
```diff
@@ -0,0 +1,10 @@
+// Same as loader-globals.ts except includes all additional extension loaders
+
+import './recorder'
+import './surveys'
+import './exception-autocapture'
+import './tracing-headers'
```
pauldambra, line 6: and web vitals too now

package.json
```diff
@@ -23,8 +23,8 @@
         "prepare": "husky install",
         "deprecate-old-versions": "node scripts/deprecate-old-versions.mjs"
     },
-    "main": "dist/module.js",
-    "module": "dist/es.js",
+    "main": "dist/main.js",
+    "module": "dist/module.js",
```
pauldambra, line 27: ELI5? 

(i have managed to ignore changes to JS bundling over the last few chaotic years 🤣)

pauldambra:
feels like there are some QoL improvements you could pull out of this (like the loadScript refactor) that would make it easier to ship them and then easier to review this 🙈

src/entrypoints/exception-autocapture.ts
pauldambra: updated two bits - broke tests - i'll follow up when i've figured out why 🙈

pauldambra: sorted

package.json
benjackwhite: I think we just had it poorly named beforehand. The module wasn't really the module loader 🤷

benjackwhite:
> feels like there are some QoL improvements you could pull out of this (like the loadScript refactor) that would make it easier to ship them and then easier to review this 🙈

That's fair. Didn't want to spend too much time here... But yeah I can split out the bundle changes from the functional changes

pauldambra (approved):
runs locally

pauldambra:
Screenshot 2024-06-26 at 15 31 49

this removal ok?

pauldambra:
i suppose only risk here is someone is importing from `dist/es.js` and their build breaks when updating?

benjackwhite:
> i suppose only risk here is someone is importing from `dist/es.js` and their build breaks when updating?

I think thats fine, especially as it is npm only in that case so they would have the minor upgrade and we don't document that as an option anywhere

PR #1267, opened 2024-06-21
fix: unwind recent changes things seem less stable

pauldambra:
There have been a bunch of preparatory changes for buffer changes, and a bunch of bug fixes recently in the recording file

But things seem less stable than when we started

This unwinds those changes

And only keeps that we make sure not to schedule full snapshots while we're idle

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -61,27 +61,6 @@ const ACTIVE_SOURCES = [
     IncrementalSource.Drag,
 ]
 
-const STYLE_SOURCES = [
-    IncrementalSource.StyleSheetRule,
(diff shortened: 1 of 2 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -827,8 +779,9 @@ export class SessionRecording {
 
         this._updateWindowAndSessionIds(event)
 
-        if (this.isIdle && !allowedWhenIdle(event)) {
-            // When in an idle state we keep recording, but don't capture all events
+        // When in an idle state we keep recording, but don't capture the events
+        // bu we allow custom events even when idle
```
daibhin, line 783:
```suggestion
        // but we allow custom events even when idle
```

daibhin (approved):

PR #1268, opened 2024-06-21
chore: update pnpm action setup worfklow version

pauldambra:
we were using 2
it gives warnings about node 16 when running
latest is 4
lets use 4


marandaneto:
https://github.com/PostHog/posthog-android/blob/main/.github/dependabot.yml
example: https://github.com/PostHog/posthog-android/pull/143
add dependabot to open PRs automatically with GH action updates

marandaneto (approved):

PR #1272, opened 2024-06-25
fix: dropped styles on route transition

daibhin:
## Problem

Relates to https://posthoghelp.zendesk.com/agent/tickets/14512 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/18047) (and probably a bunch more support tickets)

This could also be related to https://github.com/rrweb-io/rrweb/issues/1230

Addresses an issue in recordings where styles were being dropped during some route changes in SPAs

### Reproduction
It's not fully clear to me how this happens or what the correct behaviour is but the problem exists when:

1. You have a relative link element on the page:
```
<link href="./_app/immutable/assets/filename.css" rel="stylesheet">
```
2. This gets loaded into the documents `styleSheets` array with the href set as `https://domainname.com/_app/immutable/assets/filename.css`

3. Sometime later the domain changes away from the index page to another root e.g. `https://domainname.com/home`
4. When taking the next full snapshot [rrweb tries to fetch the stylesheet associated with the href](https://github.com/rrweb-io/rrweb/blob/4f44bd17aa888c041e9a4c8c0f4c921af92a4b12/packages/rrweb-snapshot/src/snapshot.ts#L657-L659):
```
const stylesheet = Array.from(doc.styleSheets).find((s) => {
  return s.href === (n as HTMLLinkElement).href;
});
```
5. `n.href` now evaluates to `https://domainname.com/home/_app/immutable/assets/filename.css` which no longer matches the `href` of the stylesheet on the document and hence cannot be "found". Presumably the href on the document stylesheet does not change because the SPA is not doing a full page transition

## Changes

This PR adds a backup check to look for the CSS on the root of the domain
- It only does this for CSS files
- It only does the lookup if a stylesheet cannot initially be found

This feels like the safest change for now but will get the opinion of others in the rrweb community too

### DIFF
--- a/patches/rrweb@2.0.0-alpha.13.patch
+++ b/patches/rrweb@2.0.0-alpha.13.patch
@@ -167,7 +167,7 @@ index ea868845c4fad3276aa8e5f74abfd3568b466d11..965505de44975e718d431a4e9a62e753
  
  export { WorkerFactory as default };
 diff --git a/es/rrweb/packages/rrweb/src/record/observers/canvas/canvas-manager.js b/es/rrweb/packages/rrweb/src/record/observers/canvas/canvas-manager.js
-index da2c103fe6b1408a5996f0eb3bf047571e434cc2..3b34097a089791bd1043f313dcab9ad4c3733830 100644
+index da2c103fe6b1408a5996f0eb3bf047571e434cc2..f5b060c7e0728a3e2c6cf62b01d97282f2484ac3 100644
 --- a/es/rrweb/packages/rrweb/src/record/observers/canvas/canvas-manager.js
 +++ b/es/rrweb/packages/rrweb/src/record/observers/canvas/canvas-manager.js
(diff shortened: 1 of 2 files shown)


marandaneto (approved):

pauldambra:
i'd love to have an error handler we could pass in to rrweb so we could log errors like this - they're so difficult to find!

daibhin:
@pauldambra totally agree, that would be great to have. This _technically_ isn't an error. It is expected that a stylesheet shouldn't be found for links to JS files. I'm somewhat special casing things here in cases when there is a `.css` suffix on the file

daibhin:
@MathiasWP that's super weird, I don't see a reference to `e.href` anywhere in the code I added. Will look into this straight away

daibhin:
A fix has just been merged in https://github.com/PostHog/posthog-js/pull/1275. Thanks for reporting this @MathiasWP

PR #1273, opened 2024-06-25
feat(survey): Allow events to repeatedly activate surveys

Phanatic:
## Changes
This is repeated feedback (heheh) we've received from customers about events activating surveys repeatedly.
To that end, I've added a new option that survey authors can select so that we can show a survey every time an event in their selection occurs, not just the first time.

Posthog PR is at https://github.com/PostHog/posthog/pull/23238

## Checklist


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -625,6 +625,35 @@ export const getDisplayOrderQuestions = (survey: Survey): SurveyQuestion[] => {
     return reverseIfUnshuffled(survey.questions, shuffle(survey.questions))
 }
 
+export const hasEvents = (survey: Survey): boolean => {
+    return !!(
+        survey.conditions &&
+        survey.conditions?.events &&
+        survey.conditions?.events?.values &&
+        survey.conditions?.events?.values?.length > 0
+    )
```
jurajmajerik, line 634: same as https://github.com/PostHog/posthog/pull/23238#discussion_r1660835388

src/posthog-surveys-types.ts
```diff
@@ -138,6 +138,7 @@ export interface Survey {
         selector?: string
         seenSurveyWaitPeriodInDays?: number
         urlMatchType?: SurveyUrlMatchType
+        repeatedActivation?: boolean
```
jurajmajerik, line 141: same as https://github.com/PostHog/posthog/pull/23238/files#r1660845077

Phanatic:
Thanks for the review @jurajmajerik, I've made the following changes : 
1. Move the `repeatedActivation` property into the `events` object so its clear what the scope of this property is. 
2. Simplify the condition to check for events.

jurajmajerik (approved):
🙌

src/utils/survey-event-receiver.ts
pauldambra, line 1: fly-by that +7% bump in bundle size is pretty big... could this event receiver be in the lazily loaded surveys code?

i only scanned it but we have an SimpleEventEmitter as well... the names sound similar but there might not be enough duplication to justify trying to avoid the duplication 🙈

PR #1275, opened 2024-06-26
fix: use includes over contains

daibhin:
## Changes

Use includes to check substring. `contains` is throwing a weird error as reported in https://github.com/PostHog/posthog-js/pull/1272#issuecomment-2191304203

|Before|After|
|----|----|
| Screenshot 2024-06-26 at 11 20 36 | Screenshot 2024-06-26 at 11 20 42 |

|Before|After|
|----|----|
| Screenshot 2024-06-26 at 11 20 16 | Screenshot 2024-06-26 at 11 21 09 |

### DIFF
--- a/patches/rrweb@2.0.0-alpha.13.patch
+++ b/patches/rrweb@2.0.0-alpha.13.patch
@@ -224,20 +224,22 @@ index da2c103fe6b1408a5996f0eb3bf047571e434cc2..f5b060c7e0728a3e2c6cf62b01d97282
                  }, [bitmap]);
              }));
 diff --git a/es/rrweb/packages/rrweb-snapshot/es/rrweb-snapshot.js b/es/rrweb/packages/rrweb-snapshot/es/rrweb-snapshot.js
-index 0fc7d4bcafc9be822347f9437658478fd77d9972..a2400dbd745664cc94335f9a0d04bb52af91ee13 100644
+index 0fc7d4bcafc9be822347f9437658478fd77d9972..5dcd1dcbd5380567110378269f3bc8d7b7ca6440 100644
 --- a/es/rrweb/packages/rrweb-snapshot/es/rrweb-snapshot.js
 +++ b/es/rrweb/packages/rrweb-snapshot/es/rrweb-snapshot.js
-@@ -641,9 +641,17 @@ function serializeElementNode(n, options) {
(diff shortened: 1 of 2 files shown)


patches/rrweb@2.0.0-alpha.13.patch
```diff
@@ -224,20 +224,22 @@ index da2c103fe6b1408a5996f0eb3bf047571e434cc2..f5b060c7e0728a3e2c6cf62b01d97282
...
      if (tagName === 'link' && inlineStylesheet) {
 -        const stylesheet = Array.from(doc.styleSheets).find((s) => {
+-            return s.href === n.href;
++        const href = n.href
 +        let stylesheet = Array.from(doc.styleSheets).find((s) => {
-             return s.href === n.href;
++            return s.href === href;
          });
-+        if (!stylesheet && n.href.contains('.css')) {
++        if (!stylesheet && href.includes('.css')) {
```
marandaneto, line 240: probably not an issue but includes is case sensitive
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/includes

marandaneto (approved):

patches/rrweb@2.0.0-alpha.13.patch
daibhin: Should be ok for our use case. I'm not sure `.CSS` is a valid file suffix

patches/rrweb@2.0.0-alpha.13.patch
```diff
@@ -224,20 +224,22 @@ index da2c103fe6b1408a5996f0eb3bf047571e434cc2..f5b060c7e0728a3e2c6cf62b01d97282
...
      }
      if (tagName === 'link' && inlineStylesheet) {
 -        const stylesheet = Array.from(doc.styleSheets).find((s) => {
+-            return s.href === n.href;
++        const href = n.href
 +        let stylesheet = Array.from(doc.styleSheets).find((s) => {
-             return s.href === n.href;
++            return s.href === href;
          });
-+        if (!stylesheet && n.href.contains('.css')) {
```
marandaneto, line 238: not part of the mozilla spec at least https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String so probably depends on browser yeah

pauldambra:
Lgtm

PR #1277, opened 2024-06-28
feat: warn when distinct id invalid for replay

pauldambra:
see https://posthog.sentry.io/issues/5544450836/?query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D&referrer=issue-stream&statsPeriod=1h&stream_index=2

it's all one customer but we should warn in this scenario


daibhin (approved):

daibhin (commented):
Should we drop the events or is that too much?

pauldambra:
well, we're going to drop them at capture, and return a 400 (since the last couple of days changes) so I think we're ok with a log error to point someone at as well

currently this is only one customer, so even though it's pretty high volume from them, they seem to track latest version, so this will maybe help them understand what is happening 🙈

PR #1278, opened 2024-06-28
fix(Surveys): do not activate survey if events have changed

Phanatic:
## Changes

Fixes https://github.com/PostHog/posthog/issues/23336
A customer ran into an issue where changing the event that activates a survey causes any old events observed to also activate the survey. In their own words : 
```
This bug is quite easy to replicate :
1. I create a pop-up survey with the only condition being that a certain event should be triggered
2. I do this event
3. The pop-up shows up
4. I change the event for the same pop-up survey
5. I refresh my application (F5)
6. The pop-up still shows up although I didn’t trigger the new event
```

Basically, what happens is that after Step 2 in the above list, we store an array of all activated surveys in the local storage on the client browser. And this persists across browser refreshes. However, if the event that activates a survey has changed, we still show the survey because we store the survey id from the last time an event matching it was observed. 
This can be confusing, rightfully so. 

The fix here is to : 
1. Save the `events observed`, not `surveys activated` in browser session. 
2. On calling `register`,  maintain a mapping of event names to the surveys that are activated by them.
3. When the call to `getSurveys` is made, delaly load the surveys by  matching the surveys to events and return the surveys.

## Checklist


dmarticus (approved):

Phanatic:
This is not necessary anymore, because this fix is in #1273

PR #1279, opened 2024-07-01
fix(surveys): polishing the popup survey UI

dmarticus:
## Changes

Closes issues 1,3, and 4 from this ticket https://github.com/PostHog/posthog/issues/20851.

Fixing 4 requires some changes to posthog, too: https://github.com/PostHog/posthog/pull/23460

Issue 2 I wasn't able to reproduce; I need more context from @annikaschmid and/or @jurajmajerik 

Quick note on browser compatibility: I'm using the `:has` pseudo-selector for determining whether to apply margins or not.  Per [caniuse](https://caniuse.com/css-has), this has been supported in all major browsers since 2023, but is this a risky change still?

### After

...

## Checklist
- I manually tested this for each survey type

### DIFF
--- a/src/extensions/surveys/surveys-utils.tsx
+++ b/src/extensions/surveys/surveys-utils.tsx
@@ -57,6 +57,9 @@ export const style = (appearance: SurveyAppearance | null) => {
               border-color: ${appearance?.borderColor || '#c9c6c6'};
               margin-top: 14px;
           }
+          .survey-box:has(.survey-question:empty) textarea {
+              margin-top: 0;
+          }
           .form-submit {
               box-sizing: border-box;
               margin: 0;
@@ -197,6 +200,9 @@ export const style = (appearance: SurveyAppearance | null) => {
           .multiple-choice-options {
               margin-top: 13px;
               font-size: 14px;
+          }
+          .survey-box > div:has(.survey-question:empty) + .multiple-choice-options {
+              margin-top: 0;
           }
           .multiple-choice-options .choice-option {
               display: flex;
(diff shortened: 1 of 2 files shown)


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -57,6 +57,9 @@ export const style = (appearance: SurveyAppearance | null) => {
               border-color: ${appearance?.borderColor || '#c9c6c6'};
               margin-top: 14px;
           }
+          .survey-box:has(.survey-question:empty):not(:has(.description)) textarea {
+              margin-top: 0;
+          }
```
dmarticus, line 62: I only want to remove the margin in the event where there's no description (the selector is _gone_) AND there's no question (the selector is _empty_).  Otherwise, the margins should remain.

jurajmajerik: > I'm using the :has pseudo-selector ... supported in all major browsers since 2023, but is this a risky change still?

This is fine in my opinion. It's a small styling detail that just won't work in a tiny minority of cases.

jurajmajerik:
> Issue 2 I wasn't able to reproduce

This is more like a feature request than a bug - we currently don't allow customization of the confirmation message text. Feel free to address in a follow up :)

dmarticus:
@jurajmajerik yeah, I haven't finished part 4 yet, sorry!  I thought I'd gotten it working but it's not ready.

dmarticus:
> Same goes for the top margin: image
> 
> I thought maybe I wasn't linking `posthog-js` properly, but not the case - I can see the new stylesheet under Sources/surveys-utils.tsx and the confirmation message _is_ being sized correctly.

Wait, so are you saying you can't reproduce the top margin being re-sized correctly? 

This is what I see when I run it with posthog-js locally 

image

jurajmajerik:
There's indeed something going wrong with my `posthog-js` linking. I've just sunk an hour trying to figure it out, but I'm gonna cut my losses - I trust you this is working :)

jurajmajerik (approved):

PR #1282, opened 2024-07-03
feat: add payload host denylist

pauldambra:
We've seen replay message be rejected for being too large
And on inspection they contain many megabytes of replay data _for other vendors_
There's no point capturing those payloads 🙈 

* adds a default deny list
* adds a hook that we could use to start overriding it via decide response
* ...

### DIFF
--- a/src/extensions/replay/external/README.md
+++ b/src/extensions/replay/external/README.md
@@ -1,0 +1,5 @@
+files in here are intended for the lazy loaded portion of replay
+
+you aren't supposed to import them from outside the entrypoint file
+
+they could cause an increase in bundle size
--- a/src/extensions/replay/external/denylist.ts
+++ b/src/extensions/replay/external/denylist.ts
@@ -1,0 +1,32 @@
+import { NetworkRecordOptions } from '../../../types'
+
+function hostnameFromURL(url: string | URL | RequestInfo): string | null {
+    try {
+        if (typeof url === 'string') {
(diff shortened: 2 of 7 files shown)


src/extensions/replay/external/README.md
pauldambra, line 1: giving us a place we can start moving code out of the entrypoint file 
as we do we should add a lint rule restricting what can import it

src/extensions/replay/external/denylist.ts
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+export function isHostOnDenyList(url: string | URL | Request, options: NetworkRecordOptions) {
+    const hostname = hostnameFromURL(url)
+    const defaultNotDenied = { hostname, isHostDenied: false }
+
+    if (!options.payloadHostDenyList?.length || !hostname?.trim().length) {
+        return defaultNotDenied
+    }
```
pauldambra, line 23: fail open

src/extensions/replay/external/denylist.ts
```diff
@@ -0,0 +1,32 @@
...
+
+    for (const deny of options.payloadHostDenyList) {
+        if (hostname.endsWith(deny)) {
```
pauldambra, line 26: went with `endsWith` over regex or something...

so that we can say e.g. `.lr-ingest.io` and not get any of their payloads

PR #1284, opened 2024-07-03
Add esm data to package json to fix #908

### LINKED ISSUE #908: missing `pkg.exports` in package.json
It is not possible to use `posthog-js` package within an ESM project in Node environment, because the package.json is missing the exports property (https://publint.dev/posthog-js@1.92.1) while Node doesn't read the module property and thus it fails to treat it as an ES package.

You get the following error if you try: 
```
import { PostHogProvider } from "posthog-js/react/dist/esm/index.js";
         ^^^^^^^^^^^^^^^
SyntaxError: Named export 'PostHogProvider' not found. The requested module 'posthog-js/react/dist/esm/index.js' is a CommonJS module, which may not support all module.exports as named exports.
CommonJS modules can always be imported via the default export, for example using:

import pkg from 'posthog-js/react/dist/esm/index.js';
const { PostHogProvider } = pkg;
```
This is within a Remix Vite project, but I guess it's relevant for any ESM project that runs SSR in Node and doesn't prebundle dependencies. 

I had to apply the following patch to get it working:

```
diff --git a/dist/es.js b/dist/es.mjs
similarity index 100%
rename from dist/es.js
rename to dist/es.mjs
diff --git a/dist/es.js.map b/dist/es.mjs.map
similarity index 100%
rename from dist/es.js.map
rename to dist/es.mjs.map
diff --git a/package.json b/package.json
index 4d35188aea0eed88a85d9c9c3a140ea10f578034..c10df882ec27ca97a29c814735f8ea53ad847dc2 100644
--- a/package.json
+++ b/package.json
@@ -23,6 +23,18 @@
     },
     "main": "dist/module.js",
     "module": "dist/es.js",
+    "exports": {
+        ".": {
+            "types": "./dist/module.d.ts",
+            "import": "./dist/es.mjs",
+            "require": "./dist/module.js"
+        },
+        "./react": {
+            "types": "./react/dist/types/index.d.ts",
+            "import": "./react/dist/esm/index.mjs",
+            "require": "./react/dist/umd/index.js"
+        }
+    },
     "types": "dist/module.d.ts",
     "files": [
         "lib/*",
diff --git a/react/README.md b/react/README.md
deleted file mode 100644
index 6d2737d54f58c39c85eea2afcfb6dd583686aef9..0000000000000000000000000000000000000000
diff --git a/react/dist/esm/index.js b/react/dist/esm/index.mjs
similarity index 100%
rename from react/dist/esm/index.js
rename to react/dist/esm/index.mjs
diff --git a/react/package.json b/react/package.json
index f313bf7c1e42c7d43be66390329013fd342558a0..8932551985ed093f2f2ab25ee48cdb0543b97a71 100644
--- a/react/package.json
+++ b/react/package.json
@@ -21,6 +21,13 @@
     },
     "main": "dist/umd/index.js",
     "module": "dist/esm/index.js",
+    "exports": {
+        ".": {
+            "types": "./dist/types/index.d.ts",
+            "import": "./dist/esm/index.mjs",
+            "require": "./dist/umd/index.js"
+        }
+    },
     "types": "dist/types",
     "files": [
         "dist/*",
```

matt-hernandez:
## Changes
Add necessary `package.json` data to root and `/react` folder to fix compat issues between posthog and Remix with Vite. Fixes #908 

This fix is based on what I had to do at my job to make `posthog-js` and `posthog-js/react` play nicely with our Remix w/ Vite app. Locally, tests passed, however e2e tests had some flakiness. My gut instinct tells me that the intermittent failures were not due to the changes to any of the `package.json` files. 

Also this fix grew past what I thought it would originally be. By converting things to an ES module, several other files had to be renamed in order to be compatible with CommonJS format, specifically for ESLint. Will be watching Github actions to see if things complete without failure.
...

## Checklist
- [ *] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [ *] Accounted for the impact of any changes across different browsers
- [ *] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)


package.json
```diff
@@ -118,5 +131,8 @@
         "patchedDependencies": {
             "rrweb@2.0.0-alpha.13": "patches/rrweb@2.0.0-alpha.13.patch"
         }
+    },
+    "engines": {
+        "node": ">=20.1.0"
     }
```
pauldambra, line 137: why do we need to bump the node version here?

we try to be pretty conservative with changes since it can be hard to predict consequence on other folks CI

is it this change that means we need the changes to eslint?

pauldambra:
hey @matt-hernandez 👋 

thanks for this! it's a little beyond me so a few questions to get started 😊

package.json
```diff
@@ -26,6 +26,19 @@
...
+            "types": "./dist/module.d.ts",
+            "import": "./dist/es.js",
+            "require": "./dist/module.js"
+        },
+        "./react": {
+            "types": "./react/dist/types/index.d.ts",
+            "import": "./react/dist/esm/index.js",
+            "require": "./react/dist/umd/index.js"
+        }
+    },
```
pauldambra, line 41: i've been sort of ignoring the module/bundling changes in JS since they seem pretty volatile and i was hoping for things to settle and i could learn it once 🤣 

can you explain what this does/means?

i'm mostly wondering if we'll have unexpected consequences in other environments than you're testing for, i've found sometimes apparently benign changes throw up weird errors for folk

(my guess is this is safe but 🙈)

package.json
pauldambra: (i'm pretty sure we can support at least >=18 since that's what i'm using locally, i haven't tested with 16)

matt-hernandez:
@pauldambra The node version requirement is due to Cypress not running e2e tests with ESM unless it runs in node > 20.1.0. There may be another workaround, but in local, that was the only way I could get Cypress working.

The exports of the package.json files combined with `”type”: “module”` help to label this package as an ESM compatible package, while also pointing bundlers that rely on CommonJS/UMD format which files to go to for their respective format type.

Unfortunately, by converting this repo to full ESM, the changes ballooned. Pretty much anything with CommonJS format is having to change. Eslint is also completely borked after the conversion to ESM. And I think the only stable way to fix it would be to bump up ESLint 9+ and convert `.eslintrc.js` to the more recommended `eslint.config.js` file with flat config.

I’m converting this to a draft in the meantime since changing everything to ESM represents more than just a few metadata changes.

package.json
pauldambra: > The node version requirement is due to Cypress not running e2e tests with ESM unless it runs in node > 20.1.0. There may be another workaround, but in local, that was the only way I could get Cypress working.

ok, the version requirement bump is likely to be painful for an SDK. Node 18 isn't end of life until Apr 2025 

but equally don't want to leave folk with your problem hanging...

maybe we can do this another way - we have an examples folder - you can see for e.g. the nextjs folder at https://github.com/PostHog/posthog-js/tree/main/playground/nextjs

could you open another pr with a vite/remix example - (as small as possible 😊)

that should let us see the problem and help us keep it working in future

(i've never touched remix or i'd offer to make one - but i'm guessing you'll do a better job 💖 )

matt-hernandez:
@pauldambra I've created PR #1293 to demo the issue from #908

matt-hernandez:
I’ll close this one since we have a different one open that’s demonstrating the underlying problem

PR #1287, opened 2024-07-05
feat: move heatmaps to their own event

pauldambra:
see https://github.com/PostHog/posthog/issues/23395

this moves heatmap data from being a passenger on other events to being its own event

https://github.com/PostHog/posthog/pull/23505 is merged so ingestion can now handle existing passenger events or these new $$heatmap events

just a final round of testing for this SDK side now remains

### DIFF
--- a/src/heatmaps.ts
+++ b/src/heatmaps.ts
@@ -6,7 +6,7 @@ import { PostHog } from './posthog-core'
 import { document, window } from './utils/globals'
 import { getParentElement, isTag } from './autocapture-utils'
 import { HEATMAPS_ENABLED_SERVER_SIDE, TOOLBAR_ID } from './constants'
-import { isObject, isUndefined } from './utils/type-utils'
+import { isEmptyObject, isObject, isUndefined } from './utils/type-utils'
 import { logger } from './utils/logger'
 
 const DEFAULT_FLUSH_INTERVAL = 5000
@@ -192,7 +192,7 @@ export class Heatmaps {
     }
 
     private flush(): void {
-        if (!this.buffer) {
+        if (!this.buffer || isEmptyObject(this.buffer)) {
             return
         }
 
--- a/src/heatmaps.ts
+++ b/src/heatmaps.ts
@@ -9,7 +9,7 @@ import { HEATMAPS_ENABLED_SERVER_SIDE, TOOLBAR_ID } from './constants'
 import { isObject, isUndefined } from './utils/type-utils'
 import { logger } from './utils/logger'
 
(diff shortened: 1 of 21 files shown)


src/heatmaps.ts
```diff
@@ -6,9 +6,13 @@ import { PostHog } from './posthog-core'
 import { document, window } from './utils/globals'
 import { getParentElement, isTag } from './autocapture-utils'
 import { HEATMAPS_ENABLED_SERVER_SIDE, TOOLBAR_ID } from './constants'
-import { isUndefined } from './utils/type-utils'
+import { isEmptyObject, isObject, isUndefined } from './utils/type-utils'
 import { logger } from './utils/logger'
 
+const DEFAULT_FLUSH_INTERVAL = 5000
+const HEATMAPS = 'heatmaps'
+const LOGGER_PREFIX = '[' + HEATMAPS + ']'
```
daibhin, line 14: Why the separate const here?

src/types.ts
```diff
@@ -283,8 +293,6 @@ export interface CaptureOptions {
     $set?: Properties /** used with $identify */
     $set_once?: Properties /** used with $identify */
     _url?: string /** Used to override the desired endpoint for the captured event */
-    /** Some events are sent as passengers inside other events - e.g. heatmaps and web vitals, not all ingestion routes can process passengers */
```
daibhin, line 286: Are web vitals the only passenger event types now? Is it worth making them full events

src/heatmaps.ts
```diff
@@ -160,4 +190,14 @@ export class Heatmaps {
 
         this.buffer[url].push(properties)
     }
+
+    private flush(): void {
+        if (!this.buffer || isEmptyObject(this.buffer)) {
+            return
+        }
+
+        this.instance.capture('$$heatmap', {
```
daibhin, line 199: Why does heatmaps use the double $?

daibhin (approved):
Couple of clarifying questions but all good otherwise

src/heatmaps.ts
pauldambra: force of habit :)
as the number of log messages grows by copy pasta, the file size doesn't increase by the multiple of the string bytes

src/types.ts
pauldambra: they aren't passengers... just a dangling comment

src/heatmaps.ts
pauldambra: i believe it makes it not chargeable - it's like our marker for internal events instead of posthog events - trying to figure that out currently cos I can drop it completely in pipeline once we've extracted the data

PR #1291, opened 2024-07-08
chore: fix error tests listener registration

pauldambra:
I can't consistently make this happen locally

But I did once notice that I don't always get

```
possible EventEmitter memory leak detected
```

i only noticed after the fact but if we didn't register an event listener then the tests wouldn't pass 

i've checked console logs in prod for our team and don't see this so it's safe to patch in tests (🙈)

### DIFF
--- a/cypress/support/e2e.ts
+++ b/cypress/support/e2e.ts
@@ -32,7 +32,7 @@ beforeEach(() => {
 
     cy.readFile('dist/recorder.js').then((body) => {
         cy.intercept('/static/recorder.js*', { body }).as('recorder')
-        cy.intercept('/static/recorder-v2.js*', { body }).as('recorder')
+        cy.intercept('/static/recorder-v2.js*', { body }).as('recorderv2')
     })
 
     cy.readFile('dist/recorder.js.map').then((body) => {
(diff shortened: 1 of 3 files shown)


cypress/support/setup.ts
```diff
@@ -19,14 +21,18 @@ export const start = ({
...
+    EventEmitter.prototype._maxListeners = 100
+
     const decideResponse = {
         editorParams: {},
         featureFlags: ['session-recording-player'],
         supportedCompression: ['gzip-js'],
         excludedDomains: [],
         autocaptureExceptions: false,
         ...decideResponseOverrides,
-        config: { ...decideResponseOverrides.config },
```
pauldambra, line 29: nobody sends this, and `DecideResponse` doesn't have a config prop so 🔪

cypress/support/e2e.ts
```diff
@@ -32,7 +32,7 @@ beforeEach(() => {
 
     cy.readFile('dist/recorder.js').then((body) => {
         cy.intercept('/static/recorder.js*', { body }).as('recorder')
-        cy.intercept('/static/recorder-v2.js*', { body }).as('recorder')
+        cy.intercept('/static/recorder-v2.js*', { body }).as('recorderv2')
```
pauldambra, line 35: we should remove this tbh but fly-by rename so it doesn't clash

cypress/support/e2e.ts
```diff
@@ -48,7 +48,7 @@ beforeEach(() => {
     })
 
     cy.readFile('dist/exception-autocapture.js').then((body) => {
-        cy.intercept('/static/exception-autocapture.js*', { body })
+        cy.intercept('/static/exception-autocapture.js*', { body }).as('exception-autocapture-script')
```
pauldambra, line 51: better safe than sorry

PR #1292, opened 2024-07-08
feat(Surveys): Allow actions to activate surveys

Phanatic:
## Description
This PR allows us to activate surveys based on user actions performed in the browser.

### Not supported
1. Matching person properties in Actions,  when matching surveys. We will not implement this as we don't want to download person properties on the client for surveys.
2. Matching event properties in Actions, when matching surveys. We could do this in a follow-up PR because the event properties are available on the client-side.

## Changes

Here are the set of changes I needed to make to implement action matching in posthog-js

### AutoCapture
1. Autocapture can now register a set of html element selectors. 
4. These html selectors, are checked against the target element **that** was captured
5. If there is a match, those element selectors are sent along as `$element_selectors` property for a given event.

### Survey Event Receiver
1. Instead of maintaining a map of Survey-ID to its actions or events, we maintain a map of Action/Event to the surveys that are activated by it.
1. We now look for, and register any surveys that are activated by events with the `ActionMatcher`.
2. Changed `on` to be `onEvent` and added a new `onAction` event handler.
3. `onAction` listens on the `actionCaptured` event in `ActionMatcher` described below.

### ActionMatcher
This is a pared down version of the action-matcher found in the ingestion pipeline, main differences are : 
1. Uses the `$element_selectors` property introduced in AutoCapture to match events to an action step.
2. Uses the `elementsJson` that is passed along as part of CaptureResult to check other properties like tagname, text, url etc
4. Raises an `actionCaptured` event when a `CaptureResult` matches a known action step.

### Flowchart for matching actions on the client-side.
``` mermaid
flowchart TD
    A[User clicks button] -->|AutoCapture| B(Analyze element)
    B --> |Element matches known selector|C(Set $element_selectors property)
    C -->D[Capture Action]
    B --> |Element does not match known selector|D[Capture Action]
    D --> |Raise eventCaptured with CaptureResult|E[Event Handlers]
    E --> F[Survey Event Receiver onEvent] 
    F --> H[Match event name to known event names]
    H --> |eventName matches a survey's events|I[Add survey to activatedSurveys]
    E --> G[Action Matcher]
    G --> |Check Action Step Matches CaptureResult|J[Raise actionCaptured with actionName]
    J --> K[Survey Event Receiver onAction]
    K --> L[Match action name to known action names]
    L --> |actionName matches a survey's actions|M[Add survey to activatedSurveys]
```
...

## Checklist
- [ ] Remove all debugging code 😄 
[...]

PR #1294, opened 2024-07-09
feat(Surveys): Add ActionMatcher to match CaptureResult to any known actions

Phanatic:
## Changes

_Broken out of #1292 into its own PR for easier reviewing and merging_

This PR allows us to match user activities performed in the browser to any known actions.

### Not supported
1. Matching person properties in Actions,  when matching surveys. We will not implement this as we don't want to download person properties on the client for surveys.
2. Matching event properties in Actions, when matching surveys. We could do this in a follow-up PR because the event properties are available on the client-side.

## Changes

Here are the set of changes I needed to make to implement action matching in posthog-js

### AutoCapture
1. Autocapture can now register a set of html element selectors. 
4. These html selectors, are checked against the target element **that** was captured
5. If there is a match, those element selectors are sent along as `$element_selectors` property for a given event.

### ActionMatcher
This is a pared down version of the action-matcher found in the ingestion pipeline, main differences are : 
1. Uses the `$element_selectors` property introduced in AutoCapture to match events to an action step.
2. Uses the `elementsJson` that is passed along as part of CaptureResult to check other properties like tagname, text, url etc
4. Raises an `actionCaptured` event when a `CaptureResult` matches a known action step.

### Flowchart for matching actions on the client-side.
``` mermaid
flowchart TD
    A[User clicks button] -->|AutoCapture| B(Analyze element)
    B --> |Element matches known selector|C(Set $element_selectors property)
    C -->D[Capture Action]
    B --> |Element does not match known selector|D[Capture Action]
    D --> |Raise eventCaptured with CaptureResult|E[Event Handlers]
    E --> F[Action Matcher]
    F --> |Check Action Step Matches CaptureResult|G[Raise actionCaptured with actionName]
```

## Checklist

### DIFF
--- a/src/__tests__/autocapture.test.ts
+++ b/src/__tests__/autocapture.test.ts
@@ -583,6 +583,27 @@ describe('Autocapture system', () => {
             expect(props['$elements'][0]).toHaveProperty('attr__data-props', 'prop'.repeat(256) + '...')
         })
 
+        it('assigns element_selector if htmlelement matches known selectors', () => {
+            const elTarget = document.createElement('img')
+            elTarget.id = 'primary_button'
+            const elParent = document.createElement('span')
+            elParent.appendChild(elTarget)
+            autocapture.setElementSelectors(new Set<string>('#primary_button'))
+            const elGrandparent = document.createElement('a')
+            elGrandparent.setAttribute('href', 'https://test.com')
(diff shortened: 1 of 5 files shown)


jurajmajerik (approved):
🚢

PR #1296, opened 2024-07-09
fix: capture network recording type

pauldambra:
see https://posthog.com/questions/hiding-specific-query-parameters-in-session-replay-waterfall-beta-view

allows caller to edit the request, properties being readonly was an unintended consequence of inheriting from PerformanceEntry and never mutating it ourselves

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -479,9 +479,8 @@ export type NetworkRequest = {
 // we mirror PerformanceEntry since we read into this type from a PerformanceObserver,
 // but we don't want to inherit its readonly-iness
 type Writable<T> = { -readonly [P in keyof T]: T[P] }
-type CapturedNetworkRequestBase = Writable<Omit<PerformanceEntry, 'toJSON'>>
 
-type CapturedNetworkRequest = CapturedNetworkRequestBase & {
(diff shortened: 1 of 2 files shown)


src/types.ts
```diff
@@ -475,7 +475,13 @@ export type NetworkRequest = {
...
 //     readonly startTime: DOMHighResTimeStamp;
 // NB: properties below here are ALPHA, don't rely on them, they may change without notice
-export type CapturedNetworkRequest = Omit<PerformanceEntry, 'toJSON'> & {
+
+// we mirror PerformanceEntry since we read into this type from a PerformanceObserver,
+// but we don't want to inherit its readonly-iness
+type Writable<T> = { -readonly [P in keyof T]: T[P] }
+type CapturedNetworkRequestBase = Writable<Omit<PerformanceEntry, 'toJSON'>>
+
+type CapturedNetworkRequest = CapturedNetworkRequestBase & {
```
pauldambra, line 484:
```suggestion

export type CapturedNetworkRequest = Writable<Omit<PerformanceEntry, 'toJSON'>> & {
```

daibhin (approved):

PR #1302, opened 2024-07-12
feat: add url_ignorelist for autocapture config

pauldambra:
see https://posthog.slack.com/archives/C075D3C5HST/p1720808391615839

and https://posthog.com/questions/url-allowlist-for-exclude

@corywatilo made a totally reasonable assumption about how this works... and it's easy to make it work that way

This PR extends autocapture config so you can pass a url_ignorelist instead of (or as well as) a url_allowlist


Phanatic (approved):
Looks great, ship it!

PR #1307, opened 2024-07-16
feat: split large incremental snapshots

pauldambra:
we see very large incremental snapshots 
large enough we can't ingest them
they often have large arrays of adds or attribute mutations
those mutations need to be applied in the same order but they don't need to be in one snapshot

todo

- [x] run locally with an artificially low size limit so we can see it work
- [x] do we need to do any timestamp fangling to preserve order


src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -165,27 +166,154 @@ export function truncateLargeConsoleLogs(_event: eventWithTime) {
 
 export const SEVEN_MEGABYTES = 1024 * 1024 * 7 * 0.9 // ~7mb (with some wiggle room)
 
-// recursively splits large buffers into smaller ones
+function sliceList(list: any[], sizeLimit: number): any[][] {
```
pauldambra, line 169: if the list length is 2 and they are both bigger than the limit what does this do

src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -165,27 +166,154 @@ export function truncateLargeConsoleLogs(_event: eventWithTime) {
...
+    const size = estimateSize(list)
```
daibhin, line 170: We compute the size in this function but then always need it again in the calling functions. Two examples:
https://github.com/PostHog/posthog-js/pull/1307/files#diff-deb180b570d4a9e728cd83fd418e54a748f69492d07c9809924a3821ca4de4d5R187
https://github.com/PostHog/posthog-js/pull/1307/files#diff-deb180b570d4a9e728cd83fd418e54a748f69492d07c9809924a3821ca4de4d5R243

If estimating the size is an expensive operation perhaps we should look to return the value so it doesn't need to be recomputed

src/extensions/replay/sessionrecording-utils.ts
daibhin: Should we possibly try recursively slicing the chunks in https://github.com/PostHog/posthog-js/pull/1307/files#diff-deb180b570d4a9e728cd83fd418e54a748f69492d07c9809924a3821ca4de4d5R178 as we used to do in https://github.com/PostHog/posthog-js/pull/1307/files#diff-deb180b570d4a9e728cd83fd418e54a748f69492d07c9809924a3821ca4de4d5L172-L188

src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -165,27 +166,154 @@ export function truncateLargeConsoleLogs(_event: eventWithTime) {
...
+                    size: estimateSize(remove),
+                    data: [
+                        {
+                            type: INCREMENTAL_SNAPSHOT_EVENT_TYPE,
+                            data: {
+                                source: MUTATION_SOURCE_TYPE,
+                                adds: [],
+                                texts: [],
+                                attributes: [],
+                                removes: remove,
```
daibhin, line 252: Tiny tiny nit but I feel like `removes` should be at the top given it happens first and the order is so important here. Wouldn't like any opportunity for someone to mix that up given we don't really have any guarantees over the ordering here

src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -165,27 +166,154 @@ export function truncateLargeConsoleLogs(_event: eventWithTime) {
...
+                                adds: add,
+                                texts: [],
+                                attributes: [],
+                                removes: [],
+                            },
+                            timestamp: bufferedData.timestamp - alteration + timestampWiggleMarker++,
+                        },
+                    ],
+                    sessionId: buffer.sessionId,
+                    windowId: buffer.windowId,
```
daibhin, line 276: Might be an opportunity to pull some of this into a function. Feels like we do almost the same thing 4 times

src/extensions/replay/sessionrecording-utils.ts
pauldambra: so, i used to split in two, and then recursively split each half in two until every part was below the size limit

now we calculate how many chunks so that ever chunk is under the limit and then split into that many chunks

(recursion always feels too "clever" to me :))

src/extensions/replay/sessionrecording-utils.ts
pauldambra: i think this has either changed enough the comment doesn't apply any more, or I don't understand the comment 🙈 🤣

pauldambra:
OK, I am as sure as I can be from local testing that this works
But i'm just uncomfortable editing rrweb data because we've seen playback be intolerant to data edits in ways we didn't expect

am sleeping on this

src/extensions/replay/sessionrecording-utils.ts
pauldambra: good call here... 0.1% less bundle increase

src/extensions/replay/sessionrecording-utils.ts
daibhin: It's not a biggie, just noticed that we estimated the size while slicing the list but they re-estimate in https://github.com/PostHog/posthog-js/pull/1307/files#diff-deb180b570d4a9e728cd83fd418e54a748f69492d07c9809924a3821ca4de4d5R184-R186. Could possibly just return the already estimated size

daibhin (approved):
Looks good but agreed it's hard to reason about for all cases without some production data flowing.

Could you hook it up locally and test it out for some large payload sites? Otherwise maybe we need to figure out a way to roll it out slowing in production

PR #1309, opened 2024-07-19
fix: ignore ridonculously large web vitals values

pauldambra:
We see web vitals captured with values many times greater than one hour.

This is 99.999999% definitely a bug in the web vitals library or browsers.

Certainly it's not useful to know you have outliers where someone's FCP was 2 days.

Let's limit ourselves to a max of one hour, but let people override that if they really, really want to

### DIFF
--- a/src/extensions/web-vitals/index.ts
+++ b/src/extensions/web-vitals/index.ts
@@ -7,6 +7,7 @@ import { assignableWindow, window } from '../../utils/globals'
 import Config from '../../config'
 
 export const FLUSH_TO_CAPTURE_TIMEOUT_MILLISECONDS = 8000
+export const ONE_HOUR_IN_MILLIS = 60 * 60 * 1000
 const LOGGER_PREFIX = '[Web Vitals]'
 type WebVitalsEventBuffer = { url: string | undefined; metrics: any[]; firstMetricTimestamp: number | undefined }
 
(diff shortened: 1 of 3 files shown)


src/extensions/web-vitals/index.ts
```diff
@@ -152,6 +167,11 @@ export class WebVitalsAutocapture {
     private _startCapturing = () => {
         const { onLCP, onCLS, onFCP, onINP } = assignableWindow.postHogWebVitalsCallbacks
 
+        if (!onLCP || !onCLS || !onFCP || !onINP) {
+            logger.error(LOGGER_PREFIX + 'web vitals callbacks not loaded - not starting')
+            return
+        }
```
pauldambra, line 173: we sometimes see an error where this is undefined... 🤷

src/extensions/web-vitals/index.ts
```diff
@@ -116,6 +124,13 @@ export class WebVitalsAutocapture {
             return
         }
 
+        // we observe some very large values sometimes, we'll ignore them
+        // since the likelihood of LCP > 1 hour being correct is very low
+        if (this._applyMaxLimit && metric.value >= ONE_HOUR_IN_MILLIS) {
```
pauldambra, line 129: an hour is pretty massive tbh

daibhin (approved):
Happy to put this out there but I guess it begs the question... why not a 30 minute limit? Or 15? Both also seem unreasonable but I guess we have to cutoff somewhere if there's nothing else indicating that these are caused by a bug

pauldambra:
fair... i just picked a value lower than daily maximums i'd seen, but it's slightly neater to let people configure the value rather than have a boolean flag 👍

PR #1281, opened 2024-07-02
feat(surveys): lazy load survey event receiver

Phanatic:
## Changes

This change allows us to lazy load the SurveyEventReciever type only when Surveys are enabled for an application.

## Checklist
- [x] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [x] Accounted for the impact of any changes across different browsers
- [x] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

PR #1311, opened 2024-07-20
fix: don't load preact until you have to

pauldambra:
we need to keep the bundle as small as possible, it's already ~3x bigger than when people first start complaining about the size

https://github.com/PostHog/posthog-js/pull/1281/ is in merge hell because we abandoned it

this is another attempt to avoid everyone downloading preact, it reduces bundle size by 6%

### DIFF
--- a/src/__tests__/surveys.test.ts
+++ b/src/__tests__/surveys.test.ts
@@ -9,14 +9,18 @@ import {
     SurveyQuestionBranchingType,
     SurveyQuestion,
 } from '../posthog-surveys-types'
-import { getDisplayOrderChoices, getDisplayOrderQuestions } from '../extensions/surveys/surveys-utils'
+import {
+    canActivateRepeatedly,
+    getDisplayOrderChoices,
+    getDisplayOrderQuestions,
+} from '../extensions/surveys/surveys-utils'
 import { PostHogPersistence } from '../posthog-persistence'
 import { PostHog } from '../posthog-core'
 import { DecideResponse, PostHogConfig, Properties } from '../types'
 import { window } from '../utils/globals'
 import { RequestRouter } from '../utils/request-router'
 import { assignableWindow } from '../utils/globals'
-import { expectScriptToExist, expectScriptToNotExist } from './helpers/script-utils'
+import { generateSurveys } from '../extensions/surveys'
 
 describe('surveys', () => {
     let config: PostHogConfig
@@ -34,6 +38,7 @@ describe('surveys', () => {
             'enabled-internal-targeting-flag-key': true,
             'disabled-internal-targeting-flag-key': false,
         },
+        surveys: true,
     } as unknown as DecideResponse
 
     const firstSurveys: Survey[] = [
@@ -119,6 +124,16 @@ describe('surveys', () => {
     beforeEach(() => {
         surveysResponse = { surveys: firstSurveys }
 
+        const loadScriptMock = jest.fn()
+
+        loadScriptMock.mockImplementation((_path, callback) => {
+            assignableWindow.__PosthogExtensions__ = assignableWindow.__Posthog__ || {}
+            assignableWindow.extendPostHogWithSurveys = generateSurveys
+            assignableWindow.__PosthogExtensions__.canActivateRepeatedly = canActivateRepeatedly
+
+            callback()
+        })
+
         config = {
             token: 'testtoken',
             api_host: 'https://app.posthog.com',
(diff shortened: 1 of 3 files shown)


src/__tests__/surveys.test.ts
```diff
@@ -147,7 +162,12 @@ describe('surveys', () => {
             },
         } as unknown as PostHog
```
pauldambra, line 163: to keep this PR smaller, i'm not replacing the mock posthog here
(we really should be, but it can wait)

src/__tests__/surveys.test.ts
```diff
@@ -757,33 +779,6 @@ describe('surveys', () => {
         })
     })
 
-    describe('decide response', () => {
```
pauldambra, line 760: since we're mocking load script we don't need to test whether or not it behaves correctly...

src/entrypoints/surveys.ts
```diff
@@ -1,8 +1,11 @@
 import { generateSurveys } from '../extensions/surveys'
 
 import { window } from '../utils/globals'
+import { canActivateRepeatedly } from '../extensions/surveys/surveys-utils'
 
 if (window) {
+    ;(window as any).__PosthogExtensions__ = (window as any).__PosthogExtensions__ || {}
+    ;(window as any).__PosthogExtensions__.canActivateRepeatedly = canActivateRepeatedly
```
pauldambra, line 8: maybe this should even be __PosthogExtensions__.__Surveys__ but 🤷 

but regardless this is the mechanism we use to push code out of the main bundle so that folk who don't use surveys (or replay or web vitals etc etc) don't have to download the code for that feature

accessing `canActivateRepeatedly` causes the bundler to pull in preact and you only need that with surveys enabled

src/posthog-surveys.ts
```diff
@@ -12,7 +12,9 @@ import { SurveyEventReceiver } from './utils/survey-event-receiver'
 import { assignableWindow, document, window } from './utils/globals'
 import { DecideResponse } from './types'
 import { logger } from './utils/logger'
-import { canActivateRepeatedly } from './extensions/surveys/surveys-utils'
+import { isNullish } from './utils/type-utils'
+
+const LOGGER_PREFIX = '[Surveys]'
```
pauldambra, line 17: QOL we put a prefix on logs so that they are easier to read

src/posthog-surveys.ts
```diff
@@ -50,15 +52,14 @@ function getRatingBucketForResponseValue(responseValue: number, scale: number) {
 }
 
 export class PostHogSurveys {
-    instance: PostHog
     private _decideServerResponse?: boolean
     public _surveyEventReceiver: SurveyEventReceiver | null
 
-    constructor(instance: PostHog) {
-        this.instance = instance
+    constructor(private readonly instance: PostHog) {
```
pauldambra, line 58: QOL: TS lets you declare a field like this when it's a constructor parameter

src/posthog-surveys.ts
```diff
@@ -178,7 +178,7 @@ export class PostHogSurveys {
                 const eventBasedTargetingFlagCheck =
                     hasEvents || hasActions ? activatedSurveys?.includes(survey.id) : true
 
-                const overrideInternalTargetingFlagCheck = canActivateRepeatedly(survey)
+                const overrideInternalTargetingFlagCheck = this._canActivateRepeatedly(survey)
```
pauldambra, line 181: this single line effectively reduces bundle size by 6%

src/posthog-surveys.ts
```diff
@@ -258,7 +258,15 @@ export class PostHogSurveys {
             return nextQuestionIndex
         }
 
-        console.warn('Falling back to next question index due to unexpected branching type') // eslint-disable-line no-console
```
pauldambra, line 261: try not to disable eslint rules, here it was warning you not to use console and it was IMHO right 😊

src/posthog-surveys.ts
```diff
@@ -258,7 +258,15 @@ export class PostHogSurveys {
...
         return nextQuestionIndex
     }
+
+    // this method is lazily loaded onto the window to avoid loading preact and other dependencies if surveys is not enabled
+    private _canActivateRepeatedly(survey: Survey) {
+        if (isNullish(assignableWindow.__PosthogExtensions__.canActivateRepeatedly)) {
+            logger.warn(LOGGER_PREFIX, 'canActivateRepeatedly is not defined, must init before calling')
+        }
+        return assignableWindow.__PosthogExtensions__.canActivateRepeatedly(survey)
+    }
```
pauldambra, line 271: convenience method for accessing the lazily loaded canActivateRepeatedly

dmarticus (approved):
This change seems reasonable to me, thanks for keeping at it and making the PR!  @pauldambra, Is there anything else you needed from the Feature Success team other than a second pair of eyes on the review?  I kicked CI again and it passed this time so all seems good from my perspective.

PR #1313, opened 2024-07-22
feat: Implement referring domain denylist

robbie-c:
## Changes
Add a config option where people can pass domains (as string or regex) that posthog will ignore when it comes to setting the referring domain.

When the referrer is an ignored domain, we don't update the stored referrer. When the very first referrer is an ignored domain, we treat it as direct traffic.

For backwards compatibility, the list of ignored domains is empty by default. We may want to consider always adding the current domain to this list, but this is out of scope for this PR.

## Checklist

### DIFF
--- a/src/__tests__/posthog-core.test.ts
+++ b/src/__tests__/posthog-core.test.ts
@@ -1,16 +1,48 @@
-import { PostHogConfig } from '../types'
-import { uuidv7 } from '../uuidv7'
 import { defaultPostHog } from './helpers/posthog-instance'
+import type { PostHogConfig } from '../types'
+import { uuidv7 } from '../uuidv7'
+
+const mockReferrerGetter = jest.fn()
+const mockURLGetter = jest.fn()
+jest.mock('../utils/globals', () => {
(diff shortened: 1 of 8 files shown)


src/__tests__/posthog-core.test.ts
```diff
@@ -52,21 +84,113 @@ describe('posthog core', () => {
...
                 onCapture.mockClear()
                 ;(console.error as any).mockClear()
-                posthog.capture(eventName, properties)
-                expect(onCapture.mock.calls).toEqual([])
-                expect(console.error).toHaveBeenCalledTimes(1)
+                for (let i = 0; i < 50; i++) {
+                    posthog.capture(eventName, eventProperties)
+                }
+                expect(onCapture).toHaveBeenCalledTimes(1)
+                expect(onCapture.mock.calls[0][0]).toBe('$$client_ingestion_warning')
```
robbie-c, line 97: @benjackwhite this test changed because it fails (on main) if you run it in isolation. The change to test setup meant that it was no longer using state from the previous test.

PR #1314, opened 2024-07-22
fix: Fix the referrer changing mid-session

robbie-c:
## Changes

As part of the personless event changes, we accidentally introduced a bug where the referrer could change mid-session. This shouldn't happen, and led to some annoying situation where e.g. if a payments workflow navigated people to checkout.stripe.com and then back to the original site, the referrer would switch to stripe, which meant that any events after that couldn't be attributed to the original referrer.

This fixes that bug, and adds some tests.

(P.S. I realised how this was broken when working on https://github.com/PostHog/posthog-js/pull/1313, IMO with this fix merged that PR becomes a lot less useful)

## Checklist

### DIFF
--- a/src/__tests__/posthog-core.test.ts
+++ b/src/__tests__/posthog-core.test.ts
@@ -108,7 +108,6 @@ describe('posthog core', () => {
                 // arrange
                 const token = uuidv7()
                 mockReferrerGetter.mockReturnValue('https://referrer.example.com/some/path')
-                mockURLGetter.mockReturnValue('')
                 const { posthog, onCapture } = setup({
                     token,
                     persistence_name: token,
@@ -129,7 +128,6 @@ describe('posthog core', () => {
                 // arrange
                 const token = uuidv7()
                 mockReferrerGetter.mockReturnValue('https://referrer1.example.com/some/path')
-                mockURLGetter.mockReturnValue('')
                 const { posthog: posthog1 } = setup({
                     token,
                     persistence_name: token,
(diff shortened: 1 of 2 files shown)


src/__tests__/posthog-core.test.ts
```diff
@@ -52,21 +84,108 @@ describe('posthog core', () => {
...
                 onCapture.mockClear()
                 ;(console.error as any).mockClear()
-                posthog.capture(eventName, properties)
-                expect(onCapture.mock.calls).toEqual([])
-                expect(console.error).toHaveBeenCalledTimes(1)
+                for (let i = 0; i < 50; i++) {
+                    posthog.capture(eventName, eventProperties)
+                }
+                expect(onCapture).toHaveBeenCalledTimes(1)
+                expect(onCapture.mock.calls[0][0]).toBe('$$client_ingestion_warning')
```
robbie-c, line 97: This test was broken before, it relied on state from the previous test and failed when run in isolation.

src/__tests__/posthog-core.test.ts
```diff
@@ -52,21 +84,108 @@ describe('posthog core', () => {
...
+
+                // assert
+                const { $set_once, properties } = onCapture.mock.calls[0][1]
+                expect($set_once['$initial_referrer']).toBe('https://referrer.example.com/some/path')
+                expect($set_once['$initial_referring_domain']).toBe('referrer.example.com')
+                expect(properties['$referrer']).toBe('https://referrer.example.com/some/path')
+                expect(properties['$referring_domain']).toBe('referrer.example.com')
+            })
+
+            it('should not update the referrer within the same session', () => {
```
robbie-c, line 128: This test fails without the fix in this PR

src/posthog-persistence.ts
```diff
@@ -227,7 +227,7 @@ export class PostHogPersistence {
     }
 
     update_referrer_info(): void {
-        this.register(Info.referrerInfo())
+        this.register_once(Info.referrerInfo(), undefined)
```
robbie-c, line 230: The one-line actual fix

benjackwhite (approved):

PR #1321, opened 2024-07-25
chore: finally remove given

pauldambra:
This moves the final files using given to not using given

Which means they can be converted to typescript

I've refactored a little but tried to keep the changes pretty minimal just to get given removed and the tests moved to TS


daibhin (approved):

daibhin:
Going to merge this because Paul is out for the week and don't want him to come back to a bunch of merge conflicts

PR #1328, opened 2024-07-29
feat(flags): give users the ability to suppress override warning logs, if desired

dmarticus:
## Changes

We've supported manually overriding feature flag values in the client for years (and our toolbar [natively supports it](https://posthog.com/docs/toolbar/override-feature-flags)), but since this override doesn't persist to the backend, we've historically always just treated this feature as a debug feature only.  However, there's been a influx of support requests (on both Twitter and in our support inbox) that are asking about this feature, and one user explicitly mentioned that the feature [would be perfect for his use case if it weren't for the debug warning and logs](https://posthoghelp.zendesk.com/agent/tickets/15724).  I started a [discussion about this in Slack](https://posthog.slack.com/archives/C034XD440RK/p1722281920032039), and if we decide that it's safe to give users the ability to remove these logs, this PR enables it.

## Checklist

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -293,22 +293,25 @@ export class PostHogFeatureFlags {
     }
 
     /*
-     * Override feature flags for debugging.
+     * Override feature flags on the client-side.  Useful for setting non-persistent feature flags, or for testing/debugging
+     * feature flags in the PostHog app.
      *
      * ### Usage:
      *
      *     - posthog.feature_flags.override(false)
      *     - posthog.feature_flags.override(['beta-feature'])
-     *     - posthog.feature_flags.override({'beta-feature': 'variant', 'other-feature': True})
+     *     - posthog.feature_flags.override({'beta-feature': 'variant', 'other-feature': true})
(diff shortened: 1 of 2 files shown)


src/posthog-featureflags.ts
```diff
@@ -293,22 +293,25 @@ export class PostHogFeatureFlags {
...
+     * @param {boolean} [suppressWarning=false] Optional parameter to suppress the override warning.
      */
-    override(flags: boolean | string[] | Record<string, string | boolean>): void {
+    override(flags: boolean | string[] | Record<string, string | boolean>, suppressWarning: boolean = false): void {
         if (!this.instance.__loaded || !this.instance.persistence) {
             return logger.uninitializedWarning('posthog.feature_flags.override')
         }
 
-        this._override_warning = false
+        this._override_warning = suppressWarning
```
Phanatic, line 314: this `_override_warning` is confusing because for a value of `false`, we log the debug warning. 
https://github.com/PostHog/posthog-js/blob/25713e9d26b43bce0d6c246999287d82edc115f2/src/posthog-featureflags.ts#L115-L122
Good fix here.

Phanatic (approved):
Ship it

PR #1331, opened 2024-07-30
fix(surveys): Accept eventPayload in onEvent handler

Phanatic:
## Changes

This is a bug fix for the scenario where we weren't removing a survey from the activated surveys list once the survey is shown.

## Checklist


dmarticus (approved):
Rip it!!

PR #1340, opened 2024-08-05
fix(surveys): textarea border box was being set in the preview but not in the surveys rendered on user websites

dmarticus:
## Changes

Issue identified here: https://posthog.slack.com/archives/C074A08LP1B/p1722845257000019?thread_ts=1722508627.577449&cid=C074A08LP1B

This issue came from me trying to fix the textarea size for free response surveys.  My fix made the survey preview rendering work correctly, but it was too wide for the popover rendering (which is more important; that's the thing the end users see!)

Before: 

image

and

Screenshot 2024-08-05 at 7 31 55 AM

I fixed this to add an explicit border box into the textarea.  In the preview, it was inheriting a border box already, so this wasn't a problem, but since the rendered survey has a different DOM tree, with different stylings, it didn't work for end users.

This fix should work for both.

After (on page):

Screenshot 2024-08-05 at 7 47 27 AM

After (in preview section):

image

### DIFF
--- a/src/extensions/surveys/surveys-utils.tsx
+++ b/src/extensions/surveys/surveys-utils.tsx
@@ -67,6 +67,7 @@ export const style = (appearance: SurveyAppearance | null) => {
               border-color: ${appearance?.borderColor || '#c9c6c6'};
               margin-top: 14px;
               width: 100%;
+              box-sizing: border-box;
           }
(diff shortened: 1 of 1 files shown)


Phanatic (approved):
Ship it

PR #1345, opened 2024-08-07
chore: make password capture more explicit

daibhin:
## Changes

Since https://github.com/PostHog/posthog-js/pull/1334 we mask passwords by default.

I want to take this a step further and always include password input masking unless it is explicitly opted out using the `dangerouslyCapturePasswordInputs`. While I can't see any reasonably good reason that you'd want to unmask a password input, we should still allow it if someone is certain that's what they want to do. It will be slightly more difficult / confusing to configure but that feels like a worthy trade-off to avoid customers accidentally unmasking password inputs by leaving the `password: true` field out of their `maskInputOptions` config


marandaneto:
On Mobile, for certain types of views such as for sure containing PII data (password/email, etc), it's not possible to disable the masking, see https://posthog.com/docs/session-replay/ios#masking-in-swiftui
Even if people allow it, it's still "us" leaking the data hence no opt-out, it's too risky and I understand that I may have to relax that in the future, so far nobody asked for it, only on how to mask it, which is enabled by default.

marandaneto:
If people can do `maskInputOptions: {password: false},` already, `dangerouslyCapturePasswordInputs` feels like just making the API more complicated/degrading UX, I'd rather document this better if needed, not a blocker, but happy to hear another opinion.

src/extensions/replay/sessionrecording.ts
```diff
@@ -600,10 +600,23 @@ export class SessionRecording {
...
+                sessionRecordingOptions.maskInputOptions = {
+                    ...sessionRecordingOptions.maskInputOptions,
+                    password: !value,
+                }
+            } else if (key in sessionRecordingOptions) {
+                if (key === 'maskInputOptions') {
+                    sessionRecordingOptions.maskInputOptions = {
+                        ...value,
+                        // never use user supplied value
+                        password: sessionRecordingOptions.maskInputOptions?.password,
```
benjackwhite, line 613: ? Isn't this always using it??

benjackwhite (changes requested):

src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -69,6 +69,7 @@ export declare type recordOptions<T> = {
     maskTextSelector?: string
     maskAllInputs?: boolean
     maskInputOptions?: MaskInputOptions
+    dangerouslyCapturePasswordInputs?: boolean
```
benjackwhite, line 72: tbh I think this is overkill... I don't see how the issue is the existing naming. The issue is that the default was to not mask them whereas now the default is always to mask them and you have to explicitly mark it as false? 

I think we should instead leave it as it was but explicitly check the object and mark it as "true" if it is anything other than explicitly "false"

benjackwhite (changes requested):

src/extensions/replay/sessionrecording-utils.ts
daibhin: Fair point. I've fixed things up to ensure a `password` value is always present. If not set we ensure the default of _true_ is used

marandaneto (approved):

benjackwhite (approved):

PR #1348, opened 2024-08-09
feat: Add prev pageview duration

robbie-c:
## Changes

Add prev pageview duration, for a customer, inspired by this slack thread https://posthog.slack.com/archives/C05LJK1N3CP/p1723144788366199

We already do have some logic for measuring the time between the same event (that gets stored in storage rather than in memory), but that doesn't add e.g. pageview duration to a pageleave event, and doesn't match up with the other prev_pageview props like prev_pageview_pathname, which the paths view on web analytics already uses.

## Checklist

### DIFF
--- a/src/page-view.ts
+++ b/src/page-view.ts
@@ -44,6 +44,11 @@ export class PageViewManager {
         const previousTimestamp = this._prevPageviewTimestamp
         const scrollContext = this._instance.scrollManager.getContext()
 
+        if (!previousTimestamp) {
+            // this means there was no previous pageview
+            return {}
+        }
+
         let properties: PageViewEventProperties = {}
         if (scrollContext) {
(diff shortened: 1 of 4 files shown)


src/page-view.ts
```diff
@@ -16,72 +17,82 @@ interface PageViewEventProperties {
...
     }
 
-    doPageLeave(): PageViewEventProperties {
-        return this._previousScrollProperties()
+    doPageLeave(timestamp: Date): PageViewEventProperties {
+        return this._previousPageViewProperties(timestamp)
     }
 
-    private _previousScrollProperties(): PageViewEventProperties {
+    private _previousPageViewProperties(timestamp: Date): PageViewEventProperties {
```
robbie-c, line 42: Probably easier to view this diff without whitespace changes

daibhin (approved):

PR #1350, opened 2024-08-09
fix(segment): Allow overriding user id in Segment integration

Phanatic:
## Changes

This PR changes the way we identify users in the Segment Integration for PostHog.
https://github.com/PostHog/posthog-js/blob/d5889f0d97a68d9a6b964a24207d631e0834e46e/src/extensions/segment-integration.ts#L119-L126

As we can see in the above block, we set the `USER_STATE` to `identified` ONLY if the user is identified by Segment. 
This can cause issues for feature flags since they're not persisted if the USER_STATE is `anonymous` causing all sorts of confusing behavior for our customers.

This PR makes a change to the Segment integration to allow us to set the `USER_STATE` to `identified` even if the segment user is unidentified, callers can explicitly set the `user_id` used by segment without identifying the user in segment

## Checklist

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -188,6 +188,9 @@ export interface PostHogConfig {
     advanced_disable_toolbar_metrics: boolean
     feature_flag_request_timeout_ms: number
     get_device_id: (uuid: string) => string
+    segment_config?: {
+        user_id?: string
+    }
     name: string
     _onCapture: (eventName: string, eventData: CaptureResult) => void
     capture_performance?: boolean | PerformanceCaptureConfig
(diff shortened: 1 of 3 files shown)


src/types.ts
```diff
@@ -188,6 +188,9 @@ export interface PostHogConfig {
     advanced_disable_toolbar_metrics: boolean
     feature_flag_request_timeout_ms: number
     get_device_id: (uuid: string) => string
+    segment_config?: {
+        user_id?: string
+    }
```
dmarticus, line 193: how does this value get set?  Like I see how we're reading it in`segment-integration`, but if we're adding this new payload to the `PostHogConfig`, how does the value appear here in the first place?

Phanatic: This should be clarified in 2ec3981

dmarticus (approved):

Issue #1326, opened 2024-07-29
React SDK - Session Recorder - Original fetch args is modified causing application to crash


samuellawerentz:
## Bug description
We use the React SDK to initialize posthog. We have a `fetch` interceptor which modifies the request before the API call is made. 

After integrating with posthog, the `fetch` API is already modified and the args are not passed properly, because of this our interceptor is not able to function properly. The issue is with the `recorder.js` file.

TLDR; you are modifying the original fetch and not forwarding all the args in this format `...args` . And hence our fetch interceptor which expects two args, receive only one causing the app to crash.

## How to reproduce

1. Load react posthog
2. modify the fetch API like `window.fetch = ...`
3. App crashes

Please let me know, how to resolve this or what is the next step.

I'm attaching a screenshot of the `recorder.js` file for reference, where you are overriding the default fetch.
<img width="1428" alt="FreezeFrame 2024-07-16 at 9 29 28 PM" src="https://github.com/user-attachments/assets/6a2cf5c4-35a5-4faf-8408-662964e4a1c2">

I have already sent an email, but was not able to track it.

pauldambra:
<img width="365" alt="Screenshot 2024-07-29 at 17 47 05" src="https://github.com/user-attachments/assets/226a8920-02bd-4c0a-9ce4-477e86c1177e">

@samuellawerentz if I understand you correctly changing to using these two args would work for you?

(we'd have to test that doesn't have any other unexpected effect but it seems reasonable - cc @daibhin)

samuellawerentz:
This is our code
```js
const originalFetch = window.fetch

export const fetchInterceptor = () => {
  window.fetch = async (...args) => {
    const href = decodeURI(args[0]) // <- It breaks around here, as we are expecting a string as the first arg

    // Replace region with the appropriate region from the state
    if (href?.includes('{{region}}')) {
      ... some logic ...
      args[0] = new URL(href.replace('region', region)
    }

    const response = await originalFetch(...args)
    return response
  }
}

```

samuellawerentz:
Yes, the first arg we expect it to be URL

pauldambra:
Ah... "It breaks around here, as we are expecting a string as the first arg"

the first arg of fetch is not guaranteed to be a string

from: https://developer.mozilla.org/en-US/docs/Web/API/Window/fetch#resource
[...]

PR #1351, opened 2024-08-10
fix: pass original fetch args along

pauldambra:
When wrapping fetch we take a url, init pair and create a request object

When we pass on and call the original fetch method we pass the request object. this is _technically_ equivalent to the original but not literally equivalent which can cause problems if there are further wrappers on fetch

see: https://github.com/PostHog/posthog-js/issues/1326

We actively don't want to alter the fetch details so it should be 100% safe to use the original args when calling the original fetch 🤷

### DIFF
--- a/src/entrypoints/recorder.ts
+++ b/src/entrypoints/recorder.ts
@@ -507,7 +507,7 @@ function initFetchObserver(
                 }
 
                 after = win.performance.now()
-                res = await originalFetch(req)
+                res = await originalFetch(url, init)
                 before = win.performance.now()
 
                 const responseHeaders: Headers = {}


daibhin (approved):
Change makes sense. I think we should be returning exactly the arguments we were given. Probably an oversight earlier on to have returned the constructed Request so worth fixing forward.

There's a risk we break this the other way e.g. customers have come to rely on the output of our fetch returning a `Request` object. We won't really know that without shipping.

My suggestion would be to get this out there as a minor bump. At least then people are more likely to be able to track the version bump if it does cause an issue for them. Secondly, there's always the escape hatch of staying on the previous version until they've had a chance to fix their reliant code (which should be trivial either way).

The only other alternative I can think of is passing `res = await originalFetch(req, init)` back but that feels less good. We'd be preserving current functionality at the expense of correctness. Given the points above it's probably just worth fixing everything now and dealing with any consequential issues that arise.

weilandia:
Our site went down tonight because of an issue with Posthog overwriting fetch in this file. Not sure that this it eh culprit, but the changes made to window.fetch made RTKQuery unusable and left us with a very confusing FE bug that came down to the Posthog assets loading and breaking fetch.

pauldambra:
oh yuck! so sorry @weilandia! will revert this change I've seen a few reports of similar and I'd be amazed if it was any of the other changes yesterday

weilandia:
Thanks!

PR #1353, opened 2024-08-12
chore: limit buffer splitting to rate limit

daibhin:
## Changes

...

## Checklist

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -662,7 +662,7 @@ export class SessionRecording {
                     this._tryAddCustomEvent('$pageview', { href })
                 }
             } catch (e) {
-                logger.error('Could not add $pageview to rrweb session', e)
+                logger.error(LOGGER_PREFIX + ' ' + 'Could not add $pageview to rrweb session', e)
             }
         })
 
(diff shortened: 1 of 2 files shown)


daibhin:
@pauldambra curious to get your thoughts on this one (still have tests to fix etc). Problem was that a customer was being clientside rate limited because we were splitting messages into more parts than their rate limit allowed. Proposal was to have two limits on the split method (a target size but that does not allow splitting into more parts than the rate limit allows)

src/extensions/replay/sessionrecording.ts
```diff
@@ -863,7 +863,11 @@ export class SessionRecording {
         }
 
         if (this.buffer.data.length > 0) {
-            const snapshotEvents = splitBuffer(this.buffer)
+            const captureEventsPerSecond = this.instance.rateLimiter.captureEventsPerSecond
+            const snapshotEvents = splitBuffer(this.buffer, captureEventsPerSecond)
+            if (snapshotEvents.length >= captureEventsPerSecond) {
```
pauldambra, line 868: 🤯 oh... knowing that recording is sending more events than, erm , events. We should either not limit them client side or have a much higher limit 🤔

daibhin:
Went a different approach in https://github.com/PostHog/posthog-js/pull/1383

PR #1355, opened 2024-08-13
feat: default to gzip compression instead of base64

frankh:
## Changes
it's weird that we do our first requests before the decide response with base64 "compression", as

1. we've supported gzip for literally years, so there's basically zero risk of somehow hitting an endpoint that doesn't support it
2. base64 is not compression, it increases the payload size compared to if we did nothing at all
3. rust capture does not support base64

## Checklist

### DIFF
--- a/functional_tests/mock-server.ts
+++ b/functional_tests/mock-server.ts
@@ -4,6 +4,7 @@ import { ResponseComposition, rest } from 'msw'
 import { setupServer } from 'msw/lib/node'
 import { RestContext } from 'msw'
 import { RestRequest } from 'msw'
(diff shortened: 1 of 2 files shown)


benjackwhite (approved):
Will be interesting if no tests fail...

frankh:
had to fix the mock parsing but looks good now

webjunkie:
Curious: What is the reason we don't use headers/http level compression?

benjackwhite:
> Curious: What is the reason we don't use headers/http level compression?

You can't for requests. (You can't know what the server accepts before sending the request). Thats why you tell the server what you accept with headers like `accept-encoding`

PR #1358, opened 2024-08-13
chore: Run ts on test files

robbie-c:
## Changes
We're not running typescript type-checking on our test files.

In the past I've found this is pretty mixed but overall positive, on the one hand you make refactoring easier and end up catching problems you might not have caught, but on the other hand your test files end up with more type coercion which is quite noisy.

Here's my attempt at adding it to posthog-js. I created a new tsconfig.json for the test files which is less strict. If I add strict:true to the test tsconfig, there's another 309 errors left to fix - let's define this at out of scope, it's still a huge improvement in type safety without it

## Checklist


src/__tests__/surveys.test.ts
```diff
@@ -614,11 +615,12 @@ describe('surveys', () => {
         })
 
         it('returns event based surveys that observed an event', () => {
+            // TODO this test fails when run in isolation
```
robbie-c, line 618: @Phanatic Can you take a look at this? (separate to this PR)

benjackwhite (approved):
I think this is totally worth it

PR #1363, opened 2024-08-16
feat: support specific endpoint for flags

raquelmsmith:
## Problem

Ad blocker people go to posthog.com to figure our what our tracking URLs are. If we use the same domain for event tracking and flags, when they determine whatever URL we are using (either default or proxied) and add it to the ad blocker, our website functionality breaks down because flags are blocked.

I think having two domains could solve this. `tracking.whatever-url.com` and `feature-flags-do-not-block.whatever-url.com`. I would honestly consider naming the subdomain exactly that so that it is very clear.

If they end up blocking our tracking URL that is totally fine - that's part of the reason why people use ad blockers in the first place. So this will allow ad blockers to block our default ingestion endpoints, but will _hopefully_ prevent them from blocking decide requests, which are required for website functionality.

## Changes

Adds a new config option `flags_api_host` that is used for any requests that hit `/decide`. If not provided, uses the default `api_host`. 

Also uses this secondary flags_api_host, if set, to get `static/array.js` so we can make the decide calls.

...

## Checklist
    - I don't think there should be an impact?
    - All existing tests still pass.

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -117,6 +117,7 @@ let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 &&
 
 export const defaultConfig = (): PostHogConfig => ({
     api_host: 'https://us.i.posthog.com',
+    flags_api_host: undefined,
     ui_host: null,
     token: '',
     autocapture: true,
@@ -899,6 +900,7 @@ export class PostHog {
 
         if (this.requestRouter.region === RequestRouterRegion.CUSTOM) {
             properties['$lib_custom_api_host'] = this.config.api_host
+            properties['$lib_custom_flags_api_host'] = this.config.flags_api_host
         }
 
         if (
--- a/src/utils/request-router.ts
+++ b/src/utils/request-router.ts
@@ -66,7 +66,7 @@ export class RequestRouter {
             return (this.uiHost || this.apiHost.replace(`.${ingestionDomain}`, '.posthog.com')) + path
         }
 
-        const useFlagsAPIHost = (path.startsWith(flagsPath) || path.startsWith(libPath)) && this.flagsApiHost
+        const useFlagsAPIHost = (path.indexOf(flagsPath) === 0 || path.indexOf(libPath) === 0) && this.flagsApiHost
         if (this.region === RequestRouterRegion.CUSTOM) {
             return useFlagsAPIHost ? this.flagsApiHost + path : this.apiHost + path
         }
--- a/src/utils/request-router.ts
+++ b/src/utils/request-router.ts
@@ -18,7 +18,6 @@ export type RequestRouterTarget = 'api' | 'ui' | 'assets'
 
 const ingestionDomain = 'i.posthog.com'
 const flagsPath = '/decide'
-const libPath = '/static/array'
 
 export class RequestRouter {
     instance: PostHog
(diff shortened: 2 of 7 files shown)


src/posthog-core.ts
```diff
@@ -117,6 +117,7 @@ let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 &&
 
 export const defaultConfig = (): PostHogConfig => ({
     api_host: 'https://us.i.posthog.com',
+    flags_api_host: undefined,
```
benjackwhite, line 120: Not sure how I feel about `flags_api_host`. Decide is much more than flags.

Also more generally I don't see adblockers caring enough to only block the other domains and not this. We're entering a whack-a-mole situation for sure.

Also also will this even help?? The api_host is what is used to load the JS from the snippet which will be blocked. If the snippet doesn't load then there is no benefit to having a separate flags host as there is no code to call it anways 😓 

I realise I'm proposing problems without solutions. I'm just not super keen on this solution that adds more config for a user for potentially no real gains.

benjackwhite: The best solution I can think of is having the tracker be on a subpath of the same domain rather than a separate domain seeing as typically adblockers will allow the domain if it is the same as the webpage you are on.

benjackwhite (changes requested):

src/posthog-core.ts
raquelmsmith: I included `static/array.js` in the filter, so that would go through the (hopefully) non-blocking domain as well. 

I don't think ad blockers would be looking specifically for array.js? I guessing what they look for is any URL that sends out lots of info when someone is on a page - that's our `/e/` and `/s/` paths. Those will all still go through the main `api_host` url, which is fine if an ad blocker blocks it (which is partially what an ad blocker is for anyway). 

`/decide` and `static/array.js` are NOT tracking endpoints so hopefully people aren't looking specifically for those and blocking them when found. If we name our endpoint something silly (`feature-flags-do-not-block.posthog.com`) then hopefully they will (correctly) ignore that domain traffic. I'm also down for putting this on a completely separate domain if we want to be more careful.

src/utils/request-router.ts
```diff
@@ -17,6 +17,8 @@ export enum RequestRouterRegion {
 export type RequestRouterTarget = 'api' | 'ui' | 'assets'
 
 const ingestionDomain = 'i.posthog.com'
+const flagsPath = '/decide'
+const libPath = '/static/array'
```
benjackwhite, line 21: This makes no sense. _this code_ lives in `array.js`. It's only the snippet that loads array.js (the snippet has no idea of any of this code).

This is why I'm generally not in favour of this approach - the snippet is always loaded via the standard endpoint. If that gets blocked then everything is blocked anways.

If you revert the arrayjs stuff then it is more mergeable and at least is an option for people avoiding the snippet (i.e. npm install) but even then I still think this is incredibly unlikely to help...

raquelmsmith: Oh hmmmmm okay. Sorry I have no idea how posthog-js works but I'm learning... I'll think about it.

raquelmsmith: Thoughts on having the snippet itself use this config to get array.js?

raquelmsmith: I think a very small change to the snippet will make it use the `flags_api_host` instead of the standard `api_host` - if the flags host is provided. Basically this part of the snippet:

```
p.src=(s.flags_api_host?s.flags_api_host:s.api_host)+"/static/array.js"
```

(it's somewhat verbose as the [Nullish coalescing operator (`??`)](https://caniuse.com/mdn-javascript_operators_nullish_coalescing) is not IE11 compatible. (when do we stop supporting IE11???????)

I think that's what would fetch array.js initially, and if it's a non-blocked host then it'll get through. 

I think we do still want to do this. If someone wants to use our feature previews, for instance, but while using a tracking/ad blocker, then they can't.

PR #1369, opened 2024-08-19
chore: Move chrome tests off of browserstack, add firefox

robbie-c:
## Changes

We can run chrome on the actions runner without needing to use browserstack. This means that we only use 2 of our 5 browserstack concurrent tests per PR rather than 3, which lets us have 2 PRs running tests in parallel.

I also added headless firefox because why not.

## Checklist


pauldambra (approved):
🦾 love it

(can make the firefox job required after this too?)

robbie-c:
I don't have powers to do that but if you do then please go ahead :)

PR #1370, opened 2024-08-19
feat: default api host to new ingestion domain

raquelmsmith:
## Changes

We changed our ingestion endpoint to `us/eu.i.posthog.com`, but many people still have configs with the api_host set to app.posthog.com. If these people tried to use the toolbar, they'd get an authentication error like [this person](https://posthoghelp.zendesk.com/agent/tickets/15339) (internal link).

This PR makes sure that anyone with their api_host set to `app.posthog.com` gets their UI host properly set to `us.posthog.com`. Also makes sure it still works for EU with the newer eu.i.posthog.com ingestion endpoint.

...

## Checklist


zlwaterfield (approved):

zlwaterfield:
Thanks for doing this!

src/utils/request-router.ts
```diff
@@ -27,14 +27,25 @@ export class RequestRouter {
...
+        let host = this.instance.config.ui_host?.replace(/\/$/, '')
+
+        if (!host) {
+            // No ui_host set, get it from the api_host. But api_host differs
+            // from the actual UI host, so replace the ingestion subdomain with just posthog.com
+            host = this.apiHost.replace(`.${ingestionDomain}`, '.posthog.com')
+        }
 
         if (host === 'https://app.posthog.com') {
             return 'https://us.posthog.com'
```
raquelmsmith, line 46: this is a fall-back if they have their ui_host set to app.posthog.com for some reason

PR #1371, opened 2024-08-20
revert: "fix: pass original fetch args along"

pauldambra:
Reverts PostHog/posthog-js#1351

we've seen multiple reports of fetch wrapping causing errors, it would be amazing if it wasn't this change causing it.

it's safe to revert if it's not this change and important to revert if it is...

### DIFF
--- a/src/entrypoints/recorder.ts
+++ b/src/entrypoints/recorder.ts
@@ -507,7 +507,7 @@ function initFetchObserver(
                 }
 
                 after = win.performance.now()
-                res = await originalFetch(url, init)
+                res = await originalFetch(req)
                 before = win.performance.now()
 
                 const responseHeaders: Headers = {}


neilkakkar (approved):

daibhin (approved):

ricardobeat:
The same error has been fixed in dd-trace a while ago by calling fetch with the original input `Request` object: https://github.com/DataDog/dd-trace-js/pull/4258/files

pauldambra:
@ricardobeat interestingly that's a fix in DD for a different issue. their thread mentions it is difficult to test for but I'll see if i can poke about and recreate for us since we might be open to the same error as well (although haven't had it reported)

ricardobeat:
@pauldambra the core issue is the same, wrapping fetch incorrectly. I meant to point to a potentially better solution vs the one at https://github.com/PostHog/posthog-js/pull/1351/commits/24979cb0a97f86acd247a3027a61ffb2bf74f055 which has been reverted, preserving a `Request` object if received is more reliable.

PR #1372, opened 2024-08-20
fix: capture a $pageview event on opting in

skoob13:
## Changes

The initial `$pageview` is not captured when customers use the opt in/out flow. It creates a discrepancy in event count like [this](https://posthog.slack.com/archives/C0113360FFV/p1723716455610139) one. The PR changes the behavior to capture the initial `$pageview` event after a user opts in if the event hasn't yet been sent.

**Before**

Screenshot 2024-08-20 at 18 01 59

**After**

Screenshot 2024-08-20 at 18 12 59

## Checklist


src/posthog-core.ts
```diff
@@ -1977,12 +1980,15 @@ export class PostHog {
...
+        // Don't capture if captureEventName is null or false
+        if (isUndefined(options?.captureEventName) || options?.captureEventName) {
+            this.capture(options?.captureEventName ?? '$opt_in', options?.captureProperties, { send_instantly: true })
         }
 
-        this.capture(options?.captureEventName ?? '$opt_in', options?.captureProperties, { send_instantly: true })
+        if (this.config.capture_pageview && !this._initialPageviewCaptured && document) {
+            this._initialPageviewCaptured = true
+            this.capture('$pageview', { title: document.title }, { send_instantly: true })
+        }
```
benjackwhite, line 1991: Could we move this block into a private method like `captureInitialPageview` instead. that way it isn't likely to get forgotten about later in one place or the other

benjackwhite (changes requested):
Generally looks good. Just a change to make it easier to maintain in the future

src/posthog-core.ts
skoob13: Thanks @benjackwhite! I added a private method.

benjackwhite (approved):

benjackwhite:
Approved - dont forget to bump the label

PR #1373, opened 2024-08-20
fix(segment): Posthog can identify after segment identifies a user, not just during bootstrap.

Phanatic:
## Changes

The posthog x segment integration sets the user_state to `identified` and sets the `distinct_id` to `user.id()` only during bootstrap of the segment plugin. as shown here : https://github.com/PostHog/posthog-js/blob/9c8fe0e35f6566be7d0e37642c33ed7727fc68bb/src/extensions/segment-integration.ts#L119-L126

If the client application calls  `segment.identify()` **__after__**  `posthog.init()`, then the user_state is still `anonymous` and no feature flags are loaded.

This PR changes the behavior of the segment plugin, upon segment calling `identify()`, to : 
1. Set the USER_STATE to `identified`
2. Set the distinct_id for the current user session
3. Reload all feature flags for the new distinct_id

## Checklist


playground/segment/segment.html
```diff
@@ -9,7 +9,8 @@
                 analytics.load("GOEDfA21zZTtR7clsBuDvmBKAtAdZ6Np");
 
                 analytics.ready(() => {
-                    posthog.init('phc_mxO23D02CR3QU7V1u95ae3YEGNAFGuvyUlRpVNwX89B', {
+                    posthog.init('phc_lPwXKkvW2VtByfTsibDzFKAxn02rdVgMUiHWkZ5kW31', {
+                        api_host:"http://localhost:8000",
```
dmarticus, line 13:
```suggestion
                        api_host: "http://localhost:8000",
```

dmarticus (approved):

PR #1381, opened 2024-08-23
chore: Bump typescript version to 5.5 and also bump related libraries

robbie-c:
## Changes

As part of improving the CI, I wanted to use a library which is TS5+, I figured it made sense to do that in a separate PR.

I also bumped express as although it was only dev dependency, we were getting an unhelpful warning which this fixes.

## Checklist


robbie-c:
The other PR with these commits got merged, so closing this

PR #1383, opened 2024-08-23
chore: skip rate limiting of snapshot events

daibhin:
## Changes

Adaptation of https://github.com/PostHog/posthog-js/pull/1353 based on feedback in https://github.com/PostHog/posthog-js/pull/1353#discussion_r1721878554 and https://posthog.slack.com/archives/C03PB072FMJ/p1724078964640479

- We did not want to adapt the split limit because we know there's too many events, it feels weird to have workaround code
- Alternative approach: do the rate limiting before calling `_captureSnapshot`
  - Cool side effect of calling `clientRateLimitContext` is that it tracks one token which we can think of as representing the `$snapshot` event
  - That means when it comes to calling `_captureSnapshot` we can include `skip_client_rate_limiting` no matter how many individual snapshot events were created during the splitting of the buffer


src/extensions/replay/sessionrecording.ts
```diff
@@ -863,15 +863,21 @@ export class SessionRecording {
...
 
         if (this.buffer.data.length > 0) {
-            const snapshotEvents = splitBuffer(this.buffer)
-            snapshotEvents.forEach((snapshotBuffer) => {
-                this._captureSnapshot({
-                    $snapshot_bytes: snapshotBuffer.size,
-                    $snapshot_data: snapshotBuffer.data,
-                    $session_id: snapshotBuffer.sessionId,
-                    $window_id: snapshotBuffer.windowId,
+            const clientRateLimitContext = this.instance.rateLimiter.clientRateLimitContext()
```
pauldambra, line 866: I thought this would get us the global bucket? 

What I don't want (for example) is someone has an error in their code and exception autocapture rate limits because they're throwing in a loop

And then we can't capture a session recording.

I think we need to have a separate bucket for `$snapshot` events and a different rate limit for this - it's way chattier than other events

Ultimately I think we might want a separate bucket for errors but that can come with time.

I might be misunderstanding what's going on here ofc...

The benefit of client side rate limiting being event type aware would be it can pick the bucket and we don't need to change the recorder.

I'm picturing 

```
public clientRateLimitContext(checkOnly = false, suffix: string): {
        isRateLimited: boolean
        remainingTokens: number
    } {
        // This is primarily to prevent runaway loops from flooding capture with millions of events for a single user.
        // It's as much for our protection as theirs.
        const now = new Date().getTime()
        const bucket = this.instance.persistence?.get_property(CAPTURE_RATE_LIMIT + '/' + suffix) ?? {
            tokens: this.captureEventsBurstLimit,
            last: now,
        }
```

or something equivalent to that so that we have a totally separate bucket.

equally you could make this endpoint aware so that it automatically has a different rate limit per endpoint

the other thing... this is pretty new... maybe the limits are wrong and that's the simplest fix is they should be a little higher so that replay doesn't trigger them

(Sorry brain dumping since i'll be afk for hours now :))

src/extensions/replay/sessionrecording.ts
```diff
@@ -903,6 +909,7 @@ export class SessionRecording {
             _url: this.instance.requestRouter.endpointFor('api', this._endpoint),
             _noTruncate: true,
             _batchKey: SESSION_RECORDING_BATCH_KEY,
+            skip_client_rate_limiting: true,
```
pauldambra, line 912: in the short term i think we should just get this line in

i can't think of a problem caused by volume of snapshots that wasn't from a pre-batching version of the SDK

so i think we can skip client rate limiting for replay and then figure out a good solution with the pressure off

pauldambra (approved):
:LGTM:

yep, rrweb is so wary of missing events that we want to step carefully here 🙈

PR #1386, opened 2024-08-26
fix: report requests even without timing

pauldambra:
related to https://posthoghelp.zendesk.com/agent/tickets/16517 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/20544)

In that ticket xhr calls to an API are not being recorded.

ultimately this is because if there is no performance entry reported by the browser we don't send any information back to posthog :/ no bueno

In that site I can capture those calls with a performance observer by pasting the following in the terminal

```
// Function to log performance entries
function logPerformanceEntries(list) {
    const entries = list.getEntries();
    entries.forEach((entry) => {
        console.log('Performance entry:', entry);
    });
}

// Create a PerformanceObserver instance
const observer = new PerformanceObserver(logPerformanceEntries);

// Observe the 'resource' type entries which include fetch and XHR requests
observer.observe({ type: 'resource', buffered: true });
```

that's not quite how the recorder works but should be equivalent

but when we try to get the call in the recorder none is returned

i can't see any code in the site calling `performance.clearResourceTimings()` 

but the existence of `performance.clearResourceTimings` means we can't guarantee timings are available anyway.

so we should report what we can

----

maybe we should

* roll this out behind config so we can check the impact on posthog and just the reporters site at first to check we're not breaking display assumptions

----

tested locally and can still capture performance data and network payloads/headers


src/entrypoints/recorder.ts
```diff
@@ -326,34 +342,50 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
...
-const exposesServerTiming = (event: PerformanceEntry): event is PerformanceResourceTiming =>
-    event.entryType === 'navigation' || event.entryType === 'resource'
-
-function prepareRequest(
-    entry: PerformanceResourceTiming,
-    method: string | undefined,
-    status: number | undefined,
-    networkRequest: Partial<CapturedNetworkRequest>,
+const exposesServerTiming = (event: PerformanceEntry | null): event is PerformanceResourceTiming =>
+    !!event && (event.entryType === 'navigation' || event.entryType === 'resource')
```
daibhin, line 346: Don't know exactly when it is better to use the helpers but maybe this is one of those times
```suggestion
    !isNull(event) && (event.entryType === 'navigation' || event.entryType === 'resource')
```

src/entrypoints/recorder.ts
```diff
@@ -326,34 +342,50 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
...
+    entry: PerformanceResourceTiming | null
+    method: string | undefined
+    status: number | undefined
+    networkRequest: Partial<CapturedNetworkRequest>
     isInitial?: boolean
-): CapturedNetworkRequest[] {
+    start?: number
+    end?: number
+}): CapturedNetworkRequest[] {
+    start = entry ? entry.startTime : start
```
daibhin, line 365: Might save some complexity on line 376 and 383
```suggestion
    start = entry ? entry.startTime : start ?? 0
```

daibhin: Actually ignore me, looks like we want it to be undefined if it is

src/entrypoints/recorder.ts
```diff
@@ -326,34 +342,50 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
...
+    end,
+}: {
+    entry: PerformanceResourceTiming | null
+    method: string | undefined
+    status: number | undefined
+    networkRequest: Partial<CapturedNetworkRequest>
     isInitial?: boolean
-): CapturedNetworkRequest[] {
+    start?: number
+    end?: number
```
daibhin, line 363: Any reason to make these optional? Looks like this method is only called with the results of `window.performance.now()` as inputs for these values which always returns a number

daibhin (approved):

src/entrypoints/recorder.ts
pauldambra: yep... only one or two characters difference when bundled (e.g. you'd have maybe`!sQ(r)` vs `!!r`) but more intention revealing

src/entrypoints/recorder.ts
pauldambra: only that the values start undefined... _really_ you shouldn't ever get to the point of needing them and the value not be present but 🙈

daibhin (approved):
Is there a test we could write that would simulate the missing performance entry?

daibhin:
(Separately, but somewhat related to this change)

I was looking at the reported issue where the performance events were missing for a customer. Given we initialise our own PerformanceObserver maybe we could store the events it captures and use that as a lookup guide instead of the `performance.getEntriesByType('resource')` we rely on today

pauldambra:
> Is there a test we could write that would simulate the missing performance entry?

playing with more testing of the network recorder at the moment...

i'm sort of worried that this suffers a common problem of this kind of "wrapping external APIs" code. if we split things so that we can test the behaviour either the scaffolding is complex or we lose all encapsulation 🙈

pauldambra:
> Given we initialise our own PerformanceObserver maybe we could store the events it captures and use that as a lookup guide instead of the performance.getEntriesByType('resource') we rely on today

🤔

pauldambra:
I'll wait and get https://github.com/PostHog/posthog-js/pull/1395 in before this so we at least have one more level of testing on top of it

marandaneto (approved):

PR #1391, opened 2024-08-29
fix: Don't track hash params by default

benjackwhite:
## Changes

Hash params are typically not necessary to be tracked and often even undeseriable as they can be used for storing tokens etc. that never leave the browser.

I don't think we should track these by default. This is technically a breaking change so a little hard to guage but I strongly feel like the default should be not to track

We probably want to offer a config option to allow this...

## Checklist

### DIFF
--- a/src/utils/event-utils.ts
+++ b/src/utils/event-utils.ts
@@ -34,6 +34,7 @@ export const CAMPAIGN_PARAMS = [
 export const Info = {
     location: function (): { href?: string; hrefIncludingHash?: string; host?: string; pathname?: string } {
         return {
+            // NOTE: Hash params are typically sensitive and not designed to be sent to servers so we exclude them here by default
             href: location?.href?.split('#')[0],
             hrefIncludingHash: location?.href,
             host: location?.host,
--- a/src/utils/event-utils.ts
+++ b/src/utils/event-utils.ts
@@ -32,6 +32,15 @@ export const CAMPAIGN_PARAMS = [
 ]
 
 export const Info = {
+    location: function (): { href?: string; hrefIncludingHash?: string; host?: string; pathname?: string } {
+        return {
+            href: location?.href?.split('#')[0],
+            hrefIncludingHash: location?.href,
+            host: location?.host,
+            pathname: location?.pathname,
+        }
+    },
+
     campaignParams: function (customParams?: string[]): Record<string, string> {
         if (!document) {
             return {}
@@ -148,7 +157,7 @@ export const Info = {
         // we're being a bit more economical with bytes here because this is stored in the cookie
         return {
             r: this.referrer(),
-            u: location?.href,
+            u: Info.location().href,
         }
     },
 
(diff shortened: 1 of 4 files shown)


src/utils/event-utils.ts
```diff
@@ -32,6 +32,16 @@ export const CAMPAIGN_PARAMS = [
 ]
 
 export const Info = {
```
pauldambra, line 34: I know it's not this PR but "info" makes me sad 🤣 
we should rename it "misc" :p

src/utils/event-utils.ts
```diff
@@ -32,6 +32,16 @@ export const CAMPAIGN_PARAMS = [
...
+    location: function (): { href?: string; hrefIncludingHash?: string; host?: string; pathname?: string } {
+        return {
+            // NOTE: Hash params are typically sensitive and not designed to be sent to servers so we exclude them here by default
+            href: location?.href?.split('#')[0],
+            hrefIncludingHash: location?.href,
```
pauldambra, line 39: almost feels like we should capture the hash params separately so that if folk really want them they can capture them

instead of having href _and_ hash

I don't think that massively affects queries so no strong opinion but would sit nicely to have `query_params`, `pathname`, and `hash_params` as separate things

(no need to take that as blocking here i think generally we could do with better url handling)

pauldambra (approved):
agree it would be nice for this to be configurable... 
arguably if we allow the config we should add a scrub function so folk can capture via allowlist or denylist so they don't capture sensitive info unless they _really_ want to

since it looks like hash params are not being queried for by many folk we can probably not take on the complexity now and ask them if they need it and add it on demand

benjackwhite:
> agree it would be nice for this to be configurable... arguably if we allow the config we should add a scrub function so folk can capture via allowlist or denylist so they don't capture sensitive info unless they _really_ want to
> 
> since it looks like hash params are not being queried for by many folk we can probably not take on the complexity now and ask them if they need it and add it on demand

I feel like this isn't specific to URLs. We should just have a hook for people to hook into and modify the final captured event payload. That way they can do whatever level of modifications they want and avoids the need for more and more complex config options tbh...

pauldambra:
> We should just have a hook for people to hook into and modify the final captured event payload. That way they can do whatever level of modifications they want and avoids the need for more and more complex config options tbh...

+100 🧠

robbie-c:
> We should just have a hook for people to hook into and modify the final captured event payload. That way they can do whatever level of modifications they want and avoids the need for more and more complex config options tbh...

I'm in favour of this too. This would be cool:

```tsx
import {stripQueryParams, stripProperties} from "@posthog/js-transforms"

posthog.init("my-token", {
  transforms: [
    stripQueryParams(["some-oauth-token", "some-ad-id"]),
    stripProperties(["$browser"])
  ]
}
```

robbie-c (approved):

daibhin (approved):

PR #1395, opened 2024-08-30
fix: network capture tests should fail if we exhaust the body

pauldambra:
part of responding to https://github.com/PostHog/posthog/issues/24471 due to the unexpected consequences of https://github.com/PostHog/posthog-js/pull/1351

it is _very_ hard to test for all unexpected consequences of changes to code 🙈 

but we do at least _now_ know for sure that that we cannot rely on request/response objects being readable and that we should not read them and then pass them on

Screenshot 2024-08-30 at 10 03 11

Here's the test failing in CI and showing we would have avoided all of the painful impact for folk

```
  1 failing
  1) Session recording
       network capture - when fetch wrapper is not badly behaved
         it sends network payloads:
     TypeError: The following error originated from your test code, not from Cypress. It was caused by an unhandled promise rejection.
  > Failed to execute 'fetch' on 'Window': Cannot construct a Request with a Request object that has already been used.
```

### DIFF
--- a/src/entrypoints/recorder.ts
+++ b/src/entrypoints/recorder.ts
@@ -220,6 +220,8 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
     const recordRequestHeaders = shouldRecordHeaders('request', options.recordHeaders)
     const recordResponseHeaders = shouldRecordHeaders('response', options.recordHeaders)
 
+    logger.info('initialising xhr observer', { recordRequestHeaders, recordResponseHeaders })
+
     const restorePatch = patch(
         win.XMLHttpRequest.prototype,
         'open',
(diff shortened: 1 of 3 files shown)


pauldambra:
@daibhin just a heads up on this in case you spot anything, but don't worry too much, it's very much a spike at the moment

src/entrypoints/recorder.ts
```diff
@@ -420,13 +422,18 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
-            .finally(() => clearTimeout(timeout))
+        try {
+            r.clone()
+                .text()
+                .then(
+                    (txt) => resolve(txt),
+                    (reason) => reject(reason)
+                )
+                .finally(() => clearTimeout(timeout))
+        } catch (e) {
```
pauldambra, line 433: this is a genuine fix for an actual bug, not just a test addition

errors in `.text()` are caught in the `then` and rejected

but `r.clone()` can throw and we weren't catching that

src/entrypoints/recorder.ts
```diff
@@ -420,13 +422,18 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+            clearTimeout(timeout)
+            resolve('[SessionReplay] Failed to read body')
```
pauldambra, line 435: so instead we should catch that error, and resolve with something meaningful, that means our captured network request would have this string as its body

and everything else should carry on

pauldambra: it's not the only way to make this fail though.. it isn't _the_ bug

marandaneto (approved):

PR #1396, opened 2024-09-02
fix: flushing the buffer for debug signal while idle extends session activity

pauldambra:
In stepping through what we do around idleness in the recorder 

* we only check and extend session if it’s an interactive event
* i.e. if the user is not idle
* Reading the code of `checkAndGetSessionAndWindowId` you must check and extend session in order to check idle timeout 
* So you can’t have an idle timeout in recording unless you return from idle 🤯
* While idle we only allow custom events  they go into the in-memory buffer  when that happens…
* the buffer is flushed if the session id has changed or if the buffer holds > 0.9MBish  and if not we set a two second timer which will flush the buffer  
* when flushing the buffer so long as the buffer is not empty we capture a snapshot with the buffer contents 
* that calls capture on the posthog instance  
* which among other things makes a non-readonly check of the session id and window id 
* which might cause the session id to change generally 
* _and_ for the message being captured **overwriting the data being sent with the new session and window id** 
* and checking the session/window id extends the activity timer
* so if a session has frequent enough custom events while idle it would never idle timeout

This PR ensures we don't flush the buffer when idle but still allows buffering (allowed) snapshot data while idle

### DIFF
--- a/src/__tests__/extensions/replay/sessionrecording.test.ts
+++ b/src/__tests__/extensions/replay/sessionrecording.test.ts
@@ -1364,7 +1364,7 @@ describe('SessionRecording', () => {
             })
         })
 
-        it('emits custom events even only when returning from idle', () => {
+        it('buffers custom events without capturing while idle', () => {
             // force idle state
             sessionRecording['isIdle'] = true
             // buffer is empty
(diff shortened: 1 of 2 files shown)


src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -1362,7 +1364,40 @@ describe('SessionRecording', () => {
...
+                size: 47,
+                windowId: 'windowId',
+            })
+            emitInactiveEvent(startingTimestamp + 100, true)
+            expect(posthog.capture).not.toHaveBeenCalled()
+
+            expect(sessionRecording['flushBufferTimer']).toBeUndefined()
+        })
+
+        it('buffers custom events without capturing while idle', () => {
```
daibhin, line 1400: Looks like these two tests have the same name

pauldambra: you found my deliberate mistake 🤣

daibhin (approved):

PR #1397, opened 2024-09-03
feat: add a `page` function

anirudh24seven:
## Changes
Adding a new page function as requested in https://github.com/PostHog/posthog/issues/635
...

## Checklist


src/posthog-core.ts
```diff
@@ -864,6 +864,26 @@ export class PostHog {
...
+     *
+     *    posthog.page('https://example.com')
+     *
+     * @param {String} url The URL of the page. This will be included in the properties as '$current_url'.
+     * @param {Object} [properties] A set of properties to include with the event you're sending. These describe the user who did the event or details about the event itself.
+     * @param {Object} [config] Optional configuration for this capture request.
+     * @param {String} [config.transport] Transport method for network request ('XHR' or 'sendBeacon').
+     * @param {Date} [config.timestamp] Timestamp is a Date object. If not set, it'll automatically be set to the current time.
+     */
+    page(url: string, properties?: Properties | null, options?: CaptureOptions): CaptureResult | undefined {
```
marandaneto, line 883: some places are calling `posthog.capture('$pageview')`, so you can refactor to use this new method.

anirudh24seven: @marandaneto From my simple understanding, I am assuming that the newly created page function can only replace capture function calls if it has a $current_url property. Or to put it differently, I am assuming that url is a mandatory parameter to the page function.

Based on this assumption, I could find only one instance of code which had to be refactored, as demonstrated in the below PR.

src/posthog-core.ts
```diff
@@ -864,6 +864,26 @@ export class PostHog {
...
+        properties['$current_url'] = url
```
marandaneto, line 884: if `properties` is undefined/null here, you should create the `properties` first, otherwise `page(url)` would fail

anirudh24seven: Makes sense. Will do that.

anirudh24seven: @marandaneto I've done the needful. Please let me know if more changes are needed.

src/posthog-core.ts
```diff
@@ -864,6 +864,28 @@ export class PostHog {
...
+     *
+     * ### Usage:
+     *
+     *    posthog.page('https://example.com')
+     *
+     * @param {String} url The URL of the page. This will be included in the properties as '$current_url'.
+     * @param {Object} [properties] A set of properties to include with the event you're sending. These describe the user who did the event or details about the event itself.
+     * @param {Object} [config] Optional configuration for this capture request.
+     * @param {String} [config.transport] Transport method for network request ('XHR' or 'sendBeacon').
+     * @param {Date} [config.timestamp] Timestamp is a Date object. If not set, it'll automatically be set to the current time.
```
marandaneto, line 881: those params dont exist here in the method signature

anirudh24seven: @marandaneto I would like your help with a point that I am confused by.

If you observe the docs for the capture method in lines 761-763 in the same file, it defines some config params, while the param is named options in the function argument.

For the capture method:
1. Is config the same as options?
2. Should config be renamed to options? Shall I do it in this PR itself?

I am trying to reuse the docs of capture method for the new page method.

For the new page method:
1. Should I similarly rename config to be options in the docs?
(or)
2. Should I remove the config lines in the docs for this new method altogether?

marandaneto: I think the code docs are not up to date.
`config` is the equivalent of `options`.
timestamp and transport are part of `options`, so they could be removed I guess.
Maybe @pauldambra knows more about this, maybe we just need to update the code docs.

pauldambra: Yep, it'll be comments getting out of date.

You're welcome to update the comment on the main capture block has well.

src/__tests__/posthog-core.ts
```diff
@@ -337,6 +337,14 @@ describe('posthog core', () => {
         })
     })
 
+    describe('page', () => {
+        it('has the correct url', () => {
+            const testUrl = 'https://example.com'
+            const captureData = posthogWith(defaultConfig, defaultOverrides).page(testUrl, {}, {})
```
pauldambra, line 343: elsewhere we test that after capture particular values have been passed to _send_request

i.e. capture ran as expected.
here we're testing the return value from capture 

so we should either

* test that properties are set / passed through and end up being passed out of capture
* or simply mock capture and test that `page` calls capture (and so only needs to rely on capture being correct)

that second lets us test that 

* properties are passed through
* options are passed through
* url is set correctly

it'd make the contract of page much clearer in the tests as well

---

why? i could reimplement page without calling capture but returning the expected data directly and the tests would pass but the library would be broken :)

pauldambra: would be good to test `(url, {}, {})`, `(url, {some: 'prop')`, `(url)` (or some similar combination of possible inputs to clarify the contract

src/posthog-core.ts
```diff
@@ -864,6 +864,28 @@ export class PostHog {
...
+     */
+    page(url: string, properties?: Properties | null, options?: CaptureOptions): CaptureResult | undefined {
```
pauldambra, line 883: we'll also need to update usage docs on posthog.com

and we should have a cypress test that covers pageview events (even implicitly) which we can duplicate to validate that `page` causes actual events to be sent

src/__tests__/posthog-core.ts
anirudh24seven: @pauldambra This makes sense. To paraphrase, `page` is tightly coupled to the functionality of `capture` and the test cases should reflect that, so that when `page` is accidentally rewritten as another implementation without calling `capture` (decoupled from `capture`), the test cases should fail, right?

pauldambra: yep exactly... all `page` should do is maybe transform the args passed to it and call capture with them

so we only need to test that it calls capture as expected with some example inputs

that way the tests correctly describe and protect the contract

src/posthog-core.ts
anirudh24seven: > we'll also need to update usage docs on posthog.com
To confirm, is this at https://github.com/PostHog/posthog.com/blob/master/contents/docs/libraries/js/index.mdx?

> and we should have a cypress test that covers pageview events (even implicitly) which we can duplicate to validate that page causes actual events to be sent

Does this mean something similar to `posthog-js/cypress/e2e/capture.cy.ts`? I'll start on something and raise a PR.

anirudh24seven:
@pauldambra I have made the requested changes but I have 2 concerns:

1. I am not sure whether the modification I have made to the cypress test is relevant. I am not sure if I fully understood your expectation. Should I add a new command in `posthog-js/cypress/support/commands.ts` to test the `page` function?
2. I still have not run the tests locally as I am running into setup issues related to the tests. Will update this PR once I confirm that all the unit tests and e2e tests pass.

anirudh24seven:
Status of cypress tests:

cypress/e2e/capture.cy.ts
```diff
@@ -25,6 +25,20 @@ describe('Event capture', () => {
         cy.phCaptures().should('include', 'custom-event')
     })
 
+    it('captures pageviews with current url', () => {
+        start({})
+
+        cy.get('[data-cy-custom-event-button]').click()
```
pauldambra, line 31: i don't _think_ the nuxt client affects these cypress pages...

i believe you can get to the page's posthog instance, i'd be tempted to call `cy.posthog.page('https://not-the-page-i-loaded;)` (assuming that's how you get to the posthog intance)

and then assert that URL has been captured on a pageview

a little more explicit and certain that way

src/__tests__/posthog-core.ts
```diff
@@ -337,6 +337,82 @@ describe('posthog core', () => {
...
+            const testUrl = 'https://example.com'
+            const posthog = posthogWith({})
+            const captureMock = jest.spyOn(posthog, 'capture').mockImplementation()
+            posthog.page(testUrl)
+            expect(captureMock).toHaveBeenCalledWith(
+                '$pageview',
+                expect.objectContaining({ $current_url: testUrl }),
+                undefined
+            )
+        })
```
pauldambra, line 365: you can collapse these two tests into one 👍

pauldambra (approved):
couple of changes and this is good to go 🚀

cypress/e2e/capture.cy.ts
anirudh24seven: @pauldambra I've made the changes. Please have a look.

anirudh24seven:
@pauldambra Can you please have a look at this PR once again, if time permits?

pauldambra:
@anirudh24seven can you update with `main` 

all looking solid to me otherwise 💪

Issue #1344, opened 2024-08-06
Survey mode API trigger a bug when solo loaded


ThibaudTeachup:
Hello everyone !

I started using Posthog for posting surveys on my website but recently, if I only send one API mode survey on my website, the method getActiveMatchingSurveys() doesn't return anything anymore and I get an error like this :

![image](https://github.com/user-attachments/assets/f55da4ae-10c7-412e-9c82-8709afc52d35)

I was wondering if someone knows why this error occurs or if it's a current bug on the posthog js library.

Thanks in advance !

kalpesh-aalore:
@ThibaudTeachup did you found any answer I am having the same issue it comes as posthog takes time to initialized

ThibaudTeachup:
@kalpesh-aalore I unfortunatly didn't find any solution to this problem. Yeah it looks like an init problem since sometimes it works fine for no reason.

pauldambra:
Awesome, thanks for letting us know!

PR #1400, opened 2024-09-05
fix: warn on unavailable lazy load not throw

pauldambra:
relates to #1344

we're checking the presence of a method on something that _could_ be undefined

but we don't correctly type all the lazy load stuff, so TS doesn't complain

this at least safely checks the presence and will warn instead 
we're likely calling the method too early (before it has had a chance to be added to the window) this might even be safe, so we definitely shouldn't throw

### DIFF
--- a/src/posthog-surveys.ts
+++ b/src/posthog-surveys.ts
@@ -267,6 +267,7 @@ export class PostHogSurveys {
     private _canActivateRepeatedly(survey: Survey) {
         if (isNullish(assignableWindow.__PosthogExtensions__?.canActivateRepeatedly)) {
             logger.warn(LOGGER_PREFIX, 'canActivateRepeatedly is not defined, must init before calling')
+            return false // TODO does it make sense to have a default here?
         }
         return assignableWindow.__PosthogExtensions__.canActivateRepeatedly(survey)
     }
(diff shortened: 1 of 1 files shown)


jurajmajerik (approved):
makes sense, thanks 🙏

src/posthog-surveys.ts
```diff
@@ -265,8 +265,9 @@ export class PostHogSurveys {
 
     // this method is lazily loaded onto the window to avoid loading preact and other dependencies if surveys is not enabled
     private _canActivateRepeatedly(survey: Survey) {
-        if (isNullish(assignableWindow.__PosthogExtensions__.canActivateRepeatedly)) {
+        if (isNullish(assignableWindow.__PosthogExtensions__?.canActivateRepeatedly)) {
             logger.warn(LOGGER_PREFIX, 'canActivateRepeatedly is not defined, must init before calling')
+            return false // TODO does it make sense to have a default here?
```
pauldambra, line 270: @jurajmajerik @dmarticus i'm unfamiliar with this (and lazy 🤣 )

is it safe to return false here? will we accidentally disable surveys or something?

jurajmajerik: I think this is fine.

This method is called here: https://github.com/PostHog/posthog-js/blob/fix/warn-not-throw/src/posthog-surveys.ts#L182

If you don't return false early than an error will be thrown on return.

@Phanatic can you double check please? I believe you've added this recently.

Phanatic (approved):
This looks good, ship it!

PR #1401, opened 2024-09-06
feat: web vitals 3000

pauldambra:
or at least web vitals 1000

see https://github.com/PostHog/posthog/issues/24597

this PR 

* ~~adds sample rate config (supporting client or remote config)~~
* adds metrics allowlist (supporting client or remote config)
* add more INP info to events


src/entrypoints/web-vitals.ts
```diff
@@ -1,15 +1,15 @@
...
-
 const postHogWebVitalsCallbacks = {
     onLCP,
     onCLS,
     onFCP,
     onINP,
 }
 
-assignableWindow.postHogWebVitalsCallbacks = postHogWebVitalsCallbacks
+assignableWindow.__PosthogExtensions__ = assignableWindow.__PosthogExtensions__ || {}
```
pauldambra, line 12: every time we add random things onto the window an angel dies (probably)

so I'm slowly moving things onto a single global object on window

this will mean we can start to type that object and have fewer `(window as any)` flying around the place

src/entrypoints/web-vitals.ts
```diff
@@ -1,15 +1,15 @@
...
+assignableWindow.__POSTHOG_EXTENSIONS__postHogWebVitalsCallbacks = postHogWebVitalsCallbacks
```
pauldambra, line 13: Yuck... look at what the implicit "any" did here... and made me typo this 🤦 🤣

src/extensions/web-vitals/index.ts
```diff
@@ -22,9 +24,30 @@ export class WebVitalsAutocapture {
 
     constructor(private readonly instance: PostHog) {
         this._enabledServerSide = !!this.instance.persistence?.props[WEB_VITALS_ENABLED_SERVER_SIDE]
+
         this.startIfEnabled()
     }
 
+    public get sampleRate(): number | undefined {
```
pauldambra, line 31: i'm wondering about taking this out

we have sampling in recording and it makes sense to have that separate since you need to sample by session not by event

but once we add it to web vitals the obvious question from someone is "why not autocapture or pageleave", so I think the feature should be global sampling to keep config from exploding

daibhin (approved):

PR #1404, opened 2024-09-09
feat: web vitals 3000 fallback

brunomguimaraes:
https://github.com/PostHog/posthog-js/pull/1401

### DIFF
--- a/src/extensions/web-vitals/index.ts
+++ b/src/extensions/web-vitals/index.ts
@@ -195,17 +195,20 @@ export class WebVitalsAutocapture {
         let onCLS: WebVitalsMetricCallback | undefined
         let onFCP: WebVitalsMetricCallback | undefined
         let onINP: WebVitalsMetricCallback | undefined
-
+    
         const posthogExtensions = assignableWindow.__PosthogExtensions__
         if (!isUndefined(posthogExtensions)) {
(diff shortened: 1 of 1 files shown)


src/extensions/web-vitals/index.ts
```diff
@@ -200,7 +200,10 @@ export class WebVitalsAutocapture {
         if (!isUndefined(posthogExtensions)) {
             ;({ onLCP, onCLS, onFCP, onINP } = posthogExtensions.postHogWebVitalsCallbacks)
         }
-
+        else if (assignableWindow.postHogWebVitalsCallbacks) {
```
pauldambra, line 203: hey @brunomguimaraes this shouldn't be necessary, what's the problem this solves?

pauldambra: Ugh, dang... It's because we load the latest web vitals extension into older clients. What a pain 😭 

AFK right now will look at this properly when back

pauldambra:
hey @brunomguimaraes this wouldn't fix the issue for old clients since they wouldn't have this new code

i've opened #1406 instead, thanks for flagging this though

src/extensions/web-vitals/index.ts
pauldambra: :)

I should have caught it... have opened https://github.com/PostHog/posthog-js/pull/1407 to try and work towards making this all safer

PR #1406, opened 2024-09-09
fix: some clients need web vitals directly on window

pauldambra:
In #1401 we moved the web vitals callbacks onto an object on the window instead of placing them directly on the window... there are more and more extensions being added on to the window - and that is no bueno 

but, existing clients installed with npm may now request web vitals and look for it directly on the window but it isn't there any more - also no bueno

we can't fix this in the main bundle since its only old clients and they won't have that code, so it has to be fixed in the entrypoint file - which is what generates the web vitals file the old clients are lazily loading


mariusandra (approved):

brunomguimaraes:
Thanks for this change @pauldambra !

PR #1407, opened 2024-09-09
feat: type safe posthog extensions

pauldambra:
We keep adding more and more things directly to window
We get no type help from TS
And we're not being super unique with naming so could easily clash with someone else

Let's be a little more explicit about where and what we add to the window


src/utils/globals.ts
```diff
@@ -11,6 +14,27 @@
...
+        onLCP: (metric: any) => void
+        onCLS: (metric: any) => void
+        onFCP: (metric: any) => void
+        onINP: (metric: any) => void
+    }
+    tracingHeadersPatchFns?: {
+        _patchFetch: (sessionManager: SessionIdManager) => () => void
+        _patchXHR: (sessionManager: any) => () => void
+    }
+}
```
pauldambra, line 36: here's the secret sauce... we've added a bunch of stuff directly to window and you can't really see what without reading a bunch of files, and you don't get any typesafety at all since it's all `(window as any)` 

we'd have to be careful here not to bring in the lazy bundles and inflate the bundle unnecessarily

but this is at least a step better

marandaneto (approved):
noice

src/entrypoints/surveys.ts
```diff
@@ -1,12 +1,10 @@
...
 import { canActivateRepeatedly } from '../extensions/surveys/surveys-utils'
 
-if (window) {
-    ;(window as any).__PosthogExtensions__ = (window as any).__PosthogExtensions__ || {}
-    ;(window as any).__PosthogExtensions__.canActivateRepeatedly = canActivateRepeatedly
-    ;(window as any).extendPostHogWithSurveys = generateSurveys
-}
+assignableWindow.__PosthogExtensions__ = assignableWindow.__PosthogExtensions__ || {}
+assignableWindow.__PosthogExtensions__.canActivateRepeatedly = canActivateRepeatedly
+assignableWindow.extendPostHogWithSurveys = generateSurveys
```
benjackwhite, line 8: For completenesss sake it would be nice if we moved this over as well.

src/utils/globals.ts
```diff
@@ -1,3 +1,8 @@
+import { SessionIdManager } from '../sessionid'
+import { ErrorEventArgs, ErrorProperties, Properties } from '../types'
+import { PostHog } from '../posthog-core'
+import { SurveyManager } from '../extensions/surveys'
```
benjackwhite, line 4: Errrr surely this isn't what you want...

benjackwhite (changes requested):

src/utils/globals.ts
benjackwhite: Although I am surprised the bundle didn't grow much it still seems like we should never be importing from extensions...

pauldambra: Ah... the fun of offsite brain tiredness. Should have a linter maybe to save in future... 🤔

pauldambra: nicer if we had interfaces instead of these `any`s but alreayd a lot changing in here so 🙈

benjackwhite: Technically we could do `import type` and that would give you what you want but thats also at risk of getting changed and importing the actual data too...

benjackwhite (approved):

PR #1409, opened 2024-09-10
feat(experiments): Apply no-code experiments to the webpage.

Phanatic:
## Changes

This PR introduces a new extension to posthog-js called `web-experiments` which allows posthog to **apply no-code experiments** to elements on a web page.  This PR needs https://github.com/PostHog/posthog/pull/24872 to merge so it can function. 

To enable this feature, developers have to opt-in by setting `disable_web_experiments` to false in `posthog.init` while in Beta. 

There are multiple ways to target an experiment within the browser here : 
1. Current URL can be matched with `exact, contains, not_contains, regex, not_regex`
2. UTM targeting works by parsing the Current URL and evaluating conditions on individual variants.

## Checklist

### Coming up in following PRs
* Support for dynamic elements on a webpage, via MutationObserver
* Anti-flicker snippet support to prevent flickering of the webpage when applying web experiments.


src/web-experiments-types.ts
```diff
@@ -0,0 +1,38 @@
+export interface WebExperimentTransform {
+    attributes:
+        | {
+              name: string
+              value: string
+          }[]
+        | null
```
dmarticus, line 7: making this value nullable instead of optional is interesting; what's the significance of going with this approach over something like this?
```
attributes?: { name: string; value: string }[]
```
Is making this value nullable vs undefined important?  Feels like we could avoid the extra null checks elsewhere by making this optional.

src/web-experiments-types.ts
```diff
@@ -0,0 +1,38 @@
...
+    conditions: {
+        url?: string
+        urlMatchType?: WebExperimentUrlMatchType
+        utm: {
+            utm_source?: string
+            utm_medium?: string
+            utm_campaign?: string
+            utm_term?: string
+        }
+    } | null
```
dmarticus, line 27: similar comment about making this payload undefined vs null as I had above; why extra an extra null state that needs more defensive coding?

src/web-experiments.ts
```diff
@@ -0,0 +1,261 @@
...
+
+                                case 'cssClass':
+                                    htmlElement.className = attribute.value
+                                    break
+
+                                default:
+                                    htmlElement.setAttribute(attribute.name, attribute.value)
+                            }
+                        })
+                    }
```
dmarticus, line 244: obviously users won't try to do this, and it's possible we're making this impossible on the server side, but should we sanitize the contents of these values that we're setting onto the DOM?

dmarticus (approved):
looks reasonable overall, nice work!  A couple of questions about data modeling and one about sanitizing HTML content but the rest looks fine.

src/web-experiments-types.ts
Phanatic: TBH, I don't have a strong preference here. I like your approach better, will adopt.

src/web-experiments.ts
Phanatic: good idea about sanitizing payload. We can do that on the server-side with nh like we do for survey HTML

PR #1410, opened 2024-09-10
chore: add page title to custom events

daibhin:
## Changes

In working through some of the design changes proposed in https://github.com/PostHog/posthog/issues/24561 we are looking to track the title of a given page in the playback tabs

Screenshot 2024-09-10 at 20 26 01

Given we don't want to rely on `$pageview` events where possible in Replay because we can't always guarantee a customer uses events, it seems like the only alternative is to add the title to the custom rrweb events.

Open to pushback here. Other options include:
1) Never showing the title during playback and just relying on the url
2) Only enhancing the tab with the title if a customer collects events

## Checklist

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -303,7 +303,7 @@ export class SessionRecording {
                             if (!href) {
                                 return
                             }
-                            this._tryAddCustomEvent('$pageview', { href })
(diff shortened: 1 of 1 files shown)


pauldambra:
I wonder if we could do this with a plugin on the replayer side... the full snapshots will have the doc title in and changes to the title will come through in mutations

(which has the benefit it would work on old recordings too)

daibhin:
Closing in favour of https://github.com/PostHog/posthog/pull/24943

PR #1411, opened 2024-09-10
fix: teardown and restart recording on session id change

pauldambra:
sometimes we see the artificial full snapshot on session idle timeout end up in the previous session not the new session - no bueno

in an attempt to not keep adding conditionals here i want to explore if tearing down rrweb and restarting recording is more stable

---

locally this seems way more reliable, but I want to get https://github.com/PostHog/posthog-js/pull/1416 in so I can retest without that noise


src/extensions/replay/sessionrecording.ts
```diff
@@ -579,12 +590,11 @@ export class SessionRecording {
...
 
-        if (
-            returningFromIdle ||
-            ([FULL_SNAPSHOT_EVENT_TYPE, META_EVENT_TYPE].indexOf(event.type) === -1 &&
-                (windowIdChanged || sessionIdChanged || isUndefined(this._fullSnapshotTimer)))
-        ) {
-            this._tryTakeFullSnapshot()
+        if (sessionIdChanged || windowIdChanged) {
+            this.stopRecording()
+            this.startIfEnabledOrStop()
```
daibhin, line 595: Haven't looked into it but wondering if we need to consider anything here with regards to ingestion controls (sampling, feature flags, minimum duration) that might affect the new recording not starting

daibhin (commented):
Looks like a simple enough change, happy to approve once you're ready for a full review

src/extensions/replay/sessionrecording.ts
pauldambra: sampling and feature flags should just work...

min duration is interesting... i don't know if we'd restart the counter - we should... 👀

src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -426,26 +427,6 @@ describe('SessionRecording', () => {
             })
         })
 
-        it('when the first event is an incremental it does take a manual full snapshot', () => {
```
pauldambra, line 429: not manually fangling this any more so we don't need to test it happens

src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -764,9 +744,8 @@ describe('SessionRecording', () => {
             expect(posthog.capture).toHaveBeenCalledWith(
                 '$snapshot',
                 {
-                    $snapshot_bytes: 80,
+                    $snapshot_bytes: 60,
                     $snapshot_data: [
-                        createFullSnapshot(),
```
pauldambra, line 769: throughout these tests we're no longer manually adding full snapshots so we don't have to assert their presence

daibhin (approved):

src/extensions/replay/sessionrecording.ts
pauldambra: yep... session start is read from storage, most recent snapshot data from the buffer, so we're safe with min duration  too 👍

PR #1414, opened 2024-09-12
chore: auto publish alpha versions

pauldambra:
Sometimes we want to test changes without releasing them generally
Let's allow ourselves to publish alphas more easily

This trigger type only fires if the workfllow file is on the default branch so I can't test it exclusively in one PR

This PR introduces a workflow file that runs when a specified label is added

The workflow

* checks the person has write access to posthog-js
* pushes an alpha prerelease to NPM

# To Follow-up

it also isn't clever enough to e.g. push alpha-1, alpha-2 etc so if someone ran it twice on one version of posthog-js then it would fail but my goal today is to push an alpha of https://github.com/PostHog/posthog-js/pull/1411 so I can figure out how to use it and then test it

to follow-up to this PR would then be detecting a sensible alpha version to publish

### DIFF
--- a/.github/workflows/label-alpha-release.yml
+++ b/.github/workflows/label-alpha-release.yml
@@ -3,72 +3,73 @@ name: Autorelease alpha version
 on:
   pull_request:
     types:
-      - labeled  # Triggers when any label is added to the PR
+      - labeled
 
 jobs:
-    check-permissions:
-      name: "Check user has permission to release"
-      runs-on: ubuntu-latest
-      steps:
-        - uses: actions-cool/check-user-permission@v2
-          id: check-permissions
-          with:
-            require: 'write'
+  check-permissions:
+    name: "Check user has permission to release"
+    runs-on: ubuntu-latest
+    outputs:
+      check-result: ${{ steps.check-permissions.outputs.check-result }}
+      user-permission: ${{ steps.check-permissions.outputs.user-permission }}
+      require-result: ${{ steps.check-permissions.outputs.require-result }}
+    steps:
+      - uses: actions-cool/check-user-permission@v2
+        id: check-permissions
+        with:
+          require: 'write'
 
-        - run: echo "user-permission = ${{ steps.check-permissions.outputs.user-permission }}"
-        - run: echo "require-result = ${{ steps.check-permissions.outputs.require-result }}"
-        - run: echo "check-result = ${{ steps.check-permissions.outputs.check-result }}"
-    warn-when-failed:
-        name: Warn when the user does not have the required permissions
-        runs-on: ubuntu-latest
-        if: |
-          steps.check-permissions.outputs.check-result == false
-        steps:
-          - run: echo "The user does not have the required permissions to trigger this action."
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -91,3 +95,8 @@ To release an alpha or beta version, you'll need to use the CLI locally:
     ```
 
 5. Enjoy the new prerelease version. You can now use it locally, in a dummy app, or in the [main repo](https://github.com/posthog/PostHog).
+
+#### Automagically
```
pauldambra, line 99:
```suggestion
#### Automagically (in progress - don't use yet!)
```

daibhin (approved):

README.md
```diff
@@ -91,3 +95,8 @@ To release an alpha or beta version, you'll need to use the CLI locally:
     ```
 
 5. Enjoy the new prerelease version. You can now use it locally, in a dummy app, or in the [main repo](https://github.com/posthog/PostHog).
+
+#### Automagically (in progress - don't use yet!)
```
pauldambra, line 99:
```suggestion
#### Automagically
```

README.md
```diff
@@ -91,3 +95,8 @@ To release an alpha or beta version, you'll need to use the CLI locally:
...
+
+Use the release-alpha label on your PR to have an alpha version bumped and published automatically.
+Remember that these are public and folk might use them, so make sure they're not _too_ alpha 🙈
```
pauldambra, line 102:
```suggestion
Use the "release alpha" label on your PR to have an alpha version published automatically. This automation currently doesn't check whether an alpha exists for the version it will try to publish. If you need to publish two alphas from one PR you'll need to fix that

Remember that these versions are public and folk might use them, so make sure they're not _too_ alpha 🙈
```

.github/workflows/label-alpha-release.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Update version in package.json
+        if: steps.bump-type.outputs.bump-type != 'null'
+        run: |
+          mv package.json package.old.json
+          jq --indent 4 '.version = "${{ steps.versions.outputs.new-version }}"' package.old.json > package.json
+          rm package.old.json
+
+      - name: Publish the alpha in the npm registry
+        run: npm publish --access public
```
benjackwhite, line 71: my memory is hazey but in js-lite we set a tag to ensure it is off the main stream. Is that maybe missing here? https://github.com/PostHog/posthog-js-lite/blob/main/.github/workflows/release.yml#L59

I think the version containing alpha doesn't put it off the main track

benjackwhite (changes requested):

.github/workflows/label-alpha-release.yml
pauldambra: https://docs.npmjs.com/cli/v10/commands/npm-publish#tag

> If you ask npm to install a package and don't tell it a specific version, then it will install the specified tag.

i did not read the docs that way... but you're totally right

benjackwhite (approved):

PR #1416, opened 2024-09-13
fix: no keep alive for large payloads

pauldambra:
we set keepalive to true for any request that is a POST

the fetch spec at https://fetch.spec.whatwg.org/#http-network-or-cache-fetch says

> If contentLength is non-null and httpRequest’s [keepalive](https://fetch.spec.whatwg.org/#request-keepalive-flag) is true, then:
> If the sum of contentLength and inflightKeepaliveBytes is greater than 64 kibibytes, then return a [network error](https://fetch.spec.whatwg.org/#concept-network-error).

read more at e.g. https://javascript.info/fetch-api#:~:text=But%20the%20keepalive%20option%20tells,for%20keepalive%20requests%20is%2064KB.

it is likely that a replay payload would be >64kb compressed

not all browsers obey this spec equally (of course) so it won't always fail, so it's not broken consistently enough to stand out.

but locally I can see network requests failing for this reason

fetch keepalive is intended for situations including those like analytics (👋 posthog) where you might want to allow a request to finish even though the page is navigating so we don't want to turn it off completely

lets estimate body size and set keepalive when its safe (which it should be for most analytics events)

### DIFF
--- a/src/request.ts
+++ b/src/request.ts
@@ -17,6 +17,7 @@ const CONTENT_TYPE_FORM = 'application/x-www-form-urlencoded'
 type EncodedBody = {
     contentType: string
     body: string | BlobPart
+    estimatedSize: number
 }
 
 export const extendURLParams = (url: string, params: Record<string, any>): string => {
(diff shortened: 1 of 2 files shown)


src/request.ts
```diff
@@ -133,7 +140,14 @@ const _fetch = (options: RequestOptions) => {
...
         headers,
-        keepalive: options.method === 'POST',
+        // if body is greater than 64kb, then fetch with keepalive will error
+        // see 8:10:5 at https://fetch.spec.whatwg.org/#http-network-or-cache-fetch,
+        // but we do want to set keepalive sometimes as it can  help with success
+        // when e.g. a page is being closed
+        // so let's get the best of both worlds and only set keepalive for POST requests
+        // where the body is less than 64kb
+        // NB this is fetch keepalive and not http keepalive
+        keepalive: options.method === 'POST' && (estimatedSize || 0) < 64 * 1024,
```
daibhin, line 150: Could abstract this but doesn't seem necessary given the comment explains a lot
```suggestion
        keepalive: options.method === 'POST' && (estimatedSize || 0) < MAX_KEEPALIVE_KB,
```

daibhin (approved):

src/request.ts
pauldambra: yeah, i'm always torn by the tension between "clarify intent" and "have fewest parts"
if you make it a const it's arguably clearer but i need to look at multiple places to be sure i understand

marandaneto (approved):

PR #1426, opened 2024-09-20
feat: Treat self referrals as direct

robbie-c:
## Changes
We handle the `$referrer` property in a potentially confusing way when an internal link is clicked. The latest example is https://posthog.slack.com/archives/C03C60FT1J7/p1726670676997629.

This happens because we use `document.referrer` directly, and this can be self-referential in the following example
1. User is on `https://referrer.com`
2. User clicks a link to `https://tracked.com`
3. Referrer is `referrer.com`
4. User is inactive long enough for the session to expire
5. User clicks an internal link to `https://tracked.com`
6. Referrer is `tracked.com`

This PR changes this, so that the referrer in 6) would be `$direct`

To be clear about this: it is a *breaking change*, however I believe it's more in line with what the expected behaviour would be. One might argue that the original behaviour is a bug

## Checklist

### DIFF
--- a/src/__tests__/posthog-core.test.ts
+++ b/src/__tests__/posthog-core.test.ts
@@ -4,6 +4,7 @@ import { uuidv7 } from '../uuidv7'
 
 const mockReferrerGetter = jest.fn()
 const mockURLGetter = jest.fn()
+const mockHostGetter = jest.fn()
 jest.mock('../utils/globals', () => {
     const orig = jest.requireActual('../utils/globals')
     return {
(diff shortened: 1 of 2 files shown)


pauldambra:
Asked a question about this in the private slack [posthog.slack.com/archives/C03C60FT1J7/p1726670676997629](https://posthog.slack.com/archives/C03C60FT1J7/p1726670676997629)

robbie-c:
Closing PR following discussion on slack

robbie-c:
See https://github.com/PostHog/posthog.com/pull/9408#pullrequestreview-2318570882 instead

PR #1430, opened 2024-09-23
chore: build when publishing

pauldambra:
We run tests when publishing that assume the lib was built...

So this should fix it


benjackwhite (changes requested):
Better would be to swap the order in the `prePublish` command of the package json tbh

benjackwhite (approved):

PR #1433, opened 2024-09-23
fix: get computed style error

pauldambra:
see https://posthog.sentry.io/issues/5858293571/events/latest/?project=1899813&query=is%3Aunresolved+issue.priority%3A%5Bhigh%2C+medium%5D+getComputedStyle&referrer=latest-event&statsPeriod=90d&stream_index=0&utc=true

and https://posthoghelp.zendesk.com/agent/tickets/17649 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/20566)

we very infrequently see a `Window.getComputedStyle: Argument 1 does not implement interface Element.` error. 

there are some small differences between similar routes in heatmaps and autocapture

the error seems to be in autocapture, so let's make heatmaps safer

### DIFF
--- a/src/autocapture.ts
+++ b/src/autocapture.ts
@@ -4,6 +4,7 @@ import {
     getClassNames,
     getDirectAndNestedSpanText,
     getElementsChainString,
+    getEventTarget,
     getSafeText,
     isAngularStyleAttr,
     isDocumentFragment,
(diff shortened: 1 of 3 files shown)


src/autocapture.ts
```diff
@@ -243,25 +244,13 @@ export class Autocapture {
         }
     }
 
-    private _getEventTarget(e: Event): Element | null {
```
pauldambra, line 246: this is useful dark magic, so we lift it out of the class so that heatmaps can use it

daibhin (approved):

PR #1436, opened 2024-09-24
feat: compress replay data

pauldambra:
We want to compress replay payloads since they're sometimes massive and very compressible

We do compress the entire payload when it is travelling over the network and when in storage, but at points in processing it is not compressed and we can leave some data compressed to make life easier on our infra

But blob ingestion needs to read some metadata from some events so we don't want to just compress everything

So, we now support posthog-js partially compressing selected payloads 

This only compresses when it is enabled so we can test this ourselves to see what the what

pairs with https://github.com/PostHog/posthog/pull/25183

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -125,10 +125,12 @@ export type compressedEvent =
 export type compressedEventWithTime = compressedEvent & {
     timestamp: number
     delay?: number
+    // marker for compression version
+    cv: '2024-10'
 }
 
 function gzipToString(data: unknown): string {
-    return strFromU8(gzipSync(strToU8(JSON.stringify(data))))
+    return strFromU8(gzipSync(strToU8(JSON.stringify(data))), true)
 }
 
 // rrweb's packer takes an event and returns a string or the reverse on unpact,
@@ -140,11 +142,13 @@ function compressEvent(event: eventWithTime, ph: PostHog): eventWithTime | compr
             return {
                 ...event,
                 data: gzipToString(event.data),
+                cv: '2024-10',
             }
         }
         if (event.type === EventType.IncrementalSnapshot && event.data.source === IncrementalSource.Mutation) {
             return {
                 ...event,
+                cv: '2024-10',
                 data: {
                     ...event.data,
                     texts: gzipToString(event.data.texts),
@@ -157,6 +161,7 @@ function compressEvent(event: eventWithTime, ph: PostHog): eventWithTime | compr
         if (event.type === EventType.IncrementalSnapshot && event.data.source === IncrementalSource.StyleSheetRule) {
             return {
                 ...event,
+                cv: '2024-10',
                 data: {
                     ...event.data,
                     adds: gzipToString(event.data.adds),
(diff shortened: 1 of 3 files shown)


marandaneto (approved):

src/extensions/replay/sessionrecording.ts
```diff
@@ -88,6 +89,95 @@ const newQueuedEvent = (rrwebMethod: () => void): QueuedRRWebEvent => ({
...
+export type compressedEventWithTime = compressedEvent & {
+    timestamp: number
+    delay?: number
+    // marker for compression version
+    cv: '2024-10'
+}
+
+function gzipToString(data: unknown): string {
+    return strFromU8(gzipSync(strToU8(JSON.stringify(data))), true)
+}
```
richard-better, line 134: Have you considered streaming, for at least the full snapshots? This seems memory intensive to do the Object->String->Array->Gzip->String transformation synchronously 

(Though I imagine it would need a bigger refactor to the network code)

benjackwhite: Streaming in the browser? 🤔 certainly not a thing I've ever heard of?

src/extensions/replay/sessionrecording.ts
```diff
@@ -88,6 +89,95 @@ const newQueuedEvent = (rrwebMethod: () => void): QueuedRRWebEvent => ({
...
+                data: gzipToString(event.data),
+                cv: '2024-10',
+            }
+        }
+        if (event.type === EventType.IncrementalSnapshot && event.data.source === IncrementalSource.Mutation) {
+            return {
+                ...event,
+                cv: '2024-10',
+                data: {
+                    ...event.data,
```
benjackwhite, line 153: Not sure what gains this really gives.
Why not pull out from the data any counters you need and then just compress the entire object? In theory it would _never_ need to be decompressed and can then be written straight to the S3 file for example and then decompressed at read time.
_Or_ at least it would be a simple single decompression of the whole payload rather than multiple ones?

benjackwhite (commented):
Couldn't help but comment on an interesting PR...

src/extensions/replay/sessionrecording.ts
richard-better: It's new-ish (at least for sending, reading the response as a stream has been available for a while): https://developer.chrome.com/docs/capabilities/web-apis/fetch-streaming-requests#using_with_writable_streams

src/extensions/replay/sessionrecording.ts
pauldambra: so... we know that when compressed the data fits into kafka but when uncompressed it doesn't always (and we reject it unnecessarily)

when I looked at pulling the metadata out (and it's fair I don't have this anywhere but in my head)

i at least had to change the SDK, django capture, blobby and maybe playback depending on how it was implemented

and we'd still have multiple compress/decompress steps since we're doing that at network boundaries anyway

this way i can test the impact with a relatively small change

src/extensions/replay/sessionrecording.ts
pauldambra: we'd not... that's interesting.
certainly we've talked about having a fast path for full snapshots at least since if all we capture are full snapshots we can play something back

src/extensions/replay/sessionrecording.ts
pauldambra: > what gains this really gives.

it should mean that the data that is currently not making into kafka from rust or django capture does make it in
without having to change anything else
so i can probe if that's the last piece of the puzzle for playback issues

pauldambra: #BiasForImpact :p

PR #1438, opened 2024-09-27
fix: no custom events when idle

pauldambra:
We sometimes see custom events extending idle sessions undesirably see https://posthoghelp.zendesk.com/agent/tickets/16379 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/29949) for an example

Since we only started allowing custom events during idle periods as a debug aid, let's just stop allowing them

We still allow session idle custom events, only to make sure they make it through. but no others


src/extensions/replay/sessionrecording.ts
```diff
@@ -655,6 +665,7 @@ export class SessionRecording {
                 })
                 // proactively flush the buffer in case the session is idle for a long time
                 this._flushBuffer()
+                this.isIdle = true
```
pauldambra, line 668: move this check to after flushing the buffer since we don't want to drop any of the things in the buffer because we're idle 🙈

marandaneto (approved):

src/extensions/replay/sessionrecording.ts
```diff
@@ -345,6 +360,13 @@ export class SessionRecording {
         this.windowId = windowId
 
         this.buffer = this.clearBuffer()
+
+        if (this.sessionIdleThresholdMilliseconds >= this.sessionManager.sessionTimeoutMs) {
+            logger.warn(
+                LOGGER_PREFIX +
+                    ` session_idle_threshold_ms (${this.sessionIdleThresholdMilliseconds}) is greater than the session timeout (${this.sessionManager.sessionTimeoutMs}). Session will never be detected as idle`
+            )
+        }
```
pauldambra, line 369: while testing this I discovered it was valid to set your max threshold to less than the idle threshold

you'd never go idle

and maybe you'd sit wondering why not 🤣 
so, we make both configurable and log when they're weird

src/extensions/replay/sessionrecording.ts
```diff
@@ -807,6 +834,11 @@ export class SessionRecording {
         this._tryAddCustomEvent('$posthog_config', {
             config: this.instance.config,
         })
+
+        logger.info(LOGGER_PREFIX + ' started', {
```
pauldambra, line 838: moved the log here since we were logging started twice when we weren't really just the previous location is idempotent and called multiple times

PR #1443, opened 2024-09-29
feat(web experiments): Emit web_experiment_applied event and do not render experiments for bots

Phanatic:
## Changes

1. Replace references to `className` with `css` which assigns the value of this field to the `style` attribute of an HTML element. 
2. Introduce the ability to `preview` an Experiment variant in the browser without deploying the experiment. 
3. Do not render web experiments if the viewer is a bot. 
4. Emit the $web_experiment_applied event so we know if the visual changes actually got applied to a page.

## Checklist

### DIFF
--- a/src/web-experiments.ts
+++ b/src/web-experiments.ts
@@ -83,8 +83,8 @@ export class WebExperiments {
         if (location?.search) {
             const experimentID = getQueryParam(location?.search, '__experiment_id')
             const variant = getQueryParam(location?.search, '__experiment_variant')
-            WebExperiments.logInfo(`previewing web experiments ${experimentID} && ${variant}`)
             if (experimentID && variant) {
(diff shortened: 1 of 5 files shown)


src/web-experiments.ts
```diff
@@ -248,14 +315,36 @@ export class WebExperiments {
...
         })
     }
+
+    _is_bot(): boolean | undefined {
+        if (navigator && this.instance) {
+            return isLikelyBot(navigator, this.instance.config.custom_blocked_useragents)
+        } else {
+            return undefined
+        }
+    }
```
dmarticus, line 349: useful!

dmarticus (approved):
seems reasonable, easy to follow, nice tests – all good from me!

jurajmajerik:
Sorry for not reviewing earlier, this is looking good!

PR #1445, opened 2024-09-30
feat: default replay partial compression on

pauldambra:
Partial compression of large rrweb events will make them more reliably ingestable

We've had this on for days now, and some users have tested it

Let's default it on

---

Clicked around in PH app and saved ~1MB in 2 minutes of recording

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -39,7 +39,9 @@ const BASE_ENDPOINT = '/s/'
 const FIVE_MINUTES = 1000 * 60 * 5
 const TWO_SECONDS = 2000
 export const RECORDING_IDLE_THRESHOLD_MS = FIVE_MINUTES
-export const RECORDING_MAX_EVENT_SIZE = 1024 * 1024 * 0.9 // ~1mb (with some wiggle room)
+const ONE_KB = 1024
+const PARTIAL_COMPRESSION_THRESHOLD = ONE_KB
+export const RECORDING_MAX_EVENT_SIZE = ONE_KB * ONE_KB * 0.9 // ~1mb (with some wiggle room)
 export const RECORDING_BUFFER_TIMEOUT = 2000 // 2 seconds
 export const SESSION_RECORDING_BATCH_KEY = 'recordings'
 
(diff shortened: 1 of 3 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -144,6 +146,11 @@ function gzipToString(data: unknown): string {
 // but we want to be able to inspect metadata during ingestion, and don't want to compress the entire event
 // so we have a custom packer that only compresses part of some events
 function compressEvent(event: eventWithTime, ph: PostHog): eventWithTime | compressedEventWithTime {
+    const originalSize = estimateSize(event)
+    if (originalSize < PARTIAL_COMPRESSION_THRESHOLD) {
+        return event
+    }
```
pauldambra, line 152: we pretty frequently get small enough snapshots that they're bigger when compressed so we can leave them be

marandaneto (approved):

kitsirota:
Hi @pauldambra,

We were going from 1.165.0 to 1.178.0 with a production release today, but we ended up being blocked by the console stack traces relating to undefined length across any event or page change.  I can confirm that without any changes to the our codebase, this PR results in endless console errors.

We actually had to roll a release back this afternoon, but after debugging this further we finally narrowed down the issue to this PR.  The errors here show up on every pageview, event, navigation, etc.
posthog-error

The workaround to proceed was to adjust our posthog.init statement to add compress_events: false
`        posthog.init(POSTHOG_KEY, {
            api_host: POSTHOG_API_HOST,
            ui_host: POSTHOG_UI_HOST,
            debug: POSTHOG_DEBUG,
            disable_surveys: POSTHOG_DISABLE_SURVEYS,
            secure_cookie: POSTHOG_SECURE_COOKIE,
            capture_pageview: POSTHOG_RECORD_PAGEVIEWS,
            capture_pageleave: POSTHOG_CAPTURE_PAGELEAVE,
            session_recording: {
                compress_events: false,
            },
        })`

Have you received any other reports of this issue?  Ive been going through different package versions for a while and the exceptions are showing up with this PR and all releases from 1.166.0

pauldambra:
hey @kitsirota 

We've not had any reports of that and it's been the default for a month now.

Are you able to open a ticket from in-app support? We'd need to take a look at the site and see what was happening.

There was a similar looking bug but that was fixed in 1.166.2 so shouldn't be possible in 1.178.0

PR #1451, opened 2024-10-03
feat: start session recording on url trigger

richard-better:
## Changes

Client part of https://github.com/PostHog/posthog/issues/25340

## Checklist


src/extensions/replay/sessionrecording.ts
```diff
@@ -343,13 +361,37 @@ export class SessionRecording {
...
+
+        if (TRIGGER_STATUSES.includes(currentValue)) {
+            return currentValue as TriggerStatus
+        }
+
+        return 'pending'
+    }
+
+    private set urlTriggerStatus(status: TriggerStatus) {
+        this.instance?.persistence?.register({
```
richard-better, line 390: both instance, and persistence are optional according to the types. Is it possible we don't have them set, should I add a fallback to a class property for that case?

src/extensions/replay/sessionrecording.ts
```diff
@@ -343,13 +361,37 @@ export class SessionRecording {
...
+            [SESSION_RECORDING_URL_TRIGGER_ACTIVATED]: status,
```
richard-better, line 391: What would be the best way to evict/expire this value when the recording session times out, so the new recordings have to hit the condition again?

src/extensions/replay/sessionrecording.ts
```diff
@@ -1090,6 +1144,36 @@ export class SessionRecording {
...
+        if (typeof window === 'undefined' || !window.location.href) {
+            return
+        }
+
+        const url = window.location.href
+
+        if (
+            this._urlTriggers.some((trigger) => {
+                switch (trigger.matching) {
+                    case 'regex':
```
richard-better, line 1157: Are we okay with just regex to start?

I looked at the ActionMatcher from surveys, but it seems it's doing a lot, we don't need yet. Would you prefer to integrate with that anyway, or is this fine for now?

src/extensions/replay/sessionrecording.ts
```diff
@@ -544,6 +586,10 @@ export class SessionRecording {
             })
         }
 
+        if (response.sessionRecording?.urlTriggers) {
```
richard-better, line 589: Is this something that should be configurable via the client as well?

src/extensions/replay/sessionrecording.ts
```diff
@@ -36,6 +44,7 @@ import { gzipSync, strFromU8, strToU8 } from 'fflate'
 
 const BASE_ENDPOINT = '/s/'
 
+const ONE_MINUTE = 1000 * 60
 const FIVE_MINUTES = 1000 * 60 * 5
```
pauldambra, line 48: very picky... and will have almost no impact individually but a little code golf helps the bundle size over time so you could have 

```suggestion
const FIVE_MINUTES = ONE_MINUTE * 5
```

i tend to dislike that style of writing and would rather fix the bigger problems in the bundle sizing but little things like that tend to help (although the bundler can behave surprisingly 🤣)

src/extensions/replay/sessionrecording.ts
```diff
@@ -343,13 +361,37 @@ export class SessionRecording {
...
         if (isBoolean(this.isSampled)) {
             return this.isSampled ? 'sampled' : 'disabled'
         } else {
             return 'active'
         }
     }
 
+    private get urlTriggerStatus(): 'activated' | 'pending' | 'disabled' {
+        if (this.receivedDecide && this._urlTriggers.length === 0) {
+            return 'disabled'
```
pauldambra, line 377: i guess disabled means url triggering is disabled and not recording is disabled... 

not blocking if you disagree - I'd be tempted to call it something like 'no-url-triggering'  to avoid (me) people getting confused in future but very very picky :)

src/extensions/replay/sessionrecording.ts
pauldambra: i keep meaning to test this... I _think_ that persistence will always at least be memory and so even though technically we might not have an instance and it might not have persistence we will always have them if recording is running

we do local flag fallback elsewhere to protect us but I _think_ it's actually unnecessary since I believe we always have memory persistence

(i've just never made time to check if that's true and simplify if so)

src/extensions/replay/sessionrecording.ts
pauldambra: we could write the status and the session id to storage to be super explicit...
otherwise there's an onsessionid handler registered elsewhere in this class and i bet it's sufficient to clear the status in that

src/extensions/replay/sessionrecording.ts
pauldambra: i think ultimately yes although I like to keep things remote only if there's a chance it will change

and my guess is this helps non-engineers more than engineers since it's a "no-code" trigger so client side maybe not super helpful for them anyway

src/extensions/replay/sessionrecording.ts
pauldambra: yeah, this is totally fine for now! 

way better to ship a little useful bit than delay while we figure out more complex behaviour! 💪

src/extensions/replay/sessionrecording.ts
```diff
@@ -909,11 +955,19 @@ export class SessionRecording {
...
+        this._checkUrlTrigger()
+
         // we're processing a full snapshot, so we should reset the timer
         if (rawEvent.type === EventType.FullSnapshot) {
             this._scheduleFullSnapshot()
         }
 
+        // Clear the buffer if waiting for a trigger, and only keep data from after the current full snapshot
+        if (rawEvent.type === EventType.FullSnapshot && this.urlTriggerStatus === 'pending') {
+            this.clearBuffer()
```
pauldambra, line 968: i guess this is still to-be-implemented behaviour?

(adding note just in case but assuming so since still draft)

pauldambra (commented):
this looks great 🚀 

didn't run it yet but 💯

src/extensions/replay/sessionrecording.ts
richard-better: Hmm, I was assuming this already does this, since the flow is essentiually: 

- We receive a new snapshot
- We clear the buffer
- We add the full snapshot to the buffer

Am I missing something here?

pauldambra: No, I was being silly! You're totally right

src/extensions/replay/sessionrecording.ts
```diff
@@ -906,11 +964,19 @@ export class SessionRecording {
             this._pageViewFallBack()
         }
 
+        // Check if the URL matches any trigger patterns
+        this._checkUrlTrigger()
```
pauldambra, line 968: we _could_ check this only when the URL changes (we have a hook in this class already)

or return early in `checkUrlTrigger` if there's no reason to run the check

wandering towards early optimisation though - not blocking at all

pauldambra (approved):
tested locally and this works now

PR #1452, opened 2024-10-03
feat: report reason for recording start

pauldambra:
in this private slack thread https://posthog.slack.com/archives/C06R7TKD508/p1727988694238459

A user has

* 10% sampling set on replay
* a batch export of events that identify interesting sessions
* every session has a session id regardless of it was recorded
* so the user has to check every session id one at a time to see which was recorded

adding this as an event has cost implication for users (if small)
and it is tricky to pick up a definitive start reasons for some flows as start is idempotent and we call it more than once

so, let's add a session super property so folk can look for that

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -34,6 +34,14 @@ import { isLocalhost } from '../../utils/request-utils'
 import { MutationRateLimiter } from './mutation-rate-limiter'
 import { gzipSync, strFromU8, strToU8 } from 'fflate'
 
+type SessionStartReason =
+    | 'sampling_override'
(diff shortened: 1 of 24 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -1099,5 +1105,12 @@ export class SessionRecording {
      * */
     public overrideLinkedFlag() {
         this._linkedFlagSeen = true
+        this._reportStarted('linked flag override')
```
marandaneto, line 1108:
```suggestion
        this._reportStarted('linked_flag_override')
```

just so that it matches the others

marandaneto (approved):

PR #1454, opened 2024-10-07
fix: Remove posthog error capture from replay

pauldambra:
We shouldn't be capturing auto posthog errors unless the integration is on... will remove this from replay for now


JakeRuth:
thanks for this and building in public! Excited to get this change in a new version of the psothogjs lib, it will really reduce spam in our live metrics

neilkakkar (approved):

PR #1456, opened 2024-10-07
fix: empty query param when calling external dependency toolbar.js

lucasra1:
## Changes

Changed the URL to use `t=<num>` as query param instead of empty/undefined param name (`v=161.1&=<num>`).

## Checklist

## Backstory

We use posthog in a nextjs application and use nextjs's rewrites as a proxy for better tracking. We noticed that the toolbar does not work on our site. when checking the logs we noticed that the rewrites did not work for the toolbar:

image

```
 code: 'ERR_INVALID_URL',
  input: 'https__ESC_COLON_//eu-assets.i.posthog.com/static/__ESC_COLON_path*'
```

But the `web-vitals.js` loaded without a problem...

image

Further investigation lead to the empty query param for the `toolbar.js` rotating token to bust  the cdn cache.

`?v=1.166.1?&=1728316200000`

The current stable version of next.js (`14.2.14`) has a "bug" that can't handle empty query params correctly in rewrites. The most recent canary version of nextjs seems to not have this bug anymore.

But still I think its better when posthog does not use an empty query param as this seems like something that was not intentional in the first place and secondly can result in unexpected behavior in frameworks like nextjs

### DIFF
--- a/src/__tests__/utils/external-scripts-loader.test.ts
+++ b/src/__tests__/utils/external-scripts-loader.test.ts
@@ -59,9 +59,7 @@ describe('external-scripts-loader', () => {
             assignableWindow.__PosthogExtensions__.loadExternalDependency(mockPostHog, 'toolbar', callback)
             const scripts = document!.getElementsByTagName('script')
             const new_script = scripts[0]
-            expect(new_script.src).toMatchInlineSnapshot(
-                `"https://us-assets.i.posthog.com/static/toolbar.js?v=1.0.0?&=1726067100000"`
-            )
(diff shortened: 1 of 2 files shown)


lucasra1:
Please keep open and review. Thanks

lucasra1:
Please keep open and review. Thanks

pauldambra (approved):

pauldambra:
fixes #1458

lucasra1:
@pauldambra I think the github workflow needs to be approved. https://github.com/PostHog/posthog-js/actions/runs/11551538865

pauldambra:
ugh... that should really work once merged 🤔

PR #1460, opened 2024-10-10
fix(errors): Better define schema, align with python

neilkakkar:
## Changes

...

## Checklist

### DIFF
--- a/src/extensions/exception-autocapture/error-conversion.ts
+++ b/src/extensions/exception-autocapture/error-conversion.ts
@@ -11,7 +11,42 @@ import {
 import { defaultStackParser, StackFrame } from './stack-trace'
 
 import { isEmptyString, isNumber, isString, isUndefined } from '../../utils/type-utils'
-import { ErrorEventArgs, ErrorProperties, SeverityLevel, severityLevels } from '../../types'
+import { ErrorEventArgs, SeverityLevel, severityLevels } from '../../types'
+
+export interface ErrorProperties {
+    $exception_type: string
+    $exception_message: string
+    $exception_level: SeverityLevel
+    $exception_list?: Exception[]
+    // $exception_source?: string
+    // $exception_lineno?: number
+    // $exception_colno?: number
+    $exception_DOMException_code?: string
+    $exception_is_synthetic?: boolean
+    // $exception_stack_trace_raw?: string
+    $exception_handled?: boolean
+    $exception_personURL?: string
+    $exception_language?: string
+}
+
+export interface Exception {
+    type?: string
+    value?: string
+    mechanism?: {
+        handled?: boolean
+        type?: string
+    }
+    module?: string
+    thread_id?: number
+    stacktrace?: {
+        frames?: StackFrame[]
+    }
+}
+
+export interface ErrorConversions {
+    errorToProperties: (args: ErrorEventArgs) => ErrorProperties
+    unhandledRejectionToProperties: (args: [ev: PromiseRejectionEvent]) => ErrorProperties
+}
 
 /**
  * based on the very wonderful MIT licensed Sentry SDK
@@ -59,7 +94,15 @@ function errorPropertiesFromError(error: Error): ErrorProperties {
     return {
         $exception_type: error.name,
         $exception_message: error.message,
-        $exception_stack_trace_raw: JSON.stringify(frames),
+        $exception_list: [
+            {
+                type: error.name,
+                value: error.message,
+                stacktrace: {
+                    frames,
+                },
+            },
+        ],
         $exception_level: 'error',
     }
 }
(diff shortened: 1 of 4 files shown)


src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -11,7 +11,42 @@ import {
...
+    // $exception_source?: string
+    // $exception_lineno?: number
+    // $exception_colno?: number
+    $exception_DOMException_code?: string
+    $exception_is_synthetic?: boolean
+    // $exception_stack_trace_raw?: string
+    $exception_handled?: boolean
+    $exception_personURL?: string
+    $exception_language?: string
+}
```
daibhin, line 30: One for the future but would be cool to have a shared schema similar to what we have for TS <-> Python we have for queries

src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -177,13 +221,7 @@ export function errorToProperties([event, source, lineno, colno, error]: ErrorEv
         $exception_level: isSeverityLevel(errorProperties.$exception_level)
             ? errorProperties.$exception_level
             : 'error',
-        ...(source
-            ? {
-                  $exception_source: source, // TODO get this from URL if not present
-              }
-            : {}),
-        ...(lineno ? { $exception_lineno: lineno } : {}),
-        ...(colno ? { $exception_colno: colno } : {}),
```
daibhin, line 186: Let's remove these from the other places they're referenced in a follow up PR

daibhin (approved):

src/extensions/exception-autocapture/error-conversion.ts
neilkakkar: will rm the commented out stuff 👀

src/extensions/exception-autocapture/error-conversion.ts
daibhin: I think they're mentioned in the new Rust service. They're also in the taxonomy descriptions but that's less important because arguably they should stick around there...

neilkakkar: oh it's optional, all good, need it to stay there to support old versions that might send events 💀 , but yeah we don't need to use this at all imo

src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -11,7 +11,42 @@ import {
...
+    $exception_personURL?: string
+    $exception_language?: string
+}
+
+export interface Exception {
+    type?: string
+    value?: string
+    mechanism?: {
+        handled?: boolean
+        type?: string
```
neilkakkar, line 37:
```suggestion
        type?: string
        source?: string
        synthetic?: boolean
```

src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -11,7 +11,42 @@ import {
...
+    }
+    module?: string
+    thread_id?: number
```
neilkakkar, line 40:
```suggestion
    thread_id?: number
    language?: string
```

src/extensions/exception-autocapture/error-conversion.ts
```diff
@@ -177,13 +224,7 @@ export function errorToProperties([event, source, lineno, colno, error]: ErrorEv
         $exception_level: isSeverityLevel(errorProperties.$exception_level)
             ? errorProperties.$exception_level
             : 'error',
-        ...(source
-            ? {
-                  $exception_source: source, // TODO get this from URL if not present
```
daibhin, line 182: I came here to ask if we should get rid of this... ahead of me :)

neilkakkar:
have changed things a decent amount, now the only top level prop is $exception_list and sometimes $exception_level..

src/posthog-core.ts
```diff
@@ -1831,9 +1831,17 @@ export class PostHog {
...
-                  $exception_type: error.name,
-                  $exception_message: error.message,
                   $exception_level: 'error',
+                  $exception_list: [
+                      {
+                          type: error.name,
+                          value: error.message,
+                          mechanism: {
+                              handled: true,
+                              synthetic: false,
```
daibhin, line 1841: Should this be `synthetic: true`? We don't know for certain if it's just a string

daibhin (approved):

src/posthog-core.ts
neilkakkar: captureException only takes an error as a param though, not string 🤔 - and in this case we can't do stack resolution either, so not sure how we'd handle the synthetic case.. `error.name` would be undefined as well.

src/extensions/sentry-integration.ts
```diff
@@ -97,6 +97,7 @@ export function createEventProcessor(
             // added manually to avoid any dependency on the lazily loaded content
             $exception_message: any
             $exception_type: any
```
neilkakkar, line 99: i haven't yet removed $exception_message from the sentry integration, but eventually yes will do.

src/posthog-core.ts
daibhin: Won't that only apply to people using Typescript? I guess someone could give us a string but I'm not even sure how we'd handle that right now... Let's revisit it once we have a few more examples. This can be the happy path for now 👍

PR #1463, opened 2024-10-12
feat: dead click detection

pauldambra:
opening to see what CI thinks of this


src/__tests__/extensions/dead-clicks-autocapture.test.ts
```diff
@@ -0,0 +1,70 @@
+import { PostHog } from '../../posthog-core'
+import { assignableWindow } from '../../utils/globals'
+import { createPosthogInstance } from '../helpers/posthog-instance'
+import { uuidv7 } from '../../uuidv7'
+import { DeadClicksAutocapture } from '../../extensions/dead-clicks-autocapture'
+
+describe('DeadClicksAutocapture', () => {
```
pauldambra, line 7: this outer wrapper delegates entirely to the lazy loaded one so we can test it interacts with a mock as desired and ignore the behaviour

src/autocapture.ts
```diff
@@ -37,6 +37,202 @@ function limitText(length: number, text: string): string {
     return text
 }
 
+export function getAugmentPropertiesFromElement(elem: Element): Properties {
```
pauldambra, line 40: a bunch of functions move out of the class so we can use them directly because they (mostly) don't rely on state of the class

it should help with bundle size a tiny bit since the names are now minifiable

react/src/context/PostHogProvider.tsx
pauldambra, line 1: I'll yank this to another PR... it was stopping the nextjs playground app work for me

robbie-c:
Does selecting text trigger this? It's not a dom mutation or a scroll, but it's not dead either

src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -0,0 +1,209 @@
...
+                this._checkClicks()
+            }, 1000)
+        }
+    }
+
+    private _startScrollObserver() {
+        // setting the third argument to `true` means that we will receive scroll events for other scrollable elements
+        // on the page, not just the window
+        // see https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#usecapture
+        assignableWindow.addEventListener('scroll', this._onScroll, true)
```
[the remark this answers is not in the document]
pauldambra: I _think_ we don't need to - we already track scrolls this way elsewhere
(but I'll double-check!)

pauldambra:
@robbie-c you get 2 hedgehog points for spotting that! fix incoming

src/entrypoints/dead-clicks-autocapture.ts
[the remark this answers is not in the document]
pauldambra: > You don't need to worry about the value of passive for the basic [scroll](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll_event) event. Since it can't be canceled, event listeners can't block page rendering anyway.

no, we're good here... although we should check through if any of the event listeners in posthog-js would benefit from being passive 👍

src/autocapture.ts
```diff
@@ -37,6 +37,202 @@ function limitText(length: number, text: string): string {
...
+export function getDefaultProperties(eventType: string): Properties {
+    return {
+        $event_type: eventType,
+        $ce_version: 1,
+    }
+}
+
+export function getPropertiesFromElement(
+    elem: Element,
+    maskAllAttributes: boolean,
```
pauldambra, line 81: renamed this parameter which has always been wrong

robbie-c:
Related to text selection, I guess any interaction with form elements (e.g. clicking a select input) wouldn't change the dom but isn't dead either

pauldambra:
> Related to text selection, I guess any interaction with form elements (e.g. clicking a select input) wouldn't change the dom but isn't dead either

pauldambra:
> Related to text selection, I guess any interaction with form elements (e.g. clicking a select input) wouldn't change the dom but isn't dead either

fair point... made any autocapturable element not eligible for dead click detection

src/extensions/dead-clicks-autocapture.ts
```diff
@@ -0,0 +1,91 @@
+import { PostHog } from '../posthog-core'
+import { DEAD_CLICKS_ENABLED_SERVER_SIDE } from '../constants'
+import { isBoolean, isObject } from '../utils/type-utils'
+import { assignableWindow, document, LazyLoadedDeadClicksAutocapture } from '../utils/globals'
+import { logger } from '../utils/logger'
+import { DecideResponse } from '../types'
+
+const LOGGER_PREFIX = '[Dead Clicks]'
+
+export class DeadClicksAutocapture {
```
pauldambra, line 10: am also thinking about https://github.com/PostHog/posthog-js/pull/1439 when making this structure.
we have almost exactly the same logic repeated for every lazy loaded extension

i'm interested in extending this simple wrapper on lazy loaded extension to replay (would save up to 13% of bundle)  and then maybe abstracting it so every lazy loaded extension could be added the same way

richard-better (approved):

cypress/e2e/capture.cy.ts
benjackwhite, line 1: Adding cypress tests - you are an inspiration to us all 🙌

src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -0,0 +1,269 @@
...
+        }
+    }
+    return null
+}
+
+function checkTimeout(value: number | undefined, thresholdMs: number) {
+    return isNumber(value) && value >= thresholdMs
+}
+
+class _LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture {
```
benjackwhite, line 46: Nit - I prefer the C# style `ILazy...` to indicate an interface. For our code style maybe 
```suggestion
class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocaptureInterface {
```

benjackwhite (approved):
Didn't look at the implentation of the dead click detection tooooo closely but did a general pass through and I think the approach and changes all look good

src/entrypoints/dead-clicks-autocapture.ts
pauldambra: I used to make C# interfaces into a sentence `IAmABankAccount` 
So.... careful what you wish for

PR #1467, opened 2024-10-14
chore: refactor some cypress setup

pauldambra:
It always bothered me when I copy-pastied a new script into the cypress setup

well, no more


cypress/support/setup.ts
```diff
@@ -24,7 +24,7 @@ export const start = ({
     // sometimes we have too many listeners in this test environment
     // that breaks the event emitter listeners in error tracking tests
     // we don't see the error in production, so it's fine to increase the limit here
-    EventEmitter.prototype._maxListeners = 100
```
pauldambra, line 27: this is equivalent but one is less naughty

PR #1468, opened 2024-10-15
feat: swap the default to identified_only

raquelmsmith:
> [!warning]
> This should only be merged after we support the field response from /decide.

## Changes

We want to make the `person_profiles: identified_only` config the default for everyone. It will save everyone money - even if they don't change their config - and is better for our customers and our pipeline. 

Instead of just pushing out a change to posthog-js for this, we want to be able to control the behavior and roll it out slowly, just in case there are any problems. 

So, this PR changes the default, but also listens to a field from decide that will say if we should be defaulting people into `identified_only`, or using the old default `always`. If someone has set their config manually, then we will continue using that setting.

## Checklist

### DIFF
--- a/src/__tests__/posthog-core.ts
+++ b/src/__tests__/posthog-core.ts
@@ -345,6 +345,20 @@ describe('posthog core', () => {
 
             expect(posthog.compression).toEqual('gzip-js')
         })
+        it('uses defaultIdentifiedOnly from decide response', () => {
+            const posthog = posthogWith({})
+
+            posthog._afterDecideResponse({ defaultIdentifiedOnly: true } as DecideResponse)
+            expect(posthog.config.person_profiles).toEqual('identified_only')
+
+            posthog._afterDecideResponse({ defaultIdentifiedOnly: false } as DecideResponse)
+            expect(posthog.config.person_profiles).toEqual('always')
+        })
+        it('defaultIdentifiedOnly does not override person_profiles if already set', () => {
+            const posthog = posthogWith({ person_profiles: 'always' })
+            posthog._afterDecideResponse({ defaultIdentifiedOnly: true } as DecideResponse)
+            expect(posthog.config.person_profiles).toEqual('always')
+        })
 
         it('enables compression from decide response when only one received', () => {
             const posthog = posthogWith({})
--- a/src/__tests__/posthog-core.ts
+++ b/src/__tests__/posthog-core.ts
@@ -387,6 +387,7 @@ describe('posthog core', () => {
                 properties: () => ({ distinct_id: 'abc', persistent: 'prop', $is_identified: false }),
                 remove_event_timer: jest.fn(),
                 get_property: () => 'anonymous',
+                props: { $groups: {} },
             } as unknown as PostHogPersistence,
             sessionPersistence: {
                 properties: () => ({ distinct_id: 'abc', persistent: 'prop' }),
@@ -425,7 +426,7 @@ describe('posthog core', () => {
                 $window_id: 'windowId',
                 $session_id: 'sessionId',
                 $is_identified: false,
-                $process_person_profile: true,
+                $process_person_profile: false,
             })
         })
 
@@ -447,7 +448,7 @@ describe('posthog core', () => {
                 $session_id: 'sessionId',
                 $lib_custom_api_host: 'https://custom.posthog.com',
                 $is_identified: false,
-                $process_person_profile: true,
+                $process_person_profile: false,
             })
         })
 
@@ -465,7 +466,7 @@ describe('posthog core', () => {
                 posthog._calculate_event_properties('custom_event', { event: 'prop' }, new Date())[
                     '$process_person_profile'
                 ]
-            ).toEqual(true)
+            ).toEqual(false)
         })
 
         it('only adds token and distinct_id if event_name is $snapshot', () => {
@@ -496,7 +497,7 @@ describe('posthog core', () => {
             expect(posthog._calculate_event_properties('custom_event', { event: 'prop' }, new Date())).toEqual({
                 event_name: 'custom_event',
                 token: 'testtoken',
-                $process_person_profile: true,
+                $process_person_profile: false,
             })
         })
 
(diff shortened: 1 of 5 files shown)


src/__tests__/posthog-core.ts
```diff
@@ -387,6 +401,7 @@ describe('posthog core', () => {
                 properties: () => ({ distinct_id: 'abc', persistent: 'prop', $is_identified: false }),
                 remove_event_timer: jest.fn(),
                 get_property: () => 'anonymous',
+                props: { $groups: {} },
```
raquelmsmith, line 404: adding this made most the tests pass with the new `identified_only` default config. Does this indicate that something else needs to happen somewhere to make sure real users always have this in their props? Or is it just a test fixture thing? (I imagine the latter because people can use `identifed_only` with no problem, but want to be certain)

src/posthog-core.ts
```diff
@@ -179,7 +179,7 @@ export const defaultConfig = (): PostHogConfig => ({
     bootstrap: {},
     disable_compression: false,
     session_idle_timeout_seconds: 30 * 60, // 30 minutes
-    person_profiles: 'always',
+    person_profiles: 'identified_only',
```
raquelmsmith, line 182: the main change for swapping the default

src/posthog-core.ts
```diff
@@ -390,6 +392,10 @@ export class PostHog {
         this.config = {} as PostHogConfig // will be set right below
         this._triggered_notifs = []
 
+        if (config.person_profiles) {
+            this._initialPersonProfilesConfig = config.person_profiles
+        }
```
raquelmsmith, line 397: in order to roll this out slowly, we need to know if the person has already set their config - that it's not just using the default. So we need to store the initial config somewhere.

src/posthog-core.ts
```diff
@@ -408,6 +414,8 @@ export class PostHog {
             this.config.persistence === 'sessionStorage'
                 ? this.persistence
                 : new PostHogPersistence({ ...this.config, persistence: 'sessionStorage' })
+
+        // should I store the initial person profiles config in persistence?
```
raquelmsmith, line 418: question

src/types.ts
```diff
@@ -398,6 +398,7 @@ export interface DecideResponse {
     isAuthenticated: boolean
     siteApps: { id: number; url: string }[]
     heatmaps?: boolean
+    defaultIdentifiedOnly?: boolean
```
raquelmsmith, line 401: this will be returned with `/decide`

src/posthog-core.ts
```diff
@@ -538,6 +546,12 @@ export class PostHog {
             this.analyticsDefaultEndpoint = response.analytics.endpoint
         }
 
+        this.config.person_profiles = this._initialPersonProfilesConfig
+            ? this._initialPersonProfilesConfig
+            : response['defaultIdentifiedOnly']
+            ? 'identified_only'
+            : 'always'
```
raquelmsmith, line 553: This is the bit that responds to a value passed from /decide. But I'm not sure if mutating the config like this is allowed....

robbie-c: Thinking out loud, but it might be better to do this with `set_config`, as that's a clear place for future code to listen for this change and respond to it

src/__tests__/posthog-core.ts
robbie-c: It feels like the latter, but I'll stare at this code a little to convince myself a bit more

robbie-c:
It's really nice that we're changing the default, but there's a slight caveat with using the decide response to allow us to roll this out gradually.

Once someone has sent an event that requires person processing, we persist a flag called `ENABLE_PERSON_PROCESSING` to ensure that future events also have person processing.

This PR does it the way around that will work btw :) but e.g. if we kept the default as `always` and then used decide to put people on `identified_only`, that most likely wouldn't work, as the first pageview event would be sent with person processing before the decide response arrives.

src/posthog-core.ts
robbie-c: The way I've handled this so far is to save whether or not they've sent an event with person processing. If they have then we persist `ENABLE_PERSON_PROCESSING` to true.

I'm just wondering whether saving their config gives us anything beyond this.

Here's a couple of cases:

Case 1
* posthog-js starts up with the user setting 'always'
* some events are sent, with person profiles, so `ENABLE_PERSON_PROCESSING` is set
* user closes the browser
* new visit, posthog-js starts up with no setting so uses the default `identified_only`
* some events are sent, as `ENABLE_PERSON_PROCESSING` was set then these have person profiles

Case 2
* posthog-js starts up with the user setting 'always'
* no events are sent, so `ENABLE_PERSON_PROCESSING` is not set
* user closes the browser
* new visit, posthog-js starts up with no setting so uses the default `identified_only`
* some events are sent, these are anonymous (if we'd persisted the setting from before, they'd have person profiles)

In case 2 there's a difference, but I'd argue that those events should be sent as anonymous. Wdyt?

robbie-c: TLDR no I don't think so :D

robbie-c: Ooooook so it turns out we don't actually handle this case correctly. I added a fix and some tests to this PR. I hope you don't mind me piggybacking like this, I can very much move it to its own PR if you prefer.

src/__tests__/posthog-core.ts
robbie-c: Ok, it was the absence of `props` in the mock that made the test fail, it didn't actually need `$groups: {}`, so I've slightly changed this line to have a more minimal mock

robbie-c (approved):
LGTM! I added some stuff, so please make sure those changes LGTY

src/posthog-core.ts
raquelmsmith: No this is great, thanks!

src/posthog-core.ts
raquelmsmith: Sweet I pushed up a change to use that. Thanls!

raquelmsmith:
Looks good to me! Thanks for the help @robbie-c!!

raquelmsmith:
Oh @robbie-c I forgot to specify what type of version bump this is.... will it bump automatically? Or just with the next one?

PR #1472, opened 2024-10-15
use synthetic stack trace if not originally captured

daibhin:
## Changes

Why miss out on a stack trace if you call `captureException` with a string?

This changes things to throw an error to capture the current stack trace and strips the final frame (e.g. inside our `captureException` function)

### DIFF
--- a/src/extensions/exception-autocapture/stack-trace.ts
+++ b/src/extensions/exception-autocapture/stack-trace.ts
@@ -51,7 +51,6 @@ export interface StackFrame {
 }
 
 const WEBPACK_ERROR_REGEXP = /\(error: (.*)\)/
-const STRIP_FRAME_REGEXP = /captureException/
 const STACKTRACE_FRAME_LIMIT = 50
 
 const UNKNOWN_FUNCTION = '?'
@@ -209,14 +208,6 @@ export function reverseAndStripFrames(stack: ReadonlyArray<StackFrame>): StackFr
     const localStack = Array.from(stack)
 
     localStack.reverse()
-
-    if (STRIP_FRAME_REGEXP.test(getLastStackFrame(localStack).function || '')) {
-        localStack.pop()
-
-        if (STRIP_FRAME_REGEXP.test(getLastStackFrame(localStack).function || '')) {
-            localStack.pop()
-        }
-    }
 
     return localStack.slice(0, STACKTRACE_FRAME_LIMIT).map((frame) => ({
         ...frame,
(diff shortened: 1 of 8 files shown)


src/extensions/exception-autocapture/stack-trace.ts
```diff
@@ -51,7 +51,6 @@ export interface StackFrame {
 }
 
 const WEBPACK_ERROR_REGEXP = /\(error: (.*)\)/
-const STRIP_FRAME_REGEXP = /captureException/
```
daibhin, line 54: This is a superior approach but I couldn't get it to work. The function name I was seeing was `t.value` in the stack trace. Perhaps there's some Javascript function construction magic needed to ensure the function name gets caught in the stack. Open to any thoughts if people have them?

src/posthog-core.ts
```diff
@@ -1826,14 +1826,15 @@ export class PostHog {
 
     /** Capture a caught exception manually */
     captureException(error: Error, additionalProperties?: Properties): void {
```
neilkakkar, line 1828: do we need to update the type here as well now?

src/extensions/exception-autocapture/stack-trace.ts
neilkakkar: probably because array.js is minified? Or the loaded extensions are if we're using the npm version of main package..

neilkakkar (approved):
nice

src/posthog-core.ts
daibhin: I was torn... technically I'd still like people to send us errors but it is a little misleading to ourselves to think that the input will always be an error. What I really want is to type it as `unknown` but for a user to be shown `Error` as the type

neilkakkar:
oh wait, this doesn't work well for the case where currently we're not seeing stack traces, it's a different captureException function that's called: https://github.com/PostHog/posthog-js/blob/master/src/extensions/exception-autocapture/index.ts#L76

src/extensions/exception-autocapture/stack-trace.ts
daibhin: Probably true - very meta given our current work 😂

daibhin:
@neilkakkar you're right, this actually only fixes the issue where the stack trace is missing after a user manually called `captureException` with a string / object that does not have a stack

The global `onerror` and `unhandledrejection` you linked should always have an error and thus not need a synthetic exception The `captureException` you linked is actually the callback used in the wrapped functions ([here](https://github.com/PostHog/posthog-js/blob/main/src/entrypoints/exception-autocapture.ts#L15) and [here](https://github.com/PostHog/posthog-js/blob/main/src/entrypoints/exception-autocapture.ts#L36)) to parse that error.

The case where the stack trace is missing from a genuine error is probably more likely related to this issue: https://sentry.io/answers/script-error/

PR #1276, opened 2024-06-26
chore: upgrade rrweb to alpha.16

daibhin:
## Changes

rrweb moved to vite bundling so the patch file will look different this time around

Previous patch file: https://github.com/PostHog/posthog-js/blob/v1.141.4/patches/rrweb%402.0.0-alpha.13.patch

----

I skipped the patch replacing the inline worker with the base64 worker
https://github.com/PostHog/posthog-js/blob/27b044c616efc8adeb6b3d1766ea5a1f25cd4708/patches/rrweb%402.0.0-alpha.13.patch#L5-L164

It looks like the migration to Vite means that the base64 worker is back. From the built `rrweb.js` file in the /node_modules you can see:
```
[...]

PR #1475, opened 2024-10-16
chore: skip if Array.from is missing

daibhin:
## Changes

When upgrading rrweb in https://github.com/PostHog/posthog-js/pull/1276 we had some [failing tests on IE11](https://github.com/PostHog/posthog-js/actions/runs/11342401043/job/31542713953?pr=1276):

```
TypeError: Object doesn't support property or method 'from'
```

This is because the `Array.from` method is [not supported on IE11](https://caniuse.com/?search=array%20from) and it seems like rrweb internals have changed such that it is now called as part of our tests when upgrading.

Proposal is to remove it in this PR and then upgrade rrweb in the separate PR linked

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -610,13 +610,13 @@ export class SessionRecording {
     }
 
     private _startCapture() {
-        if (isUndefined(Object.assign)) {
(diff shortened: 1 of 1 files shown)


pauldambra (approved):
🥇

PR #1476, opened 2024-10-16
chore: test that autocapture allowlists are unioned

pauldambra:
see this question in https://posthog.slack.com/archives/C02E3BKC78F/p1729035026457899

> are the allowlist settings for autocapture unioned

i wasn't sure of the answer so let's add a test to show that they are


pauldambra:
will merge this in without review since it's a test, but happy to revisit if folk spot something

PR #1478, opened 2024-10-16
feat: safer string truncation

pauldambra:
We truncate strings by slicing them at a given length both in replay and in general capture property processing

`String.prototype.slice()` and `String.prototype.substring()` are not unicode aware 

so

```
const str = 'A 😄 B';
console.log(str.slice(2, 3)); // Outputs broken part of emoji
```

Major browsers and Node 20 have `toWellFormed`

Our CI/tests/etc don't use Node 20 so even though we could use toWellFormed in prod it would be a pain to add. Moving to Node20 would probably fix some people's builds and break others. Let's avoid confusion for now.

Luckily the one consistent thing about JS is that it is not consistent.

`Array.from(someString)` is unicode aware and correctly splits a string into an array of valid characters. So we can then slice that array. To get a string that is sort-of the right length.

Our string truncation in the JS is set arbitrarily so it doesn't matter that this mechanism returns strings that may not have a `.length` that exactly matches the `maxLength` passed to truncation


pauldambra:
@daibhin since this would maybe help with your upgrade PR...

I'm considering adding a (lazy loaded?) polyfill for array.from

daibhin (approved):
Interestingly I'm still seeing failures on IE11 in https://github.com/PostHog/posthog-js/pull/1276 despite my changes :/ Regardless, this seems good

pauldambra:
ok, corejs array.fom polyfill adds 11% to the bundle... that's no bueno

pauldambra:
interesting... looks like the array.from polyfill in IE11 bundle makes this timeout 🤯

PR #1480, opened 2024-10-16
chore: build an es5 bundle and move main to es6

pauldambra:
stacked on top of #1479 

not supporting IE11 is a 10% bundle saving
so, we should move anyone that needs to support IE11 to a new bundle


daibhin (approved):
The size savings here are incredible!

robbie-c (approved):

PR #1490, opened 2024-10-23
feat(web-analytics): Add flag to send server hash instead of distinct id

robbie-c:
## Changes
Part of https://github.com/PostHog/posthog/issues/25117
Main repo PR (currently WIP, this PR can be merged before that one) https://github.com/PostHog/posthog/pull/25915

Everything is behind a config option with a dire warning, so this doesn't need to be final.

## Checklist

### DIFF
--- a/playground/nextjs/pages/_app.tsx
+++ b/playground/nextjs/pages/_app.tsx
@@ -42,7 +42,7 @@ export default function App({ Component, pageProps }: AppProps) {
                 <meta name="viewport" content="width=device-width, initial-scale=1" />
                 {/* CSP - useful for testing our documented recommendations. NOTE: Unsafe is only needed for nextjs pre-loading */}
                 <meta
-                    http-equiv="Content-Security-Policy"
+                    httpEquiv="Content-Security-Policy"
                     content={`
                     default-src 'self';
                     connect-src 'self' ${localhostDomain} https://*.posthog.com https://lottie.host;
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -97,8 +97,8 @@ type SessionRecordingStatus = 'disabled' | 'sampled' | 'active' | 'buffering' | 
 export interface SnapshotBuffer {
     size: number
     data: any[]
-    sessionId: string | null
-    windowId: string | null
+    sessionId: string
+    windowId: string
 }
 
 interface QueuedRRWebEvent {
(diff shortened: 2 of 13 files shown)


playground/nextjs/pages/_app.tsx
```diff
@@ -42,7 +42,7 @@ export default function App({ Component, pageProps }: AppProps) {
                 <meta name="viewport" content="width=device-width, initial-scale=1" />
                 {/* CSP - useful for testing our documented recommendations. NOTE: Unsafe is only needed for nextjs pre-loading */}
                 <meta
-                    http-equiv="Content-Security-Policy"
```
robbie-c, line 45: driveby

src/extensions/replay/sessionrecording.ts
```diff
@@ -97,8 +97,8 @@ type SessionRecordingStatus = 'disabled' | 'sampled' | 'active' | 'buffering' |
 export interface SnapshotBuffer {
     size: number
     data: any[]
-    sessionId: string
-    windowId: string
+    sessionId: string | null
```
robbie-c, line 100: I'm not expecting session replay to work in cklsh mode at the moment, that would follow later

benjackwhite: If you aren't expecting it to work - why change the types here? Without a sessionId, replay should basically not run...

src/posthog-core.ts
```diff
@@ -915,6 +924,13 @@ export class PostHog {
         let properties = { ...event_properties }
         properties['token'] = this.config.token
 
+        if (this.config.__preview_experimental_cookieless_mode) {
+            // Set a flag to tell the plugin server to use cookieless server hash mode
+            // on naming: $cklsh = cookieless server hash
+            // I didn't call it $cookieless, as it's possible to use cookies in this mode (after consent is given)
+            properties['$cklsh'] = true
```
benjackwhite, line 931: Given this is a magic prop could we move it to constants and give it a nice name?

benjackwhite (changes requested):
Small comments - nothing blocking althouh intrigued about the server side given I wasn't much aware of this 😅

src/extensions/replay/sessionrecording.ts
pauldambra: yeah, i did quite a bit of rearranging so that I could make sure session id was never null. i don't want to lose that typecheck without adding safety back in. it might be fine here in this PR but could let a bug sneak in in a future PR

(disclaimer: i have not looked at the code just this thread)

robbie-c: Cool, I've changed this to prevent the session id manager or session replay from loading when using this mode, and reverted the type changes

rafaeelaudibert (approved):
lgtm

src/posthog-core.ts
robbie-c: Done

benjackwhite (approved):

benjackwhite (approved):

PR #1493, opened 2024-10-23
feat: Only set missing campaign params to null if there is at least one non-null

robbie-c:
## Changes

See this ticket https://posthoghelp.zendesk.com/agent/tickets/19698 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/21794)

We definitely want to keep something similar to the changed logic. i.e. if one session has utm_source = source, and the next session has utm_medium = medium.

However, we could prevent overwriting the campaign params if there are none set.

Had a discussion with @neilkakkar about it here https://posthog.slack.com/archives/C05LJK1N3CP/p1729764244044729

## Checklist

### DIFF
--- a/src/posthog-persistence.ts
+++ b/src/posthog-persistence.ts
@@ -12,7 +12,7 @@ import {
     PERSISTENCE_RESERVED_PROPERTIES,
 } from './constants'
 
-import { isObject, isUndefined } from './utils/type-utils'
+import { isEmptyObject, isObject, isUndefined } from './utils/type-utils'
 import { Info } from './utils/event-utils'
 import { logger } from './utils/logger'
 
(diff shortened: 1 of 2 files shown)


src/posthog-persistence.ts
```diff
@@ -221,7 +221,11 @@ export class PostHogPersistence {
 
     update_campaign_params(): void {
         if (!this.campaign_params_saved) {
-            this.register(Info.campaignParams(this.config.custom_campaign_params))
+            const campaignParams = Info.campaignParams(this.config.custom_campaign_params)
+            // only save campaign params if there were any
+            if (!isEmptyObject(stripEmptyProperties(campaignParams))) {
+                this.register(Info.campaignParams(this.config.custom_campaign_params))
+            }
             this.campaign_params_saved = true
```
pauldambra, line 229: should we only be marked them as saved when we did send something or doesn't matter?

pauldambra (approved):

PR #1494, opened 2024-10-23
fix(surveys): Reset Survey storage after posthog.reset is called

Phanatic:
## Changes

We found through a customer issue that calling `posthog.reset()` doesn't remove the tracking properties added by PostHog Surveys from local storage. This PR adds a `reset` method to posthog surveys and calls it from posthog-core, whenever `posthog.reset()` is called.

## Checklist


dmarticus (approved):
nice work!

PR #1495, opened 2024-10-24
chore: set platform on stack frame

daibhin:
## Changes

We need to know what language an individual frame is so that we can avoid an untagged enum switching as part of the Rust service.

For now this assumes all stacks in the JS SDK are `javascript`

### DIFF
--- a/src/extensions/exception-autocapture/stack-trace.ts
+++ b/src/extensions/exception-autocapture/stack-trace.ts
@@ -33,6 +33,7 @@ export type StackLineParserFn = (line: string) => StackFrame | undefined
 export type StackLineParser = [number, StackLineParserFn]
 
 export interface StackFrame {
+    lang: string
     filename?: string
     function?: string
     module?: string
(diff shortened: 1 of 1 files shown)


src/extensions/exception-autocapture/stack-trace.ts
```diff
@@ -33,6 +33,7 @@ export type StackLineParserFn = (line: string) => StackFrame | undefined
 export type StackLineParser = [number, StackLineParserFn]
 
 export interface StackFrame {
+    lang: string
```
neilkakkar, line 36: i think this is same as `platform` (and useful to re-use imo when we extend because s£ntry SDKs _should_ populate this)

daibhin: Hmm good point. The docs aren't super clear
> This can override the platform for a single frame. Otherwise, the platform of the event is assumed. This can be used for multi-platform stack traces, such as in React Native.

My worry with platform is that it ends up populating something like "iPadOS", in which case we don't know whether it is a Swift or Objective-C frame 🤔

neilkakkar (approved):

src/extensions/exception-autocapture/stack-trace.ts
daibhin: Looking through the code it seems like `platform` is set to the language or something like `node`. In our instance the "platform" values are only ever `javascript` or `python`.

Will rename this and go with `platform` to maintain consistency but use serde to map from platform to lang when deserializing

PR #1496, opened 2024-10-24
fix: allow canvas local config

pauldambra:
A user wants to be able to dynamically configure canvas recording... so, allow local config

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -303,15 +303,19 @@ export class SessionRecording {
         return enabled_client_side ?? enabled_server_side
     }
 
-    private get canvasRecording(): { enabled: boolean; fps: number; quality: number } | undefined {
+    private get canvasRecording(): { enabled: boolean; fps: number; quality: number } {
+        const canvasRecording_client_side = this.instance.config.session_recording.captureCanvas
         const canvasRecording_server_side = this.instance.get_property(SESSION_RECORDING_CANVAS_RECORDING)
-        return canvasRecording_server_side && canvasRecording_server_side.fps && canvasRecording_server_side.quality
-            ? {
-                  enabled: canvasRecording_server_side.enabled,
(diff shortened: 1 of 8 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -303,15 +303,19 @@ export class SessionRecording {
...
-            ? {
-                  enabled: canvasRecording_server_side.enabled,
-                  fps: canvasRecording_server_side.fps,
-                  quality: canvasRecording_server_side.quality,
-              }
-            : undefined
+
+        const enabled = canvasRecording_client_side?.recordCanvas ?? canvasRecording_server_side?.enabled ?? false
+        const fps = canvasRecording_client_side?.canvasFps ?? canvasRecording_server_side?.fps ?? 0
+        const quality = canvasRecording_client_side?.canvasQuality ?? canvasRecording_server_side?.quality ?? 0
```
daibhin, line 312: I would be careful about letting users arbitrarily set these two values. We could end up with very large recordings if they set the values too high, particularly the FPS. At the very least I would limit them to sensible ranges and emit a warning log

daibhin (changes requested):

src/extensions/replay/sessionrecording.ts
pauldambra: ah... fair. 

have a feeling for reasonable range?

pauldambra: fps... between 0 and 12?
quality between 0 and 1?

daibhin: I was very conservative rolling it out initially 😂 

https://github.com/PostHog/posthog/blob/2a0a953fe91fb43d51f0f86fd0fae0a747db3017/posthog/api/decide.py#L373-L375

They sound reasonable. Unlikely many people will change the defaults. We can always dial it back in future SDK versions if it causes issues

daibhin (approved):
Nice 👌

PR #1499, opened 2024-10-25
chore: add recorder to es5 bundle

pauldambra:
we've had a customer request to support replay in the ES5 bundle so they can try replay in IE11

rrweb doesn't officially support IE11, so 🤞


pauldambra:
it's not going to work to add recorder into our ES5 bundle

PR #1500, opened 2024-10-28
feat: add recording url blocklist

richard-better:
## Changes

Client for https://github.com/PostHog/posthog/pull/25845 
Enables pausing the recording when certain urls are visited.

## Checklist

### DIFF
--- a/src/__tests__/extensions/replay/sessionrecording.test.ts
+++ b/src/__tests__/extensions/replay/sessionrecording.test.ts
@@ -45,6 +45,7 @@ import {
 } from '@rrweb/types'
 import Mock = jest.Mock
 import { ConsentManager } from '../../../consent'
+import { waitFor } from '@testing-library/preact'
 
 // Type and source defined here designate a non-user-generated recording event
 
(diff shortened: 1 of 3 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -1198,6 +1225,34 @@ export class SessionRecording {
...
+    private _pauseRecording() {
+        if (this.status === 'paused') {
+            return
+        }
+
+        this._urlBlocked = true
+
+        this.clearBuffer()
+        clearInterval(this._fullSnapshotTimer)
+        this._tryAddCustomEvent('recording paused', { reason: 'url blocker' })
```
pauldambra, line 1237: i think the _urlBlocked above means we don't get this custom event because we return early 

so, we could add the custom event and then set the flag... 

makes me wonder about the buffer clear here as well
i don't know that we have a good way to do this but... we're dropping unemitted data here without knowing it's to do with the blocked page maybe

so i'm thinking what will happen is

* add some data to the buffer by moving their mouse and clicking a button
* it isn't big or old enough so we don't send it
* user navigates to the blocked page because of the button click
* we clear the buffer and lose the mouse data we didn't send yet

i don't think we have a good way of doing this but in a perfect world we'd send all of the data before the blocked page in the buffer and clear everything after it

(should err on the side of caution while we don't have a good way of being sure though but would be nice to lose as little as possible)

src/extensions/replay/sessionrecording.ts
pauldambra, line 1: would be good to add/update tests otherwise this looks good to me!

src/extensions/replay/sessionrecording.ts
```diff
@@ -980,7 +1009,11 @@ export class SessionRecording {
         }
 
         // Check if the URL matches any trigger patterns
-        this._checkUrlTrigger()
+        this._checkTriggerConditions()
+
+        if (this.status === 'paused' && !isRecordingPausedEvent(rawEvent)) {
```
richard-better, line 1014: So I've added this to make sure this one event can go through even if the recording is paused, but... it seems like `onRRwebEmit` never gets called with this event?

src/extensions/replay/sessionrecording.ts
```diff
@@ -1198,6 +1232,39 @@ export class SessionRecording {
...
+        if (this.status !== 'paused') {
+            return
+        }
+
+        this._urlBlocked = false
+
+        this._tryTakeFullSnapshot()
+
+        this._scheduleFullSnapshot()
+        this._tryAddCustomEvent('recording resumed', { reason: 'left blocked url' })
```
richard-better, line 1264: Same deal as the paused event, this one never makes it to `onRRwebEmit` 🤔 
@pauldambra, do you have any idea what I'm doing wrong here?

src/extensions/replay/sessionrecording.ts
richard-better: Asked a question re the custom event in a different thread, but I've changed to around, and from my testing it was reliably sending all events before hitting the block condition.

src/extensions/replay/sessionrecording.ts
pauldambra: i ran it locally and these get called for me (in tests iirc there's no real rrweb so custom events don't work - could it be that)

Screenshot 2024-10-31 at 11 43 20

i got lots of paused events... don't know if it's because i had breakpoints set or if i'm getting a paused on every mutation?

src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -2156,4 +2157,77 @@ describe('SessionRecording', () => {
...
+                    '$snapshot',
+                    {
+                        $session_id: sessionId,
+                        $window_id: 'windowId',
+                        $snapshot_bytes: expect.any(Number),
+                        $snapshot_data: [
+                            { type: 3, data: { source: 1 } },
+                            { type: 3, data: { source: 2 } },
+                            // This should be in the buffer
+                            // { type: 5, data: { tag: 'recording paused', payload: { reason: 'url blocker' } } },
```
pauldambra, line 2200: ah, yes, this'll be because mocked rrweb i bet... may well be a way we can improve the mock to pass along custom events - although feel free to not spend too much time on that

if I really care then i put tests into cypress so i'm exercising a real rrweb

pauldambra (approved):
this works great!

i did get a flash of the blocked page as i navigated away

i think i'm ok telling people they should combine masking and blocking since its best effort and then improve it over time

pauldambra:
really silly scope creep idea... we could add ph-no-capture to the body and remove it again as we hit/leave no-record pages 🤔

pauldambra:
further scope creep... "skipping inactivity" while we're on a blocked page is the wrong player message.
since we have the custom events 🙌 the player can be aware it's on a blocked page and change that message for 💯 🦔 points

PR #1504, opened 2024-10-29
feat: Add timezone to events

robbie-c:
## Changes

Add timezone to events (timezone looks like Europe/London).

In older browsers that don't support Intl, this property won't be sent.

This is useful for a couple of things:
* I want to add a time of day query, which shows you how active your site is at different times of day, and it'd be cool to have a toggle to show it by the user's timezone (so you could see that e.g. around the world your app is always popular at 8am)
* I can use this for the cookieless rotating salt, to figure out which day's salt to use

## Checklist

### DIFF
--- a/src/utils/event-utils.ts
+++ b/src/utils/event-utils.ts
@@ -182,6 +182,14 @@ export const Info = {
         return props
     },
 
+    timezone: function (): string | undefined {
+        try {
+            return Intl.DateTimeFormat().resolvedOptions().timeZone
+        } catch {
+            return undefined
+        }
+    },
+
     properties: function (): Properties {
         if (!userAgent) {
             return {}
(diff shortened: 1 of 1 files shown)


pauldambra (approved):
this is great... IMHO the solution to passing dates around and knowing what's happening

src/utils/event-utils.ts
```diff
@@ -194,6 +202,7 @@ export const Info = {
                 $browser: Info.browser(userAgent, navigator.vendor),
                 $device: Info.device(userAgent),
                 $device_type: Info.deviceType(userAgent),
+                $timezone: Info.timezone(),
```
marandaneto, line 205: @robbie-c would be interesting to add `$locale` as well
did not test but something like:
```
function getBrowserLocale(): string | undefined {
    try {
        return navigator.language || navigator.languages[0];
    } catch {
        return undefined;
    }
}
```

PR #1505, opened 2024-10-30
chore: refactor towards dead clicks autocapture

pauldambra:
Lifts a set of refactors out of https://github.com/PostHog/posthog-js/pull/1463/

That PR Introduces dead clicks autocapture but has a consistent failing cypress test for surveys which should not have been impacted


Phanatic (approved):
Transfering approvals

PR #1507, opened 2024-10-31
chore: add the dead clicks autocapture without initialising it

pauldambra:
https://github.com/PostHog/posthog-js/pull/1463 adds deadclicks functionality but a surveys cypress test consistently fails on that PR and we don't know why

this PR brings the implementation of dead clicks autocapture out of that PR without wiring it up to see if the surveys test fails when the code is present but not used


pauldambra:
this PR has served its discovery purpose

PR #1508, opened 2024-10-31
fix: dead click fixes from watching in prod

pauldambra:
turned on dead clicks in prod
watched what we captured realised some things could be improved

### DIFF
--- a/src/entrypoints/dead-clicks-autocapture.ts
+++ b/src/entrypoints/dead-clicks-autocapture.ts
@@ -4,7 +4,7 @@ import { isNull, isNumber, isUndefined } from '../utils/type-utils'
 import { autocaptureCompatibleElements, getEventTarget } from '../autocapture-utils'
 import { DeadClicksAutoCaptureConfig, Properties } from '../types'
 import { autocapturePropertiesForElement } from '../autocapture'
-import { isDisabledElement, isElementInToolbar, isElementNode, isTag } from '../utils/element-utils'
+import { isElementInToolbar, isElementNode, isTag } from '../utils/element-utils'
 
 const DEFAULT_CONFIG: Required<DeadClicksAutoCaptureConfig> = {
     element_attribute_ignorelist: [],
@@ -154,10 +154,6 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
         }
 
         if (isElementInToolbar(click.node)) {
-            return true
-        }
-
-        if (isDisabledElement(click.node)) {
             return true
         }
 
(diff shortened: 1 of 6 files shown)


src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -157,6 +157,10 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
             return true
         }
 
+        if (isDisabledElement(click.node)) {
```
pauldambra, line 160: we captured a dead click after someone clicked on a disabled button, that feels wrong to me

src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -205,7 +209,9 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
                 this._config.selection_change_threshold_ms
             )
             const mutationTimeout = checkTimeout(click.mutationDelayMs, this._config.mutation_threshold_ms)
-            const absoluteTimeout = checkTimeout(click.absoluteDelayMs, this._config.mutation_threshold_ms)
+            // we want to timeout eventually even if nothing else catches it...
+            // we leave a little longer than the maximum threshold to give the other checks a chance to catch it
+            const absoluteTimeout = checkTimeout(click.absoluteDelayMs, this._config.mutation_threshold_ms * 1.1)
```
pauldambra, line 214: we saw absolute timeouts, without seeing mutation timeouts, that also seems wrong so...

src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -244,22 +250,27 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
...
+                    // TRICKY: it appears that we were moving to elementsChainAsString, but the UI still depends on elements, so :shrug:
+                    elementsChainAsString: false,
+                }).props,
+                $dead_click_scroll_delay_ms: click.scrollDelayMs,
+                $dead_click_mutation_delay_ms: click.mutationDelayMs,
+                $dead_click_absolute_delay_ms: click.absoluteDelayMs,
+                $dead_click_selection_changed_delay_ms: click.selectionChangedDelayMs,
+            },
+            {
+                timestamp: new Date(click.timestamp),
```
pauldambra, line 271: turns out this is how you set the timestamp on an event

src/utils/element-utils.ts
```diff
@@ -1,6 +1,63 @@
...
+}
+
+/*
+ * Check whether an element has nodeType Node.DOCUMENT_FRAGMENT_NODE
+ * @param {Element} el - element to check
+ * @returns {boolean} whether el is of the correct nodeType
+ */
+export function isDocumentFragment(el: Element | ParentNode | undefined | null): el is DocumentFragment {
+    return !!el && el.nodeType === 11 // Node.DOCUMENT_FRAGMENT_NODE - use integer constant for browser portability
+}
```
pauldambra, line 63: moved these from auto capture code since we now have an element utils for them to live in

src/entrypoints/dead-clicks-autocapture.ts
pauldambra: I can live with leaving it as is for now 👍

PR #1509, opened 2024-11-02
feat: allow decide to provide script name for recorder

pauldambra:
we want to allow specifying the name of the recorder script so we can roll out changes without affecting everyone all at once

this lets us use the decide response to specify the script name for the lazy loaded recorder


pauldambra:
towards https://github.com/PostHog/posthog/issues/25721

richard-better (approved):

PR #1510, opened 2024-11-03
feat: deadclicks in heatmaps

pauldambra:
we want to allow deadclicks collection in heatmaps

it gives you less information but doesn't cost money, folk that want more information can turn on standalone deadclicks capture

to keep things loose the heatmap starts its own instance of deadclicks capture, deadclicks was altered to receive varying behviour by constructor injection

---

running locally I can capture deadclicks correctly with both deadclicks and heatmaps enabled or either one enabled

### DIFF
--- a/src/extensions/dead-clicks-autocapture.ts
+++ b/src/extensions/dead-clicks-autocapture.ts
@@ -26,7 +26,7 @@ export class DeadClicksAutocapture {
     constructor(
         readonly instance: PostHog,
         readonly isEnabled: (dca: DeadClicksAutocapture) => boolean,
-        readonly onCapture: DeadClicksAutoCaptureConfig['__onCapture']
+        readonly onCapture?: DeadClicksAutoCaptureConfig['__onCapture']
     ) {
         this.startIfEnabled()
     }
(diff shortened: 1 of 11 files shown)


src/extensions/dead-clicks-autocapture.ts
```diff
@@ -36,9 +36,14 @@ export class DeadClicksAutocapture {
...
+    public startIfEnabled({
+        force,
+        onCapture,
+    }: { force?: boolean; onCapture?: (click: DeadClickCandidate, properties: Properties) => void } = {}) {
+        if (this.isEnabled || force) {
+            this.loadScript(() => {
+                this.start(onCapture)
+            })
         }
     }
```
pauldambra, line 48: not totally convinced by this signature change

passing in `onCapture` allows only one capture per instance and is at least explicit
passing in `force` to enable the capture feels messy... you could pass in the `isEnabled` method instead 🤷

src/utils/element-utils.ts
```diff
@@ -1,8 +1,8 @@
-import { TOOLBAR_ID } from '../constants'
+import { TOOLBAR_CONTAINER_CLASS, TOOLBAR_ID } from '../constants'
 
 export function isElementInToolbar(el: Element): boolean {
     // NOTE: .closest is not supported in IE11 hence the operator check
-    return el.id === TOOLBAR_ID || !!el.closest?.('#' + TOOLBAR_ID)
```
pauldambra, line 5: this was only working previously by accident because there is also `ph-no-capture` which autocapture obeys

`closest` does not escape shadow roots and TOOLBAR_ID is outside the shadow root.

switched to a selector inside the shadow root (the selector provided is at the root of the toolbar inside the shadow root

e.g.

```
<div id="TOOLBAR_ID">
    <shadow_root>
       <div class=TOOLBAR_CONTAINER_CLASS>
```

richard-better (approved):

PR #1512, opened 2024-11-05
fix: rotate session id proactively

pauldambra:
We reviewed the session idle timeout code and some of the naming was off, so this changes the names

We also periodically get reports that some events sneak through after many tens of hours of idle activity before the session id rotates. This is very surprising from reading the code but is clearly happening, let's make it (even more) impossible by:

* every time we see activity
* set a timeout to reset the session id after slightly more than the idle timeout
* if we see activity again we clear and recreate that timeout
* otherwise we reset the session id and any return from idle will create a new session id and not be able to leak to the old one

----

What I'm worried about here after talking it through with @robbie-c  in person is that the code makes sense but we don't know it will fix the problem since we don't know what the problem is 🤔


src/sessionid.ts
```diff
@@ -162,14 +165,14 @@ export class SessionIdManager {
...
         }
-        const sessionId = this.persistence.props[SESSION_ID]
+        const sessionIdInfo = this.persistence.props[SESSION_ID]
 
-        if (isArray(sessionId) && sessionId.length === 2) {
+        if (isArray(sessionIdInfo) && sessionIdInfo.length === 2) {
             // Storage does not yet have a session start time. Add the last activity timestamp as the start time
-            sessionId.push(sessionId[0])
+            sessionIdInfo.push(sessionIdInfo[0])
         }
```
pauldambra, line 173: NB we can probably remove this upgrade code now but I won't do it here, a lot of time of passed since we added this

marandaneto (approved):

marandaneto:
> What I'm worried about here after talking it through with @robbie-c in person is that the code makes sense but we don't know it will fix the problem since we don't know what the problem is 🤔

I agree but, we know the problem exists, we cannot reproduce it, so every small interaction may lead to fixing it.
We can always roll back if it worsens it.

PR #1515, opened 2024-11-08
feat: allow config of before_send function to edit or reject events

pauldambra:
see RFC: https://github.com/PostHog/product-internal/pull/668/
see docs PR: https://github.com/PostHog/posthog.com/pull/9844

The number of config options to redact urls or edit properties is growing
It is simpler to have a single hook where you can edit an event before it is captured

This also allows the enterprising customer configure a sampling function

---

tested locally and event capture still works without a beforeSend set
and e.g. the pre-canned event sampling function works as expected


src/__tests__/utils/before-capture-utils.test.ts
```diff
@@ -0,0 +1,57 @@
...
+describe('before capture utils', () => {
+    it('can sample by event name', () => {
+        const sampleFn = sampleByEvent(['$autocapture'], 50)
+        const results = []
+        Array.from({ length: 100 }).forEach(() => {
+            const captureResult = { event: '$autocapture' } as unknown as CaptureResult
+            results.push(sampleFn(captureResult))
+        })
+        const emittedEvents = results.filter((r) => !isNull(r))
+        expectRoughlyFiftyPercent(emittedEvents)
```
robbie-c, line 19: Can we mock Math.random for this so that it won't fail (very) occasionally?
```
import prand from 'pure-rand';
const mockRandom = (seed=42) => {
    const rng = prand.xoroshiro128plus(seed);
    return () => prand.unsafeUniformIntDistribution(0, 1_000_000, rng) / 1_000_000
}
```

src/posthog-core.ts
```diff
@@ -787,7 +793,11 @@ export class PostHog {
      * @param {String} [config.transport] Transport method for network request ('XHR' or 'sendBeacon').
      * @param {Date} [config.timestamp] Timestamp is a Date object. If not set, it'll automatically be set to the current time.
      */
-    capture(event_name: string, properties?: Properties | null, options?: CaptureOptions): CaptureResult | undefined {
+    capture(
+        event_name: EventName,
+        properties?: Properties | null,
+        options?: CaptureOptions
```
marandaneto, line 799:
```suggestion
        options?: CaptureOptions | null
```

do we need this here too?

src/types.ts
```diff
@@ -5,9 +5,68 @@ import type { SegmentAnalytics } from './extensions/segment-integration'
...
+ */
+export type KnownUnEditableEvent = typeof knownUnEditableEvent[number]
+
+export const knownUnsafeEditableEvent = [
+    '$pageview',
+    '$pageleave',
+    '$identify',
+    '$set',
+    '$groupidentify',
+    '$create_alias',
```
marandaneto, line 32: I wonder if those should be `knownUnEditableEvent` since they are called max 1 time per session or if something changes, usually isn't a huge amount of events, and missing them would cause "I don't find my users" type of support tickets.

src/types.ts
```diff
@@ -214,7 +270,11 @@ export interface PostHogConfig {
     feature_flag_request_timeout_ms: number
     get_device_id: (uuid: string) => string
     name: string
+    /**
+     * this is a read-only function that can be used to react to event capture
+     */
     _onCapture: (eventName: string, eventData: CaptureResult) => void
+    beforeCapture: (cr: CaptureResult) => CaptureResult | null
```
marandaneto, line 277: Semantically speaking, the fn is executed after capture (the fn) or the act of capturing an event.
And the fn is executed before enqueuing/flushing the event.
I wonder if the naming would not cause any confusion.
`beforeEnqueuing` is strange but sounds semantically correct, anyway, just giving food for thought, your call =)

src/__tests__/utils/before-capture-utils.test.ts
robbie-c: Ha your new approach also works!

src/types.ts
pauldambra: `beforeSend`??

marandaneto: that makes sense to me, sentry uses [beforeSend](https://docs.sentry.io/platforms/javascript/configuration/filtering/#using-before-send) as well

src/posthog-core.ts
```diff
@@ -870,6 +880,21 @@ export class PostHog {
             this.setPersonPropertiesForFlags(finalSet)
         }
 
+        if (!isKnownUnEditableEvent(data.event)) {
```
pauldambra, line 883: ah, you all missed my deliberate mistake 😅 

this stops us processing these events at all... which is correct for sampling but not for masking

so... 

* all events should be processed
* isKnownUnEditableEvent reject changes which are nullish (including empty object)
* others allow those changes but might warn
* empty object should be treated as nullish
* we should still be validating the shape of the events after processing so we can warn if people do something unexpected

pauldambra: actually other than $set that still makes sense to me

src/utils/before-capture.utils.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Provides an implementation of sampling that samples based on the distinct ID.
+ * Using the provided percentage.
+ * Can be used to create a beforeCapture fn for a PostHog instance.
+ *
+ * Causes roughly 50% of distinct ids to have events sent.
+ * Not 50% of events for each distinct id.
+ *
+ * @param percent a number from 0 to 100, 100 means never sample, 0 means never send the event
+ */
+export function sampleByDistinctId(percent: number): (c: CaptureResult) => CaptureResult | null {
```
robbie-c, line 28: I'm wondering if we should have a different folder for these utils (ones we expect customers to use).

I need something similar for a customer, so I'm also wondering where the best place for that would be.

robbie-c: Here's my PR https://github.com/PostHog/posthog-js/pull/1517

I put them in a folder called `customizations` which people can import from

src/posthog-core.ts
pauldambra: I don't _think_ so... I'll pull on the thread and see if anything is sending null but I didn't change that here iirc so 🤞

src/posthog-core.ts
pauldambra: so...

if we say it's not editable then we don't let you edit this at all
you can still use sanitize props

we warn if you remove all properties
$set and $set_once can be undefined so very hard to validate them

src/types.ts
```diff
@@ -5,9 +5,64 @@ import type { SegmentAnalytics } from './extensions/segment-integration'
...
+    '$feature_enrollment_update',
+    '$feature_flag_called',
+    'survey dismissed',
+    'survey sent',
+    'survey shown',
+    '$snapshot',
+    '$identify',
+    '$groupidentify',
+    '$create_alias',
+] as const
```
benjackwhite, line 22: I don't get it 😅 Why are we being restrictive here? The whole point of this function is to give ultimate flexibility to solve any case that someone could come up with. What if I want to ensure certain properties are filtered out in an identify? What if I want to mask survey sent events to try and remove PII?

I get that these are "important events" but why don't we just document that / provide the list for people to use if they want to. The warning for the $pageview etc feels fine but why not just add these to the warning?

I guarantee someone will have a use case to apply to all events and we will have to add an extra param like "includeUnsafeEvents" or something silly...

src/types.ts
```diff
@@ -214,7 +269,15 @@ export interface PostHogConfig {
...
     name: string
+    /**
+     * this is a read-only function that can be used to react to event capture
+     */
     _onCapture: (eventName: string, eventData: CaptureResult) => void
+    /**
+     * This function - if provided - is called immediately before sending data to the server.
+     * It allows you to edit data before it is sent, or choose not to send it all.
+     */
+    beforeSend: (cr: CaptureResult) => CaptureResult | null
```
benjackwhite, line 280: Rather `before_send` or? I know `_onCapture` is not following the pattern but that feels like a mistake.

benjackwhite (commented):
Generally I don't think my comments are blocking, just felt like there is some decent complexity here in the restrictions that I personally don't get.

src/types.ts
pauldambra: well, at least some of these are billing related, no?
and $snapshot they definitely can't touch

i'm totally not against moving events between the groups

pauldambra: > and $snapshot they definitely can't touch

well,,, $snapshot they mustn't _drop_ and i'd argue shouldn't edit

pauldambra: > at least some of these are billing related, no?

🤯 although of course if you sample the survey sent event... then we don't bill you for it, but that's ok because you didn't send it 😅

pauldambra: as always you were 99% right 🥇

pauldambra: OK, I'm down to only $snapshot events are ignored here

So if you wanted to mask e.g. all URLs. You would have to give us before_send and a mask function in replay

Or we let all events through here and it can be powerful or a foot gun... 🤔

src/types.ts
```diff
@@ -214,7 +265,15 @@ export interface PostHogConfig {
     feature_flag_request_timeout_ms: number
     get_device_id: (uuid: string) => string
     name: string
+    /**
+     * this is a read-only function that can be used to react to event capture
+     */
     _onCapture: (eventName: string, eventData: CaptureResult) => void
+    /**
+     * This function - if provided - is called immediately before sending data to the server.
+     * It allows you to edit data before it is sent, or choose not to send it all.
```
marandaneto, line 274: Not really immediately before sending but rather before enqueuing unless you flip the send_mmediately flag right?

src/types.ts
```diff
@@ -214,7 +265,15 @@ export interface PostHogConfig {
...
+     */
+    before_send: (cr: CaptureResult) => CaptureResult | null
```
marandaneto, line 276: Should we deprecate `_onCapture`? I mean before send does the same, if you don't want to mutate anything, just dont.

src/types.ts
marandaneto: Feels like `sanitize_properties` would also be redundant.

pauldambra: > Not really immediately before sending but rather before enqueuing unless you flip the send_mmediately flag right?

yep, went with hopefully clear to the user over technically correct 🙈

src/customizations/before-send.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Provides an implementation of sampling that samples based on the distinct ID.
+ * Using the provided percentage.
+ * Can be used to create a beforeCapture fn for a PostHog instance.
+ *
+ * Causes roughly 50% of distinct ids to have events sent.
+ * Not 50% of events for each distinct id.
+ *
+ * @param percent a number from 0 to 100, 100 means never sample, 0 means never send the event
```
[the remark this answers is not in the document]
pauldambra: yeah, i've always assumed it was laziness because math.random returns 0 to 1 rather than purposeful

but then this is laziness because the implementation underneath checks 0 to 100

I don't know what's best here... I guess if someone doesn't look at the signature then they'll send 0.1 expecting 10% and won't get that and we can't warn them because 0.1 is a valid number 🤔

src/types.ts
pauldambra: I guess we should, yes. 

I was leaning away to let us test more widely but that's overly cautious 👍

src/customizations/before-send.ts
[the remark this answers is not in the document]
marandaneto: > I don't know what's best here... I guess if someone doesn't look at the signature then they'll send 0.1 expecting 10% and won't get that and we can't warn them because 0.1 is a valid number 🤔

that was exactly what I was afraid of but it's well-documented, not sure if auto-pilot reads comment blocks tho :D

src/types.ts
marandaneto: I mean we don't need to remove anything, and an @deprecated annotation is just an annotation anyway, but we can do the annotation later once we are more confident as well, that's fine.

pauldambra: 🤣 turns out deprecated is a github user 🤣 

👋 

----

deprecated _onCapture and moved all the test usages over to before_send

src/customizations/before-send.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Provides an implementation of sampling that samples based on the event name.
+ * Using the provided percentage.
+ * Can be used to create a beforeCapture fn for a PostHog instance.
+ *
+ * @param eventNames an array of event names to sample, sampling is applied across events not per event name
+ * @param percent a number from 0 to 100, 100 means never sample, 0 means never send the event
+ */
+export function sampleByEvent(
```
robbie-c, line 58: Just another random thought (apologies for scope creep).

When you're sampling events, it's really useful to include the sample threshold as a property, e.g. `$sample_threshold`. Then when you're trying to figure out the approximate unsampled number of events, you can do `sum(1 / $sample_threshold)`, and that works even if the threshold changes over time.

For a percentage it'd be `sum(100 / $sample_percent)`

src/posthog-core.ts
```diff
@@ -181,6 +184,8 @@ export const defaultConfig = (): PostHogConfig => ({
     session_idle_timeout_seconds: 30 * 60, // 30 minutes
     person_profiles: 'identified_only',
     __add_tracing_headers: false,
+    // by default the before capture fn is the identity function
```
benjackwhite, line 187: This comment makes no sense to me 😅

benjackwhite: Also why not make it optional and have nothing set? Feels clearer to me than having an instant return function

src/posthog-core.ts
```diff
@@ -870,6 +880,25 @@ export class PostHog {
             this.setPersonPropertiesForFlags(finalSet)
         }
 
+        const beforeSendResult = this.config.before_send(data)
```
benjackwhite, line 883: Not sure why but my gut says it would be better to make this optional and only invoke if set. Thinking ahead to where we _might_ track config settings better and it would be good to know if someone has explicitly implemented it or not

benjackwhite (approved):

src/posthog-core.ts
pauldambra: where's your functional programming elitism?! the identity function just returns what you give it

it meant I didn't have to do any check to decide whether to call it, like we do iirc with `_onCapture` being a `__NOOP` fn

src/posthog-core.ts
pauldambra: yep, on reflection, for support we'd want to know it was configured or not

pauldambra:
@robbie-c how about allowing an array of functions...

then we can do 

```
before_send: [
	sampleByDistinctId(0.5), // only half of people
	sampleByEvent(['$web_vitals], 0.1), // and they capture all events except 10% of web vitals
]
```

pauldambra:
updated docs PR here https://github.com/PostHog/posthog.com/pull/9844/files#diff-e9e9ef37e4e756e5cf3bc804e89db87a8e60864b4a39d43c61c311cd0ae13eb6R583-R596

robbie-c:
Makes sense, would you multiply the sample thresholds together?

src/customizations/before-send.ts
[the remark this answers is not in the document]
pauldambra: switched it to 0 to 1

PR #1516, opened 2024-11-08
fix(flags): support multiple children prop in PostHogFeature

havenbarnes:
## Changes

Closes https://github.com/PostHog/posthog/issues/22922

## Checklist


havenbarnes:
@neilkakkar hello! I see you looked at this ticket a bit when it first came in - would love your thoughts on this when you get back. Would we prefer a non-backwards compat change like this, or perhaps split brained logic that removes the wrapper div _only if_ multiple children are passed in?

dmarticus (approved):
this seems cool to me; nice work!  Don't forget to the add the `bump:  $version` label to this PR to automatically bump and publish the latest version of these changes once they've been merged.

PR #1517, opened 2024-11-11
feat: Add customization to add all person profile properties as setPersonPropertiesForFlags

robbie-c:
## Changes
Add a customization to set all known person properties with setPersonPropertiesForFlags

## Checklist

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,10 +1,10 @@
-import { version } from '../package.json'
+import packageInfo from '../package.json'
 
 // overridden in posthog-core,
 // e.g.     Config.DEBUG = Config.DEBUG || instance.config.debug
 const Config = {
     DEBUG: false,
-    LIB_VERSION: version,
+    LIB_VERSION: packageInfo.version,
 }
 
 export default Config
--- a/src/customizations/setAllPersonProfilePropertiesAsPersonPropertiesForFlags.test.ts
+++ b/src/customizations/setAllPersonProfilePropertiesAsPersonPropertiesForFlags.test.ts
@@ -11,6 +11,10 @@ jest.mock('../utils/globals', () => {
         ...orig,
         mockURLGetter,
         mockReferrerGetter,
+        userAgent: 'Mozilla/5.0 (Android 4.4; Mobile; rv:41.0) Gecko/41.0 Firefox/41.0',
+        navigator: {
+            vendor: '',
+        },
         document: {
             ...orig.document,
             createElement: (...args: any[]) => orig.document.createElement(...args),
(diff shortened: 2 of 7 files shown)


src/config.ts
```diff
@@ -1,10 +1,10 @@
-import { version } from '../package.json'
+import packageInfo from '../package.json'
```
robbie-c, line 1: I got a warning about this, so I fixed it as a drive-by

`Should not import the named export 'version' (imported as 'version') from default-exporting module (only default export is available soon)`

src/customizations/setAllPersonProfilePropertiesAsPersonPropertiesForFlags.test.ts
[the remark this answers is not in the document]
robbie-c: Ah I'll move it

src/utils/event-utils.ts
```diff
@@ -31,6 +31,25 @@ export const CAMPAIGN_PARAMS = [
     'rdt_cid', // reddit
 ]
 
+export const EVENT_TO_PERSON_PROPERTIES = [
+    // mobile params
+    '$app_build',
+    '$app_name',
+    '$app_namespace',
+    '$app_version',
```
marandaneto, line 39: There are many more params than that such as `$device_manufacturer`, `$device_model`, etc, I wonder if this list is exhaustive to only what this SDK specifically collects?

src/utils/event-utils.ts
```diff
@@ -31,6 +31,25 @@ export const CAMPAIGN_PARAMS = [
     'rdt_cid', // reddit
 ]
 
+export const EVENT_TO_PERSON_PROPERTIES = [
```
pauldambra, line 34: I think this could do with a comment or a better name... 

I think it's saying "these are event properties that we will copy to person properties"
but I don't know for sure 🙈

src/customizations/setAllPersonProfilePropertiesAsPersonPropertiesForFlags.ts
```diff
@@ -0,0 +1,15 @@
+import { PostHog } from '../posthog-core'
+import { CAMPAIGN_PARAMS, EVENT_TO_PERSON_PROPERTIES, Info } from '../utils/event-utils'
+import { each, extend, includes } from '../utils'
+
+export const setAllPersonProfilePropertiesAsPersonPropertiesForFlags = (posthog: PostHog): void => {
```
pauldambra, line 5: could also do with a comment IMHO, why would I want to do this?

pauldambra (approved):
👍 

couple of comments but nothing blocking

src/utils/event-utils.ts
robbie-c: I don't disagree but it's copied directly from plugin server with that name, I'll add a comment though!

src/utils/event-utils.ts
robbie-c: Those other ones aren't converted into person properties though, or are they? This list comes straight from plugin-server.

robbie-c:
I'm going to merge this now and add a couple of comments in a follow up, just as I'm in a few meetings this morning and it'd be good to get back to the customer asap

src/utils/event-utils.ts
marandaneto: I didn't check so the plugin-server should be up to date and all good.

PR #1521, opened 2024-11-12
fix: Don't use session storage in memory mode

robbie-c:
## Changes

We currently use sessionStorage even when the persistence is set to memory. This is a bug, which this PR fixes and adds some tests for.

## Checklist


daibhin (approved):
Don't have full context on the why behind this change change but looks to do what it says in the test

PR #1523, opened 2024-11-13
feat: allow triggering sessions when events occur

richard-better:
folk frequently want to control their cost and volume of recordings

this allows them to set a config to start replay "paused" and only start sending data to the server when specified events are seen

Client side of https://github.com/PostHog/posthog/pull/26168


src/extensions/replay/sessionrecording.ts
```diff
@@ -266,6 +268,9 @@ export class SessionRecording {
 
     private _urlBlocked: boolean = false
 
+    private _eventTriggers: string[] = []
+    private _removeEventTriggerCaptureHook: (() => void) | undefined = undefined
```
pauldambra, line 272: are we calling this in the stop callback?

src/extensions/replay/sessionrecording.ts
```diff
@@ -644,6 +651,10 @@ export class SessionRecording {
             this._urlBlocklist = response.sessionRecording.urlBlocklist
         }
 
+        if (response.sessionRecording?.eventTriggers) {
+            this._eventTriggers = response.sessionRecording.eventTriggers
+        }
```
pauldambra, line 656: what was interesting in testing this was that i had an existing session, received the trigger in decide so didn't record

then when it was triggered i continued the same recording, _after_ that everything makes sense since a new session waits for the trigger

but might lead to unexpected behaviour during roll-out

so we could say 

* if you get a decide response that has different triggers to those you have stored
* immediately rotate the session id

but that feels like it could be unexpected too... maybe just us being aware of that roll out interaction is enough

no strong opinion here unless it triggers one in you :)

src/extensions/replay/sessionrecording.ts
```diff
@@ -1270,6 +1282,24 @@ export class SessionRecording {
         logger.info(LOGGER_PREFIX + ' recording resumed')
     }
 
+    private _addEventTriggerListener() {
+        if (this._eventTriggers.length === 0 || !isNullish(this._removeEventTriggerCaptureHook)) {
+            return
+        }
+
+        this._removeEventTriggerCaptureHook = this.instance._addCaptureHook((eventName) => {
```
pauldambra, line 1290: we could effectively inline here and instead of calling the `_addCaptureHook` method

call the more obviously public `on` method that it delegates to

but not blocking

pauldambra (approved):
some thoughts but otherwise...

👍

PR #1525, opened 2024-11-14
fix: improve ES6 bundling and support

pauldambra:
we can be much more explicit about what we support
and test that the bundle is ES6 only
which will extend browser support by >10% of global audience


robbie-c (approved):

pauldambra:
oh, very interesting that makes tests fail 🤯

package.json
```diff
@@ -121,9 +123,32 @@
...
+            "defaults",
+            "supports es6-module",
+            "chrome > 62",
+            "firefox > 59",
+            "ios_saf >= 10.3",
+            "opera > 50",
+            "not chrome 61",
+            "not opera 48",
+            "not chrome 62",
+            "not opera 49"
```
pauldambra, line 137: this is slightly ugly
and there may be a better way of doing it
but we're balancing saying "es6 support" and "navigator.webdriver" support

rollup.config.js
```diff
@@ -17,14 +18,16 @@ const plugins = (es5) => [
     babel({
         extensions: ['.js', '.jsx', '.ts', '.tsx'],
         babelHelpers: 'bundled',
-        plugins: ['@babel/plugin-transform-nullish-coalescing-operator'],
+        plugins: [
+            '@babel/plugin-transform-nullish-coalescing-operator',
+            // Explicitly included so we transform 1 ** 2 to Math.pow(1, 2) for ES6 compatability
+            '@babel/plugin-transform-exponentiation-operator',
```
pauldambra, line 24: i think the browsers to support navigator.webdriver was tricking babel into thinking some ES2017 features were ok
including this forces babel to transform this

rollup.config.js
```diff
@@ -17,14 +18,16 @@ const plugins = (es5) => [
...
+        ],
         presets: [
             [
                 '@babel/preset-env',
                 {
-                    targets: es5
-                        ? '>0.5%, last 2 versions, Firefox ESR, not dead, IE 11'
-                        : '>0.5%, last 2 versions, Firefox ESR, not dead',
+                    targets: es5 ? browserslist.es5 : browserslist.production,
```
pauldambra, line 30:
```suggestion
                    targets: es5 ? "defaults, IE 11" : browserslist.production,
```

package.json
```diff
@@ -120,9 +122,20 @@
...
+        "production": [
+            "defaults",
+            "chrome > 62",
+            "firefox > 59",
+            "ios_saf >= 10.3",
+            "opera > 50",
+            "safari > 12"
+        ],
+        "es5": [
+            "defaults",
```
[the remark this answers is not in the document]
pauldambra: > I was curious as to what defaults

`"defaults"` is just short-hand for the values we used to explicitly set.

> I think we could copy the whole list above, given that these are ORd together, so there's no harm in adding versions.

i've been fiddling with this to get CI to pass 🙈 

i don't understand why IE 11 is freezing in testcafe now so have been changing values around 🫠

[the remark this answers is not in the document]
pauldambra: (i agree the explicit list is better than "defaults" just hitting it with a hammer to try and understand what's happening :))

pauldambra:
this failing IE 11 test is super annoying -> https://github.com/PostHog/posthog-js/pull/1542

PR #1527, opened 2024-11-15
fix: XHR req method capture

pauldambra:
We were using `req.method` _before_ the request was used and so XHR was always captured as "GET"

### DIFF
--- a/cypress/e2e/session-recording.cy.ts
+++ b/cypress/e2e/session-recording.cy.ts
@@ -280,6 +280,62 @@ describe('Session recording', () => {
                     )
                 })
             })
+
+            it('it captures XHR method correctly', () => {
+                cy.get('[data-cy-xhr-call-button]').click()
+                cy.wait('@example.com')
+                cy.wait('@session-recording')
+                cy.phCaptures({ full: true }).then((captures) => {
+                    const snapshots = captures.filter((c) => c.event === '$snapshot')
+
+                    const capturedRequests: Record<string, any>[] = []
+                    for (const snapshot of snapshots) {
+                        for (const snapshotData of snapshot.properties['$snapshot_data']) {
+                            if (snapshotData.type === 6) {
+                                for (const req of snapshotData.data.payload.requests) {
+                                    capturedRequests.push(req)
+                                }
+                            }
+                        }
+                    }
+
+                    const expectedCaptureds: [RegExp, string][] = [
+                        [/http:\/\/localhost:\d+\/playground\/cypress\//, 'navigation'],
+                        [/http:\/\/localhost:\d+\/static\/array.js/, 'script'],
+                        [
+                            /http:\/\/localhost:\d+\/decide\/\?v=3&ip=1&_=\d+&ver=1\.\d\d\d\.\d+&compression=base64/,
+                            'xmlhttprequest',
+                        ],
+                        [/http:\/\/localhost:\d+\/static\/recorder.js\?v=1\.\d\d\d\.\d+/, 'script'],
+                        [/https:\/\/example.com/, 'xmlhttprequest'],
+                    ]
+
+                    // yay, includes expected type 6 network data
+                    expect(capturedRequests.length).to.equal(expectedCaptureds.length)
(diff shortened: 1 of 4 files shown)


cypress/e2e/session-recording.cy.ts
```diff
@@ -280,6 +280,62 @@ describe('Session recording', () => {
                     )
                 })
             })
+
+            it('it captures XHR method correctly', () => {
```
pauldambra, line 284: these tests have a lot of duplication now, but I want to get this fix out

src/entrypoints/recorder.ts
```diff
@@ -386,7 +386,7 @@ function prepareRequest({
             timeOrigin,
             timestamp,
             method: method,
-            initiatorType: entry ? (entry.initiatorType as InitiatorType) : initiatorType,
```
pauldambra, line 389: flips the choice here... if we're providing an initiator we should use it in preference to the detected one

they _should_ always match

src/extensions/replay/sessionrecording.ts
```diff
@@ -430,10 +430,8 @@ export class SessionRecording {
     }
 
     /**
-     * Any one trigger can activate the session
-     * So, if they are all the same - return that value
-     * If either is disabled return the other's valu
-     * @private
+     * trigger is active if _either_ URL _or_ event trigger is active
+     * and is pending if _either_ URL _or_ event trigger is pending
```
pauldambra, line 434: flyby realised this comment was nonsense 🤣

src/entrypoints/recorder.ts
```diff
@@ -309,7 +309,7 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
                         .then((entry) => {
                             const requests = prepareRequest({
                                 entry,
-                                method: req.method,
+                                method: method,
```
pauldambra, line 312: this is the fix... the method is provided to the XHR function when it is called so we should use it

src/extensions/replay/sessionrecording.ts
robbie-c: Pedantic brain: what if one is active and one is pending?

cypress/e2e/session-recording.cy.ts
robbie-c: This test is pretty convincing, how would it have failed before?

robbie-c: The method = POST assertion?

robbie-c (approved):
We're deep within code I haven't looked at before so with that caveat:
* The bug itself makes sense
* The fix for it makes sense
* The test looks good

👍 👍 👍

cypress/e2e/session-recording.cy.ts
pauldambra: yep... it was always GET before (or at least overwhelmingly GET I didn't check which)

src/extensions/replay/sessionrecording.ts
pauldambra:
```suggestion
```

also... i'm describing the code... it's a redundant comment

src/extensions/replay/sessionrecording.ts
```diff
@@ -430,10 +430,6 @@ export class SessionRecording {
     }
 
     /**
-     * Any one trigger can activate the session
-     * So, if they are all the same - return that value
-     * If either is disabled return the other's valu
-     * @private
      */
```
pauldambra, line 433:
```suggestion
```

src/extensions/replay/sessionrecording.ts
```diff
@@ -430,11 +430,6 @@ export class SessionRecording {
     }
 
     /**
```
pauldambra, line 432:
```suggestion
```

cypress/e2e/session-recording.cy.ts
pauldambra: and for clarity I validated the test failed before the change 😊

PR #1528, opened 2024-11-15
chore: Added CSP headers to next app for testing what we document

benjackwhite:
## Changes

Related documentation updates https://github.com/PostHog/posthog.com/pull/9906

Confirmed that this covers everything needed for the toolbar and the general script loading. 

## Checklist

### DIFF
--- a/playground/nextjs/pages/_app.tsx
+++ b/playground/nextjs/pages/_app.tsx
@@ -46,8 +46,8 @@ export default function App({ Component, pageProps }: AppProps) {
                     default-src 'self';
                     connect-src 'self' http://localhost:8000 https://*.posthog.com;
                     script-src 'self' 'unsafe-eval' 'unsafe-inline' http://localhost:8000 https://*.posthog.com;
-                    style-src 'self' 'unsafe-inline' http://localhost:8000;
-                    img-src 'self' http://localhost:8000;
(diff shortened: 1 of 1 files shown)


playground/nextjs/pages/_app.tsx
```diff
@@ -39,6 +39,17 @@ export default function App({ Component, pageProps }: AppProps) {
             <Head>
                 <title>PostHog</title>
                 <meta name="viewport" content="width=device-width, initial-scale=1" />
+                {/* CSP - useful for testing our documented recommendations. NOTE: Unsafe is only needed for nextjs pre-loading */}
+                <meta
+                    http-equiv="Content-Security-Policy"
+                    content={`
+                    default-src 'self';
+                    connect-src 'self' http://localhost:8000 https://*.posthog.com;
```
robbie-c, line 47: Minor nit but there's a script in package.json (called dev-crossdomain) which uses https (used for testing if some features work with https). It'd be great if this could use https instead if `NEXT_PUBLIC_CROSSDOMAIN==1`.

robbie-c (approved):

playground/nextjs/pages/_app.tsx
benjackwhite: Updated. If its not correct, feel free to fix forward 👍

PR #1530, opened 2024-11-18
fix: angular change detection clashing

pauldambra:
we get periodic reports of performance issues with angular

angular change detection patches browser methods

we do change detection too

so we fight each other

this grabs methods from an iframe so that we aren't using angular's patched methods


pauldambra:
this is way easier to test thanks to a customer providing a repo here https://github.com/typedb/angular-posthog-freeze

pauldambra:
can make a smaller change

PR #1539, opened 2024-11-19
chore: specify an explicit browserslist version

pauldambra:
pulling small changes out of https://github.com/PostHog/posthog-js/pull/1525 because IE 11 tests are freezing in that PR and it's really hard to figure out why

in that PR I added an explicit browserslist version, let's do that separately


robbie-c (approved):
Looks like specifying an explicit browserslist version to me!

PR #1540, opened 2024-11-19
chore: timeout test cafe jobs

pauldambra:
https://github.com/PostHog/posthog-js/pull/1525 has a problem where the IE 11 tests are freezing

it will run for many hours

that's silly


pauldambra:
just going to merge this pre-review since it's so innocuous 🙈

PR #1541, opened 2024-11-19
chore: package json browserslist as source of truth

pauldambra:
pulling small changes out of #1525 because IE 11 tests are freezing in that PR and it's really hard to figure out why

this grabs the browserslist setting from package.json


pauldambra:
Ah, ok, this one can't come out because it's the change that demands the change in browserslist

PR #1542, opened 2024-11-19
fix: improve ES6 bundling

pauldambra:
#1525 - IE 11 tests are freezing in that PR and it's really hard to figure out why

have pulled many small changes out of that PR but the IE 11 tests still freeze

----

* improves detection/bundling of ES6 support for posthog-js by being more specific with the browser targets that are passed to the bundler
* adds a check that the array.full.js is ES6 compliant in CI

Does not update the browserslist value in package.json, since that is the change that appears to make CI freeze
I'd rather get the benefit of the ES6 compliant bundle in prod before fixing that

### DIFF
--- a/.github/workflows/es-check.yml
+++ b/.github/workflows/es-check.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   ssr:
-    name: Cypress
+    name: Build and check ES5/ES6 support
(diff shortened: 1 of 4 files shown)


marandaneto (approved):
low context n. N :D

PR #1543, opened 2024-11-19
fix: patch angular wrap detection in rrweb

pauldambra:
follow-up to https://github.com/PostHog/posthog-js/pull/1531 which fixed this for dead clicks

patch rrweb to correctly detect angular's change detection

🎉 tested locally by installing patched version in an angular application and seeing it no longer freezes 🎉 

also opened upstream PR here https://github.com/rrweb-io/rrweb/pull/1597

### DIFF
--- a/patches/@rrweb__record@2.0.0-alpha.17.patch
+++ b/patches/@rrweb__record@2.0.0-alpha.17.patch
@@ -1,8 +1,32 @@
 diff --git a/dist/record.js b/dist/record.js
-index 46ec389fefb698243008b39db65470dbdf0a3857..39d9ac821fd84f5a1494ac1e00c8d4e0e58c432a 100644
+index 46ec389fefb698243008b39db65470dbdf0a3857..891e1cf6439630d19e9b745ff428db438943a0b2 100644
 --- a/dist/record.js
 +++ b/dist/record.js
-@@ -246,6 +246,9 @@ function isCSSImportRule(rule2) {
+@@ -26,6 +26,14 @@ const testableMethods$1 = {
+   Element: [],
+   MutationObserver: ["constructor"]
+ };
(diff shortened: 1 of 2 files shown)


patches/@rrweb__record@2.0.0-alpha.17.patch
```diff
@@ -1,8 +1,47 @@
...
+ };
++const isFunction = (x) => typeof x === 'function';
++const isNativeFunction = () => isFunction(x) && x.toString().includes("[native code]");
++const isAngularZonePatchedFunction = (x) => {
++  if (!isFunction(x)) {
++      return false;
++  }
++  const prototypeKeys = Object.getOwnPropertyNames(x.prototype || {});
++  return prototypeKeys.some((key) => key.indexOf('__zone'));
++}
```
pauldambra, line 17: copies these three functions from posthog-js that are already being used in https://github.com/PostHog/posthog-js/pull/1531

patches/@rrweb__record@2.0.0-alpha.17.patch
```diff
@@ -1,8 +1,47 @@
...
+-      }
+-    )
++    accessorNames.every((accessor) => {
++      const candidate = Object.getOwnPropertyDescriptor(defaultPrototype, accessor)?.get;
++      let isUntainted = isNativeFunction(candidate);
++      if (key === 'MutationObserver') {
++        isUntainted = isUntainted && !isAngularZonePatchedFunction(candidate);
++      }
++      return isUntainted;
++    }),
```
pauldambra, line 40: when testing mutation observer it has to be native code _and_ not have angular zone prototypes

marandaneto (approved):

pauldambra:
frustratingly my test angular application is freezing before this :/

PR #1548, opened 2024-11-22
fix(surveys): Process feature_flag_keys on Survey object

Phanatic:
## Changes

As part of adaptive survey collection, we introduced a new feature flag that the surveys component should use to filter out surveys to be shown to people. 
The common pattern has become : 
1. Add a flag to the API
2. Add the same flag as a check to `getActiveMatchingSurveys`

Instead, this PR now uses the `feature_flag_keys` array to find the names of the feature flags to check in the survey component. 
With this change, one can just add a new flag to the `feature_flag_keys` in the Surveys API in posthog.com and the survey component will automatically pick it up. 

## Checklist

### DIFF
--- a/src/__tests__/surveys.test.ts
+++ b/src/__tests__/surveys.test.ts
@@ -122,6 +122,44 @@ describe('surveys', () => {
         } as unknown as Survey,
     ]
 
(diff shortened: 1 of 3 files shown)


havenbarnes (approved):
Makes sense! 🚀

PR #1550, opened 2024-11-25
feat: Add better npm import, and script entrypoint for customizations

robbie-c:
## Changes

Add a script entrypoint, `cutomizations.full.js` for adding all of the cutomizations for posthog-js.

Right now this is the sampling functions and the setAllPersonPropertiesAsPropertiesForFlags function.

I've also added a central import for npm users who want to use customizations, so now they can do `import { sampleByDistinctId } from 'posthog-js/lib/customizations'`

## Checklist

### DIFF
--- a/src/customizations/index.ts
+++ b/src/customizations/index.ts
@@ -1,0 +1,2 @@
+export * from './setAllPersonProfilePropertiesAsPersonPropertiesForFlags'
+export * from './before-send'
--- a/playground/nextjs/pages/index.tsx
+++ b/playground/nextjs/pages/index.tsx
@@ -2,7 +2,7 @@
 import { useActiveFeatureFlags, usePostHog } from 'posthog-js/react'
 import { useEffect, useState } from 'react'
 import { cookieConsentGiven, PERSON_PROCESSING_MODE } from '@/src/posthog'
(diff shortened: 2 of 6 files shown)


pauldambra:
it's a big size bump for a convenience change...

i guess they're either not pure or rollup isn't detecting them as pure... 

need some kind of `/*@__NO_SIDE_EFFECTS__*/` marker somewhere so that these are only available in NPM not in the bundles?

src/customizations/index.ts
pauldambra, line 1: sorry should have commented here 

----

it's a big size bump for a convenience change...

i guess they're either not pure or rollup isn't detecting them as pure...

need some kind of /*@__NO_SIDE_EFFECTS__*/ marker somewhere so that these are only available in NPM not in the bundles?

robbie-c: Yeah, the size bump tells me that this hasn't worked as expected.

I'll look into it

Apologies for the premature tagging for review

robbie-c: Ok I removed the tsconfig change, and this seems to have prevented the increase in bundle size. I'll figure this part out separately.

pauldambra (approved):
lovely quality of life improvement

wanna update the `before_send` docs too? #cheeky

🚀

PR #1554, opened 2024-11-26
feat: Change lib import path from /lib/src/ to /lib

robbie-c:
## Changes

Change tsconfig.json to have a root dir, this lets people import e.g. customizations from `posthog-js/lib` rather than `posthog-js/lib/src` which IMO is cleaner

## Checklist


robbie-c:
This seems not a priority, so 🪓

PR #1555, opened 2024-11-26
feat(cdp): site apps handle consent

MarconLP:
## Changes

- Adds a `disable_site_apps_destinations` property that will disable all `site_destination`'s
- Checks if the property is enabled every time `set_config` is being called.
- Updated the nextjs example to disable/enable `site_destination`'s based on the cookie banner ([link](https://github.com/PostHog/posthog-js/blob/site-apps-consent/playground/nextjs/src/posthog.ts))

TODOs:
- [ ] add tests

## Checklist


MarconLP:
close in favor of #1563

PR #1557, opened 2024-11-26
fix: catch mutation throttling errors

pauldambra:
Am seeing reports of safari recordings not including incremental snapshots. 

This is a code route that incremental snapshots go through that other snapshots don't

I think this is totally a red herring but it is at least a safe change to make


marandaneto (approved):

pauldambra:
I'm certain it's not this, so for clarity 🪓

PR #1558, opened 2024-11-26
feat: different default and max idle period

pauldambra:
allow configuring an idle timeout larger than the default

see https://posthog.slack.com/archives/C07RUA1TBT3/p1732619524510359

the motivation for the user here is that their product might be used by someone before they leave for the day, and then once they return, and those two interactions are logically one session. the second might follow-on from the first.

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -427,7 +427,7 @@ export class PostHog {
         this._retryQueue = new RetryQueue(this)
         this.__request_queue = []
 
-        this.sessionManager = new SessionIdManager(this.config, this.persistence)
+        this.sessionManager = new SessionIdManager(this)
         this.sessionPropsManager = new SessionPropsManager(this.sessionManager, this.persistence)
 
         new TracingHeaders(this).startIfEnabledOrStop()
--- a/src/sessionid.ts
+++ b/src/sessionid.ts
@@ -10,9 +10,10 @@ import { logger } from './utils/logger'
 
 import { clampToRange } from './utils/number-utils'
 
-const MAX_SESSION_IDLE_TIMEOUT = 30 * 60 // 30 minutes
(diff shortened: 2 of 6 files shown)


src/posthog-core.ts
```diff
@@ -427,7 +427,7 @@ export class PostHog {
         this._retryQueue = new RetryQueue(this)
         this.__request_queue = []
 
-        this.sessionManager = new SessionIdManager(this.config, this.persistence)
+        this.sessionManager = new SessionIdManager(this)
```
pauldambra, line 430: chose to pass in the posthog instance rather than pass in `this.register`

src/sessionid.ts
```diff
@@ -9,10 +9,12 @@ import { isArray, isNumber, isUndefined } from './utils/type-utils'
...
 import { clampToRange } from './utils/number-utils'
+import { PostHog } from './posthog-core'
 
-const MAX_SESSION_IDLE_TIMEOUT = 30 * 60 // 30 minutes
-const MIN_SESSION_IDLE_TIMEOUT = 60 // 1 minute
-const SESSION_LENGTH_LIMIT = 24 * 3600 * 1000 // 24 hours
+export const DEFAULT_SESSION_IDLE_TIMEOUT_SECONDS = 30 * 60 // 30 minutes
+export const MAX_SESSION_IDLE_TIMEOUT_SECONDS = 10 * 60 * 60 // 10 hours
+const MIN_SESSION_IDLE_TIMEOUT_SECONDS = 60 // 1 minute
+const SESSION_LENGTH_LIMIT_MILLISECONDS = 24 * 3600 * 1000 // 24 hours
```
pauldambra, line 17: renamed these since they have different units
added DEFAULT and MAX as separate values

pauldambra:
tagging @robbie-c too in case this would be bad for sessions table

robbie-c (approved):

Issue #1561, opened 2024-11-27
Significant amount of blank screen recordings on Safari since 1.186.3


fluidsonic:
Recently, we noticed quite a few blank screen recordings. The background is visible and the correct color and the user clearly uses the website, but there's nothing else to see in the replays.

I noticed it in Safari (desktop & mobile) 17 & 18 with PostHog JS library 1.186.3, 1.186.4, 1.187.2, 1.188.0, and 1.189.0. However, it does work _sometimes_ with these configurations. There are no errors reported in the console, according to the screen replay.

I noticed that there don't seem to be any problematic recordings before mid-November. I also checked for screen recordings that use PostHog library 1.186.0, 1.186.1, and 1.186.2. None of these were affected (but our number of replays in Safari for 1.186.2 is low).

This is probably a bug introduced somewhere around 1.186.3.

Let me know if there's anything else I can provide to help understand the issue.

## Update

We noticed a lot of reports in our error reporting:

```
Message: The string did not match the expected pattern.
at matches line NaN, column null ([native code]:null)
at ? line 1, column 11167 (…/static/recorder.js?v=1.189.0:1)
at ae line 1, column 14579 (…/static/recorder.js?v=1.189.0:1)
at o line 1, column 31052 (…/static/recorder.js?v=1.189.0:1)
at ? line 1, column 32265 (…/static/recorder.js?v=1.189.0:1)
at ? line 1, column 30703 (…/static/recorder.js?v=1.189.0:1)
at ? line NaN, column null ([native code]:null)
```

## Update 2

The exception above was triggered only by a single user within the past month and is probably not related to this issue. May just be a coincidence.

kieri097:
We have been experiencing similar issues starting mid/lateNovember ; background shows and user activity is recorded correctly, but the interface itself does not show during playback (@pauldambra is already looking into it).

I just went through our recordings and the issues start appearing after we updated from 1.181.0 to 1.187.2 ; so if for you the last working version was 1.186.1, i guess that would fit with your error version triangulation .

We also made some other changes during the same update, still trying to figure out if we changed anything on our end that might be behind it...

No errors on our side though, where are you seeing those?

pauldambra:
Hey @kieri097 I'm certain yours is a different issue - I believe you're reporting you have a black background and see mouse moves, that was indeed fixed in 1.187.2.

[...]

PR #1562, opened 2024-11-27
fix: catch errors detecting dialog state when recording

pauldambra:
see https://github.com/PostHog/posthog-js/issues/1561 which points at an error in using the `:modal` pseudo selector

it does appear that safari started supporting the dialog element in 15.4 and the `:modal ` selector in 15.6


marandaneto (approved):

rafaeelaudibert (approved):
🚢

PR #1563, opened 2024-11-27
feat(cdp): allow disabling site destinations

MarconLP:
## Changes

- Only load `site_destination`'s if `isOptedOut` is false
- site destinations will be loaded if `.opt_in_capturing()` is being called at a later point.
- Updated the nextjs example to call `opt_in_capturing` & `opt_out_capturing` based on the cookie banner ([link](https://github.com/PostHog/posthog-js/blob/site-apps-consent/playground/nextjs/src/posthog.ts))

2024-11-27 at 15 18 40

TODOs:
- [x] add tests

## Checklist

### DIFF
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -60,6 +60,7 @@ if (typeof window !== 'undefined') {
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
         ...configForConsent(),
     })
+
     // Help with debugging(window as any).posthog = posthog
 }
 
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -60,7 +60,7 @@ if (typeof window !== 'undefined') {
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
         ...configForConsent(),
     })
-
+    ;(window as any).posthog = posthog
     // Help with debugging(window as any).posthog = posthog
 }
 
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -28,7 +28,6 @@ export const configForConsent = (): Partial<PostHogConfig> => {
     return {
         persistence: consentGiven ? 'localStorage+cookie' : 'memory',
         disable_surveys: !consentGiven,
-        disable_site_apps_destinations: !consentGiven,
         autocapture: consentGiven,
         disable_session_recording: !consentGiven,
     }
@@ -37,8 +36,10 @@ export const configForConsent = (): Partial<PostHogConfig> => {
 export const updatePostHogConsent = (consentGiven: boolean) => {
     if (consentGiven) {
         localStorage.setItem('cookie_consent', 'true')
+        posthog.opt_in_capturing()
     } else {
         localStorage.removeItem('cookie_consent')
+        posthog.opt_out_capturing()
     }
 
     posthog.set_config(configForConsent())
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -28,6 +28,7 @@ export const configForConsent = (): Partial<PostHogConfig> => {
     return {
         persistence: consentGiven ? 'localStorage+cookie' : 'memory',
         disable_surveys: !consentGiven,
+        disable_site_apps_destinations: !consentGiven,
         autocapture: consentGiven,
         disable_session_recording: !consentGiven,
     }
@@ -49,6 +50,7 @@ if (typeof window !== 'undefined') {
         session_recording: {
             recordCrossOriginIframes: true,
         },
+        opt_in_site_apps: true,
         debug: true,
         disable_web_experiments: false,
         scroll_root_selector: ['#scroll_element', 'html'],
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -60,7 +60,6 @@ if (typeof window !== 'undefined') {
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
         ...configForConsent(),
     })
-    ;(window as any).posthog = posthog
     // Help with debugging(window as any).posthog = posthog
 }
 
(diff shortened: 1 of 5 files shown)


playground/nextjs/src/posthog.ts
```diff
@@ -57,7 +60,7 @@ if (typeof window !== 'undefined') {
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
         ...configForConsent(),
     })
-
+    ;(window as any).posthog = posthog
```
mariusandra, line 63: Is this still needed for anything? We should probably remove if not, as site apps/destinations should work without it.

src/__tests__/site-apps.ts
```diff
@@ -73,42 +79,6 @@ describe('SiteApps', () => {
...
-            posthog.config = {
-                ...defaultConfig,
-                opt_in_site_apps: true,
-                advanced_disable_decide: false,
-            } as PostHogConfig
-
-            siteAppsInstance = new SiteApps(posthog)
-
-            expect(siteAppsInstance.enabled).toBe(true)
-        })
```
mariusandra, line 86: Does `advanced_disable_decide` still do anything? I see that it's used in the code below, so probably worth testing for it somehow.

src/__tests__/site-apps.ts
```diff
@@ -330,7 +296,42 @@ describe('SiteApps', () => {
...
+            } as DecideResponse
+
+            siteAppsInstance.afterDecideResponse(response)
+
+            posthog.consent.isOptedOut = () => false
+
+            siteAppsInstance.loadIfEnabled()
+
+            expect(typeof assignableWindow['__$$ph_site_app_5_callback']).toBe('function')
+            expect(typeof assignableWindow['__$$ph_site_app_6_callback']).toBe('function')
```
mariusandra, line 334: Is this `loadIfEnabled` correct? You're explicitly loading the site apps/destinations after consent. And they do load. However, you're calling the "load" function directly in the test. I don't expect users to call this function. 

We should test that opting in the way users opt in starts the site apps.

src/site-apps.ts
```diff
@@ -74,18 +72,33 @@ export class SiteApps {
...
+                    // if consent isn't given, skip site destinations
+                    if (this.instance.consent.isOptedOut() && type === 'site_destination') {
+                        checkIfAllLoaded()
+                        continue
+                    }
+                    // if the site app is already loaded, skip it
+                    if (!isUndefined(assignableWindow[`__$$ph_site_app_${id}_posthog`])) {
+                        checkIfAllLoaded()
+                        continue
+                    }
```
mariusandra, line 101: Nit: I'd just `continue` here, and do a global `checkIfAllLoaded()` below the loop.

src/site-apps.ts
```diff
@@ -101,17 +114,19 @@ export class SiteApps {
...
             }
         } else {
             this.loaded = true
-            this.enabled = false
         }
     }
 
-    // TODO: opting out of stuff should disable this
+    afterDecideResponse(response: DecideResponse): void {
+        this._decideServerResponse = response['siteApps']
```
mariusandra, line 129: The variable `_decideServerResponse` makes me think this contains the whole response, not just the `siteApps` key

mariusandra (changes requested):

mariusandra:
So... I finally had a look, and I think we're most of the way there.

One more thing: there's a chance if you don't opt in, that the event collector will collect forever. Is there ever a "definitely opted out" event to hook into, which would stop it?

src/__tests__/site-apps.ts
MarconLP: > Does advanced_disable_decide still do anything?

Yes! Added an additional test.

playground/nextjs/src/posthog.ts
MarconLP: No, that was just for debugging. Removed!

src/site-apps.ts
MarconLP: renamed to "_decideServerSiteAppsResponse"

src/__tests__/site-apps.ts
MarconLP: updated the test to call the `opt_out_capturing` and `opt_in_capturing` functions instead.

MarconLP:
> One more thing: there's a chance if you don't opt in, that the event collector will collect forever. Is there ever a "definitely opted out" event to hook into, which would stop it?

That is not a thing. It's either user is opted in/out by default and consent has been toggled.		

Currently, we'll clear the `missedInvocations` queue after we've loaded all of the `site_app`'s and set `loaded` to `true` (this will include cases where consent wasn't given and `site_destination`'s haven't been loaded yet), which will disable the eventCollector.

We could reuse the `missedInvocations` queue to only clear after all `site_destination`'s have been loaded and `consentIsGiven`, since it's already configured to only hold 1000 events at a time?
```
if (this.missedInvocations.length > 1000) {
    this.missedInvocations = this.missedInvocations.slice(10)
}
```

mariusandra:
Wait... so what happens if:
- User clicks on an ad and lands on a site
- Waits 30seconds before clicking "agree to cookies and stuff" (all JS loads within the first 5sec)
- what happens now?

They way I understand the last comment, is that they won't count as a converted user (we drop all events captured, as when site apps finished loading, we didn't have consent yet). Is that how it works, or did I misunderstand something again? If that is how it works though, it's definitely far from ideal and needs fixing.

benjackwhite:
Sorry to be a pain but I will really want to get [this](https://github.com/PostHog/posthog-js/pull/1580) in first

MarconLP:
I updated the code to save initial campaign/referrer props parameters to `this.storage`.

1. `persistence: 'memory'` & `isOptedOut: true`
2. User visits [posthog.com?gclid=ad_click_1](http://posthog.com/?gclid=ad_click_1)
	- We won't capture a $pageview event, but we'll store `gclid: 'ad_click_1'` in memory
3. User visits the pricing page
	- URL changes to [posthog.com/pricing](https://posthog.com/pricing)
4. User accepts the cookie banner
 	- We'll send an initial `$pageview` event and since the `gclid` is stored in `this.storage` we'll append that as an event property. Will also be included as an `$initial_` person property.
6. User visits the signup page - we capture a $pageview event
 	- This event will be captured normally

If the user reloads the page between step 2 - 4 we'll lose the gclid property.

MarconLP:
closing in favor of https://github.com/PostHog/posthog-js/pull/1619

PR #1565, opened 2024-11-27
fix: patch to angular detection in rrweb

pauldambra:
It's really hard to reproduce the bug, and in the one site where I can reproduce it it's really hard to debug what's happening.

Explanation in-line in the review, but this is an attempt to make this detection safer in Safari


pauldambra:
ugh, intellij files :/

PR #1566, opened 2024-11-27
fix: patch to angular detection in rrweb

pauldambra:
It's really hard to reproduce the bug, and in the one site where I can reproduce it it's really hard to debug what's happening.

Explanation in-line in the review, but this is an attempt to make this detection safer in Safari

### DIFF
--- a/patches/@rrweb__record@2.0.0-alpha.17.patch
+++ b/patches/@rrweb__record@2.0.0-alpha.17.patch
@@ -1,23 +1,28 @@
 diff --git a/dist/record.js b/dist/record.js
-index 46ec389fefb698243008b39db65470dbdf0a3857..bfebdc6a3887a3ddae46b21d21896092d2963fe9 100644
+index 46ec389fefb698243008b39db65470dbdf0a3857..70db907755d68b08232e25e1b255a974f56f3c65 100644
 --- a/dist/record.js
 +++ b/dist/record.js
-@@ -26,6 +26,14 @@ const testableMethods$1 = {
+@@ -26,6 +26,19 @@ const testableMethods$1 = {
    Element: [],
    MutationObserver: ["constructor"]
  };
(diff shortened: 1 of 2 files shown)


patches/@rrweb__record@2.0.0-alpha.17.patch
```diff
@@ -1,23 +1,28 @@
...
-+      return false;
++  try {
++    if (!isFunction(x)) {
++        return false;
++    }
++    const prototypeKeys = Object.getOwnPropertyNames(x.prototype || {});
++    return prototypeKeys.some((key) => key.indexOf('__zone'));
++  } catch {
++    // we've seen some intermittent problems in Safari since introducing this check
++    return false
```
pauldambra, line 19: this is the new thing, so let's wrap it and return false
very speculative

patches/@rrweb__record@2.0.0-alpha.17.patch
```diff
@@ -26,7 +31,20 @@ index 46ec389fefb698243008b39db65470dbdf0a3857..bfebdc6a3887a3ddae46b21d21896092
...
+@@ -65,10 +78,10 @@ function getUntaintedPrototype$1(key) {
+     if (!win) return defaultObj.prototype;
+     const untaintedObject = win[key].prototype;
+     document.body.removeChild(iframeEl);
+-    if (!untaintedObject) return defaultPrototype;
++    if (!untaintedObject) return defaultObj.prototype;
+     return untaintedBasePrototype$1[key] = untaintedObject;
+   } catch {
+-    return defaultPrototype;
++    return defaultObj.prototype;
```
pauldambra, line 43: we return slightly different things on different lines
we didn't change this but it's nicer to always return the same thing
NB in a site where I could reproduce the problem it was `if (!untaintedObject) return defaultPrototype;` that was returning

marandaneto (approved):

PR #1567, opened 2024-11-28
chore(flags): Add new debugging property `$used_bootstrap_value`, `$feature_flag_bootstrapped_response` and `$feature_flag_bootstrapped_payload` to `$feature_flag_called` event

havenbarnes:
## Changes

Same change for bootstrapping as https://github.com/PostHog/posthog-js-lite/pull/320

## Checklist

### DIFF
--- a/src/__tests__/featureflags.ts
+++ b/src/__tests__/featureflags.ts
@@ -32,6 +32,7 @@ describe('featureflags', () => {
             get_property: (key) => instance.persistence.props[key],
             capture: () => {},
             decideEndpointWasHit: false,
+            receivedFlagValues: false,
             _send_request: jest.fn().mockImplementation(({ callback }) =>
                 callback({
                     statusCode: 200,
@@ -67,7 +68,7 @@ describe('featureflags', () => {
     })
 
     it('should return flags from persistence even if decide endpoint was not hit', () => {
-        featureFlags.instance.decideEndpointWasHit = false
+        featureFlags.instance.receivedFlagValues = false
 
         expect(featureFlags.getFlags()).toEqual([
             'beta-feature',
@@ -80,7 +81,7 @@ describe('featureflags', () => {
 
     it('should warn if decide endpoint was not hit and no flags exist', () => {
         ;(window as any).POSTHOG_DEBUG = true
-        featureFlags.instance.decideEndpointWasHit = false
+        featureFlags.instance.receivedFlagValues = false
         instance.persistence.unregister('$enabled_feature_flags')
         instance.persistence.unregister('$active_feature_flags')
 
@@ -101,7 +102,7 @@ describe('featureflags', () => {
     })
 
     it('should return the right feature flag and call capture', () => {
-        featureFlags.instance.decideEndpointWasHit = false
+        featureFlags.instance.receivedFlagValues = false
 
         expect(featureFlags.getFlags()).toEqual([
             'beta-feature',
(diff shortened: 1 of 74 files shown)


src/posthog-featureflags.ts
```diff
@@ -244,7 +244,14 @@ export class PostHogFeatureFlags {
                 }
                 this.instance.persistence?.register({ [FLAG_CALL_REPORTED]: flagCallReported })
 
-                this.instance.capture('$feature_flag_called', { $feature_flag: key, $feature_flag_response: flagValue })
+                this.instance.capture('$feature_flag_called', {
+                    $feature_flag: key,
+                    $feature_flag_response: flagValue,
+                    $feature_flag_payload: this.getFeatureFlagPayload(key) || null,
+                    $feature_flag_bootstrapped_response: this.instance.config.bootstrap?.featureFlags?.[key] || null,
```
dmarticus, line 251: Not blocking because I'm asking you to do something a bit more involved, but I think you should probably make this change to the React Native SDK as well, given that we want to have feature parity between the two client-side SDKs as much as we can reasonably can.  [Here's the relevant code](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-core/src/index.ts#L1234-L1260), although I think you might need to change some stuff in `posthog-core` to get it to play nicely.

dmarticus (approved):
stamped with a comment about making sure this has parity for the other client-side SDKs.

src/posthog-featureflags.ts
havenbarnes: I think that should be here already! https://github.com/PostHog/posthog-js-lite/pull/320

dmarticus: 🤦 wow now I feel like a jerk for calling it out... nice work man!

havenbarnes: All good! just updated with $used_bootstrap_value!

dmarticus (approved):
:shipit:

src/__tests__/featureflags.ts
```diff
@@ -32,6 +32,7 @@ describe('featureflags', () => {
             get_property: (key) => instance.persistence.props[key],
             capture: () => {},
             decideEndpointWasHit: false,
+            receivedFlagValues: false,
```
havenbarnes, line 35: In this PR we are essentially find+replacing the existing flag `decideEndpointWasHit` for this new, more accurately named `receivedFlagValues` flag. The prior will now mean that the endpoint truly was hit with no errors, and the latter will mean that there are flag values available for the SDK to return. 
`decideEndpointWasHit` is now really only used for passing back the `$used_bootstrap_value` flag in $feature_flag_called events.

havenbarnes:
@dmarticus added `$used_bootstrap_value` to this, see comment above for explanation of the big diff

dmarticus (approved):
third ship is the charm

testcafe/check-testcafe-results.js
```diff
@@ -47,7 +47,7 @@ If they seem to be failing unexpectedly, check grafana for ingestion lag at http
     log(JSON.stringify(files, null, 2))
 
     // the deadline is the same for each assert, as the ingestion lag will be happening in parallel
-    const deadline = Date.now() + 1000 * 60 * 30 // 30 minutes
+    const deadline = Date.now() + 1000 * 60 * 120 // 30 minutes
```
dmarticus, line 50: pedantic, but you change the comment to match the time (it's 120 minutes now).

havenbarnes: I was hoping to not leave this around, but may do so depending on what others think

dmarticus: yeah fair 'nuff

PR #1569, opened 2024-11-28
fix: infinite loop when combining controls

pauldambra:
see https://posthoghelp.zendesk.com/agent/tickets/21246 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/23530)

When waiting for a linked flag to activate recordings, and also pausing on URL, the browser could get stuck in a loop trying to emit a custom event that it had paused when it had not yet started, 

so the note it emitted triggered the pause check which again tried to pause

which caused a custom event 

etc

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -392,16 +392,16 @@ export class SessionRecording {
             return 'disabled'
         }
 
+        if (this._urlBlocked) {
+            return 'paused'
+        }
+
         if (!isNullish(this._linkedFlag) && !this._linkedFlagSeen) {
             return 'buffering'
         }
 
         if (this.triggerStatus === 'trigger_pending') {
             return 'buffering'
-        }
-
-        if (this._urlBlocked) {
-            return 'paused'
         }
 
         if (isBoolean(this.isSampled)) {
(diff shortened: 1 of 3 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -392,6 +392,10 @@ export class SessionRecording {
             return 'disabled'
         }
 
+        if (this._urlBlocked) {
+            return 'paused'
+        }
```
pauldambra, line 397: the fix is to have this as the earliest check in status
rrweb continues to emit, because it doesn't know that it is paused
and we continue to check
until the URL changes and we resume

src/extensions/replay/sessionrecording.ts
```diff
@@ -501,13 +501,15 @@ export class SessionRecording {
             if (isNullish(this._removePageViewCaptureHook)) {
                 // :TRICKY: rrweb does not capture navigation within SPA-s, so hook into our $pageview events to get access to all events.
                 //   Dropping the initial event is fine (it's always captured by rrweb).
-                this._removePageViewCaptureHook = this.instance._addCaptureHook((eventName) => {
+                this._removePageViewCaptureHook = this.instance.on('eventCaptured', (event) => {
```
pauldambra, line 504: changed this to not rely on a semi private method

src/extensions/replay/sessionrecording.ts
```diff
@@ -501,13 +501,15 @@ export class SessionRecording {
...
                     // so we catch all errors.
                     try {
-                        if (eventName === '$pageview') {
-                            const href = window ? this._maskUrl(window.location.href) : ''
-                            if (!href) {
+                        if (event === '$pageview') {
+                            const href = event?.properties.$current_url
+                                ? this._maskUrl(event?.properties.$current_url)
+                                : ''
+                            if (!href || this.status !== 'active') {
```
pauldambra, line 512: pageview already tells you what URL it is, so we don't need to ask the window

pauldambra: fly by change that had been annoying me

marandaneto (approved):

PR #1570, opened 2024-11-28
fix: zone detection

pauldambra:
mutation observer setup is _sometimes_ failing in Safari

I think that the underlying problem is we can't reliably get the mutation observer from an iframe

but we are at least _always_ trying to get it from an iframe instead of only when angular has monkeyed around with it

let's reduce the number of people affected by checking correctly

i've tested locally that this doesn't freeze the test angular app that would freeze if this didn't work

and i've tested locally that i can record safari (as much as i can, it wasn't _always_ failing)

### DIFF
--- a/patches/@rrweb__record@2.0.0-alpha.17.patch
+++ b/patches/@rrweb__record@2.0.0-alpha.17.patch
@@ -1,45 +1,32 @@
 diff --git a/dist/record.js b/dist/record.js
-index 46ec389fefb698243008b39db65470dbdf0a3857..6c62dab78676066344afb27976dd7d83a7e17c58 100644
+index 46ec389fefb698243008b39db65470dbdf0a3857..efc65d8c393de8d0ad35083cba315aef91a04c2c 100644
 --- a/dist/record.js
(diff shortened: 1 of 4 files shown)


patches/@rrweb__record@2.0.0-alpha.17.patch
```diff
@@ -1,37 +1,32 @@
...
-+    if (!isFunction(x)) {
-+        return false;
-+    }
-+    const prototypeKeys = Object.getOwnPropertyNames(x.prototype || {});
-+    return prototypeKeys.some((key) => key.indexOf('__zone'));
++    return !!globalThis.Zone
 +  } catch {
-+    // we've seen some intermittent problems in Safari since introducing this check
 +    return false
 +  }
```
pauldambra, line 15: turns out it is way simpler to just check if the global zone is present
if it's not, then angular didn't taint this

just like with the previous solution, folk can configure this to not be called zone, but then they can run posthog outside of the zone 🤷

thmsobrmlr (approved):

Issue #1440, opened 2024-09-27
BigInt serialization error in encodePostData


zimrain:
Our React app uses BigInt in many places. Only autocapture is on, no custom events. We're getting the following: "TypeError: Do not know how to serialize a BigInt"

Source map points back to `JSON.stringify` call in `encodePostData` here: https://github.com/PostHog/posthog-js/blob/26150b197dc1a2f5241b8d29ee7a359331a1a0a4/src/request.ts#L48

jonahallibone:
same error here

TommySoFM:
Having the same issue here too.
Have to turn off auto-capture to stop the error from appearing

robbie-c:
Hi @zimrain - it'd be pretty easy for us to serialize BigInts as a string, but can you explain how a BigInt might have gotten into an event payload? E.g. are you setting it as a user property? It'd be good to know what steps I'd need to take to reproduce this issue and confirm my fix on.

zimrain:
Just auto-capture, we didn't have any manual events/properties

robbie-c:
Do you use data attributes to add extra properties? https://posthog.com/docs/product-analytics/autocapture#tracking-metadata

PR #1573, opened 2024-11-29
fix: Don't crash on bigints

robbie-c:
## Changes
Fixes https://github.com/PostHog/posthog-js/issues/1440

I'm a bit wary that there's another root cause we're missing, I would like to hear from those users where the BigInts came from in the first place, but in the mean-time let's not crash

## Checklist

### DIFF
--- a/src/__tests__/request.test.ts
+++ b/src/__tests__/request.test.ts
@@ -343,6 +343,21 @@ describe('request', () => {
                 'application/x-www-form-urlencoded'
             )
         })
+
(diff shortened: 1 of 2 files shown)


pauldambra (approved):
one less crash in the world 🚢

dandyling:
@robbie-c We are using protobufs i64, and that is converted to bigint in the front-end

robbie-c:
@dandyling were you experiencing this crash before, and does this fix it?

PR #1575, opened 2024-12-02
fix: parse Chrome instead of Android as browser

daibhin:
## Changes

Based on some Googling it seems like the latter value in this user agent (`Chrome/28.0.1500.94 `) is actually the browser.

_Maybe_ we don't want to make this change because it will affect peoples existing browser attribution but I think it's totally reasonable to ship a fix forward here


pauldambra (approved):
👍

PR #1577, opened 2024-12-03
feat: RemoteConfig loader

benjackwhite:
## Changes

...

## Checklist

### DIFF
--- a/src/decide.ts
+++ b/src/decide.ts
@@ -1,9 +1,19 @@
 import { PostHog } from './posthog-core'
-import { Compression, DecideResponse } from './types'
+import { Compression, DecideResponse, RemoteConfig } from './types'
 import { STORED_GROUP_PROPERTIES_KEY, STORED_PERSON_PROPERTIES_KEY } from './constants'
 
 import { logger } from './utils/logger'
-import { document } from './utils/globals'
+import { assignableWindow, document } from './utils/globals'
+
+// TODO: Add check for global config existing.
+// Modify the whole "afterDecideResponse" function to be a method of the PostHog class
+
+// 1. Option is to load the config endpoint (if __preview is enabled) and then if flags are needed call decide, and then call "afterDecideResponse" with the combined response :thinking:
+// 2. Other option is to separate out the values so that we have a "config response" and a "flags response" separately...
+
+// TODO: Fix WebExperiments to wait for flags instead of only decide
+// TODO: Fix Surveys to do the same
+// TODO: Background refresh of the config - every 5 minutes to match CDN cache - at least when active
 
 export class Decide {
     constructor(private readonly instance: PostHog) {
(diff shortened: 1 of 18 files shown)


src/decide.ts
```diff
@@ -63,6 +129,31 @@ export class Decide {
...
+        }
+
+        this.instance._onRemoteConfig(config)
+
+        if (config.hasFeatureFlags !== false) {
+            // TRICKY: This is set in the parent for some reason...
+            this.instance.featureFlags.setReloadingPaused(false)
+            // If the config has feature flags, we need to call decide to get the feature flags
+            // This completely separates it from the config logic which is good in terms of separation of concerns
+            this.instance.featureFlags.reloadFeatureFlags()
```
pauldambra, line 156: don't really need to change this now... i wonder how much order of operations is important here.

would be nice for `this.instance.featureFlags` to be listening to `onRemoteConfig` and run these things itself to break this dependency

but only noticing it because this has all moved

src/types.ts
```diff
@@ -338,6 +338,12 @@ export interface PostHogConfig {
      * whether to wrap fetch and add tracing headers to the request
      * */
     __add_tracing_headers?: boolean
+
+    /**
+     * PREVIEW - MAY CHANGE WITHOUT WARNING - DO NOT USE IN PRODUCTION
+     * whether to wrap fetch and add tracing headers to the request
+     * */
```
pauldambra, line 345:
```suggestion
     * */
```

src/decide.ts
```diff
@@ -1,17 +1,83 @@
...
+                return
+            }
+
+            if (disableRemoteCalls) {
+                logger.warn('Remote config is disabled. Falling back to local config.')
+                return
+            }
+
+            // Attempt 2 - if we have the external deps loader then lets load the script version of the config that includes site apps
+            this._loadRemoteConfigJs((config) => {
```
pauldambra, line 61: this is tricky to test the unhappy path locally

if I turn on __preview_remote_config against current master then my local posthog returns a 200 with all the react boostrap which would show 404

but instead fails here because the mime type is unexpected

and then we call onRemoteConfig where I'd expect us to fallback to decide

other than that (admittedly edgy case) this all seems solid

src/decide.ts
```diff
@@ -1,17 +1,83 @@
...
+        }
+    }
+
+    private _loadRemoteConfigJSON(cb: (config?: RemoteConfig) => void): void {
+        this.instance._send_request({
+            method: 'GET',
+            url: this.instance.requestRouter.endpointFor('assets', `/array/${this.instance.config.token}/config`),
+            callback: (response) => {
+                cb(response.json as RemoteConfig | undefined)
+            },
```
pauldambra, line 31: i guess we need to handle errors here?

pauldambra (approved):
seems to work when hooked up with the parent PR though

so other than a though about error handling which might be fine 🚢

src/decide.ts
benjackwhite: Yeah the changes for the config aren't in yet so it makes sense that that path fails. Feels okay given the whole point of this is for me to gradually test and improve it

src/decide.ts
benjackwhite: I guess but on the other hand this is the same as the current behaviour. If it fails, there's not a whole lot we can do about it. Same as if decide fails.

src/decide.ts
benjackwhite: Yeah I'd be tempted to clean this up later. The main thing is I really dont want to do fundamental changes whilst we still rely on the old decide path.

PR #1580, opened 2024-12-05
feat: Load site functions via RemoteConfig

benjackwhite:
## Changes

Needs https://github.com/PostHog/posthog-js/pull/1581
Needs https://github.com/PostHog/posthog/pull/26671

- [x] Moves the site functions to _only_ be loaded the new way
- [x] Changes the loading mentality so that the caller is responsible for initializing and passing things in (this is pretty important as we don't want the server side code to be calling non-top-level methods as they might change over time).

## Checklist

### DIFF
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -62,10 +62,12 @@ if (typeof window !== 'undefined') {
         persistence: cookieConsentGiven() ? 'localStorage+cookie' : 'memory',
         person_profiles: PERSON_PROCESSING_MODE === 'never' ? 'identified_only' : PERSON_PROCESSING_MODE,
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
+        opt_in_site_apps: true,
         __preview_remote_config: true,
         ...configForConsent(),
     })
-    // Help with debugging(window as any).posthog = posthog
+    // Help with debugging
+    ;(window as any).posthog = posthog
 }
 
 export const posthogHelpers = {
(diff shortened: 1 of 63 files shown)


playground/nextjs/src/posthog.ts
```diff
@@ -62,10 +62,12 @@ if (typeof window !== 'undefined') {
         persistence: cookieConsentGiven() ? 'localStorage+cookie' : 'memory',
         person_profiles: PERSON_PROCESSING_MODE === 'never' ? 'identified_only' : PERSON_PROCESSING_MODE,
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
+        opt_in_site_apps: true,
         __preview_remote_config: true,
         ...configForConsent(),
     })
-    // Help with debugging(window as any).posthog = posthog
+    // Help with debugging
+    ;(window as any).posthog = posthog
```
benjackwhite, line 70: Decide this should always be on

src/types.ts
```diff
@@ -341,6 +341,7 @@ export interface PostHogConfig {
 
     /**
      * PREVIEW - MAY CHANGE WITHOUT WARNING - DO NOT USE IN PRODUCTION
+     * whether to wrap fetch and add tracing headers to the request
```
benjackwhite, line 344:
```suggestion
     * enables the new RemoteConfig approach to loading config instead of decide
```

mariusandra (approved):
Looks good. The only question I had was if legacy site apps still work, but reading the code more it seems like that's handled. So ✅

benjackwhite:
> Looks good. The only question I had was if legacy site apps still work, but reading the code more it seems like that's handled. So ✅

They _should_ do. The actual path for loading them is basically identical to before. I'll test again for sure before merging thouh.

PR #1582, opened 2024-12-06
chore: cypress test with fake angular tainting prototypes

pauldambra:
it has been a pain to see this bug in prod and not be able to track or reproduce
let's see if CI can protect us in future


pauldambra:
this new test doesn't fail in webkit, and the experimental webkit support in cypress is v    e        r      y         s            l        o     w

PR #1587, opened 2024-12-08
feat: send snapshot library

pauldambra:
part of https://github.com/PostHog/posthog/pull/26747

we want to be able to filter / report on recordings from different platforms and we can't completely determine them from data already sent. let's send from the web sdk first

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -46,6 +46,7 @@ import { MutationRateLimiter } from './mutation-rate-limiter'
 import { gzipSync, strFromU8, strToU8 } from 'fflate'
 import { clampToRange } from '../../utils/number-utils'
 import { includes } from '../../utils'
+import Config from '../../config'
 
 const LOGGER_PREFIX = '[SessionRecording]'
 const logger = createLogger(LOGGER_PREFIX)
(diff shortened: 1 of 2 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -1185,6 +1186,7 @@ export class SessionRecording {
                     $snapshot_data: snapshotBuffer.data,
                     $session_id: snapshotBuffer.sessionId,
                     $window_id: snapshotBuffer.windowId,
+                    $snapshot_library: `web/ver/${Config.LIB_VERSION}/script_name/${this.scriptName}`,
```
marandaneto, line 1189: Mobile SDKs do $lib and $lib_version, can we reuse those?

src/extensions/replay/sessionrecording.ts
```diff
@@ -1185,6 +1186,8 @@ export class SessionRecording {
                     $snapshot_data: snapshotBuffer.data,
                     $session_id: snapshotBuffer.sessionId,
                     $window_id: snapshotBuffer.windowId,
+                    $lib: 'web',
+                    $lib_version: Config.LIB_VERSION,
```
marandaneto, line 1190: I noticed the web uses the `v{version}` pattern and mobile only `{version}, not a problem I guess?

marandaneto (approved):

src/extensions/replay/sessionrecording.ts
pauldambra: ah, yep, it's just a bad mock.. updated to avoid confusion!

PR #1588, opened 2024-12-08
feat: lazy load session recording

pauldambra:
move session recording entirely to lazy loaded not just the rrweb portions

tested locally and can still record

still need to figure out how to safely release this since it will be published to NPM before the new lazy bundle is available on the CDN

### DIFF
--- a/src/__tests__/extensions/replay/sessionrecording.test.ts
+++ b/src/__tests__/extensions/replay/sessionrecording.test.ts
@@ -48,6 +48,7 @@ import LazyLoadedSessionRecording, {
     RECORDING_IDLE_THRESHOLD_MS,
     RECORDING_MAX_EVENT_SIZE,
 } from '../../../entrypoints/session-recorder'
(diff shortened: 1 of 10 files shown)


src/__tests__/posthog-core.ts
```diff
@@ -825,7 +842,10 @@ describe('posthog core', () => {
             posthog.sessionRecording = {
                 afterDecideResponse: jest.fn(),
                 startIfEnabledOrStop: jest.fn(),
-            } as unknown as SessionRecording
+            } as OnlyValidKeys<
```
pauldambra, line 845: interesting... i expected this change to highlight the incorrect setup here that the `as unknown` is hiding

src/__tests__/extensions/replay/sessionrecording.test.ts
pauldambra, line 1: need to add tests on the loader separately from the loaded file

pauldambra:
too unsafe to roll out, needs a different approach

PR #1596, opened 2024-12-11
fix: Refactor decide loading

benjackwhite:
## Changes

This is a beafy refactor that came out of me having issues around how the RemoteConfig was working.

I basically found our code is super buggy around choosing when to load flags because there were two different things calling decide with a lot of bool properties that weren't really accurate to what they were doing

The main improvements here are:
- Move all the decide logic into feature flags
- Make it more deterministic so that the load is always called unless it is actively disabled.
- Fixes some weirdness I think that existed where in certain cases the flags would actually _not_ be loaded correctly (you can sort of see this that the anon-id wasn't included in one test when it totally should have been!) 

## Checklist


playground/nextjs/pages/_app.tsx
```diff
@@ -33,7 +33,9 @@ export default function App({ Component, pageProps }: AppProps) {
         }
     }, [])
 
-    const localhostDomain = process.env.NEXT_PUBLIC_CROSSDOMAIN ? 'https://localhost:8000' : 'http://localhost:8000'
+    const localhostDomain = process.env.NEXT_PUBLIC_CROSSDOMAIN
```
benjackwhite, line 36: Helper if runnning on localhost:8010

src/__tests__/remote-config.test.ts
```diff
@@ -0,0 +1,108 @@
+import { RemoteConfigLoader } from '../remote-config'
+import { RequestRouter } from '../utils/request-router'
+import { PostHog } from '../posthog-core'
+import { PostHogConfig, RemoteConfig } from '../types'
+import '../entrypoints/external-scripts-loader'
+import { assignableWindow } from '../utils/globals'
+
+describe('RemoteConfigLoader', () => {
```
benjackwhite, line 8: This was Decide but now it only handles config loading

src/posthog-core.ts
```diff
@@ -545,6 +548,14 @@ export class PostHog {
     }
 
     _onRemoteConfig(config: RemoteConfig) {
+        if (!(document && document.body)) {
```
benjackwhite, line 551: This was in decide before but should actually be here as this is the lower level call.

dmarticus (approved):
seems good by my judgement, thanks for taking the time to dive in and pull things apart!

functional_tests/feature-flags.test.ts
```diff
@@ -178,7 +178,7 @@ describe('FunctionalTests / Feature Flags', () => {
             expect(getRequests(token)['/decide/']).toEqual([
                 // This is the initial call to the decide endpoint on PostHog init, with all info added from `loaded`.
                 {
-                    // $anon_distinct_id: 'anon-id', // no anonymous ID is sent because this was overridden on load
+                    $anon_distinct_id: 'anon-id',
```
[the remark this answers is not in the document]
benjackwhite: I'm not certain but I can't see any reason why we _wouldn't_ have wanted the anon id involved. The comment about it not being set because it was overridden is not true at all. It just wasn't set because decide.ts was the previous thing that triggered first and for some reason it didn't include it.

src/__tests__/remote-config.test.ts
```diff
@@ -0,0 +1,108 @@
...
+    let posthog: PostHog
+
+    beforeEach(() => {
+        // clean the JSDOM to prevent interdependencies between tests
```
pauldambra, line 12: move/delete this comment?

PR #1597, opened 2024-12-11
feat(Surveys): Enable partial survey responses

Phanatic:
## Changes

This PR enables the SDK to send partial survey responses as users fill out the survey, not at the end once all questions are answered : https://github.com/PostHog/posthog/issues/19300
We introduced a new config option called `disable_survey_partial_response` which is true by default. Customers have to set this to `false` to get the feature enabled

### Not supported
* Continuing filling  out a survey after a browser session refresh. We will send question responses as they are answered, and don't allow you to fill out a few questions, come back later, and then fill them out. 

## Checklist


src/posthog-core.ts
```diff
@@ -145,6 +145,7 @@ export const defaultConfig = (): PostHogConfig => ({
     disable_persistence: false,
     disable_web_experiments: true, // disabled in beta.
     disable_surveys: false,
+    disable_survey_partial_response: false,
```
marandaneto, line 148:
```suggestion
    disable_survey_partial_response: true,
```

should be disabled by default right?

PR #1605, opened 2024-12-16
chore: add websockets example to nextjs playground

pauldambra:
taking baby steps to avoid a conflict with someone else's changes to the playground

this adds a chat page using websockets to the playground, next step is to wrap the websocket client so we use network requests over sockets in replay


pauldambra:
going to merge pre review since it's just the playground, i'm still on this so if folk have feedback i can roll it into next PR

PR #1607, opened 2024-12-16
fix: Load RemoteConfig by default but don't use it

benjackwhite:
## Changes

Part of the rollout testing. Because this is replacing decide we can't use decide to gradually rollout. Instead I'm turning it on for everyone but only using it if the preview flag is enabled. That way I can poke around implementations and see if its working as well as checking backend metrics

## Checklist

### DIFF
--- a/src/remote-config.ts
+++ b/src/remote-config.ts
@@ -31,14 +31,6 @@ export class RemoteConfigLoader {
     }
 
     load(): void {
-        // Call decide to get what features are enabled and other settings.
-        // As a reminder, if the /decide endpoint is disabled, feature flags, toolbar, session recording, autocapture,
-        // and compression will not be available.
-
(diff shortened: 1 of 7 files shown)


src/remote-config.ts
```diff
@@ -31,40 +31,36 @@ export class RemoteConfigLoader {
...
-            this.onRemoteConfig(assignableWindow._POSTHOG_CONFIG)
-            return
-        }
-
-        if (this.instance.config.advanced_disable_decide) {
-            // This setting is essentially saying "dont call external APIs" hence we respect it here
-            logger.warn('Remote config is disabled. Falling back to local config.')
-            return
-        }
+        try {
```
benjackwhite, line 34: diff is just the try catch and moving the "preview" check into the callback function

PR #1610, opened 2024-12-16
chore: Small tweaks after making `fetch` the default request implementation

### LINKED ISSUE #1487: TypeError: Load failed when page is unloaded before posthog events are finished sending
We are getting thousands of "TypeError: Load failed" errors in Sentry and this appears to be because PostHog has a request in flight while we are redirecting the user to an external URL.

It looks like this is because [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) is preferred _over_ [fetch](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API). 
https://github.com/PostHog/posthog-js/blob/38744a512abef2f4eb7327e7a3c23d184c3993ca/src/request.ts#L202-L216

Indeed, `XHR` is the default transport:
https://github.com/PostHog/posthog-js/blob/38744a512abef2f4eb7327e7a3c23d184c3993ca/src/request.ts#L237

From my understanding, [XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest) does not have a mechanism like [keepalive](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive) that will not cancel the request when the page is unloaded.

Fetch is available on [every major supported browser](https://caniuse.com/fetch). Could `fetch` be made the default? If not, could an option be provided to make it the default for our instance?

Alternatively, it could be helpful to have a method that returns a promise that resolves when all the pending events have been sent so we can block the redirect on that request, but that seems unnecessary with [keepalive](https://developer.mozilla.org/en-US/docs/Web/API/Request/keepalive).

rafaeelaudibert:
## Changes

Lets reorder our request implementation to use fetch as the default in case someone specifies an unknown method, rather than using XHR

Closes #1487

## Checklist

### DIFF
--- a/cypress/e2e/session-recording.cy.ts
+++ b/cypress/e2e/session-recording.cy.ts
@@ -255,7 +255,7 @@ describe('Session recording', () => {
                         [/https:\/\/example.com/, 'fetch'],
                     ]
 
-                    // yay, includes expected type 6 network data
+                    // yay, includes expected network data
                     expect(capturedRequests.length).to.equal(expectedCaptureds.length)
                     expectedCaptureds.forEach(([url, initiatorType], index) => {
                         expect(capturedRequests[index].name).to.match(url)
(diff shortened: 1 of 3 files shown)


pauldambra (approved):
👍

PR #1612, opened 2024-12-17
fix: Change default transport to fetch

benjackwhite:
## Changes

...

## Checklist

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -433,7 +433,7 @@ export class PostHogFeatureFlags {
 
         if (!existing_early_access_features || force_reload) {
             this.instance._send_request({
-                transport: 'XHR',
+                transport: 'fetch',
                 url: this.instance.requestRouter.endpointFor(
                     'api',
                     `/api/early_access_features/?token=${this.instance.config.token}`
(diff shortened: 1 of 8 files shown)


src/posthog-featureflags.ts
```diff
@@ -433,7 +433,7 @@ export class PostHogFeatureFlags {
 
         if (!existing_early_access_features || force_reload) {
             this.instance._send_request({
-                transport: 'XHR',
+                transport: 'fetch',
```
pauldambra, line 436: can this just not specify their own transport and use the default instead?

src/posthog-surveys.ts
```diff
@@ -118,7 +118,7 @@ export class PostHogSurveys {
                     `/api/surveys/?token=${this.instance.config.token}`
                 ),
                 method: 'GET',
-                transport: 'XHR',
+                transport: 'fetch',
```
pauldambra, line 121: can this just not specify their own transport and use the default instead?

src/web-experiments.ts
```diff
@@ -151,7 +151,7 @@ export class WebExperiments {
                 `/api/web_experiments/?token=${this.instance.config.token}`
             ),
             method: 'GET',
-            transport: 'XHR',
+            transport: 'fetch',
```
pauldambra, line 154: can this just not specify their own transport and use the default instead?

didn't realise there were three when i started commenting on each 🤣

src/posthog-featureflags.ts
benjackwhite: Probably.

pauldambra: https://github.com/PostHog/posthog-js/pull/1614

PR #1620, opened 2024-12-19
fix: type error accessing null object and added test case

supermar1010:
## Changes
We stumbled onto this bug, with rxdb [dispatching mouse events](https://github.com/pubkey/rxdb/blob/e045257ed3bef3e8c01032d5a91e6260ce0b8c2e/src/plugins/replication/replication-helper.ts#L116) without data and thus producing a posthog crash. 

[This typecast](https://github.com/supermar1010/posthog-js/blob/7d11d7f33ecce50367777694172d01d22187314d/src/heatmaps.ts#L180) is why the error wasn't spotted by ts. The type of `e.target` is originally `EventTarget | null`, but looses the `| null` in this cast.

## Checklist

### DIFF
--- a/src/__tests__/heatmaps.test.ts
+++ b/src/__tests__/heatmaps.test.ts
@@ -89,6 +89,27 @@ describe('heatmaps', () => {
         expect(posthog.heatmaps!.getAndClearBuffer()).toBeDefined()
     })
 
+    it('should handle empty mouse moves', async () => {
+        posthog.heatmaps?.['_onMouseMove']?.(new Event('mousemove'))
+
+        jest.advanceTimersByTime(posthog.heatmaps!.flushIntervalMilliseconds + 1)
+
+        expect(beforeSendMock).toBeCalledTimes(1)
+        expect(beforeSendMock.mock.lastCall[0]).toMatchObject({
+            event: '$$heatmap',
+            properties: {
+                $heatmap_data: {
+                    'http://replaced/': [
+                        {
+                            target_fixed: false,
+                            type: 'mousemove',
+                        },
+                    ],
+                },
+            },
+        })
+    })
+
     it('should send rageclick events in the same area', async () => {
         posthog.heatmaps?.['_onClick']?.(createMockMouseEvent())
         posthog.heatmaps?.['_onClick']?.(createMockMouseEvent())
--- a/src/__tests__/heatmaps.test.ts
+++ b/src/__tests__/heatmaps.test.ts
@@ -144,16 +144,24 @@ describe('heatmaps', () => {
     })
 
     it('should ignore clicks if they come from the toolbar', async () => {
+        const testElementToolbar = document.createElement('div')
+        testElementToolbar.id = '__POSTHOG_TOOLBAR__'
+
         posthog.heatmaps?.['_onClick']?.(
             createMockMouseEvent({
-                target: { id: '__POSTHOG_TOOLBAR__' } as Element,
+                target: testElementToolbar,
             })
         )
         expect(posthog.heatmaps?.['buffer']).toEqual(undefined)
 
+        const testElementClosest = document.createElement('div')
+        testElementClosest.closest = () => {
+            return {}
+        }
+
         posthog.heatmaps?.['_onClick']?.(
             createMockMouseEvent({
-                target: { closest: () => ({}) } as unknown as Element,
+                target: testElementClosest,
             })
         )
         expect(posthog.heatmaps?.['buffer']).toEqual(undefined)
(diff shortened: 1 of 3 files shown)


src/utils/element-utils.ts
```diff
@@ -1,8 +1,8 @@
 import { TOOLBAR_CONTAINER_CLASS, TOOLBAR_ID } from '../constants'
 
-export function isElementInToolbar(el: Element): boolean {
+export function isElementInToolbar(el: Element | null): boolean {
```
pauldambra, line 3: this is great, thank you!

src/utils/element-utils.ts
```diff
@@ -1,8 +1,8 @@
...
     // NOTE: .closest is not supported in IE11 hence the operator check
```
pauldambra, line 4: totally fly-by note for myself that this comment is probably out of date now that we're making ES6 and ES5 bundles separately

src/__tests__/heatmaps.test.ts
```diff
@@ -89,6 +89,27 @@ describe('heatmaps', () => {
...
+        jest.advanceTimersByTime(posthog.heatmaps!.flushIntervalMilliseconds + 1)
+
+        expect(beforeSendMock).toBeCalledTimes(1)
+        expect(beforeSendMock.mock.lastCall[0]).toMatchObject({
+            event: '$$heatmap',
+            properties: {
+                $heatmap_data: {
+                    'http://replaced/': [
+                        {
+                            target_fixed: false,
```
pauldambra, line 104: so, does this mean that the mousemove event has no clientX or clientY?

if so we're actually better ignoring them completely since the data is invalid without a coordinate.

src/utils/element-utils.ts
pauldambra, line 1: i wish GitHub let you comment outside of the diff :)

https://github.com/PostHog/posthog-js/blob/a334cc31249d7303cc757fcef2fdecba93aeae94/src/heatmaps.ts#L163

we can remove the cast here entirely now I think... one less `as` in the world is a good thing

pauldambra (changes requested):
1 question and 1 requested change

but this is 💯, absolutely love you took the time to PR 🙌

src/__tests__/heatmaps.test.ts
supermar1010: Yes, it's a mouse move event without any clientX or clientY. According to [mozilla dev](https://developer.mozilla.org/en-US/docs/Web/API/Event/Event) everything is optional, so it is valid to send an event with any data.

Yeah ignoring is probably a good call!

src/utils/element-utils.ts
supermar1010: Done

src/utils/element-utils.ts
```diff
@@ -1,8 +1,11 @@
 import { TOOLBAR_CONTAINER_CLASS, TOOLBAR_ID } from '../constants'
 
-export function isElementInToolbar(el: Element): boolean {
-    // NOTE: .closest is not supported in IE11 hence the operator check
-    return el.id === TOOLBAR_ID || !!el.closest?.('.' + TOOLBAR_CONTAINER_CLASS)
+export function isElementInToolbar(el: EventTarget | null): boolean {
+    if (el instanceof Element) {
```
supermar1010, line 4: Needs an if here now, because `EventTarget` does not have `id` and `closest`.

src/__tests__/heatmaps.test.ts
pauldambra: you're welcome to add that to this PR - if you're interested?

i can do as follow-up otherwise

src/utils/element-utils.ts
supermar1010: That doesn't work type wise. `isElementNode` only accepts Element as a param, but el is EventTarget. So I don't see an alternative to the instanceof right now

pauldambra: Ah, of course... cool cool and good general support for instanceof so pretty safe 🤞

src/__tests__/heatmaps.test.ts
pauldambra: we're already dropping events with no x and y during  ingestion, so it's only a nice-to-have not to omit them (although good to not do work)

so only open question for me is if you'd like to handle that in this PR?

supermar1010: I think I'd leave that to you, I'm not sure where you'd implement that filtering.

supermar1010:
Unit tests should run successful now!

pauldambra (approved):
LGTM

PR #1623, opened 2024-12-19
chore: fix changelog after automation error

pauldambra:
CD task was broken because of linting which meant we never published 1.202.3 and 1.202.4 is mungled up with 1.202.5

this updates the changelog to represent what is in each release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,17 @@
 ## 1.202.5 - 2024-12-19
 
-
+- chore: lint a file that needs it (#1622)
 
 ## 1.202.4 - 2024-12-19
 
-
+- fix: do not even send heatmap with no x or y (#1621)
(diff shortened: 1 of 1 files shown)


rafaeelaudibert (approved):
Was 1.202.4 published? If not, we could simply put all of this inside `1.202.3` instead

It's terrible that npm allowed us to publish 1.202.4 without publishing 1.202.3

pauldambra:
> Was 1.202.4 published? If not, we could simply put all of this inside 1.202.3 instead

yep, 1.202.3 was never published, 1.202.4 was published but the automation that updates the changelog can't cope with anything but a clean set of releases (it doesn't even cope with two merges too close together 🙈)

> It's terrible that npm allowed us to publish 1.202.4 without publishing 1.202.3

they only enforce that you don't try and publish a duplicate iiuc

PR #1624, opened 2024-12-19
chore: add feature name prop to event

daibhin:
## Changes

Adds the `$name` prop to an early access enrolment event cc @joethreepwood 

## Checklist

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -8,6 +8,7 @@ import {
     Properties,
     JsonType,
     Compression,
+    EarlyAccessFeature,
 } from './types'
 import { PostHogPersistence } from './posthog-persistence'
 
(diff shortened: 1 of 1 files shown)


benjackwhite (approved):
Is name perhaps a little to generic? On the other hand maybe having it as a generic property could be a good thing if we start using it more 🤔

daibhin:
> Is name perhaps a little to generic? On the other hand maybe having it as a generic property could be a good thing if we start using it more 🤔

No I think you're right. I'd convinced myself `$name` was ok given it was scoped to the event but forgot that you are often searching for event properties. Doesn't make sense to be so generic in that case. Updated to `$early_access_feature_name`

joethreepwood:
Just as a small follow-up here as I'm struggling to see how this event works now, actually. 

I've basically set up email flows. One for onboarding, one for offboarding. The triggers look like: 

**Onboarding:** Triggers if`$feature_enrollment_update` occurs where `$feature_enrollment` equals `TRUE`
**Offboarding:** Triggers if`$feature_enrollment_update` occurs where `$feature_enrollment` equals `FALSE`

My understanding was that this would basically enable the campaigns to trigger if someone opted In/Out of a beta feature. 

In practice though, 48 hours in and we've seen that the Onboarding email hasn't triggered at all, while the Offboarding email has triggered far too many times and for all the wrong customers. I've looked at replays for impacted customers and the event doesn't seem to trigger anywhere near the beta menu. 

I can see in [the code](https://github.com/PostHog/posthog-js/blob/main/src/posthog-featureflags.ts) that there may be another key or event property I need to use, but I'm hoping I can get a quick steer so I can solve this myself?

joethreepwood:
Solved Async with David!

PR #1625, opened 2024-12-19
chore: build against lowest and current version of TS

pauldambra:
i advised a customer to upgrade and they couldn't because we had an implicit minimum TS version, luckily they could upgrade without a major TS verson upgrade the friday before xmas by applying a minor update

let's test that discovered minimum so we know if we break it in future

we discovered you can't run tsc in a project lower than TS 4.7.2 that imports posthog
(for the benefit of folk coming across this issue - this doesn't affect JS projects or the HTML snippet)

---

this builds the library under different node versions and then imports it into a file and runs tsc on that file

here's that failing using the customer's too low version

Screenshot 2024-12-19 at 21 37 20

### DIFF
--- a/.github/workflows/minimum-version-ts-check.yaml
+++ b/.github/workflows/minimum-version-ts-check.yaml
@@ -13,7 +13,7 @@ jobs:
         strategy:
           matrix:
             node-version: [ 18, 20 ]
-            typescript-version: [ '5.5.4' ]
+            typescript-version: [ '4.7.2', '5.5.4' ]
         steps:
             - name: Checkout repository
               uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


.github/workflows/minimum-version-ts-check.yaml
```diff
@@ -0,0 +1,46 @@
...
+            - main
+
+jobs:
+    unit:
+        name: Check minimum TS version
+        runs-on: ubuntu-latest
+        strategy:
+          matrix:
+            node-version: [ 16, 18, 20, 22, 23 ]
+            typescript-version: ['4.7.2', '5.5.4' ]
```
rafaeelaudibert, line 16: Maybe you could do an explicit `latest` here instead? This way we don't need to keep updating this every time Microsoft decides they want to improve speed by 0.0001%

rafaeelaudibert (approved):

PR #1626, opened 2024-12-20
fix: recordings that are paused for their whole duration

pauldambra:
see: https://posthoghelp.zendesk.com/agent/tickets/22367 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/24535)

The user is using the URL pause list to entirely block domains. We wrote the feature thinking about pages within otherwise captured recordings so we don't completely stop emitting when paused. Which meant in combination with full page loads we would eventually emit an e.g. 40 minute session, with an apparent length of 0 seconds, that had an initial empty full snapshot and some network calls but nothing else

no bueno

let's not do that. we now explicitly check if the recording is paused and do not flush to the backend if so.

### DIFF
--- a/src/__tests__/extensions/replay/sessionrecording.test.ts
+++ b/src/__tests__/extensions/replay/sessionrecording.test.ts
@@ -47,7 +47,6 @@ import {
 } from '@rrweb/types'
 import Mock = jest.Mock
 import { ConsentManager } from '../../../consent'
-import { waitFor } from '@testing-library/preact'
 import { SimpleEventEmitter } from '../../../utils/simple-event-emitter'
 
 // Type and source defined here designate a non-user-generated recording event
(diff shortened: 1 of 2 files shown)


src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -2312,48 +2311,55 @@ describe('SessionRecording', () => {
             )
         })
 
-        it('flushes buffer and includes pause event when hitting blocked URL', async () => {
+        it('does not flush buffer and includes pause event when hitting blocked URL', async () => {
```
pauldambra, line 2314: that means that flushing before pausing is no longer safe.

really we should detect if we've ever flushed and then flush or not based on that but it's better to avoid this completely

src/extensions/replay/sessionrecording.ts
```diff
@@ -1269,17 +1269,15 @@ export class SessionRecording {
             return
         }
 
+        // we can't flush the buffer here since someone might be starting on a blocked page,
+        // and we need to be sure that we don't record that page
+        // so we might not get the below custom event but events will report the paused status
+        // which will allow debugging of sessions that start on blocked pages
         this._urlBlocked = true
-        document?.body?.classList?.add('ph-no-capture')
```
pauldambra, line 1273: this snuck past me in the initial implemention. naughty paul
this doesn't just add belt and braces for not capturing the DOM, it stops other capture like autocapture too... no bueno

src/extensions/replay/sessionrecording.ts
```diff
@@ -1290,7 +1288,6 @@ export class SessionRecording {
         }
 
         this._urlBlocked = false
-        document?.body?.classList?.remove('ph-no-capture')
```
marandaneto, line 1293: All the other changes make sense, this one does not feel related to the original issue?

marandaneto (approved):

src/extensions/replay/sessionrecording.ts
pauldambra: no, it's totally fly-by, commented on above. i don't think we should be doing that

PR #1631, opened 2024-12-27
chore: playwright tests

pauldambra:
Cypress tests don't run in browserstack because they were so slow in browserstack/multiple browsers

Which means session replay tests (and most e2e tests) are only exercised in the electron browser. But we can have e.g. safari specific bugs and would never catch them.

Since we know Cypress was annoyingly slow here let's not use it.

This PR ports all session recording tests from Cypress to Playwright

### Why Playwright?

* mind share
* parallel tests
* faster than cypress (Cypress ~3m46s (1 browser), Playwright ~1m30s (3 browsers) (admittedly a different number of tests))
* by default runs tests in Chrome, Firefox, and WebKit
* we're already using it in visual regression tests


src/extensions/replay/sessionrecording.ts
```diff
@@ -1323,6 +1323,7 @@ export class SessionRecording {
      * */
     public overrideLinkedFlag() {
         this._linkedFlagSeen = true
+        this._tryTakeFullSnapshot()
```
pauldambra, line 1326: genuine bug for this luckily pretty rarely called method

src/extensions/replay/sessionrecording.ts
```diff
@@ -1337,6 +1338,7 @@ export class SessionRecording {
             // short-circuits the `makeSamplingDecision` function in the session recording module
             [SESSION_RECORDING_IS_SAMPLED]: true,
         })
+        this._tryTakeFullSnapshot()
```
pauldambra, line 1341: genuine bug for this luckily pretty rarely called method

PR #1632, opened 2024-12-27
feat: Deprecate `sanitize_properties`

rafaeelaudibert:
## Changes

We can achieve the exact same outcome we achieve with `sanitize_properties` by using `before_send` which is a much more compreheensive function, with better usability and ergonomics.

We can't remove this function without a major, so let's deprecate it and issue a console warning when it's used.

## Checklist

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -1036,6 +1036,7 @@ export class PostHog {
 
         const sanitize_properties = this.config.sanitize_properties
         if (sanitize_properties) {
+            logger.error('sanitize_properties is deprecated. Use before_send instead')
             properties = sanitize_properties(properties, event_name)
         }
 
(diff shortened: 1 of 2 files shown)


src/posthog-core.ts
```diff
@@ -574,8 +574,8 @@ export class PostHog {
             this.compression = includes(config['supportedCompression'], Compression.GZipJS)
                 ? Compression.GZipJS
                 : includes(config['supportedCompression'], Compression.Base64)
-                  ? Compression.Base64
-                  : undefined
+                ? Compression.Base64
+                : undefined
```
rafaeelaudibert, line 578: I don't know why my local setup keeps tweaking this when committing. I know CI failed before because of this, but I can't fix this because my local hook keeps updating it 😮‍💨

rafaeelaudibert: Ok, `pnpm install` did it, I had an outdated version of `prettier` in that repo apparently

rafaeelaudibert:
Website update to suggest `before_send` over `sanitize_properties`: https://github.com/PostHog/posthog.com/pull/10229

pauldambra (approved):
yep, this is even better since `sanitize_properties` is called on properties and $set_once but not on $set

PR #1634, opened 2025-01-02
feat: Add pageview and prev pageview tracking

robbie-c:
## Changes
* Keep track of the last pagview event id
* Send the pageview event id with every event

This came from a discussion with @orian who wanted to be able to associate all of the query completed events that were triggered by the same pageview

## Checklist

### DIFF
--- a/src/page-view.ts
+++ b/src/page-view.ts
@@ -2,8 +2,11 @@ import { window } from './utils/globals'
 import { PostHog } from './posthog-core'
 import { isUndefined } from './utils/type-utils'
 import { clampToRange } from './utils/number-utils'
+import { extend } from './utils'
 
 interface PageViewEventProperties {
+    $pageview_id?: string
+    $prev_pageview_id?: string
     $prev_pageview_pathname?: string
     $prev_pageview_duration?: number // seconds
     $prev_pageview_last_scroll?: number
@@ -18,6 +21,7 @@ interface PageViewEventProperties {
 
 export class PageViewManager {
     _currentPath?: string
+    _currentPageviewId?: string
     _prevPageviewTimestamp?: Date
     _instance: PostHog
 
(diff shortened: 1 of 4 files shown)


src/utils/index.ts
```diff
@@ -52,15 +52,18 @@ export function each(obj: any, iterator: (value: any, key: any) => void | Breake
     }
 }
 
-export const extend = function (obj: Record<string, any>, ...args: Record<string, any>[]): Record<string, any> {
+export const extend = function <T1 extends Record<string, any>, T2 extends Record<string, any>>(
+    obj: T1,
+    ...args: T2[]
+): T1 & T2 {
```
rafaeelaudibert, line 58: Minot typing nit, but we might want to revert this

This type is not generally correct, it's only true if `T1` and `T2` dont overlap. A more accurate type might be `Omit<T1, keyof T2> & T2`

The former implies that `extend({a: 5}, {a: "str"})` would return `{a: number} & {a: string}` which besides equating to `never`, it's not a reasonable type. The right return type will be just `{a: string}`

Even that is wrong because of undefined and such, but it would be slightly more correct

Second minor problem: your new type forces all `args` to have the same shape, whilst before it was much more permissible, allowing a list of objects with different types

rafaeelaudibert: Haven't reviewed the "actual" code, will do so on Monday :)

robbie-c: Happy to revert it, I used this for something that didn't end up making it into this PR, so it's not even useful!

robbie-c: It's reverted, I made a better version of this type in a separate PR

src/posthog-core.ts
```diff
@@ -983,9 +989,11 @@ export class PostHog {
         if (!this.config.disable_scroll_properties) {
             let performanceProperties: Record<string, any> = {}
```
pauldambra, line 990: fly-by since i've never actively read this section... not jazzed about `performanceProperties` since it's not really about performance. could be `pageViewProperties` or something but really very nitty 😅

src/page-view.ts
```diff
@@ -17,40 +20,45 @@ interface PageViewEventProperties {
 }
 
 export class PageViewManager {
-    _currentPath?: string
-    _prevPageviewTimestamp?: Date
+    _currentPageview?: { timestamp: Date; pageViewId: string | undefined; pathname: string | undefined }
```
pauldambra, line 23: brain is not fully engaged so sorry if this is a silly question

this is entirely in memory, no? so if my app does full page loads on navigation (i.e. not a SPA) then I won't have a previous pageview id?

or am i misunderstanding?

otherwise the logic for tracking it looks valid to me

robbie-c: This is true - I'll add a comment about why this is OK

pauldambra: nice! thank you! 💖

pauldambra (approved):
didn't run it but that looks safe and correct to me 🚀

orian:
To query events and group the query ClickHouse' events table:

```
SELECT 
    JSONExtractString(properties, '$pageview_id') as pageview_id, 
    uuid, event, 
    timestamp,
    JSONExtractString(properties, 'duration'),
    JSONExtractString(properties, '$time'),
    properties
FROM events
WHERE team_id=1
    AND event='query completed' and timestamp > '2025-01-06'
    AND pageview_id <> ''
--    AND pageview_id = '<pageview id here>'
ORDER BY pageview_id
LIMIT 100;
```

PR #1638, opened 2025-01-03
test(ci): move more of the browser tests to playwright

pauldambra:
let's move more browser tests to playwright...

this does seem to behave differently in webkit than other browsers :/

you can see here that Cypress is taking 4 minutes to run just the survey tests in one browser
playwright 6 minutes in 3 browsers for all the rest


playwright/ua.spec.ts
```diff
@@ -0,0 +1,27 @@
...
+        sessionRecording: {
+            endpoint: '/ses/',
+        },
+        capturePerformance: true,
+    },
+    url: './playground/cypress-full/index.html',
+}
+
+test.describe('User Agent Blocking', () => {
+    test('should pick up that our automated cypress tests are indeed bot traffic', async ({ page, context }) => {
```
rafaeelaudibert, line 18:
```suggestion
    test('should pick up that our automated playwright tests are indeed bot traffic', async ({ page, context }) => {
```

rafaeelaudibert (approved):

PR #1644, opened 2025-01-08
feat: add Better Uptime string to blocked UA

matanyall:
## Changes
added `'better uptime bot'` to `blocked-uas.ts`. The user agent comes from their docs: https://betterstack.com/docs/uptime/frequently-asked-questions/#what-user-agent-does-uptime-use

Let me know if this is the correct way to do this.

## Checklist


pauldambra (approved):
👍 thanks for the contribution

PR #1649, opened 2025-01-14
chore: deduplicate subsequent identify calls

joshsny:
Often duplicate $identify calls are made due to the SDK being incorrectly setup in the app, or due to re-rendering of providers.

The current behaviour is to only deduplicate $identify calls with no properties, but most $identify calls include properties like email, so they are not deduplicated. This PR extends deduplication to calls with properties.

Note: this is my first contribution to js sdk so would appreciate a thorough review.
...

## Checklist
- [x ] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -1431,6 +1431,12 @@ export class PostHog {
             // let the reload feature flag request know to send this previous distinct id
             // for flag consistency
             this.featureFlags.setAnonymousDistinctId(previous_distinct_id)
+
+            this._cachedIdentify = PostHog._getIdentifyHash(
+                new_distinct_id,
+                userPropertiesToSet,
+                userPropertiesToSetOnce
+            )
         } else if (userPropertiesToSet || userPropertiesToSetOnce) {
             if (
                 this._cachedIdentify !==
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -69,7 +69,7 @@ import { Info } from './utils/event-utils'
 import { logger } from './utils/logger'
 import { SessionPropsManager } from './session-props'
 import { isLikelyBot } from './utils/blocked-uas'
-import { extendURLParams, jsonStringify, request, SUPPORTS_REQUEST } from './request'
+import { extendURLParams, request, SUPPORTS_REQUEST } from './request'
 import { Heatmaps } from './heatmaps'
 import { ScrollManager } from './scroll-manager'
 import { SimpleEventEmitter } from './utils/simple-event-emitter'
@@ -83,6 +83,7 @@ import { PostHogExceptions } from './posthog-exceptions'
 import { SiteApps } from './site-apps'
 import { DeadClicksAutocapture, isDeadClicksEnabledForAutocapture } from './extensions/dead-clicks-autocapture'
 import { includes, isDistinctIdStringLike } from './utils/string-utils'
+import { getIdentifyHash } from './utils/identify-utils'
 
 /*
 SIMPLE STYLE GUIDE:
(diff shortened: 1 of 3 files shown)


src/posthog-core.ts
```diff
@@ -1431,8 +1432,19 @@ export class PostHog {
...
-            // If the distinct_id is not changing, but we have user properties to set, we can go for a $set event
-            this.setPersonProperties(userPropertiesToSet, userPropertiesToSetOnce)
+            if (
+                this._cachedIdentify !==
+                PostHog._getIdentifyHash(new_distinct_id, userPropertiesToSet, userPropertiesToSetOnce)
+            ) {
+                // If the distinct_id is not changing, but we have user properties to set, we can go for a $set event
+                this.setPersonProperties(userPropertiesToSet, userPropertiesToSetOnce)
+
+                this._cachedIdentify = PostHog._getIdentifyHash(
```
joshsny, line 1442: This could be persisted?

src/posthog-core.ts
```diff
@@ -1444,6 +1462,14 @@ export class PostHog {
         }
     }
 
+    private static _getIdentifyHash(
```
pauldambra, line 1465: i would have this as a function outside of the class since its static

for two reasons...

I think something like `hashOf` that takes arbitrary args is a little nicer to read than `PostHog._getIdentifyHash`

But also (and admittedly it doesn't matter in this PR but for clarity) our minifier handles class fields badly and so we'll have `_getIdentifyHash` in the bundle unminified... whereas a "bare" function will get minified.

pauldambra: probably in the `utils` folder as much as i'm allergic to utils 🤣

src/posthog-core.ts
```diff
@@ -306,7 +307,7 @@ export class PostHog {
         this.analyticsDefaultEndpoint = '/e/'
         this._initialPageviewCaptured = false
         this._initialPersonProfilesConfig = null
-
+        this._cachedIdentify = null
```
pauldambra, line 310: was trying to decide if we should store this in local storage but i think i prefer this isn't cached outside of memory... 

i'm not super familiar with identify and person processing so this feels like a safe step to take to test the waters

src/posthog-core.ts
joshsny: The reason I've scoped this specifically to identify hashes is to ensure that the hash stays stable and that we get type safety when it's invoked. Want to avoid a situation where someone passes the wrong args to hash.

Shall I pull off the class for minification, but keep a specific fn name like `getIdentifyHash`?

src/posthog-core.ts
```diff
@@ -1464,6 +1490,8 @@ export class PostHog {
         // Update current user properties
         this.setPersonPropertiesForFlags(userPropertiesToSet || {})
 
+        // if exactly this $set call has been sent before, don't send it again - determine based on hash of properties
+
```
pauldambra, line 1494: this comment being separate from the logic that decides whether its a duplicate threw me a little, can't decide if that means the logic should be here or we can remove the comment 🙈

src/posthog-core.ts
joshsny: Yeah was also thinking about that, the current implementation helps in SPA setups, but hard page reloads would only be fixed by local storage...

src/posthog-core.ts
```diff
@@ -1430,9 +1431,26 @@ export class PostHog {
...
+            this._cachedIdentify = PostHog._getIdentifyHash(
+                new_distinct_id,
+                userPropertiesToSet,
+                userPropertiesToSetOnce
+            )
         } else if (userPropertiesToSet || userPropertiesToSetOnce) {
-            // If the distinct_id is not changing, but we have user properties to set, we can go for a $set event
-            this.setPersonProperties(userPropertiesToSet, userPropertiesToSetOnce)
+            if (
+                this._cachedIdentify !==
```
pauldambra, line 1442: i think it's ok to do this as an effectively "breaking change"... folk might be depending on the `$set` for duplicate identify calls but that'd be surprising to me

and we could always add an opt-out option if people ask for it 🤔

src/posthog-core.ts
```diff
@@ -1430,9 +1431,26 @@ export class PostHog {
...
+                PostHog._getIdentifyHash(new_distinct_id, userPropertiesToSet, userPropertiesToSetOnce)
+            ) {
+                // If the distinct_id is not changing, but we have user properties to set, we can go for a $set event
```
pauldambra, line 1445: we could `logger.info` in the case we're deduplicating
the logger only writes when in debug mode but can be very useful when troubleshooting

more code, but might be helpful

pauldambra (commented):
brain dumped from a quick scan... i've not really worked with identify so haven't thought about whether we should 🙈

src/posthog-core.ts
pauldambra: works for me...

src/posthog-core.ts
joshsny: Moved and updated the comment to reflect the logic

src/posthog-core.ts
joshsny: Good call, added a log

src/posthog-core.ts
joshsny: Hopefully, it should be okay, since we were already ignoring duplicate identify calls without properties. But in reality, few people make identify calls without properties so it's possible someone does rely on this behaviour...

pauldambra (approved):
was thinking about this... 
i don't think it's breaking (even though it's _technically_ breaking) because this is the behaviour that I think humans expect, so I don't think we need the hit on bundle size of adding config to opt in and slow rollout and all that jazz.

we should label as minor change though (that's what does CI here)

I didn't approve because I don't know _why_ we shouldn't do it - which is too many double negatives - let's update docs on posthog.com and get @joethreepwood to put it in changelog

benjackwhite (approved):
Personally I think this makes total sense. I can't think of a reason why it wouldn't want to be this way, plus we have a warning logged for it so good from me

PR #1652, opened 2025-01-15
feat: Load surveys via RemoteConfig

benjackwhite:
## Changes

We have RemoteConfig loading that includes surveys so time to start using it!

### Follow up
- [ ] Remove draft surveys from the RemoteConfig
- [ ] Do we need to persist surveys? Seems to be a speed thing but the RemoteConfig loads so fast now that it feels weird. Should we instead persist the remote config?

## Checklist

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -1285,13 +1285,13 @@ export class PostHog {
     }
 
     /** Get list of all surveys. */
-    getSurveys(callback: SurveyCallback, forceReload = false): void {
-        this.surveys.getSurveys(callback, forceReload)
+    getSurveys(callback: SurveyCallback): void {
+        this.surveys.getSurveys(callback)
     }
 
     /** Get surveys that should be enabled for the current user. */
-    getActiveMatchingSurveys(callback: SurveyCallback, forceReload = false): void {
-        this.surveys.getActiveMatchingSurveys(callback, forceReload)
+    getActiveMatchingSurveys(callback: SurveyCallback): void {
+        this.surveys.getActiveMatchingSurveys(callback)
     }
 
     /** Render a survey on a specific element. */
--- a/src/posthog-surveys.ts
+++ b/src/posthog-surveys.ts
@@ -62,7 +62,7 @@ export class PostHogSurveys {
     private _surveysEnabledRemotely?: boolean
     private _surveyEventReceiver: SurveyEventReceiver
     private _surveyManager: any
-    private _loadedSurveys: Survey[] = []
+    private _loadedSurveys?: Survey[]
 
     constructor(private readonly instance: PostHog) {
         this._surveyEventReceiver = new SurveyEventReceiver(this.instance)
@@ -111,6 +111,7 @@ export class PostHogSurveys {
                 if (response.statusCode !== 200 || !response.json) {
                     return callback([])
                 }
+
                 const surveys = response.json.surveys || []
 
                 this.onSurveys(surveys)
@@ -148,7 +149,8 @@ export class PostHogSurveys {
             return callback([])
         }
 
-        const existingSurveys: Survey[] = this._loadedSurveys ?? (this.instance.get_property(SURVEYS) as Survey[]) ?? []
+        const existingSurveys: Survey[] | undefined =
+            this._loadedSurveys ?? (this.instance.get_property(SURVEYS) as Survey[]) ?? undefined
 
         if (!existingSurveys) {
             return this.reloadSurveys(callback)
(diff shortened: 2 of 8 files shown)


src/posthog-surveys.ts
```diff
@@ -83,76 +86,84 @@ export class PostHogSurveys {
...
+                this.onSurveys(surveys)
+
+                return callback(surveys)
+            },
+        })
+    }
+
+    private onSurveys(surveys: Survey[]) {
+        this._loadedSurveys = surveys
+        this._surveysEnabledRemotely = surveys.length > 0
```
marandaneto, line 126: I think `_surveysEnabledRemotely` should be `true` if `surveys` isn't undefined, and empty is a valid state (no surveys but the API call has been made), otherwise it'd be semantically wrong, at least the naming.

src/posthog-surveys.ts
```diff
@@ -83,76 +86,84 @@ export class PostHogSurveys {
...
+
+        const eventOrActionBasedSurveys = surveys.filter(
+            (survey: Survey) =>
+                (survey.conditions?.events &&
+                    survey.conditions?.events?.values &&
+                    survey.conditions?.events?.values?.length > 0) ||
```
marandaneto, line 132:
```suggestion
                (survey.conditions?.events?.values?.length > 0) ||
```

not needed the 2 checks since the last one is already using the safe operator `?.`

src/posthog-surveys.ts
```diff
@@ -83,76 +86,84 @@ export class PostHogSurveys {
...
+                (survey.conditions?.actions &&
+                    survey.conditions?.actions?.values &&
+                    survey.conditions?.actions?.values?.length > 0)
```
marandaneto, line 135:
```suggestion
                (survey.conditions?.actions?.values?.length > 0)
```

marandaneto:
> Do we need to persist surveys? Seems to be a speed thing but the RemoteConfig loads so fast now that it feels weird. Should we instead persist the remote config?

Neil created [this ticket](https://github.com/PostHog/posthog/issues/27496) which aims to improve the API perf., so I'd say we can keep it simple and not cache anything for now and improve the API first in case we see this becomes an issue.
WDYT?

src/posthog-core.ts
```diff
@@ -1285,13 +1285,13 @@ export class PostHog {
     }
 
     /** Get list of all surveys. */
-    getSurveys(callback: SurveyCallback, forceReload = false): void {
-        this.surveys.getSurveys(callback, forceReload)
+    getSurveys(callback: SurveyCallback): void {
```
marandaneto, line 1288: Removing `forceReload` from public APIs will break customers that are [using the methods directly](https://posthog.com/docs/surveys/implementing-custom-surveys#fetching-surveys-manually).
Is there a reason to remove the `forceReload`?

It'll contract whats written here:

> Surveys are requested on first load and then cached by default by the JavaScript SDK. If you want to force an API call to get an updated list of surveys, pass true to the forceReload parameter. You only need to do this if you want changed surveys to appear mid-session for users.

src/posthog-surveys.ts
benjackwhite: The thing is I actually think we should just ditch that property altogether. Server side that property is simply "are there surveys to load" but if we now include them in remtoe config I actually think that property is pointless...

src/posthog-surveys.ts
benjackwhite: True. Just moved code but happy to improve it whilst im here

src/posthog-core.ts
benjackwhite: Ewww - I wish we hadn't allowed people to do that tbh. But I can add it back in. We have a bug currently that causes multiple loads anyways

marandaneto:
@benjackwhite is it ready for a final review? I'd address [this](https://github.com/PostHog/posthog-js/pull/1652#discussion_r1916255721) comment at least.

benjackwhite:
> @benjackwhite is it ready for a final review? I'd address [this](https://github.com/PostHog/posthog-js/pull/1652#discussion_r1916255721) comment at least.

It's not really as I haven't had time to properly look at it (and wont have time for another week) hence leaving it in waiting unless someone wants to take it off me 😅

marandaneto:
> > @benjackwhite is it ready for a final review? I'd address [this](https://github.com/PostHog/posthog-js/pull/1652#discussion_r1916255721) comment at least.
> 
> It's not really as I haven't had time to properly look at it (and wont have time for another week) hence leaving it in waiting unless someone wants to take it off me 😅

we can finish this once we roll out the changes on react native first and double-check that the remote config is working well.
i think @lucasheriques would be up for it, will double check today

PR #1654, opened 2025-01-15
feat: Add the ability to mask some PII-containing URL parameters

robbie-c:
## Changes

Add a config option to mask URL parameters. This will also mask any properties that are derived from them.

Why: We want to add a cookieless mode, allowing users to be tracked before consent is given. To do this, we need to not capture any PII-containing properties (such as ad IDs like `gclid`), as consent would be needed for those.

## Checklist

### DIFF
--- a/src/__tests__/request-utils.test.ts
+++ b/src/__tests__/request-utils.test.ts
@@ -106,13 +106,13 @@ describe('request utils', () => {
                 'https://www.example.com/?foo=1#query=2',
             ],
             [
-                "does not add a hash when the isn't one",
+                "does not add a hash when there isn't one",
                 'https://www.example.com/?foo=1',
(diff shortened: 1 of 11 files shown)


src/__tests__/request-utils.test.ts
```diff
@@ -48,6 +48,81 @@ describe('request utils', () => {
         })
     })
 
+    describe('maskQueryParams', () => {
```
daibhin, line 51: Maybe add a test around masking multiple query params?

daibhin (approved):

src/__tests__/request-utils.test.ts
robbie-c: Done!

PR #1658, opened 2025-01-17
fix(): prevent person processing if /decide fails to fetch remote config

havenbarnes:
## Changes

Context:
- https://posthoghelp.zendesk.com/agent/tickets/22908 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/28448) 
- https://posthoghelp.zendesk.com/agent/tickets/22134 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/25748)

In digging through the code it appears that there are two cases where this could potentially happen:
1. An event is emitted in between `_init` setting `this.config = {}` ([here](https://github.com/PostHog/posthog-js/blob/main/src/posthog-core.ts#L406)) and it getting re-set to the proper value (this seems like an extremely unlikely race condition but we may as well protect against it in `_hasPersonProcessing`, and while writing unit tests for a potential fix here I discovered that all events would fail to send if `posthog.config` was an empty object anyway)
2. `/decide` fails to return the config values, leaving the SDK to fall back to `person_profiles: 'always'` ([here](https://github.com/PostHog/posthog-js/blob/main/src/posthog-core.ts#L589-L595))

To protect against the second edge case we make the following change: Remove the dependency on the `defaultIdentifiedOnly` field from `/decide`. This is the new default, so we should not need this conditional moving forward anyway.

## Checklist

### DIFF
--- a/src/__tests__/personProcessing.test.ts
+++ b/src/__tests__/personProcessing.test.ts
@@ -750,31 +750,5 @@ describe('person processing', () => {
             expect(beforeSendMock.mock.calls[0][0].properties.$process_person_profile).toEqual(false)
             expect(beforeSendMock.mock.calls[1][0].properties.$process_person_profile).toEqual(false)
         })
-
-        it('should persist when the default person mode is overridden by decide', async () => {
-            // arrange
-            const persistenceName = uuidv7()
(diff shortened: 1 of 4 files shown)


dmarticus (approved):
nice work on a tricky bug!

PR #1660, opened 2025-01-19
fix: patch how rrweb loads iframe

pauldambra:
patches https://github.com/rrweb-io/rrweb/pull/1633 into our copy of the recorder

moves the iframe from the body to the head
and marks it as hidden
both changes are intended to make this more consistent in safari, since we see it fail particularly in angular apps running in safari


pauldambra:
it's definitely not this

PR #1661, opened 2025-01-20
fix: expose getNextSurveyStep to use in posthog

lucasheriques:
## Changes

No functional changes.

Exposes `getNextSurveyStep` to the exported functions so we don't need to duplicate that logic on `posthog/posthog` for survey previewing purposes

## Checklist
- [x] Tests for new code (no code changes, just exposing a function)


ioannisj (approved):
LG, some sorted imports so pre-commit hooks may not running for you?

lucasheriques:
@ioannisj they always run 🤔 so not sure why did the imports change order

PR #1663, opened 2025-01-21
chore: better logging for surveys loaders

marandaneto:
## Changes

It won't fix but It'll help figuring out https://posthoghelp.zendesk.com/agent/tickets/21804 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/25745)

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -373,7 +373,10 @@ export function usePopupVisibility(
     const [isSurveySent, setIsSurveySent] = useState(false)
 
     useEffect(() => {
-        if (isPreviewMode || !posthog) {
+        if (!posthog) {
+            logger.error('usePopupVisibility hook called without a PostHog instance.')
+        }
+        if (isPreviewMode) {
             return
         }
 
@@ -555,6 +558,7 @@ export function Questions({
         displayQuestionIndex: number
     }) => {
         if (!posthog) {
+            logger.error('onNextButtonClick called without a PostHog instance.')
             return
         }
 
@@ -655,7 +659,10 @@ export function FeedbackWidget({
     const widgetRef = useRef<HTMLDivElement>(null)
 
     useEffect(() => {
-        if (readOnly || !posthog) {
+        if (!posthog) {
+            logger.error('FeedbackWidget called without a PostHog instance.')
+        }
+        if (readOnly) {
             return
         }
 
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -375,6 +375,7 @@ export function usePopupVisibility(
     useEffect(() => {
         if (!posthog) {
             logger.error('usePopupVisibility hook called without a PostHog instance.')
+            return
         }
         if (isPreviewMode) {
             return
@@ -661,6 +662,7 @@ export function FeedbackWidget({
     useEffect(() => {
         if (!posthog) {
             logger.error('FeedbackWidget called without a PostHog instance.')
+            return
         }
         if (readOnly) {
             return
--- a/src/extensions/surveys/surveys-utils.tsx
+++ b/src/extensions/surveys/surveys-utils.tsx
@@ -2,10 +2,14 @@ import { VNode, cloneElement, createContext } from 'preact'
 import { PostHog } from '../../posthog-core'
 import { MultipleSurveyQuestion, Survey, SurveyAppearance, SurveyQuestion } from '../../posthog-surveys-types'
 import { document as _document, window as _window } from '../../utils/globals'
+import { createLogger } from '../../utils/logger'
 // We cast the types here which is dangerous but protected by the top level generateSurveys call
 const window = _window as Window & typeof globalThis
 const document = _document as Document
 const SurveySeenPrefix = 'seenSurvey_'
+
+const logger = createLogger('[Surveys]')
+
 export const style = (appearance: SurveyAppearance | null) => {
     const positions = {
         left: 'left: 30px;',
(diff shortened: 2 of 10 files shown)


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -534,7 +538,11 @@ export const sendSurveyEvent = (
     survey: Survey,
     posthog?: PostHog
 ) => {
-    if (!posthog) return
+    // TODO: does it need readOnly like dismissedSurveyEvent?
```
marandaneto, line 541: @lucasheriques since you are working on the preview bits, do we need the readOnly here too? is the preview sending dismissed events?

src/posthog-core.ts
```diff
@@ -598,7 +598,7 @@ export class PostHog {
         this.sessionRecording?.onRemoteConfig(config)
         this.autocapture?.onRemoteConfig(config)
         this.heatmaps?.onRemoteConfig(config)
-        this.surveys?.onRemoteConfig(config)
+        this.surveys.onRemoteConfig(config)
```
marandaneto, line 601: surveys is created on the constructor, never undefined|null

src/extensions/surveys/surveys-utils.tsx
lucasheriques: no the preview doesn't send and kind of events

marandaneto: @lucasheriques so should we remove the checks from the other methods?

marandaneto:
@pauldambra just because you're the bundler master and I want to avoid something like https://posthog.slack.com/archives/C07QD3LT8U9/p1737472825108809

ioannisj (approved):
Just a minor comment, but looks good with my limited know-how :)

src/extensions/surveys.tsx
```diff
@@ -373,7 +373,10 @@ export function usePopupVisibility(
     const [isSurveySent, setIsSurveySent] = useState(false)
 
     useEffect(() => {
-        if (isPreviewMode || !posthog) {
+        if (!posthog) {
+            logger.error('usePopupVisibility hook called without a PostHog instance.')
```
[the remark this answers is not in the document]
marandaneto: yes, good catch, copy-pasta error

src/posthog-surveys.ts
```diff
@@ -154,27 +156,55 @@ export class PostHogSurveys {
...
+                    if (err) {
+                        return logger.error('Could not load surveys script', err)
+                    }
 
-                this._surveyManager = assignableWindow.__PosthogExtensions__?.generateSurveys?.(this.instance)
-            })
+                    this._surveyManager = phExtensions.generateSurveys?.(this.instance)
+                })
+            } else {
+                logger.error('PostHog loadExternalDependency extension not found. Cannot load remote config.')
```
pauldambra, line 195: you can bundle such that loadExternalDependency is not available but surveys is

e.g. `array-full.no-external.ts` 

i'm not completely sure how that would look to the code (it's a while since I checked)

pauldambra:
you can always check https://github.com/PostHog/posthog-js/pull/1663#issuecomment-2605067543 once it's run (and it will update each time you push) to track the impact :)

src/posthog-surveys.ts
pauldambra: so for example here https://github.com/PostHog/posthog-js/blob/912e97dcbfb71dbbd20711a37f1ddc30973de712/src/extensions/replay/sessionrecording.ts#L760-L770

* we check for presence of rrweb
* if it is present then we don't have to load anything
* and we call start
* otherwise we try to load it
* and call start if that succeeds

(i bet we're not consistent about being that careful 🙈)

marandaneto: Does that mean the current code already has a bug then? All I did was add conditions so I could do logging for each of the objects that are in the not-desired state.
I can fix it if that's the case.

marandaneto: Done!

src/posthog-surveys.ts
```diff
@@ -154,27 +156,57 @@ export class PostHogSurveys {
...
+                        return
+                    }
+
+                    this._surveyManager = phExtensions.generateSurveys?.(this.instance)
+                })
+            } else {
+                logger.error('PostHog loadExternalDependency extension not found. Cannot load remote config.')
+            }
+        } else {
+            this._surveyManager = generateSurveys(this.instance)
```
pauldambra, line 200: do we know `loadIfEnabled` is only called once?

we can maybe short circuit too if `this._surveyManager` exists but I haven't thought through what all this is doing :)

marandaneto: Good point, didn't think of that since the goal was just more logging, but the surveys bit has room for lots of improvements.

pauldambra (approved):
didn't run it, but looks safe 👍

src/posthog-surveys.ts
marandaneto: can be called by `onRemoteConfig`, `set_config` and `init`.
`set_config` can be called in many cases, so yeah it makes sense to short circuit.

marandaneto: done, thanks.

lucasheriques (approved):
looks good to me

daibhin:
Not sure if it's intended but I'm seeing a ton of repeated logs added in this PR. Looks like it's firing once a second
Screenshot 2025-01-23 at 10 22 47

marandaneto:
> Not sure if it's intended but I'm seeing a ton of repeated logs added in this PR. Looks like it's firing once a second Screenshot 2025-01-23 at 10 22 47

good point, here it is https://github.com/PostHog/posthog-js/pull/1679

PR #1670, opened 2025-01-22
fix: patch for css parsing performance

pauldambra:
patches in this PR https://github.com/rrweb-io/rrweb/pull/1615


veryayskiy (approved):
Patch it all!

Issue #769, opened 2023-08-04
Uninitialized PostHog client scenario when initializing PostHogProvider with client


morsela:
In a React (NextJS) app we have a setup with Segment & Posthog looking like this
```
useEffect(() => {
    const analytics = AnalyticsBrowser.load({ writeKey });
    analytics.ready(() => {
      posthog.init(process.env.NEXT_PUBLIC_POSTHOG_CLIENT_ID!, {
        api_host: "https://app.posthog.com",
        segment: window.analytics,
        opt_in_site_apps: true,
        loaded: (posthog) => {
          setPosthogInstance(posthog);
        },
      });
    });

    setAnalytics(analytics);
  }, [writeKey]);
```

the initialized instance is then used for the initialization of the PostHogProvider context
i.e 
```
<PostHogProvider client={posthogInstance}>
```

We started seeing issues in production (TypeError: Cannot read properties of undefined (reading 'props')) from a group call down the line.

I seems like https://github.com/PostHog/posthog-js/commit/ab59da77ce8f82e29363ab967a5cc99c26057d4c introduced a non-optional posthog client, but didn't account for handling the case where the client is given as the PostHogProvider param is still uninitialized, leading to [returning an uninitialized `posthogJs`](https://github.com/PostHog/posthog-js/blob/ab59da77ce8f82e29363ab967a5cc99c26057d4c/react/src/context/PostHogProvider.tsx#L41)

FatehAK:
I'm facing the same issue when firing the `posthog.group` call

rafaeelaudibert:
I'm not positive about what could be done here to avoid this. If you're passing a `client` to our `Provider`, we assume that you control the client and, therefore, should control its initialization.

We can't throw an error when it's passed uninitialized because the recommended approach is to initialize it using a `useEffect` which will only run during the next rerender cycle. Similarly, we can't init it ourselves because we don't have your token.

I'm merging https://github.com/PostHog/posthog-js/pull/1675 soon, which will help with some situations, but arguably doesn't solve this.

I'll close this assuming that this is intended behavior.

PR #1675, opened 2025-01-22
docs: Improve documentation/logs around PostHogProvider

rafaeelaudibert:
We're now logging more reasonable logs around our PostHogProvider to help customers identify problems with their setup. We're also typing that component much better to prevent people from getting it wrong at the type level.

Inspired by #769, but not a solution


react/src/context/PostHogProvider.tsx
```diff
@@ -4,41 +4,63 @@ import posthogJs, { PostHogConfig } from 'posthog-js'
...
+/**
+ * Props for the PostHogProvider component.
+ * This is a discriminated union type that ensures mutually exclusive props:
+ *
+ * - If `client` is provided, `apiKey` and `options` must not be provided
+ * - If `apiKey` is provided, `client` must not be provided, and `options` is optional
+ */
+type PostHogProviderProps =
+    | { client: PostHog; apiKey?: undefined; options?: undefined }
+    | { apiKey: string; options?: Partial<PostHogConfig>; client?: undefined }
```
rafaeelaudibert, line 18: This is somewhat of a breaking change at the type level. I know we've had other similar type-level breaking changes in the past. This is extremely easy to fix, all onboard with a minor update?

pauldambra: shouldn't they be `never` if they aren't supposed to be provided?

will folk have to change _working_ _correct_ code with this change?

rafaeelaudibert: > shouldn't they be never if they aren't supposed to be provided?

Done, force-pushed the change

pauldambra: i really wish this was bundled as `@posthog/react` and we could version it separately
i guess this never hits people through the html snippet?
so you have to run `my-package-manager update` to get the change
but then we'd be making you change code without a major upgrade... which is yucky

if we were going to be good citizens we'd add a new provider and deprecate the old one... is there a not yuck way of doing that?

rafaeelaudibert: > i really wish this was bundled as @posthog/react and we could version it separately

Oh, absolutely. I might do that, and then we would keep this component as is, and add a big deprecation notice telling people to move to `@posthog/react` instead. Thoughts?

> i guess this never hits people through the html snippet?

They're not using React if they're using the HTML snippet, so yeah, we can do slightly-more-dangerous stuff here if we want to. That's why I wouldn't be opposed to simply merging this and moving to the performance issue I'm delaying :)

> but then we'd be making you change code without a major upgrade... which is yucky

Is it though? IMO, it's reasonable to expect someone would have to change the code if we had a bad bug, and I wouldn't argue for a major release. This isn't a bad bug, of course, but if anyone was using the Library in that way, it's kind of a bug.

> if we were going to be good citizens we'd add a new provider and deprecate the old one... is there a not yuck way of doing that?

yuck all around, I'd be against this, I'd argue it's more painful for customers to understand why the heck do we have 2 providers vs. having to remove 2 props on a minor bump.

tiagoskaneta:
Quick note here that this was definitely a breaking change. As the typescript definition changed we caught this during build but we never saw the warning. The fix is trivial but at least a note in the release note would have been appreciated.

pauldambra:
hey @tiagoskaneta thanks for the feedback 💖 we want to get to a place where we can version posthog/react and posthog/web separately... the changelog is auto-generated, but we could have edited it.

out of interest since the setup before allowed technically "incorrect" setup... how did this manifest for you? is it better now?

tiagoskaneta:
> out of interest since the setup before allowed technically "incorrect" setup... how did this manifest for you? is it better now?

We were using the `<PostHogProvider>` without any argument, which I understand now to be an "incorrect" setup. After the change, during typescript checking (triggered by a dependabot bump) we caught the breaking change. As no significant change was flagged in the release notes, I checked the repo and got to this PR.
In our case, we were already instantiating and initializing posthog, so the fix was just to add `client={posthog}`, so simple enough and no harm done 😃

PR #1676, opened 2025-01-22
feat: Enhance PostHogProvider to prevent double initialization on `React.StrictMode`

### LINKED ISSUE #723: "posthog was already loaded elsewhere"
I'm having an issue with posthog logging this error. we are using gatsby but not the gatsby plugin, current posthog version is 1.69.0. we're initialising in `gatsby-browser.tsx` with:

```js
export const wrapRootElement: GatsbyBrowser['wrapRootElement'] = ({ element }) => (
  <PostHogProvider
    apiKey={apiKey}
    options={posthogOptions}
  >
    {element}
  </PostHogProvider>
);
```

The simply importing `usePosthog` and `useFeatureFlagEnabled` in a few places. I can't currently see how it would be loaded twice?

rafaeelaudibert:
Added state management to track if the PostHog client has already been initialized, addressing issues that arise when running under React.StrictMode. This change includes warnings for users if the client is loaded multiple times, improving the overall robustness of the PostHogProvider component.

Closes #723


robbie-c (commented):
I guess the intention was that useMemo would protect against this, but in strict mode useMemo calls the calculation twice.

Using useState to store the state makes sense to me, but is a boolean flag enough state? What if the user changes the api key or the client?

Do we want to start handling changes in options correctly, and update the posthog-js config?

rafaeelaudibert:
> I guess the intention was that useMemo would protect against this, but in strict mode useMemo calls the calculation twice.

I think `useMemo` was there to allow people to send a new `client` or different options, but we don't handle that, so, maybe?

Also, `useMemo` only calls it twice on React 16+, and this code predates that

> Using useState to store the state makes sense to me, but is a boolean flag enough state? What if the user changes the api key or the client?
> 
> Do we want to start handling changes in options correctly, and update the posthog-js config? Sorry to offer you some scope creep :D

Those are great points. I do believe people haven't actively complained about it yet, so I'm happy to merge this as is, stepping on some toes, and then coming back with a fix if it's needed.

rafaeelaudibert:
Quoting myself, yes

> Those are great points. I do believe people haven't actively complained about it yet, so I'm happy to merge this as is, stepping on some toes, and then coming back with a fix if it's needed.

I'm not introducing a new bug, but I might be making it worse, I'll try and fix this.

rafaeelaudibert:
@robbie-c Mind checking [bed8367](https://github.com/PostHog/posthog-js/pull/1676/commits/bed83675e6fe0b4c6f110ec548209e23e11dc80d) out? Should be a better solution

react/src/context/PostHogProvider.tsx
```diff
@@ -50,11 +64,56 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
...
 
         if (apiKey) {
+            // If the client was already initialized, we'll attempt to return the existing instance
+            // This is likely to occur when someone is developing locally under `React.StrictMode`
+            //
+            // There's no built-in way to detect React.StrictMode, so we set a flag to check if the client was already initialized
+            // and return the existing instance if it was.
+            //
+            // This might also be triggered if someone attempts to change their `apiKey` or
+            // set a different set of `options`. This is handled separately below.
```
robbie-c, line 74: This approach makes sense to me! I do think adding some tests would be prudent (is the same client returning the second time calling with the same args, does set_config get called correctly, etc).

Otherwise LGTM!

robbie-c (approved):

react/src/context/PostHogProvider.tsx
rafaeelaudibert: Added a couple tests to assert behavior - and found a slight bug :)

PR #1677, opened 2025-01-22
fix: cache nested loops

pauldambra:
we're clearly hitting a pathological case in the CSS processing for some folks
my guess is something is repeating the check due to run away mutations or something in a tight loop... or maybe it is just bad at very large strings 🤷 

quoting my old colleague Chris Sewart about peformance

you can do it faster
do it less
or not do it

let's cache our results as we go so that we only run the nested check or the regex replace once per input

you can see the test here https://github.com/PostHog/posthog-rrweb/pull/17

we see results like

```
First execution time: 3.3014169996604323ms
Average execution time for the second execution: 1.8079558699950575ms
```

that second execution will be O(1)
while first execution is O(n)

so where O(n) is large this should be a much bigger improvement


veryayskiy (approved):
//

PR #1680, opened 2025-01-23
fix: Exclude more domains from cross subdomain tracking by default

rafaeelaudibert:
## Changes

We created this when Heroku was the only big player with cross-subdomain problems. We have Vercel and Netlify as big-ish players now, so let's check them as well.


src/utils/index.ts
```diff
@@ -294,6 +294,8 @@ export const registerEvent = (function () {
     return register_event
 })()
 
+// NOTE: Update PostHogConfig docs if you change this list
```
rafaeelaudibert, line 297: Hint of what my next PR is going to be

src/utils/index.ts
```diff
@@ -294,6 +294,8 @@ export const registerEvent = (function () {
...
+const EXCLUDED_FROM_CROSS_SUBDOMAIN_COOKIE = ['herokuapp.com', 'vercel.app', 'netlify.app']
```
robbie-c, line 298: Let's not go nuts here, there's tons of sites we *could* include, but really the correct answer is people setting the config option as to whether the cookie is cross-domain.

It might be worth adding a comment here.

rafaeelaudibert: I agree we shouldn't go nuts, but we do init our configuration to this function result, so this is the default behavior. I do think we should make a small amount of effort to avoid people from getting bit.

I can add a comment here explaining that we will not try to catch all bullets.

robbie-c (approved):

PR #1681, opened 2025-01-23
docs: Add documentation to all of our init options

rafaeelaudibert:
We usually try to document our settings on our website, but not all of them are present there. This means that we have some configs that are completely undocumented which makes it hard for our customers to use them.

Let's improve that by adding JSDocs to all of our existing settings. This will surely cause the next developers to also include documentation if they ever add a new config. I hope so, at least.

I've deprecated `verbose` in favor of `debug`. `verbose` wasn't been used before, but some people might have set it, so I'm renaming it to `debug` if present.

I've also removed any of the existing documentation to previously deprecated properties in an effort to force people to use the new methods.

I noticed that the docs for `person_profiles` were asserting that `always` was the default, which isn't for a couple of months now, so let's update that documentation as well.

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -122,6 +122,9 @@ const PRIMARY_INSTANCE_NAME = 'posthog'
 // should only be true for Opera<12
 let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 && userAgent?.indexOf('Mozilla') === -1
 
+// NOTE: Remember to update `types.ts` when changing a default value
+// to guarantee documentation is up to date, make sure to also update our website docs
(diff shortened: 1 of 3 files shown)


src/posthog-core.ts
rafaeelaudibert, line 1: Because this PR involves:
- deprecating some columns
- removing preview config
- organizing types

it might make sense to split it into three? Open to thoughts

robbie-c: I'd prefer that - or even 2, based on whether you changed the public interface (such as deprecating fields and adding new ones) or didn't (fixing comments, moving things between files)

rafaeelaudibert: On it

rafaeelaudibert: Ok, #1694 is merged, this can be reviewed now, this is only docs changes :)

dmarticus (approved):

PR #1682, opened 2025-01-23
fix: correctly cache split css

pauldambra:
oops, we mutate the value that I'm using as a cache key, so we never hit the cache 🫠

### DIFF
--- a/patches/@rrweb__record@2.0.0-alpha.18.patch
+++ b/patches/@rrweb__record@2.0.0-alpha.18.patch
@@ -1,5 +1,5 @@
 diff --git a/dist/record.js b/dist/record.js
-index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..62e51b1a9df6df14115b31e21a7360b25591e249 100644
+index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..1932f81f2e330cc904e8349a74911e6b3c12669a 100644
 --- a/dist/record.js
 +++ b/dist/record.js
 @@ -68,10 +68,10 @@ function getUntaintedPrototype$1(key) {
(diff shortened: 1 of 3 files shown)


patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -68,6 +68,8 @@ index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..62e51b1a9df6df14115b31e21a7360b2
 +      // we know there's a result
 +      return splitCache.get(cssText)
 +    }
++    // cssText is mutated so we keep the original for when we cache the result
++    const og = cssText
```
pauldambra, line 72: these patches are horrible to read

I changed this to store the string

(i was going to hash it too, but i'd rather fix the actual performance problem)

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -136,17 +138,17 @@ index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..62e51b1a9df6df14115b31e21a7360b2
...
              break;
-@@ -451,7 +524,8 @@ function splitCssText(cssText, style) {
+@@ -451,7 +526,8 @@ function splitCssText(cssText, style) {
        }
      }
    }
 -  splits.push(cssText);
 +  splits.push(cssText); // either the full thing if no splits were found, or the last split
-+  splitCache.set(cssText, splits);
++  splitCache.set(og, splits);
```
pauldambra, line 147: and then use the `og` value as cache key here

patches/@rrweb__record@2.0.0-alpha.18.patch
rafaeelaudibert: I can't understand it, but I'll go for a ✅

rafaeelaudibert (approved):

PR #1686, opened 2025-01-24
fix: Implement `isDeepEqual` manually

### LINKED ISSUE #1685: v1.209.0 causing build failures with ES Build
I have started getting build failure with newer version.

```
#11 33.54 node_modules/posthog-js/react/dist/esm/index.js (3:9): "isDeepStrictEqual" is not exported by "__vite-browser-external", imported by "node_modules/posthog-js/react/dist/esm/index.js".
33.54 file: /app/node_modules/posthog-js/react/dist/esm/index.js:3:9
1: import posthogJs from 'posthog-js';
2025-01-23 15:54:25.756 PST
2: import React, { createContext, useState, useMemo, useContext, useEffect, useRef, useCallback, Children } from 'react';
2025-01-23 15:54:25.756 PST
3: import { isDeepStrictEqual } from 'util';
```

rafaeelaudibert:
We were using a built-in Node function, but that was silly because Node functions aren't available in the browser. Most bundlers could probably work that out, but Vite doesn't, so let's implement that ourselves and avoid requiring a bundler to sort this out.

Closes #1685

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -117,6 +117,12 @@ module.exports = {
                 'compat/compat': 'off',
             },
         },
+        {
+            files: ['react/**/*.ts'],
+            rules: {
+                'posthog-js/no-direct-null-check': 'off',
+            },
+        },
     ],
     root: true,
 }
--- a/react/src/context/PostHogProvider.tsx
+++ b/react/src/context/PostHogProvider.tsx
@@ -3,7 +3,7 @@ import posthogJs, { PostHogConfig } from 'posthog-js'
 
 import React, { useMemo, useState } from 'react'
 import { PostHog, PostHogContext } from './PostHogContext'
-import { isDeepStrictEqual } from 'util'
+import { isDeepEqual } from '../utils/object-utils'
 
 type AlreadyInitialized =
     | {
@@ -34,6 +34,11 @@ type PostHogProviderProps =
  *
  * These initialization methods are mutually exclusive - you must use one or the other,
  * but not both simultaneously.
+ *
+ * We strongly suggest you memoize the `options` object to ensure that you don't
+ * accidentally trigger unnecessary re-renders. We'll properly detect if the options
+ * have changed and only call `posthogJs.set_config` if they have, but it's better to
+ * avoid unnecessary re-renders in the first place.
  */
 export function PostHogProvider({ children, client, apiKey, options }: WithOptionalChildren<PostHogProviderProps>) {
     // Used to detect if the client was already initialized
@@ -84,7 +89,7 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
 
                 // Changing options is better supported because we can just call `posthogJs.set_config(options)`
                 // and they'll be good to go with their new config. The SDK will know how to handle the changes.
-                if (options && !isDeepStrictEqual(options, alreadyInitialized.previousOptions)) {
+                if (options && !isDeepEqual(options, alreadyInitialized.previousOptions)) {
                     posthogJs.set_config(options)
                 }
 
--- a/react/src/context/__tests__/PostHogProvider.test.jsx
+++ b/react/src/context/__tests__/PostHogProvider.test.jsx
@@ -56,6 +56,30 @@ describe('PostHogProvider component', () => {
             expect(posthogJs.set_config).toHaveBeenCalledWith(updatedOptions)
         })
 
+        it('should NOT call set_config when we pass new options that are the same as the previous options', () => {
+            const { rerender } = render(
+                <PostHogProvider apiKey={apiKey} options={initialOptions}>
+                    <div>Test</div>
+                </PostHogProvider>
+            )
+
+            // First render should initialize
+            expect(posthogJs.init).toHaveBeenCalledWith(apiKey, initialOptions)
+
+            // Rerender with new options
+            const sameOptionsButDifferentReference = { ...initialOptions }
+            act(() => {
+                rerender(
+                    <PostHogProvider apiKey={apiKey} options={sameOptionsButDifferentReference}>
+                        <div>Test</div>
+                    </PostHogProvider>
+                )
+            })
+
+            // Should NOT call set_config
+            expect(posthogJs.set_config).not.toHaveBeenCalled()
+        })
+
         it('should warn when attempting to change apiKey', () => {
             const consoleSpy = jest.spyOn(console, 'warn').mockImplementation()
             const newApiKey = 'different-api-key'
--- a/react/src/utils/object-utils.ts
+++ b/react/src/utils/object-utils.ts
@@ -1,0 +1,19 @@
+export function isDeepEqual(obj1: any, obj2: any): boolean {
+    if (obj1 === obj2) return true
+
+    if (typeof obj1 !== 'object' || obj1 === null || typeof obj2 !== 'object' || obj2 === null) {
+        return false
+    }
+
+    const keys1 = Object.keys(obj1)
+    const keys2 = Object.keys(obj2)
+
+    if (keys1.length !== keys2.length) return false
+
+    for (const key of keys1) {
+        if (!keys2.includes(key)) return false
+        if (!isDeepEqual(obj1[key], obj2[key])) return false
+    }
+
+    return true
+}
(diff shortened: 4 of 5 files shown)


.eslintrc.js
```diff
@@ -117,6 +117,12 @@ module.exports = {
                 'compat/compat': 'off',
             },
         },
+        {
+            files: ['react/**/*.ts'],
+            rules: {
+                'posthog-js/no-direct-null-check': 'off',
```
pauldambra, line 123: why turn this off instead of using `isNull`?

react/src/context/PostHogProvider.tsx
```diff
@@ -34,6 +34,11 @@ type PostHogProviderProps =
  *
  * These initialization methods are mutually exclusive - you must use one or the other,
  * but not both simultaneously.
+ *
+ * We strongly suggest you memoize the `options` object to ensure that you don't
+ * accidentally trigger unnecessary re-renders. We'll properly detect if the options
+ * have changed and only call `posthogJs.set_config` if they have, but it's better to
+ * avoid unnecessary re-renders in the first place.
```
pauldambra, line 41: can't we memoise the provider and provide the equality check ourselves so consumers don't have to?

react/src/context/__tests__/PostHogProvider.test.jsx
```diff
@@ -56,6 +56,30 @@ describe('PostHogProvider component', () => {
             expect(posthogJs.set_config).toHaveBeenCalledWith(updatedOptions)
         })
 
+        it('should NOT call set_config when we pass new options that are the same as the previous options', () => {
```
pauldambra, line 59: nice!

react/src/utils/object-utils.ts
```diff
@@ -0,0 +1,19 @@
+export function isDeepEqual(obj1: any, obj2: any): boolean {
```
robbie-c, line 1: What's the provenance of this? Is it directly copied from the node library? If you wrote or modified it, it needs tests.

react/src/utils/object-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+    if (obj1 === obj2) return true
+
+    if (typeof obj1 !== 'object' || obj1 === null || typeof obj2 !== 'object' || obj2 === null) {
+        return false
+    }
+
+    const keys1 = Object.keys(obj1)
+    const keys2 = Object.keys(obj2)
+
+    if (keys1.length !== keys2.length) return false
```
robbie-c, line 11:
```suggestion
    if (keys1.length !== keys2.length) {
      return false
    }
```
We generally prefer this style in the codebase. (I'm surprised prettier didn't change this!)

react/src/utils/object-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+
+    for (const key of keys1) {
+        if (!keys2.includes(key)) return false
+        if (!isDeepEqual(obj1[key], obj2[key])) return false
```
robbie-c, line 15:
```suggestion
        if (!keys2.includes(key) || !isDeepEqual(obj1[key], obj2[key])) {
          return false
        }
```

react/src/context/__tests__/PostHogProvider.test.jsx
robbie-c: +1

.eslintrc.js
robbie-c: Side note, but what's the background on us using `isNull` etc rather than `===`? Browser compatibility? Bytes?

react/src/utils/object-utils.ts
pauldambra, line 1: JS engines have surprising optimisations these days so I wondered if just comparing json stringify would be faster

turns out absolutely no 🤣 this is about 10x the speed of `json.stringify(a) == json.stringify(b)`

https://jsbench.me/dpm6ahv74o/1

.eslintrc.js
pauldambra: the bundle reduced by 1% when we switched to `isNull` and `isUndefined` compared to the straight comparison

cos `fe(x, y)` is fewer characters than `x === y` and we use it a lot

ofc likely that the react stuff is being bundled and run through a minifier by someone else but consistency is probably better than varying

pauldambra: we use the checks a lot so the minified code has (something like) `fe(x)` at each call site 

whereas the direct check is (something like) `x === void 0` when minified 

so once you call `isUndefined` more times than the character count of adding the declaration to the bundle you  add roughly half as many characters by moving to a function

it's the same reason we have some global things as variables in the `utils` folder because then they'll minify and we have `xY` in the bundle not `navigator`

react/src/utils/object-utils.ts
```diff
@@ -0,0 +1,19 @@
...
+        return false
+    }
+
+    const keys1 = Object.keys(obj1)
+    const keys2 = Object.keys(obj2)
+
+    if (keys1.length !== keys2.length) return false
+
+    for (const key of keys1) {
+        if (!keys2.includes(key)) return false
```
joshsny, line 14: Isn't this making the for loop O(n^2) for n keys? We probably want to consider using a set or something

react/src/utils/object-utils.ts
joshsny: I've attempted to write this function a few times and always been wrong, so would be interested in the answer to this!

joshsny (commented):
Looks good, we need to be very careful with that deep equal function

react/src/utils/object-utils.ts
joshsny: Maybe we don't care because we don't have many keys - but a little worried we might use it elsewhere (e.g for properties) where it might get inefficient

.eslintrc.js
rafaeelaudibert: Hmm, in my mind I did that to avoid crossing the boundary from `posthog-js/react` to `posthog-js`. I now see that doesn't make sense because this is a single package for those who're installing it, so I'm happy to change :)

pauldambra: Ah.. fair.. yep, something to bear in my mind if/when we start splitting this up

react/src/utils/object-utils.ts
pauldambra: we've had to handle recursive json in other places in the SDK too... which is why i wondered about json stringify comparison performance

abhyuditjain:
How about this:
```
export function isDeepEqual(obj1: any, obj2: any, visited = new WeakMap()): boolean {
    if (obj1 === obj2) return true;

    if (obj1 === null || typeof obj1 !== 'object' || obj2 === null || typeof obj2 !== 'object') {
        return false;
    }

    // check visited to avoid infinite recursion
    if (visited.has(obj1) && visited.get(obj1) === obj2) return true;
    visited.set(obj1, obj2);

    const keys1 = Object.keys(obj1);
    const keys2 = Object.keys(obj2);

    if (keys1.length !== keys2.length) return false;

    // set for O(1) lookups
    const keys2Set = new Set(keys2);

    for (const key of keys1) {
        if (!keys2Set.has(key) || !isDeepEqual(obj1[key], obj2[key], visited)) {
            return false;
        }
    }

    return true;
}
```

Avoids `keys2.includes()` for each key and has a visited set for circular references.

.eslintrc.js
pauldambra: > So if someone was experimenting with a new minifier, this is something they could change to see if it made a difference.

that'd be a very good minifier... it's more of a compiler at that point 🤯

pauldambra: there's two sources of size in the repo

* unnecessary repetition of primitives - like this example
* heavy use of classes and no minification of class fields/properties

react/src/utils/object-utils.ts
rafaeelaudibert: True. We're discussing further in https://posthog.slack.com/archives/C03P7NL6RMW/p1737730568567229.

For our current use case, where we're only comparing `PostHogConfig` objects this isn't a problem because I don't expect people to have more than ~10 entries, and the extra work of creating a set/heap is more than just (10^2)

We might optimize this in the future for "the general usecase", but I'm happy to keep as is for now

react/src/context/PostHogProvider.tsx
pauldambra: yep, yep, if I don't hit the big "request changes" button then it's always just riffing :)

robbie-c (approved):

react/src/utils/object-utils.ts
joshsny: yeah that makes complete sense!

rafaeelaudibert:
We'll have a follow-up PR moving this new helper function to the `PosthogProvider` file. It's good enough for our use-case - nobody should have huge/circular configs, so this is fine, but this function can't be used whenever we need a general `deepCompare`.

PR #1687, opened 2025-01-24
fix: angular detection can be less blunt

pauldambra:
from investigation of how rrweb has done this over time and how safari behaves with different versions manually patched

i learned that Angular Zone advertises its unpatched versions in a known/discoverable location

let's just use them instead of going to an iframe when they're present

### DIFF
--- a/patches/@rrweb__record@2.0.0-alpha.18.patch
+++ b/patches/@rrweb__record@2.0.0-alpha.18.patch
@@ -1,8 +1,55 @@
 diff --git a/dist/record.js b/dist/record.js
-index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..1932f81f2e330cc904e8349a74911e6b3c12669a 100644
+index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..5d583e81bb7b7e49b7a091ca0c217a37116861b1 100644
 --- a/dist/record.js
 +++ b/dist/record.js
-@@ -68,10 +68,10 @@ function getUntaintedPrototype$1(key) {
+@@ -26,14 +26,33 @@ const testableMethods$1 = {
+   Element: [],
+   MutationObserver: ["constructor"]
+ };
++/*
++Angular zone patches many things and can pass the untainted checks below, causing performance issues
++Angular zone, puts the unpatched originals on the window, and the names for hose on the zone object.
++So, we get the unpatched versions from the window object if they exist.
++You can rename Zone, but this is a good enough proxy to avoid going to an iframe to get the untainted versions.
++see: https://github.com/angular/angular/issues/26948
++*/
++function angularZoneUnpatchedAlternative$1(key) {
++  const angularUnpatchedVersionSymbol = (
++    globalThis
++  )?.Zone?.__symbol__?.(key);
++  if (
++    angularUnpatchedVersionSymbol &&
++    (globalThis)[angularUnpatchedVersionSymbol]
++  ) {
++    return (globalThis)[
++      angularUnpatchedVersionSymbol
++    ];
++  } else {
++    return undefined;
++  }
++}
+ const untaintedBasePrototype$1 = {};
+-const isAngularZonePresent$1 = () => {
+-  return !!globalThis.Zone;
+-};
+ function getUntaintedPrototype$1(key) {
+   if (untaintedBasePrototype$1[key])
+     return untaintedBasePrototype$1[key];
+-  const defaultObj = globalThis[key];
++  const defaultObj = angularZoneUnpatchedAlternative$1(key) || globalThis[key];
+   const defaultPrototype = defaultObj.prototype;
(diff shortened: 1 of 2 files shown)


pauldambra:
https://github.com/PostHog/posthog-rrweb/pull/24

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -1,8 +1,55 @@
...
++/*
++Angular zone patches many things and can pass the untainted checks below, causing performance issues
++Angular zone, puts the unpatched originals on the window, and the names for hose on the zone object.
++So, we get the unpatched versions from the window object if they exist.
++You can rename Zone, but this is a good enough proxy to avoid going to an iframe to get the untainted versions.
++see: https://github.com/angular/angular/issues/26948
++*/
++function angularZoneUnpatchedAlternative$1(key) {
++  const angularUnpatchedVersionSymbol = (
++    globalThis
```
pauldambra, line 18: angular puts a function on Zone you can call with the name of the thing
and you get back undefined or where it has stored the thing

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -1,8 +1,55 @@
...
++  )?.Zone?.__symbol__?.(key);
++  if (
++    angularUnpatchedVersionSymbol &&
++    (globalThis)[angularUnpatchedVersionSymbol]
++  ) {
++    return (globalThis)[
```
pauldambra, line 24: so, if there's an unpatched version we can use that

(confirmed this returns expected values on customer site)

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -1,8 +1,55 @@
...
++}
+ const untaintedBasePrototype$1 = {};
+-const isAngularZonePresent$1 = () => {
+-  return !!globalThis.Zone;
+-};
+ function getUntaintedPrototype$1(key) {
+   if (untaintedBasePrototype$1[key])
+     return untaintedBasePrototype$1[key];
+-  const defaultObj = globalThis[key];
++  const defaultObj = angularZoneUnpatchedAlternative$1(key) || globalThis[key];
```
pauldambra, line 39: so we use the angular one if its present or the default global
even if this was an expensive lookup (it's not) the results are cached

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -7774,7 +7822,55 @@ index d9c57625633fb87da27fc5948c1b15a7c4e4caa5..1932f81f2e330cc904e8349a74911e6b
...
+-};
++
++/*
++Angular zone patches many things and can pass the untainted checks below, causing performance issues
++Angular zone, puts the unpatched originals on the window, and the names for hose on the zone object.
++So, we get the unpatched versions from the window object if they exist.
++You can rename Zone, but this is a good enough proxy to avoid going to an iframe to get the untainted versions.
++see: https://github.com/angular/angular/issues/26948
++*/
++function angularZoneUnpatchedAlternative(key) {
```
pauldambra, line 7840: same here because we're patching a bundle that is tree-shaken badly

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -1,8 +1,55 @@
...
++  const defaultObj = angularZoneUnpatchedAlternative$1(key) || globalThis[key];
+   const defaultPrototype = defaultObj.prototype;
+   const accessorNames = key in testableAccessors$1 ? testableAccessors$1[key] : void 0;
+   const isUntaintedAccessors = Boolean(
+@@ -57,7 +76,7 @@ function getUntaintedPrototype$1(key) {
+       }
+     )
+   );
+-  if (isUntaintedAccessors && isUntaintedMethods && !isAngularZonePresent$1()) {
++  if (isUntaintedAccessors && isUntaintedMethods) {
```
pauldambra, line 48: and so no angular check here

patches/@rrweb__record@2.0.0-alpha.18.patch
```diff
@@ -1,8 +1,55 @@
...
 --- a/dist/record.js
 +++ b/dist/record.js
-@@ -68,10 +68,10 @@ function getUntaintedPrototype$1(key) {
+@@ -26,14 +26,33 @@ const testableMethods$1 = {
+   Element: [],
+   MutationObserver: ["constructor"]
+ };
++/*
++Angular zone patches many things and can pass the untainted checks below, causing performance issues
++Angular zone, puts the unpatched originals on the window, and the names for hose on the zone object.
```
daibhin, line 11:
```suggestion
+Angular zone, puts the unpatched originals on the window, and the names for those on the zone object.
```

daibhin (approved):
Feels like it couldn't hurt 😅 Given this is an Angular specific issue I think it's relatively safe to test it out

PR #1688, opened 2025-01-24
fix: Fix hitting call stack limit when deep comparing circular objects

robbie-c:
## Changes

Incorporates @abhyuditjain's suggestion from https://github.com/PostHog/posthog-js/pull/1686#issuecomment-2612699873

I was surprised that WeakMap was as well-supported as it is, and we can actually use it! https://caniuse.com/?search=JavaScript%20WeakMap

## Checklist

### DIFF
--- a/react/src/utils/__tests__/object-utils.test.ts
+++ b/react/src/utils/__tests__/object-utils.test.ts
@@ -30,7 +30,8 @@ describe('object-utils', () => {
             [true, [1, 2, 3], [1, 2, 3]],
             [false, [1, 2, 3], [1, 2, 4]],
             [true, { a: circularArray1 }, { a: circularArray1 }],
-            // [false, { a: circularArray1 }, { a: circularArray2 }], // TODO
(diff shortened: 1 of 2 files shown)


react/src/utils/__tests__/object-utils.test.ts
```diff
@@ -30,7 +30,8 @@ describe('object-utils', () => {
             [true, [1, 2, 3], [1, 2, 3]],
             [false, [1, 2, 3], [1, 2, 4]],
             [true, { a: circularArray1 }, { a: circularArray1 }],
-            // [false, { a: circularArray1 }, { a: circularArray2 }], // TODO
+            [true, { a: circularArray1 }, { a: circularArray2 }],
+            [true, circularArray1, [circularArray1]],
```
rafaeelaudibert, line 34: Funny how these can be the same, JS is really a nice piece of technology

rafaeelaudibert (approved):
Well, if it's this easy, then I guess we don't need to go the `deepComparePostHogConfig`-only route, nice :)

PR #1689, opened 2025-01-24
feat: Add new rule to guarantee events are tracked as passive

rafaeelaudibert:
We'll never call `preventDefault` on our events and, for that reason, it's smarter to tag our events as `passive: true` letting the browser know that it can continue processing the event, it doesn't need to wait for us.

This should improve performance on lower-end devices.

I've released this for `scroll` events which are more critical - called hundreds of times a second - and I'm monitoring to see if we'll decrease event ingestion count or not. If I don't see a decrease, I'll ship this as well :)


robbie-c (commented):
LGTM!

Is the bundle size smaller if you make a `{ passive: true}` object somewhere and import it?

src/entrypoints/recorder.ts
```diff
@@ -278,53 +278,58 @@ function initXhrObserver(cb: networkCallback, win: IWindow, options: Required<Ne
...
+                                    url: url.toString(),
+                                    initiatorType: 'xmlhttprequest',
+                                })
+                                cb({ requests })
+                            })
+                            .catch(() => {
+                                //
+                            })
+                    },
+                    { passive: true }
```
pauldambra, line 330: can i be precious and we not change this... changes in here can really break customers' sites so i change it very carefully 🙈

robbie-c: Ah good catch

src/utils/index.ts
```diff
@@ -236,7 +236,7 @@ export const registerEvent = (function () {
         }
 
         if (element.addEventListener && !oldSchool) {
-            element.addEventListener(type, handler, !!useCapture)
+            element.addEventListener(type, handler, { capture: !!useCapture, passive: true })
         } else {
```
rafaeelaudibert, line 240: So, I was trying to understand what this function is for, and it's intended to support extremely old browsers. If you check https://caniuse.com/?search=addEventListener you'll see this is only ever useful for IE 6-8 which we supposedly don't support 

```
"browserslist": [
    "> 0.5%, last 2 versions, Firefox ESR, not dead, IE 11"
]
```

Of course, we have the `oldSchool` flag which triggers this code for  `window.load` when trying to initialize `posthog-js`. https://github.com/PostHog/posthog-js/blob/make-all-event-listeners-passive/src/posthog-core.ts#L2315-L2318

That's only triggered, however, if `document.addEventListener` doesn't exist which, again, is only if you're running IE6-8. Anything against removing this code?

I'll instead replace it with a generic `addEventListener` function intended to avoid having to repeat `passive: true` everywhere and reduce our bundle size slightly.

@pauldambra @robbie-c

pauldambra: nope... that's super early code for sure...

we should check the ES5 bundle still works but hopefully CI just covers that.

eslint-rules/no-add-event-listener.js
```diff
@@ -0,0 +1,59 @@
...
+                                    ? context.getSourceCode().getText(node.arguments[2]) === 'true'
+                                        ? ', { capture: true }'
+                                        : `, ${context.getSourceCode().getText(node.arguments[2])}`
+                                    : ''
+
+                            // Add import if needed (note: this is a basic implementation, it won't always work)
+                            const importFix = fixer.insertTextBefore(
+                                context.getSourceCode().ast,
+                                "import { addEventListener } from './utils'\n"
+                            )
```
rafaeelaudibert, line 44: Wondering if this makes any sense. Most often than not it'll be correct, but it will be a broken import sometimes

Maybe we shouldn't be autofixing this at all?

rafaeelaudibert:
@robbie-c @pauldambra Added a couple of new commits to reduce bundle size, and saved some bytes everywhere - even compared to before!

eslint-rules/no-add-event-listener.js
pauldambra: yeah, I'd leave the fix (if we can't detect the correct path). most folks' editors should be able to create the import on demand anyway

rafaeelaudibert: I'll see if `eslint` allows me to detect what file I'm in, if so, I'll keep it

robbie-c (approved):
What @pauldambra said about the ES5 bundle still working for people. Do you still have your IE11 setup? Can you make sure that we're still tracking scrolls correctly?

(I'm 99% confident we will be, just... you know.. changes with a large impact if something goes wrong)

pauldambra (approved):
yep, running locally I still see heatmaps and scrolls and things being captured

CI is green 👍

rafaeelaudibert:
@pauldambra's tests are reassuring, and scroll still works on my IE11 from the change last week, I'll try the changes here locally, let's hope my IE11 installation is trustworthy

rafaeelaudibert:
Hmm, I can't run the demo NextJS app on IE11 because it uses `react-refresh` and that doesn't support IE11. Trying something else, testing IE11 sucks

rafaeelaudibert:
Ok, found a way, [`e1e109f` (#1689)](https://github.com/PostHog/posthog-js/pull/1689/commits/e1e109fbd3e417a3a63e53e03fdc65cc7397bd31), but it's hacky, it does work, though! 🎉

rafaeelaudibert:
Will be merging this tomorrow morning, more time to monitor before I leave for the day

PR #1695, opened 2025-01-27
fix: survey URL targeting should re-evaluate after the URL changes

lucasheriques:
## Changes

Adds a new effect to `SurveyPopup` to continuously check if the URL has changed to display a survey or not.

Right now, if you add a URL filter, go there, and then change the page, the survey is still persisted on the UI.

But, to match the filter properly, it should instead be hidden, as that survey is only supposed to show on a specific URL.

@dmarticus tagged you as reviewer because maybe there's some historical reason as why it was made to work like that instead of hiding it once the URL changes

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -9,13 +9,12 @@ import {
     SurveyRenderReason,
     SurveyType,
 } from '../posthog-surveys-types'
-import { CaptureResult } from '../types'
 
 import * as Preact from 'preact'
 import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks'
 import { document as _document, window as _window } from '../utils/globals'
 import { createLogger } from '../utils/logger'
-import { isArray, isNull, isNumber } from '../utils/type-utils'
+import { isNull, isNumber } from '../utils/type-utils'
 import { createWidgetShadow, createWidgetStyle } from './surveys-widget'
 import { ConfirmationMessage } from './surveys/components/ConfirmationMessage'
 import { Cancel } from './surveys/components/QuestionHeader'
@@ -446,6 +445,51 @@ export function usePopupVisibility(
         }
     }, [])
 
+    // Add URL change listener to hide survey when URL no longer matches
+    useEffect(() => {
+        if (isPreviewMode || !survey.conditions?.url || !posthog) {
+            return
+        }
+
+        const checkUrlMatch = () => {
+            const urlMatchType = survey.conditions?.urlMatchType ?? 'icontains'
+            const url = survey.conditions?.url
+            if (!url) {
+                return
+            }
+            const urlCheck = surveyUrlValidationMap[urlMatchType](url)
+            if (!urlCheck) {
+                setIsPopupVisible(false)
+                removeSurveyFromFocus(survey.id)
+            }
+        }
+
+        // Listen for browser back/forward and hash changes (for hash-based routing)
+        window.addEventListener('popstate', checkUrlMatch)
+        window.addEventListener('hashchange', checkUrlMatch)
+
+        // Listen for SPA navigation
+        const originalPushState = window.history.pushState
+        const originalReplaceState = window.history.replaceState
+
+        window.history.pushState = function (...args) {
+            originalPushState.apply(this, args)
+            checkUrlMatch()
+        }
+
+        window.history.replaceState = function (...args) {
+            originalReplaceState.apply(this, args)
+            checkUrlMatch()
+        }
+
+        return () => {
+            window.removeEventListener('popstate', checkUrlMatch)
+            window.removeEventListener('hashchange', checkUrlMatch)
+            window.history.pushState = originalPushState
+            window.history.replaceState = originalReplaceState
+        }
+    }, [isPreviewMode, survey, removeSurveyFromFocus, posthog])
+
     return { isPopupVisible, isSurveySent, setIsPopupVisible }
 }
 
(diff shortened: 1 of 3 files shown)


src/extensions/surveys.tsx
```diff
@@ -483,6 +485,70 @@ export function SurveyPopup({
...
+
+        // Add our handler to the existing before_send configuration
+        const originalBeforeSend = posthog.config.before_send
+        if (isArray(originalBeforeSend)) {
+            posthog.config.before_send = [...originalBeforeSend, surveyBeforeSend]
+        } else if (originalBeforeSend) {
+            posthog.config.before_send = [originalBeforeSend, surveyBeforeSend]
+        } else {
+            posthog.config.before_send = surveyBeforeSend
+        }
```
lucasheriques, line 531: I decided to use the existing PostHog capture events for `$pageview` as they work more reliably across different kinds of apps (SPA, SSR, etc). Let me know if that's okay

src/extensions/surveys.tsx
```diff
@@ -483,6 +485,70 @@ export function SurveyPopup({
...
+        const handleUrlChange = () => {
+            checkUrlMatch()
+        }
+
+        // Listen for browser back/forward
+        window.addEventListener('popstate', handleUrlChange)
+
+        // Create our before_send handler
+        const surveyBeforeSend = (eventData: CaptureResult | null) => {
+            if (eventData?.event === '$pageview' || eventData?.event === '$pageleave') {
```
marandaneto, line 517: if autocapture is disabled, those events aren't fired, so this bug would still happen

src/extensions/surveys.tsx
marandaneto, line 531: not sure this is the way to go, `before_send` is a customers hook, feels like this code should live somewhere else and not depend on this hook.

lucasheriques: changed the approach to instead change the `window.history` methods instead, which is default and also used by other similar libraries (sentry, plausible)

src/extensions/surveys.tsx
lucasheriques: changed the approach

src/extensions/surveys.tsx
```diff
@@ -444,6 +445,51 @@ export function usePopupVisibility(
         }
     }, [])
 
+    // Add URL change listener to hide survey when URL no longer matches
+    useEffect(() => {
+        if (isPreviewMode || !survey.conditions?.url || !posthog) {
```
marandaneto, line 450: do we need `posthog`? its not being used in this block

src/extensions/surveys.tsx
```diff
@@ -444,6 +445,51 @@ export function usePopupVisibility(
...
+            return
+        }
+
+        const checkUrlMatch = () => {
+            const urlMatchType = survey.conditions?.urlMatchType ?? 'icontains'
+            const url = survey.conditions?.url
+            if (!url) {
+                return
+            }
+            const urlCheck = surveyUrlValidationMap[urlMatchType](url)
```
marandaneto, line 460: not sure if 100% the same but https://github.com/PostHog/posthog-js/blob/4a0fd3142b78dee170274d590b08646039b41c46/src/posthog-surveys.ts#L274-L277
can we extract to a method and reuse in both places?

src/extensions/surveys.tsx
```diff
@@ -444,6 +445,51 @@ export function usePopupVisibility(
...
+            if (!urlCheck) {
+                setIsPopupVisible(false)
+                removeSurveyFromFocus(survey.id)
+            }
+        }
+
+        // Listen for browser back/forward and hash changes (for hash-based routing)
+        window.addEventListener('popstate', checkUrlMatch)
+        window.addEventListener('hashchange', checkUrlMatch)
```
marandaneto, line 469: > The hashchange event is fired when the fragment identifier of the URL has changed (the part of the URL beginning with and following the # symbol).

add a comment about the hashchange, it wasn't obvious to me

src/extensions/surveys.tsx
```diff
@@ -444,6 +445,51 @@ export function usePopupVisibility(
...
+
+        window.history.pushState = function (...args) {
+            originalPushState.apply(this, args)
+            checkUrlMatch()
+        }
+
+        window.history.replaceState = function (...args) {
+            originalReplaceState.apply(this, args)
+            checkUrlMatch()
+        }
```
marandaneto, line 483: since we also have the popstate and hashchange events, does that mean we're doing the check `checkUrlMatch` twice in some cases? or its either the former (popstate, hashchange) or only pushState/replaceState?

lucasheriques: no, they're mutually exclusive. `hashchange` events are meant for certain SPA frameworks that use hash-based routing, but they don't fire both at the same time.

src/extensions/surveys.tsx
lucasheriques: added!

src/extensions/surveys.tsx
lucasheriques: yeah we can, good point, just did with https://github.com/PostHog/posthog-js/pull/1695/commits/31ce8500ca75e5f095b974e1d59c49c8c7553153

marandaneto (approved):
looks good, just [this](https://github.com/PostHog/posthog-js/pull/1695/files#r1932636765) comment otherwise LGTM

src/extensions/surveys.tsx
lucasheriques: removed, thanks for pointing

Issue #1304, opened 2024-07-15
Feature flag override does not fetch new payload


Bonitis:
When using the toolbar to test feature flag variants, the payload is not refetched with the override values. The override variant is correctly stored in `$override_feature_flags` but it will only return the selected variant name, and the payload never changes.

It would be nice if there was an `$override_feature_flag_payloads` entry with the variant selected so that the payloads could be swapped for testing different experiments. That or have the decide endpoint fetch the override variant and replace the payload under `$feature_flag_payloads`.

dmarticus:
+1 to something like this (from https://posthoghelp.zendesk.com/agent/tickets/21407 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/24000)).

This... doesn't seem _that_ hard – maybe I'll hit it as a side quest this sprint.

evidovly:
Hey! We noticed this issue still occurs. Is there a way to resolve it? Thanks!

PR #1697, opened 2025-01-28
feat: deprecate `featureFlags.override` in favor of `featureFlags.overrideFeatureFlags`, a new function that supports overriding flags and flag payloads

dmarticus:
## Changes

The context for these changes are this feature request (https://github.com/PostHog/posthog-js/issues/1304) and this slack thread: https://posthog.slack.com/archives/C07Q2U4BH4L/p1738088710814869.  TL;DR – if folks want to use `override` to manage the value of the feature flag in the client side, I don't see any reason not to.

To make this change ergonomically (I was frustrated by the semantics of `override`, since to override payloads I'd have to do something like `override($flagValue, $some_bool, $payload)`, instead of just letting users do `override($flagValue, $payload)` without needing to specify the override warning), I ended up deprecating `override` and replaced it with the more ergonomic `overrideFeatureFlags`.  Even though `override` is deprecated, though, it still _works_ – it just calls that new method under the hood.  I added the `@deprecated` flag and a warning so that folks on the latest version of posthog-js move away from it.

Closes https://github.com/PostHog/posthog-js/issues/1304

Updated the docs around this change to support using the new method, too: https://github.com/PostHog/posthog.com/pull/10528/files

## Checklist


src/posthog-featureflags.ts
```diff
@@ -149,7 +150,18 @@ export class PostHogFeatureFlags {
...
+        if (!overriddenPayloads) {
+            return flagPayloads || {}
+        }
+
+        const finalPayloads = extend({}, flagPayloads || {})
+        const overriddenKeys = Object.keys(overriddenPayloads)
+        for (let i = 0; i < overriddenKeys.length; i++) {
+            finalPayloads[overriddenKeys[i]] = overriddenPayloads[overriddenKeys[i]]
+        }
+        return finalPayloads
```
dmarticus, line 164: same pattern that we do in `getFlags`

src/posthog-featureflags.ts
```diff
@@ -379,14 +399,25 @@ export class PostHogFeatureFlags {
...
+                this.instance.persistence.register({ [PERSISTENCE_OVERRIDE_FEATURE_FLAGS]: flags })
+            }
+
+            if (payloads) {
+                this.instance.persistence.register({ [PERSISTENCE_OVERRIDE_FEATURE_FLAG_PAYLOADS]: payloads })
+            }
+        }
+
+        if (triggerFlagEvent) {
+            this._fireFeatureFlagsCallbacks()
```
dmarticus, line 420: this should be sufficient to trigger hooks like `useFeatureFlagEnabled`, which is what I was yapping about here: https://posthog.slack.com/archives/C07Q2U4BH4L/p1738088710814869

src/posthog-featureflags.ts
```diff
@@ -369,8 +381,16 @@ export class PostHogFeatureFlags {
      *
      * @param {Object|Array|String} flags Flags to override with.
      * @param {boolean} [suppressWarning=false] Optional parameter to suppress the override warning.
+     * @param {Object} [payloads] Optional parameter to override the payloads for the feature flags.
+     * @param {boolean} [triggerFlagEvent] Optional parameter to trigger the _fireFeatureFlagsCallbacks() event.
+     * If set to true, calling `override` will trigger the callback, which is useful for triggering things like the `useFeatureFlagEnabled` hook.
```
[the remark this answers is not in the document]
havenbarnes: I think so and would even go as far as to say we might not even want to add an argument for them? Even our docs seem to communicate overrides as being the same as bootstraps outside of how long the value is persisted: https://posthog.com/docs/feature-flags/bootstrapping#overriding-feature-flags

havenbarnes (approved):
Nice improvement and test coverage!

src/posthog-featureflags.ts
```diff
@@ -356,39 +368,63 @@ export class PostHogFeatureFlags {
         this._fireFeatureFlagsCallbacks(errorsLoading)
     }
 
-    /*
-     * Override feature flags on the client-side.  Useful for setting non-persistent feature flags, or for testing/debugging
-     * feature flags in the PostHog app.
+    /**
+     * @deprecated Use overrideFeatureFlags instead. This will be removed in a future version.
```
dmarticus, line 372: added the deprecated tag, so users of this latest version will see the warning in their IDE like this
image

src/posthog-featureflags.ts
```diff
@@ -356,39 +368,63 @@ export class PostHogFeatureFlags {
...
+     */
+    override(flags: boolean | string[] | Record<string, string | boolean>, suppressWarning: boolean = false): void {
+        logger.warn('override is deprecated. Please use overrideFeatureFlags instead.')
```
dmarticus, line 375: logging a deprecation warning too; is this overkill?  cc @havenbarnes

src/posthog-featureflags.ts
```diff
@@ -356,39 +368,63 @@ export class PostHogFeatureFlags {
...
+     *     - posthog.feature_flags.overrideFeatureFlags(false)
+     *     - posthog.feature_flags.overrideFeatureFlags(['beta-feature'])
+     *     - posthog.feature_flags.overrideFeatureFlags({'beta-feature': 'variant'})
+     *     - posthog.feature_flags.overrideFeatureFlags({'beta-feature': 'variant'}, {
+     *         suppressWarning: true,
+     *         payloads: { 'beta-feature': { someData: true } }
+     *       })
      */
-    override(flags: boolean | string[] | Record<string, string | boolean>, suppressWarning: boolean = false): void {
+    overrideFeatureFlags(
```
dmarticus, line 393: implement the new method here

src/posthog-featureflags.ts
havenbarnes: Seems reasonable to me, good to have the forcing function so no one gets completely blindsided later when this gets removed

src/posthog-featureflags.ts
havenbarnes: nit: if we're here changing the shape of the method signature anyway, should we go ahead and transition to a single object parameter (I'd making `flags` optional too) to keep things nice and clean? 

Reason I suggest this is that if someone is only interested in overriding payloads, they will always have to pass some empty value in for `flags` which is a tad annoying

havenbarnes (approved):
Looks good, left one nit on the new method's arguments. Also we'll wanna update the code example here https://posthog.com/docs/feature-flags/bootstrapping#overriding-feature-flags

src/posthog-featureflags.ts
dmarticus: handled this

dmarticus:
> Looks good, left one nit on the new method's arguments. Also we'll wanna update the code example here [posthog.com/docs/feature-flags/bootstrapping#overriding-feature-flags](https://posthog.com/docs/feature-flags/bootstrapping#overriding-feature-flags)

yup, doing that here: https://github.com/PostHog/posthog.com/pull/10528

iamnafets:
This seems to be backwards incompatible. When I loaded the toolbar in prod against version 1.217.1, I get this error whenever I try to override a flag: 

image

This is fixed for me by upgrading posthog-js but I'd recommend against making changes that break existing versions in prod without an update path.

dmarticus:
yup, that was my bad – thanks for the feedback!

PR #1698, opened 2025-01-28
chore: add surveys device types check

marandaneto:
## Changes

Draft for https://github.com/PostHog/posthog/issues/24495 (SDK side)
docs https://github.com/PostHog/posthog.com/pull/10529

...

## Checklist


src/posthog-surveys-types.ts
```diff
@@ -123,7 +123,7 @@ export interface SurveyResponse {
 
 export type SurveyCallback = (surveys: Survey[]) => void
 
-export type SurveyUrlMatchType = 'regex' | 'not_regex' | 'exact' | 'is_not' | 'icontains' | 'not_icontains'
+export type SurveyMatchType = 'regex' | 'not_regex' | 'exact' | 'is_not' | 'icontains' | 'not_icontains'
```
marandaneto, line 126: i think we could rename this to something generic and kill `WebExperimentUrlMatchType` as well

marandaneto: but then this needs to be moved out of the surveys bundle

src/posthog-surveys.ts
```diff
@@ -131,13 +135,39 @@ export function getNextSurveyStep(
...
+    return matchType ?? 'icontains'
+}
+
 // use urlMatchType to validate url condition, fallback to contains for backwards compatibility
 export function doesSurveyUrlMatch(survey: Survey): boolean {
     if (!survey.conditions?.url) {
         return true
     }
+    // if we dont know the url, assume it is not a match
+    const href = assignableWindow?.location?.href
```
marandaneto, line 148: is it ok to use `assignableWindow` here?

lucasheriques: I think if you're getting the `location.href`, which is a default field in the `window` object, you should probably just use the `window.location.href`. but @pauldambra you're my local posthog-js expert, what do you think?

src/posthog-surveys-types.ts
lucasheriques: I think this makes sense. However, since it's generic, I'd say we name it `UrlMatchType` and move it to `src/types.ts`

marandaneto: No `Url` in its name, since its not used only for Urls

src/posthog-surveys.ts
marandaneto: `window` does not exist in that context (or at least there's no code using it) and the code uses `assignableWindow` instead so I guess this is the right thing to do? but yeah @pauldambra \o/

src/posthog-surveys-types.ts
pauldambra: oooh, we were working towards something similar in replay 👀 🤔

pauldambra: URL matching and the like

src/utils/request-utils.ts
```diff
@@ -22,9 +22,9 @@ export const convertToURL = (url: string): HTMLAnchorElement | null => {
     return location
 }
 
-export const isUrlMatchingRegex = function (url: string, pattern: string): boolean {
+export const isMatchingRegex = function (value: string, pattern: string): boolean {
```
pauldambra, line 25: nit: feels more like string utils than request utils

especially now it's more obviously got nothing to do with URLs

src/posthog-surveys.ts
pauldambra: i think it doesn't matter here too much... `assignableWindow` just does some helpful type fangling and in this instance you don't (i don't think) need it

there is a `window` that should be being exported alongside `assignableWindow`

i guess it's more expressive to use that since you're not attempting to assign

pauldambra: https://github.com/PostHog/posthog-js/blob/2a06f91fe83d905f6728d2d44cc65c83a5cb5259/src/utils/globals.ts#L113

src/utils/request-utils.ts
marandaneto: makes sense to move it now

lucasheriques (approved):
looks good!

we already have tests in place for URL matching right?

PR #1701, opened 2025-01-29
fix: should use event listener

pauldambra:
or else the pnpm complains


rafaeelaudibert (approved):
Ahn, is this new code? It must've been commited after I started my branch or else my code would've gotten it

lucasheriques:
learned something new today 🤔

@rafaeelaudibert yup it's my code, merged on a PR yesterday

PR #1703, opened 2025-01-29
docs: Add missing types

### LINKED ISSUE #1702: Possible unintended SessionRecordingOptions type change
I started getting type errors on my initialization call when setting session recording masking options starting in v1.211.0. Tracked down the change that is causing those errors in [this diff here where old lines 404-407 were deleted in `src/types.js`](https://github.com/PostHog/posthog-js/commit/58550b9e853d666be9392668e31c739505d107ce#diff-c54113cf61ec99691748a3890bfbeb00e10efb3f0a76f03a0fd9ec49072e410aL405).

Are these options being deprecated or was this an unintentional deletion? [Documentation](https://posthog.com/docs/session-replay/privacy#input-elements) does still list these options

rafaeelaudibert:
These were incorrectly removed in 58550b9e853d666be9392668e31c739505d107ce

I could say it was Cursor fault, but I should've paid more attention

Closes #1702

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -905,16 +905,32 @@ export interface SessionRecordingOptions {
     /**
      * Derived from `rrweb.record` options
      * @see https://github.com/rrweb-io/rrweb/blob/master/guide.md
-     * @default undefined
      */
     maskTextSelector?: string | null
 
     /**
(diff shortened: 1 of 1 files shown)


pauldambra (approved):
🚢

PR #1704, opened 2025-01-30
fix: make feedback tab survey display responsive

### LINKED ISSUE #27982: Surveys feedback button doesnt work well when resizing the screen
## Bug description

Image

1. create a feedback button survey
2. load it on a website
3.  resize the screen - notice how the position gets broken

### Debug info

```shell
```

bvosk:
## Changes

Closes  posthog/posthog#27982

Before
before

After
after

## Discussion

Ultimately the issue here is that there are some styles in `src/extensions/surveys.tsx` that override the ones in `src/extensions/surveys/surveys-utils.tsx`. So this was just a matter of removing the override causing the problem.

In addition to the lack of responsive behavior, this same style was causing some positioning to be off on some device widths. Here's an example of that with the `center` position:

center_bug

## Checklist
  - Added a playwright test for responsive behavior
- ⚠️ I did not do this. I'm not sure how to account for this. I'd love some assistance with that if possible.
  - This change should be backward compatible

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -14,7 +14,7 @@ import {
 import { addEventListener } from '../utils'
 
 import * as Preact from 'preact'
-import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks'
+import { useContext, useEffect, useMemo, useState } from 'preact/hooks'
 import { document as _document, window as _window } from '../utils/globals'
 import { doesSurveyUrlMatch, SURVEY_LOGGER as logger } from '../utils/survey-utils'
 import { isNull, isNumber } from '../utils/type-utils'
@@ -1008,7 +1008,6 @@ export function FeedbackWidget({
     const [isFeedbackButtonVisible, setIsFeedbackButtonVisible] = useState(true)
     const [showSurvey, setShowSurvey] = useState(false)
     const [styleOverrides, setStyleOverrides] = useState<React.CSSProperties>({})
-    const widgetRef = useRef<HTMLDivElement>(null)
 
     useEffect(() => {
         if (!posthog) {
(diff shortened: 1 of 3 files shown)


src/extensions/surveys.tsx
```diff
@@ -701,7 +701,6 @@ export function FeedbackWidget({
 }): JSX.Element {
     const [showSurvey, setShowSurvey] = useState(false)
     const [styleOverrides, setStyle] = useState({})
-    const widgetRef = useRef<HTMLDivElement>(null)
```
bvosk, line 704: I ended up removing this as it's no longer needed. It muddies up the diff quite a bit, but have a look at the individual commits if you'd like to separate this change from the rest.

src/extensions/surveys/surveys-utils.tsx
```diff
@@ -13,7 +13,7 @@ const logger = createLogger('[Surveys]')
 export const style = (appearance: SurveyAppearance | null) => {
     const positions = {
         left: 'left: 30px;',
-        right: 'right: 30px;',
+        right: 'right: 60px;',
```
bvosk, line 16: Repeating my commit message here for discussion. This wasn't strictly necessary to fix the reported issue, but I believe this fix created this problem of overlap with the feedback button. Perhaps fixing this overlap motivated the style override in the first place. Not sure.

> The 30px tucks the survey right in the bottom right corner of the page when it's presented as a popover. But overlaps with the feedback button when presented in that manner.
> 
> I'm opting to solve this by adding additional clearance in both configurations in favor of simplicity. It could be handled conditionally, but that would require some more logic.
> 
> The "more logic" is more than I originally thought because - the way things are today - a user can configure a popover survey with a `appearance.widgetType`  of tab. So properly handling this requires either adding validation to make that impossible or passing the whole survey - not just `SurveyAppearance` into this `style` utility function. Passing on that for now although there's a case for it.

Here's how it looks with the unchanged `right: 30px` style.

tab_without_margin

lucasheriques:
hey @bvosk thanks for the contribution! apologizes it took me a while to check it out.

From the GIFs it looks great!

there are a couple conflicts, do you mind resolving them when you have some time? I'll also review the code and test it out locally tomorrow

bvosk:
Sure thing @lucasheriques. Went ahead and rebased it as requested.

Thanks very much!

lucasheriques (approved):
this looks great, thank you so much @bvosk

finally this is working as expected ❤️

PR #1708, opened 2025-01-31
feat: allow configuring request queue interval

pauldambra:
the request queue has a fixed interval where it sends events, this allows us to batch events to reduce load on ingestion and reduce network calls on client devices

but for customers where they may have short page visits or automated page redirects this means relying on the unload mechanism which browsers don't guarantee

let's allow configuring that interval between 250 and 5000ms

### DIFF
--- a/src/request-queue.ts
+++ b/src/request-queue.ts
@@ -1,21 +1,31 @@
-import { QueuedRequestOptions } from './types'
+import { QueuedRequestWithOptions, RequestQueueConfig } from './types'
 import { each } from './utils'
 
 import { isArray, isUndefined } from './utils/type-utils'
+import { clampToRange } from './utils/number-utils'
+
+export const DEFAULT_FLUSH_INTERVAL_MS = 3000
 
 export class RequestQueue {
     // We start in a paused state and only start flushing when enabled by the parent
     private isPaused: boolean = true
-    private queue: QueuedRequestOptions[] = []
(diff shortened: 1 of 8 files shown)


src/request-queue.ts
```diff
@@ -1,21 +1,31 @@
-import { QueuedRequestOptions } from './types'
+import { QueuedRequestWithOptions, RequestQueueConfig } from './types'
 import { each } from './utils'
 
 import { isArray, isUndefined } from './utils/type-utils'
+import { clampToRange } from './utils/number-utils'
+
+export const DEFAULT_FLUSH_INTERVAL_MS = 3000
```
benjackwhite, line 7: Feels like a solution that doesn't actually relate to the problem...

The problem sounds more to me like customers want a way to flush data and wait for the response before unloading?

benjackwhite (approved):
No problem with the implementation, just a comment about the source problem

src/request-queue.ts
pauldambra: In the specific case they have an auto-refresh after 2 seconds, so they wouldn't want to delay too long

it's also a much bigger change to wait on refreshes...

what I'd love to do instead is have this queue store to local storage as well as memory, and then this all becomes much safer

(so, yep, this is for sure a workaround I can give this customer quickly, much more than something I'd expect many people to set)

PR #1710, opened 2025-01-31
chore: some rrweb type changes

pauldambra:
Need to export some extra types from posthog-js so that I can use them in PostHog

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -125,12 +125,12 @@ const newQueuedEvent = (rrwebMethod: () => void): QueuedRRWebEvent => ({
     attempt: 1,
 })
 
-type compressedFullSnapshotEvent = {
+export type compressedFullSnapshotEvent = {
     type: EventType.FullSnapshot
     data: string
 }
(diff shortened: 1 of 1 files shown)


src/extensions/replay/sessionrecording-utils.ts
```diff
@@ -32,12 +32,25 @@ export function estimateSize(sizeable: unknown): number {
...
+export const MOUSE_MOVE_SOURCE = 1
+export const MOUSE_INTERACTION_SOURCE = 2
+export const SCROLL_SOURCE = 3
+export const VIEWPORT_RESIZE_SOURCE = 4
+export const INPUT_SOURCE = 5
+export const TOUCH_MOVE_SOURCE = 6
+export const MEDIA_INTERACTION_SOURCE = 7
+export const STYLE_SHEET_RULE_SOURCE = 8
+export const DRAG_SOURCE = 12
+
```
pauldambra, line 53: we had some of these defined here, but not all the ones we use
so we may as well have one way of doing things

rafaeelaudibert (approved):

PR #1617, opened 2024-12-18
fix: update init() type (#1616)

thexeos:
## Changes

- `PostHog.init()`'s TypeScript signature changed from `PostHog | undefined` to just `PostHog`

PR #1712, opened 2025-02-03
refactor: Simplify `Posthog.init()` signature

### LINKED ISSUE #1616: PostHog.init() always returns a value, TS types are misleading
The `PostHog.init()` method always returns a value (`PostHog`) and never returns `undefined`, contrary to what TS types suggest. The current behavior is as follows:

### Returns the primary PostHog instance (`this`) when:
- No name is provided
- The name is 'posthog'
- The token is undefined/empty (with a warning)
- The instance is already loaded (with a warning)

### Returns a new named PostHog instance when:
- A custom name is provided

Relevant code:

https://github.com/PostHog/posthog-js/blob/6da785b18c526f5008c2ec9f0bc98555ca784404/src/posthog-core.ts#L357-L562

rafaeelaudibert:
In a recent version, we updated the body of this function to always return a `Posthog` instance but we forgot to update the type, let's fix that

This is a slight type-level breaking change at `posthog-core.ts`. This is more serious than the recent changes we did to code that was only present in the React SDK. Should we ship this? I assume you'll have thoughts @pauldambra 

Originally written by https://github.com/thexeos but I couldn't get his PR to pass CI - secrets are not shared with contributors PRs

Closes #1616
Supersedes #1617

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -368,7 +368,7 @@ export class PostHog {
         token: string,
         config?: OnlyValidKeys<Partial<PostHogConfig>, Partial<PostHogConfig>>,
         name?: string
-    ): PostHog | undefined {
+    ): PostHog {
         if (!name || name === PRIMARY_INSTANCE_NAME) {
             // This means we are initializing the primary instance (i.e. this)
             return this._init(token, config, name)


src/posthog-core.ts
pauldambra, line 1: >  I assume you'll have thoughts @pauldambra

🤣 i love my js sdk, don't hurt them!

---

I think this is a safe direction to make the change. You're either already ignoring that this could be undefined or handling it.

At least now we'd never return undefined. Potentially as with the previous change someone's TS breaks

e.g. if they have `const x: PostHog | undefined = init(blah)` then TS might complain

but importantly i think this is safe in the running code!

pauldambra (approved):

PR #1713, opened 2025-02-03
fix: checking status is too blunt for paused status

pauldambra:
We're checking if status is paused before pausing..

Makes sense, except we have multiple controls and only one status

So, it's possible to be pausable while sampling has determined you shouldn't be recording at all 🤦 . So if `sampling` had determined we should not record this session then we'd init a pause on every rrweb event... which probably causes rrweb events, and down the spiral to madness

Let's check a paused specific flag instead of the top level status


rafaeelaudibert (approved):

pauldambra:
cc @veryayskiy for visibility

PR #1716, opened 2025-02-04
fix(flags): store setOnce properties in locally persisted feature flag props

havenbarnes:
## Changes

Context: https://posthoghelp.zendesk.com/agent/tickets/24489 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/26418)

Include `setOnce` properties in the locally persisted feature flag person properties: https://posthog.com/docs/feature-flags/adding-feature-flag-code#automatic-overrides

These are stored in local storage, but only used by feature flags for passing into the `/decide` call. This means that we should be safe from `setOnce` properties inadvertently overwriting previous property values.

## Checklist

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -1427,7 +1427,10 @@ export class PostHog {
             this.persistence.set_property(USER_STATE, 'identified')
 
             // Update current user properties
-            this.setPersonPropertiesForFlags(userPropertiesToSet || {}, false)
+            this.setPersonPropertiesForFlags(
+                { ...(userPropertiesToSetOnce || {}), ...(userPropertiesToSet || {}) },
+                false
+            )
 
             this.capture(
                 '$identify',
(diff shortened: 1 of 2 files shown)


src/posthog-core.ts
```diff
@@ -1427,7 +1427,10 @@ export class PostHog {
             this.persistence.set_property(USER_STATE, 'identified')
 
             // Update current user properties
-            this.setPersonPropertiesForFlags(userPropertiesToSet || {}, false)
+            this.setPersonPropertiesForFlags(
```
marandaneto, line 1430: Since `$set` and `$set_once` are different props, I wonder if the server does something different for each set of props.
Right now, we are merging them, which may cause issues if the above is true.
Also, userPropertiesToSet has precedence over userPropertiesToSetOnce, I think this should be part of the method docs at the least.
Your call here, just food for thought.

havenbarnes: > Since $set and $set_once are different props, I wonder if the server does something different for each set of props.

I did check this, `/decide` doesn't do any person property writing so we should be safe there, it is only used for flag evaluation purposes during that single request. `/decide` also just has a single person_properties input, it doesn't have a concept for set_once

> Also, userPropertiesToSet has precedence over userPropertiesToSetOnce, I think this should be part of the method docs at the least.

Ok, will call this out in method docs!

havenbarnes: Added

marandaneto (approved):

PR #1719, opened 2025-02-05
fix(flags): set default state values in feature flag hooks

havenbarnes:
## Changes

Context: https://posthog.slack.com/archives/C07F4BF6WLT/p1738270875752869

TLDR:
To suppress a NextJS hydration error, we were previously forgoing default state values in our feature flags hooks. This was necessary because the `posthog` client methods like `isFeatureEnabled` use browser-only APIs like `localStorage` or `document`, which cannot run in SSR. 

Currently, we "fix" this issue by using solution #1 from the [nextjs docs](https://nextjs.org/docs/messages/react-hydration-error#possible-ways-to-fix-it). However, this results in the first render on the client side returning `undefined` from our feature flag hooks, which is both potentially incorrect (if flags are already loaded) and can cause bugs. Further, this hydration error does not actually apply to anyone using our React library for CSR-only.

However, `posthog-js` does not actually support SSR in any case - `usePosthog` uses the SSR-violating `useContext` hook, and our documentation even requires the `'use client'` [directive to initialize the SDK](https://posthog.com/docs/libraries/next-js#router-specific-instructions).

**Changes in this PR:** apply the default value directly in `useState` in our feature flag hooks. Since the hooks are only meant to be run within CSR anyway, this should not affect any users

Note: You cannot fix a problem like this by setting different default values if running in server vs client: NextJS will exhibit a hydration error if the initial render on the server-side does not match the initial render on the client-side: https://nextjs.org/docs/messages/react-hydration-error#possible-ways-to-fix-it

I've tested this works as expected with the existing (non-SSR) nextjs playground - no more `undefined` on initial render!

## Checklist

### DIFF
--- a/react/src/hooks/useActiveFeatureFlags.ts
+++ b/react/src/hooks/useActiveFeatureFlags.ts
@@ -1,10 +1,10 @@
 import { useEffect, useState } from 'react'
 import { usePostHog } from './usePostHog'
 
-export function useActiveFeatureFlags(): string[] | undefined {
+export function useActiveFeatureFlags(): string[] {
     const client = usePostHog()
 
-    const [featureFlags, setFeatureFlags] = useState<string[] | undefined>(() => client.featureFlags.getFlags())
+    const [featureFlags, setFeatureFlags] = useState<string[]>(() => client.featureFlags.getFlags())
 
     useEffect(() => {
         return client.onFeatureFlags((flags) => {
--- a/react/src/hooks/useActiveFeatureFlags.ts
+++ b/react/src/hooks/useActiveFeatureFlags.ts
@@ -5,8 +5,6 @@ export function useActiveFeatureFlags(): string[] | undefined {
     const client = usePostHog()
 
     const [featureFlags, setFeatureFlags] = useState<string[] | undefined>(() => client.featureFlags.getFlags())
-    // would be nice to have a default value above however it's not possible due
-    // to a hydration error when using nextjs
 
     useEffect(() => {
         return client.onFeatureFlags((flags) => {
--- a/react/src/hooks/useActiveFeatureFlags.ts
+++ b/react/src/hooks/useActiveFeatureFlags.ts
@@ -1,10 +1,12 @@
+'use client'
+
 import { useEffect, useState } from 'react'
 import { usePostHog } from './usePostHog'
 
 export function useActiveFeatureFlags(): string[] | undefined {
     const client = usePostHog()
 
-    const [featureFlags, setFeatureFlags] = useState<string[] | undefined>()
+    const [featureFlags, setFeatureFlags] = useState<string[] | undefined>(() => client.featureFlags.getFlags())
     // would be nice to have a default value above however it's not possible due
     // to a hydration error when using nextjs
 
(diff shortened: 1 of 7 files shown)


react/src/hooks/useActiveFeatureFlags.ts
```diff
@@ -4,9 +4,7 @@ import { usePostHog } from './usePostHog'
 export function useActiveFeatureFlags(): string[] | undefined {
     const client = usePostHog()
 
-    const [featureFlags, setFeatureFlags] = useState<string[] | undefined>()
-    // would be nice to have a default value above however it's not possible due
-    // to a hydration error when using nextjs
+    const [featureFlags, setFeatureFlags] = useState<string[] | undefined>(() => client.featureFlags.getFlags())
```
rafaeelaudibert, line 7: `getFlags()` returns `string[]` so you can simplify the type here

react/src/hooks/useFeatureFlagPayload.ts
```diff
@@ -5,9 +5,7 @@ import { usePostHog } from './usePostHog'
 export function useFeatureFlagPayload(flag: string): JsonType | undefined {
     const client = usePostHog()
 
-    const [featureFlagPayload, setFeatureFlagPayload] = useState<JsonType>()
-    // would be nice to have a default value above however it's not possible due
-    // to a hydration error when using nextjs
+    const [featureFlagPayload, setFeatureFlagPayload] = useState<JsonType>(() => client.getFeatureFlagPayload(flag))
```
rafaeelaudibert, line 8: I just noticed `JsonType` can't be `undefined`, but if you look at the implementation of `getFeatureFlagPayload` it can totally be. We might wanna fix that - the fix is either to extend `JsonType` to list `undefined` as a possible value or to return `null` inside `getFeatureFlagPayload`

rafaeelaudibert: I do think the latter is a breaking change (we'll return `null` where we previously returned `undefined`), so we'll need to go with the first option.

rafaeelaudibert (changes requested):
That's a great find! I was slightly afraid that we had added that comment in an attempt to fix a problem that someone reported with SSR, but it turns out this has been like this since the beginning. I'm sure I'll build a `@posthog/nextjs` library soon, so this is much more than enough.

Nice job! And thanks for explaining what the problem was in Slack, that was helpful :)

I have two type-related comments that I'd like us to fix before we can release this

rafaeelaudibert (approved):
Well, "progress over perfection". Feel free to merge this as is, we can fix the types in a follow-up PR.

react/src/hooks/useFeatureFlagPayload.ts
havenbarnes: Yeah that makes sense, thanks for the callout!

react/src/hooks/useActiveFeatureFlags.ts
havenbarnes: thanks, updated

react/src/hooks/useFeatureFlagPayload.ts
havenbarnes: Updated the type

PR #1722, opened 2025-02-07
fix: check url match twice because of potential race condition

lucasheriques:
## Changes

Also checks if the URL condition is matching before actually rendering it.

This is to test wether or not there's some sort of race condition where the survey is rendered on the correct URL, but the URL changes, and before we process that, the survey is already rendered.

[Instance event where it happened](https://us.posthog.com/project/55348/person/6f932fa8c52e49cd8a581c7aa6a42487#activeTab=events). Survey is set to not sohw up in this event, but it did.

...

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -12,6 +12,7 @@ import {
 
 import * as Preact from 'preact'
 import { useContext, useEffect, useMemo, useRef, useState } from 'preact/hooks'
+import { addEventListener } from '../utils'
 import { document as _document, window as _window } from '../utils/globals'
 import { createLogger } from '../utils/logger'
 import { isNull, isNumber } from '../utils/type-utils'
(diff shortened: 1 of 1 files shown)


src/extensions/surveys.tsx
```diff
@@ -77,6 +77,10 @@ export class SurveyManager {
                 return
             }
         }
+        // Adding this duplicate check to see if there's some sort of race-condition between getActiveMatchingSurveys and actual URL changes, as sometimes Surveys are shown up when they shouldn't be
+        if (!doesSurveyUrlMatch(survey)) {
+            return
+        }
```
marandaneto, line 83: should that be in the `callSurveysAndEvaluateDisplayLogic` method then so it'd hopefully fix for handleWidgetSelector as well?

lucasheriques: yeah that makes more sense - just moved now

marandaneto (approved):

PR #1726, opened 2025-02-10
feat: override exception capture clientside

daibhin:
## Changes

Right now there is no way to disable exception autocapture clientside.

Adds a config option that takes precedence over the server value

### DIFF
--- a/src/extensions/exception-autocapture/index.ts
+++ b/src/extensions/exception-autocapture/index.ts
@@ -22,10 +22,7 @@ export class ExceptionObserver {
     }
 
     public get isEnabled(): boolean {
-        if (
-            !isUndefined(this.instance.config.capture_exceptions) &&
-            isBoolean(this.instance.config.capture_exceptions)
-        ) {
+        if (isBoolean(this.instance.config.capture_exceptions)) {
             return this.instance.config.capture_exceptions
         }
         return this.remoteEnabled ?? false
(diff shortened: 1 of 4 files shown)


src/extensions/exception-autocapture/index.ts
```diff
@@ -21,20 +21,19 @@ export class ExceptionObserver {
...
         return this.remoteEnabled ?? false
     }
 
-    get isCapturing() {
-        return !!(window?.onerror as any)?.__POSTHOG_INSTRUMENTED__
-    }
-
     get hasHandlers() {
-        return this.originalOnUnhandledRejectionHandler || this.unwrapOnError
+        return !isUndefined(this.unwrapOnError)
```
pauldambra, line 32: same, is the assumption here if we have one then we have both?

pauldambra (approved):
couple of questions but looks like code

src/extensions/exception-autocapture/index.ts
```diff
@@ -21,20 +21,19 @@ export class ExceptionObserver {
...
 
-    get isEnabled() {
+    public get isEnabled(): boolean {
+        if (isBoolean(this.instance.config.capture_exceptions)) {
+            return this.instance.config.capture_exceptions
+        }
         return this.remoteEnabled ?? false
     }
 
-    get isCapturing() {
```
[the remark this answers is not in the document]
daibhin: That was my thinking

src/extensions/exception-autocapture/index.ts
daibhin: Yeah. I guess we could look up either but then would would have to handle the two separately e.g. consider adding a handler to the two functions separately, which would complicate the logic elsewhere. I'm happy to go with this as an assumption for now but certainly willing to reconsider if it causes issues

daibhin: Basing it somewhat off the fact that `isCapturing` previously only tested `window.onerror` anyways

PR #1727, opened 2025-02-11
fix: canvas quality value

pauldambra:
when clamping canvas quality to range we aren't first checking it's a number 😓 

this improves that and adds a matrix test to validate the config parsing

### DIFF
--- a/src/extensions/replay/sessionrecording.ts
+++ b/src/extensions/replay/sessionrecording.ts
@@ -337,9 +337,14 @@ export class SessionRecording {
         const canvasRecording_client_side = this.instance.config.session_recording.captureCanvas
         const canvasRecording_server_side = this.instance.get_property(SESSION_RECORDING_CANVAS_RECORDING)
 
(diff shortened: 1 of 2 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -337,14 +337,20 @@ export class SessionRecording {
...
+            const parsed = parseFloat(quality)
+            quality = isNaN(parsed) ? 0.4 : parsed
+        }
 
         return {
             enabled,
-            fps: clampToRange(fps, 0, 12, 'canvas recording fps'),
-            quality: clampToRange(quality, 0, 1, 'canvas recording quality'),
+            fps: clampToRange(fps, 0, 12, 'canvas recording fps', 4),
+            quality: clampToRange(quality, 0, 1, 'canvas recording quality', 0.4),
```
robbie-c, line 353: Could these be constants? I have no idea what a quality of `0.4` means 😓

robbie-c (approved):

src/extensions/replay/sessionrecording.ts
pauldambra: done! 👍

PR #1729, opened 2025-02-13
feat(flags): Add remote config feature flag method that bypasses cached values

havenbarnes:
## Changes

Adds `getRemoteConfigPayload` to feature flags, which re-calls `/decide` and returns the single remote config feature flag payload:

Screenshot 2025-02-12 at 7 00 48 PM
Screenshot 2025-02-12 at 7 00 41 PM

Context: https://posthog.slack.com/archives/C074A08LP1B/p1739376378507109

## Checklist
  - Looking into this!

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -364,7 +364,7 @@ export class PostHogFeatureFlags {
      * Fetches the payload for a remote config feature flag. This method will bypass any cached values and fetch the latest
      * value from the PostHog API.
      *
-     * Note: Encrypted remote config payloads will not be redacted, not decrypted in the response.
+     * Note: Encrypted remote config payloads will not be redacted, nor decrypted in the response.
      *
      * ### Usage:
      *
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -9,6 +9,7 @@ import {
     JsonType,
     Compression,
     EarlyAccessFeature,
+    RemoteConfigFeatureFlagCallback,
 } from './types'
 import { PostHogPersistence } from './posthog-persistence'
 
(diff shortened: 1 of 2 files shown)


src/posthog-featureflags.ts
```diff
@@ -359,6 +360,37 @@ export class PostHogFeatureFlags {
         return payloads[key]
     }
 
+    /*
+     * Fetches the payload for a remote config feature flag. This method will bypass any cached values and fetch the latest
+     * value from the PostHog API.
+     *
+     * Note: Encrypted remote config payloads will not be redacted, not decrypted in the response.
```
haacked, line 367:
```suggestion
     * Note: Encrypted remote config payloads will not be redacted, nor decrypted in the response.
```

src/posthog-featureflags.ts
```diff
@@ -359,6 +360,37 @@ export class PostHogFeatureFlags {
...
+     */
+    getRemoteConfigPayload(key: string, callback: RemoteConfigFeatureFlagCallback): void {
+        const token = this.instance.config.token
+        this.instance._send_request({
+            method: 'POST',
+            url: this.instance.requestRouter.endpointFor('api', '/decide/?v=3'),
+            data: {
+                distinct_id: this.instance.get_distinct_id(),
+                token,
+            },
```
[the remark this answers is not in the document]
havenbarnes: Not needed for this

src/posthog-featureflags.ts
```diff
@@ -359,6 +360,37 @@ export class PostHogFeatureFlags {
...
+     *     getRemoteConfigPayload("home-page-welcome-message", (payload) => console.log(`Fetched remote config: ${payload}`))
+     *
+     * @param {String} key Key of the feature flag.
+     * @param {Function} [callback] The callback function will be called once the remote config feature flag payload has been fetched.
+     */
+    getRemoteConfigPayload(key: string, callback: RemoteConfigFeatureFlagCallback): void {
+        const token = this.instance.config.token
+        this.instance._send_request({
+            method: 'POST',
+            url: this.instance.requestRouter.endpointFor('api', '/decide/?v=3'),
```
haacked, line 380: Why isn't this going to the new `/remote_config` endpoint? I'm assuming it's because this is for a client library and it won't have access to the `personal_api_key`? Should the server libraries support this scenario? I'm going to assume no because they're more likely to use the `personal_api_key`.

haacked (approved):
LGTM! One question, but I don't think it should block this.

src/posthog-featureflags.ts
```diff
@@ -359,6 +360,37 @@ export class PostHogFeatureFlags {
...
+     * Note: Encrypted remote config payloads will not be redacted, nor decrypted in the response.
+     *
+     * ### Usage:
+     *
+     *     getRemoteConfigPayload("home-page-welcome-message", (payload) => console.log(`Fetched remote config: ${payload}`))
+     *
+     * @param {String} key Key of the feature flag.
+     * @param {Function} [callback] The callback function will be called once the remote config feature flag payload has been fetched.
+     */
+    getRemoteConfigPayload(key: string, callback: RemoteConfigFeatureFlagCallback): void {
```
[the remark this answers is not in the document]
havenbarnes: Yeah, was bummed to not do this yet but understandable!

src/posthog-featureflags.ts
havenbarnes: Your assumption is 100% correct! And yeah, server libs may as well always use `/remote_config`, because even if they are not using personal_api_key, it'd still work for non-encrypted payloads and will bypass any caching / events that the decide code may do

haacked: Oh! If `/remote_config` works without the personal API key for unencrypted payloads, why not use that everywhere? It sounds like exactly the behavior you want here. And it would allow us to take some load off `/decide` in theory (I assume remote_config calls decide code under the hood today, but could be separated in the future. I’d check but I’m AFK. 😂).

I don’t feel too strongly about it, just curious.

PR #1739, opened 2025-02-14
feat(web-analytics): Handle set_once latest props in a cross-subdomain way

robbie-c:
## Changes

There are a few changes here

Change 1: the biggest one is that we send set_once props for the *latest* props now. Previously, we just sent *initial* props. This means that this user journey now works

* start on subdomain1.example.com
* go to subdomain2.example.com
* call identify()
* check latest props

Previously, they would not necessarily have latest props, which we saw in a ticket here https://posthog.slack.com/archives/C05LJK1N3CP/p1739469826803839. Now we are using the session-scoped props to set these.

Change 2: reduce how often we send set_once properties. Previously we were sending these with every event with person properties, now we only send them with the first one.

Change 3: the set_once properties will now get proper masking of personal data properties in saved URLs.

## Checklist

### DIFF
--- a/src/__tests__/personProcessing.test.ts
+++ b/src/__tests__/personProcessing.test.ts
@@ -34,6 +34,31 @@ const INITIAL_CAMPAIGN_PARAMS_NULL = {
     $initial_wbraid: null,
 }
 
+const CAMPAIGN_PARAMS_NULL = {
+    _kx: null,
+    dclid: null,
+    fbclid: null,
+    gad_source: null,
+    gbraid: null,
+    gclid: null,
+    gclsrc: null,
+    $host: null,
+    igshid: null,
+    irclid: null,
+    li_fat_id: null,
+    mc_cid: null,
+    msclkid: null,
+    rdt_cid: null,
+    ttclid: null,
+    twclid: null,
+    utm_campaign: null,
+    utm_content: null,
+    utm_medium: null,
+    utm_source: null,
+    utm_term: null,
+    wbraid: null,
+}
+
 jest.mock('../utils/globals', () => {
     const orig = jest.requireActual('../utils/globals')
     const mockURLGetter = jest.fn()
@@ -215,20 +240,27 @@ describe('person processing', () => {
             expect(identifyCall[0].event).toEqual('$identify')
             expect(identifyCall[0].$set_once).toEqual({
                 ...INITIAL_CAMPAIGN_PARAMS_NULL,
+                ...CAMPAIGN_PARAMS_NULL,
                 $initial_current_url: 'https://example.com?utm_source=foo',
                 $initial_host: 'example.com',
                 $initial_pathname: '/',
                 $initial_referrer: 'https://referrer.com',
                 $initial_referring_domain: 'referrer.com',
                 $initial_utm_source: 'foo',
+                $current_url: 'https://example.com?utm_source=foo',
+                $host: 'example.com',
+                $pathname: '/',
+                $referrer: 'https://referrer.com',
+                $referring_domain: 'referrer.com',
+                utm_source: 'foo',
             })
         })
 
         it('should preserve initial referrer info across a separate session', async () => {
             // arrange
-            const { posthog, beforeSendMock } = await setup('identified_only')
             mockReferrerGetter.mockReturnValue('https://referrer1.com')
             mockURLGetter.mockReturnValue('https://example1.com/pathname1?utm_source=foo1')
+            const { posthog, beforeSendMock } = await setup('identified_only')
 
             // act
             // s1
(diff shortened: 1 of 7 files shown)


src/__tests__/personProcessing.test.ts
```diff
@@ -258,37 +291,36 @@ describe('person processing', () => {
...
                 $initial_referrer: 'https://referrer1.com',
                 $initial_referring_domain: 'referrer1.com',
                 $initial_utm_source: 'foo1',
+                $current_url: 'https://example2.com/pathname2?utm_source=foo2',
+                $host: 'example2.com',
+                $pathname: '/pathname2',
+                $referrer: 'https://referrer2.com',
+                $referring_domain: 'referrer2.com',
+                utm_source: 'foo2',
             })
```
rafaeelaudibert, line 307: Nice

src/posthog-core.ts
```diff
@@ -1068,17 +1069,32 @@ export class PostHog {
...
-        let setOnceProperties = extend({}, this.persistence.get_initial_props(), dataSetOnce || {})
+        const initialProps = this.persistence.get_initial_props()
+        const sessionProps = this.sessionPropsManager?.getSetOnceInitialSessionPropsProps()
+        let setOnceProperties = extend({}, initialProps, sessionProps || {}, dataSetOnce || {})
         const sanitize_properties = this.config.sanitize_properties
         if (sanitize_properties) {
             logger.error('sanitize_properties is deprecated. Use before_send instead')
             setOnceProperties = sanitize_properties(setOnceProperties, '$set_once')
         }
+        this._personProcessingSetOncePropertiesSent = true
```
rafaeelaudibert, line 1097: Should this really be set here? I'd prefer we set this once we actually call `capture`. If someone calls `_calculate_set_once_properties` but doesn't actually send the props then we'll miss them

src/session-props.ts
```diff
@@ -78,20 +80,38 @@ export class SessionPropsManager {
...
         }
     }
+
+    getSessionProps() {
+        const result: Record<string, any> = {}
+        const p = this._props()
+        each(p, function (v, k) {
+            result['$client_session_' + stripLeadingDollar(k)] = v
+        })
+    }
```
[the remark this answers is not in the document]
robbie-c: Tbh I should just delete this function, it's not used, brb

rafaeelaudibert, line 112: Not returning anything ⚠️

src/session-props.ts
```diff
@@ -6,52 +6,54 @@
...
+        instance?: PostHog
+    ) => LegacySessionSourceProps | CurrentSessionSourceProps
 
     constructor(
         instance: PostHog,
         sessionIdManager: SessionIdManager,
         persistence: PostHogPersistence,
-        sessionSourceParamGenerator?: (instance?: PostHog) => SessionSourceProps
+        sessionSourceParamGenerator?: (instance?: PostHog) => LegacySessionSourceProps
     ) {
```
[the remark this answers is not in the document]
rafaeelaudibert: Ahn, not sure greptile is correct here? Take a look

src/utils/event-utils.ts
```diff
@@ -194,46 +194,56 @@ export const Info = {
...
-    initialPersonPropsFromInfo: function (info: Record<string, any>): Record<string, any> {
-        const { r: initial_referrer, u: initial_current_url } = info
-        const referring_domain =
-            initial_referrer == null
-                ? undefined
-                : initial_referrer == '$direct'
-                  ? '$direct'
-                  : convertToURL(initial_referrer)?.host
+    personPropsFromInfo: function (info: Record<string, any>): Record<string, any> {
+        const { r: r, u: u } = info
```
rafaeelaudibert, line 216: lol

```suggestion
        const { r, u } = info
```

rafaeelaudibert: Or better

```suggestion
        const { r: infoReferrer, u: infoUrl } = info
```

rafaeelaudibert (commented):
This is complex code so I can't say I fully understand the changes. The tests are exactly what I'd like to see/this to behave, so looks good? Greptile found a small bug, can you fix?

src/session-props.ts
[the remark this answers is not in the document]
robbie-c: It's done

robbie-c: I just killed this function instead as it's not used

src/utils/event-utils.ts
robbie-c: Done

src/posthog-core.ts
robbie-c: Hmm, I liked it here because then it was being read/written in the same function. It's a private function and only called from `capture`. What do you think?

rafaeelaudibert: I'm afraid it might become a bug in the future, but it does make sense for now. Maybe add a warning at the top of this function saying that it's not a pure function and you should make sure you're sending the results returned from the first ever call?

rafaeelaudibert (approved):

PR #1742, opened 2025-02-16
fix: replace useMemo for react 19 support

### LINKED ISSUE #1741: Not usable with React 19 because of useMemo
`useMemo` is removed in React 19, so the `useMemo` in `PostHogProvider` breaks any app using React 19 or higher.

wes337:
## Changes

Replace `useMemo` with equivalent through `useState` and `useEffect`. React 19 has removed the `useMemo` function entirely.

Closes #1741 

## Checklist
- [X] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [X] Accounted for the impact of any changes across different browsers
- [X] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

### DIFF
--- a/react/src/context/PostHogProvider.tsx
+++ b/react/src/context/PostHogProvider.tsx
@@ -46,9 +46,8 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     // We're not storing a simple boolean here because we want to be able to detect if the
     // apiKey or options have changed.
     const [alreadyInitialized, setAlreadyInitialized] = useState<AlreadyInitialized>(false)
-    const [posthog, setPosthog] = useState<PostHog>()
 
-    useEffect(() => {
+    const getPosthogClient = () => {
         if (client) {
             if (apiKey) {
                 console.warn(
@@ -101,8 +100,7 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
                 })
 
                 // Return the same already-initialized global client
-                setPosthog(posthogJs)
-                return
+                return posthogJs
             }
 
             // If it's the first time running this, but it has been loaded elsewhere, warn the user about it.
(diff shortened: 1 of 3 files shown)


src/extensions/surveys.tsx
rafaeelaudibert, line 1: This file is using Preact, it shouldn't be changed

src/extensions/surveys/components/QuestionTypes.tsx
rafaeelaudibert, line 1: Same as above, this is Preact

react/src/context/PostHogProvider.tsx
rafaeelaudibert, line 1: Yikes, this is tricky. I'll investigate the changes next week and see if this is the best/recommended approach

rafaeelaudibert (changes requested):
I'm investigating if there's any other approach, this feels weird, but I understand the rationale - and I know this would work!

Can you remove the changes to the Preact files? Those are definitely not necessary.

react/src/context/PostHogProvider.tsx
wes337: Yeah I understand. It's difficult to know the full effect of this change -- but it is in theory the same thing.

For now in my application we have just stopped using the `<PostHogProvider>`

rafaeelaudibert:
@wes337 Do you have a link to the announcement where they've **removed** `useMemo`? I know the function isn't **needed** when React Compiler is enabled but I haven't seen any announcements saying that `useMemo` has been removed from the tree and that we need to move away from it.

If I look at React's source code I can totally see it: https://github.com/facebook/react/blob/v19.0.0/packages/react/src/ReactHooks.js#L146

lucasheriques:
I also don't think `useMemo` will be removed - I haven't seen that announced anywhere @wes337

wes337:
You're right, I must have been mistaken about it getting removed, I could have sworn I saw this somewhere. I'm not having the issue with React 19 anymore, -- so it was something related to the upgrade in my own app.

Sorry about this, and thanks for taking the time to review and look into it!

PR #1743, opened 2025-02-17
feat: add more debug signals

pauldambra:
it can be tricky to know what is happening when users report errors

let's add more debug signals

Screenshot 2025-02-17 at 12 40 01

needs a follow-up to add these to core definitions

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -996,8 +996,8 @@ export class PostHog {
                 properties['$sdk_debug_replay_internal_buffer_size'] = this.sessionRecording['buffer'].size
             }
             properties['$sdk_debug_retry_queue_size'] = this._retryQueue?.['queue']?.length
-        } catch {
-            properties['$sdk_debug_error_capturing_properties'] = true
+        } catch (e: any) {
+            properties['$sdk_debug_error_capturing_properties'] = String(e)
         }
 
         if (this.requestRouter.region === RequestRouterRegion.CUSTOM) {
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -991,6 +991,13 @@ export class PostHog {
 
         if (this.sessionRecording) {
             properties['$recording_status'] = this.sessionRecording.status
+            try {
+                properties['$sdk_debug_replay_internal_buffer_length'] = this.sessionRecording['buffer'].data.length
+                properties['$sdk_debug_replay_internal_buffer_size'] = this.sessionRecording['buffer'].size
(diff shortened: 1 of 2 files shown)


src/posthog-core.ts
```diff
@@ -989,8 +989,15 @@ export class PostHog {
...
 
-        if (this.sessionRecording) {
-            properties['$recording_status'] = this.sessionRecording.status
+        try {
+            if (this.sessionRecording) {
+                properties['$recording_status'] = this.sessionRecording.status
+                properties['$sdk_debug_replay_internal_buffer_length'] = this.sessionRecording['buffer'].data.length
+                properties['$sdk_debug_replay_internal_buffer_size'] = this.sessionRecording['buffer'].size
+            }
+            properties['$sdk_debug_retry_queue_size'] = this._retryQueue?.['queue']?.length
```
pauldambra, line 998: this should give us a view on how much data replay is currently holding
and whether the SDK retry queue is high

hopefully giving signal that e.g. network requests to send data are failing in replay 
when smaller event data might be succeeding

src/posthog-core.ts
```diff
@@ -989,8 +989,15 @@ export class PostHog {
             properties['$window_id'] = windowId
         }
 
-        if (this.sessionRecording) {
-            properties['$recording_status'] = this.sessionRecording.status
+        try {
```
oliverb123, line 992: Is the try/catch here just an abundance of caution?

src/posthog-core.ts
```diff
@@ -989,8 +989,15 @@ export class PostHog {
...
+            if (this.sessionRecording) {
+                properties['$recording_status'] = this.sessionRecording.status
+                properties['$sdk_debug_replay_internal_buffer_length'] = this.sessionRecording['buffer'].data.length
+                properties['$sdk_debug_replay_internal_buffer_size'] = this.sessionRecording['buffer'].size
+            }
+            properties['$sdk_debug_retry_queue_size'] = this._retryQueue?.['queue']?.length
+        } catch {
```
oliverb123, line 999: Any harm grabbing the message from the error and sticking it in the props too?

oliverb123 (approved):

src/posthog-core.ts
pauldambra: only laziness because it's a pain getting the messages off errors in js 🤣

pauldambra: 🤔 we do `defaultExceptionMessage: String(error),` in error tracking... ok you can be way lazier than I thought

PR #1745, opened 2025-02-18
fix: last seen survey date logic

lucasheriques:
## Changes

currently, if you set up a survey to wait for 1 day to show again, it'll be displayed straight way.

that's because the logic that tests this case is wrong - if only 5 minutes have passed, it'll still render it.

So I extracted that logic into a separate function to fix the logic and make it easier to create tests for it
...

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -32,6 +32,7 @@ import {
     getContrastingTextColor,
     getDisplayOrderQuestions,
     getSurveySeen,
+    hasWaitPeriodPassed,
     sendSurveyEvent,
     style,
     SurveyContext,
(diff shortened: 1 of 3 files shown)


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -694,6 +694,20 @@ const getSurveyInteractionProperty = (survey: Survey, action: string): string =>
     return surveyProperty
 }
 
+export const hasWaitPeriodPassed = (
+    lastSeenSurveyDate: string | null,
+    waitPeriodInDays: number | undefined
+): boolean => {
+    if (!waitPeriodInDays || !lastSeenSurveyDate) {
+        return true
+    }
```
marandaneto, line 703: i think either the method has to be renamed or the return has to be inverted.
`hasWaitPeriodPassed` returns true if no waitPeriodInDays, so its semantically inverted

src/extensions/surveys.tsx
```diff
@@ -69,13 +70,9 @@ export class SurveyManager {
...
-        if (surveyWaitPeriodInDays && lastSeenSurveyDate) {
-            const today = new Date()
-            const diff = Math.abs(today.getTime() - new Date(lastSeenSurveyDate).getTime())
-            const diffDaysFromToday = Math.ceil(diff / (1000 * 3600 * 24))
-            if (diffDaysFromToday < surveyWaitPeriodInDays) {
-                return
-            }
+
+        if (!hasWaitPeriodPassed(lastSeenSurveyDate, surveyWaitPeriodInDays)) {
+            return
```
marandaneto, line 75: see https://github.com/PostHog/posthog-js/pull/1745/files#r1959159402

marandaneto (approved):
looks solid

PR #1747, opened 2025-02-18
fix: initializing surveys twice and calling survey shown/dismissed more than once per survey

lucasheriques:
# Fix race conditions in survey initialization

## Problem
Two intermittent race conditions causing duplicate survey displays:
1. Multiple concurrent calls to `/surveys` endpoint
2. Multiple `SurveyManager` instances being created during PostHog's remote config loading process

These issues don't occur consistently because they depend on:
- Network response timing between config loading attempts
- Browser's script loading/execution speed
- Whether PostHog needs to fall back to different config loading strategies

## Solution
Added synchronization mechanisms:
1. `_isFetchingSurveys` flag to prevent concurrent API calls
2. `_isInitializingSurveys` flag with proper cleanup to ensure only one `SurveyManager` instance is created

Note: Multiple "Already initializing surveys" logs are expected due to PostHog's config loading retry mechanism, but only one initialization actually occurs.

## Testing
Verify:
1. Single network call to `/surveys`
2. Single interval ID in console logs
3. Surveys display only once

## Checklist


src/posthog-surveys.ts
```diff
@@ -214,34 +216,63 @@ export class PostHogSurveys {
...
+                        this._surveyManager = generateSurveys(this.instance)
+                    }
+                } finally {
+                    this._isInitializingSurveys = false
+                }
             }
-        } else {
-            this._surveyManager = generateSurveys(this.instance)
+        } catch (e) {
+            this._isInitializingSurveys = false
```
marandaneto, line 274: should this be within the finally block instead?

src/posthog-surveys.ts
```diff
@@ -214,34 +216,63 @@ export class PostHogSurveys {
             return
         }
 
-        const generateSurveys = phExtensions.generateSurveys
-
-        if (!this._decideServerResponse) {
-            logger.warn('Decide not loaded yet. Not loading surveys.')
+        if (this._isInitializingSurveys) {
```
marandaneto, line 219: should this check be moved up like the 1st thing?

src/posthog-surveys.ts
```diff
@@ -261,38 +291,50 @@ export class PostHogSurveys {
...
-                                survey.conditions?.actions?.values &&
-                                survey.conditions?.actions?.values?.length > 0)
-                    )
-
-                    if (eventOrActionBasedSurveys.length > 0) {
-                        this._surveyEventReceiver?.register(eventOrActionBasedSurveys)
-                    }
+            // Prevent concurrent API calls
+            if (this._isFetchingSurveys) {
+                return
```
marandaneto, line 296:
```suggestion
                return callback([])
```

src/posthog-surveys.ts
```diff
@@ -261,38 +291,50 @@ export class PostHogSurveys {
...
-                            (survey.conditions?.actions &&
-                                survey.conditions?.actions?.values &&
-                                survey.conditions?.actions?.values?.length > 0)
-                    )
-
-                    if (eventOrActionBasedSurveys.length > 0) {
-                        this._surveyEventReceiver?.register(eventOrActionBasedSurveys)
-                    }
+            // Prevent concurrent API calls
+            if (this._isFetchingSurveys) {
```
marandaneto, line 295: similar to https://github.com/PostHog/posthog-js/pull/1747/files#r1961284501

src/posthog-surveys.ts
```diff
@@ -261,38 +291,50 @@ export class PostHogSurveys {
...
+            try {
+                this._isFetchingSurveys = true
+                this.instance._send_request({
+                    url: this.instance.requestRouter.endpointFor(
+                        'api',
+                        `/api/surveys/?token=${this.instance.config.token}`
+                    ),
+                    method: 'GET',
+                    callback: (response) => {
+                        this._isFetchingSurveys = false
```
marandaneto, line 308: i think this should be done after the callback is executed so L312 and 330, otherwise you still might have race conditions in the `_surveyEventReceiver` and `persistence`

src/posthog-surveys.ts
```diff
@@ -261,38 +291,50 @@ export class PostHogSurveys {
...
-            })
+            try {
+                this._isFetchingSurveys = true
+                this.instance._send_request({
+                    url: this.instance.requestRouter.endpointFor(
+                        'api',
+                        `/api/surveys/?token=${this.instance.config.token}`
+                    ),
+                    method: 'GET',
+                    callback: (response) => {
```
marandaneto, line 307: if there's no connection, is the `callback` still called so we can reset the `_isFetchingSurveys`?
or is it gonna throw and the `catch` will reset it?

marandaneto:
https://github.com/PostHog/posthog-js/blob/8a82b350fc308deb39746d5d968ffeda6fb3e5ba/src/posthog-surveys.ts#L224-L226
i think this should only be init if the `_surveyManager` was also properly inited, `_surveyManager` might fail because of eg 'Could not load surveys script' but the `_surveyEventReceiver` is now in a limbo state

src/posthog-surveys.ts
```diff
@@ -214,34 +216,63 @@ export class PostHogSurveys {
...
-                loadExternalDependency(this.instance, 'surveys', (err) => {
-                    if (err) {
-                        logger.error('Could not load surveys script', err)
-                        return
-                    }
+            if (!this._decideServerResponse) {
+                logger.warn('Decide not loaded yet. Not loading surveys.')
+                this._isInitializingSurveys = false
+                return
+            }
```
rafaeelaudibert, line 233: Can probably be outside the `try/catch`

src/posthog-surveys.ts
```diff
@@ -214,34 +216,63 @@ export class PostHogSurveys {
...
 
-                    this._surveyManager = phExtensions.generateSurveys?.(this.instance)
-                })
+            if (this._surveyEventReceiver == null) {
+                this._surveyEventReceiver = new SurveyEventReceiver(this.instance)
+            }
```
rafaeelaudibert, line 237: Can probably be outside the try/catch as well?

src/posthog-surveys.ts
rafaeelaudibert: I agree, and you can then remove all of the inner `this._isInitializingSurveys = false` calls

src/posthog-surveys.ts
rafaeelaudibert: Javascript is monothread, I don't see a way for this code to yield control to a separate function in between here and line 330, so this **should** be fine, and actually looks slightly cleaner to me

src/posthog-surveys.ts
rafaeelaudibert: This needs to be tested, but I believe `callback` will still be called even if there's no connection

rafaeelaudibert (commented):
Dropping a couple of very cheap 2 Brazilian cents

src/posthog-surveys.ts
lucasheriques: agreed, made the change

src/posthog-surveys.ts
lucasheriques: added a 10s timeout to it, and [created tests](https://github.com/PostHog/posthog-js/pull/1747/commits/b50c1724d4e59040e2f1e2b51350390f6d0135c3#diff-04488c5fc8f15661dc5ea1cd4d4be9e1b0cf35cefaa2c3b84c8f5a29fd6e487fR204-R215) to make sure its happening as expected

lucasheriques:
@marandaneto also changed to now only instantiate `SurveyEventReceiver` once, on `loadIfEnabled`

src/posthog-surveys.ts
```diff
@@ -261,38 +282,51 @@ export class PostHogSurveys {
...
+
+            try {
+                this._isFetchingSurveys = true
+                this.instance._send_request({
+                    url: this.instance.requestRouter.endpointFor(
+                        'api',
+                        `/api/surveys/?token=${this.instance.config.token}`
+                    ),
+                    method: 'GET',
+                    timeout: 10000, // 10 second timeout
```
marandaneto, line 298: we should add this under the options, eg feature_flag_request_timeout_ms

src/posthog-surveys.ts
marandaneto: agree, and this [test](https://github.com/PostHog/posthog-js/pull/1747/files#r1962233991) actually calls the callback in the test, so its not really testing what it would happen in case of a timeout/no connection

marandaneto:
> SurveyEventReceiver

if `getSurveys` is called before `loadIfEnabled` is fully initialized, either `getSurveys` should await for `loadIfEnabled`, or it should return nothing, but it seems wrong to instant things elsewhere.

`loadIfEnabled` calls `generateSurveys`, and only `generateSurveys` calls `getSurveys` automatically, so at this point everything should be init.

IMO the test should be wrong then unless I miss something

lucasheriques:
@marandaneto agree - `getSurveys` shouldn't be called if `loadIfEnabled` wasn't.

I changed the test to first call `loadIfEnabled` and instantiate the event receiver.

Also, couple of other changes:

- [Added the config for surveys request timeout](https://github.com/PostHog/posthog-js/pull/1747/commits/c2549f5713d35026f2a048ae5f528c63df56f1a3) (plus [add default value for survey timeout](https://github.com/PostHog/posthog-js/pull/1747/commits/1021524dc5fbf3ba705a247f3419130d9edf3912))
- [remove double check for the survey manager - not needed](https://github.com/PostHog/posthog-js/pull/1747/commits/56e2c7add55bac914748fb80dfd7d5b8fdb4c301)
- [move _isInitializingSurveys to inside the callback](https://github.com/PostHog/posthog-js/pull/1747/commits/92d7ee469cac86788dd5c88547843f0ef7607567): as you pointed, it's an aync call, so it should be only called if the survey is sucessfully loaded. on the `finally` block, it could be called even without finishing it

src/posthog-surveys.ts
```diff
@@ -214,34 +221,50 @@ export class PostHogSurveys {
...
+                            this._isInitializingSurveys = false
+                            return
+                        }
+
+                        this._surveyManager = phExtensions.generateSurveys?.(this.instance)
+                        this._isInitializingSurveys = false
+                        if (this._surveyEventReceiver == null) {
+                            this._surveyEventReceiver = new SurveyEventReceiver(this.instance)
+                        }
+                        logger.info('Surveys loaded successfully')
```
marandaneto, line 250: could it be that `phExtensions.generateSurveys` returns null? if so, we have to check if `_surveyManager` has a value and then log its successfully loaded or not

src/posthog-surveys.ts
```diff
@@ -214,34 +221,50 @@ export class PostHogSurveys {
...
+                            logger.error('Could not load surveys script', err)
+                            this._isInitializingSurveys = false
+                            return
+                        }
+
+                        this._surveyManager = phExtensions.generateSurveys?.(this.instance)
+                        this._isInitializingSurveys = false
+                        if (this._surveyEventReceiver == null) {
+                            this._surveyEventReceiver = new SurveyEventReceiver(this.instance)
+                        }
```
marandaneto, line 249: do we actually need this guard? the whole control flow now is protected, will this method ever be called again if `_surveyManager` was created and consequently `_surveyEventReceiver`?

src/posthog-surveys.ts
```diff
@@ -214,34 +221,50 @@ export class PostHogSurveys {
...
+                        logger.info('Surveys loaded successfully')
+                    })
+                } else {
+                    logger.error('PostHog loadExternalDependency extension not found. Cannot load remote config.')
+                    this._isInitializingSurveys = false
+                }
             } else {
-                logger.error('PostHog loadExternalDependency extension not found. Cannot load remote config.')
+                this._surveyManager = generateSurveys(this.instance)
```
marandaneto, line 257: same as https://github.com/PostHog/posthog-js/pull/1747/files#r1964979053

src/posthog-surveys.ts
lucasheriques: this is executed after the callback that should load it into phExtensions.

but I agree we should be more defensive here - so I [added an additional check](https://github.com/PostHog/posthog-js/pull/1747/commits/899bff9c3acc6930a313cbb22005be7bc8794477), before seeing if there's an error, to see if this value is undefined.

src/posthog-surveys.ts
lucasheriques: good point, no need for it anymore. removed it

src/posthog-surveys.ts
lucasheriques: for this one, it actually guarantees that it's not undefined. since we first check if `generateSurveys` is undefined on [this if](https://github.com/PostHog/posthog-js/pull/1747/files#diff-e2c6b58ed91904007d3de10b7836a7dc30e4239c8e4d88ccd9f705e061920e88R232-R234). so if it is, we guarantee that it has a value

src/posthog-surveys.ts
```diff
@@ -214,34 +221,46 @@ export class PostHogSurveys {
...
+
+                if (loadExternalDependency) {
+                    loadExternalDependency(this.instance, 'surveys', (err) => {
+                        if (err || !phExtensions.generateSurveys) {
+                            logger.error('Could not load surveys script', err)
+                            this._isInitializingSurveys = false
+                            return
+                        }
+
+                        this._surveyManager = phExtensions.generateSurveys(this.instance)
```
marandaneto, line 245: you still have to check the return of this function, the signature of `generateSurveys` is `generateSurveys?: (posthog: PostHog) => any | undefined` so it can return undefined and it will be wrongly set to "initialized" but its actually not

marandaneto (approved):

PR #1749, opened 2025-02-19
chore: Make `@rrweb/types` a peer dependency

### LINKED ISSUE #1662: rrweb dependency issue
Version 1.207.1 seems to have an issue with rrweb. In the [version release notes](https://github.com/PostHog/posthog-js/releases/tag/v1.207.1) it says updated to 1.8 alpha so maybe there is an issue with that version.

Image

rafaeelaudibert:
Some users aren't willing to turn `skipLibCheck` off which causes errors with us importing types from `@rrweb/types` but making that a developer dependency. In some other cases, namely, those using the Angular compiler, they actually can't turn this off, which is blocking their use of newer posthog-js versions

We can fix that in two ways:
- Let's add `@rrweb/types` as a peer dependency, this way users will know that they need to install the library themselves to get types to work
- Let's stop using `@rrweb/record` types to prevent people from having to install it. This increases our maintenance burden because we need to keep the types ourselves, but it implies they won't need to bundle a heavy library if they don't need it.

---

> How did you check this work?

I checked the new `module.d.ts` file, it doesn't include `@rrweb/record` anymore, just `@rrweb/types` which is just TS definitions so it shouldn't matter for bundle size

Closes #1662

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,6 +114,9 @@
         "typescript": "^5.5.4",
         "yargs": "^17.7.2"
     },
+    "peerDependencies": {
+        "@rrweb/types": "2.0.0-alpha.17"
+    },
     "lint-staged": {
         "*.{ts,tsx,js,json}": "prettier --write",
         "*.js": "eslint --cache --fix",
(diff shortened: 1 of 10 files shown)


rafaeelaudibert:
I've confirmed with the customer that this will partly solve #1662. They simply need to add `@rrweb/types` as a devDependency and they're all set!

pauldambra (approved):
CI is green, so I'm happy

PR #1750, opened 2025-02-19
fix: Remove circular dependency

rafaeelaudibert:
When bundling our SDK we were warned about circular dependency errors, because:

`src/utils/type-utils` depends on `src/utils/string-utils`
`src/utils/string-utils` depends on `src/utils`
`src/utils` depends on `src/string-utils`

We can remove the dependency by making `string-utils` self-contained copying the function it was importing from the index. AND also (follow-up below) by extracting this function to a new `regex-utils`

Turns out that function isn't being used anywhere else anyway.

### DIFF
--- a/src/utils/string-utils.ts
+++ b/src/utils/string-utils.ts
@@ -1,4 +1,11 @@
-import { isValidRegex } from '.'
+export const isValidRegex = function (str: string): boolean {
+    try {
+        new RegExp(str)
+    } catch {
+        return false
+    }
+    return true
+}
 
 export function includes<T = any>(str: T[] | string, needle: T): boolean {
     return (str as any).indexOf(needle) !== -1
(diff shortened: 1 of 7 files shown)


marandaneto (approved):

src/utils/string-utils.ts
```diff
@@ -1,4 +1,11 @@
-import { isValidRegex } from '.'
+export const isValidRegex = function (str: string): boolean {
```
marandaneto, line 1: well the file is called string-utils and this is more a general utils or regex utils, so semantically, it does not belong to this file but if it fixes a circular dep, all good from my side

marandaneto (approved):

src/utils/string-utils.ts
marandaneto: true only `isMatchingRegex` is being used directly.
and I am the one to blame apparently https://github.com/PostHog/posthog-js/pull/1698#discussion_r1934416531 lol

rafaeelaudibert: I'm happy to move them both to `regex-utils`, even though it's only used here. It'll make it more semantic :).

PR #1752, opened 2025-02-19
feat: allow widget surveys to be repeated indefinitely

lucasheriques:
## Changes

add a new check for surveys repeating: if theyir type is `Widget` and the schedule is `Always`

## Checklist


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -645,6 +652,10 @@ export const hasEvents = (survey: Survey): boolean => {
 }
 
 export const canActivateRepeatedly = (survey: Survey): boolean => {
+    if (survey.schedule === SurveySchedule.Always && survey.type === SurveyType.Widget) {
```
marandaneto, line 655: Can we add tests for this? ideally all the new code that is testable, we'd add tests to cover more of the surveys codebase

lucasheriques: @marandaneto i agree, just added to this function and also to the one it depends too

marandaneto (approved):

PR #1758, opened 2025-02-22
feat(feature-flags): support quota limiting for feature flags

dmarticus:
with https://github.com/PostHog/posthog/pull/28564, we now start to respond different with the /decide and /local_evaluation APIs if users have gone over their quota limit. Now we need to change the SDKs to handle these new responses.

### DIFF
--- a/src/__tests__/posthog-core.loaded.test.ts
+++ b/src/__tests__/posthog-core.loaded.test.ts
@@ -106,4 +106,78 @@ describe('loaded() with flags', () => {
             expect(instance._send_request).toHaveBeenCalledTimes(1)
         })
     })
+
+    describe('quota limiting', () => {
+        let mockLogger: jest.SpyInstance
+
+        beforeEach(async () => {
+            mockLogger = jest.spyOn(console, 'warn').mockImplementation()
+            instance = await createPosthog()
+        })
+
+        afterEach(() => {
+            mockLogger.mockRestore()
+        })
+
+        it('does not process feature flags when quota limited', async () => {
+            instance._send_request = jest.fn(({ callback }) =>
+                callback?.({
+                    statusCode: 200,
+                    json: {
+                        quotaLimited: ['feature_flags'],
+                        featureFlags: { 'test-flag': true },
+                    },
+                })
+            )
+
+            const receivedFeatureFlagsSpy = jest.spyOn(instance.featureFlags, 'receivedFeatureFlags')
+
+            instance.featureFlags._callDecideEndpoint()
+            jest.runAllTimers()
+
+            expect(receivedFeatureFlagsSpy).not.toHaveBeenCalled()
+        })
+
+        it('processes feature flags when not quota limited', async () => {
+            const mockFlags = { 'test-flag': true }
+            instance._send_request = jest.fn(({ callback }) =>
+                callback?.({
+                    statusCode: 200,
+                    json: {
+                        featureFlags: mockFlags,
+                    },
+                })
+            )
+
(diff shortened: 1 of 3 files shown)


src/__tests__/posthog-core.loaded.test.ts
```diff
@@ -106,4 +106,78 @@ describe('loaded() with flags', () => {
...
+
+        it('does not process feature flags when quota limited', async () => {
+            instance._send_request = jest.fn(({ callback }) =>
+                callback?.({
+                    statusCode: 200,
+                    json: {
+                        quotaLimited: ['feature_flags'],
+                        featureFlags: { 'test-flag': true },
+                    },
+                })
```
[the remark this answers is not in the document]
dmarticus: wrong – everything should be `statusCode`, actually.

src/__tests__/posthog-core.loaded.test.ts
```diff
@@ -15,7 +15,7 @@ describe('loaded() with flags', () => {
             ...config,
             loaded: (ph) => {
                 ph.capture = jest.fn()
-                ph._send_request = jest.fn(({ callback }) => callback?.({ status: 200, json: {} }))
+                ph._send_request = jest.fn(({ callback }) => callback?.({ statusCode: 200, json: {} }))
```
dmarticus, line 18: had to fix a squiggly red line

src/posthog-featureflags.ts
```diff
@@ -302,6 +307,15 @@ export class PostHogFeatureFlags {
                 }
 
                 this._flagsLoadedFromRemote = !errorsLoading
+
+                if (response.json && response.json.quotaLimited?.includes(QuotaLimitedResource.FeatureFlags)) {
+                    // log a warning and then early return
+                    logger.warn(
+                        'You have hit your feature flags quota limit, and will not be able to load feature flags until the quota is reset.  Please visit https://posthog.com/docs/billing/limits-alerts to learn more.'
```
dmarticus, line 314: This warning feels a bit too long – what do folks think?

src/posthog-featureflags.ts
```diff
@@ -302,6 +307,15 @@ export class PostHogFeatureFlags {
                 }
 
                 this._flagsLoadedFromRemote = !errorsLoading
```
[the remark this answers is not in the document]
dmarticus: yeah, but the variable name `_flagsLoadedFromRemote` is a bit misleading too – it implies that we're _trying_ to use the flags from the `/decide` service, not that we actually end up using them.

src/__tests__/posthog-core.loaded.test.ts
haacked: Squiggly red lines are the worst!

src/posthog-featureflags.ts
haacked: It seems to be a lot longer than the other libraries. EIther we should go more detailed in all of them (with the URL), or keep them all short and make sure searching PostHog will find that page easily.

haacked (approved):
:shipit: Looks good to me.

src/posthog-featureflags.ts
dmarticus: Yeah, I have a docs PR for this here: https://github.com/PostHog/posthog.com/pull/10653, but that's mostly for the API and it doesn't include information on resolution.  It feels appropriate to put that in the logs for the SDKs.  I think I'll do this everywhere, actually – if it were me using the SDK I'd be please to see a link to how to resolve a warning log in the log itself.

haacked: Agreed!

PR #1761, opened 2025-02-23
fix: dont show survey on url change, only hide it

lucasheriques:
## Changes

showing a survey should be the responsibility only of `showSurvey`, we shouldn't really call it anywhere else

## Checklist

### DIFF
--- a/src/__tests__/extensions/surveys.test.ts
+++ b/src/__tests__/extensions/surveys.test.ts
@@ -5,8 +5,8 @@ import {
     renderFeedbackWidgetPreview,
     renderSurveysPreview,
     SurveyManager,
+    useHideSurveyOnURLChange,
     usePopupVisibility,
-    useToggleSurveyOnURLChange,
 } from '../../extensions/surveys'
 import { createShadow } from '../../extensions/surveys/surveys-utils'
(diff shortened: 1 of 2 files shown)


src/__tests__/extensions/surveys.test.ts
```diff
@@ -1043,67 +1043,6 @@ describe('useToggleSurveyOnURLChange', () => {
         expect(mockSetSurveyVisible).not.toHaveBeenCalled()
     })
 
-    it('should show survey when URL changes to match conditions', () => {
```
marandaneto, line 1046: if there was a test for it, I wonder if there was a reason for it and now we are changing it?

marandaneto: Is there an issue caused by this code path so I can get better context?

lucasheriques: this was a test i added on my [last PR](https://github.com/PostHog/posthog-js/pull/1755), however, i was doing some tests locally, and this is causing delayed surveys to overcome the delay and also skip the `survey sent` event, since thats handled by the `showSurvey()` method

marandaneto (approved):

PR #1764, opened 2025-02-25
chore: use quesiton ID instead of question index if it's available

lucasheriques:
## Changes

send question events using question id instead of index

depends on https://github.com/PostHog/posthog/pull/29226 to be merged first

## Checklist


marandaneto:
ps: update docs https://posthog.com/docs/surveys/implementing-custom-surveys

marandaneto:
I think this PR should clean some code around `currentQuestionIndex`, `originalQuestionIndex` and `displayQuestionIndex`.
we only need the question index right now for question sequence, but not for keeping track of shuffled questions.

lucasheriques:
@marandaneto i think we should actually leave those changes to another PR - especially around `displayIndex`. because those still matter for preview surveys before creating them (question's don't have an `id` on this point).

while we can generate the id in the frontend, it's gonna take even more work to fix the preview as that currently relies on indexes too, and the logic is built around that. it just doesn't matter for actual events anymore.

we can't make changes here until we also make them on the main repo. so maybe we can fix that later? what do you think?

marandaneto:
> @marandaneto i think we should actually leave those changes to another PR - especially around `displayIndex`. because those still matter for preview surveys before creating them (question's don't have an `id` on this point).
> 
> while we can generate the id in the frontend, it's gonna take even more work to fix the preview as that currently relies on indexes too, and the logic is built around that. it just doesn't matter for actual events anymore.
> 
> we can't make changes here until we also make them on the main repo. so maybe we can fix that later? what do you think?

that is fine, it can be a separate PR, but I recall all the confusion that `originalQuestionIndex` brought, and this is an SDK field only and is leaking to clickhouse events, so I'd definitely follow up right after

marandaneto:
`currentQuestionIndex` and `displayQuestionIndex` is fine to stay since you need them to control the next question flow, only `originalQuestionIndex` is unnecessary now, right? that's what I meant actually.

lucasheriques:
@marandaneto agree on `originalQuestionIndex`, [removed all references from it](https://github.com/PostHog/posthog-js/pull/1764/commits/e181e4129786d15c40bd5c413155b24882f47448)

src/extensions/surveys.tsx
```diff
@@ -759,7 +759,7 @@ export function Questions({
         survey.appearance?.backgroundColor || defaultSurveyAppearance.backgroundColor
     )
     const [questionsResponses, setQuestionsResponses] = useState({})
-    const { isPreviewMode, previewPageIndex, onPopupSurveyDismissed, isPopup, onPreviewSubmit, onPopupSurveySent } =
```
marandaneto, line 762: don we need the `isPreviewMode` anymore?

marandaneto (approved):
left a last comment otherwise LGTM

src/extensions/surveys.tsx
lucasheriques: not on this part anymore, only to prevent the submission now (another component)

marandaneto:
> @marandaneto agree on `originalQuestionIndex`, [removed all references from it](https://github.com/PostHog/posthog-js/pull/1764/commits/e181e4129786d15c40bd5c413155b24882f47448)

I still see `originalQuestionIndex` in 2 cases including `SurveyQuestionBase` which is the one that will leak to the database, right? can we somehow move this out from the model.

PR #1767, opened 2025-02-27
fix: dont require double click/page reload to show feedback surveys

lucasheriques:
## Changes

adds a new option on SurveyContext to do something after a survey is submitted.

also, right now, for widget-type surveys (tab style), we have to click on the button twice to show the survey again after dismissing it. this PR fixes that

and, for selector, it only opens again if you do a full page reload

## Checklist

### DIFF
--- a/src/__tests__/extensions/surveys.test.ts
+++ b/src/__tests__/extensions/surveys.test.ts
@@ -1088,7 +1088,7 @@ describe('onSurveyDismissedOrSent callback', () => {
         } as unknown as PostHog
     })
 
-    test('onSurveyDismissedOrSent is called when Cancel button is clicked in SurveyPopup', () => {
+    test('onPopupSurveyDismissed is called when Cancel button is clicked in SurveyPopup', () => {
         // Create a simple survey for testing
         const survey = {
             id: 'test-survey',
@@ -1127,7 +1127,7 @@ describe('onSurveyDismissedOrSent callback', () => {
                 survey: survey,
                 removeSurveyFromFocus: mockRemoveSurveyFromFocus,
                 isPopup: true,
-                onSurveyDismissedOrSent: mockOnSurveyDismissedOrSent,
+                onPopupSurveyDismissed: mockOnSurveyDismissedOrSent,
                 posthog: posthog,
             })
         )
--- a/src/__tests__/extensions/surveys.test.ts
+++ b/src/__tests__/extensions/surveys.test.ts
@@ -1,10 +1,11 @@
 /* eslint-disable compat/compat */
 import { act, fireEvent, render, renderHook } from '@testing-library/preact'
 import {
+    SurveyManager,
+    SurveyPopup,
     generateSurveys,
     renderFeedbackWidgetPreview,
     renderSurveysPreview,
-    SurveyManager,
     useHideSurveyOnURLChange,
     usePopupVisibility,
 } from '../../extensions/surveys'
@@ -613,6 +614,7 @@ describe('usePopupVisibility URL changes should hide surveys accordingly', () =>
             feature_flag_keys: null,
             linked_flag_key: null,
             targeting_flag_key: null,
+            internal_targeting_flag_key: null,
             appearance: {},
         }) as Survey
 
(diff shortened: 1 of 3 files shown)


ioannisj (approved):
LG. Just a suggestion/observation - We already have `handleCloseSurveyPopup` can we use that for dismissing and just add a new `onSurveySent`?

src/extensions/surveys.tsx
```diff
@@ -655,6 +655,7 @@ interface SurveyPopupProps {
     removeSurveyFromFocus: (id: string) => void
     isPopup?: boolean
     onPreviewSubmit?: (res: string | string[] | number | null) => void
+    onSurveyDismissedOrSent?: () => void
```
marandaneto, line 658: I think we'd benefit more from 2 callbacks.
one for dismissed and one for sent.
sent (show something with a voucher, etc), dismissed (show something else, do nothing).

src/extensions/surveys.tsx
```diff
@@ -768,6 +777,7 @@ export function Questions({
         const nextStep = getNextSurveyStep(survey, displayQuestionIndex, res)
         if (nextStep === SurveyQuestionBranchingType.End) {
             sendSurveyEvent({ ...questionsResponses, [responseKey]: res }, survey, posthog)
+            onSurveyDismissedOrSent()
```
marandaneto, line 780: curious why we call this function here but not `handleCloseSurveyPopup`?

marandaneto:
> LG. Just a suggestion/observation - We already have `handleCloseSurveyPopup` can we use that for dismissing and just add a new `onSurveySent`?

yeah I wonder if `handleCloseSurveyPopup` is called for dismissed/sent, just one of them? which one? are we duplicating something here?

src/extensions/surveys.tsx
lucasheriques: `handleCloseSurveyPopup` is for when pressing the X button (dismissing the survey)

sending the survey is another action than closing/dismissing it

lucasheriques:
@marandaneto @ioannisj 

- `handleCloseSurveyPopup` is called for dismissing the survey only (`survey dismissed`)

- [onNextButtonClick](https://github.com/PostHog/posthog-js/blob/fix/dont-require-double-click-to-show-feedback-survey/src/extensions/surveys.tsx#L758) is actually the one that is called for `survey sent`, but it's defined down in the component tree.

I agree this is probably confusing, and there's a refactor somewhere we can do to make the logic be more sound

but it's not on the scope of this PR specifically - where I'm mainly looking for adding a callback that will be triggered on both cases (dismissing or sending the survey)

marandaneto:
> @marandaneto @ioannisj
> 
> * `handleCloseSurveyPopup` is called for dismissing the survey only (`survey dismissed`)
> * [onNextButtonClick](https://github.com/PostHog/posthog-js/blob/fix/dont-require-double-click-to-show-feedback-survey/src/extensions/surveys.tsx#L758) is actually the one that is called for `survey sent`, but it's defined down in the component tree.
> 
> I agree this is probably confusing, and there's a refactor somewhere we can do to make the logic be more sound
> 
> but it's not on the scope of this PR specifically - where I'm mainly looking for adding a callback that will be triggered on both cases (dismissing or sending the survey)

I think the confusion here is that there are one action and 2 callback options
`handleCloseSurveyPopup` -> dismissing
`onSurveyDismissedOrSent` -> dismissing and sending

I'd rather do
`handleCloseSurveyPopup` which exists
`handleSentSurveyPopup` sending
whenever needed, call both functions (if you need both actions), how does that sound?

src/extensions/surveys.tsx
marandaneto: ok, i thought people could use as part of this https://posthog.com/docs/surveys/implementing-custom-surveys

marandaneto (approved):
left a last question [here](https://github.com/PostHog/posthog-js/pull/1767#issuecomment-2688166999)
i might be misunderstanding the behaviour of both callbacks, so feel free to explain and merge it if that's the case

src/extensions/surveys.tsx
```diff
@@ -736,8 +746,14 @@ export function Questions({
         survey.appearance?.backgroundColor || defaultSurveyAppearance.backgroundColor
     )
     const [questionsResponses, setQuestionsResponses] = useState({})
-    const { isPreviewMode, previewPageIndex, handleCloseSurveyPopup, isPopup, onPreviewSubmit } =
-        useContext(SurveyContext)
+    const {
+        isPreviewMode,
+        previewPageIndex,
+        onPopupSurveyDismissed: handleCloseSurveyPopup,
```
marandaneto, line 752: isnt `handleCloseSurveyPopup` renamed?

src/extensions/surveys.tsx
```diff
@@ -806,7 +823,14 @@ export function Questions({
                                     : {}
                             }
                         >
-                            {isPopup && <Cancel onClick={() => handleCloseSurveyPopup()} />}
+                            {isPopup && (
+                                <Cancel
+                                    onClick={() => {
+                                        handleCloseSurveyPopup()
```
marandaneto, line 829: https://github.com/PostHog/posthog-js/pull/1767/files#r1973794231

src/extensions/surveys.tsx
```diff
@@ -736,8 +746,14 @@ export function Questions({
...
     const [questionsResponses, setQuestionsResponses] = useState({})
-    const { isPreviewMode, previewPageIndex, handleCloseSurveyPopup, isPopup, onPreviewSubmit } =
-        useContext(SurveyContext)
+    const {
+        isPreviewMode,
+        previewPageIndex,
+        onPopupSurveyDismissed: handleCloseSurveyPopup,
+        isPopup,
+        onPreviewSubmit,
+        onPopupSurveySent: onSurveyDismissedOrSent,
```
marandaneto, line 755: isnt `onSurveyDismissedOrSent` renamed?

src/__tests__/extensions/surveys.test.ts
```diff
@@ -1071,6 +1073,81 @@ describe('useHideSurveyOnURLChange', () => {
     })
 })
 
+describe('onSurveyDismissedOrSent callback', () => {
```
marandaneto, line 1076: tests are using the old callback names

PR #1768, opened 2025-02-27
feat: no recording blob urls

pauldambra:
I noticed this while investigating something else... we don't need to do any parsing of blob/object URL bodies in the network recorder. we simply shouldn't be attempting to record their body


src/entrypoints/recorder.ts
```diff
@@ -122,19 +122,43 @@ function shouldRecordBody({
...
     }
 
+    /**
+     * particularly in canvas applications we see many requests to blob URLs
+     * e.g. blob:https://video_url
+     * these blob/object URLs are local to the browser, we can never capture that body
+     * so we can just return false here
+     */
+    function isBlobURL(url: string | URL | RequestInfo) {
+        try {
```
veryayskiy, line 145: Not sure why do you need try catch here though

veryayskiy (approved):

src/entrypoints/recorder.ts
pauldambra: i am very extra super careful in this file
if we break things in here we can break fetch/xhr in customer files

PR #1770, opened 2025-02-28
chore: add posthog built rrweb

pauldambra:
Adds the option of loading the posthog built rrweb instead of rrweb one

Doesn't use it anywhere. But will make it available for testing


pauldambra:
you can see here why we can't switch everyone over quite yet  to the unpatched recorder

image

veryayskiy (approved):
Small rebellion

PR #1771, opened 2025-02-28
feat: capture console errors

daibhin:
## Changes

Need to test this locally because some errors are posted to the console so I don't want to double track exceptions


src/entrypoints/exception-autocapture.ts
```diff
@@ -46,9 +46,36 @@ const wrapUnhandledRejection = (captureFn: (props: Properties) => void) => {
...
+    const originalConsoleError = con.error
+
+    con.error = function (...args: any[]): boolean {
+        const event = args.join(' ')
+        const error = args.find((arg) => arg instanceof Error)
+        const errorProperties = error
+            ? errorToProperties({ event, error })
+            : errorToProperties({ event }, { syntheticException: new Error('PostHog syntheticException') })
+
+        captureFn(errorProperties)
```
pauldambra, line 64: might be cool to indicate if replay and console capture is running
to help with linking between the two

but i guess we have 99% of that anyway

kdawgwilk:
This would make the error tracking for me much more valuable, would love to see this

src/entrypoints/exception-autocapture.ts
daibhin: I did try to consolidate the two but the replay one looks to be within rrweb :/

oliverb123 (approved):

nihar-jb:
Is this PR live in posthog? I tried to throw console error manually but posthog is not capturing those errors.

PR #1772, opened 2025-03-01
feat: better positioning for survey selectors

lucasheriques:
# Make surveys appear next to buttons instead of in the corner

This PR changes where surveys show up when triggered from buttons on the page, making them easier to use.

## What changed

- Surveys now pop up right next to the button you clicked instead of always appearing in the bottom-right corner
- Added a consistent small gap (12px) between the button and survey
- Made sure surveys work well on small screens by keeping them fully visible
- Fixed the X button display issue by removing some unnecessary styling

## Before & After

Before: Surveys would show up in the bottom-right corner, far away from the button you clicked. (like for all other current surveys)

After: Surveys now appear right next to the button you clicked, making it clear what triggered the survey.

| Survey on top | Survey on bottom |
|--------|--------|
| CleanShot 2025-02-28 at 22 44 42@2x | CleanShot 2025-02-28 at 22 45 29@2x | 

## Testing

I tested this on phones, tablets and desktops to make sure it looks good everywhere. The survey now properly appears above the button when there's not enough room below it.

## Why no automated tests?

We went with manual testing because:

1. This is mostly about how things look on screen, which is hard to test with code, especially when we don't have an easy way atm to mock surveys (no other tests do this)
2. The positioning depends on screen size and button location - things that are tricky to simulate in tests
3. The best way to check if it works was to actually try it on different devices and screen sizes
4. The changes don't affect how the surveys work, just where they appear

## Checklist

### DIFF
--- a/playground/nextjs/pages/survey.tsx
+++ b/playground/nextjs/pages/survey.tsx
@@ -1,6 +1,6 @@
+import type { Survey } from 'posthog-js'
 import { usePostHog } from 'posthog-js/react'
 import { useEffect, useState } from 'react'
-import type { Survey } from 'posthog-js'
 
 export default function SurveyForm() {
     const posthog = usePostHog()
@@ -29,6 +29,7 @@ export default function SurveyForm() {
     return (
         <>
             <div className="flex items-center gap-2 flex-wrap">
+                <button className="survey-feedback-button">Feedback</button>
                 <select value={selectedSurvey} onChange={handleChange}>
                     <option value="">Select a survey</option>
                     {arraySurveyItems}
@@ -44,6 +45,28 @@ export default function SurveyForm() {
                 <div id="survey-container">
                     <h1>hello world</h1>
                 </div>
+            </div>
+
+            {/* Add spacer to push the bottom button down */}
+            <div style={{ height: '70vh', display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+                <p>Scroll down to see the bottom feedback button</p>
+            </div>
+
+            {/* Bottom feedback button */}
+            <div style={{ padding: '20px', textAlign: 'center', borderTop: '1px solid #eee' }}>
+                <button
+                    className="survey-feedback-button"
+                    style={{
+                        padding: '10px 20px',
+                        background: '#1D4AFF',
+                        color: 'white',
+                        borderRadius: '4px',
+                        border: 'none',
+                        cursor: 'pointer',
+                    }}
+                >
+                    Feedback (Bottom)
+                </button>
             </div>
         </>
     )
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -35,6 +35,7 @@ import {
     hasWaitPeriodPassed,
     sendSurveyEvent,
     style,
+    SURVEY_DEFAULT_Z_INDEX,
     SurveyContext,
 } from './surveys/surveys-utils'
 import { prepareStylesheet } from './utils/stylesheet-loader'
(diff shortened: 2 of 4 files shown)


playground/nextjs/pages/survey.tsx
```diff
@@ -45,6 +46,28 @@ export default function SurveyForm() {
...
+                        background: '#1D4AFF',
+                        color: 'white',
+                        borderRadius: '4px',
+                        border: 'none',
+                        cursor: 'pointer',
+                    }}
+                >
+                    Feedback (Bottom)
+                </button>
+            </div>
```
lucasheriques, line 70: no need to pay attention to this code really - it's just a button I added to the nextjs playground

src/extensions/surveys.tsx
```diff
@@ -888,7 +896,73 @@ export function FeedbackWidget({
...
+                    left = 20 // 20px padding from left edge
+                }
+
+                // Determine if we should show above or below
+                let showAbove = false
+
+                // Check if there's enough space below (need at least 300px)
+                // If not enough space below, show above
+                if (buttonRect.bottom + 300 > viewportHeight) {
+                    showAbove = true
```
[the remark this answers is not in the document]
lucasheriques: survey max height is actually `300` so this doesn't apply

playground/nextjs/pages/survey.tsx
```diff
@@ -29,6 +29,7 @@ export default function SurveyForm() {
     return (
         <>
             <div className="flex items-center gap-2 flex-wrap">
+                <button className="survey-feedback-button">Feedback</button>
```
marandaneto, line 32: can we move this to a const, we are repeatin

src/extensions/surveys.tsx
```diff
@@ -604,6 +604,14 @@ export function usePopupVisibility(
                 sessionRecordingUrl: posthog.get_session_replay_url?.(),
             })
             localStorage.setItem('lastSeenSurveyDate', new Date().toISOString())
+            setTimeout(() => {
+                const inputField = document
+                    .querySelector('.PostHogWidget' + survey.id)
```
marandaneto, line 609: lets move PostHogWidget to a const, we are repeating many times over

src/extensions/surveys/surveys-utils.tsx
```diff
@@ -48,7 +48,7 @@ export const style = (appearance: SurveyAppearance | null) => {
               text-align: left;
               max-width: ${parseInt(appearance?.maxWidth || '300')}px;
               width: 100%;
-              z-index: ${parseInt(appearance?.zIndex || '99999')};
+              z-index: ${parseInt(appearance?.zIndex || '2147482647')};
```
marandaneto, line 51: lets change it here too https://github.com/PostHog/posthog-js/blob/92ee6c5bad352b8879f44749b7f96a6556ff4a1a/src/extensions/surveys-widget.ts#L40

marandaneto: the line above should be using `appearance?.zIndex` as well I guess

lucasheriques: not really as those are the settings for the widget tab, not the survey customization, and we currently don't allow appearance changes for the tab. we probably can, but for now i'll just leave as const

playground/nextjs/pages/survey.tsx
lucasheriques: this is just on our internal nextjs playground, imo not needed as it's just the class i used for my survey. I left the code just because it might be helpful to other people in the future

src/extensions/surveys.tsx
lucasheriques: moved to a function as I think it's more clear [5743518](https://github.com/PostHog/posthog-js/pull/1772/commits/57435189533c9b68840ead8ed6bddfa3cd2b0b05)

src/extensions/surveys/surveys-utils.tsx
lucasheriques: also made the zindex default value a const

playground/nextjs/pages/survey.tsx
marandaneto: true that, my bad, i just did a search and saw the same className across places

marandaneto (approved):

PR #1774, opened 2025-03-03
chore: refactor playwright helper for clarity

pauldambra:
this made sense to me when I first wrote it, but then i tried to explain it to @joshsny and thought it did the opposite of what it actually did

a little fiddling to hopefully make it more clear to the future traveller


joshsny (approved):
😇

PR #1778, opened 2025-03-03
feat: Make `getQueryParams` more resilient

rafaeelaudibert:
An SDK user has a slightly broken URL where they have two leading question marks, which is causing our `queryParam` parsing to break. Right now, because we split on `?` and get the second split, we're simply getting an empty string.

We're fixing this in two ways:
- where possible use built-in browser features to parse query params (with fallback to IE11 and op_mini)
- trim leading question marks to avoid this from happening again

See https://posthoghelp.zendesk.com/agent/tickets/25469 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/28561)

### DIFF
--- a/src/utils/request-utils.ts
+++ b/src/utils/request-utils.ts
@@ -41,11 +41,16 @@ export const formDataToQuery = function (formdata: Record<string, any> | FormDat
     return tph_arr.join(arg_separator)
 }
 
+// NOTE: Once we get rid of IE11/op_mini we can start using URLSearchParams
 export const getQueryParam = function (url: string, param: string): string {
     const withoutHash: string = url.split('#')[0] || ''
-    const queryParams: string = withoutHash.split('?')[1] || ''
 
-    const queryParts = queryParams.split('&')
+    // Split only on the first ? to sort problem out for those with multiple ?s
+    // and then remove them
+    const queryParams: string = withoutHash.split(/\?(.*)/)[1] || ''
+    const cleanedQueryParams = queryParams.replace(/^\?+/g, '')
+
+    const queryParts = cleanedQueryParams.split('&')
     let keyValuePair
 
     for (let i = 0; i < queryParts.length; i++) {
--- a/src/__tests__/request-utils.test.ts
+++ b/src/__tests__/request-utils.test.ts
@@ -44,6 +44,8 @@ describe('request utils', () => {
             ['gets param when no match with trailing slash', '/', 'name', ''],
             ['gets param when no match and there are params', '/?test=123', 'name', ''],
             ['gets param when no match and there are params with trailing slash', '/?test=123', 'name', ''],
+            ['gets param when we have duplicate question marks', '??test=123', 'test', '123'],
+            ['gets trailing param when we have duplicate question marks', '??test=123&name=john', 'name', 'john'],
         ])('%s', (_name, url, param, expected) => {
             expect(getQueryParam(`https://example.com${url}`, param)).toEqual(expected)
         })


src/utils/request-utils.ts
```diff
@@ -42,8 +42,31 @@ export const formDataToQuery = function (formdata: Record<string, any> | FormDat
...
+    const urlObj = new URL(url) // eslint-disable-line compat/compat
+
+    // Remove leading ? from search string if present
+    // This is not really necessary if there's a single ? in the search string
+    // but it's helpful if someone has some malformed URL with multiple ?s
+    const search = urlObj.search.replace(/^\?*/g, '')
+    const searchParams = new URLSearchParams(search) // eslint-disable-line compat/compat
+
+    return searchParams.get(param) ?? ''
+}
```
rafaeelaudibert, line 61: Open to feedback on whether this is needed. I think using the builtins are better than rolling our own code like we have on the fallback, but happy to be told otherwise and just stick to the previous thing we had going

Also, I'm not decoding the params anymore, so I might need to add that back 🤔

src/utils/request-utils.ts
```diff
@@ -42,8 +42,45 @@ export const formDataToQuery = function (formdata: Record<string, any> | FormDat
 }
 
 export const getQueryParam = function (url: string, param: string): string {
+    // Check if URL and URLSearchParams are available in the global scope
+    // Fall back to manual parsing if URL/URLSearchParams aren't supported
+    if (typeof URL === 'undefined' || typeof URLSearchParams === 'undefined') {
+        return getQueryParamFallback(url, param)
```
robbie-c, line 48: Is there any downside to just always calling that function? It's nicer to use the built-in parsing, and if we could guarantee that it was available, we'd just use that, but given we have to have a fallback anyway, why add extra bytes to the bundle with two implementations?

src/utils/request-utils.ts
```diff
@@ -42,8 +42,45 @@ export const formDataToQuery = function (formdata: Record<string, any> | FormDat
...
+}
+
+const getQueryParamFallback = function (url: string, param: string): string {
     const withoutHash: string = url.split('#')[0] || ''
-    const queryParams: string = withoutHash.split('?')[1] || ''
+
+    // By limiting the split to 2, we ensure that we only split on the first ?
+    // This is not really necessary if there's a single ? in the search string
+    // but it's helpful if someone has some malformed URL with multiple ?s
+    const queryParams: string = withoutHash.split('?', 2)[1] || ''
```
robbie-c, line 83: Do we want to remove a potential leading `?` after this?

src/utils/request-utils.ts
rafaeelaudibert: Just because I'd prefer to use the built-in if it exists, but I'm fine with keeping the old-but-proven-to-work function instead, and just fix it (which I did anyway)

src/utils/request-utils.ts
rafaeelaudibert: Yeah, sounds good, let me add that too

src/utils/request-utils.ts
rafaeelaudibert: Which is basically what I commented here: https://github.com/PostHog/posthog-js/pull/1778#discussion_r1977941224

rafaeelaudibert: Would you prefer we kept that old code instead and never add the new one?

src/utils/request-utils.ts
robbie-c: I'd rather only have 1 version. Bundle size is important. Sorry man =/ I'd love to kill off the old version and only have one that uses built-in methods, but if we need to support older browsers then we can't do that.

rafaeelaudibert: Not a problem, will revert

robbie-c (approved):

PR #1779, opened 2025-03-03
feat: Block Lighthouse cryptic user agent

rafaeelaudibert:
Believe it or not, this is Lighthouse's user agent. They don't usually include "lighthouse" in it anymore because some people were gaming the system for LH scores, so they use some cryptic UAs now. I'll just ignore this outright because not many people will actually have that UA, and this might cause some big bills for some customers.

https://github.com/GoogleChrome/lighthouse/blob/main/core/config/constants.js#L42

### DIFF
--- a/src/utils/blocked-uas.ts
+++ b/src/utils/blocked-uas.ts
@@ -17,7 +17,6 @@ export const DEFAULT_BLOCKED_UA_STRS = [
     'bot.php',
     '(bot;',
     'bot/',
-    'chrome-lighthouse',
     'crawler',
     'dataforseobot',
     'deepscan',
@@ -80,6 +79,13 @@ export const DEFAULT_BLOCKED_UA_STRS = [
     'mediapartners-google',
     'storebot-google',
     'bytespider',
+
+    // Believe it or not, these are all from Lighthouse
+    // We'll make them an exhaustive match to decrease false positives
+    // https://github.com/GoogleChrome/lighthouse/blob/main/core/config/constants.js
+    'chrome-lighthouse',
+    'mozilla/5.0 (linux; android 11; moto g power (2022)) applewebkit/537.36 (khtml, like gecko) chrome/119.0.0.0 mobile safari/537.36', // Mobile UA for Lighthouse
+    'mozilla/5.0 (macintosh; intel mac os x 10_15_7) applewebkit/537.36 (khtml, like gecko) chrome/119.0.0.0 safari/537.36', // Desktop UA for Lighthouse
 ]
 
 /**
--- a/src/__tests__/utils.test.ts
+++ b/src/__tests__/utils.test.ts
@@ -165,6 +165,12 @@ describe('utils', () => {
                 'Mozilla/5.0 (compatible; SeznamBot/4.0; +https://o-seznam.cz/napoveda/vyhledavani/en/seznambot-crawler/)',
             ],
             ['BrightEdge Crawler/1.0 (crawler@brightedge.com)'],
+            [
+                'Mozilla/5.0 (Linux; Android 11; moto g power (2022)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Mobile Safari/537.36',
+            ],
+            [
+                'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36',
+            ],
         ])('blocks based on user agent', (botString) => {
             expect(isBlockedUA(botString, [])).toBe(true)
             expect(isBlockedUA(botString.toLowerCase(), [])).toBe(true)


robbie-c (approved):
TIL, that's wild

src/utils/blocked-uas.ts
```diff
@@ -80,6 +79,12 @@ export const DEFAULT_BLOCKED_UA_STRS = [
     'mediapartners-google',
     'storebot-google',
     'bytespider',
+
+    // Believe it or not, these are all from Lighthouse
+    // https://github.com/GoogleChrome/lighthouse/blob/main/core/config/constants.js
+    'chrome-lighthouse',
+    'moto g power (2022)', // Mobile UA for Lighthouse: Mozilla/5.0 (Linux; Android 11; moto g power (2022)) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Mobile Safari/537.36
```
robbie-c, line 86: put these in the tests file?

src/utils/blocked-uas.ts
```diff
@@ -80,6 +79,12 @@ export const DEFAULT_BLOCKED_UA_STRS = [
...
+    'Intel Mac OS X 10_15_7', // Desktop UA for Lighthouse: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/119.0.0.0 Safari/537.36
```
robbie-c, line 87: oh, this should be lower case

robbie-c (changes requested):

robbie-c:
I did a bit more googling on this - I'm now not sure we should do it.

These are real user agents (afaik) and it's not impossible that we'd block real traffic.

I'm wondering if we need to take another approach. Where did it come up that these user agents were hitting a customer? I know that this one might be related https://posthoghelp.zendesk.com/agent/tickets/25267, (moved to PostHog: https://us.posthog.com/project/2/support/tickets/33353) where ahrefs uses lighthouse in one of its crawlers. I wonder if we could just ask ahrefs to identify itself somehow (and not through ip ranges, ideally)

rafaeelaudibert:
> I did a bit more googling on this - I'm now not sure we should do it.
> 
> These are real user agents (afaik) and it's not impossible that we'd block real traffic.

It is possible, yes, but I expect traffic from a specific user-agent to be negligible. That's why I updated the code to actually include the whole user agent rather than just a subset of it.


> I'm wondering if we need to take another approach. Where did it come up that these user agents were hitting a customer? I know that this one might be related [posthoghelp.zendesk.com/agent/tickets/25267](https://posthoghelp.zendesk.com/agent/tickets/25267) (moved to PostHog: https://us.posthog.com/project/2/support/tickets/33353), where ahrefs uses lighthouse in one of its crawlers. I wonder if we could just ask ahrefs to identify itself somehow (and not through ip ranges, ideally)

This is from https://posthoghelp.zendesk.com/agent/tickets/25646 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/28564) and it's possible they've enabled ahrefs or something else because it started on a specific day (February 3rd) and it's pretty consistent on how many accesses they have every day.

---

I feel confident on doing this if we go with the "include the whole user agent" approach, but I'm happy to just close this and let customers know about it if you disagree. They can always use `custom_uagent_blocklist` to block this themselves - which I told the customer to do already.

rafaeelaudibert:
> If we wanted, we could also include a flag to block these user agents too, I'm just not sure it should be the default.

Yeah, not sure. They already have the power to add specific user-agents they wanna block, I don't see the value on having a `block_lighthouse` flag. Maybe just improved documentation on the matter - listing the offending user-agents - would be enough.

> IMO it's pretty bad that a crawler hasn't identified itself.

100%, not very friendly


> I'll send this to [support@ahrefs.com](mailto:support@ahrefs.com), any thoughts:
> 
> Hi there,
> 
> I work for PostHog, and one of our mutual customers has asked for help blocking the ahrefs crawler from their analytics data.
> 
> You use lighthouse as part of your crawling, which by default sets the user agent to a realistic user agent (see [GoogleChrome/lighthouse@`main`/core/config/constants.js#L42](https://github.com/GoogleChrome/lighthouse/blob/main/core/config/constants.js?rgh-link-date=2025-03-04T14%3A04%3A28Z#L42)) which we aren't able to block. We've seen these user agents coming from your IP addresses (e.g. 5.39.1.230).
> 
> While we could block events coming from your IP addresses, it'd be very hard for us to do this client-side, where the IP address is not known, which would be an imperfect solution for some of our products.
> 
> Could you change the user agent, or at least provide your customers an option to change the user agent when crawling their own site?
> 
> Thanks, Robbie

Can you CC me? I like the email

The support ticket I'm working on in specific isn't coming from Ahrefs though, but it might be something similar, I've let them know

[177.244.11.68](https://ipinfo.io/ips/177.244.11.0/24)

rafaeelaudibert:
Making this a draft while we validate our solution with ahrefs

LuizHAP:
Thanks for helping this guys @rafaeelaudibert and @robbie-c. Meanwhile u guys thiking about it, I opened a ticket with Arcjet (https://app.arcjet.com/) and they will share with me what they thoughts around it, because maybe they suffered with the same issue

davidmytton:
Thanks for flagging this to us @LuizHAP.

All the known bot user agents we track are open source at https://github.com/arcjet/well-known-bots. In the case where the user agent is [one of those spoofed user agents](https://github.com/GoogleChrome/lighthouse/blob/cc206582e63ddc617635696473fd5d88f816a001/core/config/constants.js#L42), Arcjet will do an IP lookup against our reputation database. If it's from a known crawler like Ahrefs then we'll return a deny decision.

rafaeelaudibert:
> Thanks for flagging this to us @LuizHAP.
> 
> All the known bot user agents we track are open source at [arcjet/well-known-bots](https://github.com/arcjet/well-known-bots?rgh-link-date=2025-03-10T20%3A11%3A32Z). In the case where the user agent is [one of those spoofed user agents](https://github.com/GoogleChrome/lighthouse/blob/cc206582e63ddc617635696473fd5d88f816a001/core/config/constants.js#L42), Arcjet will do an IP lookup against our reputation database. If it's from a known crawler like Ahrefs then we'll return a deny decision.

Hey, @davidmytton, thank you for sharing that!

We're considering adding an IP lookup as a last resort. Ideally, we'd get the "bad" actors to improve their behavior but we both know that's not that easy. Besides the Lighthouse UAs, do you have something else on the list of known spoofed UAs?

davidmytton:
> We're considering adding an IP lookup as a last resort. Ideally, we'd get the "bad" actors to improve their behavior but we both know that's not that easy. Besides the Lighthouse UAs, do you have something else on the list of known spoofed UAs?

The UA matching is the first step. If there's no match then we use our private IP reputation database. We also do [verification for known providers](https://docs.arcjet.com/bot-protection/reference#bot-verification) e.g. we verify whether a request claiming to be Google is actually Google [through their rDNS lookup & IP ranges](https://developers.google.com/search/docs/crawling-indexing/verifying-googlebot). We do this for several providers who offer those lookup options so clients can't pretend to be good bots!

pauldambra:
@benjackwhite the arcjet tool above is interesting... too heavy to include in the SDK but could potentially offer a transformation that uses it to augment during ingestion

(dropping this into your brain, but obvs feel free to ignore 🤣)

PR #1780, opened 2025-03-03
feat(web-analytics): Add $session property with an object of props on session_id lifecycle

robbie-c:
## Changes

See https://posthog.slack.com/archives/C06GG249PR6/p1741000095036129 for the feature request

Adds $session_entry_X properties to events, which includes some properties that are scoped to the lifecycle of the session id

Screenshot 2025-03-04 at 09 11 04

I debated using a `$session` object and putting all the props in there, but I don't know how well materialisation would work on nested properties.

## Checklist


benjackwhite (approved):
Very cool

PR #1781, opened 2025-03-04
fix(flags): pass along $initial_ person properties in decide requests automatically

havenbarnes:
## Changes

Addresses https://posthoghelp.zendesk.com/agent/tickets/26377 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/28119)

When making initial decide requests, especially when the user is anonymous, PostHog users expect to be able to use feature flag conditions on internal initial props like $initial_pathname. This change automatically tacks them on to decide requests as intended, along with any custom person properties for flags:

Screenshot 2025-03-03 at 6 20 57 PM

Screenshot 2025-03-03 at 6 13 39 PM

## Checklist


havenbarnes:
@robbie-c would this cause issues with cookieless mode?

robbie-c:
They won't be available in cookieless mode, so if we wanted these properties we'd have to get them from the person profile instead.

My understanding is that this change is to support anonymous events? So that shouldn't be a problem.

(Initial properties won't work if you use both cookieless and anonymous events. Those initial properties need to be stored *somewhere*. It's fine for feature flags to have this same limitation)

haacked (approved):
:shipit: LGTM!

haacked:
@havenbarnes Just curious, why wasn't the version bumped? Will these changes be available automatically in the JS snippet   (https://posthog.com/docs/getting-started/install)?

havenbarnes:
> @havenbarnes Just curious, why wasn't the version bumped? Will these changes be available automatically in the JS snippet ([posthog.com/docs/getting-started/install](https://posthog.com/docs/getting-started/install))?

@haacked it should have been since I added the `bump minor` label.. 🤔

haacked:
@havenbarnes Oh, I saw this commit and got curious: https://github.com/PostHog/posthog-js/pull/1781/commits/62d81a2450b6ef526e585da0788f8f1398ea046a

I didn't know about the "bump minor" label.

PR #1783, opened 2025-03-04
feat: posthog feature component should report variant too

pauldambra:
Was confusing that this wasn't happening, so a user wanted to filter by variant but couldn't

So, adds reporting the variant to the `PosthogFeature` component

Screenshot 2025-03-04 at 15 22 02
Screenshot 2025-03-04 at 15 22 59


playground/nuxtjs/package.json
```diff
@@ -13,6 +13,7 @@
         "nuxt": "^3.4.3"
     },
     "dependencies": {
-        "posthog-js": "^1.55.1"
-    }
+        "posthog-js": "file:../../dist/"
+    },
+    "packageManager": "pnpm@9.15.4"
```
pauldambra, line 18: ran nuxt as well as next because typo
so fixed this while i was here 🤣

pauldambra:
associated with https://posthoghelp.zendesk.com/agent/tickets/25101 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/29550) (although doesn't resolve that issue i don't think)

react/src/components/PostHogFeature.tsx
```diff
@@ -46,12 +46,36 @@ export function PostHogFeature({
...
+}: {
+    flag: string
+    posthog: PostHog
+    flagVariant?: string | boolean
+}) {
+    posthog.capture('$feature_interaction', {
+        feature_flag: flag,
+        feature_flag_variant: typeof flagVariant === 'string' ? flagVariant : undefined,
+        $set: { [`$feature_interaction/${flag}`]: flagVariant ?? true },
+    })
```
rafaeelaudibert, line 62: We're sending `undefined` now, is this ok? If you wanted to prevent sending that altogether you could do the following

```suggestion
    const properties = { feature_flag: flag, $set: { [`$feature_interaction/${flag}`]: flagVariant ?? true } }
    if (flagVariant === 'string') properties.feature_flag_variant = flagVariant
    
    posthog.capture('$feature_interaction', properties)
```

react/src/components/PostHogFeature.tsx
```diff
@@ -46,12 +46,36 @@ export function PostHogFeature({
...
+    posthog,
+    flagVariant,
+}: {
+    flag: string
+    posthog: PostHog
+    flagVariant?: string | boolean
+}) {
+    posthog.capture('$feature_view', {
+        feature_flag: flag,
+        feature_flag_variant: typeof flagVariant === 'string' ? flagVariant : undefined,
```
rafaeelaudibert, line 76: Same as above

rafaeelaudibert (approved):
LGTM, but check whether we should be sending that key all the time or only when there's an actual variant

haacked (approved):
:shipit: LGTM!

PR #1786, opened 2025-03-05
fix(flags): do not reset $anon_distinct_id after first decide call if advanced_disable_feature_flags_on_first_load is true

havenbarnes:
## Changes

Follow up to https://github.com/PostHog/posthog-js/pull/1785, this PR addresses the same case where advanced_disable_feature_flags_on_first_load is set. The second decide call needs $anon_distinct_id to be present in order for authentication persistence to work as intended

## Checklist

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -293,7 +293,9 @@ export class PostHogFeatureFlags {
                     // successful request
                     // reset anon_distinct_id after at least a single request with it
                     // makes it through
-                    this.$anon_distinct_id = undefined
+                    if (!this._additionalReloadRequested) {
+                        this.$anon_distinct_id = undefined
+                    }
                     errorsLoading = false
                 }
 


haacked (approved):
:shipit:

haacked:
Couldn't help myself from digging into why this test is failing. Check out https://github.com/PostHog/posthog-js/blob/main/src/__tests__/setup.js

```js
beforeEach(() => {
    console.error = (...args) => {
        throw new Error(`Unexpected console.error: ${args}`)
    }
    console.warn = (...args) => {
        throw new Error(`Unexpected console.warn: ${args}`)
    }
})
```

That runs before each test and makes it so `console.error` and `console.warn` throw an `Error`.

It turns out that the `js-dom` package writes to the console.

```js
function validCORSHeaders(xhr, response, flag, properties, origin) {
  const acaoStr = response.headers["access-control-allow-origin"];
  const acao = acaoStr ? acaoStr.trim() : null;
  if (acao !== "*" && acao !== origin) {
    properties.error = "Cross origin " + origin + " forbidden";
    dispatchError(xhr);
    return false;
  }
  const acacStr = response.headers["access-control-allow-credentials"];
  const acac = acacStr ? acacStr.trim() : null;
  if (flag.withCredentials && acac !== "true") {
    properties.error = "Credentials forbidden";
    dispatchError(xhr);
    return false;
  }
  return true;
}

function dispatchError(xhr) {
  const errMessage = xhr.properties.error;
  requestErrorSteps(xhr, "error", DOMException.create(xhr._globalObject, [errMessage, "NetworkError"]));

  if (xhr._ownerDocument) {
    const error = new Error(errMessage);
    error.type = "XMLHttpRequest"; // TODO this should become "resource loading" when XHR goes through resource loader

    xhr._ownerDocument._defaultView._virtualConsole.emit("jsdomError", error);
  }
}
```

That's the error we're getting, "Credentials forbidden". I'm not sure what the proper fix is. Is it to override `xhr._ownerDocument._defaultView._virtualConsole`? Is it to remove this `setup.ts`? Or, is it possible the changes we've made are allowing an xhr request to be made that wasn't being made before and causing this issue?

haacked:
Looks like a flaky test. I just reran the job for you and everything works. 😆

PR #1792, opened 2025-03-05
fix: Fix site app loading

robbie-c:
## Changes

There were some bugs with site app loading, which this PR fixes

* It was possible to process batched events for a site app before the init call had returned.
* It was possible for one app to init immediately, and clear out the batched events for yet-to-be-initialized apps

I made one existing test fail by asserting that processEvents had been called the right number of times on both apps, and made a new failing test where the init call returns immediately. These changes fix both tests.

See also https://posthog.slack.com/archives/C06GG249PR6/p1741183829699909

## Checklist

### DIFF
--- a/src/site-apps.ts
+++ b/src/site-apps.ts
@@ -83,32 +83,28 @@ export class SiteApps {
     }
 
     setupSiteApp(loader: SiteAppLoader) {
-        const app: SiteApp = {
-            id: loader.id,
-            loaded: false,
-            errored: false,
-        }
-        this.apps[loader.id] = app
-
-        const onLoaded = (success: boolean) => {
-            this.apps[loader.id].errored = !success
-            this.apps[loader.id].loaded = true
-
-            logger.info(`Site app with id ${loader.id} ${success ? 'loaded' : 'errored'}`)
-
-            if (success && this.bufferedInvocations.length) {
+        const app = this.apps[loader.id]
+        const processBufferedEvents = () => {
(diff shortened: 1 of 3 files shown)


src/site-apps.ts
```diff
@@ -83,32 +83,28 @@ export class SiteApps {
...
             }
 
-            for (const app of Object.values(this.apps)) {
-                if (!app.loaded) {
-                    // If any other apps are not loaded, we don't want to stop buffering
-                    return
-                }
+            if (Object.keys(this.apps).every((id) => this.apps[id].processedBuffer || this.apps[id].errored)) {
+                this.stopBuffering?.()
             }
```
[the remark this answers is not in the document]
robbie-c: stopBuffering removes itself as a function so can only be called once

rafaeelaudibert, line 96: Why iterate over the keys and then get the value?
```suggestion
            if (Object.values(this.apps).every((app) => app.processedBuffer || app.errored)) {
                this.stopBuffering?.()
            }
```

rafaeelaudibert (approved):
LGTM, left a minor comment on a slightly-improved way to iterate over the apps, won't affect usage

rafaeelaudibert:
@robbie-c Small nit: I'd call this is a patch fix not a minor :)

src/site-apps.ts
robbie-c: Good spot, this was leftover from a previous version

PR #1794, opened 2025-03-06
feat: add chunk_id to stack frames

daibhin:
## Changes

- Read chunk ids from global object
- Add chunk ids to frames


oliverb123 (approved):
:shipit:

PR #1795, opened 2025-03-06
fix: Reverse error message

robbie-c:
## Changes

I believe that this is the wrong way around. The docs says here https://posthog.com/docs/libraries/react#posthog-provider 
> The provider can either take an initialized client instance OR an API key and an optional options object.

## Checklist


rafaeelaudibert (approved):
Oh yeah, you're right. I guess I got confused with the error on line 108 and just used the same, ✅

rafaeelaudibert:
Hey @abemelek you're right. We're deploying a new version removing that log message completely, thank you!

Issue #1801, opened 2025-03-07
Same scripts are injected multiple times


janruo:
Posthog-js seems to include the same `<script>` tags multiple times. This is a screenshot from posthog.com itself, and the same thing happens on other sites using Posthog as well:

![Image](https://github.com/user-attachments/assets/379af24f-61ba-4970-bd39-59594029157f)

`dead-clicks-autocapture.js` is injected five times, and `exception-autocapture.js` twice.

I don't think it's causing any actual problems, but surely it's not supposed to be doing that?

rafaeelaudibert:
Hey, @janruo! Are you still seeing this?  We believe #1804 fixed it (v1.230.2+)

janruo:
Looks like it's fixed

pauldambra:
sorry to miss adding the update here 😅

PR #1804, opened 2025-03-09
fix: simple loader race protection

pauldambra:
the external script loader can add a script more than once

while the script might be idempotent it has to be parsed each time and will execute each time

let's add a simple check for the script's presence in the DOM and not add it if it's already there


pauldambra:
see https://github.com/PostHog/posthog-js/issues/1801

marandaneto (approved):

rafaeelaudibert (approved):

PR #1805, opened 2025-03-10
set host

daibhin:
## Changes

Full context in https://github.com/PostHog/posthog/pull/29698


src/posthog-core.ts
```diff
@@ -1275,8 +1276,12 @@ export class PostHog {
     }
 
     /** Get the list of early access features. To check enrollment status, use `isFeatureEnabled`. */
-    getEarlyAccessFeatures(callback: EarlyAccessFeatureCallback, force_reload = false): void {
-        return this.featureFlags.getEarlyAccessFeatures(callback, force_reload)
+    getEarlyAccessFeatures(
+        callback: EarlyAccessFeatureCallback,
+        force_reload = false,
+        stages: EarlyAccessFeatureStage[]
```
robbie-c, line 1282: Since it's a new argument, I'd make `stages` optional. What should this function do if it's not present? In `this.featureFlags.getEarlyAccessFeatures` you send an empty string, is this handled sensibly in the python back end? (I'm guessing it must be? as this is current behaviour)

daibhin: Yeah, Monday fog. Should certainly be optional.

Backend [handles the case when the params aren't passed](https://github.com/PostHog/posthog/pull/29698/files#diff-52e0735255b49a93744178e2cf6e847f677539f02c291a918d2bc47a78f4f629R246) as older clients will still need support

robbie-c (approved):

PR #1809, opened 2025-03-11
fix: prettier needed running

pauldambra:
Somehow got a PR all the way to merged and being published before the publish failed because prettier hadn't been run

wat

labelled as patch so that my failed release gets retried

prettier fail here https://github.com/PostHog/posthog-js/actions/runs/13784405731/job/38548993523

🤣 GH actions warns on prettier errors without failing the step 🫠 

Screenshot 2025-03-11 at 19 19 57

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,13 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+protected_branch='main'
+current_branch=$(git branch --show-current)
+
+if [ $protected_branch = $current_branch ]
+then
+    echo "Pushing to $protected_branch is a sin! Instead, create a pull request and repent."
+    exit 1 # push will not execute
+else
+    exit 0 # push will execute
+fi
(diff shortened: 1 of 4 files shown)


.husky/pre-push
pauldambra, line 1: branch protections are limited and let admins push to them
so we can use this script like in the main repo to stop that

package.json
```diff
@@ -136,7 +136,8 @@
         "*.js": "eslint --cache --fix",
         "*.{ts,tsx}": [
             "eslint src --fix",
-            "eslint playwright --fix"
+            "eslint playwright --fix",
+            "prettier --write"
```
pauldambra, line 140: why even push if this needs doing

src/__tests__/utils/external-scripts-loader.test.ts
```diff
@@ -43,7 +43,7 @@ describe('external-scripts-loader', () => {
             const scripts = document!.getElementsByTagName('script')
             expect(scripts.length).toBe(1)
             expect(scripts[0].src).toMatchInlineSnapshot(`"https://us-assets.i.posthog.com/static/recorder.js?v=1.0.0"`)
-
```
pauldambra, line 46: the offending line

danielbachhuber (approved):

PR #1811, opened 2025-03-11
feat(flags): Add version info to $feature_flag_called request

haacked:
This PR adds `/decide?v=4` support to `posthog-js`. What we get with this support are the following:

## A new `getFeatureFlagDetails(key)` method

This returns the full details of a feature flag. For example:

```json
{
    "beta-feature-flag": {
    "key": "beta-feature-flag",
    "enabled": true,
    "variant": "variant-3",
    "reason": {
        "code": "condition_match",
        "condition_index": 0,
        "description": null
    },
    "metadata": {
        "id": 8,
        "payload": "{\"paylod\": \"3\"}",
        "version": 19,
        "description": ""
    }
}
```

This includes an evaluation `reason`. It also includes `metadata` such as the db id of the flag and its version.

There's also the corresponding `getFlagsWithDetails()` method to get all the flag details.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

1. `$feature_flag_version`: The version of the feature flag.
2. `$feature_flag_id`: The database id of the flag. A flag can be deleted and recreated with the same key, so this helps us determine if that's the case.
3. `$feature_flag_reason`: The reason the flag evaluated as it did. For now, it's just the evaluation code, but once https://github.com/PostHog/posthog/issues/29445 is completed, this will include a human friendly description.
4. `$feature_flag_original_response`: This contains the original response in the situation that the client overrode the flag via a call to `overrideFeatureFlags()`.
5.  `$feature_flag_original_payload`: This contains the original payload in the situation that the client overrode the payload via a call to `overrideFeatureFlags()`.

Sending the original values will help debug in the rare situation that a customer is confused by a flag result and forgot they were overriding it. 😆 

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`. Any requests for `/decide?v=4` today will return a `v=3` response. Once https://github.com/PostHog/posthog/pull/29751 is deployed, then these new additions will "light up".

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -32,6 +32,7 @@ export const SESSION_RECORDING_EVENT_TRIGGER_ACTIVATED_SESSION = '$session_recor
 export const SESSION_RECORDING_EVENT_TRIGGER_STATUS = '$session_recording_event_trigger_status'
 export const ENABLED_FEATURE_FLAGS = '$enabled_feature_flags'
 export const PERSISTENCE_EARLY_ACCESS_FEATURES = '$early_access_features'
+export const PERSISTENCE_FEATURE_FLAG_DETAILS = '$feature_flag_details'
 export const STORED_PERSON_PROPERTIES_KEY = '$stored_person_properties'
 export const STORED_GROUP_PROPERTIES_KEY = '$stored_group_properties'
 export const SURVEYS = '$surveys'
@@ -71,6 +72,7 @@ export const PERSISTENCE_RESERVED_PROPERTIES = [
     ENABLED_FEATURE_FLAGS,
     USER_STATE,
     PERSISTENCE_EARLY_ACCESS_FEATURES,
+    PERSISTENCE_FEATURE_FLAG_DETAILS,
     STORED_GROUP_PROPERTIES_KEY,
     STORED_PERSON_PROPERTIES_KEY,
     SURVEYS,
--- a/src/types.ts
+++ b/src/types.ts
@@ -1390,6 +1390,7 @@ export type FeatureFlagMetadata = {
     version: number | undefined
     description: string | undefined
     payload: JsonType | undefined
+    original_payload?: JsonType | undefined // Only used when overriding a flag payload.
 }
 
 export type EvaluationReason = {
--- a/src/types.ts
+++ b/src/types.ts
@@ -1380,7 +1380,11 @@ export type FeatureFlagsCallback = (
 export type FeatureFlagDetail = {
     key: string
     enabled: boolean
+    // Only used when overriding a flag payload.
+    original_enabled?: boolean | undefined
     variant: string | undefined
+    // Only used when overriding a flag payload.
+    original_variant?: string | undefined
     reason: EvaluationReason | undefined
     metadata: FeatureFlagMetadata | undefined
 }
@@ -1390,7 +1394,8 @@ export type FeatureFlagMetadata = {
     version: number | undefined
     description: string | undefined
     payload: JsonType | undefined
-    original_payload?: JsonType | undefined // Only used when overriding a flag payload.
+    // Only used when overriding a flag payload.
+    original_payload?: JsonType | undefined
 }
 
 export type EvaluationReason = {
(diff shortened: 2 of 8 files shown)


haacked:
Here's an example showing the additional information:

Screenshot 2025-03-13 at 7 30 32 PM

src/constants.ts
```diff
@@ -71,6 +72,7 @@ export const PERSISTENCE_RESERVED_PROPERTIES = [
     ENABLED_FEATURE_FLAGS,
     USER_STATE,
     PERSISTENCE_EARLY_ACCESS_FEATURES,
+    PERSISTENCE_FEATURE_FLAG_DETAILS,
```
haacked, line 75: If you're wondering why I am excluding this, if we don't, it would get rendered like so:

Screenshot 2025-03-13 at 7 06 56 PM

Yikes! Useful info, but maybe too much and we'd need to figure out how to render it more nicely.

haacked:
Made one last (hopefully) change to how we include the `reason` in the response:

image

I think the whole structure is useful. Later, when we implement a human friendly description, it can be included in there.

I need to dig into this, but by having structured data in there, I hope this'll make it possible to filter on `$feature_flag_called` events by `reason.code` so we can see all flags that matched a condition, etc.

dmarticus:
> Made one last (hopefully) change to how we include the `reason` in the response:
> 
> image
> 
> I think the whole structure is useful. Later, when we implement a human friendly description, it can be included in there.
> 
> I need to dig into this, but by having structured data in there, I hope this'll make it possible to filter on `$feature_flag_called` events by `reason.code` so we can see all flags that matched a condition, etc.
> 
> Happy to roll that back if others don't like it and would rather it just be the human friendly description.

Can the `condition_index` field be better somehow?  Or just omitted?  I don't think it adds clarity – if we're just going to end up deprecating this field anyway in the favor of a human-friendly description then I guess I'm fine with it, but I'd rather omit it entirely for now, at least in the event table.

src/constants.ts
dmarticus: yeah, super reasonable to not include all this info by default.  Would be awesome if we had a UI component to make this nice, but it's out of scope for this change IMO.

src/types.ts
```diff
@@ -1376,6 +1377,33 @@ export type FeatureFlagsCallback = (
...
+    version: number | undefined
+    description: string | undefined
+    payload: JsonType | undefined
+    // Only used when overriding a flag payload.
+    original_payload?: JsonType | undefined
+}
+
+export type EvaluationReason = {
+    code: string
+    condition_index: number | undefined
```
dmarticus, line 1403: I made this note in the top-level PR review, but I'm not sure this field helps add context.

src/types.ts
```diff
@@ -1376,6 +1377,33 @@ export type FeatureFlagsCallback = (
...
+    id: number
+    version: number | undefined
+    description: string | undefined
+    payload: JsonType | undefined
+    // Only used when overriding a flag payload.
+    original_payload?: JsonType | undefined
+}
+
+export type EvaluationReason = {
+    code: string
```
dmarticus, line 1402: the code isn't super human-readable (e.g. `condition_match` or `no_condition_match`, seems like a low lift to have a helper that renders better strings in the UI. This is probably a posthog app change (taxonomy), but in e.g. the persons view we have this mapping to show human-friendly match reasons
```
const featureFlagMatchMapping = {
    [FeatureFlagMatchReason.ConditionMatch]: 'Matches',
    [FeatureFlagMatchReason.NoConditionMatch]: "Doesn't match any conditions",
    [FeatureFlagMatchReason.OutOfRolloutBound]: 'Out of rollout bound',
    [FeatureFlagMatchReason.NoGroupType]: 'Missing group type',
    [FeatureFlagMatchReason.SuperConditionMatch]: 'Matches early access condition',
    [FeatureFlagMatchReason.Disabled]: 'Disabled',
}

```

dmarticus (approved):
looks great; only major note is that I think we should leverage some of our existing stuff in the PostHog app (see `RelatedFeatureFlags.tsx`) to make the evaluation/match reasons more human friendly.  I think that can come in your taxonomy PR, though.

src/types.ts
haacked: Agreed. I updated it to just send the reason.description.

PR #1812, opened 2025-03-12
chore: add new question index to id map

lucasheriques:
## Changes

add a new map from question index to its ID to help build queries from just the survey event, without relying on hardcoded ids

## Checklist

### DIFF
--- a/src/extensions/surveys/surveys-utils.tsx
+++ b/src/extensions/surveys/surveys-utils.tsx
@@ -34,6 +34,16 @@ export function getSurveyResponseKey(questionId: string) {
     return `$survey_response_${questionId}`
 }
 
+function getSurveyIndexToIdMap(survey: Survey) {
+    return survey.questions.reduce<Record<number, string>>((acc, question, index) => {
+        if (!question.id) {
+            return acc
+        }
+        acc[index] = getSurveyResponseKey(question.id)
+        return acc
+    }, {})
+}
+
 export const style = (appearance: SurveyAppearance | null) => {
     const positions = {
         left: 'left: 30px;',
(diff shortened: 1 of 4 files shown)


src/extensions/surveys/surveys-utils.tsx
```diff
@@ -34,6 +34,16 @@ export function getSurveyResponseKey(questionId: string) {
     return `$survey_response_${questionId}`
 }
 
+function getSurveyIndexToIdMap(survey: Survey) {
+    return survey.questions.reduce<Record<number, string>>((acc, question, index) => {
```
marandaneto, line 38: will the index matter here if the question is shuffled? just trying to avoid a bug If possible

marandaneto: indexes ideally would match the $survey_questions as well

marandaneto: Can we check if `$survey_questions` is being used in any insight, and if not, just fix that one?
eg
```
$survey_questions: [
	{
	id: UUID
	index: 1
	text: the text
	}
]
```

lucasheriques: removed this in favor of just using `survey_questions`

src/extensions/surveys/surveys-utils.tsx
```diff
@@ -34,6 +34,16 @@ export function getSurveyResponseKey(questionId: string) {
     return `$survey_response_${questionId}`
 }
 
+function getSurveyIndexToIdMap(survey: Survey) {
```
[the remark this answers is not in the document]
lucasheriques: removed in favor of `survey_questions`

lucasheriques:
@marandaneto @ioannisj confirmed with [product analytics team](https://posthog.slack.com/archives/C0368RPHLQH/p1741808883101149), `$survey_questions` should be safe to change, so using it now to storing the questions. works fine with shuffle questions too

marandaneto (approved):

PR #1826, opened 2025-03-13
feat: remote image masking

pauldambra:
let's set blockSelector: img for folk that want it

pairs with a nice malbec or https://github.com/PostHog/posthog/pull/29898

(tested locally with the nextjs playground)


joshsny (approved):
LGTM 👍

PR #1829, opened 2025-03-17
feat: export `PostHogProviderProps` type

kidonng:
## Changes

I'm writing a wrapper for the React provider to only enable PostHog in production:

```ts
import { PostHogProvider as PostHogProviderOriginal } from 'posthog-js/react';
import type { ComponentProps } from 'react';

export function PostHogProvider({
  children,
  ...props
}: ComponentProps<typeof PostHogProviderOriginal>) {
  return process.env.NODE_ENV === 'production' ? (
    <PostHogProviderOriginal {...props}>{children}</PostHogProviderOriginal>
  ) : (
    children
  );
}
```

But ideally I can just reference the exported props:

```ts
import {
  PostHogProvider as PostHogProviderOriginal,
  // Does not exist yet
  type PostHogProviderProps
} from 'posthog-js/react';

export function PostHogProvider({
  children,
  ...props
}: PostHogProviderProps) {
  return process.env.NODE_ENV === 'production' ? (
    <PostHogProviderOriginal {...props}>{children}</PostHogProviderOriginal>
  ) : (
    children
  );
}
```

This pull request export the `PostHogProviderProps` type to allow easier access and further customizations.

While we are at it, I also replaced the custom `WithOptionalChildren` type with the identical [`React.PropsWithChildren`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/a4450792b76f9faa16b0f91bcac5d6be1fcb66e6/types/react/index.d.ts#L1394) type. This change can be omitted if not desired.

## Checklist

### DIFF
--- a/react/src/context/PostHogProvider.tsx
+++ b/react/src/context/PostHogProvider.tsx
@@ -12,8 +12,6 @@ type AlreadyInitialized =
       }
     | false
 
-type WithOptionalChildren<T> = T & { children?: React.ReactNode | undefined }
-
 /**
  * Props for the PostHogProvider component.
  * This is a discriminated union type that ensures mutually exclusive props:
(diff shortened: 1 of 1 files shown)


kidonng:
I still assume this would be useful.

PR #1832, opened 2025-03-18
fix: check surveys timeout in the survey manager level

lucasheriques:
## Changes

[fixes issues on this thread:](https://posthog.slack.com/archives/C07QD3LT8U9/p1742237034541509)

instead of handling timeouts in the SurveyPopup level (And rendering it even before it's needed), instead, we handle it directly on the `SurveyManager` level.

this way we can also guarantee no duplicate timeouts will be triggered - as we now track those timeouts in the class. And clear them every time we re-render a survey.

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -179,10 +179,8 @@ export class SurveyManager {
     }
 
     private clearSurveyTimeout(surveyId: string) {
-        if (this.surveyTimeouts.has(surveyId)) {
-            clearTimeout(this.surveyTimeouts.get(surveyId))
-            this.surveyTimeouts.delete(surveyId)
-        }
+        clearTimeout(this.surveyTimeouts.get(surveyId))
+        this.surveyTimeouts.delete(surveyId)
     }
 
     private handlePopoverSurvey = (survey: Survey): void => {
@@ -215,6 +213,7 @@ export class SurveyManager {
                 if (!doesSurveyUrlMatch(survey)) {
                     return this.removeSurveyFromFocus(survey.id)
                 }
+                // rendering with surveyPopupDelaySeconds = 0 because we're already handling the timeout here
                 Preact.render(
                     <SurveyPopup
                         key={'popover-survey'}
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -179,8 +179,11 @@ export class SurveyManager {
     }
 
     private clearSurveyTimeout(surveyId: string) {
-        clearTimeout(this.surveyTimeouts.get(surveyId))
-        this.surveyTimeouts.delete(surveyId)
+        const timeout = this.surveyTimeouts.get(surveyId)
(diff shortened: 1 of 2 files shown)


src/extensions/surveys.tsx
```diff
@@ -187,18 +195,39 @@ export class SurveyManager {
...
+                        removeSurveyFromFocus={this.removeSurveyFromFocus}
+                        isPopup={true}
+                    />,
+                    shadow
+                )
+            }
+            const timeoutId = setTimeout(() => {
+                if (!doesSurveyUrlMatch(survey)) {
+                    return this.removeSurveyFromFocus(survey.id)
+                }
```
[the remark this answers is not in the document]
lucasheriques: we need to check it in the timeout because the URL can change after scheduling the survey. and we need to guarantee it will not show up if it's no longer a match

src/extensions/surveys.tsx
```diff
@@ -187,18 +195,38 @@ export class SurveyManager {
...
+                Preact.render(
+                    <SurveyPopup
+                        key={'popover-survey'}
+                        posthog={this.posthog}
+                        survey={{ ...survey, appearance: { ...survey.appearance, surveyPopupDelaySeconds: 0 } }}
```
marandaneto, line 222: why do we force `surveyPopupDelaySeconds: 0` here?

lucasheriques: because we're already handling the delay within the SurveyManager, so no need to handle it again in the SurveyPopup logic.

I'm still keeping the SurveyPopup logic for backwards compatibility, though shouldn't be used anymore by surveys managed by us (i.e. not triggered via `renderSurvey`)

src/extensions/surveys.tsx
```diff
@@ -179,10 +179,8 @@ export class SurveyManager {
     }
 
     private clearSurveyTimeout(surveyId: string) {
-        if (this.surveyTimeouts.has(surveyId)) {
-            clearTimeout(this.surveyTimeouts.get(surveyId))
-            this.surveyTimeouts.delete(surveyId)
-        }
+        clearTimeout(this.surveyTimeouts.get(surveyId))
+        this.surveyTimeouts.delete(surveyId)
```
marandaneto, line 183: you still do a dual lookup since delete has to check if it exists or not before deleting
```
const timeout = this.surveyTimeouts.get(surveyId)
if (time) {
    clearTimeout(timeout)
    this.surveyTimeouts.delete(surveyId)
}
```

now at least it only does the 2nd lookup if needed

lucasheriques: even then, it's a O(1) operation, so i don't think we gain much from this. i'll add though since no strong objections

marandaneto (approved):

src/extensions/surveys.tsx
marandaneto: thats how systems get slow, one extra lookup here, another there, etc

PR #1835, opened 2025-03-19
fix: do not allow non string `$current_url`  to be provided

pauldambra:
we've had a few support tickets in replay that boil down to 

if you pass an invalid value to $current_url we accept it, use it, and then our UI explodes

let's add special casing to not allow someone to send e.g. `new URL(window.location.href)` when they think they're being helpful

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -864,7 +864,7 @@ export class PostHog {
         }
 
         if (properties?.$current_url && !isString(properties?.$current_url)) {
-            logger.error('Invalid $current_url property provided to posthog.capture. removing provided value')
+            logger.error('Invalid `$current_url` property provided to `posthog.capture`. Input must be a string. Ignoring provided value.')
             delete properties?.$current_url
         }
 
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -863,6 +863,11 @@ export class PostHog {
             return
         }
 
+        if (properties?.$current_url && !isString(properties?.$current_url)) {
+            logger.error('Invalid $current_url property provided to posthog.capture. removing provided value')
+            delete properties?.$current_url
+        }
+
         // update persistence
         this.sessionPersistence.update_search_keyword()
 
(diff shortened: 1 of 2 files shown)


src/posthog-core.ts
```diff
@@ -863,6 +863,11 @@ export class PostHog {
             return
         }
 
+        if (properties?.$current_url && !isString(properties?.$current_url)) {
+            logger.error('Invalid $current_url property provided to posthog.capture. removing provided value')
```
rafaeelaudibert, line 867:
```suggestion
            logger.error('Invalid `$current_url` property provided to `posthog.capture`. Input must be a string. Ignoring provided value.')
```

pauldambra: LLM writes the message
LLM complains about the message

🙈 

(message much better like this :))

rafaeelaudibert (approved):
🚀 but CI is failing

src/posthog-core.ts
```diff
@@ -863,6 +863,13 @@ export class PostHog {
             return
         }
 
+        if (properties?.$current_url && !isString(properties?.$current_url)) {
+            logger.error(
+                'Invalid `$current_url` property provided to `posthog.capture`. Input must be a string. Ignoring provided value.'
+            )
+            delete properties?.$current_url
```
pauldambra, line 870: most annoying feature of GitHub is you can't thread conversations on the main page of the PR @haacked 

> Could we just detect this situation and unpack it for them? After all, they were trying to be helpful. 😆

We sort of will since we always add the correct URL anyway
(and in the replay UI we try and grab the right href if you send `{ href: 'blah' }` instead of 'blah')

but the most recent case I saw we just got `{}` as the URL for some reason, so i thought rather than try and detect  what value they were sending we'll just delete it and set the "correct" value

pauldambra: (obvs happy to be cleverer if i'm missing something!)

haacked: Sounds good to me!

pauldambra: Screenshot 2025-03-19 at 19 54 24

we get a relatively wide range of values if we query

```
select mat_$current_url
from events
where timestamp <= now()
and timestamp >= now() - interval 7 day
and event = '$pageview'
and mat_$current_url like '{%'
```

haacked: Oh wow!

PR #1837, opened 2025-03-19
feat(flags): Support `/decide?v=4` and include version info flag called events

haacked:
This PR adds `/decide?v=4` support to `posthog-js`.

> [!NOTE]
> This PR adds _support_ for `v=4`, but is still pointing to `v=3`. I'll update it to point to v4 in a follow-up PR.

What we get with this support are the following:

## A new `getFeatureFlagDetails(key)` method

This returns the full details of a feature flag. For example:

```json
{
    "beta-feature-flag": {
    "key": "beta-feature-flag",
    "enabled": true,
    "variant": "variant-3",
    "reason": {
        "code": "condition_match",
        "condition_index": 0,
        "description": null
    },
    "metadata": {
        "id": 8,
        "payload": "{\"paylod\": \"3\"}",
        "version": 19,
        "description": ""
    }
}
```

This includes an evaluation `reason`. It also includes `metadata` such as the db id of the flag and its version.

There's also the corresponding `getFlagsWithDetails()` method to get all the flag details.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

1. `$feature_flag_version`: The version of the feature flag.
2. `$feature_flag_id`: The database id of the flag. A flag can be deleted and recreated with the same key, so this helps us determine if that's the case.
3. `$feature_flag_reason`: The reason the flag evaluated as it did. For now, it's just the evaluation code, but once https://github.com/PostHog/posthog/issues/29445 is completed, this will include a human friendly description.
4. `$feature_flag_original_response`: This contains the original response in the situation that the client overrode the flag via a call to `overrideFeatureFlags()`.
5.  `$feature_flag_original_payload`: This contains the original payload in the situation that the client overrode the payload via a call to `overrideFeatureFlags()`.

Sending the original values will help debug in the rare situation that a customer is confused by a flag result and forgot they were overriding it. 😆 

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

### DIFF
--- a/playwright/session-recording/session-recording-network-recorder.spec.ts
+++ b/playwright/session-recording/session-recording-network-recorder.spec.ts
@@ -134,7 +134,7 @@ test.beforeEach(async ({ context }) => {
                               // webkit isn't capturing this failed request in the pre-wrapped fetch performance observer records
                               // [/https:\/\/localhost:\d+\/array\/test%20token\/config.js/, 'script'],
                               [
-                                  /https:\/\/localhost:\d+\/decide\/\?v=3&ip=1&_=\d+&ver=1\.\d\d\d\.\d+&compression=base64/,
+                                  /https:\/\/localhost:\d+\/decide\/\?v=4&ip=1&_=\d+&ver=1\.\d\d\d\.\d+&compression=base64/,
                                   'fetch',
                               ],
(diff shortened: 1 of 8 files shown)


dmarticus (commented):
nice work!

dmarticus (approved):
nice work!

PR #1839, opened 2025-03-19
feat: do not start recorder active-active

pauldambra:
while investigating a supprt ticket we saw that a user had a large spike of crawls from ahrefs

ahrefs do not provide a custom user agent that can be detected as a bot and blocked (by our SDK's default user agent blocking)

so that large spike of sessions were all recorded and exhausted the user's free limit

but all of them were recordings with no user activity (sort of... some had a little keyboard activity, presumably accidental or behaviour built into the crawler)

this made me realise... we start the recorder active and wait for a lack of user activity to tell us that it is idle

instead we can start the recorder idle and let the presence of user activity switch to active

(in reality i started "unknown" so I could skip some debug logging and make things more clear)

### DIFF
--- a/src/__tests__/extensions/replay/sessionrecording.test.ts
+++ b/src/__tests__/extensions/replay/sessionrecording.test.ts
@@ -1478,7 +1478,7 @@ describe('SessionRecording', () => {
     describe('idle timeouts', () => {
         let startingTimestamp = -1
 
-        function emitInactiveEvent(activityTimestamp: number, expectIdle: boolean = false) {
+        function emitInactiveEvent(activityTimestamp: number, expectIdle: boolean | 'unknown' = false) {
             const snapshotEvent = {
                 event: 123,
                 type: INCREMENTAL_SNAPSHOT_EVENT_TYPE,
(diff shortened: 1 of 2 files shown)


src/extensions/replay/sessionrecording.ts
```diff
@@ -290,10 +294,6 @@ export class SessionRecording {
         return this.instance.config.session_recording.session_idle_threshold_ms || RECORDING_IDLE_THRESHOLD_MS
     }
 
-    private get rrwebRecord(): rrwebRecord | undefined {
```
pauldambra, line 293: flyby change to make this a function instead of a class method
we can't minify class methods so every call to `rrwebRecord` uses the fullname

this PR was +85B and is now +15B

not much in the grand scheme of things but better than nothing

src/__tests__/extensions/replay/sessionrecording.test.ts
```diff
@@ -2163,7 +2186,6 @@ describe('SessionRecording', () => {
             sessionRecording.onRRwebEmit(createIncrementalSnapshot({ data: { source: 1 } }) as any)
 
             expect(sessionRecording['queuedRRWebEvents']).toHaveLength(0)
-            expect(sessionRecording['rrwebRecord']).not.toBeUndefined()
```
pauldambra, line 2166: this assertion only tested that the mock existed 🤷

PR #1841, opened 2025-03-20
chore(flags): use the new `/flags` endpoint to manage flag evaluation for team 2

dmarticus:
## Changes

Basically, what I'm doing is that any request with the PostHog team 2 public API key, we use `/flags`, and then for everything else we just use `/decide`.  

There's a little bit of trickiness w.r.t remote config (I added a comment to that end) since new flags relies on the fact that remote config takes care of all the other non-flag-related code (it basically does away with `/decide` altogether).  But if a customer isn't using remote config, they'll need to hit `/decide` anyway.  I should have this handled in the logic, but worth calling out.


haacked (approved):
Looks good to me!

src/posthog-featureflags.ts
```diff
@@ -390,10 +391,12 @@ export class PostHogFeatureFlags {
             data.disable_flags = true
         }
 
+        const eligibleForFlagsV2 = token === DEFAULT_POSTHOG_APP_API_KEY && this.instance.config.__preview_remote_config
```
robbie-c, line 394: Any chance we could just do

```suggestion
        const eligibleForFlagsV2 = this.instance.config.__preview_flags_v2
```

so that I can run this on localhost easily?

dmarticus: oh yeah, and then we can just set the preview to true for our own posthog JS initialization + any other instances of it.

That's a good idea; I might still want to use API tokens down the road (especially when rolling out to specific users), but I'm happy doing something like this instead.  Thanks for the suggestion!

dmarticus: @robbie-c went ahead and went with the config-based approach, although I might need to use API tokens down the road too (so I can turn this on for specific customers as needed).  I think my plan will be to have a set of tokens that have this feature toggled on, and then allow folks to specify it in their config as well.

PR #1838, opened 2025-03-19
feat(flags): Enjoy the benefits of `/decide?v=4`. `$feature_flag_called` events contain more details such as flag version

haacked:
This is a follow-up to #1837 where we added support for `/decide?v=4` in a backwards compatible fashion, but still made requests to `/decide?v=3`

With this change, we should get more useful debug information in `$feature_flag_called` events.

## Changes

Changes the library to request `/decide?v=4` instead of `/decide?v=3`. All the substantial changes were made in #1837.

## Checklist
- [x] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [x] Accounted for the impact of any changes across different browsers
- [x] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

PR #1842, opened 2025-03-21
fix(flags): Revert "feat(flags): Enjoy the benefits of `/decide?v=4`. `$feature_flag_called` events contain more details such as flag version"

dmarticus:
Reverts PostHog/posthog-js#1838

Oof, v4 responses from `/decide` were not smart about flags that were just straight disabled, let's roll back to v3


haacked (approved):
Does this need a bump patch label?

dmarticus:
> Does this need a bump patch label?

oh yeah, wrong label

PR #1845, opened 2025-03-21
fix: make callback on confirmation message instead

lucasheriques:
## Changes

after I fix I did for [this issue](https://github.com/PostHog/posthog-js/pull/1767), it introduced a bug for feedback surveys where the confirmation message is not shown for feedback survey since it's being closed in the same instant.

so, this PR does two things:

- address it so the `reset` for the feedback widget popup happens when the survey is closed
- creates an end to end test to make sure this error doesn't happen again

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -693,7 +693,7 @@ interface SurveyPopupProps {
     isPopup?: boolean
     onPreviewSubmit?: (res: string | string[] | number | null) => void
     onPopupSurveyDismissed?: () => void
-    onPopupSurveySent?: () => void
+    onCloseConfirmationMessage?: () => void
 }
 
 export function SurveyPopup({
@@ -706,7 +706,7 @@ export function SurveyPopup({
     isPopup,
     onPreviewSubmit = () => {},
     onPopupSurveyDismissed = () => {},
-    onPopupSurveySent = () => {},
+    onCloseConfirmationMessage = () => {},
 }: SurveyPopupProps) {
     const isPreviewMode = Number.isInteger(previewPageIndex)
     // NB: The client-side code passes the millisecondDelay in seconds, but setTimeout expects milliseconds, so we multiply by 1000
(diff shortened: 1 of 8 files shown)


src/extensions/surveys.tsx
```diff
@@ -819,7 +818,6 @@ export function Questions({
         const nextStep = getNextSurveyStep(survey, displayQuestionIndex, res)
         if (nextStep === SurveyQuestionBranchingType.End) {
             sendSurveyEvent({ ...questionsResponses, [responseKey]: res }, survey, posthog)
-            onPopupSurveySent()
         } else {
```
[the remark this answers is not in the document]
lucasheriques: this is handled somewhere else

src/extensions/surveys/components/BottomSection.tsx
```diff
@@ -31,6 +31,7 @@ export function BottomSection({
                 <button
                     className="form-submit"
                     disabled={submitDisabled}
+                    aria-label="Submit survey"
```
marandaneto, line 34: was this missing?

src/extensions/surveys/components/QuestionHeader.tsx
```diff
@@ -35,8 +35,14 @@ export function Cancel({ onClick }: { onClick: () => void }) {
...
 
     return (
-        <div className="cancel-btn-wrapper" onClick={onClick} disabled={isPreviewMode}>
-            <button className="form-cancel" onClick={onClick} disabled={isPreviewMode}>
+        <div className="cancel-btn-wrapper">
+            <button
+                className="form-cancel"
+                onClick={onClick}
+                disabled={isPreviewMode}
+                aria-label="Close survey"
```
marandaneto, line 43: same https://github.com/PostHog/posthog-js/pull/1845/files#r2010206457

marandaneto (approved):

src/extensions/surveys/components/BottomSection.tsx
lucasheriques: yes

src/extensions/surveys/components/QuestionHeader.tsx
lucasheriques: also missing

PR #1850, opened 2025-03-26
feat: always include stack trace parsing

daibhin:
## Problem

A bunch of customers are complaining that their exceptions don't have stack traces. We do not include the stack trace by default if the parser isn't available.

We only included the parser if autocapture was enabled for bundle size reasons. This has proven to be:
- Confusing for us to debug customer issues
- Confusing for customers who don't understand why they don't have stack traces on some manually captured exceptions
- A bad product experience overall because we can't group very effectively
- Not really necessary because it only adds at most ~3% to the bundle

## Changes

Include `errorToProperties` in the default bundle


oliverb123 (approved):
shipit

PR #1851, opened 2025-03-27
fix: only load surveys if there are surveys and its enabled

marandaneto:
## Changes

Surveys only hit the API if it's enabled

## Checklist


src/posthog-surveys-types.ts
```diff
@@ -118,10 +118,6 @@ interface SpecificQuestionBranching {
     index: number
 }
 
-export interface SurveyResponse {
```
marandaneto, line 121: not used anywhere

src/posthog-surveys.ts
```diff
@@ -75,10 +76,20 @@ export class PostHogSurveys {
...
+        } else if (isBoolean(surveys)) {
+            this._hasSurveys = surveys
+            this._decideServerResponse = true
+        }
 
-        this.loadIfEnabled()
+        if (this._decideServerResponse) {
+            logger.info(`decideServerResponse set to ${this._decideServerResponse}, hasSurveys: ${this._hasSurveys}`)
+            this.loadIfEnabled()
+        }
```
lucasheriques, line 92: with this i feel like `_decideServerResponse` is not really needed anymore? since it's only used to check that. unless i'm missing something? and then we can just use `hasSurveys` which is more meaningful too

lucasheriques (approved):
only one comment but looks great, ty for fixing

src/posthog-surveys.ts
marandaneto: its more about the logging.
decide not loaded != not having surveys

functional_tests/mock-server.ts
```diff
@@ -26,12 +26,11 @@ const handleRequest = (group: string) => (req: RestRequest, res: ResponseComposi
             } else if (gzipCompressed) {
                 const data = new Uint8Array(req._body)
                 const decoded = strFromU8(decompressSync(data))
-                console.log(decoded)
                 body = JSON.parse(decoded)
             } else {
                 body = JSON.parse(decodeURIComponent(body.split('=')[1]))
             }
-        } catch (e) {
+        } catch {
```
marandaneto, line 33: linters complained

ioannisj (approved):
LG

src/posthog-surveys.ts
marandaneto: i did the logging elsewhere and removed the _decideServerResponse anyway

src/posthog-surveys.ts
```diff
@@ -75,10 +76,20 @@ export class PostHogSurveys {
...
     onRemoteConfig(response: RemoteConfig) {
-        this._decideServerResponse = !!response['surveys']
-        logger.info(`decideServerResponse set to ${this._decideServerResponse}`)
+        // only load surveys if they are enabled and there are surveys to load
+        const surveys = response['surveys']
+        if (isArray(surveys)) {
+            this._hasSurveys = surveys.length > 0
+            this._decideServerResponse = true
+        } else if (isBoolean(surveys)) {
+            this._hasSurveys = surveys
```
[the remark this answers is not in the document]
marandaneto: decide and remote config are different
response can be:
surveys: false
surveys: true
surveys: Survey[] which means surveys is enabled
they share the same response type

src/posthog-surveys.ts
```diff
@@ -76,10 +76,20 @@ export class PostHogSurveys {
...
+        } else {
+            logger.warn('Decide not loaded yet. Not loading surveys.')
+        }
 
-        this.loadIfEnabled()
+        if (this._hasSurveys) {
+            logger.info(`decide response received, hasSurveys: ${this._hasSurveys}`)
+            this.loadIfEnabled()
+        }
```
lucasheriques, line 92: feels like we can simplify this a bit. what do you think?

```ts
    onRemoteConfig(response: RemoteConfig) {
        // only load surveys if they are enabled and there are surveys to load
        const surveys = response['surveys']
        if (isUndefined(surveys)) {
            return logger.warn('Decide not loaded yet. Not loading surveys.')
        }
        const isArrayResponse = isArray(surveys)
        this._hasSurveys = isArrayResponse ? surveys.length > 0 : surveys
        if (this._hasSurveys) {
            logger.info(`decide response received, hasSurveys: ${this._hasSurveys}`)
            this.loadIfEnabled()
        }
    }
```

lucasheriques: actually i'd move the decide response received log to be outside the `if` so we always log the response

marandaneto: yep, thx

lucasheriques (approved):

PR #1853, opened 2025-03-28
fix: do not always wrap console.error

daibhin:
## Changes

This was a big boo boo on my end. Wasn't correctly reading the config so we were wrapping the `console.error` by default when we shouldn't have been

Added a test


src/extensions/exception-autocapture/index.ts
```diff
@@ -98,13 +90,13 @@ export class ExceptionObserver {
...
-            if (!this.unwrapOnError) {
+            if (!this.unwrapOnError && this.config.capture_unhandled_errors) {
                 this.unwrapOnError = wrapOnError(this.captureException.bind(this))
             }
-            if (!this.unwrapUnhandledRejection) {
+            if (!this.unwrapUnhandledRejection && this.config.capture_unhandled_rejections) {
                 this.unwrapUnhandledRejection = wrapUnhandledRejection(this.captureException.bind(this))
             }
-            if (!this.unwrapConsoleError) {
+            if (!this.unwrapConsoleError && this.config.capture_console_errors) {
```
daibhin, line 99: This is the key line. We shouldn't be wrapping unless it is explicitly enabled in the config

oliverb123 (approved):

PR #1857, opened 2025-03-29
fix: return can render survey reason

lucasheriques:
## Changes

right now `canRenderSurvey` is useless because it doesn't return anything.

this fixes it and adds a helper button in the nextjs playground

## Checklist

### DIFF
--- a/src/posthog-surveys.ts
+++ b/src/posthog-surveys.ts
@@ -389,13 +389,15 @@ export class PostHogSurveys {
     canRenderSurvey(surveyId: string): SurveyRenderReason | null {
         if (isNullish(this._surveyManager)) {
             logger.warn('init was not called')
-            return { visible: false }
+            return { visible: false, disabledReason: 'SDK is not enabled or survey functionality is not yet loaded' }
         }
         let renderReason: SurveyRenderReason | null = null
         this.getSurveys((surveys) => {
             const survey = surveys.filter((x) => x.id === surveyId)[0]
             if (survey) {
                 renderReason = { ...this._surveyManager.canRenderSurvey(survey) }
+            } else {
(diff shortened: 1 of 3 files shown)


src/posthog-surveys.ts
```diff
@@ -386,15 +386,19 @@ export class PostHogSurveys {
         return assignableWindow.__PosthogExtensions__.canActivateRepeatedly(survey)
     }
 
-    canRenderSurvey(surveyId: string) {
+    canRenderSurvey(surveyId: string): SurveyRenderReason | null {
         if (isNullish(this._surveyManager)) {
             logger.warn('init was not called')
-            return
+            return { visible: false }
```
marandaneto, line 392: should we return `disabledReason: 'SDK wasn't enabled'`?

src/posthog-surveys.ts
```diff
@@ -386,15 +386,19 @@ export class PostHogSurveys {
...
         }
+        let renderReason: SurveyRenderReason | null = null
         this.getSurveys((surveys) => {
             const survey = surveys.filter((x) => x.id === surveyId)[0]
-            this._surveyManager.canRenderSurvey(survey)
+            if (survey) {
+                renderReason = { ...this._surveyManager.canRenderSurvey(survey) }
+            }
```
marandaneto, line 399: I think we need an `else` branch here with `visible: false` and `disabledReason: 'Survey not found'`

marandaneto (approved):
left 2 comments otherwise LGTM

src/posthog-surveys.ts
lucasheriques: agreed, addressed both comments

PR #1865, opened 2025-04-03
fix: survey feedback button anchoring to trigger

lucasheriques:
## Changes

Fixed survey widget positioning for better cross-site compatibility by improving viewport calculations. 

Replaced fixed height constraints with more flexible positioning logic based on available viewport space.

necessary to fix [this ticket](https://posthoghelp.zendesk.com/agent/tickets/27879)

Before: survey was not showing up in the correct anchored position

After:

CleanShot 2025-04-03 at 01 53 47@2x


ioannisj (approved):
Not tested but since this is all UI, I assume it works as expected

lucasheriques:
yeah it's all UI, mostly improving a few CSS calculations. We also have e2e tests for widget surveys type too to make sure it's not breaking @ioannisj

PR #1871, opened 2025-04-05
fix: add minimal bundle output (#65)

### LINKED ISSUE #65: This library is large
As it stands now, posthog-js v1.3.0 is about [52.4kb](https://bundlephobia.com/result?p=posthog-js@1.3.0) of minified code.

image

While nowhere near as bad as some of our competitors (mixpanel-browser is [95KB](https://bundlephobia.com/result?p=mixpanel-browser@2.38.0) and segment's analytics.js is [187KB](https://bundlephobia.com/result?p=analytics.js@2.9.1)), this can still be improved upon.

Should we improve on this and with what priority? We've covered the low hanging fruit and the next step is to refactor and modernise the library's internals. This would greatly help [writing proper tests](https://github.com/PostHog/posthog-js/issues/50) as well.

thexeos:
## Changes

This PR introduces build-time flags around various optional extension initializers. Once bundled, Terser can effectively remove extension code and their dependencies.

Additionally, `debug`, `info`, `warn`, and `error` log statements were removed from the minimal bundle. Critical logs have been kept.

Furthermore, importing of version number from `package.json` when building was refactored to use Rollup's replace plugin. As a result, this addressed outputting of individual type definitions for all files in `/src` that were being placed in `/dist` directory.

### Breaking Changes

Since we are adding a new named bundle, 

### What can still be improved

- It should be possible to modify `package.json` to export this bundle as a named export (e.g., `import posthog from 'posthog-js/tiny'`).
- Code and tests can be refactored to mark more properties as private, which should add another 1-2% bundle size saving.

Closes #65

## Checklist

### DIFF
--- a/src/utils/request-router.ts
+++ b/src/utils/request-router.ts
@@ -17,7 +17,7 @@ export type RequestRouterTarget = 'api' | 'ui' | 'assets'
 const ingestionDomain = 'i.posthog.com'
 
 export class RequestRouter {
-    instance: PostHog
+    private instance: PostHog
     private _regionCache: Record<string, RequestRouterRegion> = {}
 
     constructor(instance: PostHog) {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,11 @@ module.exports = {
         given: 'readonly',
         global: 'readonly',
         Buffer: 'readonly',
+<<<<<<< Updated upstream
+=======
+        MINIMAL_BUILD: 'readonly',
+        BUILD_VERSION: 'readonly',
+>>>>>>> Stashed changes
     },
     parser: '@typescript-eslint/parser',
     parserOptions: {
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,11 +33,8 @@ module.exports = {
         given: 'readonly',
         global: 'readonly',
         Buffer: 'readonly',
-<<<<<<< Updated upstream
-=======
         MINIMAL_BUILD: 'readonly',
         BUILD_VERSION: 'readonly',
->>>>>>> Stashed changes
     },
     parser: '@typescript-eslint/parser',
     parserOptions: {
--- a/package.json
+++ b/package.json
@@ -64,6 +64,7 @@
         "@rollup/plugin-commonjs": "^28.0.1",
         "@rollup/plugin-json": "^6.1.0",
         "@rollup/plugin-node-resolve": "^16.0.0",
+        "@rollup/plugin-replace": "^6.0.2",
         "@rollup/plugin-terser": "^0.4.4",
         "@rollup/plugin-typescript": "^12.1.2",
         "@rrweb/record": "2.0.0-alpha.17",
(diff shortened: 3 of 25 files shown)


pauldambra:
this is a really interesting contribution 💖

we'd definitely need to either clone it or pull multiple PRs out of it since we won't run enough tests on an external fork to accept as is

I need to think through the consequences of adding a second mechanism for bundle size reduction too (this + lazy loading of extensions)... but that's partly once we add a mechanism we're really stuck with it and partly just me being a purist 🙈

src/utils/request-router.ts
```diff
@@ -17,7 +17,7 @@ export type RequestRouterTarget = 'api' | 'ui' | 'assets'
 const ingestionDomain = 'i.posthog.com'
 
 export class RequestRouter {
-    instance: PostHog
+    private instance: PostHog
```
pauldambra, line 20: to check... have you add `private` to multiple things because it helps with bundle size?

(we're relatively careful about changing the API of things as it's hard to know what someone has coupled to)

obvs this is sort-of an OK change since JS wouldn't care, and TS would complain at build time but 👀

pauldambra:
Oh! Also, can you email me (paul AT posthog DOT com)

I want to send you a merch code for the effort regardless of how much of this we adopt! 💖

src/utils/request-router.ts
thexeos: Yes, these are all attempts at marking properties that are not currently being used anywhere else in code as "private" so that they can be safely mangled. The Rollup pipeline processes all files one-at-a-time until it gets to Babel, so there is no "big picture" to reason about what can be renamed and what must keep its names.

But again, these are just different levels of minification. Impact of renaming class members on bundle size is around ~7% in best case and ~0% in worst case.

thexeos: One thing to note: due to how es5 transpiling is done, all class member definitions turn into `Object.defineProperty`-kind of code, which is not something that Terser is able to mangle. While there is a way to mangle those as well, it is likely easier to just let it be and skip that optimization for es5 build.

robbie-c: FYI, I took a slightly different approach to this. I wanted us to have the option to mangle properties that are not part of the public API, but are used across classes/files (if TS had a `package` accessibility modifier, it would use that).

I went with a naming-convention approach, so anything prefixed with an underscore is mangled, and a linting-based approach so that anything private must be prefixed with an underscore.

Here's the linting PR https://github.com/PostHog/posthog-js/pull/1901
Here's the mangling PR https://github.com/PostHog/posthog-js/pull/1902 - @thexeos I'd love to get your eyes on this

thexeos: With TS you could use namespaces, but that's a big refactor you are looking at. It could help long-term, as more modules/extensions get added and you want to provide context isolation for those, but it is definitely outside the scope of bundle size reduction.

#1902 looks good to me. The TS transforms were basically a monkey-patch in place of establishing a code style guide and refactoring entire codebase to match it. So this approach is a good long-term solution. Your implementation also comes with the benefit of having an "audit trail" of what gets mangled, which (as a bonus) gets indexed by GitHub/Google, so anyone looking for why they can't access the property they are clearly seeing in source code can quickly figure it out without having to understand how the build process works.

PR #1873, opened 2025-04-07
feat: Only capture vitals on http and https

robbie-c:
## Changes

Check the protocol when starting up web vitals, and don't enable it if we're not on http or https. (i.e. don't use web vitals on file:// urls)

## Checklist

### DIFF
--- a/src/extensions/web-vitals/index.ts
+++ b/src/extensions/web-vitals/index.ts
@@ -3,7 +3,7 @@ import { RemoteConfig, SupportedWebVitalsMetrics } from '../../types'
 import { createLogger } from '../../utils/logger'
 import { isBoolean, isNullish, isNumber, isObject, isUndefined } from '../../utils/type-utils'
 import { WEB_VITALS_ALLOWED_METRICS, WEB_VITALS_ENABLED_SERVER_SIDE } from '../../constants'
-import { assignableWindow, window } from '../../utils/globals'
+import { assignableWindow, window, location } from '../../utils/globals'
 
 const logger = createLogger('[Web Vitals]')
 
@@ -58,6 +58,13 @@ export class WebVitalsAutocapture {
     }
 
     public get isEnabled(): boolean {
+        // Always disable web vitals if we're not on http or https
+        const protocol = location?.protocol
+        if (protocol !== 'http:' && protocol !== 'https:') {
+            return false
+        }
+
+        // Otherwise, check config
         const clientConfig = isObject(this.instance.config.capture_performance)
             ? this.instance.config.capture_performance.web_vitals
             : undefined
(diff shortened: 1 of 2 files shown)


src/extensions/web-vitals/index.ts
```diff
@@ -58,6 +58,13 @@ export class WebVitalsAutocapture {
     }
 
     public get isEnabled(): boolean {
+        // Always disable web vitals if we're not on http or https
+        const protocol = location?.protocol
+        if (protocol !== 'http:' && protocol !== 'https:') {
```
pauldambra, line 63: i _guess_ we could look and see if there are things like `capacitor:` in the prod data 

since some people might want to make this configurable or have a longer allowlist

but i don't know web vitals would work in those environments so 🙈

pauldambra:
```suggestion
        if (protocol !== 'http:' && protocol !== 'https:') {
        	logger.info("not starting etc etc")
```

maybe?

pauldambra (approved):
👍 
some waffle but otherwise LGTM

src/extensions/web-vitals/index.ts
robbie-c: > i guess we could look and see if there are things like capacitor: in the prod data

I wondered about that (and chrome extensions too), but would that tell you anything useful?

Logging makes sense, will do that.

WIll also run a quick metabase to see if there are any other protocols we should care about

pauldambra: yeah, i guess if we see protocols + data then we don't want to break it
but if we see protocols + noise then we're safe
but also don't want to default to config for every change or things get unwieldy fast

robbie-c: https://metabase.prod-us.posthog.dev/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJuYXRpdmUiOnsicXVlcnkiOiJzZWxlY3QgY291bnQoKSBhcyBjLCBwcm90b2NvbChtYXRfJGN1cnJlbnRfdXJsKSBhcyBwIGZyb20gZXZlbnRzIHdoZXJlIHRpbWVzdGFtcCA-IHRvZGF5KCkgZ3JvdXAgYnkgcCBvcmRlciBieSBjIGRlc2MiLCJ0ZW1wbGF0ZS10YWdzIjp7fX0sImRhdGFiYXNlIjo0Mn0sImRpc3BsYXkiOiJ0YWJsZSIsInBhcmFtZXRlcnMiOltdLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7fX0= FYI

robbie-c: > yeah, i guess if we see protocols + data then we don't want to break it

I'm worried about data that's in the right shape, but is meaningless and would skew vitals scores. E.g. if capacitor is loading pages from the device's local file system, that'll look like data, but the loading times would be ~instant and potentially prevent the vitals page from reporting real problems.

pauldambra: yep... 

we're talking orders of magnitude fewer events so i think safe to be a little YOLO here and can add this back for customers that turn out to be actively tracking this for some reason 🙈

rafaeelaudibert (approved):

PR #1874, opened 2025-04-07
feat: Playwright tests should run over http rather than file

robbie-c:
## Changes
Some web technologies behave differently over file:// than http://, and as our users will be using http(s), we should run tests on this.

I ran a local http server and pointed tests at that, and I fixed up the tests where file paths needed to be changed

Blocking https://github.com/PostHog/posthog-js/pull/1873

## Checklist


playwright/capture.spec.ts
```diff
@@ -52,6 +52,7 @@ test.describe('event capture', () => {
 
         // Pageview will be sent immediately
         await pollUntilEventCaptured(page, '$pageview')
+        await pollUntilCondition(page, () => captureRequests.length > 0)
```
robbie-c, line 55: this was unrelated, but this failed on a clean branch on my local machine until I added this poll here

robbie-c:
The `playwright.config.ts` changes mean it'll spin up the server automatically on start

PR #1875, opened 2025-04-07
feat: allow ANY or ALL matches for triggers

pauldambra:
pairs with remote config changes in https://github.com/PostHog/posthog/pull/30883

Current behavior of the replay ingestion controls has "grown organically" and is complicated to explain.

Users have asked for "if any trigger matches then you should record" as a behavior

This keeps minimum duration separate (since it isn't really a trigger)

currently

we start recording when `(url trigger || event trigger) && (sampling && linked flag)`
this introduces `url trigger || event trigger || sampling || linked flag`
or `url trigger && event trigger && sampling && linked flag`

as options

(it also flyby starts to extract some things from the main class which has got pretty gigantic)


src/types.ts
```diff
@@ -1306,6 +1306,7 @@ export interface RemoteConfig {
         scriptConfig?: { script?: string | undefined }
         urlBlocklist?: SessionRecordingUrlTrigger[]
         eventTriggers?: string[]
+        triggerMatchType?: 'any' | 'all' | 'og'
```
pauldambra, line 1309:
```suggestion
        triggerMatchType?: 'any' | 'all'
```

src/extensions/replay/triggerMatching.ts
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    const sampledActive = triggersStatus.isSampled === true
+    const triggerMatches = new OrTriggerMatching([
+        triggersStatus.eventTriggerMatching,
+        triggersStatus.urlTriggerMatching,
+        triggersStatus.linkedFlagMatching,
+    ]).triggerStatus(triggersStatus.sessionId)
+
+    if (sampledActive || triggerMatches === 'trigger_activated') {
```
sortafreel, line 234: I'm confused about this check. 

- I suggest `sampled` used here in a positive way (like, got in the sample, record)? As lower there's a `triggersStatus.isSampled === false` => `disabled` check.
- As `sampledActive` is checked for truthiness - what's the point of another check `sampledActive ? 'sampled' : 'active'` if we know it always true?

src/extensions/replay/triggerMatching.ts
```diff
@@ -0,0 +1,314 @@
...
+    SESSION_RECORDING_EVENT_TRIGGER_ACTIVATED_SESSION,
+    SESSION_RECORDING_URL_TRIGGER_ACTIVATED_SESSION,
+} from '../../constants'
+import { PostHog } from '../../posthog-core'
+import { FlagVariant, RemoteConfig, SessionRecordingUrlTrigger } from '../../types'
+import { isBoolean, isObject, isString } from '../../utils/type-utils'
+import { isNullish } from '../../utils/type-utils'
+import { window } from '../../utils/globals'
+
+export type TriggerType = 'url' | 'event'
```
sortafreel, line 11: I notice that in most of the cases such types are used instead of enums. I'm slightly triggered as raw strings like `trigger_activated` are used in checks and so, but it looks like it's safe enough because of TS validation. Plus, I assume these types are erased when building into the actual client script, so the size is lower?

No action is required, just thinking out loud :)

src/extensions/replay/triggerMatching.ts
```diff
@@ -0,0 +1,314 @@
...
+export type TriggerStatus = 'trigger_activated' | 'trigger_pending' | 'trigger_disabled'
```
sortafreel, line 12: Could you, pretty please, add the same description as you did for the session recording status right below? It would clarify lots of logic.

My bet would be something like this, but not sure if it's correct :) 

```ts
/**
 * Trigger status indicates the current state of a trigger in the session recording system.
 * - trigger_activated: The trigger condition has been met and recording is active for this session
 * - trigger_pending: The trigger is configured but hasn't been activated yet for this session
 * - trigger_disabled: No trigger configuration exists or the trigger is explicitly disabled
 */
 ```

src/extensions/replay/triggerMatching.ts
pauldambra: yep... they're not really raw strings when typed like this so TS protects you

https://www.totaltypescript.com/why-i-dont-like-typescript-enums says it better than i could why i prefer this approach

sortafreel (approved):

src/extensions/replay/triggerMatching.ts
pauldambra: ah, historically we report the status of the recorder by saying "sampled" when it was sampled instead of active

so inside the block we could be in because of sampling or because of triggers so we have to check again so we report the right status

iirc it's only for debug and i think i'd like to make deciding and reporting on deciding more explicit

but, here you could make it two ifs

if (sampledActive) return 'sampled'
if (triggerMatches === 'trigger_activated') return 'active

for the same result i guess

pauldambra:
Before code golf +2.85 kB (+1.79%)
After code golf +2.25 kB (+1.41%)

pauldambra:
@greptileai again

PR #1878, opened 2025-04-08
feat(segment): identify only when segment.identify() is called

ioannisj:
## Changes

Ticket: https://posthoghelp.zendesk.com/agent/tickets/28276 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/31750)

Previously, any Segment event containing a userId would trigger posthog.identify(), potentially identifying anonymous users unexpectedly. 
This change ensures that user identification only happens during explicit Segment `identify()` calls, while other events like `track`, `page`, `screen` preserve the user's anonymous state.

> Speculative fix - lacking context and know-how on testing this out

## Checklist

### DIFF
--- a/src/extensions/segment-integration.ts
+++ b/src/extensions/segment-integration.ts
@@ -68,7 +68,7 @@ const createSegmentIntegration = (posthog: PostHog): SegmentPlugin => {
         logger.warn('This browser does not have Promise support, and can not use the segment integration')
     }
 
-    const enrichEvent = (ctx: SegmentContext, eventName: string | undefined) => {
+    const enrichEvent = (ctx: SegmentContext, eventName: string | undefined, checkIdentity: boolean) => {
         if (!eventName) {
             return ctx
         }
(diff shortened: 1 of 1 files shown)


src/extensions/segment-integration.ts
```diff
@@ -77,7 +77,8 @@ const createSegmentIntegration = (posthog: PostHog): SegmentPlugin => {
             logger.info('No userId set, resetting PostHog')
             posthog.reset()
         }
-        if (ctx.event.userId && ctx.event.userId !== posthog.get_distinct_id()) {
+
+        if (checkIdentity && ctx.event.userId && ctx.event.userId !== posthog.get_distinct_id()) {
```
benjackwhite, line 81: I think the reason this was done this way was so that if the integration was added later on, the linking of distinct id would still work. 

This would now only work if you set up posthog at the same time as segment _or_ if they are sending identify events on page load or whatever. 

Probably fine but it's been a while since I've worked on this so it might be worth validating properly.

benjackwhite (approved):

PR #1879, opened 2025-04-08
chore(bundle size): getting version info without package.json

pauldambra:
in this case https://github.com/PostHog/posthog-js/pull/1871/commits/d8b8b60df25bd26b4c6a58c20251ca7b3096c3d9

this part of the change is intended to reduce the number of TS files in the dist folder

---

pros: less to download is good e.g. on train wi-fi

cons: it moves some files e.g. `/lib/src/blah` becomes `/lib/blah` so it's a breaking change

my gut is not to adopt this part of the PR (as much as it is all appreciated)
but separating things out to get other people's brains

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -16,7 +16,14 @@ const plugins = (es5) => [
         preventAssignment: true,
     }),
     resolve({ browser: true }),
-    typescript({ sourceMap: true, outDir: './dist', declaration: false }),
+    typescript({
+        sourceMap: true,
+        outDir: './dist',
+        declaration: false,
+        compilerOptions: {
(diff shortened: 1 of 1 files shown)


rafaeelaudibert (commented):
My vote is 👎🏻. I see some people using [this](https://github.com/search?q=posthog-js%2Flib%2Fsrc&type=code) in Github. 	Most of it are Posthog forks, but there is some real usage. Most of our customers are closed-source, anyway.

I don't think the benefit of reducing download size *once* - npm/yarn will cache it - is enough to warrant a breaking change on a minor.

This should be doable once/if we move to v2.0.0

PR #1880, opened 2025-04-08
chore(bundle size): strip console log messages from minimal bundle

pauldambra:

create a minimal bundle
as a first step, remove console log messages

9KB smaller without log messages ~5%
excluding the externals loader was only 1kb in comparison... will keep that in back pocket

### DIFF
--- a/src/entrypoints/array minimal.ts
+++ b/src/entrypoints/array minimal.ts
@@ -1,2 +1,1 @@
-import './external-scripts-loader'
 import './array.no-external'
--- a/src/entrypoints/array minimal.ts
+++ b/src/entrypoints/array minimal.ts
@@ -1,0 +1,2 @@
+import './external-scripts-loader'
+import './array.no-external'
--- a/src/entrypoints/array minimal.ts
+++ b/src/entrypoints/array minimal.ts
@@ -1,1 +1,2 @@
+import './external-scripts-loader'
 import './array.no-external'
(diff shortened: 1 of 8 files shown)


robbie-c:
This is cool!

I had another thought on how we could reduce log message sizes, basically copy what React logs do (used to do?) where you just get a URL with a number in it, and that URL takes you to a web page where the log message is fully explained. You could also have some guidance on the page for how to fix the error, which would normally take up way more space than you could fit in an error message.

These aren't mutually exclusive approaches, we could do this for the remaining error messages that are preserved by this change.

robbie-c (approved):

src/entrypoints/array minimal.ts
```diff
@@ -0,0 +1,2 @@
+import './external-scripts-loader'
```
robbie-c, line 1: is that a space in the file name rather than a dot? just noticed this after approving but I'm guessing it should be a dot?

pauldambra: i don't think so... or I can't see it

robbie-c: sorry I meant in `src/entrypoints/array minimal.ts`, not in `./external-scripts-loader` 🙃

pauldambra:
since we can get such a boost from minifying and have posthog-js-lite... let's skip this

PR #1881, opened 2025-04-08
chore: strip console log messages from minimal bundle

pauldambra:
stacked on top of https://github.com/PostHog/posthog-js/pull/1880

create a minimal bundle
by uglifying TS private things (which can't otherwise be mangled)


robbie-c:
See https://github.com/PostHog/posthog-js/pull/1902 for my take on this

It was more faff overall (needed lots of exceptions to the mangling rule) BUT it allows us to mangle more things overall, and have more control over what is mangled.

PR #1883, opened 2025-04-09
set host

daibhin:
## Changes

Make attribution clearer and add where missing


LICENSE
```diff
@@ -215,3 +215,30 @@ limitations under the License.
    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
    See the License for the specific language governing permissions and
    limitations under the License.
+
+
+---
+
+Portions of this codebase are derived from getsentry/sentry-javascript by Software, Inc. dba Sentry
```
robbie-c, line 222: Is it *only* the files that have a header comment (like the ones in this PR). I'm just wondering if it's useful to make it obvious which parts this applies to

rafaeelaudibert: Yeah, something akin to:

> This license only applies to files where explicitly stated in its header.

pauldambra:
Screenshot 2025-04-09 at 19 05 59
we should be careful to update from main 🤣

src/extensions/exception-autocapture/stack-trace.ts
```diff
@@ -1,5 +1,5 @@
-// copied and adapted from https://github.com/getsentry/sentry-javascript/blob/41fef4b10f3a644179b77985f00f8696c908539f/packages/browser/src/stack-parsers.ts
-// 💖open source
```
pauldambra, line 2: can we keep `💖 open source`? we do heart it :)

src/extensions/exception-autocapture/stack-trace.ts
```diff
@@ -1,5 +1,5 @@
...
+// Portions of this file are derived from getsentry/sentry-javascript by Software, Inc. dba Sentry
+// Licensed under the MIT License
```
pauldambra, line 2:
```suggestion
// Licensed under the MIT License

// 💖 open source
```

pauldambra (approved):
👍

PR #1886, opened 2025-04-10
feat: add history api autocapture pageviews

lricoy:
## Changes

This is a follow-up of https://github.com/PostHog/posthog-js-lite/pull/446, which ports the same functionality here. The tests and code are structured in a similar way but I've changed them in a attempt to make they fit the standard of this project a bit more. 

In summary, if `{ capture_pageviews: 'history_change' }` config is provided, we will send history events as page views, but only on pathname changes (no query string or hash change should trigger pageviews, even when using the back button). 

## Checklist


src/types.ts
```diff
@@ -1269,6 +1277,11 @@ export interface RemoteConfig {
      */
     capturePerformance?: boolean | PerformanceCaptureConfig
 
+    /**
+     * Whether to capture History API events (pushState, replaceState, popstate)
+     */
+    captureHistoryEvents?: boolean
+
```
pauldambra, line 1284: do we need this on remote config? 
are you adding this in team settings as well? that's super cool!

src/extensions/history-autocapture.ts
```diff
@@ -0,0 +1,129 @@
...
+        const isEnabled = () => this.isEnabled
+
+        // Only patch if not already patched
+        if (!(window.history.pushState as any)?.__posthog_wrapped__) {
+            patch(window.history, 'pushState', (originalPushState: any) => {
+                return function (this: History, ...args: any[]) {
+                    const result = originalPushState.apply(this, args)
+                    try {
+                        if (isEnabled()) {
+                            instance.capture('$pageview', { navigation_type: 'pushState' })
```
robbie-c, line 69: Just thinking through some edge cases here.

What about something like e.g. metabase where every query run updates the state? Or sites that automatically update query params when you visit (I've seen this with google ad links, e.g. when I google shoes and click the first shopping link https://arterton.co.uk/products/acme-wingtip-balmoral-oxford-8091?variant=43255266050221&utm_medium=product_sync&utm_source=google&utm_content=sag_organic&utm_campaign=sag_organic&gad_source=1&gclid=EAIaIQobChMIxsXc4PvNjAMVlJJQBh0Dew1OEAQYASABEgJ18vD_BwE it did this (it added the country code as a query param)

Maybe this should only trigger when the actual pathname changes, rather than the query or hash? 

Or maybe those users just shouldn't use this 🤷

lricoy: That is a really good point. I think most use cases would be for only the pathname, but on that same link, if I click on different sizes, it changes the query for the variant, which some people may also want to track. 

We could add an option on the config like `{ captureHistoryChange: 'always' | 'pathname' | 'never'}` instead of just a boolean but I am afraid this could grow...

Thoughts @ivanagas ?

src/types.ts
lricoy: I was actually going to ask what the pattern is here; I see most of the `extensions` are on the remote config, but I don't think we would actually **need** it there, but if it is a good thing, sure!

src/extensions/history-autocapture.ts
ivanagas: I think we should only trigger a pageview on path change if possible and keep it as a boolean. 

This seems to be what people expect. For example, the Docusaurus plugin captured when the hash changed unintentionally and people complained so [we fixed it](https://github.com/PostHog/posthog-docusaurus/pull/22). Hash or state changes are probably handled better by autocapture or custom events.

src/types.ts
pauldambra: so, you only need to add it to remote config if you want to control whether it is active remotely from team settings

we don't control pageview capture iirc from remote settings so i think safe not to add this here

src/extensions/history-autocapture.ts
lricoy: Cool! I've updated both PRs to follow that line and only capture pageviews when the path	 changes.

src/__tests__/extensions/history-autocapture.test.ts
```diff
@@ -0,0 +1,426 @@
...
+            capture,
+            config: {
+                capture_history_events: true,
+            },
+            pageViewManager: {
+                doPageView: pageViewManagerDoPageView,
+            },
+            scrollManager: {
+                resetContext: scrollManagerResetContext,
+            },
```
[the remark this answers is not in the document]
lricoy: Keeping as not using it was firing some code on PageViewManager for some reason.

src/types.ts
```diff
@@ -331,6 +331,13 @@ export interface PostHogConfig {
      */
     rageclick: boolean
 
+    /**
+     * Determines whether PostHog should automatically capture navigation events using the History API and emit them as pageviews.
+     *
+     * @default false
+     */
+    capture_history_events: boolean
```
robbie-c, line 339: How about adding another option to `capture_pageviews`, e.g. `capture_pageviews: 'history-change`?

This is what people might expect `capture_pageviews` to do anyway

lricoy: Done :)

robbie-c (approved):

lricoy:
@ivanagas @pauldambra, this looks good to be merged. Do you guys want to take another look?

lricoy:
TWIMC: I am updating the playground samples and checking some docs references before merging this, it should be good to go tomorrow

lricoy:
Merging this and updating the playground on another PR. The tests seemed fine @ivanagas :)

PR #1888, opened 2025-04-10
feat: submit partial responses for survey answers

lucasheriques:
## Changes

**don't be scared about the number of lines changed! most of them are tests**.

Part of https://github.com/PostHog/posthog/issues/19300

the actual logic is not that complicated

original RFC diagram:

```mermaid
flowchart TD
    A[Survey shown] --> B[Check store_partial_responses setting]
    B -->|Disabled| C1[User answers all questions]
    C1 --> D1[Send single 'survey sent' event when complete]
    B -->|Enabled| C2[Create unique survey_response_id]
    C2 --> D2[User answers a question]
    D2 --> E2[Send 'survey sent' event with $survey_submission_id and $survey_completed=false]
    E2 --> F2{More questions?}
    F2 -->|Yes, user continues| D2
    F2 -->|Yes, but user leaves| G2[We can rely on the already sent events to reconstruct the response]
    F2 -->|No, survey finished| H2[Send final 'survey sent' event with $survey_completed=true]
    G2 --> I2[Response shown in PostHog]
    H2 --> I2
    D1 --> J1[Response shown in PostHog]
```

The logic is almost the same, with some changes to variables names and the inclusion of the new `inProgressSurvey` logic.

How does the new in progress survey logic works:

- After sending each question, store the submission ID and responses on localStorage
- Restore that if the page is refreshed, add this condition to allow answering a survey if the survey is in progress

It also does a few addition to the 'survey dismissed' event, adding properties if an user has responded to any question before dismissing the survey.

## How we handle multiple partial responses in case of event delays, and how they are processed in the main app

```mermaid
sequenceDiagram
    participant User (Web)
    participant SDK (Web)
    participant User (Mobile)
    participant SDK (Mobile)
    participant PostHog (Events)
    participant PostHog (User Props)

    Note over User (Web), PostHog (User Props): Scenario: 3-question survey, partial responses enabled, event ingestion delayed. User hasn't responded yet ($survey_responded/survey_id = false).

    User (Web)->>SDK (Web): Opens website page matching survey conditions
    SDK (Web)->>PostHog (User Props): Check eligibility (e.g., $survey_responded/survey_id flag)
    PostHog (User Props)-->>SDK (Web): Flag is false (user hasn't responded)
    SDK (Web)->>SDK (Web): Check localStorage (isSurveyInProgress) - returns false
    SDK (Web)->>User (Web): Show Survey
    SDK (Web)->>SDK (Web): Generate surveySubmissionId: "web123"
    SDK (Web)->>SDK (Web): Store partial state in localStorage (web123, {})

    User (Web)->>SDK (Web): Answers Q1
    SDK (Web)->>SDK (Web): Update localStorage (web123, {Q1: ans})
    SDK (Web)->>PostHog (Events): Send 'survey sent' (ID: web123, Q1: ans, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, but processing/ingestion delayed...

    User (Mobile)->>SDK (Mobile): Opens app screen matching survey conditions
    SDK (Mobile)->>PostHog (User Props): Check eligibility ($survey_responded/survey_id flag)
    Note right of PostHog (User Props): Flag is still false due to ingestion delay!
    PostHog (User Props)-->>SDK (Mobile): Flag is false
    SDK (Mobile)->>SDK (Mobile): Check localStorage (isSurveyInProgress) - returns false (different device)
    SDK (Mobile)->>User (Mobile): Show Survey
    SDK (Mobile)->>SDK (Mobile): Generate surveySubmissionId: "mob456"
    SDK (Mobile)->>SDK (Mobile): Store partial state in localStorage (mob456, {})

    User (Web)->>SDK (Web): Answers Q2
    SDK (Web)->>SDK (Web): Update localStorage (web123, {Q1: ans, Q2: ans})
    SDK (Web)->>PostHog (Events): Send 'survey sent' (ID: web123, {Q1, Q2}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Mobile)->>SDK (Mobile): Answers Q1
    SDK (Mobile)->>SDK (Mobile): Update localStorage (mob456, {Q1: ans})
    SDK (Mobile)->>PostHog (Events): Send 'survey sent' (ID: mob456, {Q1}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Web)->>SDK (Web): Abandons survey (e.g., navigates away)

    User (Mobile)->>SDK (Mobile): Answers Q2
    SDK (Mobile)->>SDK (Mobile): Update localStorage (mob456, {Q1: ans, Q2: ans})
    SDK (Mobile)->>PostHog (Events): Send 'survey sent' (ID: mob456, {Q1, Q2}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Mobile)->>SDK (Mobile): Abandons survey

    loop Eventually...
        PostHog (Events)-->>PostHog (Events): Process Web Q1 event (ID: web123)
        PostHog (Events)-->>PostHog (User Props): Update user property $survey_responded/survey_id = true
        PostHog (Events)-->>PostHog (Events): Process Mobile Q1 event (ID: mob456)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
        PostHog (Events)-->>PostHog (Events): Process Web Q2 event (ID: web123)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
        PostHog (Events)-->>PostHog (Events): Process Mobile Q2 event (ID: mob456)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
    end

    Note over User (Web), PostHog (User Props): Final State:\n- PostHog has two separate partial responses:\n  - ID "web123" with Q1, Q2 answers.\n  - ID "mob456" with Q1, Q2 answers.\n- User property $survey_responded/survey_id is true.\n- Subsequent survey checks on *any* device will return false (unless survey allows repeated activation).
```

## Checklist


src/extensions/surveys.tsx
```diff
@@ -424,6 +433,16 @@ export class SurveyManager {
             : true
 
         if (!internalTargetingFlagCheck) {
+            // check if survey is in progress
+            const isSurveyInProgress = getInProgressSurveyState(survey)
+            if (isSurveyInProgress) {
+                renderReason.disabledReason = `survey is in progress`
```
marandaneto, line 439: this is odd since the property is called `disabledReason`, should this be set to null instead

marandaneto: This is not resolved

lucasheriques: I resolved because this change doesn't exist anymore, as we don't change that method and the logic is now on the survey eligibility instead

src/extensions/surveys.tsx
```diff
@@ -388,6 +390,14 @@ export class SurveyManager {
...
     private _sortSurveysByAppearanceDelay(surveys: Survey[]): Survey[] {
         return surveys.sort((a, b) => {
+            const isSurveyInProgressA = isSurveyInProgress(a)
+            const isSurveyInProgressB = isSurveyInProgress(b)
+            if (isSurveyInProgressA && !isSurveyInProgressB) {
+                return -1 // a comes before b (in progress surveys first)
+            }
+            if (!isSurveyInProgressA && isSurveyInProgressB) {
+                return 1 // a comes after b (in progress surveys first)
+            }
```
lucasheriques, line 400: changed so surveys in progress always show up first

src/extensions/surveys.tsx
```diff
@@ -990,11 +1017,30 @@ export function Questions({
...
+            if (survey.enable_partial_responses) {
+                sendSurveyEvent(
+                    { ...questionsResponses, [responseKey]: res },
+                    survey,
+                    posthog,
+                    surveySubmissionId,
+                    isSurveyCompleted
+                )
+            }
+            return
```
lucasheriques, line 1034: if partial responses is enabled, we send a survey sent event

src/extensions/surveys.tsx
```diff
@@ -956,8 +975,16 @@ export function Questions({
...
-    const [questionsResponses, setQuestionsResponses] = useState({})
-    const { previewPageIndex, onPopupSurveyDismissed, isPopup, onPreviewSubmit } = useContext(SurveyContext)
+    // Initialize responses from localStorage or empty object
+    const [questionsResponses, setQuestionsResponses] = useState(() => {
+        const inProgressSurveyData = getInProgressSurveyState(survey)
+        if (inProgressSurveyData?.responses) {
+            logger.info('Survey is already in progress, filling in initial responses')
+        }
+        return inProgressSurveyData?.responses || {}
+    })
```
lucasheriques, line 985: restore store the state of the question responses based on the data of a survey that's in progress

src/extensions/surveys.tsx
```diff
@@ -430,7 +440,11 @@ export class SurveyManager {
     }
 
     private _internalFlagCheckSatisfied(survey: Survey): boolean {
-        return canActivateRepeatedly(survey) || this._isSurveyFeatureFlagEnabled(survey.internal_targeting_flag_key)
+        return (
+            canActivateRepeatedly(survey) ||
+            this._isSurveyFeatureFlagEnabled(survey.internal_targeting_flag_key) ||
+            isSurveyInProgress(survey)
```
lucasheriques, line 446: allow surveys that are in progress to be re-submitted

src/__tests__/extensions/surveys/survey-popup.test.tsx
```diff
@@ -1,34 +1,62 @@
...
 import { Survey, SurveyQuestionType, SurveyType } from '../../../posthog-surveys-types'
+import * as uuid from '../../../uuidv7' // Import uuidv7
+
+// Mock the utility functions
+jest.mock('../../../extensions/surveys/surveys-extension-utils', () => ({
+    ...jest.requireActual('../../../extensions/surveys/surveys-extension-utils'), // Keep original implementations for non-mocked parts
+    getInProgressSurveyState: jest.fn(),
+    // No need to mock set/clear directly if we verify send/dismiss event calls
+    // setInProgressSurveyState: jest.fn(),
+    // clearInProgressSurveyState: jest.fn(),
```
marandaneto, line 14: should we just remove then instead?

src/__tests__/extensions/surveys/survey-popup.test.tsx
```diff
@@ -1,34 +1,62 @@
...
-    // Create a basic mock survey for testing
     const mockSurvey: Survey = {
-        id: 'test-survey',
-        name: 'Test Survey',
-        description: 'A test survey',
+        id: 'test-survey-partial',
+        name: 'Test Partial Survey',
+        description: 'A test survey for partial responses',
         type: SurveyType.Popover,
+        enable_partial_responses: true, // Enable partial responses for testing
```
marandaneto, line 34: should this be a different test then since the behaviour changes based on the `enable_partial_responses`?

src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -261,17 +280,41 @@ export function MultipleChoiceQuestion({
     appearance,
     onSubmit,
     onPreviewSubmit,
-}: CommonProps & {
+    initialValue,
```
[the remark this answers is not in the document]
lucasheriques: it's a choice with tradeoffs imo. Let's say I see a Survey, partially fill it, but then I only come back to it after 3-4 days.

if there was more context on the survey in the past, it's possible I won't remember anymore the responses to the previous questions.

also, if/when we implement the ability to go back to a previous question, we'd still need to store past questions to restore them anyway.

We can also start at the unanswered question (in that case question C), not sure which one is better though

src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -571,57 +572,96 @@ export const createShadow = (styleSheet: string, surveyId: string, element?: Ele
...
         $survey_id: survey.id,
         $survey_iteration: survey.current_iteration,
         $survey_iteration_start_date: survey.current_iteration_start_date,
-        $survey_questions: survey.questions.map((question, index) => ({
+        $survey_questions: survey.questions.map((question) => ({
             id: question.id,
             question: question.question,
-            index,
         })),
+        $survey_submission_id: surveySubmissionId, // Include the ID if available
```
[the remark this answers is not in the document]
lucasheriques: updated to remove comment, as I made surveySubmissionId as non null since it'll not be as it always gets generated anyway

lucasheriques:
@marandaneto for multiple devices, the survey would not show up on device B, since we only store the data a survey is inProgress on local storage. 

so once the first `survey sent` even is sent, the survey is already marked as responded for that person, so it'll not show up unless it is the same device - so it knows the survey is in progress.

the alternative to make it work on all devices would be to storing it as a person property instead, which we can do. not sure however if we charge for storing custom person properties, would have to make that as non optional. this way we can make it work for all devices.

but I think i'd release this as is, and add that as a future improvement

src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -107,13 +114,25 @@ export function RatingQuestion({
...
-    const [rating, setRating] = useState<number | null>(null)
+    const [rating, setRating] = useState<number | null>(() => {
+        if (isNumber(initialValue)) {
+            return initialValue
+        }
+        if (isArray(initialValue)) {
+            return parseInt(initialValue[0])
+        }
+        if (isString(initialValue)) {
+            return parseInt(initialValue)
```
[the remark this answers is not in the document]
lucasheriques: you're right, i updated it to handle that and also [added tests](https://github.com/PostHog/posthog-js/pull/1888/commits/86d456612788808ce6658b2aaefb74be7127d7cc)

src/__tests__/extensions/surveys/survey-popup.test.tsx
lucasheriques: not a different test, but yes should be a diffenret mock. [updated now](https://github.com/PostHog/posthog-js/pull/1888/commits/098444e1cd3f0310040c57ac05c8bd188b3bd0b1)

marandaneto:
> @marandaneto for multiple devices, the survey would not show up on device B, since we only store the data a survey is inProgress on local storage.
> 
> so once the first `survey sent` even is sent, the survey is already marked as responded for that person, so it'll not show up unless it is the same device - so it knows the survey is in progress.
> 
> the alternative to make it work on all devices would be to storing it as a person property instead, which we can do. not sure however if we charge for storing custom person properties, would have to make that as non optional. this way we can make it work for all devices.
> 
> but I think i'd release this as is, and add that as a future improvement

mm I think its worth discussing because we cannot rely on feature flags for that.
feature flags take some time to evaluate (and depends on ingesting the event in the first place).
user A device A answers 2 questions
Ingestion has a 30 min lag
user A device B sees the survey normally and might complete a survey either partially or fully.
we'll need to deduplicate those answers, get the latest, or something, otherwise it will mess up with all counts I guess?
the question here is, is that a problem? how will we mitigate the problem that the data is still correct?

marandaneto:
> @marandaneto well on that case, this is a problem that exists right now, regardless of the setup.
> 
> Say the user A responds a 1 question survey, on device A but there's a delay on the ingestion, so the feature flag is not updated yet
> 
> Then they go on device B, sees the same survey and answers it again.
> 
> **That means we now have 2 responses from the same user, on a survey where it was not possible. This is the main issue.**
> 
> On the same device we don't have that problem because we already store a seen survey in local storage too.
> 
> So it's not really a partial responses problem, this is a problem that exists today, but we haven't seen any cases of this being an issue so far

that is correct.
my point is that now we will also have mixed up partial responses from different devices, rather than the whole (and single) event twice.
the question is if our FE/BE will handle this use case nicely, or will it mess up with the results.

also, another option is a running survey that had partial responses disabled, and now it is enabled, will everything work as it should or will we not allow enabling/disabling after creation?

all of that should be considered before cutting a release to avoid breaking changes in the future, or even worse, mess up with running survey responses.

lucasheriques:
okay so here are the scenarios I though and their outcomes @marandaneto 

## Survey that has multiple responses

```mermaid
sequenceDiagram
    participant User (Web)
    participant SDK (Web)
    participant User (Mobile)
    participant SDK (Mobile)
    participant PostHog (Events)
    participant PostHog (User Props)

    Note over User (Web), PostHog (User Props): Scenario: 3-question survey, partial responses enabled, event ingestion delayed. User hasn't responded yet ($survey_responded/survey_id = false).

    User (Web)->>SDK (Web): Opens website page matching survey conditions
    SDK (Web)->>PostHog (User Props): Check eligibility (e.g., $survey_responded/survey_id flag)
    PostHog (User Props)-->>SDK (Web): Flag is false (user hasn't responded)
    SDK (Web)->>SDK (Web): Check localStorage (isSurveyInProgress) - returns false
    SDK (Web)->>User (Web): Show Survey
    SDK (Web)->>SDK (Web): Generate surveySubmissionId: "web123"
    SDK (Web)->>SDK (Web): Store partial state in localStorage (web123, {})

    User (Web)->>SDK (Web): Answers Q1
    SDK (Web)->>SDK (Web): Update localStorage (web123, {Q1: ans})
    SDK (Web)->>PostHog (Events): Send 'survey sent' (ID: web123, Q1: ans, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, but processing/ingestion delayed...

    User (Mobile)->>SDK (Mobile): Opens app screen matching survey conditions
    SDK (Mobile)->>PostHog (User Props): Check eligibility ($survey_responded/survey_id flag)
    Note right of PostHog (User Props): Flag is still false due to ingestion delay!
    PostHog (User Props)-->>SDK (Mobile): Flag is false
    SDK (Mobile)->>SDK (Mobile): Check localStorage (isSurveyInProgress) - returns false (different device)
    SDK (Mobile)->>User (Mobile): Show Survey
    SDK (Mobile)->>SDK (Mobile): Generate surveySubmissionId: "mob456"
    SDK (Mobile)->>SDK (Mobile): Store partial state in localStorage (mob456, {})

    User (Web)->>SDK (Web): Answers Q2
    SDK (Web)->>SDK (Web): Update localStorage (web123, {Q1: ans, Q2: ans})
    SDK (Web)->>PostHog (Events): Send 'survey sent' (ID: web123, {Q1, Q2}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Mobile)->>SDK (Mobile): Answers Q1
    SDK (Mobile)->>SDK (Mobile): Update localStorage (mob456, {Q1: ans})
    SDK (Mobile)->>PostHog (Events): Send 'survey sent' (ID: mob456, {Q1}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Web)->>SDK (Web): Abandons survey (e.g., navigates away)

    User (Mobile)->>SDK (Mobile): Answers Q2
    SDK (Mobile)->>SDK (Mobile): Update localStorage (mob456, {Q1: ans, Q2: ans})
    SDK (Mobile)->>PostHog (Events): Send 'survey sent' (ID: mob456, {Q1, Q2}, completed: false, $set: {$survey_responded/survey_id: true})
    Note right of PostHog (Events): Event received, processing still delayed...

    User (Mobile)->>SDK (Mobile): Abandons survey

    loop Eventually...
        PostHog (Events)-->>PostHog (Events): Process Web Q1 event (ID: web123)
        PostHog (Events)-->>PostHog (User Props): Update user property $survey_responded/survey_id = true
        PostHog (Events)-->>PostHog (Events): Process Mobile Q1 event (ID: mob456)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
        PostHog (Events)-->>PostHog (Events): Process Web Q2 event (ID: web123)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
        PostHog (Events)-->>PostHog (Events): Process Mobile Q2 event (ID: mob456)
        PostHog (Events)-->>PostHog (User Props): Update user property (idemponent or overwrites)
    end

    Note over User (Web), PostHog (User Props): Final State:\n- PostHog has two separate partial responses:\n  - ID "web123" with Q1, Q2 answers.\n  - ID "mob456" with Q1, Q2 answers.\n- User property $survey_responded/survey_id is true.\n- Subsequent survey checks on *any* device will return false (unless survey allows repeated activation).
```

And what happens in the main posthog app:

```mermaid
sequenceDiagram
    participant Events API
    participant Event Processor
    participant Event DB (ClickHouse)
    participant User Prop Store
    participant PostHog App (Survey Results View)

    Note over Events API, PostHog App (Survey Results View): Scenario: Viewing Survey Results during/after ingestion delay with multiple partial responses.

    Events API->>Event Processor: Receive Web Q1 event (ID: web123, Q1: ans, completed: false, $set: ...)
    Events API->>Event Processor: Receive Mobile Q1 event (ID: mob456, Q1: ans, completed: false, $set: ...)
    Events API->>Event Processor: Receive Web Q2 event (ID: web123, {Q1, Q2}, completed: false, $set: ...)
    Events API->>Event Processor: Receive Mobile Q2 event (ID: mob456, {Q1, Q2}, completed: false, $set: ...)

    PostHog App (Survey Results View)->>Event DB (ClickHouse): Query survey sent events for survey_id X, iteration Y (Filter/Group by surveySubmissionId)
    Event DB (ClickHouse)-->>PostHog App (Survey Results View): Returns 0 results (Events not processed yet)
    Note right of PostHog App (Survey Results View): UI shows 0 responses initially.

    Event Processor->>Event Processor: Start processing Web Q1 event (ID: web123)
    Event Processor->>Event DB (ClickHouse): Store Web Q1 event record
    Event Processor->>User Prop Store: Update user property $survey_responded/survey_id = true

    PostHog App (Survey Results View)->>Event DB (ClickHouse): Query survey sent events for survey_id X, iteration Y
    Event DB (ClickHouse)-->>PostHog App (Survey Results View): Returns 1 event group: {ID: web123, Q1: ans}
    Note right of PostHog App (Survey Results View): UI updates to show 1 partial response (web123).

    Event Processor->>Event Processor: Start processing Mobile Q1 event (ID: mob456)
    Event Processor->>Event DB (ClickHouse): Store Mobile Q1 event record
    Event Processor->>User Prop Store: Update user property (idemponent/overwrite)

    PostHog App (Survey Results View)->>Event DB (ClickHouse): Query survey sent events for survey_id X, iteration Y
    Event DB (ClickHouse)-->>PostHog App (Survey Results View): Returns 2 event groups: {ID: web123, Q1: ans}, {ID: mob456, Q1: ans}
    Note right of PostHog App (Survey Results View): UI updates to show 2 partial responses (web123 and mob456).

    Event Processor->>Event Processor: Start processing Web Q2 event (ID: web123)
    Event Processor->>Event DB (ClickHouse): Store Web Q2 event record (contains Q1+Q2 data)

    PostHog App (Survey Results View)->>Event DB (ClickHouse): Query survey sent events for survey_id X, iteration Y
    Note right of Event DB (ClickHouse): Query likely gets latest event per submission ID or aggregates. Assuming it shows latest state.
    Event DB (ClickHouse)-->>PostHog App (Survey Results View): Returns 2 event groups: {ID: web123, Q1: ans, Q2: ans}, {ID: mob456, Q1: ans}
    Note right of PostHog App (Survey Results View): UI updates, showing progress on web123 response.

    Event Processor->>Event Processor: Start processing Mobile Q2 event (ID: mob456)
    Event Processor->>Event DB (ClickHouse): Store Mobile Q2 event record (contains Q1+Q2 data)

    PostHog App (Survey Results View)->>Event DB (ClickHouse): Query survey sent events for survey_id X, iteration Y
    Event DB (ClickHouse)-->>PostHog App (Survey Results View): Returns 2 event groups: {ID: web123, Q1: ans, Q2: ans}, {ID: mob456, Q1: ans, Q2: ans}
    Note right of PostHog App (Survey Results View): Final state: UI shows 2 distinct partial responses for the same user.
```

here are the key implications for this scenario, where multiple answers are processed:

- Multiple Partial Responses: distinct surveySubmissionId entries for the same user and survey iteration, each potentially capturing different partial answers submitted around the same time from different devices.
- No data loss, but duplicated answers: we won't lose the partial data submitted from either device, but you get separate records for what was essentially the user interacting with the "same" survey instance conceptually.

In the end, we end up with 2 responses from the same user, but they are treated as distinct responses.

---

As for your other point on changing partial responses after a survey is launched.

Enabling partial responses on an existing survey is safe. The main effect is that any previously abandoned (but not dismissed) surveys will have their stored partial answers sent to PostHog as soon as the user interacts with them again on the same device after the feature is enabled. There's no inconsistencies introduced

src/extensions/surveys/components/QuestionTypes.tsx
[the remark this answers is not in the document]
lucasheriques: I took a look on what would it take to update this, but it's not as trivial, as the logic that handles which question is shown depends on:

- wether the questions are shuffled or not
- branching logic for questions

We rely right now on the question index to determine which question is shown, and compute that based on the branching logic when going forward (or shuffle it it that's enabled).

The easy but not recommended way to solve this is to store the current question index in the survey in progress state, and initiate to that when coming back. But this won't work for shuffled questions, since their order can change (unless we also store the order of the questions in the in progress state too).

Other way would be to revamp that matching logic to use question IDs instead, than we can just store the ID, but it's gonna take more work and also require changes on all SDKs. Eventually we want to do that though.

But I think for now, it's ok the user will start back at question 0, since the initial values will be there anyway

marandaneto (approved):

PR #1889, opened 2025-04-10
chore: rename survey extension utils and fix logging

lucasheriques:
## Changes

no functional changes, just moving some files and fixing the logger by unifying it in a single util.

that's necessary as the logger is not being created properly since the surveys.js file is downloaded async


marandaneto (approved):
LGTM
just pay attention on bundle size since depending on what you do, it might increase bundle size (moving from async to sync etc)
also no idea if it conflicts with https://github.com/PostHog/posthog-js/pull/1880

marandaneto:
-19kb o/

lucasheriques:
yeah I thought there was some dependencies happening here that were not ideal. I'm already doing that as a part of [survey partial responses with survey progress stored in session](https://github.com/PostHog/posthog-js/pull/1888), but to reduce files changed there decided to create a separate PR first

PR #1891, opened 2025-04-10
chore: add canRenderSurveyAsync

marandaneto:
## Changes

`canRenderSurvey` is sync and only returns a valid result if surveys were already loaded, otherwise none.
This method does not accept the `forceReload` param and can't because again its sync.
Adds a new method called canRenderSurveyAsync that accepts a callback and the forceReload params.
Deprecates the canRenderSurvey method because of the side effect, if you rely on it early in the app before the surveys were loaded, its always returning nothing, so its a broken UX.

## Checklist


src/posthog-core.ts
```diff
@@ -1370,11 +1370,28 @@ export class PostHog {
...
         return this.surveys.canRenderSurvey(surveyId)
     }
 
+    /**
+     * Checks the feature flags associated with this Survey to see if the survey can be rendered.
+     * @param surveyId The ID of the survey to check.
+     * @param callback A callback function that will be called with the SurveyRenderReason object indicating if the survey can be rendered.
+     * @param forceReload If true, the survey will be reloaded from the server, Default: false
+     */
+    canRenderSurveyAsync(surveyId: string, callback: SurveyRenderReasonCallback, forceReload = false) {
```
lucasheriques, line 1391: maybe we can add a default callback to something like:

```suggestion
    canRenderSurveyAsync(surveyId: string, callback: SurveyRenderReasonCallback = (renderReason) => alert(renderReason?.visible
                            ? `Survey can be rendered`
                            : `Survey cannot be rendered: ${renderReason?.disabledReason || 'No reason provided'}`), forceReload = false) {
```

this way we don't always need to pass the callback if we just want to quickly check the value

marandaneto: @ioannisj opinions?

marandaneto, line 1391: @pauldambra master of JS, any ideas if not just deprecating the older method and adding a new one? is the `Async` suffix a name that makes sense if its a callback method?

pauldambra: we don't support IE11 in the default bundle any more so you can use promises
we _should_ autopolyfill and can manually polyfill in the ES5 entrypoint if not

marandaneto: aha TIL

pauldambra: funnily enough I think I'd expect `Sync` in the name if it took a callback and `Async` if it returned a promise

but as above we can use promises now
so maybe that helps figure out the path forwards?

so, you could add canRenderSurveysAsync which returns a promise instead

ioannisj (approved):
Looks good to me, no comments from my part

PR #1893, opened 2025-04-11
feat: Improve how PostHogProvider works with SSR

robbie-c:
## Changes

Change the useState -> useMemo, and move the actual init call to a useEffect.

This should remove the need to do this in nextjs:

```tsx
export function PHProvider({ children }) {
  useEffect(() => {
    posthog.init(process.env.NEXT_PUBLIC_POSTHOG_KEY, {
      api_host: process.env.NEXT_PUBLIC_POSTHOG_HOST,
      person_profiles: 'identified_only',
      capture_pageview: false,
      capture_pageleave: true // Enable pageleave capture
    })
  }, []);

  return <PostHogProvider client={posthog}>{children}</PostHogProvider>
}
```
and instead, people can just do 
```tsx
export function PHProvider({ children }) {
  return <PostHogProvider
    apiKey={process.env.NEXT_PUBLIC_POSTHOG_KEY}
    options={{
      api_host: process.env.NEXT_PUBLIC_POSTHOG_HOST,
      person_profiles: 'identified_only',
      capture_pageview: false,
      capture_pageleave: true // Enable pageleave capture
    }}>{children}</PostHogProvider>
}
```

## Checklist

### DIFF
--- a/react/src/context/PostHogProvider.tsx
+++ b/react/src/context/PostHogProvider.tsx
@@ -48,7 +48,6 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     const alreadyInitializedRef = useRef<AlreadyInitialized>(false)
 
     const posthog = useMemo(() => {
-        console.log('useMemo', { client, apiKey, options, alreadyInitializedRef })
         if (client) {
             if (apiKey) {
                 console.warn(
--- a/react/src/context/PostHogProvider.tsx
+++ b/react/src/context/PostHogProvider.tsx
@@ -5,12 +5,10 @@ import React, { useEffect, useMemo, useRef } from 'react'
 import { PostHog, PostHogContext } from './PostHogContext'
 import { isDeepEqual } from '../utils/object-utils'
 
-type AlreadyInitialized =
-    | {
-          previousAPIKey: string
-          previousOptions: Partial<PostHogConfig>
-      }
-    | false
+interface PreviousInitialization {
+    apiKey: string
+    options: Partial<PostHogConfig>
+}
 
 type WithOptionalChildren<T> = T & { children?: React.ReactNode | undefined }
 
@@ -45,7 +43,7 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     // This is used to prevent double initialization when running under React.StrictMode
     // We're not storing a simple boolean here because we want to be able to detect if the
     // apiKey or options have changed.
-    const alreadyInitializedRef = useRef<AlreadyInitialized>(false)
+    const previousInitializationRef = useRef<PreviousInitialization | null>(null)
 
     const posthog = useMemo(() => {
         if (client) {
@@ -80,9 +78,9 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
             // if the user has passed their own client, assume they will also handle calling init().
             return
         }
-        const alreadyInitialized = alreadyInitializedRef.current
+        const previousInitialization = previousInitializationRef.current
 
-        if (!alreadyInitialized) {
+        if (!previousInitialization) {
             // If it's the first time running this, but it has been loaded elsewhere, warn the user about it.
             if (posthogJs.__loaded) {
                 console.warn('[PostHog.js] `posthog` was already loaded elsewhere. This may cause issues.')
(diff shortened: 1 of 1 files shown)


react/src/context/PostHogProvider.tsx
```diff
@@ -45,60 +45,45 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     // This is used to prevent double initialization when running under React.StrictMode
     // We're not storing a simple boolean here because we want to be able to detect if the
     // apiKey or options have changed.
-    const [alreadyInitialized, setAlreadyInitialized] = useState<AlreadyInitialized>(false)
+    const alreadyInitializedRef = useRef<AlreadyInitialized>(false)
 
     const posthog = useMemo(() => {
+        console.log('useMemo', { client, apiKey, options, alreadyInitializedRef })
```
pauldambra, line 51:
```suggestion
```

robbie-c: god damnit

react/src/context/PostHogProvider.tsx
```diff
@@ -45,7 +45,7 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     // This is used to prevent double initialization when running under React.StrictMode
     // We're not storing a simple boolean here because we want to be able to detect if the
     // apiKey or options have changed.
-    const [alreadyInitialized, setAlreadyInitialized] = useState<AlreadyInitialized>(false)
+    const alreadyInitializedRef = useRef<AlreadyInitialized>(false)
```
pauldambra, line 48: we say it's not a simple bool then init it to false... should it be `AlreadyInitialized | null` instead?

robbie-c: Yeah happy to make that change! (I didn't change this part, but yeah I agree that it'd be neater)

pauldambra: the fun of properly reading things for the first time 🙈

pauldambra (approved):
makes me think we should have ssr in the playground example so we can test this...

but all seems to compile and run so 🚢

robbie-c:
> makes me think we should have ssr in the playground example so we can test this...

I used the react-router example, which does do SSR!

lricoy:
I think those failing opt-in tests are flaky; I made a note to look at them; they will probably pass if you run them again. Locally they "fix" themselves if I rebuild the lib.

lricoy (approved):

react/src/context/PostHogProvider.tsx
```diff
@@ -45,7 +43,7 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
     // This is used to prevent double initialization when running under React.StrictMode
     // We're not storing a simple boolean here because we want to be able to detect if the
     // apiKey or options have changed.
-    const [alreadyInitialized, setAlreadyInitialized] = useState<AlreadyInitialized>(false)
+    const previousInitializationRef = useRef<PreviousInitialization | null>(null)
```
rafaeelaudibert, line 46: Since you're not matching against the `ref` object directly, you're checking the internal props, this makes more sense as a `useState`, don't you think? Does that break SSR?

react/src/context/PostHogProvider.tsx
```diff
@@ -108,20 +90,38 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
             posthogJs.init(apiKey, options)
```
rafaeelaudibert, line 90: Inside the `useEffect` you must call `posthog` rather `posthogJs` because we wanna initialize *that* client. I think this is working right now because we always return `posthogJs` from inside of it, but it makes more sense semantically to init that instance, right?

react/src/context/PostHogProvider.tsx
```diff
@@ -54,51 +52,35 @@ export function PostHogProvider({ children, client, apiKey, options }: WithOptio
...
+        )
+        return posthogJs
+    }, [client, apiKey, JSON.stringify(options)]) // Stringify options to be a stable reference
 
-                // Return the same already-initialized global client
-                return posthogJs
-            }
+    // TRICKY: The init needs to happen in a useEffect rather than useMemo, as useEffect does not happen during SSR. Otherwise
+    // we'd end up trying to call posthogJs.init() on the server, which can cause issues around hydration and double-init.
+    useEffect(() => {
```
rafaeelaudibert, line 76: Since we have a `useEffect`, is there anything we should be doing when this provider is unmounted? Maybe we should call `reset`? Not sure if it makes sense as a default tho

react/src/context/PostHogProvider.tsx
rafaeelaudibert: Oh, how I'd love to have a conversation with Guillermo/someone at Meta, this is just nonsense.

React: be reactive
Also React: force you to use declarative refs to work around it being reactive

PR #1896, opened 2025-04-11
WIP: Update nextjs playground to show how in-app to browser conversions might work

robbie-c:
## Changes
Update nextjs playground to show how in-app to browser conversions might work.

Essentially we just make sure that we keep the distinct id and session id in the URL, so that if it's opened in another browser, we bring that state with us

## Checklist

### DIFF
--- a/playground/nextjs/pages/mobile-app-attribution.tsx
+++ b/playground/nextjs/pages/mobile-app-attribution.tsx
@@ -1,0 +1,26 @@
+import { useEffect } from 'react'
+import { posthog } from '@/src/posthog'
+
+const MobileAppAttribution = () => {
+    useEffect(() => {
+        if (
+            navigator.userAgent.toLowerCase().includes('linkedin') ||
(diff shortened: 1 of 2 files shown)


playground/nextjs/pages/mobile-app-attribution.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+const MobileAppAttribution = () => {
+    useEffect(() => {
+        if (
+            navigator.userAgent.toLowerCase().includes('linkedin') ||
+            navigator.userAgent.toLowerCase().includes('facebook')
+        ) {
+            const params = new URLSearchParams(window.location.search)
+            params.set('__ph_distinct_id', posthog.get_distinct_id())
+            params.set('__ph_is_identified', posthog._isIdentified() ? 'true' : 'false')
```
robbie-c, line 12: We should remove the _ from this before telling people to use it, we strongly discourage people from using function that begin with _ as they are not part of the public interface and might change

PR #1897, opened 2025-04-11
fix: refactor canRenderSurvey and move logic to SurveyManager

lucasheriques:
## Changes

both `getActiveMatchingSurveys` and `canRenderSurvey` share a lot of similarities around business logic. Since their goal is to get check if a survey is eligible for an user or not.

However, we had duplicated business logic happening on both.

So, I decided to remove `canRenderSurvey` from `posthog-surveys.ts` to `extenstions/surveys.tsx`, to centralize the logic in one place and also reduce the main bundle size, by only downloading it if there are surveys enabled.

I also did the same to `getActiveMatchingSurveys`. No need to have this logic be on `posthog-surveys.ts` since it's only relevant if surveys are actually loaded.

## Checklist


src/__tests__/extensions/surveys.test.ts
```diff
@@ -552,74 +552,6 @@ describe('SurveyManager', () => {
         })
     })
 
-    describe('canRenderSurvey', () => {
```
lucasheriques, line 555: tests removed as I moved that method to `posthog-surveys` to stop duplicating rendering logic

src/posthog-surveys.ts
```diff
@@ -333,40 +321,75 @@ export class PostHogSurveys {
...
-            if (survey) {
-                renderReason = { ...this._surveyManager.canRenderSurvey(survey) }
-            } else {
-                renderReason = { visible: false, disabledReason: 'Survey not found' }
-            }
+            survey = surveys.find((x) => x.id === surveyId) ?? null
         })
-        return renderReason
+        return survey
     }
```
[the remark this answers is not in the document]
lucasheriques: added a check to only run if surveys functionality is already loaded

src/posthog-core.ts
```diff
@@ -1365,6 +1365,10 @@ export class PostHog {
         this.surveys.getActiveMatchingSurveys(callback, forceReload)
     }
 
+    checkSurveyEligibility(surveyId: string): { eligible: boolean; reason?: string } {
```
marandaneto, line 1368: I'd not increase the API surface since this is exactly what `canRenderSurvey` does already

marandaneto:
> So, I decided to remove canRenderSurvey from the extensions/survey.tsx file, and move it to posthog-surveys.ts directly. While this slightly increases the bundle size for customers who are not using surveys (434 bytes), it reduces the overall size when surveys are loaded.

> Plus, this centralizes the logic on wether or not a survey is eligible in a single function.

can we do it the other way around? so we get the refactoring gains without increasing the main bundle?

lucasheriques:
@marandaneto @ioannisj I decided to move more business logic to the `SurveyManager` instead, so we can better centralize the logic and also reduce the main bundle size. Now it has -2KB, which are only downloaded if we actually load surveys functionality.

src/posthog-core.ts
```diff
@@ -14,6 +14,7 @@ import {
 import { DeadClicksAutocapture, isDeadClicksEnabledForAutocapture } from './extensions/dead-clicks-autocapture'
 import { ExceptionObserver } from './extensions/exception-autocapture'
 import { errorToProperties } from './extensions/exception-autocapture/error-conversion'
+import { HistoryAutocapture } from './extensions/history-autocapture'
```
marandaneto, line 17: Most of your PRs auto-format the imports, which adds noise to the PR review.
Can you double-check that your auto-format config is the same as everyone else's?

src/posthog-surveys.ts
```diff
@@ -368,9 +305,16 @@ export class PostHogSurveys {
...
-            this._surveyManager.renderSurvey(survey, document?.querySelector(selector))
-        })
+        const survey = this._getSurveyById(surveyId)
+        const elem = document?.querySelector(selector)
+        if (!survey) {
+            logger.warn('Survey not found')
+            return
+        }
+        if (!elem) {
+            logger.warn('Element not found')
```
marandaneto, line 315:
```suggestion
            logger.warn('Survey element not found')
```

marandaneto (approved):

marandaneto:
> @marandaneto @ioannisj I decided to move more business logic to the `SurveyManager` instead, so we can better centralize the logic and also reduce the main bundle size. Now it has -2KB, which are only downloaded if we actually load surveys functionality.

nice, left 2 comments otherwise LGTM

src/posthog-core.ts
lucasheriques: it's caused by a setting I have on VS Code. I [created a PR](https://github.com/PostHog/posthog-js/pull/1914) to prevent it from having effect in this repo

PR #1900, opened 2025-04-12
chore: Remove giant Info object

robbie-c:
## Changes

This allows the minifier (terser) to do a much better job, as it doesn't need to worry about preserving property names. We should see a reduced bundle size, as none of our long function names actually appear in the bundle any more, and dead code elimination works

## Checklist


src/__tests__/utils/user-agent-utils.test.ts
```diff
@@ -0,0 +1,317 @@
+import uaParserDeviceTestCases from './device.test.json'
```
robbie-c, line 1: these tests are moved (but otherwise unchanged) from event-utils.test.ts

src/__tests__/utils/event-utils.test.ts
```diff
@@ -28,348 +30,34 @@ describe(`event-utils`, () => {
...
-            const properties = Info.properties({
-                maskPersonalDataProperties: true,
-                customPersonalDataProperties: ['other'],
-            })
+            const properties = getEventProperties(true, ['other'])
             expect(properties['$current_url']).toEqual('https://www.example.com/path?gclid=<masked>&other=<masked>')
         })
     })
 
-    describe('user agent', () => {
```
robbie-c, line 54: moved to user-agent-utils.test.ts

src/utils/event-utils.ts
```diff
@@ -70,261 +70,205 @@ export const EVENT_TO_PERSON_PROPERTIES = [
...
+        : []
 
-        return this._campaignParamsFromUrl(maskQueryParams(document.URL, paramsToMask, MASKED), customTrackedParams)
-    },
+    return _getCampaignParamsFromUrl(maskQueryParams(document.URL, paramsToMask, MASKED), customTrackedParams)
+}
 
-    _campaignParamsFromUrl: function (url: string, customParams?: string[]): Record<string, string> {
-        const campaign_keywords = CAMPAIGN_PARAMS.concat(customParams || [])
+function _getCampaignParamsFromUrl(url: string, customParams?: string[]): Record<string, string> {
```
rafaeelaudibert, line 89: Does it even make sense to prefix them with an underscore? Not exporting them should be enough to indicate they're private

src/__tests__/utils/event-utils.test.ts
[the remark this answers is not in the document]
robbie-c: True, I'm making another commit to this stuff, so I'll follow up there

src/utils/event-utils.ts
robbie-c: Yeah agreed, I'll do that in the follow-up

PR #1901, opened 2025-04-14
feat: Enforce that private members of classes start with an underscore

robbie-c:
## Changes

This is not immediately useful on its own, but in the future this would allow us to use terser's property mangler on these properties to signfiicantly reduce the bundle size.

See https://github.com/PostHog/posthog-js/pull/1902 for where number go down

## Checklist


rafaeelaudibert (approved):
+1 from me, but I'm not known for being careful with my changes, so wait until you have the right set of approves, I like the convention, though!

PR #1902, opened 2025-04-14
feat: Reduce bundle size with property mangling

robbie-c:
## Changes

DO NOT MERGE unless you're me. I want to get eyes on it from all teams who touch posthog-js

TLDR I saved 13% of the bundle size of posthog-js using property mangling, I want to get an Nth opinion on whether this is a good idea as property mangling is a bit risky

We use terser to minify our code, but by default, it preserves class property names, so our minified bundle has lots of strings like `posthog.someLongClassInstanceName.someEvenLongerPropertyNameThatSoundsCool=42` . Terser can reduce that down to `posthog.a.b=42` but you have to tell it where it's safe to do so.

Here's what I did:
* I added an eslint rule where all private properties/methods had to be prefixed with _ and schlepped through the codebase and fixed it (in a different PR [here](https://github.com/PostHog/posthog-js/pull/1901))
* I added a property mangler rule to mangle all properties that start with a single underscore _ (and not a double underscore __ as we use these for some complex internals)
* I excluded from the mangler any properties that seemed to need to be preserved. This includes anything that is referenced in the snippet, and some things in setup/teardown. See PRs for the list

See https://posthog.slack.com/archives/C03P7NL6RMW/p1744652082051639 for internal discussion.

Check the file in the diff `terser-mangled-names.json` to see the affected property names.

## Checklist


marandaneto:
@lucasheriques quite a few survey files were affected. Please pair up with @robbie-c to make sure there's nothing broken.
We can also split the testing if it makes sense.

rollup.config.js
```diff
@@ -51,9 +52,88 @@ const plugins = (es5) => [
...
         toplevel: true,
         compress: {
-            // 5 is the default if unspecified
             ecma: es5 ? 5 : 6,
         },
+        mangle: es5
+            ? undefined // Don't mangle properties in the es5 build, as it relies on helpers with don't work well with mangling.
+            : {
+                  // Note:
+                  // PROPERTY MANGLING CAN BREAK YOUR CODE
```
daibhin, line 61: Without knowing much about how this works it would be super useful if we could let people know somehow that they've written a function that will be mangled (and possibly make them safe list it). It feels like that's the best way to protect against peoples code being minified unintendedly (not sure that's a word)

daibhin: The latter would be great but probably not necessary if it is too much hassle. I just worry that people don't necessarily see the README and end up dealing with the consequences

daibhin: Maybe even an eslint rule would suffice if there's a prewritten one

robbie-c: > The latter would be great
Done! :)

daibhin (approved):

lucasheriques (commented):
Didn't take a look on the code just yet, but I tested this with the Nextjs playground and didn't find anything not working as expected. Will test more today and also review the PR later, but for me this looks great and it's a insane win

Amazing job @robbie-c

rollup.config.js
```diff
@@ -49,12 +55,172 @@ const plugins = (es5) => [
...
+                  // But we use it anyway because it's incredible for bundle size, you just need to develop with it in mind.
+                  // Any properties that start with _ will be mangled, which can be a problem if anything with that pattern is
+                  // part of the public interface, or if any API responses we use matches that regex.
+                  // Fix specific instances of this by adding the property to the reserved list.
+                  properties: {
+                      regex: /^_(?!_)/, // only mangle properties that start with a single _
+                      reserved: [
```
pauldambra, line 77: :chefs-kiss:

robbie-c:
@pauldambra you mentioned doing an alpha build of this for local testing, I vaguely remember you doing this before.

My approach would be to create a new CI workflow to handle this based on a label (ideally I'd want to make the alpha build without merging the PR, so it'd run on every commit before the PR is merged), sanity check that I'm not missing some existing tooling?

pauldambra:
Screenshot 2025-04-28 at 15 09 18

i added this... iirc if you toggle off and back on it'll do a release

was the fastest thing so I could test 🙈

robbie-c:
That's perfect!

PR #1910, opened 2025-04-21
feat: Avoid tracking pageview from prerenders

rafaeelaudibert:
Chrome and some frameworks such as Wordpress prerender a page because they want the loading time to be fast. We were, however, including that render (which runs Javascript) as a pageview even though that page might never actually be shown to the user.

We're now being slightly more cautious about it and not triggering the pageview when we can safely detect the page isnt visible. We then attach a listener and only ever trigger the `pageview` in case the page is actually displayed

Inspired by the behavior in https://github.com/plausible/analytics/blob/beb3ae087b7e7310a13cb012d109ab38d31301d5/tracker/src/plausible.js

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -293,6 +293,7 @@ export class PostHog {
     webPerformance = new DeprecatedWebPerformanceObserver()
 
     _initialPageviewCaptured: boolean
+    _visibilityStateListener: (() => void) | null
     _personProcessingSetOncePropertiesSent: boolean = false
     _triggered_notifs: any
     compression?: Compression
@@ -327,6 +328,7 @@ export class PostHog {
         this.__loaded = false
         this.analyticsDefaultEndpoint = '/e/'
         this._initialPageviewCaptured = false
+        this._visibilityStateListener = null
         this._initialPersonProfilesConfig = null
         this._cachedPersonProperties = null
         this.featureFlags = new PostHogFeatureFlags(this)
@@ -2154,9 +2156,34 @@ export class PostHog {
     }
 
     _captureInitialPageview(): void {
-        if (document && !this._initialPageviewCaptured) {
+        if (!document) {
+            return
+        }
+
+        // If page is not visible, add a listener to detect when the page becomes visible
+        // and trigger the pageview only then
+        // This is useful to avoid `prerender` calls from Chrome/Wordpress/SPAs
+        // that are not visible to the user
+
+        if (document.visibilityState !== 'visible') {
+            if (!this._visibilityStateListener) {
+                this._visibilityStateListener = this._captureInitialPageview.bind(this)
+                addEventListener(document, 'visibilitychange', this._visibilityStateListener)
+            }
+
+            return
+        }
+
+        // Extra check here to guarantee we only ever trigger a single `$pageview` event
+        if (!this._initialPageviewCaptured) {
             this._initialPageviewCaptured = true
             this.capture('$pageview', { title: document.title }, { send_instantly: true })
+
+            // After we've captured the initial pageview, we can remove the listener
+            if (this._visibilityStateListener) {
+                document.removeEventListener('visibilitychange', this._visibilityStateListener)
+                this._visibilityStateListener = null
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/posthog-core.ts
```diff
@@ -2152,9 +2154,32 @@ export class PostHog {
...
+        // If page is not visible, add a listener to detect when the page becomes visible
+        // and trigger the pageview only then
+        // This is useful to avoid `prerender` calls from Chrome/Wordpress/SPAs
+        // that are not visible to the user
+
+        if (document.visibilityState !== 'visible') {
+            if (!this._hasVisibilityStateListener) {
+                addEventListener(document, 'visibilitychange', this._captureInitialPageview.bind(this))
+                this._hasVisibilityStateListener = true
+            }
```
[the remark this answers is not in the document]
rafaeelaudibert: ah, yuck

src/posthog-core.ts
```diff
@@ -2152,9 +2154,34 @@ export class PostHog {
     }
 
     _captureInitialPageview(): void {
```
rafaeelaudibert, line 2156: And yes, I'm aware this is a lot of new bytes 😢

src/posthog-core.ts
```diff
@@ -2152,9 +2154,34 @@ export class PostHog {
...
+        // If page is not visible, add a listener to detect when the page becomes visible
+        // and trigger the pageview only then
+        // This is useful to avoid `prerender` calls from Chrome/Wordpress/SPAs
+        // that are not visible to the user
+
+        if (document.visibilityState !== 'visible') {
+            if (!this._visibilityStateListener) {
+                this._visibilityStateListener = this._captureInitialPageview.bind(this)
+                addEventListener(document, 'visibilitychange', this._visibilityStateListener)
+            }
```
marandaneto, line 2170: what if you do this during the SDK init instead? so the listener is ready once `_captureInitialPageview` is called

rafaeelaudibert: Hm, yeah. I thought adding this here made it simpler and simplified the main function's concern, but now that I need this flag it might make sense to put it back there, I'll reconsider

rafaeelaudibert: Ok, I cant do this because there are two places that call this function: either on `init` or when calls `opt_in_capturing` so it makes sense to keep this here.

rafaeelaudibert:
@pauldambra @robbie-c mind taking a look here?

src/posthog-core.ts
```diff
@@ -293,6 +293,7 @@ export class PostHog {
     webPerformance = new DeprecatedWebPerformanceObserver()
 
     _initialPageviewCaptured: boolean
+    _visibilityStateListener: (() => void) | null
```
pauldambra, line 296: funnily enough i'm pretty sure we have at least two visibility state listeners now
i think maybe this is the third or fourth

so we could have a single listener that uses the event emitter which the other places react to

that _might_ be smaller (but is of course more complex ⚖️)

src/posthog-core.ts
```diff
@@ -2154,9 +2156,34 @@ export class PostHog {
...
+        if (!document) {
+            return
+        }
+
+        // If page is not visible, add a listener to detect when the page becomes visible
+        // and trigger the pageview only then
+        // This is useful to avoid `prerender` calls from Chrome/Wordpress/SPAs
+        // that are not visible to the user
+
+        if (document.visibilityState !== 'visible') {
```
pauldambra, line 2168: much prefer this check btw
very conflicting information out there about whether `preprender` is a valid value, it's certainly not in the spec

(i wonder if we should have visibility state by default on events)

rafaeelaudibert: Yeah, types actually fail if I try using `prerender`, so this is definitely the right way to do

src/posthog-core.ts
rafaeelaudibert: I could keep calling this function on **any** visibility change because it's already guaranteeing that we only send the event once, so... maybe? I'll check

robbie-c (approved):
LGTM - at least based on how the docs say it should work

src/posthog-core.ts
rafaeelaudibert: This is actually just the second visibility listener, so I'm not too worried

PR #1912, opened 2025-04-21
fix: prevent multiple stylesheet creation for feedback survey

lucasheriques:
## Changes

Fixes https://posthoghelp.zendesk.com/agent/tickets/29337 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/30939)

Surveys are evaluated for display logic every second.

However, Feedback Button/Widget surveys are mounted on the DOM once - which is intended, as they're more permanent.

Yet, every second we're adding the stylesheet again, since the survey matches the condition. We don't have to worry about the DOM element itself since Preact takes care of that - but we should prevent adding multiple <style> tags.

## Checklist

### DIFF
--- a/src/extensions/surveys.tsx
+++ b/src/extensions/surveys.tsx
@@ -246,10 +246,16 @@ export class SurveyManager {
         // Ensure widget container exists if it doesn't
         const shadow = retrieveWidgetShadow(survey, this._posthog)
         const stylesheetContent = style(survey.appearance)
-        const stylesheet = prepareStylesheet(document, stylesheetContent, this._posthog)
+        const styleId = 'ph-survey-widget-style' // Identifier for the style element
 
-        if (stylesheet) {
-            shadow.appendChild(stylesheet)
+        // Check if the stylesheet already exists
+        if (!shadow.querySelector(`style[data-style-id="${styleId}"]`)) {
+            const stylesheet = prepareStylesheet(document, stylesheetContent, this._posthog)
+
+            if (stylesheet) {
+                stylesheet.setAttribute('data-style-id', styleId) // Add identifier
+                shadow.appendChild(stylesheet)
+            }
         }
 
         Preact.render(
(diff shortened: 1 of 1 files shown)


src/extensions/surveys.tsx
```diff
@@ -246,10 +246,16 @@ export class SurveyManager {
...
-            shadow.appendChild(stylesheet)
+        // Check if the stylesheet already exists
+        if (!shadow.querySelector(`style[data-style-id="${styleId}"]`)) {
+            const stylesheet = prepareStylesheet(document, stylesheetContent, this._posthog)
+
+            if (stylesheet) {
+                stylesheet.setAttribute('data-style-id', styleId) // Add identifier
+                shadow.appendChild(stylesheet)
+            }
         }
```
rafaeelaudibert, line 259: Can we:
1. Extract `data-style-id` to a separate constant to avoid duplicating text and increasing bundle size?
2. Call it `ph-data-style-id` instead?

rafaeelaudibert (changes requested):
see comment

src/extensions/surveys.tsx
lucasheriques: done

src/extensions/surveys.tsx
```diff
@@ -246,10 +246,16 @@ export class SurveyManager {
         // Ensure widget container exists if it doesn't
         const shadow = retrieveWidgetShadow(survey, this._posthog)
         const stylesheetContent = style(survey.appearance)
-        const stylesheet = prepareStylesheet(document, stylesheetContent, this._posthog)
+        const WIDGET_STYLE_ID = 'ph-data-style-id'
```
rafaeelaudibert, line 249: Does it make sense to set it as `ph-data-style-id="ph-data-style-id"`? Same key/value? If so, LGTM!

rafaeelaudibert (approved):

src/extensions/surveys.tsx
lucasheriques: yeah no need to have different values, just need a consistent id to check if it's there

PR #1921, opened 2025-04-25
chore: survey 10kb win

lucasheriques:
## Changes

bundle size without surveys 📉

### DIFF
--- a/src/posthog-surveys-types.ts
+++ b/src/posthog-surveys-types.ts
@@ -199,6 +199,7 @@ export interface Survey {
     current_iteration: number | null
     current_iteration_start_date: string | null
     schedule?: SurveySchedule | null
+    enable_partial_responses?: boolean | null
 }
 
 export interface SurveyActionType {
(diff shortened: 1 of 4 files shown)


daibhin (approved):

daibhin:
Nice!! Would be cool if we could lint against things like this... one for another time

src/posthog-surveys-types.ts
```diff
@@ -199,6 +199,7 @@ export interface Survey {
     current_iteration: number | null
     current_iteration_start_date: string | null
     schedule?: SurveySchedule | null
+    enable_partial_responses?: boolean | null
```
pauldambra, line 202: this feels separate to this change... am i being silly

src/posthog-surveys.ts
```diff
@@ -1,6 +1,5 @@
 import { SURVEYS } from './constants'
 import { SurveyManager } from './extensions/surveys'
-import { getSurveySeenStorageKeys } from './extensions/surveys/surveys-extension-utils'
```
pauldambra, line 3: you're right @daibhin this should be lintable...

src/posthog-surveys-types.ts
lucasheriques: it is, but needed anyway since we already have the property on the BE on main posthog so wanted to get it in already

src/posthog-surveys.ts
lucasheriques: yeah something like not allowing importing from ./extensions/* from main posthog-* files

PR #1924, opened 2025-04-27
fix: prevent survey events to affect multiple surveys

lucasheriques:
## Changes

There's a bug right now that, if a survey sent/dismissed event is sent, it closes any other surveys that might be visible.

While we only support one popup survey being shown, for feedback surveys there's no such restriction.

So this PR changes the event emitters for PHSurveySent and PHSurveyClosed to also check for the survey id to only close the survey in the event context. This is tested using a e2e test.

There's also another change, in preparation for a bigger one: it changes the survey class container name from being `PostHogWidget${surveyId}` or `PostHogSurvey${surveyId}` depending on the type to be a unified `PostHogSurvey-${surveyId}`.

This is needed to have constant class names between surveys no matter the type. There's no functional changes. But it does require updating most of our e2e tests to the new value.

## Checklist

### DIFF
--- a/playwright/surveys/feedback-widget.spec.ts
+++ b/playwright/surveys/feedback-widget.spec.ts
@@ -77,16 +77,16 @@ test.describe('surveys - feedback widget', () => {
         await start(startOptions, page, context)
         await surveysAPICall
 
-        await expect(page.locator('.PostHogWidget123').locator('.survey-form')).not.toBeVisible()
-        await page.locator('.PostHogWidget123').locator('.ph-survey-widget-tab').click()
-        await expect(page.locator('.PostHogWidget123').locator('.survey-form')).toBeVisible()
-        await expect(page.locator('.PostHogWidget123').locator('.survey-question')).toHaveText('Feedback for us?')
-        await expect(page.locator('.PostHogWidget123').locator('.survey-question-description')).toHaveText(
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-form')).not.toBeVisible()
+        await page.locator('.PostHogSurvey-123').locator('.ph-survey-widget-tab').click()
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-form')).toBeVisible()
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-question')).toHaveText('Feedback for us?')
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-question-description')).toHaveText(
             'tab feedback widget'
         )
 
-        await page.locator('.PostHogWidget123').locator('.survey-form').locator('textarea').fill('hello posthog!')
-        await page.locator('.PostHogWidget123').locator('.survey-form').locator('.form-submit').click()
+        await page.locator('.PostHogSurvey-123').locator('.survey-form').locator('textarea').fill('hello posthog!')
+        await page.locator('.PostHogSurvey-123').locator('.survey-form').locator('.form-submit').click()
         await pollUntilEventCaptured(page, 'survey sent')
     })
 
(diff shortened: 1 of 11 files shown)


playwright/surveys/feedback-widget.spec.ts
```diff
@@ -279,41 +280,161 @@ test.describe('surveys - feedback widget', () => {
...
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-form')).toBeVisible()
+
+        // close first survey
+        await page.locator('.PostHogSurvey-123').locator('.survey-form').locator('.form-cancel').click()
+
+        await pollUntilEventCaptured(page, 'survey dismissed')
+
+        // check if the second survey is still visible
+        await expect(page.locator('.PostHogSurvey-123').locator('.survey-form')).not.toBeVisible()
+    })
```
lucasheriques, line 439: there are the tests for the bug. other changes are just the selector changes

playwright/surveys/thank-you-message.spec.ts
```diff
@@ -74,12 +74,12 @@ test.describe('surveys - feedback widget', () => {
...
-        await expect(page.locator('.PostHogSurvey123 .ratings-emoji')).toHaveCount(5)
-        await page.locator('.PostHogSurvey123 .ratings-emoji').first().click()
-        await page.locator('.PostHogSurvey123 .form-submit').click()
+        await expect(page.locator('.PostHogSurvey-123 .ratings-emoji')).toHaveCount(5)
+        await page.locator('.PostHogSurvey-123 .ratings-emoji').first().click()
+        await page.locator('.PostHogSurvey-123 .form-submit').click()
 
-        await expect(page.locator('.PostHogSurvey123 .thank-you-message')).toBeVisible()
+        await expect(page.locator('.PostHogSurvey-123 .thank-you-message')).toBeVisible()
         await page.waitForTimeout(5000)
```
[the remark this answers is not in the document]
lucasheriques: no need for tests

playwright/surveys/question-types.spec.ts
```diff
@@ -139,18 +139,18 @@ test.describe('surveys - core display logic', () => {
...
+        await page.locator('.PostHogSurvey-12345').locator('#surveyQuestion0Choice1').click()
+        await page.locator('.PostHogSurvey-12345').locator('#surveyQuestion0Choice2').click()
+        await page.locator('.PostHogSurvey-12345').locator('.form-submit').click()
 
-        await page.locator('.PostHogSurvey12345').locator('textarea').type('Great job!')
-        await page.locator('.PostHogSurvey12345').locator('.form-submit').click()
-        await page.locator('.PostHogSurvey12345').locator('.form-submit').click()
+        await page.locator('.PostHogSurvey-12345').locator('textarea').type('Great job!')
+        await page.locator('.PostHogSurvey-12345').locator('.form-submit').click()
+        await page.locator('.PostHogSurvey-12345').locator('.form-submit').click()
```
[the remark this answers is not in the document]
lucasheriques: this is expected

playwright/surveys/feedback-widget.spec.ts
```diff
@@ -279,41 +280,161 @@ test.describe('surveys - feedback widget', () => {
...
 
         // Verify thank you message appears again
-        await expect(page.locator('.PostHogWidget123').locator('.thank-you-message-header')).toBeVisible()
+        await expect(page.locator('.PostHogSurvey-123').locator('.thank-you-message-header')).toBeVisible()
 
         // Verify second event was sent
         await pollUntilEventCaptured(page, 'survey sent')
     })
+
+    test('if multiple surveys being shown, sending one of them does not close the other one', async ({
```
lucasheriques, line 321: start of the new tests that handles the bug

playwright/surveys/feedback-widget.spec.ts
lucasheriques, line 439: end of the new tests that handles the bug

src/extensions/surveys.tsx
```diff
@@ -817,32 +820,32 @@ export function usePopupVisibility(
             localStorage.setItem('lastSeenSurveyDate', new Date().toISOString())
             setTimeout(() => {
                 const inputField = document
-                    .querySelector(getPosthogWidgetClass(survey.id))
+                    .querySelector(`.${getSurveyContainerClass(survey)}`)
```
marandaneto, line 823: should `getSurveyContainerClass` already start with the `.` so you don't have to concatenate in multiple places? it was like this before as well

marandaneto (approved):
left a question otherwise LGTM

src/extensions/surveys.tsx
lucasheriques: yeah there's two places where it requires the . vs one that doesn't, so will change

lucasheriques: i changed instead so the function has the option to return the [class as a selector](https://github.com/PostHog/posthog-js/pull/1924/commits/e1f6dca0de4a4ec8734e7c0b4db6a85756a9be6f)

PR #1928, opened 2025-04-29
chore: Add homepage fields and copy LICENSE to posthog-js-react

ordehi:
## Changes

Add LICENSE file to react/ and set homepage in /package.json and react/package.json. This improves license compliance and package metadata clarity without affecting runtime behavior.


package.json
```diff
@@ -4,7 +4,8 @@
     "description": "Posthog-js allows you to automatically capture usage and send events to PostHog.",
     "repository": "https://github.com/PostHog/posthog-js",
     "author": "hey@posthog.com",
-    "license": "MIT",
+    "license": "Apache-2.0",
```
rafaeelaudibert, line 7: I've checked https://docs.npmjs.com/cli/v11/configuring-npm/package-json#license and in our case where we have a multilicense we should do this

```suggestion
    "license": "SEE LICENSE IN LICENSE",
```

react/LICENSE
rafaeelaudibert, line 1: I've taken a look here and our `react` folder does NOT include any Sentry/Mixpanel related code, so we can get away with a single license here.

I suggest we go with MIT since that's what our main posthog repo is licensed under, but you can always ask Hector if you're in doubt - I wouldn't, MIT is fine.

react/package.json
```diff
@@ -9,7 +9,8 @@
         "directory": "react"
     },
     "author": "hey@posthog.com",
-    "license": "MIT",
+    "license": "Apache-2.0",
```
rafaeelaudibert, line 12: And then make this MIT if you follow my suggestion above

rafaeelaudibert (approved):

rafaeelaudibert (changes requested):
Meant to request changes

PR #1930, opened 2025-04-30
refactor: Extract event names to constants

rafaeelaudibert:
We repeat these event names in several places. Most of them are in tests, yes, but there's a fair bit of repetition in some places, so let's extract them to a new `events.ts` file and import them from there instead

### DIFF
--- a/functional_tests/identify.test.ts
+++ b/functional_tests/identify.test.ts
@@ -7,6 +7,7 @@ import { PostHog } from '../src/posthog-core'
 import { logger } from '../src/utils/logger'
 import { uuidv7 } from '../src/uuidv7'
 import { getRequests } from './mock-server'
+import { IDENTIFY_EVENT, SET_EVENT } from '../src/events'
 
 describe('FunctionalTests / Identify', () => {
     let token: string
@@ -25,7 +26,7 @@ describe('FunctionalTests / Identify', () => {
         await waitFor(() =>
             expect(getRequests(token)['/e/']).toContainEqual(
                 expect.objectContaining({
-                    event: '$identify',
+                    event: IDENTIFY_EVENT,
                     properties: expect.objectContaining({
                         distinct_id: 'test-id',
                         $anon_distinct_id: anonymousId,
@@ -47,7 +48,7 @@ describe('FunctionalTests / Identify', () => {
         await waitFor(() =>
             expect(getRequests(token)['/e/']).toContainEqual(
                 expect.objectContaining({
-                    event: '$identify',
+                    event: IDENTIFY_EVENT,
                     $set: { email: 'first@email.com' },
                     $set_once: expect.objectContaining({
                         location: 'first',
@@ -66,7 +67,7 @@ describe('FunctionalTests / Identify', () => {
         await waitFor(() =>
             expect(getRequests(token)['/e/']).toContainEqual(
                 expect.objectContaining({
-                    event: '$set',
+                    event: SET_EVENT,
                     properties: expect.objectContaining({
                         $browser: 'Safari',
                         $browser_version: null,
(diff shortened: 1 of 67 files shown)


functional_tests/identify.test.ts
```diff
@@ -7,6 +7,7 @@ import { PostHog } from '../src/posthog-core'
 import { logger } from '../src/utils/logger'
 import { uuidv7 } from '../src/uuidv7'
 import { getRequests } from './mock-server'
+import { IDENTIFY_EVENT, SET_EVENT } from '../src/events'
```
rafaeelaudibert, line 10: I wonder if yall prefer it this way or something like `EVENTS.IDENTIFY` instead

rafaeelaudibert:
meh, didnt save as many bytes as I thought this would, thoughts?

image

src/events.ts
pauldambra, line 1: i can see these identifiers being  minified in the built output

didn't check all of them but `$web_vitals` looked to still be present as a string in the output

i think we had this same surprise with user agent string parts where 2 or 3 repetitions don't have much impact 

there we went further and did things like

```
const CLICK = 'click'
const RAGE_CLICK = '$rage' + CLICK
```

am always torn by this kind of code golf though

(NB has to be (unless rollup/terser has changed) string + string not `"${string}${string}"`

---

regardless I think this is worth it since these event names need to be correct when they're used. we do have them typed for intellisense but generally allow `and also string` so can't totally protect against typos

functional_tests/identify.test.ts
lucasheriques: the problem with TS enums are mostly when they are used as numbers. as long as we keep enums as string values only they're fine

src/events.ts
rafaeelaudibert: That's too much codegolf for me. I don't think I saw that big of a reduction in bundle size, BUT I'll merge this as is for the sake of standardizing these fields.

functional_tests/identify.test.ts
rafaeelaudibert: > repeated hardcoded strings\numbers though

In a way `const RAGECLICK_EVENT = '$rageclick'` is also repeating it, which does not spark joy

---

Given the comments and my great dislike of enums, I'll vote for keeping this as is. We can easily change this in the future, I'm actually pretty sure a future contributor will change that exact thing :)

playwright/before_send.spec.ts
```diff
@@ -38,18 +39,20 @@ test.describe('before_send', () => {
...
                             return null
                         }
                     }
-                    if (cr.event === '$autocapture') {
-                        return {
-                            ...cr,
-                            event: 'redacted',
-                        }
-                    }
+
```
pauldambra, line 55: do you remember why this change? 
this test seems to be failing now

pauldambra: which may be totally unrelated ofc

pauldambra (approved):
other than the playwright test which looks like a real fail (or at least not a flake) LGTM

playwright/before_send.spec.ts
rafaeelaudibert: Just thought it made sense to test it earlier, but... might have messed it up? I don't understand how that could change anything, but... happy to revert

rafaeelaudibert: Let's see if [`c6e4b07` (#1930)](https://github.com/PostHog/posthog-js/pull/1930/commits/c6e4b0789b01b451eefdddd8062b5035640fa466) does the trick

rafaeelaudibert: Didnt fix it, will investigate what's happening later tonight, got too many PRs open rn

rafaeelaudibert:
Ugh, will need to reopen this, too many conflicts and that weird failure.

PR #1934, opened 2025-05-02
fix: dont mangle some surveys properties

lucasheriques:
## Changes

[more context here](https://posthog.slack.com/archives/C03P7NL6RMW/p1746218789588879)

property mangling is causing some issues on surveys. for now, adding the names used on it as an exception til we can actually diagnose the issue.


dmarticus (approved):
oof

PR #1942, opened 2025-05-06
chore: include response within the survey_quesiton property

lucasheriques:
## Changes

includes the survey response within the $survey_questions property in the survey sent/dismissed events.

this will a key case for handling zapier integrations where we're not able to use the `getSurveyResponse` function to retrieve survey questions. as this allow us to get both question and response with a single property at a small cost of duplicating a small byte of info.

### DIFF
--- a/src/extensions/surveys/surveys-extension-utils.tsx
+++ b/src/extensions/surveys/surveys-extension-utils.tsx
@@ -590,6 +590,7 @@ export const sendSurveyEvent = ({
         $survey_questions: survey.questions.map((question) => ({
             id: question.id,
             question: question.question,
+            response: question.id ? responses[getSurveyResponseKey(question.id)] : null,
         })),
         $survey_submission_id: surveySubmissionId,
         $survey_completed: isSurveyCompleted,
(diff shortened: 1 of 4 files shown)


kappa90 (approved):

lricoy (approved):

src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -633,6 +634,7 @@ export const dismissedSurveyEvent = (survey: Survey, posthog?: PostHog, readOnly
         $survey_questions: survey.questions.map((question) => ({
             id: question.id,
             question: question.question,
+            response: question.id ? inProgressSurvey?.responses[getSurveyResponseKey(question.id)] : null,
```
marandaneto, line 637: do we need this in survey dismissed?

lucasheriques: since we show the same pattern for both right now, i'd like to keep it the same way. this is for the case of getting the responses in case the user dismisses the survey after answering a couple of questions

src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -590,6 +590,7 @@ export const sendSurveyEvent = ({
         $survey_questions: survey.questions.map((question) => ({
             id: question.id,
             question: question.question,
+            response: question.id ? responses[getSurveyResponseKey(question.id)] : null,
```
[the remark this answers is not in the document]
lucasheriques: I edited the issue to include the response too

marandaneto (approved):

PR #1944, opened 2025-05-07
fix: false flag should not trigger linked flags

pauldambra:
finally a false flag that isn't a conspiracy theory!

it used to be safe to check for the presence of a flag 

but we recently started returning flags that are enabled but false

so, it's not safe any more and replay linked flags was broken as a result

i had seen that we changed that API but wrote this code so long ago I totally missed this consequence

### DIFF
--- a/playwright/session-recording/session-recording-linked-flags.spec.ts
+++ b/playwright/session-recording/session-recording-linked-flags.spec.ts
@@ -51,7 +51,15 @@ test.describe('Session recording - linked flags', () => {
             },
             decideResponseOverrides: {
                 sessionRecording: { linkedFlag: 'my-linked-flag' },
-                featureFlags: { 'my-linked-flag': false },
+                flags: {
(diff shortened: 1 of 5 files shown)


pauldambra:
@marandaneto @ioannisj looks like Android and iOS SDKs don't make this mistake - nice!

Do flutter and RN simply use Android and iOS behavior or do I need to look there as well

marandaneto:
> @marandaneto @ioannisj looks like Android and iOS SDKs don't make this mistake - nice!
> 
> Do flutter and RN simply use Android and iOS behavior or do I need to look there as well

Flutter uses Android and iOS, RN has its own flag evaluation, but I think [its correct already](https://github.com/PostHog/posthog-js-lite/blob/1570b3dddadc3ec0cc2b89948353cb8475ed9dfc/posthog-react-native/src/posthog-rn.ts#L377-L407)

Issue #1940, opened 2025-05-06
Use CSS files instead of hardcoded strings for Survey styles


lucasheriques:
Right now, surveys Popover and Feeedback Widget styles are defined in a [big JS string](https://github.com/PostHog/posthog-js/blob/4c4872919b64db9124ec6e5ad806524f39169a36/src/extensions/surveys/surveys-extension-utils.tsx#L50-L345).

This is prone to errors as there's no CSS validation and syntax highlighting. Not to mention much harder to make changes and use advanced CSS features.

Instead, let's adapt our rollup builder to process and use CSS files instead.

[Started on this draft PR](https://github.com/PostHog/posthog-js/pull/1923/files), but creating this issue to slowly list out everything that is needed.

- [x] Add dependencies to rollup to build CSS files
- [ ] Pass the current styles to CSS files
- [ ] Replace the big CSS string with the actual files

PR #1923, opened 2025-04-26
feat: use CSS files for surveys instead of hardcoded strings

lucasheriques:
- **question transitions**
- **lets use css files**
- **missing styles**
- **css files are now working**
- **simplifying more styles**

## Changes

...

## Checklist
- [ ] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [ ] Accounted for the impact of any changes across different browsers
- [ ] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)
- [ ] Took care not to unnecessarily increase the bundle size

PR #1945, opened 2025-05-07
chore: make rollup able to bundle CSS files

lucasheriques:
Part 1 of https://github.com/PostHog/posthog-js/issues/1940

No changes for now. Shouldn't change bundle size as it's only adding dev dependencies.

Tagging as surveys as it's the only feature that will use it.


rollup.config.js
```diff
@@ -19,6 +23,17 @@ const plugins = (es5) => [
...
+    postcss({
+        plugins: [
+            postcssImport(),
+            postcssNesting(),
+            cssnano({
+                preset: ['default', { discardComments: { removeAll: true } }],
+            }),
+        ],
+        minimize: true,
+        inject: false,
```
pauldambra, line 35: am assuming that some code elsewhere is going to be handling loading the CSS file and we won't ever depend on this from e.g. array.js

lucasheriques: yes that's correct. I have a messy [draft PR](https://github.com/PostHog/posthog-js/pull/1923/files) that is already doing it.

it's loading the `.css` files on both `extensions/surveys.tsx` and `extensions/surveys-widget.tsx`

[here's an example](https://github.com/PostHog/posthog-js/pull/1923/files#diff-ca75b1ebdb1e4a75293f5a5f4d0bdc7b94c5ca7243b5ac9ebb49147a6dd34d83R54-R55)

pauldambra: which I _think_ would also mean that we'd have to update docs for "using posthog in an extension" to say "you can't include surveys"

since Chrome store rejects extensions that depend on loading files... 🤔 but surely there's a way to use CSS in an extension...

ok, there's stuff to consider here but this is also relatively easy to change so 👍

lucasheriques: I think, when it's bundled, the CSS is actually loaded as a string. Then I just apply this string into `<style>` tag. so it's not really loading an external file. [I know I even had to do this check when doing tests](https://github.com/PostHog/posthog-js/pull/1923/files#diff-ca75b1ebdb1e4a75293f5a5f4d0bdc7b94c5ca7243b5ac9ebb49147a6dd34d83R604-R606) as I had to mock the CSS files out.

so I don't think we'd have this restriction

pauldambra: > I think, when it's bundled, the CSS is actually loaded as a string

but (and i'm happy to learn i'm being stupid 😅 ) where is it loaded from

if I just have `array.js` and then `array.js` loads the `surveys.js` then I don't have a `surveys.css` yet, right?

so something has to load it?

pauldambra: Ah, cool beans.

So, we need to make sure we use it so that it's only in the lazy loaded bundle but otherwise all cool

thank you!

pauldambra (approved):
LGTM

PR #1952, opened 2025-05-12
fix: add new relic to payload host denylist

pauldambra:
see https://posthoghelp.zendesk.com/agent/tickets/30095 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/32280)

we're capturing (if enabled) new relic payloads in replay and that's unnecessary, let's not


robbie-c (approved):

robbie-c:
Ah looks like a trivial but legit test failure in `config › network request options › payloadHostDenyList`

rafaeelaudibert (approved):

PR #1959, opened 2025-05-15
fix: center positioning for surveys

lucasheriques:
## Changes

center positioning is slightly to the right because i'm using transform for animations now too. switch to use just left/right/margins now

CleanShot 2025-05-15 at 11 34 27@2x


pauldambra (approved):
👍

PR #1961, opened 2025-05-15
feat: chat

thmsobrmlr:
## Changes

Add chat MVP

Screenshot 2025-05-23 at 12 59 51
Screenshot 2025-05-23 at 12 59 40

...

## Checklist


src/posthog-core.ts
```diff
@@ -1385,6 +1391,11 @@ export class PostHog {
         return this.sessionManager?.onSessionId(callback) ?? (() => {})
     }
 
+    /** Get list of all chats. */
+    getChats(): void {
+        this.chat.getChat()
+    }
```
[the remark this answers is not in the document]
pauldambra: I'd not add this to the top level API of posthog-js while it's still in beta.
you can always call `posthog.chat.getChat` in the meanwhile
and keeps things tidier in case it changes before release

src/utils/chat-utils.ts
pauldambra, line 1: I think you could live without this file as well and just create the logger where you're going to use it :)

src/posthog-chat.ts
```diff
@@ -0,0 +1,133 @@
...
+        /** No calls if chat is not enabled */
+        if (!this.isEnabled) {
+            return
+        }
+
+        try {
+            this._instance._send_request({
+                url: this._instance.requestRouter.endpointFor(
+                    'api',
+                    `/api/chat/?token=${this._instance.config.token}&distinct_id=${this._instance.get_distinct_id()}`
```
pauldambra, line 105: this is fine for an alpha but...

does this make the content of all chats enumerable?

the API token used in the SDK is write-only which is what makes it safe to share

worth looking at how other tools solve this, i'd expect to exchange a short-lived token into the browser so that chat is very hard / impossible to read from arbitrary devices but easy to read from the device that started it

there needs to be some private thing here, right?

pauldambra (changes requested):
can definitely have less impact on bundle size!
that's why i'm marking request changes 😊 

some thoughts on reading chats, but that doesn't matter until we're moving past PoC

src/posthog-chat.ts
```diff
@@ -0,0 +1,133 @@
...
+                        this.conversationId = chat.id
+                    }
+                },
+            })
+        } catch (e) {
+            logger.error('PostHogChat getChat', e)
+            throw e
+        }
+    }
+}
```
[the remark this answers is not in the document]
veryayskiy: where can I see an example?

src/posthog-chat.ts
veryayskiy: That's how surveys work: https://github.com/PostHog/posthog-js/blob/main/src/posthog-surveys.ts#L195

The whole thing is copy from surveys :)

src/posthog-core.ts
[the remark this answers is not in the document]
veryayskiy: Good point

src/posthog-chat.ts
pauldambra: but surveys only loads the question right?
so it's only information you choose to expose

the answers are sent as events which the public token can't read

so i can enumerate the survey (maybe) but they're not private

whereas if we expose chats then i can enumerate what the customer wrote
and distinct ids are guessable so i can enumerate what customer might have chats

src/posthog-chat.ts
```diff
@@ -0,0 +1,133 @@
...
+                    if (chats.length > 0) {
+                        const chat = chats[0]
+                        this.messages = chat.messages || []
+                        this.conversationId = chat.id
+                    }
+                },
+            })
+        } catch (e) {
+            logger.error('PostHogChat getChat', e)
+            throw e
```
pauldambra, line 130: i wouldn't throw as well, risk that the SDK breaks the customer site or itself

we do have an onError callback i think, but i think it's also focussed on capture 🤔 
(something that generally our SDKS could be better at tbh)

src/posthog-chat.ts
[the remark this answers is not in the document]
pauldambra: surveys isn't a good example of something that does our bundling well

we always want the minimum amount of code in the main bundle - particularly for an alpha feature

`assignableWindow.__PosthogExtensions__.loadChat = loadChat` 

you want something on here
look at trace headers or deadclicks 

i can go through it on tuesday if it helps

src/posthog-chat.ts
pauldambra: i don't know tbh

i think this is new for us and we should look at what other providers do (or figure it out)

src/posthog-chat.ts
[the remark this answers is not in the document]
veryayskiy: yeah, that would indeed help!

PR #1963, opened 2025-05-15
feat: allow using person properties in survey links

lucasheriques:
## Changes

allows the usage of user properties in link-type questions so our customers can embed a person info as a URL parameter.

So `https://posthog.com/?email={{email}}` becomes `https://posthog.com/?email=test%40posthog.com`.

SDK part of https://github.com/PostHog/posthog/issues/26620

## Checklist

### DIFF
--- a/src/extensions/surveys/components/QuestionTypes.tsx
+++ b/src/extensions/surveys/components/QuestionTypes.tsx
@@ -81,35 +81,26 @@ export function OpenTextQuestion({
     )
 }
 
-// create a function that replaces any search params like
-// https://example.com/test?name={{$user_id}}
-// {{user_id}} should be replaced by posthog.get_property('$user_id')
 export function parseUserPropertiesInLink(link: string, posthog?: PostHog) {
     if (!link || !posthog || !posthog.get_property) {
         return link
     }
 
-    const regex = /\{\{(.*?)\}\}|\{(.*?)\}/g
+    const regex = /\{\{(.*?)\}\}/g
     let newLink = link
     let match
 
     // Iterate over all matches and replace them
     while (!isNull((match = regex.exec(link)))) {
-        const placeholder = match[0] // The full placeholder e.g. {{property_name}} or {property_name}
-        const propertyNameWithPotentialWhitespace = match[1] || match[2] // The actual property name
+        const placeholder = match[0] // The full placeholder e.g. {{property_name}}
+        const propertyNameWithPotentialWhitespace = match[1] // Property name is now always in match[1]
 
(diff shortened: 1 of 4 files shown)


lucasheriques:
@ioannisj you're right, it will. I think this SDK part is still safe to merge, but we have to figure out a way to handle mobile SDKs in the main posthog still. Either with warnings or coming out with a way to target a survey only in the JS sdk (web or mobile web).

I'll think about that and see if I can come up with a solution as we discussed in the daily

src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -79,6 +81,35 @@ export function OpenTextQuestion({
...
+    while (!isNull((match = regex.exec(link)))) {
+        const placeholder = match.at(0) // The full placeholder e.g. {{property_name}}
+        const propertyNameWithPotentialWhitespace = match.at(1) // Property name is now always in match[1]
+
+        if (placeholder && propertyNameWithPotentialWhitespace) {
+            const propertyName = propertyNameWithPotentialWhitespace.trim()
+            if (propertyName !== '') {
+                const propertyValue =
+                    posthog.get_property(propertyName) ||
+                    posthog.get_property(STORED_PERSON_PROPERTIES_KEY)?.[propertyName]
```
marandaneto, line 103: Is this the right thing, or is there a better way?
The API cannot return the parameter name-value since it does not know who the requester is.
We need to load this dynamically on the SDK side, but the SDK might not have all the info. E.g., user logs in device A and calls identify with all props, or user logs out and logs in again but not with all the person props again, although the person props are merged server side already.
User logs in device B and loads surveys, but all the identified props aren't cached in that device.
How would be the UX for this use case specifically?

src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -536,6 +537,7 @@ interface SurveyContextProps {
     isPopup: boolean
     onPreviewSubmit: (res: string | string[] | number | null) => void
     surveySubmissionId: string
+    posthog: PostHog | undefined
```
marandaneto, line 540: should we not allow `undefined` if we actually need it?

src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -545,8 +547,13 @@ export const SurveyContext = createContext<SurveyContextProps>({
     isPopup: true,
     onPreviewSubmit: () => {},
     surveySubmissionId: '',
+    posthog: undefined,
```
marandaneto, line 550: why do we pass `undefined` here?

PR #1964, opened 2025-05-16
fix: don't call receivedFeatureFlags if flags were disabled

jearle-dealnews:
## Changes

...

If someone uses the `advanced_disable_feature_flags_on_first_load` config option: 
- the `_callDecideEndpoint` function will be called to get the remote config. 
- However, if someone calls something like `setPersonProperties` while that is still processing, then `_additionalReloadRequested` will be set to true
- the callback for `_callDecideEndpoint` will end up calling `receivedFeatureFlags` with no flag data
- eventually onFeatureFlags is dispatched with no flag data which causes issues with our experiments.

## Checklist

### DIFF
--- a/src/posthog-featureflags.ts
+++ b/src/posthog-featureflags.ts
@@ -446,7 +446,9 @@ export class PostHogFeatureFlags {
                     return
                 }
 
-                this.receivedFeatureFlags(response.json ?? {}, errorsLoading)
+                if (!data.disable_flags) {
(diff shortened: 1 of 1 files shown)


jearle-dealnews:
If it's easier to have someone copy these code changes into a new branch, I'm fine with that. I see some of the checks seem to fail based on authorization. I just thought submitting a PR would be easier than explaining what I thought the problem was. Or if it's a problem with my code change, I can address those issues myself if it would help.

PR #1965, opened 2025-05-16
feat: add new options for survey positions

lucasheriques:
## Changes

adds new options for survey positions, and moves more inline styles to use CSS variables instead.

SDK part of https://github.com/PostHog/posthog/issues/30094.

This only does UI changes, and only for the new positionings.


ioannisj (approved):
Could not validate locally - survey example is throwing some errors for me. Since this is UI I trust that it's tested though. I do see some checks failing on Safari?

marandaneto (approved):

lucasheriques:
@ioannisj test cafe is very flaky, but all good now.

for now it won't have an effect until we also have the [main app PR](https://github.com/PostHog/posthog/pull/32292) that actually enables those options

PR #1967, opened 2025-05-19
fix: csp trailing slash

lricoy:
Defaulting to the trailing slash to avoid problems

### DIFF
--- a/playground/csp-violations/server.js
+++ b/playground/csp-violations/server.js
@@ -26,7 +26,7 @@ const POSTHOG_SCRIPT = POSTHOG_USE_SNIPPET
 </script>`
     : `<script src="/dist/posthog.js"></script>`
 
-const CSP_REPORT_URI = `${POSTHOG_API_HOST}/report?token=${POSTHOG_TOKEN}`
+const CSP_REPORT_URI = `${POSTHOG_API_HOST}/report/?token=${POSTHOG_TOKEN}`
 const USE_REPORT_TO = POSTHOG_API_HOST.startsWith('https://')
 
 const CSP_RULES = {


rafaeelaudibert (approved):

playground/csp-violations/server.js
```diff
@@ -26,7 +26,7 @@ const POSTHOG_SCRIPT = POSTHOG_USE_SNIPPET
 </script>`
     : `<script src="/dist/posthog.js"></script>`
 
-const CSP_REPORT_URI = `${POSTHOG_API_HOST}/report?token=${POSTHOG_TOKEN}`
+const CSP_REPORT_URI = `${POSTHOG_API_HOST}/report/?token=${POSTHOG_TOKEN}`
```
joshsny, line 29: Why do we pass this token in query params? Should we move it to an auth header?

lricoy: Good question!

This is going to be sent on a header to be fair: `reporting-endpoints:
posthog="https://us.i.posthog.com/report/?token=<token>&sample_rate=0.1&v=1"` and sent as a POST, but we can't control the body payload so that's why it is on the URL.

On the endpoint side: We need to have the authentication on the URL because we can't control how the browser sends the reports, just the endpoint to which it sends those reports. 

It is very industry standard to use the token like that, so there are probably more constraints: see [example1](https://blog.sentry.io/how-sentry-captures-csp-violations/), [example2](https://raygun.com/documentation/language-guides/browser-reporting/crash-reporting/csp/) but again, since it is a POST we should be fine.

joshsny: Yeah makes sense - I guess at the end of the day it's a public token anyway so should be good 👍

PR #1973, opened 2025-05-21
fix: modernise domain spelunking cookie setting code

pauldambra:
A user reports they sometimes get 431 errors as a result of the domain check cookies.

That's surprising since we set at most 8 and they are expired if accepted by the browser

But, we no longer support IE11 so we can remove the regex test here
And we can shift to the more "modern" max age which we can set a few seconds into the future giving us time to check without leaving the cookie around if something goes wrong

see internal links: https://posthoghelp.zendesk.com/agent/tickets/30280 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/32784) and https://posthog.slack.com/archives/C03P7NL6RMW/p1747837404395389

### DIFF
--- a/src/__tests__/storage.test.ts
+++ b/src/__tests__/storage.test.ts
@@ -21,8 +21,8 @@ describe('sessionStore', () => {
             set cookie(value) {
                 //needs to refuse known public suffixes, like a browser would
                 // value arrives like dmn_chk_1699961248575=1;domain=.uk
-                const domain = value.split('domain=')
-                if (['.uk', '.com', '.au', '.com.au', '.co.uk'].includes(domain[1])) return
+                const domain = value.split('domain=')[1].split(';')[0]
+                if (['.uk', '.com', '.au', '.com.au', '.co.uk'].includes(domain)) return
                 this.value_ += value + ';'
             },
         }
--- a/src/storage.ts
+++ b/src/storage.ts
@@ -13,7 +13,6 @@ import { logger } from './utils/logger'
 import { window, document } from './utils/globals'
 import { uuidv7 } from './uuidv7'
 
-const Y1970 = 'Thu, 01 Jan 1970 00:00:00 GMT'
 // we store the discovered subdomain in memory because it might be read multiple times
 let firstNonPublicSubDomain = ''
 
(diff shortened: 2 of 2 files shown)


src/storage.ts
```diff
@@ -48,18 +47,17 @@ export function seekFirstNonPublicSubDomain(hostname: string, cookieJar = docume
...
-        const candidateCookieValue = key + '=1;domain=.' + candidate
+        const candidateCookieValue = key + '=1;domain=.' + candidate + ';path=/'
 
-        // try to set cookie
-        cookieJar.cookie = candidateCookieValue
+        // try to set cookie, include a short expiry in seconds since we'll check immediately
+        cookieJar.cookie = candidateCookieValue + ';max-age=3'
 
-        if (R.test(cookieJar.cookie)) {
+        if (cookieJar.cookie.includes(key)) {
```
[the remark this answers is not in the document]
pauldambra: the key is very unique... that's overkill

src/__tests__/storage.test.ts
```diff
@@ -21,8 +21,8 @@ describe('sessionStore', () => {
             set cookie(value) {
                 //needs to refuse known public suffixes, like a browser would
                 // value arrives like dmn_chk_1699961248575=1;domain=.uk
-                const domain = value.split('domain=')
-                if (['.uk', '.com', '.au', '.com.au', '.co.uk'].includes(domain[1])) return
+                const domain = value.split('domain=')[1].split(';')[0]
```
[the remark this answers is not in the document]
pauldambra: it's a test mock, robot dude

src/__tests__/storage.test.ts
rafaeelaudibert, line 1: If we dont support IE11 anymore, why is it that we still have some checks around it everywhere? Should we more explicitly remove support to it? No legacy build for example?

src/__tests__/storage.test.ts
```diff
@@ -21,8 +21,8 @@ describe('sessionStore', () => {
...
+                if (['.uk', '.com', '.au', '.com.au', '.co.uk'].includes(domain)) return
```
rafaeelaudibert, line 25: It's so crazy this is a thing, like wtf AU/UK, what are you doing

rafaeelaudibert (approved):
If we're ok with dropping IE11.... yeah, go for it, much cleaner

src/__tests__/storage.test.ts
pauldambra: > No legacy build for example?

oh, maybe i explained things you now... 
we don't support ie11 in the main bundle anymore but we do have a legacy build

i don't think i have a way of checking if the legacy build is being used 🤔

rafaeelaudibert: Ok, gotcha, but will it be smart enough to polyfill something like this cookie setting? Or do we assume that it's ok to have a legacy build that doesn't work 100%?

pauldambra: i _think_ it will polyfill includes

PR #1974, opened 2025-05-21
fix: Block newer google bots

rafaeelaudibert:
Extracted from https://developers.google.com/search/docs/crawling-indexing/google-common-crawlers


lricoy (approved):

rafaeelaudibert:
The changelog is supposed to update automatically, we just need to check whether it makes sense in case automation goes 🚬

joshsny (approved):

marandaneto (approved):

PR #1978, opened 2025-05-24
feat: Add "defaults" to config

robbie-c:
## Changes

See https://posthog.slack.com/archives/C03P7NL6RMW/p1747948930053279 for the discussion

We want to make `capture_pageview: 'history_change'` the default, but don't want to make a breaking change.

This lets us update the default config without breaking anyone's code.

(We will need to update our code examples to add a pinned value)

## Checklist

### DIFF
--- a/src/posthog-core.ts
+++ b/src/posthog-core.ts
@@ -56,6 +56,7 @@ import {
     SessionIdChangedCallback,
     SnippetArrayItem,
     ToolbarParams,
+    ConfigDefaults,
 } from './types'
 import {
     _copyAndTruncateStrings,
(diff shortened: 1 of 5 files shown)


src/posthog-core.ts
```diff
@@ -996,6 +998,7 @@ export class PostHog {
         const startTimestamp = readOnly ? undefined : this.persistence.remove_event_timer(eventName)
         let properties = { ...eventProperties }
         properties['token'] = this.config.token
+        properties['$config_defaults'] = this.config.defaults
```
pauldambra, line 1001: nice

pauldambra (approved):
this is a great idea!

src/posthog-core.ts
robbie-c: nice

(yeah I figure we can use this to warn snippet users about deprecation)

PR #1991, opened 2025-06-02
chore: workspace migration

hpouillot:
## Context

The goal here is to merge `posthog-js-lite` inside `posthog-js` as discussed in the RFC [here](https://github.com/PostHog/product-internal/pull/758) and the first step to do so is to convert `posthog-js` repo to a pnpm workspace to add multiple packages.

## Changes

- [x] create pnpm workspace
- [x] move posthog-js to packages/browser
	- [x] migrate eslint configuration. https://github.com/PostHog/posthog-js/pull/1995
	- [x] migrate github workflows. https://github.com/PostHog/posthog-js/pull/1993
	- [x] update vercel deployment. https://github.com/PostHog/posthog-js/pull/1996
	- [x] update release workflow. https://github.com/PostHog/posthog-js/pull/1998

## Notes

- This PR involves mostly file moves and acts as the base branch for other sub-PRs to make reviewing easier. 
- sub-PRs depends on each other, we need to review them in the list order.
- CI fails because because workflows are updated in sub-PRs.
- I tried to keep file changes as minimal as possible to avoid merge conflicts even if sometimes it's sub-optimal.
- I will rebase those branches regularly so you can keep shipping 🚢


.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 . "$(dirname -- "$0")/_/husky.sh"
 
-npx lint-staged
+# npx lint-staged
```
hpouillot: I disabled pre-commit hook for now to avoid file changes (to have only move commit and make rebase easier)

CODEOWNERS
```diff
@@ -1,11 +1,11 @@
 # the recorder file can have unexpected impact on users' sites
-src/entrypoints/recorder.ts @team-replay
+packages/browser/src/entrypoints/recorder.ts @team-replay
```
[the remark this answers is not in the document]
pauldambra: ah, interesting...

package.json
pauldambra, line 1: is there too little in this package.json and that's why the github actions are failing?

hpouillot: yes, it looks like github actions need package manager infos at the workspace level

package.json
```diff
@@ -1,160 +1,5 @@
...
-            "eslint src --fix",
-            "eslint playwright --fix",
-            "prettier --write"
-        ]
-    },
-    "browserslist": [
-        "> 0.5%, last 2 versions, Firefox ESR, not dead, IE 11"
-    ],
     "pnpm": {
         "patchedDependencies": {
```
joshsny, line 8: should these patches be specified at the workspace root or in the browser package?

joshsny (commented):
Looking great - good first step in this direction 🙌

packages/browser/package.json
```diff
@@ -0,0 +1,158 @@
+{
+    "name": "posthog-js",
```
joshsny, line 2: will we rename this internally in the workspace - or does it stay as `posthog-js`?

hpouillot: Yes, once everything is moved, I think we should discuss package names and how code is shared in the workspace.

rafaeelaudibert:
@hpouillot This PR will be extremely tricky to merge unless we freeze development/merge on other PRs, do you think we need this by any chance?

Also, is there an RFC with what you're proposing here? This is a big change, and I'm not familiar with what the end goal is supposed to look like

hpouillot:
@rafaeelaudibert IMO, I don't think we need to completely freeze development. This branch is almost only files moves so rebase should be straightforward. The only thing we need to freeze is the CI while I update the workflows.

package.json
hpouillot: It's not ideal but pnpm expects them to be at the root level

packages/browser/package.json
```diff
@@ -0,0 +1,158 @@
...
+        "jsdom": "16.5.0",
+        "jsdom-global": "3.0.2",
+        "lint-staged": "^10.2.11",
+        "localStorage": "1.0.4",
+        "msw": "^1.3.3",
+        "node-fetch": "^2.6.11",
+        "postcss": "^8.5.3",
+        "postcss-import": "^16.1.0",
+        "postcss-nesting": "^13.0.1",
+        "posthog-js": "link:",
```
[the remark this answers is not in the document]
hpouillot: Let's keep it as it's already on main branch but we should probably remove it

packages/browser/package.json
```diff
@@ -0,0 +1,158 @@
+{
+    "name": "posthog-js",
+    "version": "1.249.3",
```
daibhin: We'll need to make sure this (and any other changing values) are updated before merging

packages/browser/package.json
```diff
@@ -0,0 +1,158 @@
...
+        "*.{ts,tsx}": [
+            "eslint src --fix",
+            "eslint playwright --fix",
+            "prettier --write"
+        ]
+    },
+    "browserslist": [
+        "> 0.5%, last 2 versions, Firefox ESR, not dead, IE 11"
+    ]
+}
```
daibhin, line 137: I'm assuming everything here is largely copy pasted

packages/browser/package.json
pauldambra: makes me think of an old colleague of mine spending 3 months trying to change how a single field dealt with trimming whitespace when multiple teams depended on it differently

I _think_ we'd have to dual publish `posthog-js` and `@posthog/web` for a long time before we could remove `posthog-js` 

but, literally spent 30 seconds thinking about it just now so 🤷

hpouillot: I agree I think we will need to keep publishing `posthog-js` as a wrapper of `@posthog/web` for sometime and add a lot of warning banners before we can remove it

daibhin: I think we already track the browser package in `posthog-js-lite` as `$lib: web` but it might be worth having a different $lib key for `posthog-js` (presumably keep this as it is) and change it in `@posthog/web` to be something like `$lib: browser`. That would allow us to track usage over time and make an informed call as to when we would even consider sunsetting the old package name

daibhin (approved):
Epic stuff! Diff looks reasonably but there's every chance I missed something in some of the config changes. Is there anything you'd like tested before shipping? Will have a play around with the branch to make sure I can get things to work locally but curious if there's any specific flows you're worried about

packages/browser/package.json
hpouillot: Yes, it should be the exact same package.json except for devDependencies

hpouillot:
Ok, I'll keep the same lib key for now. @daibhin, I think we can focus on test workflows (just to be extra sure nothing broke) and also CD workflows, as I couldn't test them.

PR #1994, opened 2025-06-03
feat: exception rate limiter by type

daibhin:
## Changes

Error tracking is susceptible to unintentionally capturing a large number of exceptions via autocapture, causing customer to rack up very large bills.

The usage pattern here is very different to the default client-side rate limits on capture and hence is implemented as a separate rate limiter:
- We want to bucket by exception type. We should continue to capture exceptions but rate limit noisy ones e.g. ResizeObserver which can often fire in a loop
- The rates will be much lower than the equivalent limiter for events. We let you capture up to 100 events / second by default but you're unlikely to want to capture more than 1 exception / second (although we let you capture up to 10)


src/extensions/exception-autocapture/exception-rate-limiter.ts
```diff
@@ -0,0 +1,61 @@
...
+
+export class ExceptionRateLimiter {
+    // We refill the buckets every 10 seconds.
+    // The default bucket size of 10 and refill rate of 1 means we
+    // capture ten exceptions of a type before burst protection kicks in,
+    // after which we capture one exception of each type every 10 seconds
+
+    private _bucketSize = 10
+    private _refillRate = 1
+    private _exceptionBuckets: Record<string, number> = {}
```
pauldambra: oh... interesting... I was wondering about adding a "throttle key" to capture
because i frequently want to capture a single instance of something per session id
so i can get the signal knowing that i don't have to worry whether every mutation in a 3 hour recording will send an event

this is very close to that.
if there was an (x: unknown) => string so i can init a rate limit and throttle by x.properties.session_id

then i'm 99% of the way there with this
(would just need to figure out what size and refill rate to set so I get close to 1 per session)

so I'd be tempted to make this less exception focussed and just be a rate limiter
and then we can maybe even get one ratelimiter instead of 2

since this is "just" a rate limiter that knows how to read exception properties, the one I want is "just" a rate limiter that knows how to get session id from an event, and the one we already have is "just" a rate limiter than knows how to loop over SVG elements in a snapshot, right?

pauldambra: that also means I the client of posthog-js need to be able to push in rate limiters from the outside
as well as we have our own built in

🤔 

but then I can do that with before_send already 🤔 

i may be just being noise

daibhin: This actually brings us from two to three rate limiters given we have the default rate limiter and the mutation rate limiter already. I know the mutation rate limiter works on rrweb events rather than PostHog events but really they're all just counters behind the scenes. I'm not against aligning them a little more just don't know if I have context on each of their uses e.g. the default rate limiter has average limits & burst limits whereas the mutation rate limiter has refill rate and bucket size. None of the rate limiters let you customize the interval 😂 

The problem with the before_send hook is that it runs _after_ the default rate limiter which is fine if your custom rate limiter is more limited but will filter out items if more permissive.

Would you like me to attempt making this more generic and consolidating it with the mutation rate limiter? I think the default rate limiter feels scarier to change 🙈

pauldambra: > Would you like me to attempt making this more generic and consolidating it with the mutation rate limiter

yes... but also i have no visibility of what all else you have going on 😅 

i bet the default rate limiter was a quick job but... we staying at two rather than climbing to three would be good

also ofc.. ship this and we all promise to be good citizens and come back and tidy up later 🙈

src/__tests__/posthog-core.loaded.test.ts
```diff
@@ -167,7 +167,7 @@ describe('loaded() with flags', () => {
             const receivedFeatureFlagsSpy = jest.spyOn(instance.featureFlags, 'receivedFeatureFlags')
 
             instance.featureFlags._callDecideEndpoint()
-            jest.runAllTimers()
+            jest.runOnlyPendingTimers()
```
daibhin, line 170: Had to change this for the exact reason explained in https://jestjs.io/docs/next/timer-mocks#run-pending-timers

src/utils/bucketed-rate-limiter.ts
```diff
@@ -0,0 +1,74 @@
...
+            refillInterval: number
+            _onBucketRateLimited?: (key: T) => void
+        }
+    ) {
+        this._onBucketRateLimited = this._options._onBucketRateLimited
+        this._bucketSize = clampToRange(this._options.bucketSize, 0, 100, 'rate limiter bucket size')
+        this._refillRate = clampToRange(
+            this._options.refillRate,
+            0,
+            Math.min(this._bucketSize, 100), // never refill more than bucket size
```
hpouillot: is this._bucketSize not always inferior to 100 as it's clamped ?

src/utils/bucketed-rate-limiter.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private _refillBuckets = () => {
+        Object.keys(this._buckets).forEach((key) => {
+            const newTokens = this._getBucket(key) + this._refillRate
+            this._setBucket(key, newTokens)
+
+            if (newTokens >= this._bucketSize) {
+                delete this._buckets[key]
+            }
```
hpouillot:
```suggestion

            if (newTokens >= this._bucketSize) {
                delete this._buckets[key]
            } else {
            	this._setBucket(key, newTokens)
            }
```

src/utils/bucketed-rate-limiter.ts
```diff
@@ -0,0 +1,74 @@
...
+    }
+
+    private _getBucket = (key: T | string) => {
+        return this._buckets[String(key)]
+    }
+    private _setBucket = (key: T | string, value: number) => {
+        this._buckets[String(key)] = value
+    }
+
+    public isRateLimited = (key: T) => {
```
hpouillot: Hum, I didn't expect this method to reduce the number of tokens. I would either split the method in two (update and check) or rename it to make sure there is no misusage.

daibhin:
Confirmed this worked. Please ignore the "Uncaught TypeError" that's actually a different bug related to suppression rules. You can assume that the TypeError is actually the exception being captured. You see it is captured 9 times before it runs out of tokens and starts logging the info message. Once 10 seconds passes it tries to capture another instance (because 1 token was refilled) but will be limited again after that

Screenshot 2025-06-05 at 17 07 18

src/utils/bucketed-rate-limiter.ts
daibhin: Good point. Simplified

src/utils/bucketed-rate-limiter.ts
daibhin: Fixed in next commit

src/utils/bucketed-rate-limiter.ts
daibhin: Good point. Updated to `consumeRateLimit`

hpouillot (approved):

Issue #1667, opened 2025-01-22
[Question] ERR_MODULE_NOT_FOUND error when using ‘posthog-js/lib/src/customisations’


lucashan13:
Hello, Posthog team!

I'm trying to use your library and I have a question.

I'm using [this documentation](https://posthog.com/docs/libraries/js#sampling-to-receive-only-40-of-events-by-person-distinct-id) to set up sampling, but I ran into a problem.

I'm getting an `ERR_MODULE_NOT_FOUND` error on `import { sampleByDistinctId } from posthog-js/lib/src/customisations`, and when I looked into it, the files built in that path are being served like ESM (using import, export), and I'm guessing that the module isn't being found because it's missing the ‘.js’ at the end of the file I'm importing.

Can you tell me why the module is being served like this? 

e.g.

```js
import { CAMPAIGN_PARAMS, EVENT_TO_PERSON_PROPERTIES, Info } from '../utils/event-utils';
import { each, extend } from '../utils';
import { includes } from '../utils/string-utils';
export var setAllPersonProfilePropertiesAsPersonPropertiesForFlags = function (posthog) {
    var allProperties = extend({}, Info.properties({
        maskPersonalDataProperties: posthog.config.mask_personal_data_properties,
        customPersonalDataProperties: posthog.config.custom_personal_data_properties,
    }), Info.campaignParams({
        customTrackedParams: posthog.config.custom_campaign_params,
        maskPersonalDataProperties: posthog.config.mask_personal_data_properties,
        customPersonalDataProperties: posthog.config.custom_personal_data_properties,
    }), Info.referrerInfo());
    var personProperties = {};
    each(allProperties, function (v, k) {
        if (includes(CAMPAIGN_PARAMS, k) || includes(EVENT_TO_PERSON_PROPERTIES, k)) {
            personProperties[k] = v;
        }
    });
    posthog.setPersonPropertiesForFlags(personProperties);
};
```

turnipdabeets:
Closing as was fixed in #4089

PR #1996, opened 2025-06-04
chore: worspace migration (vercel)

hpouillot:
Migration PR: https://github.com/PostHog/posthog-js/pull/1991

## Changes

- Remove lib folder from posthog-js
- Add vercel.json to specify install and build commands
- Update dependency path

## Notes

- We need to update the root directory on vercel to fix the deployment issue
- We can do that once the migration is done to avoid breaking other PR checks

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -41,7 +41,6 @@
     "module": "dist/module.js",
     "types": "dist/module.d.ts",
     "files": [
-        "lib/*",
         "dist/*",
         "react/dist/*",
         "react/package.json"
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -9,8 +9,9 @@
     "packageManager": "pnpm@9.15.4",
     "scripts": {
         "start": "pnpm build-react && NODE_OPTIONS=\"--max-old-space-size=8192\" pnpm build-rollup -w",
-        "build": "pnpm build-rollup && pnpm build-react",
+        "build": "pnpm build-rollup && pnpm build-lib && pnpm build-react",
         "build-rollup": "rm -rf lib && tsc -b && rollup -c --bundleConfigAsCjs",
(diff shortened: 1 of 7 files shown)


packages/browser/package.json
```diff
@@ -41,7 +41,6 @@
     "module": "dist/module.js",
     "types": "dist/module.d.ts",
     "files": [
-        "lib/*",
```
hpouillot: I removed the lib folder because it made the next build process crash. It looks like we only use it as an intermediate step to generate types. Is there any reason to keep it ?

pauldambra: > Is there any reason to keep it ?

the answer to anything "public" with posthog-js will always be: because someone might depend on it

the really important thing is the bundled files `array.js` etc, i _think_ they wouldn't be affected by this and that's the area where you can silently break other people's sites 🙈 😬 

i guess removing this here would break people's builds, and arguably that's ok if we do it rarely

still technically a breaking change - but not a breaking change in deployed production code

---

then we should at least document an issue (even if resolved) with why we did it, so that if someone asks us in M weeks time we don't have to relearn everything

pauldambra (approved):
qualified approval...

i'm sure we won't manage to make this omlette without breaking some eggs

if this is the only egg then i _think_ it would be ok... 

do document the lib problem somewhere though would be good to avoid it, and good to at least not lose the reason

packages/browser/package.json
hpouillot: Looks like there is an open issue related to this : https://github.com/PostHog/posthog-js/issues/1667

The lib folder is distributed as an ESM package but the package.json don't mention this, so it's treated as CJS.

hpouillot: It's fixed with the latest [commit](https://github.com/PostHog/posthog-js/pull/1996/commits/9fb8428d1f1d16a26f230e43b3073ec5e3439776) but it's a bit hacky

PR #1999, opened 2025-06-04
feat: improve surveys accessibility

lucasheriques:
Closes https://github.com/PostHog/posthog/issues/32213

Significant number of line changes, however, biggest ones is CSS which hopefully won't need too much attention as there's no logic. It's mainly adding new styles and focus states.

## Changes

Currently, Surveys form suffer from a couple of accessibility issues:

- There's no visible outline/focus indicators for each fields
- Keyboard navigation doesn't fully work because of the way the inputs are set up
- Incorrect use of semantic tags. We were using mostly `<div>`s, which are now replaced with proper elements like `label`, `fieldset`, `h3`, `p`, among others.
- Among other problems mentioned in the issue

For the multiple/single choice questions, we were also using an unorthodox way of showing them. On the code, they were always defined as the usual radio/checkbox input. However, in the actual visualization, they appear to be sort of a button.

I decided to change them to look like default radio/checkbox inputs. I don't think we should be innovating here, and instead we should rely on the web standards for them. I did do a bunch of micro-interactions and animations to make sure the survey feels delightful to use. This now also includes automatically closing the survey in the confirmation message by pressing enter or escape.

Here are some detailed videos on the different of each:

| Before | After |
|--------|--------|
| [Screen recording](https://share.cleanshot.com/3BpJ3TGG) | [Screen recording](https://share.cleanshot.com/FgrHhz4n)  EDIT: [Newest recording](https://share.cleanshot.com/YfPbwJF1) that simplifies animations | 

## Checklist
- [x] Took care not to unnecessarily increase the bundle size -> it's increased, but only on surveys. Unfortunately it's necessary to make all the required adjustments and new CSS rules.


rollup.config.js
```diff
@@ -28,7 +28,37 @@ const plugins = (es5) => [
             postcssImport(),
             postcssNesting(),
             cssnano({
-                preset: ['default', { discardComments: { removeAll: true } }],
+                preset: [
```
lucasheriques, line 31: feel free to look past this, main goal of this is to reduce our CSS files as much as possible

src/extensions/surveys.tsx
```diff
@@ -1208,18 +1199,21 @@ const getQuestionComponent = ({
...
+            return <OpenTextQuestion {...baseProps} question={question} key={question.id} />
+        case SurveyQuestionType.Link:
+            return <LinkQuestion {...baseProps} question={question} key={question.id} />
+        case SurveyQuestionType.Rating:
+            return <RatingQuestion {...baseProps} question={question} key={question.id} />
+        case SurveyQuestionType.SingleChoice:
+        case SurveyQuestionType.MultipleChoice:
+            return <MultipleChoiceQuestion {...baseProps} question={question} key={question.id} />
+        default:
+            logger.error(`Unsupported question type: ${(question as any).type}`)
```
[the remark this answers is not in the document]
lucasheriques: no need just for the logs, as this is not a case we expect to happen as long as TS is working correctly

src/extensions/surveys/components/ConfirmationMessage.tsx
```diff
@@ -24,26 +26,37 @@ export function ConfirmationMessage({
...
+    useEffect(() => {
+        const handleKeyDown = (event: KeyboardEvent) => {
+            if (event.key === 'Enter' || event.key === 'Escape') {
+                event.preventDefault()
+                onClose()
+            }
+        }
+        addEventListener(document, 'keydown', handleKeyDown as EventListener)
+        return () => {
+            document?.removeEventListener('keydown', handleKeyDown)
```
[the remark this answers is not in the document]
lucasheriques: it's not needed to check in the event listener

src/extensions/surveys/survey.css
```diff
@@ -20,16 +20,135 @@
...
+        color: var(--ph-survey-input-text-color);
+        background: var(--ph-survey-input-background);
+        padding: 10px;
+        border: 1.5px solid var(--ph-survey-border-color);
+        border-radius: 4px;
+        /* Slightly slower for form inputs to feel more intentional */
+        transition:
+            border-color 0.2s ease-out,
+            box-shadow 0.2s ease-out,
+            transform 0.15s ease-out;
```
adamleithp, line 66: Nits: the animations particularly the slide effect for the next screen and the expanding text boxes, they seem a bit gimmicky and unprofessional. I'd drop them entirely.

adamleithp (dismissed):
love it great job

src/extensions/surveys/survey.css
lucasheriques: do you mean dropping all animations then, or just those two?

ioannisj: I'd agree in the sense that these feel a bit opinionated and may clash with a host app's design language? Since the goal is to integrate with a wide range of host apps, I think less is more in this case? Unless people can easily remove  or replace them I'd say. I don't feel strongly about this, so up to you

lucasheriques: I'll remove all but the fade on open/close then. Agree makes sense so it works for all UIs and is more generic

lucasheriques: [here's the new screen recording](https://share.cleanshot.com/YfPbwJF1). most animations gone besides the fade on opening/closing the survey.

fivestarspicy:
@lucasheriques I think this is a big upgrade on accessibility and following standards for this. I like it!

src/extensions/surveys/survey.css
lucasheriques: @marandaneto @ioannisj I talked to Adam and all good on the new changes here. can either of you approve so we can get this in? he's swamped atm

ioannisj (approved):
https://github.com/PostHog/posthog-js/pull/1999#discussion_r2140657123

marandaneto (approved):
noice!

PR #2000, opened 2025-06-05
feat(flags): replace `/decide` with `/flags&config=true`

dmarticus:
## Changes

`/flags` now supports all the fields that `/decide` did when called with the `&config=true` param.  This means there are no more functional blockers to rolling out all latest PostHog SDKs to use `/flags` instead of `/decide` 🎉 

I'm not gonna merge this until I've added some rollout condition stuff based on hashed API tokens (similar to what I did here: https://github.com/PostHog/posthog-js-lite/pull/458)

## Checklist


src/posthog-core.ts
```diff
@@ -313,7 +316,8 @@ export class PostHog {
     private _internalEventEmitter = new SimpleEventEmitter()
 
     // Legacy property to support existing usage - this isn't technically correct but it's what it has always been - a proxy for flags being loaded
-    public get decideEndpointWasHit(): boolean {
+    // TODO DYLAN: do I need to keep this, or can I rename it?
+    public get flagsEndpointWasHit(): boolean {
```
marandaneto: its a public API so I'd say no, what you can do is to add a @deprecated annotation, add a new method, and this legacy method just forwards the call to the new method

src/posthog-featureflags.ts
```diff
@@ -391,20 +391,18 @@ export class PostHogFeatureFlags {
...
-        // (e.g. we could call `/decide` with flags disabled for the data otherwise returned by remote config, and then still call
-        // `/flags/` to get the flag evaluation data).
-        const eligibleForFlagsV2 =
+        // Flags v2 requires remote config to be enabled to skip /flags&config=true and use /flags directly
+        const useRemoteConfigWithFlags =
             this._instance.config.__preview_flags_v2 && this._instance.config.__preview_remote_config
 
-        const baseUrl = eligibleForFlagsV2 ? '/flags/?v=2' : '/decide/?v=4'
+        // if we're not using RemoteConfig (which loads the config info from the JS snippet itself), then get all the data from remote config on the server side
+        const baseUrl = useRemoteConfigWithFlags ? '/flags/?v=2' : '/flags/?v=2&config=true'
```
marandaneto: @dmarticus @lucasheriques we need to make sure that surveys work in both cases, with remote config enabled or not

playwright/error-tracking.spec.ts
```diff
@@ -6,7 +6,7 @@ test.describe('Exception capture', () => {
     test.describe('Exception autocapture enabled', () => {
         test.beforeEach(async ({ page, context }) => {
             await start(
-                { decideResponseOverrides: { autocaptureExceptions: false }, url: './playground/cypress/index.html' },
+                { flagsResponseOverrides: { autocaptureExceptions: false }, url: './playground/cypress/index.html' },
```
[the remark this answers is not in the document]
dmarticus: look man I just ported a test take it up with paul

playwright/utils/setup.ts
```diff
@@ -56,9 +56,27 @@ export async function start(
...
+                    description: 'Matched condition set 1',
+                },
+                metadata: {
+                    id: 1421,
+                    version: 1,
+                    description: undefined,
+                    payload: undefined,
+                },
+            },
+        },
```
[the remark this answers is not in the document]
dmarticus: no ❤️

src/posthog-core.ts
dmarticus: cool, thanks for looking!

src/posthog-featureflags.ts
dmarticus: I'll double-check that in my testing!

dmarticus: seems to work on my testing, I did the following scenarios:

- turning on surveys yields "surveys": true
- survey flag targeting works

What was the situation that wasn't working in the past?

marandaneto: There wasn't a problem before.
I just wanted to ensure you test both scenarios, as flipping the remote config flag would use a different endpoint (adding a new query param, actually) that responds with a more complete configuration. If surveys still work in both scenarios, that's all good.

dmarticus: Yeah, surveys still work in both cases :) `/flags&config=true` is basically `/decide` now.

src/utils/index.ts
```diff
@@ -248,3 +248,39 @@ export function addEventListener(
...
+    loggerInstance?: { warn: (message: string) => void }
+): T {
+    const hasNewField = newField in config && !isUndefined(config[newField])
+    const hasOldField = oldField in config && !isUndefined(config[oldField])
+
+    if (hasNewField) {
+        return config[newField]
+    }
+
+    if (hasOldField) {
```
haacked, line 275: Minor perf nit: No need to evaluate `hasOldField` if `hasNewField` is true so might as well move it to after the return.

```suggestion
    const hasNewField = newField in config && !isUndefined(config[newField])

    if (hasNewField) {
        return config[newField]
    }

    const hasOldField = oldField in config && !isUndefined(config[oldField])

    if (hasOldField) {
```

haacked (approved):
Whew! Lots of nitty gritty name changes. But code looks solid. One minor minor suggestion.

lucasheriques (approved):
I tested this and it looks like all is working on the surveys side! 🙏

PR #2003, opened 2025-06-05
fix: Stop autocapture turned on if eventually ad-blocked

rafaeelaudibert:
## Changes

The existing code works great if we aren't adblocked, but it has a major flaw if /decide calls are **eventually** blocked after being previously going through.

Think of the following scenario:
- Person loads PostHog and we get from the backend that we should autocapture
- Store `true` in the persistent store
- Person's adblocker is turned on and starts blocking requests to `/decide`
- Company turns autocapture off in their settings
- When person tries hitting `/decide` it fails to get the new `false` state and reuses the `true` value

Note that this is only a problem when adblocker is blocking `/decide` but not blocking `/e`. I've found at least one case where this is happening in production, so let's protect against that.

### DIFF
--- a/src/autocapture.ts
+++ b/src/autocapture.ts
@@ -19,7 +19,7 @@ import { AutocaptureConfig, COPY_AUTOCAPTURE_EVENT, EventName, Properties, Remot
 import { PostHog } from './posthog-core'
 import { AUTOCAPTURE_DISABLED_SERVER_SIDE } from './constants'
 
-import { isFunction, isNull, isObject } from './utils/type-utils'
+import { isBoolean, isFunction, isNull, isObject } from './utils/type-utils'
 import { createLogger } from './utils/logger'
 import { document, window } from './utils/globals'
 import { convertToURL } from './utils/request-utils'
@@ -325,16 +325,23 @@ export class Autocapture {
     }
 
     public get isEnabled(): boolean {
-        // If we're not positive whether the server enabled/disabled it, we don't want to enable it yet
-        // This implies that if the /decide request is not made, we will not enable autocapture
-        // but that's not a problem because that's likely an adblocker acting and it'd block our autocapture event anyway
-        //
-        // In the rare case where we've explicitly disabled calling decide, then we'll use the persisted/client value
-        if (isNull(this._isDisabledServerSide) && !this.instance.config.advanced_disable_decide) {
-            return false
+        // Special case where we dont care about the client side config,
+        // or any cached values, we only care about the server side config
+        if (this.instance.config.autocapture === 'wait_for_server') {
+            return !this._isDisabledServerSide
         }
 
+        // Else we either use, in order: server side config, server side cached value, client side config
         const persistedServerDisabled = this.instance.persistence?.props[AUTOCAPTURE_DISABLED_SERVER_SIDE]
+        const memoryDisabled = this._isDisabledServerSide
+
+        if (
+            isNull(memoryDisabled) &&
+            !isBoolean(persistedServerDisabled) &&
+            !this.instance.config.advanced_disable_decide
+        ) {
+            return false
+        }
 
         const disabledServer = this._isDisabledServerSide ?? !!persistedServerDisabled
         const disabledClient = !this.instance.config.autocapture
--- a/src/autocapture.ts
+++ b/src/autocapture.ts
@@ -19,7 +19,7 @@ import { AutocaptureConfig, COPY_AUTOCAPTURE_EVENT, EventName, Properties, Remot
 import { PostHog } from './posthog-core'
 import { AUTOCAPTURE_DISABLED_SERVER_SIDE } from './constants'
 
-import { isBoolean, isFunction, isNull, isObject } from './utils/type-utils'
+import { isFunction, isNull, isObject } from './utils/type-utils'
 import { createLogger } from './utils/logger'
 import { document, window } from './utils/globals'
 import { convertToURL } from './utils/request-utils'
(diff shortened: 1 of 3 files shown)


robbie-c:
I don't like the trade-off here

If `decide` takes a long time (which is not uncommon on poor 3g connections),  I'd rather we use `AUTOCAPTURE_DISABLED_SERVER_SIDE` if we have it, otherwise we might drop events that should have been captured.

IMO, there are a few ways we could deal with this that would be preferable:
* If users are running into problems with ad blockers, we make it clearer how they could avoid them (e.g. managed reverse proxy https://posthog.com/docs/advanced/proxy/managed-reverse-proxy)
* If the user has client-side config, use that until `decide` returns (right now this is hard because we don't distinguish between options that are set vs defaults)

rafaeelaudibert:
@robbie-c 

> If users are running into problems with ad blockers, we make it clearer how they could avoid them (e.g. managed reverse proxy [posthog.com/docs/advanced/proxy/managed-reverse-proxy](https://posthog.com/docs/advanced/proxy/managed-reverse-proxy)). There are probably other things that will break (e.g. feature flags) if decide doesn't work at all.

Funnily enough, this user in specific is using a proxy already. This PR might be a red herring, but the only way I could reproduce that behavior is by having calls go through when hitting the `/e` endpoint but fail when hitting `/decide`. This implies there is some kind of adblocking happening, but it's not very smart - it should do the opposite.

> If the user has client-side config, use that until decide returns (right now this is hard because we don't distinguish between options that are set vs defaults)

The idea here is that this user, in specific, wants to use our server-side config to programatically turn it on/off at their own demand; this way, they can autocapture while they care about it, and easily turn it off once they've gotten the information they need. That means they need to keep it enabled client-side, but they don't want the steady influx of events they're getting because the cached value is incorrect.

---

You might wanna check their support ticket and see if you can come up with a more reasonable explanation to the problem than the one I came up with: https://posthoghelp.zendesk.com/agent/tickets/31417 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/36653)

src/types.ts
```diff
@@ -320,11 +320,20 @@ export interface PostHogConfig {
...
+     *
+     * You can also pass in an object to configure autocapture.
+     * @see {AutocaptureConfig} for more details.
+     *
      * This setting does not affect capturing pageview events (see `capture_pageview`).
      *
      * @default true
      */
-    autocapture: boolean | AutocaptureConfig
+    autocapture: boolean | 'wait_for_server' | AutocaptureConfig
```
rafaeelaudibert, line 336: @robbie-c Any alternative names? My suggestions are:
- `wait_for_server`
- `await_server`
- `server`

src/autocapture.ts
```diff
@@ -325,6 +325,13 @@ export class Autocapture {
     }
 
     public get isEnabled(): boolean {
+        // Special case where we dont care about the client side config,
+        // or any cached values, we only care about the server side config
+        if (this.instance.config.autocapture === 'wait_for_server') {
```
pauldambra, line 330: fly-by without reading properly

in which case we're not waiting for server we're ignoring client

that's a pretty big change from "client config always overrides server config" which muddies the waters

i'd prefer an explicit 

`is_enabled_source_considered: ('server'|'client')[] = ['server', 'client']`

so that we're separate the config and the decision making

but it's very low context... just the name of the thing stood out for me in notifications

robbie-c: I wrote a lot here and then realised a much simpler way we could solve this user's specific problem. I'll attach what I wrote for posterity.

But maybe the quickest answer is to just have a flag like `ignore_persisted_autocapture_disabled_server_side`

<details>
<summary>old message</summary>
We don't actually have "client config always overrides server config", we have it being enabled only when both:

* we have a response from decide
* both client and server have enabled it

The problem with the current design is that waiting for decide either means that until the decide response, we might do the wrong thing.

We can solve this by:

* Changing the logic, so that if the customer sets the config option (rather than it being set by default), always use this value
* Maybe by storing the autocapture events somewhere??? and only capturing them once decide returns. It might mean we get some events out of order though.

I do think the explicit `is_enabled_source_considered` works as a solution, really there's 3 options though

* server - always disable until decide returns
* client - always use the client value, ignore decide
* prefer_client - use the client value until decide returns (this is tricky right now because we don't have a good way of telling if the value was set by the customer or by default, but we obviously could change this)
* </details>

rafaeelaudibert: Not that it wouldnt work, but it sounds like the most edge-case of the flags ever, and we're only adding it because we can't come up with an elegant solution.

The most elegant would probably be something similar to what you wrote in the old message where we let customers choose who they should trust. This is very powerful and could enable new interesting behaviors. That setting wouldn't apply to `autocapture` only, but the whole `/decide` response.

---

Also, to Paul's original complaint about this not being really `wait_for_server`, we could just call it `server` which then makes it clear we're delegating to the server answer, but I'm in favor of keeping `wait` because it implies we wont be doing anything until the server replies.

pauldambra: so the problem is we cache the setting
for some sessions
if they don't pick up that it has been disabled remotely
then they stay enabled 
until they get a successful decide response

??

is that it?

i'm almost tempted to say that the remote toggles are not instant and that 's ok. if you really want it off you go change the client side setting

🙈 

then there's no code

pauldambra: (purposefully not "requesting changes" so you all can carry on despite my joining the party)

rafaeelaudibert: Yes, and they might never get a decide response for some reason – probably adblocker blocking /decide but not blocking /e so we autocapture something they don't want.

The problem is: it feels like it doesn't work, right? It feels like our server control is flawed and there's no way customers can change it at their will unless they update their code and deploy it to prod again, doesnt sound great

robbie-c: Yeah agreed - IMO this is a bigger problem than what we can handle in this PR but:

Loading posthog-js config asynchronously is problematic because:
* sometimes the request can fail / be blocked / take a long time, so it is ignored
* what should the SDK do before it arrives?

IMO we should put some effort into ways to load this config sync, e.g.
* if using the snippet, it should be bundled with it (I think Ben was working on this, idk if this was finished and is now the way that people use the snippet or not? if so, great)
* if using a framework with SSR (e.g. next), the team-wide config should be loaded by the back-end and be included with SSR

rafaeelaudibert: Yeah, I think you're right. There's probably no right fix for this customer in specific, they just need to use it differently now.

I think Ben's code should work already, but it's currently only enabled for him because he hasnt actually tested it extensively yet: https://us.posthog.com/project/2/feature_flags/89554

IIRC there's some surveys stuff that isnt working on remote config yet. Maybe we should try and expedite that to solve this kind of problem? It'd definitely solve it because the config would be there as soon as the SDK is inited.

---

@benjackwhite no need to read this entire thread, but what are we missing with remote config? Can we help get that across the finish line?

PR #2005, opened 2025-06-06
feat: add import/export compatibility tests and update Rollup config

### LINKED ISSUE #1323: Property default does not exist on type 'PostHog'
I tried to integrate PostHog in my Astro website the following way:

```astro
---
import posthog from 'posthog-js';
---

{
  posthog.init('xxx', {
    api_host: 'https://eu.i.posthog.com',
    persistence: 'memory',
    person_profiles: 'always',
  })
}
``` 

It crashed my website with the following error:

> TypeError: __vite_ssr_import_1__.default.init is not a function

I changed the code to:

```astro
---
import posthog from 'posthog-js';
---

{
  posthog.default.init('xxx', {
    api_host: 'https://eu.i.posthog.com',
    persistence: 'memory',
    person_profiles: 'always',
  })
}
```

This works now in a live environment but shows me an error in my IDE:

> Property 'default' does not exist on type 'PostHog'.

I guess there is something wrong with the default typing?

Best,
Benny

rafaeelaudibert:
We're seeing some problems with CommonJS/ES modules compatibility - we wanna support both.

See https://github.com/PostHog/posthog-js/issues/1323 for completeness

We're now sorting this by properly building and adding an extra default export at the end of the main file.

I had Cursor write me a test file which works very well - it was breaking before. I've added it to CI.

Closes #1323

### DIFF
--- a/.github/workflows/import-export-test.yml
+++ b/.github/workflows/import-export-test.yml
@@ -1,5 +1,8 @@
 name: Import/Export Compatibility Test
 
+permissions:
+    contents: read
+
 on:
     pull_request:
     push:


.github/workflows/import-export-test.yml
```diff
@@ -0,0 +1,23 @@
...
+            - uses: actions/checkout@v4
+            - uses: pnpm/action-setup@v4
+            - uses: actions/setup-node@v4
+              with:
+                  node-version: '18'
+                  cache: 'pnpm'
+            - run: pnpm install
+            - run: pnpm build-rollup
+            - name: Test import/export compatibility
+              run: pnpm test:imports
```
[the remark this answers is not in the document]
rafaeelaudibert: Thank you 🤣

rollup.config.js
```diff
@@ -286,7 +286,12 @@ const entrypointTargets = entrypoints.map((file) => {
                           },
                       }
                     : {}),
-                ...(format === 'cjs' ? { exports: 'auto' } : {}),
+                ...(format === 'cjs'
+                    ? {
+                          exports: 'named',
+                          outro: 'module.exports = exports.posthog; module.exports.default = exports.posthog; Object.assign(module.exports, exports);',
```
hpouillot, line 292: This line is risky, it overrides posthog instance methods and properties with all the commonjs exports. If there is a name collision, it will create weird behaviors depending on the platform, that are hard to catch since we run tests before bundling.

rafaeelaudibert: That's legit, but this is also the line that actually fixes it. I couldnt get it to work without this, and neither could Cursor.

Any other suggestion? We dont do new majors on posthog-js (we probably could do it for this in specific since it only affects those using the npm version and not the snippet)

hpouillot: There is no way to have default and named exports in CJS, so I can see few options : 
- We leave commonjs as it is, we update the doc to promote named exports instead and drop support for it later.
- We make the esmodule the main entrypoint for `posthog-js` but I am not sure about all the implications and retro compatibility issues. I created an example PR here : https://github.com/PostHog/posthog-js/pull/2007/files

rafaeelaudibert: #2007 is too big of a chance, and I don't even think that's what we should do. I'll bring this discussion to the rest of the team. I ultimately believe we should be fine saying we dont support it - it's not working anyway.

PR #2013, opened 2025-06-12
fix: Add warning for when the user tries to identify using the cookieless sentinel distinct ID

robbie-c:
## Changes

Sometimes people store the distinct ID and set it again later. With cookieless, this can do weird things, so let's prevent this.

## Checklist

### DIFF
--- a/src/__tests__/posthog-core.identify.test.ts
+++ b/src/__tests__/posthog-core.identify.test.ts
@@ -242,6 +242,21 @@ describe('identify()', () => {
                 'Unique user id has not been set in posthog.identify'
             )
         })
+
+        it('does not update user', () => {
+            console.error = jest.fn()
+
+            instance.debug()
+
+            instance.identify('$posthog_cookieless')
+
+            expect(beforeSendMock).not.toHaveBeenCalled()
+            expect(instance.register).not.toHaveBeenCalled()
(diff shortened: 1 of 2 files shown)


daibhin (approved):

src/posthog-core.ts
```diff
@@ -1513,6 +1513,12 @@ export class PostHog {
             )
             return
         }
+        if (new_distinct_id === COOKIELESS_SENTINEL_VALUE) {
+            logger.critical(
+                `The string "${COOKIELESS_SENTINEL_VALUE}" was set in posthog.identify which indicates an error. This ID is only use as a sentinel value.`
```
robbie-c:
```suggestion
                `The string "${COOKIELESS_SENTINEL_VALUE}" was set in posthog.identify which indicates an error. This ID is only used as a sentinel value.`
```

src/__tests__/posthog-core.identify.test.ts
```diff
@@ -242,6 +242,21 @@ describe('identify()', () => {
                 'Unique user id has not been set in posthog.identify'
             )
         })
+
+        it('does not update user', () => {
```
robbie-c:
```suggestion
        it('does not update user when distinct ID is $posthog_cookieless', () => {
```

PR #2016, opened 2025-06-16
fix: style problems in multiple/single-choice questions

lucasheriques:
## Changes

This PR fixes two main problems.

**First one, it prevents this line break from happening:**

| Before | After |
|--------|--------|
| CleanShot 2025-06-16 at 00 36 03@2x | CleanShot 2025-06-16 at 00 35 29@2x | 

**Second, it makes sure the label of the checkboxes/radio inputs has the proper, contrasted color:**

| Before | After |
|--------|--------|
| CleanShot 2025-06-16 at 01 17 08@2x | CleanShot 2025-06-16 at 01 16 29@2x | 

### Other misc changes

- Moves the autofocus on for open text directly into the question component. It simplifies the logic, plus now it autofocus for any question, not just the first one (i.e. last question is an open ended)
- Organizes the CSS code to make uncustomizable variables be closer together. No actual change

## Checklist
- [x] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing)) -> all tests passing

### DIFF
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -817,14 +817,6 @@ export function usePopupVisibility(
                 sessionRecordingUrl: posthog.get_session_replay_url?.(),
             })
             localStorage.setItem('lastSeenSurveyDate', new Date().toISOString())
-            setTimeout(() => {
-                const inputField = document
-                    .querySelector(getSurveyContainerClass(survey, true))
-                    ?.shadowRoot?.querySelector('textarea') as HTMLElement
-                if (inputField) {
-                    inputField.focus()
-                }
-            }, 100)
         }
 
         addEventListener(window, 'PHSurveyClosed', handleSurveyClosed as EventListener)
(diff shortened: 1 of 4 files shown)


src/extensions/surveys/survey.css
```diff
@@ -285,7 +286,14 @@
         gap: 8px;
         font-size: 13px;
         align-items: center;
-        flex-wrap: wrap;
+
+        & span {
+            color: inherit;
+        }
```
lucasheriques: this fixes second first problem (white text color on the white background)

src/extensions/surveys/survey.css
```diff
@@ -285,7 +286,14 @@
...
+
+        &.choice-option-open {
+            flex-wrap: wrap;
+        }
```
lucasheriques: and this fixes the line-break problem, by only wrapping if it's an open-ended input

src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -81,20 +81,28 @@ export function OpenTextQuestion({
...
             return initialValue
         }
         return ''
     })
 
+    useEffect(() => {
+        setTimeout(() => {
+            inputRef.current?.focus()
+        }, 100)
+    }, [])
```
lucasheriques: logic to autofocus on the open question input moved to here, where it makes more sense.

marandaneto (approved):

ioannisj (approved):

PR #2022, opened 2025-06-16
fix: don't call ff events for surveys

lucasheriques:
## Changes

don't call feature flag events for survey-related feature flags

### DIFF
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -213,7 +213,7 @@ function getNextToTriggerPosition(target: HTMLElement, surveyWidth: number): Rea
 }
 
 // Keep in sync with posthog/constants.py on main repo
-const SURVEY_TARGETING_FLAG_PREFIX = 'survey-targeting'
+const SURVEY_TARGETING_FLAG_PREFIX = 'survey-targeting-'
 
 export class SurveyManager {
     private _posthog: PostHog
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -580,6 +580,7 @@ export class SurveyManager {
             manageWidgetSelectorListener: this._manageWidgetSelectorListener,
             sortSurveysByAppearanceDelay: this._sortSurveysByAppearanceDelay,
             checkFlags: this._checkFlags.bind(this),
+            isSurveyFeatureFlagEnabled: this._isSurveyFeatureFlagEnabled.bind(this),
         }
     }
 }
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -212,6 +212,9 @@ function getNextToTriggerPosition(target: HTMLElement, surveyWidth: number): Rea
     }
 }
 
+// Keep in sync with posthog/constants.py on main repo
+const SURVEY_TARGETING_FLAG_PREFIX = 'survey-targeting'
+
 export class SurveyManager {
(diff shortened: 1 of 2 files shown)


packages/browser/src/extensions/surveys.tsx
```diff
@@ -386,7 +386,7 @@ export class SurveyManager {
         if (!flagKey) {
             return true
         }
-        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey)
+        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey, { send_event: false })
```
marandaneto: that will stop for all flags, see comment [here](https://posthog.slack.com/archives/C07QD3LT8U9/p1750057416504189?thread_ts=1749823234.899909&cid=C07QD3LT8U9).
we should only set that to our own internal flags only

ioannisj: Yeah, like @marandaneto suggested we could do !startsWith("survey-targeting") here

lucasheriques: oops yeah, in my head I was like "let's not call this for all survey-related feature flags" but makes sense. changed now!

packages/browser/src/extensions/surveys.tsx
```diff
@@ -212,6 +212,9 @@ function getNextToTriggerPosition(target: HTMLElement, surveyWidth: number): Rea
     }
 }
 
+// Keep in sync with posthog/constants.py on main repo
+const SURVEY_TARGETING_FLAG_PREFIX = 'survey-targeting'
```
ioannisj:
```suggestion
const SURVEY_TARGETING_FLAG_PREFIX = 'survey-targeting-'
```

ioannisj (approved):
LG

packages/browser/src/extensions/surveys.tsx
ioannisj: Just so it's the same as constants.py

packages/browser/src/extensions/surveys.tsx
```diff
@@ -386,7 +389,9 @@ export class SurveyManager {
         if (!flagKey) {
             return true
         }
-        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey, { send_event: false })
+        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey, {
+            send_event: !flagKey.startsWith(SURVEY_TARGETING_FLAG_PREFIX),
```
marandaneto, line 393: not sure if an issue but IE11 does not support `startsWith`
I do see that we use `startsWith` elsewhere tho, your call here if either replace or keep it

packages/browser/src/extensions/surveys.tsx
```diff
@@ -386,7 +389,9 @@ export class SurveyManager {
         if (!flagKey) {
             return true
         }
-        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey, { send_event: false })
+        return !!this._posthog.featureFlags.isFeatureEnabled(flagKey, {
```
marandaneto, line 392: Add a test case?

packages/browser/src/extensions/surveys.tsx
lucasheriques: if IE11 doesn't support it, `rollup` will prob take care of it when bundling the JS

packages/browser/src/extensions/surveys.tsx
lucasheriques: done

marandaneto (approved):

PR #2026, opened 2025-06-17
feat: add distinct id tracing header

daibhin:
## Changes

https://github.com/PostHog/posthog-python/pull/263/files#r2152219291

### DIFF
--- a/packages/browser/src/entrypoints/tracing-headers.ts
+++ b/packages/browser/src/entrypoints/tracing-headers.ts
@@ -2,15 +2,16 @@ import { SessionIdManager } from '../sessionid'
 import { patch } from '../extensions/replay/rrweb-plugins/patch'
 import { assignableWindow, window } from '../utils/globals'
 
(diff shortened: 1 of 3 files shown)


oliverb123 (approved):

marandaneto:
@daibhin @oliverb123 i think it'd make more sense to implement this using standardized headers, see issue for additional context https://github.com/PostHog/posthog/issues/20026
A custom header works when the header is traveling from patched SDK A to patched SDK B; however, if not, it usually gets lost in transit (end-to-end tracing even if using a 3rd party).
baggage header also allows passing key-value things. In your case, you'd need a new header again for a new key value.
issues with header bloating, header injection if not encoded (security issue), etc.
there are a few more advantages, happy to discuss if needed.

oliverb123:
[According to the implementation report, nobody has adopted this?](https://w3c.github.io/baggage/reports/cr-2024-implementations) It looks easy to add in a backwards-compatible way, if we decide we want to. Let's not be first.

PR #2027, opened 2025-06-17
feat: enhance campaign parameter retrieval from cookies and URL

rafaeelaudibert:
Some campaign parameters can be set via cookies when using specialized libraries. Let's introduce a new feature that let's us collect this from the cookie store.

See https://learn.microsoft.com/en-us/linkedin/marketing/conversions/enabling-first-party-cookies?view=li-lms-2025-05#reading-li_fat_id-from-cookies for an example

Closes https://github.com/PostHog/posthog/issues/33811


robbie-c (approved):
Nice! I was trying to come up with cases where this is a bad idea, e.g. if we're in a Cordova or Electron app, will trying to access cookies break anything? But it seems like it should just return undefined/empty but otherwise function, which means that this will behave ok

PR #2031, opened 2025-06-20
chore: new survey config

lucasheriques:
## Changes

Adds a new setting, `disable_surveys_automatic_display`. When this is true, no surveys will be shown automatically.

Necessary for [fullscreen surveys](https://github.com/PostHog/posthog/pull/33948), so we don't trigger any surveys on that page.

Also fixes a small issue where we're automatically focusing the open text input even in the survey preview.

### DIFF
--- a/packages/browser/src/types.ts
+++ b/packages/browser/src/types.ts
@@ -474,11 +474,18 @@ export interface PostHogConfig {
     disable_cookie?: boolean
 
     /**
-     * Determines whether PostHog should disable surveys.
+     * Determines whether PostHog should disable all surveys functionality.
      *
      * @default false
      */
     disable_surveys: boolean
+
+    /**
+     * Determines whether PostHog should disable automatic display of surveys. If this is true, popup or widget surveys will not be shown when display conditions are met.
+     *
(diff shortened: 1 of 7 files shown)


ioannisj (approved):
Sounds good. No idea if there was a way to just hijack survey rendering and render nothing instead, but this is a cleaner approach either way

packages/browser/src/types.ts
```diff
@@ -474,12 +474,19 @@ export interface PostHogConfig {
...
      * @default false
      */
     disable_surveys: boolean
 
+    /**
+     * Determines whether PostHog should disable automatic display of surveys. If this is true, popup or widget surveys will not be shown when display conditions are met.
+     *
+     * @default false
+     */
+    disable_surveys_automatic_display: boolean
```
marandaneto, line 488: It's already merged, but the name and description are confusing to me.
we have `disable_surveys` and `disable_surveys_automatic_display`
why would I chose `disable_surveys_automatic_display` over `disable_surveys`?
They do the same thing, I guess? I probably lack some context, and the user probably will too

PR #2032, opened 2025-06-20
feat: allow rendering survey popups on demand

### LINKED ISSUE #1981: Ability to Hide/Show Survey Programmatically
This feature was requested once before (https://github.com/PostHog/posthog/issues/18245) but it doesn't appear to be fully implemented. `renderSurvey` was added (https://github.com/PostHog/posthog-js/pull/1324) to address the feature request but instead of adding the ability to programmatically hide/show, it takes over the rendering process. This leaves us having to provide our own styling and custom elements to support what's rendered. This feature is great if you are just wanting to take over the rendering process.

However, my request is purely around the ability to hide/show it which i believe is what the original issue was about.

I would like a function, `show`/`hide` for example, that allows me to purely render the out of the box survey from posthog. My intent is to have this survey appear when someone clicks a button instead of automatically when the user visits the page. I understand there are options in the UI to do this, but it renders a button on the page that I don't have any control over (outside of basic styling).

I have a custom button, somewhere on the page that i control. When a user clicks that button, I want to show my survey. If it were possible to provide options around the location of _where_ the survey is rendered (over the current element, default, center of screen, etc...), that would be great too.

lucasheriques:
## Changes

make it easy to render popup surveys on demand! this way, there's a simpler way for our customers to render surveys whenever they want.

Closes https://github.com/PostHog/posthog-js/issues/1981

## Checklist


packages/browser/src/posthog-core.ts
```diff
@@ -1425,11 +1425,16 @@ export class PostHog {
...
-    /** Render a survey on a specific element. */
+    /** Render an unstyled survey on a specific element. */
     renderSurvey(surveyId: string, selector: string): void {
         this.surveys.renderSurvey(surveyId, selector)
     }
 
+    /** Render a styled survey popup. Notice that this method ignores any delay or conditions set on the survey. */
+    renderSurveyPopup(surveyId: string): void {
+        this.surveys.renderSurveyPopup(surveyId)
+    }
```
marandaneto: could this be just an extra param to `renderSurvey` method?
its a bit confusing right now, 2 methods with similar names, and the difference is ignoring delay or conditions

lucasheriques: they can be, but the outcome is very different for both of them, which is why I made them different.

`renderSurvey` render the survey form fields, unstyled, in a parent container. Nothing else.

now, `renderSurveyPopup`, or new method, does:

- hooks the survey in the `SurveyManager` for tracking focus states
- render the survey exactly how it's rendered in the Survey Preview. i.e. the floating popover, with all the styles and shadow DOM
- it also ignores both delay and display conditions (answering your question - it ignores all conditions and just displays it)

which is why I decided to make it a separate method. The `renderSurvey` also needs a `selector` parameter, which this one doesn't. JS doesn't have overloaded parameters, so we'd also have to change the signature to accept either a `selector` or `styled` parameter, and potentially change it to an object.

does this make sense?

marandaneto: it does, but that required quite a big explanation, i dont think the name renderSurvey vs renderSurveyPopup will be clear for users tbh, UX here is key so people find what they need and dont run into bugs/different behaviours just because they used the wrong method.

lucasheriques: do you have a suggestion? i don't think using the same method is ideal because of the signature change it requires.

maybe we use a name similar to the event we use? the event is `survey shown`, so we could rename it to `showSurvey`, wdyt?

marandaneto: >  and the difference is ignoring delay or conditions

i thought that was the only difference, so using the same method name does not make sense

`renderSurveyPopup` would ignore conditions which is not what i would expect and it'd be different than `renderSurvey`

maybe  `renderSurveyPopup` takes a param to either ignroe or now the conditions, so at least the behaviour is aligned to `renderSurvey`, default false, so its aligned with renderSurvey

same for renderSurveyPopup

one can:
renderSurvey(..., ignoreConditions: true) // (defualt false)
renderSurveyPopup(..., ignoreConditions: true) // (defualt false)

write better comments on both methods explaining the difference and why we have the new ignoreConditions

thats what i'd think it'd help a bit with a less confusing API/UX. wdyt?

mirusky: As an user of posthog and a developer, I think:

- it should be just one method 
- default implementation should render with style
- unstyled should be an option
- it MUST follow delay and conditions 
- delay and conditions CAN be ignored by flags/options

An alternative to not break the API, can be something like: 

```js
interface DisplaySurveyOptions {
  styled: bool // default: true
  ignoreDelay: bool // default: false
  ignoreConditions: book // default: false
}

displaySurvey(surveyId: string, options?: DisplaySurveyOptions)
```

I know it would require more work, but it will give a better DX, when I first reached renderSurvey method I was a bit disappointed it returned a unstyled version of survey. 

Also it won't introduce a breaking change, neither a misleading naming.

mirusky: And by introducing displaySurvey, we deprecate renderSurvey or just leave as a shortcut/alias.

marandaneto: `displaySurvey` makes sense to me; there's still a difference between rendering a popup or in a selector.
That can be another optional parameter.

```
interface DisplaySurveyOptions {
  styled: bool // default: true
  ignoreDelay: bool // default: false
  ignoreConditions: book // default: false
  displayType: enum // (popup or inlined) default: popup
  selector: string? // default null, requires displayType == inlined
}
```

something like that?

marandaneto: cc @ioannisj for opinions since we will need this on mobile most likely in the future

ioannisj: Yeah, I think the final form of this looks reasonable enough to me. Not sure we want displayType if are passing a selector though?

ioannisj: I mean what will a displayType: popup with a selector do?

marandaneto: in this case, ignore the `selector`
also fine with just the `selector`, not sure if it will be obvious to users tho, they will have to figure out the difference between having and not having the selector, whereas with a `displayType`, you can write comments etc

mirusky: I think, 

**just selector** can lead to miss behaviors for example if someone pass selector but was expecting it to be a popup, it can cause some bugs/issues being falsely reported.

**displayType + selector** is an overload, but if it's well documented + a warning on console saying it's not necessary when popup would make DX good enough and self explanatory.

lucasheriques: @mirusky @marandaneto @ioannisj sorry for the delay here! i was on PTO but back now. i'll implement the changes we discussed here and ping again once ready

packages/browser/CLAUDE.md
```diff
@@ -0,0 +1,87 @@
+# CLAUDE.md
```
lucasheriques, line 1: just ran `/init` using claude code, as I used it for writing couple of things. thought it didn't hurt to commit, but anyone is always free to change it regardless

packages/browser/src/posthog-surveys-types.ts
```diff
@@ -260,3 +260,25 @@ export enum SurveyEventProperties {
...
+interface DisplaySurveyPopoverOptions extends DisplaySurveyOptionsBase {
+    displayType: DisplaySurveyType.Popover
+}
+
+interface DisplaySurveyInlineOptions extends DisplaySurveyOptionsBase {
+    displayType: DisplaySurveyType.Inline
+    selector: string
+}
+
+export type DisplaySurveyOptions = DisplaySurveyPopoverOptions | DisplaySurveyInlineOptions
```
lucasheriques, line 284: @marandaneto @mirusky here's the interface I thought would work best

I didn't add the `styled` flag because i don't think it makes sense:

- Popover surveys are always styled, or they would not be popovers
- We don't support styles for non popovers, so I don't think it's worth adding it would be always false

So example usage:

```typescript
// for popover
posthog.displaySurvey(<surveyId>, {
	displayType: DisplaySurveyType.Popover,
	ignoreDelay: true,
	ignoreConditions: true
}

// for inline - this replaces renderSurvey directly, so in the future we can deprecate it
posthog.displaySurvey(<surveyId>, {
	displayType: DisplaySurveyType.Inline,
	ignoreDelay: true,
	ignoreConditions: true,
	selector: "#survey",
}
```

because the way I defined the type, it means that if you use the inline option, it always requires a selector as well

packages/browser/playground/nextjs/pages/index.tsx
```diff
@@ -88,6 +89,25 @@ export default function Home() {
...
+                        if (!survey) {
+                            return
+                        }
+                        posthog?.displaySurvey(survey.id, {
+                            ignoreConditions: true,
+                            ignoreDelay: true,
+                            displayType: DisplaySurveyType.Popover,
+                        })
+                    }}
+                >
```
lucasheriques, line 108: this is just adding the code to test it in the nextjs playground, can be safely ignored

packages/browser/src/posthog-core.ts
```diff
@@ -1779,6 +1786,18 @@ export class PostHog {
         this.surveys.renderSurvey(surveyId, selector)
     }
 
+    /** Render a styled survey popup. Notice that this method ignores any delay or conditions set on the survey. */
+    displaySurvey(
+        surveyId: string,
+        options: DisplaySurveyOptions = {
+            ignoreDelay: false,
+            ignoreConditions: false,
```
marandaneto: should `ignoreDelay` be part of the conditions? or should it be separated?

packages/browser/src/posthog-surveys-types.ts
```diff
@@ -260,3 +260,25 @@ export enum SurveyEventProperties {
...
+}
+
+interface DisplaySurveyPopoverOptions extends DisplaySurveyOptionsBase {
+    displayType: DisplaySurveyType.Popover
+}
+
+interface DisplaySurveyInlineOptions extends DisplaySurveyOptionsBase {
+    displayType: DisplaySurveyType.Inline
+    selector: string
+}
```
marandaneto, line 282: we should add comment blocks to each of them explaining what they mean
we should also document somewhere here https://posthog.com/docs/surveys/

packages/browser/src/posthog-surveys.ts
```diff
@@ -325,26 +331,72 @@ export class PostHogSurveys {
...
         this._surveyManager.renderSurvey(survey, elem)
     }
+
+    displaySurvey(
+        surveyId: string,
+        options: DisplaySurveyOptions = {
+            ignoreConditions: false,
+            ignoreDelay: false,
+            displayType: DisplaySurveyType.Popover,
+        }
```
marandaneto: we can create a default const for the default param here
we are repeating/recreating the same obj with the same defaults on the core file, so they can be easily out of sync if we change one place and not the other

packages/browser/src/posthog-surveys.ts
```diff
@@ -290,12 +296,12 @@ export class PostHogSurveys {
         return this._surveyManager.checkSurveyEligibility(survey)
     }
 
-    canRenderSurvey(surveyId: string): SurveyRenderReason {
+    canRenderSurvey(survey: string | Survey): SurveyRenderReason {
```
marandaneto: can all the changes of this file be a possible breaking change with people having the bundled sdk installed (before the changes) but loading the latest version of the script?

packages/browser/src/posthog-core.ts
lucasheriques: I believe different, as we also treat it different on the survey form. Delay is under "Customization", but the others are related to "Display conditions" section

marandaneto: should it be `ignoreCustomizations` then? will we have more options soon so the `ignoreDelay` is next to many other things

packages/browser/src/posthog-surveys-types.ts
lucasheriques: I'll add the comments on the main function `displaySurvey` using the JSDOC syntax, this way they are also automatically added to the website [like this](https://posthog.com/docs/references/posthog-js#PostHog-onSurveysLoaded)

packages/browser/src/posthog-surveys.ts
lucasheriques: it won't be, as we are not adding/reming parameters. The type doesn't really matter, as only js files are downloaded anyway

packages/browser/src/posthog-core.ts
lucasheriques: I don't think so, as customizations is about a lot of things, and we can't really have one option for it all. It also includes:

- Appearance (width, padding, font, colors)
- Placeholders
- White labeling
- Shuffle questions
- Lastly, the delay

so if we add more, it'll have to be one for each

packages/browser/src/posthog-surveys.ts
lucasheriques: done!

packages/browser/src/posthog-surveys.ts
lucasheriques: i removed the renaming of the parameter though, so it matches the parameter of `_checkSurveyEligibility` too

packages/browser/src/posthog-surveys-types.ts
lucasheriques: JSDOC added now, will add the docs in the website in another PR

packages/browser/src/posthog-core.ts
lucasheriques: big thread here, marking as resolved, for the new interface, [check this comment](https://github.com/PostHog/posthog-js/pull/2032#discussion_r2337700436)

marandaneto (approved):

PR #2035, opened 2025-06-23
chore: Make nextjs playground work with regular install

robbie-c:
## Changes

...

## Checklist

### DIFF
--- a/packages/browser/playground/nextjs/package.json
+++ b/packages/browser/playground/nextjs/package.json
@@ -5,13 +5,13 @@
     "packageManager": "pnpm@9.15.4",
     "scripts": {
         "clean-react": "cd ../../react && pnpm run clean-node-modules",
-        "dev": "next dev",
+        "dev": "pnpm run clean-react && next dev",
         "dev-crossdomain": "NEXT_PUBLIC_CROSSDOMAIN=1 next dev --experimental-https",
         "install-deps": "pnpm install && pnpm run build-posthog-js && pnpm install --ignore-workspace",
         "build": "next build --no-lint",
         "start": "next start",
         "lint": "next lint",
-        "build-posthog-js": "pnpm --filter=posthog-js build"
+        "build-posthog-js": "cd ../../ && NODE_ENV=dev pnpm i && pnpm run build"
     },
     "dependencies": {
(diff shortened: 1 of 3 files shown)


packages/browser/playground/nextjs/package.json
```diff
@@ -21,8 +21,8 @@
         "cookie": "^0.7.2",
         "eslint": "^8.57.1",
         "hls.js": "^1.5.15",
-        "next": "14.2.20",
-        "posthog-js": "file:../..",
+        "next": "14.2.30",
+        "posthog-js": "file:../../dist",
```
hpouillot, line 25: Is it needed ? It should pickup the package directly ?

packages/browser/playground/nextjs/package.json
```diff
@@ -5,13 +5,13 @@
     "packageManager": "pnpm@9.15.4",
     "scripts": {
         "clean-react": "cd ../../react && pnpm run clean-node-modules",
-        "dev": "next dev",
+        "dev": "pnpm run clean-react && next dev",
         "dev-crossdomain": "NEXT_PUBLIC_CROSSDOMAIN=1 next dev --experimental-https",
-        "install-deps": "pnpm install && pnpm run build-posthog-js && pnpm install --ignore-workspace",
+        "install-deps": "pnpm run build-posthog-js && pnpm install",
```
hpouillot, line 10: vercel deployment silently fails with this config, from memory it's because of multiple react installation

hpouillot (changes requested):

packages/browser/playground/nextjs/package.json
robbie-c: This is fixed by adding `pnpm run clean-react` to `start`

packages/browser/playground/nextjs/package.json
robbie-c: This tells it to use the transpiled version of posthog-js, without this, I got an error around unexpected `export` keyword

packages/browser/playground/nextjs/package.json
robbie-c: NM, no it isn't

PR #2042, opened 2025-06-25
chore: posthog-js-lite migration

hpouillot:
This PR is the base branch for the `posthog-js-lite` migration.

RFC : https://github.com/PostHog/product-internal/pull/758
Follow-up of : https://github.com/PostHog/posthog-js/pull/1991

## Important changes with `posthog-js-lite`
- Add workspace dependencies
- Create a `@posthog/react` package (copied to posthog-js/react for retro-compat)
- Create a `@posthog/core` package (installed as dependencies)
- Use `changesets` for release management. See [here](https://github.com/PostHog/posthog-js/pull/2042/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R50) for more infos
- Use `turbo` as task runner
- Use tarballs for testing package in examples folder. See [here](https://github.com/PostHog/posthog-js/pull/2042/files#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9R1) for more infos
- Add `corepack` compatibility with packageManager field
- Upgrade to pnpm@10.12.4

## Important changes with `posthog-js`
- Temporarily remove alpha release workflow
- Upgrade to node 20 (for development)
- Fix circular dependency between `posthog-js` and `posthog-js/react`

## Last Sync with `posthog-js-lite`
- https://github.com/PostHog/posthog-js/pull/2098

## Migration Plan
- [x] Configuration
		 - [x] Merge linting / prettier configuration. https://github.com/PostHog/posthog-js/pull/2043
- [x] Build process
		- [x] Create a tsconfig base package to share config. https://github.com/PostHog/posthog-js/pull/2045
		- [x] Create a rollup utils package to be shared. https://github.com/PostHog/posthog-js/pull/2047
		- [x] Update babel config to use package specific presets. https://github.com/PostHog/posthog-js/pull/2057
		- [x] Update core package. https://github.com/PostHog/posthog-js/pull/2048
		- [x] Update node package. https://github.com/PostHog/posthog-js/pull/2052
		- [x] Update ai package. https://github.com/PostHog/posthog-js/pull/2046
		- [x] Update react-native package. https://github.com/PostHog/posthog-js/pull/2054
		- [x] Update nextjs-config package. https://github.com/PostHog/posthog-js/pull/2055
		- [x] Update web package. https://github.com/PostHog/posthog-js/pull/2058
[...]

PR #2043, opened 2025-06-25
chore(migration): add linting config

hpouillot:

## Changes

- update linting config
- disable unused rules from posthog-js


.eslintrc.js
```diff
@@ -64,6 +64,32 @@ module.exports = {
...
+                'packages/nextjs-config/**',
+                'packages/react-native/**',
+                'packages/node/**',
+                'packages/web/**',
+            ],
+            rules: {
+                'no-console': 'off',
+                '@typescript-eslint/no-unused-vars': 'off',
+                '@typescript-eslint/naming-convention': 'off',
+                'posthog-js/no-direct-undefined-check': 'off',
```
hpouillot, line 80: we should probably clean linting after merging. posthog-js rules should only apply to `posthog-js` package

packages/ai/.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 2,
```
hpouillot, line 3: we don't have the same tabWidth between `posthog-js` and `posthog-js-lite`. I will keep it inconsistent for now, to avoid too many file changes

daibhin (approved):

PR #2044, opened 2025-06-25
chore(migration): update minimum node version

hpouillot:

## Changes

- update minimum node version for the workspace

### DIFF
--- a/.github/workflows/minimum-version-ts-check.yaml
+++ b/.github/workflows/minimum-version-ts-check.yaml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node-version: [18, 20, 22, 23]
+        node-version: [20, 22, 23]
         typescript-version: ["4.7.2", "5.5.4", "latest"]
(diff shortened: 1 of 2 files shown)


.github/workflows/minimum-version-ts-check.yaml
```diff
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-22.04
     strategy:
       matrix:
-        node-version: [18, 20, 22, 23]
```
daibhin, line 15: Does this drop support for Node 18?

hpouillot: Not really, it just means you cannot develop on the workspace with node version < 20. Node support will be defined inside packages, by the package.json and the ecmascript target (tsconfig).

But I think we should add tests like this one, to make sure final bundles work as expected in different env.

hpouillot: Actually, it changes the node support because babel config uses the current node version for transpilation. I need to change this part as well.

PR #2048, opened 2025-06-26
chore(migration): update core package

hpouillot:

## Changes

- update core package
- add jest config file


packages/core/jest.config.js
daibhin, line 1: Can some of these config files be global?

packages/core/tsconfig.json
```diff
@@ -1,7 +1,12 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "@posthog-tooling/tsconfig-base",
```
daibhin, line 2: What is `@posthog-tooling` is that a separate package or something that lives inside the same repo?

hpouillot: When you publish on npm it should be an organisation, but we only use those packages internally here so we can see it as a namespace.

packages/core/jest.config.js
hpouillot: We could share the config by creating a babel-config package, import it and use this config but it seems overkill for one line. Beside, this config is only used for transpiling tests and I plan on updating this part as well.

PR #2052, opened 2025-06-26
chore(migration): update node package

hpouillot:

## Changes

- update imports with workspace dependencies
- use testing entrypoint from posthog-core
- rename outdir dist for consistency

### DIFF
--- a/packages/node/babel.config.js
+++ b/packages/node/babel.config.js
@@ -1,3 +1,3 @@
 module.exports = {
-  presets: [['@babel/preset-env', { targets: { node: 'current' } }], '@babel/preset-typescript'],
+  presets: [['@babel/preset-env', { targets: { node: '18.0' } }], '@babel/preset-typescript'],
 }


packages/core/tsconfig.json
```diff
@@ -1,7 +1,14 @@
 {
-  "extends": "../tsconfig.json",
+  "extends": "@posthog-tooling/tsconfig-base",
   "compilerOptions": {
+    "rootDir": "./src",
+    "baseUrl": "./src",
+    "outDir": "dist",
+    "module": "CommonJS",
     "composite": true,
+    "declaration": true,
```
daibhin: Somewhat separate note: changes like this are somewhat over my head. Might be a really good idea to do a brownbag on "wtf do these config options mean and which should we be using" for the JS ecosystem

daibhin (approved):
Mainly looks to be standardising config options and  updating file paths

packages/core/tsconfig.json
hpouillot: Ahah, yes, I think eng brownbag is good idea.

PR #2053, opened 2025-06-26
chore: deprecate ip config

lricoy:
## Changes

This issue arises occasionally when people attempt to use this configuration for dropping IP address ingestion, but it currently only adds an `ip=1|0` to the URL. 

Before, this could be used elsewhere to also send properties with `$ip: null`, but this doesn't work anymore.

The only ways to drop IP properties now were documented by @gewenyu99 (thanks!) on this https://posthog.com/tutorials/web-redact-properties

I saw some occurrences of the `ip=` query string being used on other parts, like ingestion: https://github.com/PostHog/posthog/blob/9da8dbe6283837eab7846a05bda1a8e416af1106/rust/capture/src/utils.rs#L111-L121, so I think we should keep it. The benefit of removing is probably less ingress/egress besides the clarity on not being used, but would need more coordination.

More context:

Leftover findings: https://posthog.slack.com/archives/C02E3BKC78F/p1748646235527809
Customer question: https://posthog.slack.com/archives/C02E3BKC78F/p1749750495599039
Team Ingestion take: https://posthog.slack.com/archives/C08JQTX5RRP/p1749755116923149

## Checklist

### DIFF
--- a/packages/browser/playwright/utils/setup.ts
+++ b/packages/browser/playwright/utils/setup.ts
@@ -118,6 +118,7 @@ export async function start(
                 const opts: Partial<PostHogConfig> = {
                     api_host: 'https://localhost:1234',
                     debug: true,
+                    ip: false, // Prevent IP deprecation warning in Playwright tests
                     before_send: (event) => {
                         const win = window as WindowWithPostHog
                         win.capturedEvents = win.capturedEvents || []


packages/browser/src/posthog-core.ts
```diff
@@ -163,7 +163,7 @@ export const defaultConfig = (defaults?: ConfigDefaults): PostHogConfig => ({
     disable_external_dependency_loading: false,
     enable_recording_console_log: undefined, // When undefined, it falls back to the server-side setting
     secure_cookie: window?.location?.protocol === 'https:',
-    ip: true,
+    ip: false,
```
lricoy, line 166: As mentioned, I don't think it's a breaking change because it's not doing anything on the servers right now.

packages/browser/src/types.ts
```diff
@@ -539,6 +539,7 @@ export interface PostHogConfig {
      *
      * @default true
```
rafaeelaudibert:
```suggestion
     * @default false
```

packages/browser/src/types.ts
rafaeelaudibert, line 1: Can you extend the documentation in here to note that besides being deprecated, it doesn't actually have any effect? You can see how we explain it for other deprecated properties at the bottom of the config docs

packages/browser/src/posthog-core.ts
```diff
@@ -607,6 +607,12 @@ export class PostHog {
             this.on('eventCaptured', (data) => this.config._onCapture(data.event, data))
         }
 
+        if (this.config.ip) {
+            logger.warn(
+                'The `ip` config option is deprecated. See https://posthog.com/tutorials/web-redact-properties#hiding-customer-ip-address for more information on how to discard IP addresses.'
+            )
+        }
```
rafaeelaudibert: Can we be clearer and say that there's NO effect at all on setting this to `true`? If I read this message I'd think I need to change my approach but that the old one still works, while that's not true at all

gewenyu99:
Great. Another mystery solved 🤣 

API pollution is a real problem, glad it's fixed.

packages/browser/src/types.ts
```diff
@@ -987,6 +980,12 @@ export interface PostHogConfig {
 
     /** @deprecated - NOT USED ANYMORE, kept here for backwards compatibility reasons */
     inapp_link_new_window?: boolean
+
+    /**
+     * @deprecated - THIS OPTION HAS NO EFFECT, kept here for backwards compatibility reasons.
```
gewenyu99, line 985: I will try to support the `@deprecated` flag in the generated specs :D

gewenyu99 (approved):

packages/browser/src/posthog-core.ts
lricoy: Agreed and done!

rafaeelaudibert (approved):

lricoy:
Fixing the tests locally ♻️

PR #2054, opened 2025-06-26
chore(migration): update react-native package

hpouillot:

## Changes

- change import path
- update tsconfig

### DIFF
--- a/packages/react-native/src/storage.ts
+++ b/packages/react-native/src/storage.ts
@@ -1,4 +1,4 @@
-import { utils } from 'posthog-core'
+import { isPromise } from 'posthog-core/utils'
 import { PostHogCustomStorage } from './types'
 
 const POSTHOG_STORAGE_KEY = '.posthog-rn.json'
@@ -16,7 +16,7 @@ export class PostHogRNStorage {
 
     const preloadResult = this.storage.getItem(POSTHOG_STORAGE_KEY)
 
-    if (utils.isPromise(preloadResult)) {
+    if (isPromise(preloadResult)) {
       this.preloadPromise = preloadResult.then((res) => {
         this.populateMemoryCache(res)
       })
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -31,6 +31,10 @@
     "./testing": {
       "types": "./dist/testing/index.d.ts",
       "default": "./dist/testing/index.js"
+    },
+    "./utils": {
+      "types": "./dist/utils.d.ts",
+      "default": "./dist/utils.js"
     }
   },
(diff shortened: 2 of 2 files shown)


packages/react-native/src/storage.ts
```diff
@@ -1,4 +1,4 @@
-import { isPromise } from '../../posthog-core/src/utils'
+import { utils } from 'posthog-core'
```
marandaneto: its better to import `isPromise` only I guess

packages/react-native/tsconfig.json
```diff
@@ -1,12 +1,13 @@
...
   "extends": "@posthog-tooling/tsconfig-base",
   "compilerOptions": {
+    "rootDir": ".",
     "declaration": true,
     "esModuleInterop": true,
     "jsx": "react-native",
     "lib": ["ESNext"],
     "module": "ESNext",
-    "moduleResolution": "node16",
+    "moduleResolution": "bundler",
```
marandaneto, line 10: we need to be sure this is not breaking anything
https://reactnative.dev/blog/2023/06/21/package-exports-support#enabling-package-exports-beta
i recall changing [something around this for a reason](https://github.com/PostHog/posthog-js-lite/pull/294/files#r1815663690), maybe this is ok, but it has to be tested

marandaneto:
somehting i've not tested is also the release process and the specific RN examples https://github.com/PostHog/posthog-js-lite/tree/main/examples

hpouillot:
> quite a few files missing if you compare [PostHog/posthog-js-lite@`main`/posthog-react-native/src](https://github.com/PostHog/posthog-js-lite/tree/main/posthog-react-native/src?rgh-link-date=2025-06-30T12%3A20%3A14Z) [PostHog/posthog-js-lite@`main`/posthog-react-native/src/optional](https://github.com/PostHog/posthog-js-lite/tree/main/posthog-react-native/src/optional?rgh-link-date=2025-06-30T12%3A20%3A14Z) is missing [PostHog/posthog-js-lite@`main`/posthog-react-native/src/frameworks](https://github.com/PostHog/posthog-js-lite/tree/main/posthog-react-native/src/frameworks?rgh-link-date=2025-06-30T12%3A20%3A14Z) is missing and probably more files?

I need to integrate changes from latest PRs from posthog-js-lite, but optionnal and framework should have the same files.

packages/react-native/src/storage.ts
hpouillot: It should not make a difference, posthog-core needs to be bundled with this package but I can add a utils entrypoint on posthog-core.

PR #2056, opened 2025-06-27
fix: init surveys manager if surveys is disabled in the project settings

### LINKED ISSUE #2050: Surveys loading are not respecting clients that are manually creating surveys
Since we can manually render surveys by the SDK clients we should not respect this setting during the load of the surveys module:
https://github.com/PostHog/posthog-js/blob/1e907ac1f358ef1dbbb94f63631a3055bd03ee60/packages/browser/src/posthog-surveys.ts#L39

We were using an older [version of SDK](https://github.com/PostHog/posthog-js/blob/4dc9c0e734c27c7a2eeb847910a1a6d3b99c9484/src/posthog-surveys.ts#L81) and once updated to the latest one was surprized that our surveys are not fetched at all.

marandaneto:
## Changes

Fix https://github.com/PostHog/posthog-js/issues/2050

## Checklist


lucasheriques (approved):
I tested this, and looks good.

For the prev changes, my reasoning was to, if surveys was disabled, just don't do anything (and just return). However I agree that, if we're loading surveys manager anyway, it makes sense to return its instance and just not do any of the automatic call and evaludate display logic.

Tested this with API Surveys and it disabled on the project and it's working well

lucasheriques (commented):
the goal of the original PR was to don't initialize Surveys if it was disabled on the project level.

It seems it also caused the unintended side effect where, a customer using only API Surveys, still couldn't access the Survey APIs.

This PR addresses the first issue (APIs like `getSurveys` and `getActiveMatchingSurveys`) are now working.

But it introduces a new issue: now, even with team survey settings disabled, the popups are still showing up. so we'll have to come up with a different logic. I'll follow up on this once I have a better approach

lucasheriques:
@marandaneto [check this Loom](https://www.loom.com/share/0a2cf5907ba446f1b0557554a3fae464) I think it should explain why this is not working as expected

marandaneto:
> @marandaneto [check this Loom](https://www.loom.com/share/0a2cf5907ba446f1b0557554a3fae464) I think it should explain why this is not working as expected

oh i see, the fix should be simple then
[`e45008d` (#2056)](https://github.com/PostHog/posthog-js/pull/2056/commits/e45008db53d0b26f54e9619064613856fef39fe3)
i was not considering that `loadIfEnabled` is called multiple times before remote config.
the surveys should only be used after everything is loaded, including remote config.
so in this case is ok to guard against waiting remote config to finish and only continue after we know about `isSurveysEnabled`
can you retest with this fix?

lucasheriques (approved):
looks good now! working as expected

PR #2061, opened 2025-06-30
chore: make it easier to run locally

veryayskiy:
## Changes

When we work with playground we mostly do it locally. So let's simplify the process.

Added a simple switch to run locally + dotenv to get the key from .env


pauldambra (approved):
all in a playground folder... zero read 👍 🤣

PR #2064, opened 2025-06-30
chore(migration): update lint and test jobs

hpouillot:

## Changes

- update lint and test jobs
- move react package in its own folder (there is a circular dependencies that breaks installation)
- copy react assets into browser package (for retrocompat)
- fix prepublish script


packages/react/src/components/__tests__/PostHogErrorBoundary.test.jsx
```diff
@@ -72,7 +72,7 @@ describe('captureException processing', () => {
         const exceptionList = captureCalls[0][1].$exception_list
         expect(exceptionList.length).toBe(1)
         const stacktrace = exceptionList[0].stacktrace
-        expect(stacktrace.frames.length).toBe(36)
+        expect(stacktrace.frames.length).toBe(44)
```
daibhin, line 75: Why did this change? I'm guessing folder structure but really not sure

packages/ai/rollup.config.mjs
```diff
@@ -1,5 +1,5 @@
 import { external, babel, resolve, commonjs, json, dts, plugins } from '@posthog-tooling/rollup-utils'
-import packageJson from './package.json'
+import packageJson from './package.json' with { type: 'json' }
```
daibhin, line 2: TIL

packages/react/src/components/__tests__/PostHogErrorBoundary.test.jsx
hpouillot: Yes, it could be because of folder structure or jest configuration changes, I don't know

PR #2065, opened 2025-06-30
fix: word break on open ended choices

lucasheriques:
## Changes

Before:

CleanShot 2025-06-30 at 14 40 46@2x

After:

CleanShot 2025-06-30 at 14 39 12@2x

## Checklist


marandaneto (approved):

corywatilo:
Might be good to align-items: start so the checkbox/radio doesn't move down as text wraps further!

PR #2072, opened 2025-07-02
chore(migration): update publish workflow

hpouillot:

## Changes

- update all package version on PR merged using changeset
- release new package version on main push
- upgrade posthog.com and posthog on new posthog-js version


.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+
+jobs:
+  release:
+    name: Publish
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        package:
+          - name: posthog-react-native
+            npm_token_secret: NPM_TOKEN_BEN_WHITE # TODO: Change to standard token once able
```
hpouillot, line 16: might be worth replacing this as well

.github/workflows/label-version-bump.yml
```diff
@@ -6,15 +6,12 @@ on:
...
     runs-on: ubuntu-22.04
     if: |
       github.event.pull_request.merged
-      && (
-          contains(github.event.pull_request.labels.*.name, 'bump patch')
-          || contains(github.event.pull_request.labels.*.name, 'bump minor')
-          || contains(github.event.pull_request.labels.*.name, 'bump major')
-      )
+      && contains(github.event.pull_request.labels.*.name, 'release')
+      && github.event.pull_request.base.ref == 'main'
```
daibhin, line 14: I probably missed it elsewhere but how do we know what bump type to use without these labels? I'm guessing we are reading the changelog in some way

daibhin (commented):
Seems reasonable, certainly a tricky one to test though

.github/workflows/label-version-bump.yml
hpouillot: https://github.com/PostHog/posthog-js/pull/2072/files/a612a57c38b7bf67df6382a7ac9d3a42be805e40#diff-833739ba7ba9c2fcca8ffb8c723547ba58fe0b4418c2ad7d8233c761fc5d220dR30

This part is a bit magical but this line updates the packages version and changelog. When you create a PR, you add a changeset (using the cli) that specifies the changes and the version bump type for the package you changed then when this command runs it consumes those changes and update package versions (by taking workspace dependencies into account).

daibhin (approved):

PR #2076, opened 2025-07-04
refactor: Improve instance initialization in PostHog class

rafaeelaudibert:
This PR does two-fold:

- Ensured proper assignment of instances to the primary PostHog instance for better debugging.
- Ensure `window.posthog` is defined even in cases where we're using the SDK (makes debugging much easier)

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -423,19 +423,31 @@ export class PostHog {
         config?: OnlyValidKeys<Partial<PostHogConfig>, Partial<PostHogConfig>>,
         name?: string
     ): PostHog {
+        let instance: PostHog
         if (!name || name === PRIMARY_INSTANCE_NAME) {
             // This means we are initializing the primary instance (i.e. this)
-            return this._init(token, config, name)
+            instance = this._init(token, config, name)
         } else {
-            const namedPosthog = instances[name] ?? new PostHog()
-            namedPosthog._init(token, config, name)
-            instances[name] = namedPosthog
-
-            // Add as a property to the primary instance (this isn't type-safe but its how it was always done)
-            ;(instances[PRIMARY_INSTANCE_NAME] as any)[name] = namedPosthog
-
-            return namedPosthog
+            instance = instances[name] ?? new PostHog()
+            instance = instance._init(token, config, name)
+            instances[name] = instance
         }
+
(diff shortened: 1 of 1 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -423,19 +423,31 @@ export class PostHog {
...
+            instance = instance._init(token, config, name)
+            instances[name] = instance
+        }
 
-            // Add as a property to the primary instance (this isn't type-safe but its how it was always done)
-            ;(instances[PRIMARY_INSTANCE_NAME] as any)[name] = namedPosthog
+        // Add as a property to the primary instance (this isn't type-safe but its how it was always done)
+        if (instances[PRIMARY_INSTANCE_NAME] && name) {
+            const anyInstance = instances[PRIMARY_INSTANCE_NAME] as any
+            anyInstance[name] = instance
```
hpouillot, line 439: Any reason to create an access at the instance level ? I'm always wary when reassigning properties as it might create conflicts. If we want to get a global access to all instances, maybe we can create a global `__posthog_instances` object ?

robbie-c:
FYI, you can access this on customer sites by adding ?__posthog_debug=true to the end of the URL. This will console.log the instance (and some other stuff), and from there you can use the browser's built in dev tools to get access to the instance

rafaeelaudibert:
> This will console.log the instance (and some other stuff), and from there you can use the browser's built in dev tools to get access to the instance

How do you get access to the `posthog` instance after it's logged? You cant reference it directly. Ive been using the log to get access to the code, then adding a breakpoint, refreshing the page, and once we hit the breakpoint I can do `window.posthog = this`. It's slightly too much work IMO, I wanna simplify that

packages/browser/src/posthog-core.ts
rafaeelaudibert: I just moved this around, this is not a change. The lines below are the changes on this PR.

rafaeelaudibert: See old line 435

hpouillot: Ok my bad, I missed this line

robbie-c:
At least in Firefox you can do `store as global variable`:

Screenshot 2025-07-07 at 13 31 58

It gets added to the global scope as tempN (e.g. `temp0`) and I can access it from the console

Screenshot 2025-07-07 at 13 33 03

PR #2079, opened 2025-07-08
chore(flags): roll `/flags` to the remaining holdouts

dmarticus:
## Changes

- prefer `/flags` over `/decide` for all of the latest versions of `posthog-js`, including the final holdouts
- removes the dependency on an experimental config setting for `/flags` since it's supported everywhere by default now

Let's rip!


packages/browser/src/posthog-featureflags.ts
```diff
@@ -546,14 +393,9 @@ export class PostHogFeatureFlags {
 
         // flags supports loading config data with the `config` query param, but if you're using remote config, you
         // don't need to add that parameter because all the config data is loaded from the remote config endpoint.
-        const useRemoteConfigWithFlags =
-            this._instance.config.__preview_flags_v2 && this._instance.config.__preview_remote_config
```
dmarticus, line 550: now that `/flags` is the default route for evaluating feature flags, we don't need to use this preview opt-in any more (although we definitely still need it for the RemoteConfig stuff).

PR #2084, opened 2025-07-11
chore(migration): rename core package

hpouillot:

## Changes

- rename posthog-core into @posthog/core


tooling/rollup-utils/src/index.ts
```diff
@@ -35,12 +35,6 @@ export function external(packageJson: any) {
         ...listDeps(packageJson.devDependencies),
     ])
 
-    // posthog-core is internal for now
```
daibhin, line 38: I'm assuming we don't need to exclude `@posthog/core` here?

hpouillot: Yes, the goal is to start publishing @posthog/core as a regular dependencies and make other packages depends on it (so we can keep it in externals)

PR #2085, opened 2025-07-11
chore(migration): update examples folder

hpouillot:

## Changes

- build @posthog/core as dual-package, using rslib
- update cloudflare
- update expo 53
- update gcp functions
- update nextjs
- update node
- update web
- update rn_macos:
		I have a build [issue](https://stackoverflow.com/questions/77738691/error-installing-boost-verification-checksum-was-incorrect-expected) caused by boost checksum. I have the same on posthog-js-lite as well


.gitignore
```diff
@@ -38,3 +38,4 @@ dist
 tsconfig.tsbuildinfo
 .turbo
 target
+.yarn
```
daibhin: Should we be using pnpm over yarn?

examples/example-gcp-functions/package.json
```diff
@@ -1,20 +1,20 @@
...
+        "start": "functions-framework --target=helloWorld"
+    },
+    "author": "",
+    "license": "ISC",
+    "dependencies": {
+        "@google-cloud/functions-framework": "^3.4.0",
+        "posthog-node": "*"
+    },
+    "resolutions": {
+        "posthog-react-native": "file:../../target/posthog-react-native.tgz",
```
daibhin: Why do we use react native and not node here?

examples/example-expo-53/package.json
```diff
@@ -1,61 +1,62 @@
...
-    "prestart": "yarn build:posthog",
-    "start": "expo start",
-    "reset-project": "node ./scripts/reset-project.js",
-    "android": "expo run:android",
-    "ios": "expo run:ios",
-    "web": "expo start --web",
-    "lint": "expo lint",
-    "build:posthog": "yarn build:posthog:publish && yarn build:posthog:link",
-    "build:posthog:publish": "cd ../../posthog-react-native && yalc publish --replace",
-    "build:posthog:link": "yalc add posthog-react-native"
```
daibhin, line 16: Can we remove yalc as a dependency?

daibhin (commented):
Might be worth adding to the README if any instructions on building / running has changed

.gitignore
hpouillot: Yes, I switched to pnpm-workspace, I wanted to check if it was easier to overrides dependencies with yarn but there is no way to do it dynamically

examples/example-gcp-functions/package.json
hpouillot: That's the issue with static resolutions, you end up copy pasting overrides... I removed this part

examples/example-expo-53/package.json
hpouillot: Yes, there is no need to use yalc anymore. Packages are exported and installed as tarballs now.

hpouillot: https://github.com/PostHog/posthog-js/pull/2085/files#diff-49aaa2819e35a856818ecec8c9fa7e1c79ad028d3f44bd749736353cfb51bac9R1

PR #2088, opened 2025-07-14
chore(tests): move from testcafe to playwright

hpouillot:
## Problem

- tests running with testcafe and browserstack are flaky

## Changes

- migrate tests from testcafe to playwright
- run playwright test in parallel (by browser)


.github/workflows/library-ci.yml
```diff
@@ -21,27 +21,6 @@ jobs:
         working-directory: packages/browser
       - run: git diff --exit-code # fail if e.g. the mangled properties list has changed, see rollup.config.js
 
-  integration:
```
pauldambra, line 24: two significant changes here

1. no longer running these tests on push to main

 i think we should still run them, just to catch the very rare risk of things being valid in the PR but not once merged

2. we've gone from two e2e suites - one of which uses posthog cloud - to one e2e suite without that distinction

 it'd be good and my instinct is faster to run them as two suites. the replay ones don't hit posthog cloud iirc and would be good to keep them clearly separate

packages/browser/package.json
```diff
@@ -27,14 +27,14 @@
         "test-watch": "jest --watch src",
         "test:typecheck": "cd src/__tests__ && tsc --noEmit --project tsconfig.json",
         "typecheck": "tsc --noEmit --project tsconfig.json",
-        "playwright": "pnpm exec playwright test --project webkit --project firefox --project chromium",
+        "playwright": "pnpm exec playwright test --project webkit --project firefox --project chromium --grep-invert 'ingestion'",
```
pauldambra: i wish json allowed comments
i'm worry that nobody will know what `grep-invert ingestion` does

i wonder if it's easier to have two commands that remove the need to do that

pauldambra:
i don't think we use testcafe for any particular reason https://github.com/PostHog/posthog-js/pull/180
so 👍 on this in principle

i'll want to run the tests locally just to validate the replay ones are still firing 😅 
shout when it's not draft

packages/browser/package.json
robbie-c: Could run a bash file with comments

hpouillot:
Closing this one, new PR is here  => https://github.com/PostHog/posthog-js/pull/2093

PR #2089, opened 2025-07-15
feat: ignore extension exceptions

daibhin:
## Changes

Relates to https://posthoghelp.zendesk.com/agent/tickets/34164 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/36203)

We capture exceptions thrown by extensions on a users browser. These are typically low value because 1) they do not relate to our customers code and 2) we cannot demangle the stack traces.

Making a config change to stop capturing exceptions containing the `chrome-extension://` protocol as the filename.

If you want to test this working it's probably easiest to create a Chrome extension that throws an exception:

```javascript
// manifest.json

{
  "manifest_version": 3,
  "name": "Page Button Error Thrower",
  "version": "1.0",
  "content_scripts": [
    {
      "matches": ["<all_urls>"],
      "js": ["content.js"]
    }
  ],
  "permissions": ["scripting"],
  "web_accessible_resources": [
    {
      "resources": ["inject.js"],
      "matches": ["<all_urls>"]
    }
  ]
}

// content.js

const script = document.createElement("script");
script.src = chrome.runtime.getURL("inject.js");
document.documentElement.appendChild(script);
script.remove();

// inject.js

const btn = document.createElement("button");
btn.textContent = "Throw Error";
btn.style.position = "fixed";
btn.style.bottom = "20px";
btn.style.right = "20px";
btn.style.zIndex = 9999;

btn.addEventListener("click", () => {
  throw new Error("This is an error thrown in the page context!");
});

document.body.appendChild(btn);
```


packages/browser/src/posthog-exceptions.ts
```diff
@@ -18,23 +19,36 @@ export class PostHogExceptions {
...
+        return enabled_client_side ?? enabled_server_side ?? false
+    }
+
     sendExceptionEvent(properties: Properties) {
         if (this._matchesSuppressionRule(properties)) {
             logger.info('Skipping exception capture because a suppression rule matched')
             return
         }
 
+        if (!this._captureExtensionExtensions && this._isExtensionException(properties)) {
```
hpouillot:
```suggestion
        if (!this._captureExtensionExceptions && this._isExtensionException(properties)) {
```

hpouillot (approved):

packages/browser/src/posthog-exceptions.ts
hpouillot: Looks like a small typo

PR #2091, opened 2025-07-16
chore: only evaluate display logic for in-app surveys

lucasheriques:
## Changes

changes the `callSurveysAndEvaluateDisplayLogic` surveys' method to only specifically evaluate surveys with relevant display logic UI (i.e. popover or feedback button surveys)

@marandaneto @ioannisj I think this doesn't require any changes for [React Native per this line](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-react-native/src/surveys/PostHogSurveyProvider.tsx#L109). [Similar for iOS too](https://github.com/PostHog/posthog-ios/blob/1c2e98eb20aa9218be12a9fded58e22cc8451514/PostHog/Surveys/PostHogSurveyIntegration.swift#L253-L257), though let me know if not!

Part of https://github.com/PostHog/posthog/pull/33948

## Checklist


marandaneto:
@ioannisj same here https://github.com/PostHog/posthog-ios/blob/1c2e98eb20aa9218be12a9fded58e22cc8451514/PostHog/Models/Surveys/PostHogSurveyEnums.swift#L12 otherwise it will be `unknown` which is fine since the logic is not gonna break anything

packages/browser/src/extensions/surveys.tsx
```diff
@@ -418,6 +418,12 @@ export class SurveyManager {
             return eligibility
         }
 
+        if (survey.type === SurveyType.ExternalSurvey) {
+            eligibility.eligible = false
+            eligibility.reason = `External surveys are never eligible to be shown in the app`
+            return eligibility
+        }
```
marandaneto: what will happen if theres a `ExternalSurvey` and an older version of this SDK? is the survey gonna be shown as a popover?

marandaneto: should we check for `survey.type === SurveyType.Popover || survey.type === SurveyType.Widget` instead so adding new types wont break anything?

lucasheriques: nothing is going to happen for two reasons:

1. when the type is `ExternalSurvey`, nothing happens, as we only have handlers for Widget or Popover surveys
2. as we discussed in the incident, at least for now, the extensions code is always the latest one we released. so even older SDKs still get the most up to date code
3. I [added a new check](https://github.com/PostHog/posthog-js/pull/2091/commits/861ba7c641046c400882641525e545851c7c4673) to explicitly see if we're dealing for an in app survey here

marandaneto: you can still bundle the whole SDK when installing via npm and in this case you wont get the latest version of the extensions, so we have to account for this use case as well.
right now i still dont know if external surveys will be shown for those users or not since your answer is always mentioning the 'latest version', can you clarify that?

marandaneto: https://posthog.com/docs/session-replay/installation
see 'bundle all required extensions'

lucasheriques: > when the type is ExternalSurvey, nothing happens, as we only have handlers for Widget or Popover surveys

this is always true, irrelevant of their SDK version

marandaneto: I dont follow why we need this PR then? If the current behavior does the right thing

lucasheriques: it's mostly to not evalute external surveys on the `getActiveMatchingSurveys`. since they don't have any display conditions, it doesn't make sense to keep them there

marandaneto: So in this case if people have an external survey and use this method manually, a survey will
Be shown wrongly?

lucasheriques: no, because `getActiveMatchingSurveys` doesn't render any surveys. it checks for surveys who are matching to users and just returns an array with the surveys. but now that will mention, `renderSurvey` will work regardless, [so I made this change to also prevent it](https://github.com/PostHog/posthog-js/pull/2091/commits/a24aa0ff8ae255b95af5eaa7fdc92facee3defdb)

marandaneto: So then theres a behavior change from older sdk versions (that was my point from the beginning), and we should notify users about it before they create any new external surveys.
Please add a label to the frontend change telling about this and that they should use the min. $sdk.version otherwise there will be a side effect which is showing the survey wrongly

lucasheriques: I wouldn't say there is a behavior change because this is a new type. External surveys didn't exist before, now they do. And still, even for older SDKs, external surveys are never rendered automatically, only if the customer would actually go and call `renderSurvey` on them manually

lucasheriques: happy to add the min sdk change, but I don't think it adds much, because external surveys do not require the SDK anyway, since we host them

marandaneto (approved):

packages/browser/src/extensions/surveys.tsx
marandaneto: how else would you suggest then? they need the SDK to avoid this side effect of:

```
posthog.getActiveMatchingSurveys((surveys) => {
      if (surveys.length > 0) {
        const survey = surveys[0]; // this can be an external survey using the SDK before this PR gets merged and released
        setSurvey(survey) // or call `renderSurvey`
      }
    });
```

this is exactly what we do here: https://posthog.com/docs/surveys/implementing-custom-surveys#fetching-surveys-manually as an example and this is what users do as well since public API.

so either i am missing something or you still dont get why this is causing a side effect for older versions, i am not sure which one is the case because you insist its not a problem.

ioannisj: Yeah, I do see the issue @marandaneto raised as well, but maybe I'm missing some context too. I think the goal is to have getActiveMatchingSurveys() not return any external surveys at all, so people don't try to render them. This will help with maintaining backwards compatibility as well

> since they don't have any display conditions

What if we do attach display conditions that will always evaluate to false for external surveys, so we don't have to worry about getActiveMatchingSurveys()?

lucasheriques: > so either i am missing something or you still dont get why this is causing a side effect for older versions, i am not sure which one is the case because you insist its not a problem.

I still think it's not a problem because, even if it's an external survey, it is still an survey. And all methods work on it as expected. People can render them, and they will work, it's just a behavior we do not want to encourage/support since it breaks the whole point of them.

I did change to make `renderSurvey` work again with external surveys, because this is what I'm actually using to render them on the external surveys PR 😅

marandaneto: i think it is a problem, you create an external survey without any display / complete conditions and you share the link with users A, suddently your app shows this survey for either all users or the wrong users which mess up with your data/responses
external surveys should not be rendered automatically, its a link where people open/answer by themselves.
if that requires a new SDK version, we should warn users before they go ahead and create external surveys to avoid the side effect (that is the whole point of this discussion).


> I did change to make renderSurvey work again with external surveys, because this is what I'm actually using to render them on the external surveys PR 😅

IMO you should then find another way/expose a private method only for that or else.

i think we should discuss on the daily because its a long back and forth.

lucasheriques:
> Another possible workaround is that the surveys API and remote config don't return external surveys, only if you pass a query parameter like ?include_external_surveys=true, just thinking out loud.

I think that's a fine idea, but it'll also require changes in the main API. For now, I'd like to merge that as is, [but I created this as an issue](https://github.com/PostHog/posthog-js/issues/2112) so I don't forget about it

packages/browser/src/posthog-surveys.ts
```diff
@@ -327,6 +328,32 @@ export class PostHogSurveys {
...
+            return
+        }
+        if (!elem) {
+            logger.warn('Survey element not found')
+            return
+        }
+        this._surveyManager.renderSurvey(survey, elem)
+    }
+
+    private _renderExternalSurvey(surveyId: string, selector: string) {
```
marandaneto, line 346: you can extract this method and renderSurvey to another method that takes a `type` list as a param, this method passes `[SurveyType.ExternalSurvey]` and the other one `IN_APP_SURVEY_TYPES`, so you dont have to duplicate the whole method body twice next to each other because of one single condition

marandaneto:
```
_renderExternalSurvey(...) {
  _renderAnySurvey(..., [SurveyType.ExternalSurvey])
}
```

```
renderSurvey(...) {
  _renderAnySurvey(..., [IN_APP_SURVEY_TYPES])
}
```

```
_renderAnySurvey(...) {
   ...
   this._surveyManager.renderSurvey(survey, elem)
}
```

lucasheriques: great idea, just did it

PR #2093, opened 2025-07-17
chore(tests): use playwright for integration tests

hpouillot:
Follow-up on PR: https://github.com/PostHog/posthog-js/pull/2088

## Changes

- remove testcafe in favor of playwright for testing integration
- create playwright fixtures
- stop testing ie11 as it is not compatible with playwright


robbie-c:
Losing IE11 testing is a blow, as we theoretically support it. It also gives us some peace of mind for other legacy browsers like UC and pre-Chrome Android, which have only partial support for ES2015. It also means that the ES5 build is not tested.

IMO we should at least test the ES5 build (maybe on Edge?). We'd then have to decide if building for ES5 and testing on a modern browser gives us enough confidence.

Actually maybe let's check metabase to see how much of a big deal this actually is, sec

robbie-c:
https://metabase.prod-us.posthog.dev/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJuYXRpdmUiOnsicXVlcnkiOiJzZWxlY3RcbiAgICBtYXRfJGJyb3dzZXIgYXMgYixcbiAgICBjb3VudCgpIGFzIGMsXG4gICAgYyAqIDEwMC4wIC8gc3VtKGMpIE9WRVIgKCkgQVMgcGNcbmZyb20gZXZlbnRzIHdoZXJlIHRpbWVzdGFtcCA-IHRvZGF5KCkgYW5kIGIgaXMgbm90IG51bGwgYW5kIGIhPSAnJyBncm91cCBieSBiIG9yZGVyIGJ5IGMgZGVzYyIsInRlbXBsYXRlLXRhZ3MiOnt9fSwiZGF0YWJhc2UiOjQyfSwiZGlzcGxheSI6InRhYmxlIiwicGFyYW1ldGVycyI6W10sInZpc3VhbGl6YXRpb25fc2V0dGluZ3MiOnt9fQ==

So UC browser is 0.01%

packages/browser/playwright/mocked/session-recording/session-recording-masking.spec.ts
```diff
@@ -1,7 +1,7 @@
 import { test, expect } from '../utils/posthog-playwright-test-base'
 import { start } from '../utils/setup'
 import { Page } from '@playwright/test'
-import { CaptureResult } from '../../src/types'
+import { CaptureResult } from '@/types'
```
daibhin, line 4: Is the way we reference the path important or just stylistic? Maybe we should lint against relative paths if there's a legitimate downside

packages/browser/playwright/integration/checks.ts
```diff
@@ -0,0 +1,54 @@
+import { expect } from '@playwright/test'
+
+export async function assertCustomEventsWorkAndAreAccessibleViaApi(events: any[]) {
+    // chromium might send a pageleave event
+    expect(events.length).toBeGreaterThanOrEqual(3)
+    expect(events.filter(({ event }) => event === 'custom-event').length).toEqual(1)
+    expect(events.filter(({ event }) => event === '$pageview').length).toEqual(1)
+    expect(events.filter(({ event }) => event === '$autocapture').length).toEqual(1)
+    expect(events.filter(({ event }) => event === '$pageleave').length).toBeGreaterThanOrEqual(0)
```
daibhin: This feels kind of irrelevant, it's not going to be less than 0

daibhin (approved):
Looks good! Is there any old test files to remove alongside the runner? Happy if that's done as a follow up PR.

.github/workflows/integration.yml
```diff
@@ -0,0 +1,56 @@
+name: Playwright Integration Tests
```
daibhin, line 1: How do we make sure this is a required check? Let me know if you don't have the correct GitHub permissions to do that

hpouillot: I'm not sure how to do that, as I don't have the permissions but we should be able to replace required TestsCafe with those ones in the repo settings

packages/browser/playwright/integration/checks.ts
hpouillot: Yes, I'll do a cleanup as a follow-up PR to remove unused scripts and assertions like this one

.github/workflows/integration.yml
daibhin: Not sure I know how either. Let's get this in once you're happy with it and I can figure out how to make it required then

packages/browser/playwright.config.ts
pauldambra, line 1: commenting randomly so we can a thread

the IE11 tests periodically capture genuine bugs where we break the ES5 build

is that 0.012% that @robbie-c the IE11 usage of the SDK?

it is a shame to lose those, it just means the Es5 build will at some point break and we won't know. we're cutting those customers loose

pauldambra:
is it errors like this pushing us to move (for posterity)

https://github.com/PostHog/posthog-js/actions/runs/16261855863/job/45908666971?pr=2042

.github/workflows/integration.yml
pauldambra, line 1: is it errors like this pushing us to move (for posterity)

https://github.com/PostHog/posthog-js/actions/runs/16261855863/job/45908666971?pr=2042

packages/browser/playwright.config.ts
hpouillot: ES5 build is tested on chromium but it's true that we won't catch bugs that are IE11 specific. 0.012% is [UC browser](https://www.ucweb.com/), I'll check the volume for IE11

hpouillot: Internet Explorer is 0.0017%

pauldambra: most of that traffic looks fake to me too

but for folk using ES5 this is us effectively removing support - the fact that they're a small proportion for us doesn't help them if it's all of their analytics

i think we should do the removal of ES5/IE11 testing in a separate PR so that people can find it if they come looking

pauldambra: wait... if it was safari causing us a problem
why not just keep testcafe for ES5/IE11 testing?

hpouillot: Interesting, the [maximum proportion](https://metabase.prod-us.posthog.dev/question#eyJkYXRhc2V0X3F1ZXJ5Ijp7InR5cGUiOiJuYXRpdmUiLCJkYXRhYmFzZSI6NDIsIm5hdGl2ZSI6eyJxdWVyeSI6InNlbGVjdFxudCwgYiwgYywgcGNcbmZyb20gKHNlbGVjdFxuICAgIHRlYW1faWQgYXMgdCxcbiAgICBtYXRfJGJyb3dzZXIgYXMgYixcbiAgICBjb3VudCgpIGFzIGMsXG4gICAgYyAqIDEwMC4wIC8gc3VtKGMpIG92ZXIocGFydGl0aW9uIGJ5IHRlYW1faWQpIEFTIHBjXG5mcm9tIGV2ZW50cyBcbndoZXJlIHRpbWVzdGFtcCA-IG5vdygpIC0gaW50ZXJ2YWwgNyBkYXlzIGFuZCBiIGlzIG5vdCBudWxsIGFuZCBiIT0gJydcbmdyb3VwIGJ5IHQsIGJcbm9yZGVyIGJ5IHQsIGMgZGVzYylcbndoZXJlIGIgPT0gJ0ludGVybmV0IEV4cGxvcmVyJ1xub3JkZXIgYnkgcGMgZGVzYyIsInRlbXBsYXRlLXRhZ3MiOnt9fX0sImRpc3BsYXkiOiJ0YWJsZSIsInBhcmFtZXRlcnMiOltdLCJ2aXN1YWxpemF0aW9uX3NldHRpbmdzIjp7InRhYmxlLmNvbHVtbl93aWR0aHMiOls0OTgsMjA2LDE1MSwxNjVdfX0=) of IE for our users is 49%...

Maybe let's keep TestCafe for IE11 😅

packages/browser/playwright/mocked/session-recording/session-recording-masking.spec.ts
hpouillot: It's mainly stylistic, especially when you start having deeply nested tests

packages/browser/playwright.config.ts
pauldambra: discussions like this is why we've not removed ES5/IE11 support in the past
if you need that, you _really_ need that

at some point it becomes too expensive to provide, but for now it's only updating the polyfill in the ES5 bundle periodically so where we can keep it I lean towards we should keep it

PR #2099, opened 2025-07-22
feat(flags): Allow capturing feature enrollment stage in update event

haacked:
When calling `updateEarlyAccessFeatureEnrollment`, allow callers to specify the stage of the enrollment.

See https://posthog.slack.com/archives/C02E3BKC78F/p1753190373231089

## Changes

Adds an optional `stage` argument to the `updateEarlyAccessFeatureEnrollment` method.

## Checklist

### DIFF
--- a/packages/browser/src/posthog-featureflags.ts
+++ b/packages/browser/src/posthog-featureflags.ts
@@ -914,7 +914,7 @@ export class PostHogFeatureFlags {
         }
 
         if (stage) {
-            properties['$feature_enrollment_state'] = stage
+            properties['$feature_enrollment_stage'] = stage
         }
 
         this._instance.capture('$feature_enrollment_update', properties)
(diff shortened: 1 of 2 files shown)


packages/browser/src/posthog-featureflags.ts
```diff
@@ -913,6 +913,10 @@ export class PostHogFeatureFlags {
             properties['$early_access_feature_name'] = feature.name
         }
 
+        if (stage) {
+            properties['$feature_enrollment_state'] = stage
```
dmarticus: state, or stage?  Feels weird to have the variable be one but the property be anotherl

haacked: Yeah, we should name it `stage` because that's what's in the UI.

haacked: I'll make that change before merging.

PR #2100, opened 2025-07-22
chore: Safer iteration of experimental `__add_tracing_headers`

Twixes:
## Changes

Turns out it's not exactly safe to monkeypatch every single request made by the frontend. Enabling `__add_tracing_headers` in https://github.com/PostHog/posthog/pull/35384 broke Zendesk ticket submissions.

Updated take on `__add_tracing_headers`: it's now an _allowlist_ of hostnames.

### DIFF
--- a/packages/browser/src/entrypoints/tracing-headers.ts
+++ b/packages/browser/src/entrypoints/tracing-headers.ts
@@ -2,7 +2,20 @@ import { SessionIdManager } from '../sessionid'
 import { patch } from '../extensions/replay/rrweb-plugins/patch'
 import { assignableWindow, window } from '../utils/globals'
 
-const addTracingHeaders = (distinctId: string, sessionManager: SessionIdManager | undefined, req: Request) => {
+const addTracingHeaders = (hostnames: string[], distinctId: string, sessionManager: SessionIdManager | undefined, req: Request) => {
+    let reqHostname: string
+    try {
+        reqHostname = new URL(req.url).hostname
+    } catch (e) {
+        // If the URL is invalid, we skip adding tracing headers
+        return
+    }
+    if (Array.isArray(hostnames) && !hostnames.includes(reqHostname)) {
+        // Skip if the hostname is not in the list (also skip if hostnames is not an array,
+        // because in the earliest version of this __add_tracing_headers was a bool)
+        return
+    }
+
     if (sessionManager) {
         const { sessionId, windowId } = sessionManager.checkAndGetSessionAndWindowId(true)
         req.headers.set('X-POSTHOG-SESSION-ID', sessionId)
@@ -11,7 +24,7 @@ const addTracingHeaders = (distinctId: string, sessionManager: SessionIdManager 
     req.headers.set('X-POSTHOG-DISTINCT-ID', distinctId)
 }
 
-const patchFetch = (distinctId: string, sessionManager?: SessionIdManager): (() => void) => {
+const patchFetch = (hostnames: string[], distinctId: string, sessionManager?: SessionIdManager): (() => void) => {
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
     return patch(window, 'fetch', (originalFetch: typeof fetch) => {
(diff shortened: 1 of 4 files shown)


packages/browser/src/types.ts
```diff
@@ -968,9 +968,10 @@ export interface PostHogConfig {
 
     /**
      * PREVIEW - MAY CHANGE WITHOUT WARNING - DO NOT USE IN PRODUCTION
-     * Whether to wrap fetch and add tracing headers to the request
+     * A list of hostnames for which to inject PostHog tracing headers to all requests
+     * (X-POSTHOG-DISTINCT-ID, X-POSTHOG-SESSION-ID, X-POSTHOG-WINDOW-ID)
      * */
-    __add_tracing_headers?: boolean
+    __add_tracing_headers?: string[]
```
pauldambra, line 974: we could be nicer and accept `boolean | string[]` 
although we do warn it might change so 🙈

packages/browser/src/entrypoints/tracing-headers.ts
```diff
@@ -2,7 +2,20 @@ import { SessionIdManager } from '../sessionid'
 import { patch } from '../extensions/replay/rrweb-plugins/patch'
 import { assignableWindow, window } from '../utils/globals'
 
-const addTracingHeaders = (distinctId: string, sessionManager: SessionIdManager | undefined, req: Request) => {
+const addTracingHeaders = (hostnames: string[], distinctId: string, sessionManager: SessionIdManager | undefined, req: Request) => {
+    let reqHostname: string
+    try {
+        reqHostname = new URL(req.url).hostname
```
pauldambra, line 16: ok, we don't include this in the ES5 bundle so `new URL` is safe here

packages/browser/src/entrypoints/tracing-headers.ts
```diff
@@ -2,7 +2,20 @@ import { SessionIdManager } from '../sessionid'
...
+    } catch (e) {
+        // If the URL is invalid, we skip adding tracing headers
+        return
+    }
+    if (Array.isArray(hostnames) && !hostnames.includes(reqHostname)) {
```
pauldambra: in a perfect world we'd support `*.posthog.com` but that can follow

packages/browser/src/utils/globals.ts
```diff
@@ -180,8 +180,8 @@ interface PostHogExtensions {
         onINP: (metric: any) => void
     }
     tracingHeadersPatchFns?: {
-        _patchFetch: (distinctId: string, sessionManager?: SessionIdManager) => () => void
-        _patchXHR: (distinctId: string, sessionManager?: SessionIdManager) => () => void
+        _patchFetch: (hostnames: string[], distinctId: string, sessionManager?: SessionIdManager) => () => void
+        _patchXHR: (hostnames: string[], distinctId: string, sessionManager?: SessionIdManager) => () => void
```
pauldambra, line 184: for the future traveller thinking of copying - this change will break clients cos we're changing the API between the core bundle and the lazy loaded extension

that's ok here because this is a preview feature and folk really shouldn't have it on in production 🙈

Twixes: I _think_ I engineered this in a way that won't break clients, unless I missed something? The type checks will break, but if someone still has `__add_tracing_headers: true`, that should work the same.
(And I don't see why someone would be hooking so deep to use `_patchFetch `/`_patchXHR` manually)

pauldambra (approved):
👍

packages/browser/playwright/mocked/tracing-headers.spec.ts
```diff
@@ -0,0 +1,247 @@
...
+        const request = fetchRequests[0]
+        const headers = request.headers()
+
+        expect(headers['x-posthog-distinct-id']).toBeTruthy()
+        expect(headers['x-posthog-session-id']).toBeTruthy()
+        expect(headers['x-posthog-window-id']).toBeTruthy()
+    })
+
+    // XHR test fails... needs more testing
+    test.skip('adds PostHog tracing headers to XMLHttpRequest', async ({ page, context }) => {
```
pauldambra, line 55: i'm surprised this fails but I never actively tested this with XHR so 🙈

packages/browser/src/utils/globals.ts
pauldambra: it's because we've changed the signature
so an NPM-fixed posthog core gets the latest lazy loaded file
it doesn't know about the signature change then calls this with `(distinctId, sessionManager)`
but the lazyLoaded files recieves that as `(hostnames: string[], distinctId: string, sessionManager?: SessionIdManager)`

so probably just 

`(distinctId: string, sessionManager?: SessionIdManager, hostnames: string[]) => () => void`

is enough

but it's a preview feature, we're fine to break it. 
only adding a note so if someone comes across the change and copies it for something like dead clicks or surveys 🙈

Twixes: Ah so the lazy-loaded stuff always get loaded from the PostHog CDN? (I thought an npm-fixed posthog-js has all the components npm-fixed)

pauldambra: no, always the latest
you can npm install and then use the `full` or `no-external` bundles

so we have things like https://github.com/PostHog/posthog-js/blob/13ec7669fd750a811faeeddae06b1b5aabdf08f3/packages/browser/src/entrypoints/recorder.ts#L10-L17

PR #2107, opened 2025-07-23
chore(ci): check if posthog-js is affected

hpouillot:
## Changes

- run integration, testcafe and playwright tests only when posthog-js is affected


hpouillot:
I'm skipping steps inside jobs so it should be fine

daibhin (approved):

PR #2111, opened 2025-07-23
fix: mark as idle only when definitely idle

pauldambra:
we set a timer that proactively marks a session as idle in posthog-js

but... if the browser window/tab returns from idle while another window/tab has been active, we don't check for activity and just reset the session

so, instead, when the timer fires we check if there has been any activity before resetting the session

see https://posthoghelp.zendesk.com/agent/tickets/33714 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/40055)


marandaneto (approved):
low context approval :)

pauldambra:
@hpouillot are we in a position to release changes? 🤞

hpouillot:
Release workflow seems to work now, feel free to merge when you want

PR #2117, opened 2025-07-24
chore(ci): optimize ci runs

hpouillot:
## Changes

- Split write mangled property names job
- Build lib only once for bundle size checks

### DIFF
--- a/.github/workflows/bundled-size.yaml
+++ b/.github/workflows/bundled-size.yaml
@@ -11,10 +11,6 @@ jobs:
 
       - uses: ./.github/actions/setup
 
-      - name: Build library
-        run: pnpm build-rollup
-        working-directory: packages/browser
-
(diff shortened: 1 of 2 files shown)


.github/workflows/bundled-size.yaml
```diff
@@ -11,10 +11,6 @@ jobs:
 
       - uses: ./.github/actions/setup
 
-      - name: Build library
-        run: pnpm build-rollup
```
robbie-c, line 15: I think `Upload bundle size visualization` needs this to run, so that the file exists

hpouillot: Yes, all packages are built during the setup step, just before

robbie-c: Ah cool, didn't realise this built stuff now! All good!

PR #2122, opened 2025-07-24
fix(ci): use workflow dispatch

hpouillot:
## Problem

- When checking out another repo, composite actions fail, see [this error](https://github.com/PostHog/posthog-js/actions/runs/16500830722/job/46658652524)

## Changes

- use independent workflow dispatch to checkout other repository


.github/workflows/posthog-upgrade.yml
```diff
@@ -0,0 +1,110 @@
...
+        with:
+          repository: "PostHog/posthog"
+          token: ${{ secrets.POSTHOG_BOT_GITHUB_TOKEN }}
+
+      - uses: pnpm/action-setup@v4
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18
```
pauldambra, line 38: should we use node 20?

pauldambra: or just lts

pauldambra (approved):
blind stamp :)

.github/workflows/posthog-upgrade.yml
hpouillot: posthog is not compatible with node 20. the node version for this repo is only used when developing, so it should be fine.
-> https://github.com/PostHog/posthog/blob/master/.nvmrc

PR #2123, opened 2025-07-24
fix(nextjs-config): update cli version

hpouillot:
## Changes

- update @posthog/cli to 0.3.6 to integrate vulnerability fix


packages/nextjs-config/CHANGELOG.md
```diff
@@ -1,11 +1,13 @@
-# 1.0.2
+# @posthog/nextjs-config
+
+## 1.0.2
```
daibhin, line 3: Why has it done a patch bump if the changeset is minor?

hpouillot: the changeset is for nextjs-config, it's the version that will be released (1.1.0) and the patch bump is for @posthog/cli (manual upgrade on posthog). It's a bit confusing, I agree 😅

daibhin: Right, but I don't see 1.1.0 mentioned anywhere. Does that come after merging?

daibhin (approved):
If you're happy then feels fine to ship. Might need the `release` label

packages/nextjs-config/CHANGELOG.md
hpouillot: Yes, when a PR is closed (with a label release), versions are bumped according to the changesets

PR #2129, opened 2025-07-25
chore/remove legacy migration

marandaneto:
- **Remove legacy migration code, left over from the V1 to V2 migration**
- **fix**
- **fix**

## Changes

...

## Checklist


packages/react-native/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Next
 
+## Removed
```
hpouillot: there is no need to update the changelog, it will be updated automatically when the PR is closed

RELEASING.md
```diff
@@ -0,0 +1,12 @@
+## Releases
```
hpouillot, line 1: Yes it will be clearer here

packages/react-native/CHANGELOG.md
marandaneto: ok, removing

hpouillot (approved):

PR #2131, opened 2025-07-28
fix(flags): fix early access features accumulation in persistence

dmarticus:
### Problem
The `getEarlyAccessFeatures()` SDK method was returning more features than the API (20 vs 7), indicating that features removed from the API were not being removed from local persistence.

### Root Cause
The SDK stores early access features in persistence to avoid repeated API calls. However, when features are removed from the API response, they weren't being cleared from persistence, causing them to accumulate over time.

### Solution
Added an explicit `unregister()` call before storing new features to ensure persistence always reflects the current API state:

```typescript
// Clear existing features before registering new ones
this._instance.persistence?.unregister(PERSISTENCE_EARLY_ACCESS_FEATURES)
this._instance.persistence?.register({ [PERSISTENCE_EARLY_ACCESS_FEATURES]: earlyAccessFeatures })
```

###  Testing

- Features in persistence are completely replaced with API response
- Old features are removed when not in the API response
- The fix handles missing persistence gracefully

### DIFF
--- a/packages/browser/src/__tests__/featureflags.test.ts
+++ b/packages/browser/src/__tests__/featureflags.test.ts
@@ -951,7 +951,7 @@ describe('featureflags', () => {
             // Set up initial features in persistence
             instance.persistence.props.$early_access_features = [
                 EARLY_ACCESS_FEATURE_FIRST,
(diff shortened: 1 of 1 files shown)


haacked (approved):
LGTM!

PR #2170, opened 2025-07-29
chore: do not mangle _surveyManager property

lucasheriques:
## Problem

We created this method, `_renderExternalSurvey`, to [handle rendering external surveys](https://github.com/PostHog/posthog/pull/33948).

However, this method currently requires an extra network call to fetch the survey, and then render it.

I figured out a better way to do that by getting the survey payload in the backend, when just calling `_surveyManager` directly.

## Changes

Removes the `_renderExternalSurvey`, and adds the `_surveyManager` as a list of property to do not be mangled (so they don't get minimized).

[See the main repo PR here to understand how we're using it now.](https://github.com/PostHog/posthog/pull/35811)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes!) -> only changing private methods


marandaneto (approved):

packages/browser/rollup.config.mjs
```diff
@@ -143,7 +143,7 @@ const plugins = (es5, noExternal) => [
                               '_instance',
                               '_surveyEventReceiver',
                               // this one is needed - we use it to render external surveys. but we want to keep it private for our usage only
-                              '_renderExternalSurvey',
+                              '_surveyManager',
```
marandaneto, line 146: i think we should fix the comment above as well

lucasheriques: the reasoning is the sort of the same, but I updated it to be more clear 👍

marandaneto: yeah i know, this part 'but we want to keep it private for our usage only' was outdated after the change, since its an instance and not a method anymore

PR #2172, opened 2025-07-30
chore: add new enable surveys config

lucasheriques:
## Problem

right now, loading surveys functionality always requires a `/flags` call to determine if there are any surveys to load.

this is is expected behavior up until now. Since we now have external surveys, we want to be able to enable surveys, even without them.

## Changes

adds new `advanced_enable_surveys` to PostHog config, which we'll use in the external surveys.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-surveys.ts
```diff
@@ -79,7 +79,7 @@ export class PostHogSurveys {
             return
         }
 
-        const isSurveysEnabled = this._isSurveysEnabled
+        const isSurveysEnabled = this._isSurveysEnabled || this._instance.config.advanced_enable_surveys
```
marandaneto: There's an edge case here, if the external scripts fail to load, eg, ad blockers, we call `_handleSurveyLoadError` -> `_notifySurveyCallbacks` and will end up hitting `/api/surveys/`, at least for external surveys, we don't want that.

lucasheriques: `/api/surveys/` endpoint is only hit when this.getSurveys(callback) is called, which only happens when context.isLoaded is true. 

When external scripts fail to load, [isLoaded is explicitly set to false](https://github.com/PostHog/posthog-js/blob/chore/add-new-enable-surveys-config/packages/browser/src/posthog-surveys.ts#L139), so the API call is never made.

[The logic is working correctly](https://github.com/PostHog/posthog-js/blob/chore/add-new-enable-surveys-config/packages/browser/src/posthog-surveys.ts#L247-L251):
- When scripts load successfully: isLoaded: true → calls this.getSurveys() → may hit API if surveys are not loaded yet
- When scripts fail to load: isLoaded: false → calls callback([], context) directly → no API call

makes sense?

marandaneto: if i understand correctly, we'd call https://github.com/PostHog/posthog-js/blob/a8f01d8fbe98b76c367ae302da123c76545f97d5/packages/browser/src/posthog-surveys.ts#L250 which then would call surveys API if `isLoaded` is false.

lucasheriques: this.getSurveys is only called if `context.isLoaded == true`, which in this case it's not

https://github.com/PostHog/posthog-js/blob/a8f01d8fbe98b76c367ae302da123c76545f97d5/packages/browser/src/posthog-surveys.ts#L137-L139

https://github.com/PostHog/posthog-js/blob/a8f01d8fbe98b76c367ae302da123c76545f97d5/packages/browser/src/posthog-surveys.ts#L244-L251

marandaneto: i read this `if` the other way around, my bad, makes sense

marandaneto (approved):

packages/browser/src/posthog-surveys.ts
lucasheriques: i changed it slightly, think it reads better this way:

https://github.com/PostHog/posthog-js/blob/7aa57da9d3037e97dc3feadb0da468c8012a205e/packages/browser/src/posthog-surveys.ts#L244-L250

PR #2173, opened 2025-07-30
fix(llmo): tools not sent to llmo for langchain

Radu-Raicea:
## Problem

Tool calls from LangChain were not being sent to LLM Observability.

## Changes

Fixed the bug that prevented us from capturing the tool calls.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/loud-taxis-appear.md
+++ b/.changeset/loud-taxis-appear.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Fix bug where tool calls were not sent in LangChain


.changeset/loud-taxis-appear.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/ai': patch
```
k11kirky, line 2: what is this file?

Radu-Raicea: It's the release process for posthog-js: https://github.com/PostHog/posthog-js/blob/6e35a639a4d06804f6844cbde15adf11a069b92b/RELEASING.md

We don't need to manually bump the version and add to the changelog anymore.

hpouillot: It's part of the new release workflow. I introduced a tool called changeset that will bump the version and update changelogs when the PR is merged (if there is a 'release' label on the PR).

PR #2176, opened 2025-07-31
fix(nextjs-config): support esm next configs

redabacha:
## Problem

Currently the @posthog/nextjs-config package does not work properly with a ESM next config (via `next.config.mjs` rather than `next.config.js`) because it is checking for the `__dirname` global which only exists in a CJS context.

## Changes

- Added a check for `__dirname` and use alternative method if unavailable
- Added tests for the above which also involved setting up some missing boilerplate
- Moved `callPosthogCli` to `utils.ts`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/sixty-pandas-unite.md
+++ b/.changeset/sixty-pandas-unite.md
@@ -1,0 +1,5 @@
+---
+'@posthog/nextjs-config': patch
+---
+
+Fix \_\_dirname is not defined error with ESM next configs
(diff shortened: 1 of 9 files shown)


hpouillot:
Hey, thanks for your PR ! I think it's better to not deal with shims ourself, I created another PR to use rslib instead of tsup to avoid doing that : https://github.com/PostHog/posthog-js/pull/2182

PR #2185, opened 2025-08-02
feat(cookieless): Split cookieless into 2 modes, 'always' and 'on_reject'

robbie-c:
## Problem

Cookieless mode took a bit of effort to use, this makes it much simpler.

## Changes

Split cookieless into 2 modes. Copied from a comment:

```tsx
    /**
     * Enables cookieless mode. In this mode, PostHog will not set any cookies, or use session or local storage. User
     * identity is handled by generating a privacy-preserving hash on PostHog's servers.
     * - 'always' - enable cookieless mode immediately on startup, use this if you do not intend to show a cookie banner
     * - 'on_reject' - enable cookieless mode only if the user rejects cookies, use this if you want to show a cookie banner. If the user accepts cookies, cookieless mode will not be used, and PostHog will use cookies and local storage as usual.
     *
     * Note that you MUST enable cookieless mode in your PostHog project's settings, otherwise all your cookieless events will be ignored. We plan to remove this requirement in the future.
     * */
```

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/playground/nextjs/src/posthog.ts
+++ b/packages/browser/playground/nextjs/src/posthog.ts
@@ -40,7 +40,6 @@ export const configForConsent = (): Partial<PostHogConfig> => {
 }
 
 export const updatePostHogConsent = (consentGiven: boolean) => {
-    console.log(`updatePostHogConsent(${consentGiven})`)
     if (consentGiven) {
         posthog.opt_in_capturing()
     } else {
--- a/packages/browser/playground/nextjs/src/posthog.ts
+++ b/packages/browser/playground/nextjs/src/posthog.ts
@@ -40,6 +40,7 @@ export const configForConsent = (): Partial<PostHogConfig> => {
 }
 
 export const updatePostHogConsent = (consentGiven: boolean) => {
+    console.log(`updatePostHogConsent(${consentGiven})`)
     if (consentGiven) {
         posthog.opt_in_capturing()
     } else {
@@ -65,14 +66,12 @@ if (typeof window !== 'undefined') {
         person_profiles: PERSON_PROCESSING_MODE === 'never' ? 'identified_only' : PERSON_PROCESSING_MODE,
         persistence_name: `${process.env.NEXT_PUBLIC_POSTHOG_KEY}_nextjs`,
         opt_in_site_apps: true,
-        opt_out_capturing_by_default: true,
-        opt_out_persistence_by_default: true,
         integrations: {
             intercom: true,
             crispChat: true,
         },
         __preview_remote_config: true,
-        __preview_experimental_cookieless_mode: false,
+        cookieless_mode: 'on_reject',
         __preview_flags_v2: true,
         ...configForConsent(),
     })
--- a/packages/browser/src/__tests__/cookieless.test.ts
+++ b/packages/browser/src/__tests__/cookieless.test.ts
@@ -1,89 +1,213 @@
 import type { PostHogConfig } from '../types'
 import { uuidv7 } from '../uuidv7'
-import { defaultPostHog } from './helpers/posthog-instance'
+import { createPosthogInstance } from './helpers/posthog-instance'
+const uuidV7Pattern = /^[0-9a-f]{8}-[0-9a-f]{4}-7[0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}$/i
+
+jest.mock('../utils/globals', () => {
+    const orig = jest.requireActual('../utils/globals')
+    const mockURLGetter = jest.fn()
+    const mockReferrerGetter = jest.fn()
+    const mockedCookieBox = { cookie: '' }
+    return {
+        ...orig,
+        mockURLGetter,
+        mockReferrerGetter,
+        mockedCookieBox: mockedCookieBox,
+        document: {
+            ...orig.document,
+            createElement: (...args: any[]) => orig.document.createElement(...args),
+            body: {},
+            get referrer() {
+                return mockReferrerGetter()
+            },
+            get URL() {
+                return mockURLGetter()
+            },
+            get cookie() {
+                return mockedCookieBox.cookie
+            },
+            set cookie(value: string) {
+                mockedCookieBox.cookie = value
+            },
+            addEventListener(_, event, callback) {
+                if (event === 'DOMContentLoaded') {
+                    callback()
+                }
+            },
+            readyState: 'complete',
+            visibilityState: 'visible',
+        },
+        get location() {
+            const url = mockURLGetter()
+            return {
+                href: url,
+                toString: () => url,
+            }
+        },
+    }
+})
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { mockURLGetter, mockedCookieBox, document } = require('../utils/globals')
 
 describe('cookieless', () => {
     const eventName = 'custom_event'
     const eventProperties = {
         event: 'prop',
     }
-    const identifiedDistinctId = 'user-1'
-    const setup = (config: Partial<PostHogConfig> = {}, token: string = uuidv7()) => {
+    const setup = async (config: Partial<PostHogConfig> = {}, token: string = uuidv7()) => {
         const beforeSendMock = jest.fn().mockImplementation((e) => e)
-        const posthog = defaultPostHog().init(token, { ...config, before_send: beforeSendMock }, token)!
+        const posthog = await createPosthogInstance(token, {
+            ...config,
+            before_send: beforeSendMock,
+        })!
         posthog.debug()
         return { posthog, beforeSendMock }
     }
 
-    it('should send events with the sentinel distinct id', () => {
-        const { posthog, beforeSendMock } = setup({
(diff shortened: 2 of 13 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -2463,6 +2495,14 @@ export class PostHog {
         return this.consent.isOptedOut()
     }
 
+    is_capturing(): boolean {
+        if (this.config.cookieless_mode === 'on_reject') {
+            return this.consent.isExplicitlyOptedOut() || this.consent.isOptedIn()
+        } else {
+            return !this.has_opted_out_capturing()
+        }
+    }
```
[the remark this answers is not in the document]
robbie-c: Added a comment explaining this

robbie-c:
@greptile-apps again

packages/browser/playground/nextjs/src/posthog.ts
```diff
@@ -40,6 +40,7 @@ export const configForConsent = (): Partial<PostHogConfig> => {
 }
 
 export const updatePostHogConsent = (consentGiven: boolean) => {
+    console.log(`updatePostHogConsent(${consentGiven})`)
```
jabahamondes: Is this console log expected?

packages/browser/playground/nextjs/src/posthog.ts
```diff
@@ -65,14 +66,12 @@ if (typeof window !== 'undefined') {
...
         opt_in_site_apps: true,
-        opt_out_capturing_by_default: true,
-        opt_out_persistence_by_default: true,
         integrations: {
             intercom: true,
             crispChat: true,
         },
         __preview_remote_config: true,
-        __preview_experimental_cookieless_mode: false,
+        cookieless_mode: 'on_reject',
```
jabahamondes, line 73: Nit: Can we add these mode into an enum so we know it’s not just a random string?

packages/browser/src/__tests__/cookieless.test.ts
```diff
@@ -1,89 +1,213 @@
...
+            return {
+                href: url,
+                toString: () => url,
+            }
+        },
+    }
+})
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+const { mockURLGetter, mockedCookieBox, document } = require('../utils/globals')
```
jabahamondes, line 51: Is there a reason we can’t use import here?

jabahamondes (approved):
LGTM! Left small comments but no blockers

packages/browser/playground/nextjs/src/posthog.ts
robbie-c: oops, no, I guess we don't cover this in linting either

packages/browser/playground/nextjs/src/posthog.ts
rafaeelaudibert: Problem with enum is that those using TS will *need* to use the enum or else it'll fail which means setup for TS vs JS will be different, I'm not really a fan of that.

We've used strings for all other configs, I agree with the behavior Robbie has here

packages/browser/src/consent.ts
```diff
@@ -34,16 +34,28 @@ export class ConsentManager {
     }
 
     public isOptedOut() {
+        if (this._config.cookieless_mode === 'always') {
+            return false
+        }
```
rafaeelaudibert: Hmm, what's the argument behind saying that someone is NOT opted out when cookieles mode is `always`? I could argue both ways

packages/browser/src/posthog-core.ts
```diff
@@ -1932,7 +1940,7 @@ export class PostHog {
         this.persistence?.set_property(USER_STATE, 'anonymous')
         this.sessionManager?.resetSessionId()
         this._cachedPersonProperties = null
-        if (this.config.__preview_experimental_cookieless_mode) {
+        if (this.config.cookieless_mode === 'always') {
```
rafaeelaudibert, line 1944: Why not check for reject+explicitly opted out here?

packages/browser/src/posthog-core.ts
```diff
@@ -2388,6 +2400,11 @@ export class PostHog {
         captureEventName?: EventName | null | false /** event name to be used for capturing the opt-in action */
         captureProperties?: Properties /** set of properties to be captured along with the opt-in action */
     }): void {
+        if (this.config.cookieless_mode === 'always') {
+            logger.warn('Consent opt in/out is not valid with cookieless_mode="always" and will be ignored')
+            return
+        }
```
rafaeelaudibert, line 2407: I wonder if we should still keep track of this even though we wont take it into consideration? This way if someone eventually decides to move to `on_reject` we'll have properly captured consent in case it happened in the past

packages/browser/src/posthog-core.ts
```diff
@@ -2417,8 +2434,23 @@ export class PostHog {
      * @public
      */
     opt_out_capturing(): void {
```
rafaeelaudibert, line 2448: Where are we doing the opposite of this - removing cookieless mode when someone who was opted out once and then opts in?

packages/browser/src/posthog-surveys.ts
```diff
@@ -72,6 +72,9 @@ export class PostHogSurveys {
             logger.info('Disabled. Not loading surveys.')
             return
         }
+        if (this._instance.config.cookieless_mode) {
+            return
+        }
```
rafaeelaudibert: Can we log something here to make sure we understand why surveys is not loading? I've seen my fair share of surveys problems and these logs have always helped me understand why they aren't working.

rafaeelaudibert: Also, why is this the case? If someone's set it to `on_reject` but the customer consented to tracking, then this isnt a problem.

packages/browser/src/posthog-core.ts
robbie-c: Hmm, I don't really want to support users moving from one mode to another, it can potentially do legally-dodgy things if you try to associate regular events with cookieless events. That's why the 2 modes are:
* always send cookieless events
* don't send any events until you know what the consent status is, then send either only cookieless or only regular events for that user

packages/browser/src/posthog-core.ts
robbie-c: You're right, at this point we should call reset() and *then* opt them in. I'll add this as a follow-up, as the first customer who wants to use this is going to use "always" which is unaffected.

packages/browser/src/__tests__/cookieless.test.ts
robbie-c: I copied this style from another test file (which I wrote) - I don't remember the details, but the mockX imports don't exist on the original file so I would guess that my IDE would complain

packages/browser/src/consent.ts
robbie-c: Hmm I wrote this before I added isCapturing, I could probably handle it there instead

packages/browser/src/posthog-core.ts
robbie-c: There's tests for this now

packages/browser/src/posthog-core.ts
robbie-c: Ah they can't have explicit opt out here because we reset the consent state a few lines earlier

lricoy (approved):
lgtm! 

Special mention to +1 on @rafaeelaudibert's surveys question and adding a log there to explain why they're not loading

packages/browser/src/consent.ts
robbie-c: (I did this)

packages/browser/src/posthog-surveys.ts
robbie-c: Added a log here.

I plan to fix more posthog features as we go - I'll do surveys first.

Error tracking will probably be second ^^

hpouillot:
I just saw a "bump minor", so I thought I'll pop in :) Release workflow changed, you'll need to add a changeset and a release tag on the PR to release a new version => https://github.com/PostHog/posthog-js/blob/main/RELEASING.md

robbie-c:
@ioannisj there's always been the option to set persistence to "memory", in which case the library doesn't write to cookie / local / session storage, which was how we used to recommend people to do cookieless.

In general, that mode wasn't very useful (a page reload creates a new user identity), and the new version of cookieless works in a different way (there's some big comments in the cookieless files e.g. https://github.com/PostHog/posthog/blob/c63bcb7a5763ec8ef05a0b59c922469b57e649a0/plugin-server/src/ingestion/cookieless/cookieless-manager.ts#L24 explaining it)

Some products won't work if the user has not allowed us to use storage or process personal data, so for now those products are just going to be disabled in cookieless mode, and I'll be reaching out to those teams to figure out if it makes sense technically and legally to do something else.

PR #2186, opened 2025-08-03
Add turbopack support to nextjs-config package

jrhe:
## Problem

Turbopack is the successor to webpack, and `@posthog/nextjs-config` does not yet have support for it. 

## Changes

This PR adds turbopack support to the nextjs-config package. It uses the relatively new `runAfterProductionCompile` nextjs hook,  so we should ensure the correct version of nextjs. Does posthog-js have a standardised approach?

It also adds a `failOnError` parameter to fail a build if source maps are not uploaded.

The code is partially AI generated using Claude Code, but has been reviewed both manually and by CodeRabbit.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [X] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### DIFF
--- a/packages/nextjs-config/src/turbopack-handler.ts
+++ b/packages/nextjs-config/src/turbopack-handler.ts
@@ -1,0 +1,35 @@
+import { PostHogNextConfigComplete } from './config'
+import * as path from 'path'
+import * as fs from 'fs'
+import { processSourcemaps } from './sourcemap-processor'
+
+/**
+ * Process sourcemaps for Turbopack builds
+ * This is called by the runAfterProductionCompile hook
+ */
+export async function processTurbopackSourcemaps(
+  posthogOptions: PostHogNextConfigComplete,
+  distDir?: string
+): Promise<void> {
+  const resolvedDistDir = path.resolve(distDir ?? '.next')
+
+  // Process both server and client sourcemaps
+  const serverDir = path.join(resolvedDistDir, 'server')
+  const clientDir = path.join(resolvedDistDir, 'static/chunks')
+
+  // Check if directories exist
+  const serverExists = fs.existsSync(serverDir)
+  const clientExists = fs.existsSync(clientDir)
+
+  if (serverExists) {
+    await processSourcemaps(serverDir, posthogOptions, true)
+  }
+
+  if (clientExists) {
+    await processSourcemaps(clientDir, posthogOptions, false)
+  }
+
+  if (!serverExists && !clientExists && posthogOptions.verbose) {
+    console.log('PostHog: No build directories found, skipping sourcemap processing')
+  }
+}
--- a/packages/nextjs-config/README.md
+++ b/packages/nextjs-config/README.md
@@ -23,6 +23,7 @@ export default withPostHogConfig(nextConfig, {
     project: "my-application", // (optional) Project name, defaults to repository name
     version: "1.0.0", // (optional) Release version, defaults to current git commit
     deleteAfterUpload: true, // (optional) Delete sourcemaps after upload, defaults to true
+    failOnError: false, // (optional) Fail on error, defaults to false
   },
 });
 ```
--- a/packages/nextjs-config/package.json
+++ b/packages/nextjs-config/package.json
@@ -23,7 +23,8 @@
     "build": "rslib build",
     "test:unit": "jest",
     "dev": "rslib build --watch",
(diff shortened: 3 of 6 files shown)


jrhe:
@hpouillot - tagging as I believe you wrote the original code and this is in your area.

packages/nextjs-config/src/turbopack-handler.ts
```diff
@@ -0,0 +1,35 @@
...
+
+/**
+ * Process sourcemaps for Turbopack builds
+ * This is called by the runAfterProductionCompile hook
+ */
+export async function processTurbopackSourcemaps(
+  posthogOptions: PostHogNextConfigComplete,
+  distDir?: string
+): Promise<void> {
+  const resolvedDistDir = path.resolve(distDir ?? '.next')
```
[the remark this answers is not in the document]
jrhe: Updated

packages/nextjs-config/src/utils/version-check.ts
```diff
@@ -0,0 +1,101 @@
...
+    return { parseable: false, supportsHook: false }
+  }
+
+  // Parse semantic version
+  const parsedVersion = semver.parse(cleanedVersion)
+  if (!parsedVersion) {
+    return { parseable: false, supportsHook: false }
+  }
+
+  // runAfterProductionCompile was introduced in Next.js 15.4.0-canary.19
```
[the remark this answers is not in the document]
jrhe: comment not used as a constant.

jrhe:
- Fixed the `runAfterProductionCompile` hook as it should have been on the compiler key
- Refactored to make a bit cleaner

This works now for server source maps. It doesn't work for static sourcemaps though as turbopack for some reason uses different names between the source file and the map file. 

To fix this we can either modify the cli to check the last line of each file for a `//# sourceMappingURL=` declaration, and then read in the relevant file; or see if we can get  a change made to nextjs such that it outputs a mapping or uses the same filenames.

packages/nextjs-config/src/utils/version-check.ts
```diff
@@ -0,0 +1,101 @@
...
+
+// Helper to clean version strings by removing leading prefixes
+export function cleanVersion(version: string): string {
+  return version.replace(/^[^0-9]*/, '').trim()
+}
+
+// Helper to get Next.js version from package.json
+export function getNextJsVersion(): string | null {
+  try {
+    const packageJsonPath = path.join(process.cwd(), 'package.json')
```
hpouillot, line 13: hum, I don't think this way of retrieving nextjs version will work with workspaces and monorepos

.claude/settings.local.json
```diff
@@ -0,0 +1,18 @@
+{
```
hpouillot: 👀

jrhe: oops!

jakeleventhal:
Looking forward to getting this merged!

packages/nextjs-config/src/builders/turbopack.ts
```diff
@@ -0,0 +1,44 @@
...
+      ...resolvedUserConfig.compiler,
+      runAfterProductionCompile: async () => {
+        try {
+          // Call user's hook first if it exists
+          const userCompiler = resolvedUserConfig.compiler as ExtendedNextConfig['compiler']
+          if (userCompiler?.runAfterProductionCompile) {
+            await userCompiler.runAfterProductionCompile()
+          }
+          // Then process sourcemaps
+          await processTurbopackSourcemaps(posthogNextConfigComplete, resolvedUserConfig.distDir)
```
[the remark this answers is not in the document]
jrhe: Next JS has a `distDir` default of nextjs . I'm don't think there's a case where this would be explicitly overriden to be unset.

packages/nextjs-config/src/utils/version-check.ts
jrhe: It works in our pnpm monorepo with the `apps/webapp` workspace having our nextjs app. I think the time this would break is if a user is building a nextjs config outside of their nextjs app workspace, but I'm not sure why someone would do this, and if they did, they'd likely have a dependency on nextjs

jrhe:
Added a couple of replies to comments - is anything blocking this that I can help with?

PR #2188, opened 2025-08-05
fix: send custom event to replay on trigger matched

pauldambra:
it's hard to show a user that a session is waiting to start recording
when in the replayer

it means they are sometimes confused about why a recording has an event in the timeline before the recording starts

so, let's show a custom event in the timeline when a trigger matches to make this easier to see


pauldambra:
nowait... this is not right 🫠

Issue #2135, opened 2025-07-03
Vercel AI SDK v5


chkn:
## Is your feature request related to a problem?

Vercel [has released](https://v5.ai-sdk.dev/docs/announcing-ai-sdk-5-beta) a beta of AI SDK v5, but the `@posthog/ai` package doesn't seem to support it yet.

## Describe the solution you'd like

Prerelease `@posthog/ai` with support for Vercel AI SDK v5.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs-config

k11kirky:
@tsotimus We will look into how to better support the gateway independently from this issue!

PR #2191, opened 2025-08-06
feat: breaking change to v5 vercel sdk

k11kirky:
BREAKING CHANGE TO VERCEL AI SDK

Upgrading to Vercel AI SDK v5 as requested in https://github.com/PostHog/posthog-js/issues/2135 as well as many support tickets.

This is a breaking change, so those wanting to stay on V4 should not bump. (we have major bump for AI package)

those wanting to upgrade can do see migration guide here: https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0
 - no changed needed to posthog implementation.
 
 Thanks to @Yash-Singh1 for https://github.com/PostHog/posthog-js/pull/2189 the base PR


packages/ai/src/vercel/middleware.ts
```diff
@@ -177,17 +180,17 @@ const mapVercelOutput = (result: any): PostHogInput[] => {
       },
     ]
   }
-
-  // Fallback to original behavior for other result types
+  // Fallback to original behavior for other result types TODO: check if we can remove this
```
rafaeelaudibert, line 183: When we doing this TODO?

rafaeelaudibert (approved):
Low context

packages/ai/src/vercel/middleware.ts
k11kirky: 👀 added to my todo list

PR #2197, opened 2025-08-07
fix(flags): Pass project API key in remote_config requests

haacked:
This change ensures that remote config requests include the project API key as a query parameter, enabling deterministic project routing when using personal API keys that have access to multiple projects.

## Problem

See https://github.com/PostHog/posthog/issues/35303 for the problem.

## Changes:

- Update remote config URL to include token parameter with project API key
- Add comprehensive test coverage for the new URL format
- Update changelog to document the fix

Port of PostHog Python implementation: https://github.com/PostHog/posthog-python/pull/303

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/node/CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # posthog-node
 
+## [Unreleased]
+
+1. fix: Always pass project API key in `remote_config` requests for deterministic project routing
```
hpouillot: For info, there is no need to update the changelog, changeset will take care of it automatically

haacked: Ah, thanks! Should I revert this particular change then, or will changeset handle the fact that it's already there?

hpouillot: I think It will append to the changelog and duplicate the note so yes, we should probably revert this change

haacked: Hmm, so the [Pull Request Template](https://github.com/PostHog/posthog-js/blob/main/.github/pull_request_template.md) has this section:

### If releasing new changes

I was just following along and ran `pnpm changeset` because the template told me to. Is the PR template incorrect?

playground/flags/README.md
```diff
@@ -0,0 +1,53 @@
...
+```
+
+### Expected Output
+
+```
+Testing remote config endpoint...
+✅ Success! Remote config payload for 'your-flag-key': { "setting": "value", "config": {...} }
+```
+
+### Implementation Notes
```
andyzzhao: Nit: this blurb feels unnecessary

andyzzhao (approved):

PR #2199, opened 2025-08-07
fix: vercel output mapping

k11kirky:
## Problem

Vercel output mappings were off sending the whole responses

## Changes

fixed mapping

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config

### DIFF
--- a/packages/ai/src/vercel/middleware.ts
+++ b/packages/ai/src/vercel/middleware.ts
@@ -372,7 +372,26 @@ export const createInstrumentationMiddleware = (
 
           flush: async () => {
             const latency = (Date.now() - startTime) / 1000
-            const outputContent = reasoningText ? `${reasoningText}\n\n${generatedText}` : generatedText
+            // Build content array similar to mapVercelOutput structure
+            const content = []
+            if (reasoningText) {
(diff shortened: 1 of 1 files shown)


fuziontech (approved):
send it!

packages/ai/src/vercel/middleware.ts
```diff
@@ -152,25 +153,59 @@ const mapVercelPrompt = (messages: LanguageModelV2Prompt): PostHogInput[] => {
...
-        type: 'function',
-        id: toolCall.toolCallId,
+const mapVercelOutput = (result: LanguageModelV2Content[]): PostHogInput[] => {
+  const content: any[] = result.map((item) => {
+    if (item.type === 'text') {
+      return { type: 'text', text: truncate(item.text) }
+    }
+    if (item.type === 'tool-call') {
+      return {
+        type: 'tool-call',
```
Radu-Raicea, line 163: We made tool calls all have a type "function" so that they can be parsed in the new flow that takes an array in the content property. I'm not sure how this is going to be displayed.

PR #2202, opened 2025-08-07
feat(react-native): Add ability to pass custom headers to fetch calls

myklam:
## Problem

The browser package for PostHog permits the user to pass custom headers to the requests that the package makes. The react-native package did not support this. The main reason I'm adding this here is that my reverse proxy requires an Authorization header to be sent as we are only tracking logged in users.

## Changes

- Adds the `request_headers` to the config
- Spreads this option with other existing headers
- Adds tests for `request_headers`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -68,6 +68,13 @@ export type PostHogOptions = PostHogCoreOptions & {
    * Defaults to false
    */
   enablePersistSessionIdAcrossRestart?: boolean
+
+  /**
+   * A list of headers that should be sent with requests to the PostHog API.
+   *
+   * @default {}
+   */
+  request_headers?: { [header_name: string]: string }
 }
 
 export class PostHog extends PostHogCore {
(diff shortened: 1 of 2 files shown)


packages/react-native/src/posthog-rn.ts
```diff
@@ -68,6 +68,13 @@ export type PostHogOptions = PostHogCoreOptions & {
    * Defaults to false
    */
   enablePersistSessionIdAcrossRestart?: boolean
+
+  /**
+   * A list of headers that should be sent with requests to the PostHog API.
+   *
+   * @default {}
+   */
+  request_headers?: { [header_name: string]: string }
```
marandaneto, line 77: lets add this to `PostHogCoreOptions` so the other packages will benefit from this too

marandaneto: We should add a note that this won't work for react native - session replay since the native SDKs (Android and iOS SDKs) send the recordings, so the headers would need to be set there as well.
So this feature has to be added there first

PR #2213, opened 2025-08-13
chore(err): add before_send to node SDK

oliverb123:


packages/node/src/client.ts
```diff
@@ -791,4 +839,28 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     )
     return await this.captureImmediate(evtMsg)
   }
+
+  private _runBeforeSend(eventMessage: EventMessage): EventMessage | null {
```
hpouillot, line 843: I am wondering if we should not put this logic in @posthog/core to share with react-native and posthog-js-lite as well ?

oliverb123: My main hesitation is just that they all have different types for stuff like `EventMessage` - I could write some generic thing, but it doesn't feel _super_ worth it

hpouillot: Make sense, It could probably part of a later code sharing push

daibhin: Could we use a generic type?

daibhin (approved):

PR #2215, opened 2025-08-13
fix: always set surveySeen on local storage for old posthog-js versions

lucasheriques:
## Problem

[This PR](https://github.com/PostHog/posthog-js/pull/1987) introduced a change: the local storage `surveySeen` key was moved from the survey extensions to `posthog-core`.

However, this caused an issue: any versions prior to `1.249.2` got an issue where the survey was popping up repeatedly for the same user. This happened because we introduced a breaking change that now requires all customers to update their posthog-js version to remove this bug.

This happens because the extensions are always downloaded from the lastest version, so a customer on a version prior to this no longer sets the local storage. But because they might not have updated `posthog-js`, their `posthog-core` will also not set it up, causing the survey to show multiple times.

## Changes

This PR re-adds setting the local storage key to survey-extension-utils, and it still keeps the same functionality in posthog-core.

This way, even older versions will also work as expected without the need to upgrade.

It also adds a helper function to ensure we don't need to set the local storage multiple times if it's already set.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code -> all tests passing

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
+++ b/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
@@ -13,7 +13,12 @@ import {
     SurveyWidgetType,
 } from '../../posthog-surveys-types'
 import { document as _document, window as _window, userAgent } from '../../utils/globals'
-import { SURVEY_LOGGER as logger, SURVEY_IN_PROGRESS_PREFIX, SURVEY_SEEN_PREFIX } from '../../utils/survey-utils'
+import {
+    getSurveySeenKey,
+    SURVEY_LOGGER as logger,
+    setSurveySeenOnLocalStorage,
+    SURVEY_IN_PROGRESS_PREFIX,
+} from '../../utils/survey-utils'
 import { isNullish, isArray } from '../../utils/type-utils'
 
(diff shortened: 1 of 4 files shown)


packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -529,15 +535,6 @@ export const getSurveySeen = (survey: Survey): boolean => {
     return false
 }
 
-export const getSurveySeenKey = (survey: Survey): string => {
```
marandaneto, line 532: i dont think we can safely remove this without breaking users?
people will download the latest extension that now does not have the `getSurveySeenKey` function anymore

lucasheriques: they will still get it, because I'm now importing it from the [shared utility here](https://github.com/PostHog/posthog-js/pull/2215/files#diff-0b4698fe086defb3ddea07f0015da089c548bcc4df94205bc33f0c18c79f051eR16-R21), so it won't break

edit: and, when the `surveys.js` is compiled, it includes all the imports directly on the file. this is really just removing duplicated code

marandaneto (approved):

PR #2228, opened 2025-08-20
feat: Category labels for filtering class methods

gewenyu99:
## Problem

We want to add some filtering to SDK references, they're hard to read:

image

This means we need category info

## Changes

Does two things:
- Tags every method appropriately
- Found some mystery methods that lacked comments, add more comments. Source is other docs.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1657,7 +1657,7 @@ export class PostHog {
      * @param callback - The callback function will be called once the feature flags are ready or when they are updated.
      *                   It'll return a list of feature flags enabled for the user, the variants,
      *                   and also a context object indicating whether we succeeded to fetch the flags or not.
-     * @returns A function that can be called to unsubscribe the listener. Used by useEffect when the component unmounts.
+     * @returns A function that can be called to unsubscribe the listener. Used by `useEffect` when the component unmounts.
      */
     onFeatureFlags(callback: FeatureFlagsCallback): () => void {
         return this.featureFlags.onFeatureFlags(callback)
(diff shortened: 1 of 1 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -1495,56 +1620,83 @@ export class PostHog {
...
+     * ```
      *
-     * @param {Function} [callback] The callback function will be called once the feature flags are ready or when they are updated.
-     *                              It'll return a list of feature flags enabled for the user, the variants,
-     *                              and also a context object indicating whether we succeeded to fetch the flags or not.
-     * @returns {Function} A function that can be called to unsubscribe the listener. Used by useEffect when the component unmounts.
+     * @param callback - The callback function will be called once the feature flags are ready or when they are updated.
+     *                   It'll return a list of feature flags enabled for the user, the variants,
+     *                   and also a context object indicating whether we succeeded to fetch the flags or not.
+     * @returns A function that can be called to unsubscribe the listener. Used by useEffect when the component unmounts.
```
rafaeelaudibert: Do we want something like this here so that we can properly use `code` on the docs?

gewenyu99: Let me check what this renders as 🤔 

I think TSDoc is handling this as expected
Screenshot 2025-08-21 at 11 49 37 AM

rafaeelaudibert: Ah, I forgot to add my code hahaha

I meant using `useEffect` with backticks :)

gewenyu99: Ah yes, that's a good one

gewenyu99:
Ughhhh flaky tests 😆

PR #2230, opened 2025-08-20
fix(llma): various bug fixes related to tools for streaming providers

Radu-Raicea:
## Problem

There are multiple bugs relating to tool for streaming providers, mostly where tool calls do not appear in the output messages, but also Vercel not showing available tools.

## Changes

Fixes:
- Anthropic Streaming missing tool call in output
- Gemini streaming missing tool call in output
- OpenAI Chat Completions API Streaming missing tool call in output
- Vercel missing tool call in output
- Vercel missing $ai_tools in both normal and streaming

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/whole-candles-ask.md
+++ b/.changeset/whole-candles-ask.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Fixes tool calls for streaming providers
--- a/packages/ai/src/anthropic/index.ts
+++ b/packages/ai/src/anthropic/index.ts
@@ -8,6 +8,7 @@ import {
   sendEventToPosthog,
   extractAvailableToolCalls,
 } from '../utils'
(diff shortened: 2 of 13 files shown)


.changeset/whole-candles-ask.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/ai': patch
```
k11kirky, line 2: this is probably more than a patch, i would do minor bump

k11kirky (approved):

.changeset/whole-candles-ask.md
Radu-Raicea: I didn't add any new features, they were bug fixes and no changes in the public API, that's why I thought it would be best put as a patch, WDYT?

PR #2238, opened 2025-08-22
feat: Improve consent APIs

robbie-c:
## Problem

I want to move posthog.com to using our own consent APIs. As part of this dogfooding, I found some places where we could make this experience better.

## Changes

* Accept "yes" and "no" in storage rather than just "1" and "0".
* Add a top-level function which can be used to determine whether the cookie banner should be shown.
* Update the playground to use this, matching how real users might use this

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/playground/nextjs/src/posthog.ts
+++ b/packages/browser/playground/nextjs/src/posthog.ts
@@ -44,7 +44,6 @@ export const configForConsent = (): Partial<PostHogConfig> => {
 }
 
 export const updatePostHogConsent = (consentGiven: ConsentState) => {
-    console.log(`PostHog consent given: ${consentGiven}`)
     if (consentGiven !== undefined) {
         if (consentGiven === 'granted') {
             posthog.opt_in_capturing()


packages/browser/playground/nextjs/src/CookieBanner.tsx
```diff
@@ -1,16 +1,16 @@
...
+export function useCookieConsent(): [ConsentState | null, (consentGiven: ConsentState) => void] {
+    const [consentGiven, setConsentGiven] = useState<ConsentState | null>(null)
 
     useEffect(() => {
         setConsentGiven(cookieConsentGiven())
     }, [])
 
     useEffect(() => {
-        if (consentGiven === null || posthog.config.cookieless_mode === 'always') {
+        if (consentGiven == null || posthog.config.cookieless_mode === 'always') {
```
jabahamondes: I think there are different meanings when consentGiven is null or undefined. Is it possible to have a helper method for those cases? For example `consentNotGiven = consent => consentGiven == null` and `consentNotAnswered =  consent => consent === undefined` and with consentGiven === null

packages/browser/playground/nextjs/src/posthog.ts
```diff
@@ -17,33 +17,41 @@ export const posthog: PostHog = POSTHOG_USE_SNIPPET
...
 
-export const updatePostHogConsent = (consentGiven: boolean) => {
-    if (consentGiven) {
-        posthog.opt_in_capturing()
-    } else {
-        posthog.opt_out_capturing()
+export const updatePostHogConsent = (consentGiven: ConsentState) => {
+    console.log(`PostHog consent given: ${consentGiven}`)
+    // eslint-disable-next-line posthog-js/no-direct-null-check
+    if (consentGiven !== null) {
```
jabahamondes: Here as well, should it enter if consentGiven is undefined?

packages/browser/playground/nextjs/src/posthog.ts
```diff
@@ -17,33 +17,41 @@ export const posthog: PostHog = POSTHOG_USE_SNIPPET
         ? (window as any).posthog
         : null
     : posthogJS
+
+// this is a bit gross, we use null to represent "not yet determined", and undefined for SSR
```
jabahamondes: Ok here is my answer haha could we use the methods still to be explicit with the meaning?

packages/browser/playground/nextjs/src/CookieBanner.tsx
robbie-c: Having thought about this a bit more, I think it makes more sense to return a string rather than a bool, so we can capture more of the state

packages/browser/src/__tests__/consent.test.ts
```diff
@@ -23,16 +23,19 @@ function deleteAllCookies() {
 }
 
 describe('consentManager', () => {
-    const createPostHog = (config: Partial<PostHogConfig> = {}) => {
```
robbie-c, line 26: drive-by fix, these tests have always been flaky, while I was here I decided to try to make them less flaky

packages/browser/playground/nextjs/src/posthog.ts
```diff
@@ -17,33 +17,43 @@ export const posthog: PostHog = POSTHOG_USE_SNIPPET
...
     }
 }
 
-export const updatePostHogConsent = (consentGiven: boolean) => {
-    if (consentGiven) {
-        posthog.opt_in_capturing()
-    } else {
-        posthog.opt_out_capturing()
+export const updatePostHogConsent = (consentGiven: ConsentState) => {
+    console.log(`PostHog consent given: ${consentGiven}`)
```
lricoy: Is the log intentional to keep?

PR #2239, opened 2025-08-22
feat(surveys): add support for conditional survey questions on react-native

ioannisj:
## Problem

Ports conditional question logic from the web sdk

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -104,6 +108,128 @@ export const getDisplayOrderChoices = (question: MultipleSurveyQuestion): string
...
+
+        if (nextStep === SurveyQuestionBranchingType.End) {
+          return SurveyQuestionBranchingType.End
+        }
+
+        return nextQuestionIndex
+      }
+    } else if (question.type === SurveyQuestionType.Rating) {
+      if (typeof response !== 'number' || !Number.isInteger(response)) {
+        throw new Error('The response type must be an integer')
```
marandaneto: what will happen when this and other methods throws? will crash the app?

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -104,6 +108,128 @@ export const getDisplayOrderChoices = (question: MultipleSurveyQuestion): string
...
+        }
+
+        if (nextStep === SurveyQuestionBranchingType.End) {
+          return SurveyQuestionBranchingType.End
+        }
+
+        return nextQuestionIndex
+      }
+    } else if (question.type === SurveyQuestionType.Rating) {
+      if (typeof response !== 'number' || !Number.isInteger(response)) {
```
marandaneto: this could be an extension, i see the `Number.isInteger` multiple times, but we dont check for `typeof response !== 'number'` everywhere

packages/react-native/src/surveys/surveys-utils.ts
ioannisj: Ah, i'll test but I imagine it would yeah. I'll just log in that case and return next index

marandaneto (approved):

marandaneto:
@ioannisj we should update the docs after https://posthog.com/docs/surveys/installation?tab=React+Native

PR #2246, opened 2025-08-26
feat: lazy load much more of session replay

pauldambra:
a lot of the session replay code predates (or at least overlapped with the introduction of) our lazy loading
so, even though we lazy load rrweb
a large chunk of the bundle is the code that interacts with rrweb

this PR moves as much as possible of that code into a lazy loaded extension

## what does this change?

* for customers that aren't enabling replay the bundle is ~14% smaller

* for customers that are enabling replay we've removed the race between starting rrweb and receiving remoteconfig, this is probably ok, since remote config needs to be fast for flags, but does mean _some_ customers might see replay start very slight later than today... but given we (want to) control what script to load via remote config and allow configuring masking via remote config this is probably desirable overall

### DIFF
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -523,8 +523,8 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         }
         const currentUrl = this._maskUrl(window.location.href)
         if (this._lastHref !== currentUrl) {
-            this._tryAddCustomEvent('$url_changed', { href: currentUrl })
             this._lastHref = currentUrl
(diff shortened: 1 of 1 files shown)


packages/browser/jest.config.js
```diff
@@ -1,4 +1,9 @@
 module.exports = {
+    testMatch: [
+        '**/__tests__/**/*.test.[jt]s?(x)',
+        '**/?(*.)+test.[jt]s?(x)',
+        '**/functional_tests/**/*.test.[jt]s?(x)',
+    ],
```
pauldambra, line 6: i had to do this because i extracted a `utils.ts` file in the tests folder and then jest was unhappy because it was trying to run it as a test file

pauldambra:
> Closed? It's still really huge.

👋 this was implemented in a different PR

PR #2245, opened 2025-08-25
feat: Node.js reference docs

gewenyu99:
## Problem

Prep for node.js reference docs. Refactors generation scripts to be general and imported from posthog-js.

## Changes

- Lift scripts that work for general typescript parsing to top level
- Add script/config in Node.js to parse built docs
- In a future PR, we'll add in annotation comments for exposed methods :)

PR #2248, opened 2025-08-26
chore: Add doc comments to generate into specs.

gewenyu99:
## Problem

We don't have doc comments on exposed methods. We want these to generate specs. Stacked on https://github.com/PostHog/posthog-js/pull/2245

## Changes

Doc comments from the existing node.js docs.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -352,7 +352,7 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
    *   sendFeatureFlags: true
    * })
    * ```
-   * 
+   *
    * @example
    * ```ts
    * // With custom feature flags options


gewenyu99:
🤔 Is package size a legitimate concern when we add code comments?

packages/node/src/client.ts
```diff
@@ -42,6 +42,35 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+   * ```ts
+   * // Basic initialization
+   * const client = new PostHogBackendClient(
+   *   'your-api-key',
+   *   { host: 'https://app.posthog.com' }
+   * )
+   * ```
+   * 
+   * @example
+   * ```
```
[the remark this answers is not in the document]
rafaeelaudibert: @gewenyu99 this makes sense, should we add ts here?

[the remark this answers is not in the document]
gewenyu99: Oh yeah totally 🤦 I forgot to do this.

packages/node/src/client.ts
```diff
@@ -163,6 +330,34 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+   * })
+   * ```
+   *
+   * @example
+   * ```ts
+   * // With feature flags
+   * await client.captureImmediate({
+   *   distinctId: 'user_123',
+   *   event: 'user_action',
+   *   sendFeatureFlags: true
```
dmarticus, line 352: NB – perhaps worth mentioning that `sendFeatureFlags` is of type `boolean | SendFeatureFlagsOptions`, and `SendFeatureFlagsOptions` lets users specify more information about how to send feature flags along with the captured data

```
export interface SendFeatureFlagsOptions {
  onlyEvaluateLocally?: boolean
  personProperties?: Record<string, any>
  groupProperties?: Record<string, Record<string, any>>
  flagKeys?: string[]
}
```

Maybe worth including another example with feature flags using the `SendFeatureFlagsOptions`.

dmarticus (approved):
one comment on `sendFeatureFlags`, otherwise looks good from the flags perspective

packages/node/src/client.ts
gewenyu99: Hey, this is actually also not documented anywhere else. Do you have more info on the properties + how to use this? I'lll add it in specs and in feature flag docs somewhere :D

dmarticus: it's documented here: https://posthog.com/docs/libraries/node#advanced-usage-v550, so somewhat buried, but each SDK that supports this behavior has docs for it on the library-specific page.

gewenyu99: Ohhh nice, that's where it is! Inkeep lied to me. Thanks!

Issue #2214, opened 2025-08-13
[LLMO] Support $ai_embedding


Radu-Raicea:
## Is your feature request related to a problem?

The Python SDK for OpenAI supports creating $ai_embedding events, but the Node SDK does not.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config

PR #2253, opened 2025-08-28
feat(llma): support-openai-embeddings

carlos-marchal-ph:
Implements embeddings for OpenAI. Solves #2214.

### DIFF
--- a/packages/ai/src/anthropic/index.ts
+++ b/packages/ai/src/anthropic/index.ts
@@ -46,6 +46,8 @@ export class PostHogAnthropic extends AnthropicOriginal {
 
 export class WrappedMessages extends AnthropicOriginal.Messages {
   private readonly phClient: PostHog
+  // TODO: Figure out where this is being assigned and type it better
+  private readonly baseURL?: string
 
   constructor(parentClient: PostHogAnthropic, phClient: PostHog) {
     super(parentClient)
@@ -221,7 +223,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
                 input: sanitizeAnthropic(mergeSystemPrompt(anthropicParams, 'anthropic')),
                 output: formattedOutput,
                 latency,
-                baseURL: (this as any).baseURL ?? '',
+                baseURL: this.baseURL ?? '',
                 params: body,
                 httpStatus: 200,
                 usage,
@@ -239,7 +241,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
                 input: sanitizeAnthropic(mergeSystemPrompt(anthropicParams, 'anthropic')),
                 output: [],
                 latency: 0,
-                baseURL: (this as any).baseURL ?? '',
+                baseURL: this.baseURL ?? '',
                 params: body,
                 httpStatus: error?.status ? error.status : 500,
                 usage: {
@@ -275,7 +277,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
               input: sanitizeAnthropic(mergeSystemPrompt(anthropicParams, 'anthropic')),
               output: formatResponseAnthropic(result),
               latency,
-              baseURL: (this as any).baseURL ?? '',
+              baseURL: this.baseURL ?? '',
               params: body,
               httpStatus: 200,
               usage: {
@@ -300,7 +302,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
             input: sanitizeAnthropic(mergeSystemPrompt(anthropicParams, 'anthropic')),
             output: [],
             latency: 0,
-            baseURL: (this as any).baseURL ?? '',
+            baseURL: this.baseURL ?? '',
             params: body,
             httpStatus: error?.status ? error.status : 500,
             usage: {
--- a/packages/ai/src/anthropic/index.ts
+++ b/packages/ai/src/anthropic/index.ts
@@ -46,12 +46,12 @@ export class PostHogAnthropic extends AnthropicOriginal {
 
 export class WrappedMessages extends AnthropicOriginal.Messages {
   private readonly phClient: PostHog
-  // TODO: Figure out where this is being assigned and type it better
-  private readonly baseURL?: string
+  private readonly baseURL: string
 
   constructor(parentClient: PostHogAnthropic, phClient: PostHog) {
     super(parentClient)
     this.phClient = phClient
+    this.baseURL = parentClient.baseURL
   }
 
   public create(body: MessageCreateParamsNonStreaming, options?: RequestOptions): APIPromise<Message>
@@ -138,7 +138,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
                 // Handle text delta events
                 if ('delta' in chunk) {
                   if ('text' in chunk.delta) {
-                    const delta = chunk?.delta?.text ?? ''
+                    const delta = chunk?.delta?.text
 
                     accumulatedContent += delta
 
@@ -223,7 +223,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
                 input: sanitizeAnthropic(mergeSystemPrompt(anthropicParams, 'anthropic')),
                 output: formattedOutput,
                 latency,
-                baseURL: this.baseURL ?? '',
+                baseURL: this.baseURL,
                 params: body,
                 httpStatus: 200,
                 usage,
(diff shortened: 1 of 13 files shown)


packages/ai/src/openai/azure.ts
```diff
@@ -589,6 +593,89 @@ export class WrappedResponses extends AzureOpenAI.Responses {
...
+
+  public create(
+    body: EmbeddingCreateParams & MonitoringParams,
+    options?: RequestOptions
+  ): APIPromise<CreateEmbeddingResponse> {
+    const {
+      posthogDistinctId,
+      posthogTraceId,
+      posthogProperties,
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
Radu-Raicea: Do we need this disable? If we remove it, can we easily fix what it's complaining about?

packages/ai/src/openai/index.ts
```diff
@@ -620,6 +631,89 @@ export class WrappedResponses extends Responses {
...
+          client: this.phClient,
+          eventType: '$ai_embedding',
+          distinctId: posthogDistinctId,
+          traceId,
+          model: openAIParams.model,
+          provider: 'openai',
+          input: withPrivacyMode(this.phClient, posthogPrivacyMode, openAIParams.input),
+          output: null, // Embeddings don't have output content
+          latency,
+          baseURL: (this as any).baseURL ?? '',
```
Radu-Raicea: Is there a way to make this pattern more type safe? This is common throughout our package.

`(this as any).baseURL`

packages/ai/src/utils.ts
```diff
@@ -343,6 +349,7 @@ function sanitizeValues(obj: any): any {
 
 export const sendEventToPosthog = async ({
   client,
+  eventType = '$ai_generation',
```
Radu-Raicea: This can be tied well with the enum suggested by Greptile.

Radu-Raicea (approved):
Overall, this looks good. The behaviour looks right, and the code is okay.

Greptile and I left a few comments for some minor code improvements. They are optional, but it's going to be hard to prioritize code improvement over new features and bug fixes, so I think that we should take every chance that we get to fix any broken windows our code touches.

Don't forget to run `pnpm changeset` to generate a new library version (if these changes are backwards compatible, it's probably a minor version update). You also need to add the "release" label to the PR for it to trigger the npm publishing.

packages/ai/src/openai/azure.ts
```diff
@@ -589,6 +593,89 @@ export class WrappedResponses extends AzureOpenAI.Responses {
...
+  }
+
+  public create(
+    body: EmbeddingCreateParams & MonitoringParams,
+    options?: RequestOptions
+  ): APIPromise<CreateEmbeddingResponse> {
+    const {
+      posthogDistinctId,
+      posthogTraceId,
+      posthogProperties,
```
Radu-Raicea: We tend to destructure these, without using them. Let's not destructure them then.

Radu-Raicea:
One last comment, we should accompany this with changes in https://github.com/PostHog/llm-analytics-apps to add the Node embeddings to the CLI, so that we can easily run embeddings tests whenever we make changes to them, without having to prepare a local setup.

packages/ai/tests/openai.test.ts
```diff
@@ -1005,4 +1036,135 @@ describe('PostHogOpenAI - Jest test suite', () => {
...
+        ],
+        model: 'text-embedding-3-small',
+        usage: {
+          prompt_tokens: 8,
+          total_tokens: 8,
+        },
+      }
+
+      const EmbeddingsMock: any = openaiModule.Embeddings || class MockEmbeddings {}
+      EmbeddingsMock.prototype.create = jest.fn().mockResolvedValue(mockOpenAiEmbeddingResponse)
```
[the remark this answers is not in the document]
carlos-marchal-ph: In this case it would save like two lines and make the tests more confusing since you would have to go the start of the file to understand how they are being mocked.

packages/ai/src/utils.ts
carlos-marchal-ph: What do you mean? I will move the types to an enum as Greptile suggests, but this line is fine as is, right? It's just setting the default value for backwards compatibility.

Radu-Raicea: I meant setting the default value as `eventType = EnumName.EnumValue`.

packages/ai/src/openai/azure.ts
carlos-marchal-ph: I removed these. I also saw in the eslint config that we specifically disable the rule that warns about stuff like this. Any idea why? Pretty easy to slip up otherwise, especially when using AI coding tools.

packages/ai/src/openai/index.ts
carlos-marchal-ph: I did my best without sinking a huge amount of time into it, it's pretty weird and neither me nor Claude could figure out where this is getting assigned or why it's accessible.

packages/ai/src/openai/azure.ts
Radu-Raicea: Is it only for our package? If yes, what happens when you enable it? Is it a manageable amount to fix?

packages/ai/src/openai/azure.ts
```diff
@@ -615,7 +622,7 @@ export class WrappedEmbeddings extends AzureOpenAI.Embeddings {
     const startTime = Date.now()
 
     const parentPromise = super.create(openAIParams, options)
-
+    type A = typeof this
```
Radu-Raicea: I think this is a debugging artifact

packages/ai/src/utils.ts
carlos-marchal-ph: Oh my bad, just saw that we are using actual enums in the code. I was used to using union types, will make the switch!

carlos-marchal-ph: Fixed!

packages/ai/src/openai/azure.ts
carlos-marchal-ph: It's globally for several packages, I removed it from ours and fixed the offenses, wasn't too hard and now we have a set way to customise linting rules for the AI package.

packages/ai/src/openai/azure.ts
carlos-marchal-ph: Yes, already gone!

packages/ai/src/anthropic/index.ts
```diff
@@ -137,7 +130,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
                 // Handle text delta events
                 if ('delta' in chunk) {
                   if ('text' in chunk.delta) {
-                    const delta = chunk?.delta?.text ?? ''
+                    const delta = chunk?.delta?.text
```
Radu-Raicea: Given that it's an `any` type, `text` could be anything, even a non-string. Maybe we should keep the `?? ''`.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -85,22 +85,15 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
 
-  public handleChainEnd(
-    outputs: ChainValues,
-    runId: string,
-    parentRunId?: string,
-    tags?: string[],
-    // eslint-disable-next-line @typescript-eslint/no-unused-vars
-    kwargs?: { inputs?: Record<string, unknown> }
-  ): void {
+  public handleChainEnd(outputs: ChainValues, runId: string, parentRunId?: string, tags?: string[]): void {
```
Radu-Raicea: I think that this is public facing, right? We probably want to keep the signature the same.

Radu-Raicea (approved):
I hope I didn't miss anything, but other than my two comments that need to be addressed, this looks good to go!

You need to add the "release" label too for it to publish a new package version.

packages/ai/src/anthropic/index.ts
carlos-marchal-ph: It shows up as `string` for me 🤔 And the conditional access is probably redundant too since we are checking right before.
Screenshot 2025-08-29 at 18 24 56

packages/ai/src/langchain/callbacks.ts
carlos-marchal-ph: Yep, that one slipped through, meant to just add an underscore before it. Fixed now!

packages/ai/src/anthropic/index.ts
Radu-Raicea: Nevermind! I think I had an older branch. If we don't need the conditional accesses, let's remove them too.

PR #2263, opened 2025-09-02
chore: add a manual changeset because i forgot

pauldambra:
follow up to https://github.com/PostHog/posthog-js/pull/2259 and https://github.com/PostHog/posthog-js/pull/2251 because i forgot the bump-foo labels don't work any more


pauldambra:
oh, wat
is that not automatic any more?

pauldambra:
ok, instructions say, let's see if that woks

Screenshot 2025-09-02 at 12 58 46

PR #2267, opened 2025-09-02
feat: allow opting in to testing session replay code lazy loading

pauldambra:
this is a pretty big change to lazy loading replay
let's not release it to everyone all at once
this lets client config opt you in so we can test this in posthog's usage

this is a 1.6% bump now
so that we can test a 16% reduction


packages/browser/src/extensions/replay/sessionrecording-wrapper.ts
```diff
@@ -0,0 +1,229 @@
...
+    }
+
+    /**
+     * this ignores the URL/Event trigger config and (if other conditions are met) causes capture to start
+     *
+     * It is not usual to call this directly,
+     * instead call `posthog.startSessionRecording({trigger: 'url' | 'event'})`
+     * */
+    public overrideTrigger(triggerType: TriggerType) {
+        // TODO what if this gets called before lazy loading is done
```
rafaeelaudibert, line 213: I don't think people will call this programatically, so it's probably fine to simply fail with an error if this is called before loading's finished, letting people know they just need to wait before calling this again. Maybe return a boolean indicating whether it worked? I'd just raise an error tho

pauldambra: Ah, yep... I hadn't decided what to do here... So thought well I'll test if it all works at all anyway first 🙈

PR #2269, opened 2025-09-02
chore: add test that would have presented noise

pauldambra:
follow-up to

https://github.com/PostHog/posthog-js/pull/2268
https://github.com/PostHog/posthog-js/pull/2267

this adds the test that would have protected me from the type mistake i released


packages/browser/src/__tests__/posthog-core-also.test.ts
pauldambra, line 1: it always makes me sad that this test file isn't called `.test.ts`

packages/browser/src/__tests__/posthog-core.loaded.test.ts
```diff
@@ -30,6 +30,20 @@ describe('loaded() with flags', () => {
         return posthog
     }
 
+    describe('replay', () => {
+        it('can set force allow localhost network capture', async () => {
+            instance = await createPosthog({
+                loaded: (ph) => {
+                    if (ph.sessionRecording) {
+                        ph.sessionRecording._forceAllowLocalhostNetworkCapture = true
```
pauldambra, line 38: in the faulty version this line wasn't possible as `_forceAllowLocalhostNetworkCapture` wasn't available on both types

PR #2270, opened 2025-09-02
fix: hosted surveys should always render

lucasheriques:
- **fix: hosted surveys should always render even if delay**
- **add test**

## Problem

Hosted surveys that are saved with a delay are not showing.

## Changes

Ignore the delay when setting the survey popup visibility. Plus remove any style position changes from them.

I'll also change the function that sanitizes the survey payload to remove the delay if the survey is hosted, however, let's also make the change in the SDK so it works regardless.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -782,7 +782,9 @@ export function usePopupVisibility(
     removeSurveyFromFocus: (survey: SurveyWithTypeAndAppearance) => void,
     surveyContainerRef?: React.RefObject<HTMLDivElement>
 ) {
-    const [isPopupVisible, setIsPopupVisible] = useState(isPreviewMode || millisecondDelay === 0)
+    const [isPopupVisible, setIsPopupVisible] = useState(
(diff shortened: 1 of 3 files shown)


packages/browser/src/extensions/surveys.tsx
```diff
@@ -907,6 +909,10 @@ function getPopoverPosition(
     position: SurveyPosition = SurveyPosition.Right,
     surveyWidgetType?: SurveyWidgetType
 ) {
+    if (type === SurveyType.ExternalSurvey) {
```
marandaneto, line 912: similar question to https://github.com/PostHog/posthog/pull/37542#issuecomment-3248144613

marandaneto (approved):
approving because of back compatibility so this is needed.
i think we UI and API should not allow creating an external survey with such conditions

PR #2272, opened 2025-09-03
fix: dynamically extract posthog params from OpenAI calls

carlos-marchal-ph:
This PR accomplishes several things:
1. Centralise logic for preparing OpenAI related events and its defaults
2. Ensure that we never send any Posthog specific parameter to OpenAI, [which causes it to fail the request](https://posthog.com/questions/400-unknown-parameter-posthog-porperties)
3. Ensure that all Posthog specific params are always passed to the event by default


packages/ai/src/openai/util.ts
```diff
@@ -0,0 +1,41 @@
...
+  }
+}
+
+function isPosthogParam(key: string): key is keyof MonitoringParams {
+  return key.startsWith('posthog')
+}
+
+function removePosthogPrefix(key: keyof MonitoringParams): keyof MonitoringEventProperties {
+  const unprefixed = key.replace(/^posthog/, '')
+  return (unprefixed.charAt(0).toLowerCase() + unprefixed.slice(1)) as keyof MonitoringEventProperties
```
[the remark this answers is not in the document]
carlos-marchal-ph: `posthog: foo` is not a valid parameter for the OpenAI client

packages/ai/src/openai/util.ts
```diff
@@ -0,0 +1,41 @@
...
+    openAIParams: openAIParams as T,
+    posthogParams: addDefaults(posthogParams as MonitoringEventProperties),
+  }
+}
+
+function isPosthogParam(key: string): key is keyof MonitoringParams {
+  return key.startsWith('posthog')
+}
+
+function removePosthogPrefix(key: keyof MonitoringParams): keyof MonitoringEventProperties {
```
Radu-Raicea: I'm not seeing the benefit of this approach. I think it's actually making the code harder to understand. Am I missing some benefit from creating a new interface that has no posthog prefix on the parameters, and spreading that object like this instead of explicitly setting each parameter?

```
await sendEventToPosthog({


          client: this.phClient,
          eventType: AIEvent.Embedding,
          ...posthogParams,
```

packages/ai/tests/azure-openai.test.ts
```diff
@@ -638,4 +638,27 @@ describe('PostHogAzureOpenAI - Embeddings test suite', () => {
       expect(groups).toEqual(testGroups)
     })
   })
+
+  conditionalTest('posthogProperties are not sent to Azure OpenAI', async () => {
```
Radu-Raicea, line 642: Let's remove the concept of `conditionalTest` from these two test suites. We shouldn't make API calls in tests, and I want to make sure that these two test suites run in the CI.

packages/ai/src/openai/util.ts
carlos-marchal-ph: I realized there’s a layer of indirection between the actual client call and where we pass parameters, which means we don’t currently forward extra options as intended. My original idea was simply to pass those options through to the client. I’ll simplify the types, but I’d still keep the parameter extraction functions, since incorrectly passing parameters can cause requests to fail.

The benefit of using spreading is that it centralizes control, so we can update which parameters get passed to the capture function in a single place. There’s already quite a bit of repetition in these files, such as the repeated calls to `sendEventToPosthog` with almost identical params. I don’t feel strongly about this and am happy to adjust if you prefer a different approach.

packages/ai/tests/azure-openai.test.ts
carlos-marchal-ph: I will tackle this as a separate PR to remove all conditional tests from all providers (same goes for the Python SDK). In any case, I think most of the tests mock the AI clients and just check for our library's transformation behaviour. I actually ran some of them with a dummy API key and they still passed, so it should be quite straightforward to fix this dependency and keep them as unit tests.

We should have some sort of integration testing against real API requests though, maybe using some sort of request recording and mocking library?

Radu-Raicea: Sure 👍 I might have modified a bunch of tests to add mocks, without removing the need for the fake API key 😅 

> We should have some sort of integration testing against real API requests though, maybe using some sort of request recording and mocking library?

Yes, I'd love to see something like that, do you have any suggestions?

carlos-marchal-ph: I do not, but I can take care of doing a bit of research and doing a PoC as part of this test fixing PR :)

packages/ai/src/openai/util.ts
Radu-Raicea: I think the part that I find a bit odd is the removal of the posthog prefix dynamically. Maybe I'm more of a stickler to being explicit rather than implicit. I'll approve this PR and let you choose whether to keep it like this or change it 👍

carlos-marchal-ph: Ah I get that, metaprogramming can be a bit of a footgun. I made it more explicit, now the list of params and their renames are stored in a constant, and we get an error if a posthog variable is not mapped explicitly.

PR #2274, opened 2025-09-03
Fix initial pageview not firing on cookieless_mode: always

robbie-c:
## Changes

Fix a bug where we might not send the initial pageview event when using `cookieless_mode: 'always'`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/cookieless.test.ts
+++ b/packages/browser/src/__tests__/cookieless.test.ts
@@ -44,11 +44,18 @@ jest.mock('../utils/globals', () => {
                 toString: () => url,
             }
         },
+        XMLHttpRequest: () => ({
+            open: jest.fn(),
+            send: jest.fn(),
+            setRequestHeader: jest.fn(),
+        }),
     }
 })
 
 // eslint-disable-next-line @typescript-eslint/no-require-imports
 const { mockURLGetter, mockedCookieBox, document } = require('../utils/globals')
+
+const delay = (timeoutMs: number) => new Promise((resolve) => setTimeout(resolve, timeoutMs))
 
 describe('cookieless', () => {
     const eventName = 'custom_event'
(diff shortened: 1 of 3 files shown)


pauldambra (approved):
looks good to me

packages/browser/src/__tests__/cookieless.test.ts
```diff
@@ -93,6 +100,33 @@ describe('cookieless', () => {
             // should ignore cookie consent, and throw in test code due to logging an error
             expect(() => posthog.opt_in_capturing()).toThrow()
         })
+
+        it.each([[true], ['history_change']])(
+            'should send the initial pageview event when capture_pageview is %p',
```
rafaeelaudibert, line 105: Hmm, somewhat misleading here because we're not varying cookieless_mode on this test, maybe improve the name?

```suggestion
            'should send the initial pageview event when cookieless_mode is "always" and capture_pageview is %p',
```

rafaeelaudibert (approved):
I assume we've tested and this properly works when we "accept/reject" cookies on `on_reject` mode?

jabahamondes (approved):

packages/browser/src/__tests__/cookieless.test.ts
robbie-c: ~~Oh that's true, good catch~~

robbie-c: No - wait - the test name is correct :D

It's under a `describe('always mode', ...)` block

And yes, the initial pageview is tested separately in the describe('on_reject', ...)` block

PR #2276, opened 2025-09-03
fix(flags): `isFeatureEnabled` discerns missing feature flags in return value

### LINKED ISSUE #2247: useFeatureFlagEnabled and isFeatureEnabled behave differently
## Bug description

I noticed  `useFeatureFlagEnabled`  and `isFeatureEnabled` behave differently when a feature flag ("FF") is disabled.

`useFeatureFlagEnabled` returns `false` when a FF is disabled.

Whereas, `isFeatureEnabled` returns `undefined`...

I would expect that `useFeatureFlagEnabled` returns `undefined` when a flag is disabled, so that the FF has three states: `true` (enabled and rolled out to that user), `false` (enabled but not rolled out to that user) and `undefined` (disabled).

This makes handling the different responses difficult because when using `isFeatureEnabled` we can evaluate `flag === false` to know that it's enabled but not available to the user.

However, this evaluation breaks when using `useFeatureFlagEnabled` because `flag === false` could mean the flag is either disabled or enabled but not available to that user.

## How to reproduce

1. Call `useFeatureFlagEnabled` with a disabled flag
2.`false` is returned

1. Call `isFeatureEnabled` with a disabled flag
2.`undefined` is returned

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Additional context
versions used: 
    "posthog-js": "^1.260.1",
    "posthog-node": "^4.18.0"

dustinbyrne:
## Problem

`isFeatureEnabled` currently returns `false` in cases where the specified feature flag is disabled or does not exist. This means cases where the feature flag exists but has not matched conditions cannot be differentiated.

## Changes

`isFeatureEnabled` now correctly returns `undefined` in cases where if the specified feature flag does not exist or is disabled.

Resolves #2247

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/shiny-bushes-stare.md
+++ b/.changeset/shiny-bushes-stare.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+`isFeatureEnabled` now returns `undefined` (instead of `false`) for missing or disabled feature flags, aligning with the documentation. Previously, `undefined` was returned only before flags had loaded, and missing/disabled flags returned `false`. This change clarifies the difference between flags that exist but don't match (`false`) and flags that don’t exist or are disabled (`undefined`).
--- a/packages/browser/src/__tests__/featureflags.test.ts
+++ b/packages/browser/src/__tests__/featureflags.test.ts
@@ -163,7 +163,7 @@ describe('featureflags', () => {
             'disabled-flag': false,
         })
         expect(featureFlags.isFeatureEnabled('beta-feature')).toEqual(true)
-        expect(featureFlags.isFeatureEnabled('random')).toEqual(false)
+        expect(featureFlags.isFeatureEnabled('random')).toEqual(undefined)
         expect(featureFlags.isFeatureEnabled('multivariate-flag')).toEqual(true)
 
         expect(instance.capture).toHaveBeenCalledTimes(3)
@@ -214,7 +214,7 @@ describe('featureflags', () => {
             $enabled_feature_flags: { x: 'y' },
         })
         expect(featureFlags.getFlagVariants()).toEqual({ x: 'y' })
-        expect(featureFlags.isFeatureEnabled('beta-feature')).toEqual(false)
+        expect(featureFlags.isFeatureEnabled('beta-feature')).toEqual(undefined)
         expect(instance.capture).toHaveBeenCalledTimes(2)
 
         instance.persistence.register({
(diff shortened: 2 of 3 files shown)


dustinbyrne:
For what it's worth, I noticed our documentation does accurately describe the intended return value here:
https://posthog.com/docs/references/posthog-js#PostHog-isFeatureEnabled

> Returns true if the flag is enabled, false if disabled, or undefined if not found. This is a convenience method that treats any truthy value as enabled.

Given this change affects the return value in a manner which is potentially backwards incompatible, we can stick with the major version bump to be safe. Though, I feel better knowing that downstream consumers _should be_ properly handling this.

.changeset/shiny-bushes-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': major
```
pauldambra: i'm tempted to say this is a fix

the behaviour of the sdk is wrong
and now it would be right

this is only breaking if you are coupled to the incorrect behaviour

when we merge the change, we open a PR against posthog main, that updates the version there and when that is released it updates our CDN. ~five minutes later everyone using `array.js` in the snippet gets this code when they next refresh

i think importantly for me we're not changing the type signature which remains`boolean | undefined`  we just would never previously have ever returned undefined

(will catch up in slack or synchronously about the avoiding breaking breaking changes)

marandaneto: yeah it sounds like a patch to me as well, the return type was unchanged, but they did mention the reasoning [here](https://github.com/PostHog/posthog-js/pull/2276#issuecomment-3254907502) which is fair (to be extra careful).
people might have implemented things based on the current implementation, and that would break their business logic.

dustinbyrne: Thanks for the feedback and the additional context, it's super helpful!

I agree that this can be a patch considering:
- the return type is unchanged
- the documentation has already described the correct return values

It may still affect the behavior of any clients who relied on the incorrect behavior depending on how they implemented it. It might be a little odd, but I suppose it _is_ possible, e.g.:
```ts
const enabled = posthog.isFeatureEnabled('foo') ?? true
```
or
```ts
if (posthog.isFeatureEnabled('foo') !== false) { ... }
```

I don't yet have a great sense of whether this is an acceptable amount of backwards incompatibility in exchange for the bug fix. Perhaps it's safer to wait until we have better options for end users to update or roll back their snippet.

pauldambra: i haven't looked at the implementation but @robbie-c added `defaults` to the config if i remember the name correctly which was intended to let us version things
so you could be super cautious and add a new "default" so that older clients keep the old behaviour and new clients get the new behaviour

it's such a small change that i'm tempted to yolo but a call for the FF team tbh

i added "request changes" just cos i didn't have time to read the PR and saw major change and panicked 🤣

haacked: Given the type and docs make it clear that there are situations you have to handle undefined, I’m fine with making this a patch release. I think the risk is really really small.

prandell:
It is absolutely crazy to me that such a breaking change like this was merged as a patch. Having a tri-state here is fundamental to being able to show loading states until a flag resolves, preventing the disabled state from rendering before flipping to enabled. Coming here and reading “yolo” after our production app was taken down for multiple hours is … not what I expected.

PR #2279, opened 2025-09-04
fix: flappy console logging in test

pauldambra:
every so often the segment integration tests fail because of unexpected console logging

this is the result of a race where the posthog instance in the test is actually trying to load surveys
and cannot, and then logs if it manages to fail to load surveys before the test assertions run

in the test in question we are explicitly disabling surveys but ignoring that when logging in onRemoteConfig

there is imho a bigger bug here that this test exposes by accident, but at least to fix the flap, there is no point logging that we aren't loading surveys when client config has told us to disable surveys

### DIFF
--- a/packages/browser/src/posthog-surveys.ts
+++ b/packages/browser/src/posthog-surveys.ts
@@ -35,11 +35,11 @@ export class PostHogSurveys {
     }
 
     onRemoteConfig(response: RemoteConfig) {
+        // only load surveys if they are enabled and there are surveys to load
         if (this._instance.config.disable_surveys) {
             return
         }
 
-        // only load surveys if they are enabled and there are surveys to load
         const surveys = response['surveys']
         if (isNullish(surveys)) {
             return logger.warn('Flags not loaded yet. Not loading surveys.')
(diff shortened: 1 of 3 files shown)


marandaneto (approved):

packages/browser/src/posthog-surveys.ts
```diff
@@ -37,7 +37,7 @@ export class PostHogSurveys {
     onRemoteConfig(response: RemoteConfig) {
         // only load surveys if they are enabled and there are surveys to load
         const surveys = response['surveys']
-        if (isNullish(surveys)) {
+        if (isNullish(surveys) && !this._instance.config.disable_surveys) {
```
marandaneto: not sure if this is needed since `onRemoteConfig` should not even be called if `disable_surveys=true` unless this is only for tests not respecting `disable_surveys` and calling it directly

pauldambra: >  since onRemoteConfig should not even be called if disable_surveys=true

I had to move that to the top since the next logline in this method caused the failure

`onRemoteConfig` is called without checking any config. so each recipient of the call needs to decide if it should process

i'm assuming this is safe here

(agree that the log assertion is a bit annoying)

pauldambra: https://github.com/PostHog/posthog-js/blob/f2ac9569d80a10961919d51a8132e0153b471a69/packages/browser/src/posthog-core.ts#L686-L694

pauldambra: since I had to change behaviour here, i moved the test into this PR too

marandaneto (approved):

PR #2280, opened 2025-09-04
chore: show bug with failing test

pauldambra:
follow-up to https://github.com/PostHog/posthog-js/pull/2279

surveys has a private isEnabled field
it is set to true if the remote config response has surveys in it

this ignores the `disable_surveys` config on client config
but it probably shouldn't

it showed up as a failing test in the linked PR
i have no idea if it actually surfaces as user experienced difference

### DIFF
--- a/packages/browser/src/__tests__/surveys.test.ts
+++ b/packages/browser/src/__tests__/surveys.test.ts
@@ -1561,7 +1561,7 @@ describe('surveys', () => {
         it('can be disabled by config despite results of onRemoteConfig', () => {
             surveys['_instance'].config.disable_surveys = true
             surveys.onRemoteConfig({
-                surveys: false,
+                surveys: ['example'],
             } as Partial<RemoteConfig> as RemoteConfig)
             expect(surveys['_isSurveysEnabled']).toBe(false)
         })
--- a/packages/browser/src/__tests__/surveys.test.ts
+++ b/packages/browser/src/__tests__/surveys.test.ts
@@ -1555,7 +1555,7 @@ describe('surveys', () => {
             surveys.onRemoteConfig({
                 surveys: false,
             } as Partial<RemoteConfig> as RemoteConfig)
-            expect(surveys['_isSurveysEnabled']).toBe(true)
+            expect(surveys['_isSurveysEnabled']).toBe(false)
         })
 
         it('can be disabled by config despite results of onRemoteConfig', () => {
--- a/packages/browser/src/__tests__/surveys.test.ts
+++ b/packages/browser/src/__tests__/surveys.test.ts
@@ -1546,7 +1546,7 @@ describe('surveys', () => {
         it('is enabled by having results in onRemoteConfig', () => {
             expect(surveys['_isSurveysEnabled']).toBe(undefined)
             surveys.onRemoteConfig({
-                surveys: true,
+                surveys: ['example'],
             } as Partial<RemoteConfig> as RemoteConfig)
             expect(surveys['_isSurveysEnabled']).toBe(true)
         })
--- a/packages/browser/src/__tests__/surveys.test.ts
+++ b/packages/browser/src/__tests__/surveys.test.ts
@@ -20,7 +20,7 @@ import {
     SurveyQuestionType,
     SurveyType,
 } from '../posthog-surveys-types'
-import { FlagsResponse, PostHogConfig, Properties } from '../types'
+import { FlagsResponse, PostHogConfig, Properties, RemoteConfig } from '../types'
 import * as globals from '../utils/globals'
 import { assignableWindow, window } from '../utils/globals'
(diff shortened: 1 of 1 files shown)


packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1537,4 +1537,33 @@ describe('surveys', () => {
...
+                surveys: true,
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(true)
+        })
+
+        it('is disabled by having no results in onRemoteConfig', () => {
+            surveys.onRemoteConfig({
+                surveys: false,
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(true)
```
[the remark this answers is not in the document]
marandaneto: should be false

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1537,4 +1537,33 @@ describe('surveys', () => {
...
+
+    describe('is enabled', () => {
+        it('starts effectively disabled', () => {
+            expect(surveys['_isSurveysEnabled']).toBe(undefined)
+        })
+
+        it('is enabled by having results in onRemoteConfig', () => {
+            expect(surveys['_isSurveysEnabled']).toBe(undefined)
+            surveys.onRemoteConfig({
+                surveys: true,
```
marandaneto: this should be an array with surveys in it
https://github.com/PostHog/posthog-js/blob/00601e3725b249a0f9a9d7d69c583928e1bbfd5d/packages/browser/src/posthog-surveys.ts#L44

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1537,4 +1537,33 @@ describe('surveys', () => {
...
+            surveys.onRemoteConfig({
+                surveys: false,
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(true)
+        })
+
+        it('can be disabled by config despite results of onRemoteConfig', () => {
+            surveys['_instance'].config.disable_surveys = true
+            surveys.onRemoteConfig({
+                surveys: false,
```
marandaneto: same as https://github.com/PostHog/posthog-js/pull/2280/files#r2321756480

packages/browser/src/__tests__/surveys.test.ts
pauldambra: oh, yep, was just for the test so I didn't need to figure out the format of the array... although i guess `['example']` works too

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1537,4 +1537,33 @@ describe('surveys', () => {
...
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(false)
+        })
+
+        it('can be disabled by config despite results of onRemoteConfig', () => {
+            surveys['_instance'].config.disable_surveys = true
+            surveys.onRemoteConfig({
+                surveys: false,
```
pauldambra:
```suggestion
                surveys: ['example'],
```

packages/browser/src/__tests__/surveys.test.ts
pauldambra:
```suggestion
                surveys: ['example'],
```

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1537,4 +1537,33 @@ describe('surveys', () => {
...
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(false)
+        })
+
+        it('can be disabled by config despite results of onRemoteConfig', () => {
+            surveys['_instance'].config.disable_surveys = true
+            surveys.onRemoteConfig({
+                surveys: ['example'],
+            } as Partial<RemoteConfig> as RemoteConfig)
+            expect(surveys['_isSurveysEnabled']).toBe(false)
```
pauldambra, line 1566: so, this should fail as things are now
because surveys will be enabled
because there was a response
onRemoteConfig is called regardless of other configuration so this code needs to be safe

pauldambra: "needs"
given i don't know that this actually bubbles up to the user

packages/browser/src/__tests__/surveys.test.ts
marandaneto: yes theres no validation at this point

pauldambra:
https://github.com/PostHog/posthog-js/pull/2279 in the description :)

PR #2284, opened 2025-09-04
fix: order of operations when replay does pageview fallback

pauldambra:
when replay is running and pageview capture is disabled
we have simple url tracking which gets emitted to the replay events table as metadata

we say 

* when emitting an event that is not a meta event
 * if rrwebEvent url is not the same as privately tracked `lastUrl`
 * then emit a `urlChanged` custom event that holds the url
 * and then update the privately tracked `lastUrl`

but if the custom event is emitted before the tracking variable is updated then we'll emit _another_ url 
and that might cause us to...

etc

instead let's update the privately tracked `lastUrl` and then emit the rrweb custom event

### DIFF
--- a/.changeset/stale-taxes-unite.md
+++ b/.changeset/stale-taxes-unite.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: order of operations when replay does pageview fallback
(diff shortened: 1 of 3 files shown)


eli-r-ph (approved):
Not a ton of context here, but the change is simple and appears safe, and rationale makes sense to me 👍

PR #2294, opened 2025-09-07
feat: add a redux logger to posthog-js

pauldambra:
this has come up a couple of times at least

* link1
* link2

the main hit for `redux logger` is this abandoned library https://web.archive.org/web/20201023122130/https://github.com/LogRocket/redux-logger/issues/314

so let's add a simple logger that we can pass a logger into
currently it defaults to console.log

but you could pass `posthog.sessionRecording.log` or `posthog.capture`


mariusandra (approved):
✅

However... this middleware will be on a very hot path. I wonder if it might be worth defaulting `maskState` and `maskAction` to `undefined`, and if either or neither are present, exiting early. This could make things just a bit faster? Though this might be premature optimization.

All redux actions usually happen very synchronously in one thread... so I have a hard time imagining one taking > 1500ms. Perhaps that part can even be skipped (less to maintain going forward, etc)?

PR #2296, opened 2025-09-08
feat: switch replay default to lazy loading

pauldambra:
we want to move replay to default lazy loading because it is heading towards a fifth of the bundle

but despite it working in our testing it is very hard to cover all angles for bugs

so, let's release it as the default without getting the bundle savings so users can switch back to eager loading if we discover a bug

and take another little safe step to that glorious future

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -526,10 +526,10 @@ export class PostHog {
         this.siteApps?.init()
 
         if (!startInCookielessMode) {
-            if (this.config.__preview_lazy_load_replay) {
-                this.sessionRecording = new SessionRecordingWrapper(this)
-            } else {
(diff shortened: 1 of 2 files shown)


packages/browser/src/extensions/replay/sessionrecording-wrapper.ts
```diff
@@ -35,9 +35,6 @@ export class SessionRecordingWrapper {
      * once a flags response is received status can be disabled, active or sampled
      */
     get status(): SessionRecordingStatus {
-        if (!this._isRecordingEnabled) {
-            return DISABLED
-        }
```
pauldambra, line 40: glad tests caught this! that wasn't right, we're not disabled until _after_ we've had a chance to get remote config

PR #2299, opened 2025-09-09
fix: zod types not infered

k11kirky:
## Problem

Several users on X have raised an issue where adding posthog llma broke their type inference using zod schemas:
https://x.com/simplyalec/status/1965213956089491696

## Changes

Added overrides - You can now call client.responses.parse({ text: { format: z.object({...}) }, ... }) and get ParsedResponse<z.infer<typeof schema>> without manual generics.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config


packages/ai/src/openai/index.ts
```diff
@@ -575,7 +593,7 @@ export class WrappedResponses extends Responses {
         }
       )
 
-      return wrappedPromise as APIPromise<ParsedResponse<ParsedT>>
+      return wrappedPromise as APIPromise<ParsedResponse<any>>
```
k11kirky: will this affect user typings? 

No.
TypeScript uses the overload signatures for callers; the implementation signature (with any) is not exposed.
In the emitted d.ts, only the overloads appear, so call sites still get the correct ParsedResponse<...> type (Zod-inferred, explicit ParsedT, or fallback).
The as APIPromise<ParsedResponse<any>> is only to satisfy the implementation and won’t affect consumer typings.

PR #2300, opened 2025-09-09
feat: add React helpers

daibhin:
## Problem

https://posthoghelp.zendesk.com/agent/tickets/37142 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/39231)

## Changes

Customer asked for a helper that handles the new error callbacks added in React 19 (https://react.dev/reference/react-dom/client/createRoot#parameters)

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

### DIFF
--- a/packages/react/src/helpers/error-helpers.ts
+++ b/packages/react/src/helpers/error-helpers.ts
@@ -1,0 +1,14 @@
+import type { ErrorInfo } from 'react'
+import { PostHog } from '../context'
+
+export const setupReactErrorHandler = (
+    client: PostHog,
+    callback?: (event: any, error: any, errorInfo: ErrorInfo) => void
+) => {
+    return (error: any, errorInfo: ErrorInfo): void => {
+        const event = client.captureException(error)
+        if (callback) {
+            callback(event, error, errorInfo)
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


packages/react/src/helpers/error-helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+import { PostHog } from '../context'
+
+export const setupReactErrorHandler = (
+    client: PostHog,
+    callback?: (event: any, error: any, errorInfo: ErrorInfo) => void
+) => {
+    return (error: any, errorInfo: ErrorInfo): void => {
+        const event = client.captureException(error)
+        if (callback) {
+            callback(event, error, errorInfo)
```
ablaszkiewicz, line 12: Shouldn't we pass original `event` from original callback from line 6?

ablaszkiewicz (approved):

packages/react/src/helpers/error-helpers.ts
daibhin: The callback would be something the user provides and the the type `(event: any, error: any, errorInfo: ErrorInfo) => void` simply means that they should expect to receive the captured PostHog event, plus the original error and errorInfo

daibhin: It's a good point though, I can probably type the event better than `any`

ablaszkiewicz: Oops I misunderstood the code

PR #2302, opened 2025-09-09
Chore: Add docs comments and code for react native specs

gewenyu99:
## Problem

We need comments for react native specs, see: https://github.com/PostHog/posthog.com/pull/12720

Similar to [these references](https://posthog.com/docs/references/posthog-js) for posthog-js

Would love your help to make sure these comments are accurate

## Changes

Adds docs comment. Adds a special case to include specific methods that aren't part of the posthog class like the posthogprovider

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## How to test these changes

This PR uses the generated specs.: https://github.com/PostHog/posthog.com/pull/12720

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -235,6 +235,31 @@ export abstract class PostHogCoreStateless {
     return this._events.on(event, cb)
   }
 
+  /**
+   * Enables or disables debug mode for detailed logging.
+   *
+   * @remarks
+   * Debug mode logs all PostHog calls to the console for troubleshooting.
+   * This is useful during development to understand what data is being sent.
+   *
+   * {@label Initialization}
+   *
+   * @example
+   * ```js
+   * // enable debug mode
+   * posthog.debug(true)
+   * ```
+   *
+   * @example
+   * ```js
+   * // disable debug mode
+   * posthog.debug(false)
+   * ```
+   *
+   * @public
+   *
+   * @param {boolean} [debug] If true, will enable debug mode.
+   */
   debug(enabled: boolean = true): void {
     this.removeDebugCallback?.()
 
@@ -923,7 +948,7 @@ export abstract class PostHogCoreStateless {
   }
 
   /**
-   * Flushes the queue
+   * Flushes the queue of pending events.
    *
    * This function will return a promise that will resolve when the flush is complete,
    * or reject if there was an error (for example if the server or network is down).
@@ -932,13 +957,19 @@ export abstract class PostHogCoreStateless {
    *
    * It's recommended to do error handling in the callback of the promise.
    *
+   * {@label Initialization}
+   *
    * @example
+   * ```js
+   * // flush with error handling
    * posthog.flush().then(() => {
    *   console.log('Flush complete')
    * }).catch((err) => {
    *   console.error('Flush failed', err)
    * })
+   * ```
    *
+   * @public
    *
    * @throws PostHogFetchHttpError
    * @throws PostHogFetchNetworkError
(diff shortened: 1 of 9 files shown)


packages/core/src/index.ts
```diff
@@ -1322,7 +1371,17 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
 
   /**
-   * * @returns {string} The stored session ID for the current session. This may be an empty string if the client is not yet fully initialized.
+   * Returns the current session_id.
+   *
+   * @remarks
+   * This should only be used for informative purposes.
+   * Any actual internal use case for the session_id should be handled by the sessionManager.
+   *
+   * {@label Session replay}
```
marandaneto: this is also used for session duration https://posthog.com/tutorials/session-metrics
not only session replay

> The stored session ID for the current session. This may be an empty string if the client is not yet fully initialized.

the new comment is removing ^ but i dont think we should

packages/react-native/scripts/generate-docs.js
```diff
@@ -0,0 +1,43 @@
...
+// Node-specific configuration
+const REACT_NATIVE_SPEC_INFO = {
+    id: 'posthog-react-native',
+    title: 'PostHog React Native SDK',
+    description: 'PostHog React Native SDK allows you to capture events and send them to PostHog from your React Native applications.',
+    slugPrefix: 'posthog-react-native',
+    specUrl: 'https://github.com/PostHog/posthog-js'
+};
+
+// Node-specific type examples (can be customized as needed)
```
marandaneto: we should use react native specific configuration then

packages/react-native/src/PostHogProvider.tsx
```diff
@@ -28,6 +42,66 @@ function PostHogNavigationHook({
   return null
 }
 
+/**
+ * PostHogProvider is a React component that provides PostHog functionality to your React Native app. You can find all configuration options in the [React Native SDK docs](https://posthog.com/docs/libraries/react-native#configuration-options).
+ * 
+ * Autocapturing navigation requires further configuration. See the [React Native SDK navigation docs](https://posthog.com/docs/libraries/react-native#capturing-screen-views) 
+ * for more information about autocapturing navigation.
+ *
+ * This is the recommended way to set up PostHog for React Native. This utilizes the Context API to pass the PostHog client around, enable autocapture, and ensure that the queue is flushed at the right time.
```
marandaneto: > and ensure that the queue is flushed at the right time.

i think thats wrong, the provider has nothing to do with the flushing lifecycle

packages/core/src/index.ts
gewenyu99: Good catch 😆

packages/react-native/scripts/generate-docs.js
gewenyu99: Oops I need to update these

packages/react-native/src/PostHogProvider.tsx
gewenyu99: Screenshot 2025-09-10 at 3 24 08 PM

I got this from the react native SDK docs. https://posthog.com/docs/libraries/react-native

I'll confirm in slack

marandaneto: answered on slack

gewenyu99: Updated
Also[ PR for docs repo](https://github.com/PostHog/posthog.com/pull/12755)

marandaneto (approved):

PR #2310, opened 2025-09-11
feat: Add config option for partitioned cookies

robbie-c:
## Problem

Our cookies are not partioned.

This means that if the user visits posthog.com, then visits customer.com, when the posthog-js library on customer.com calls api.posthog.com, the browser will attempt to include those cookies. This can trip up some customers with strict cookie rules.

See this ticket: https://posthoghelp.zendesk.com/agent/tickets/37096 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/39016)

## Changes

We can use used Partitioned cookies (see https://developer.mozilla.org/en-US/docs/Web/Privacy/Guides/Privacy_sandbox/Partitioned_cookies)

This would mean that cookies set on posthog.com are stored in a different cookie jar to those set on customer.com.

This doesn't affect cross-subdomain tracking, so posthog.com and app.posthog.com should still be able to share cookies.

This is pretty hard to test, I am relying on rolling this out to posthog.com first, and then making sure the issue in the above ticket is fixed.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-persistence.ts
+++ b/packages/browser/src/posthog-persistence.ts
@@ -56,6 +56,7 @@ export class PostHogPersistence {
     private readonly _name: string
     _disabled: boolean | undefined
     private _secure: boolean | undefined
+    private _partitioned: boolean | undefined
     private _expire_days: number | undefined
     private _default_expiry: number | undefined
     private _cross_subdomain: boolean | undefined
@@ -160,6 +161,7 @@ export class PostHogPersistence {
             this._expire_days,
             this._cross_subdomain,
             this._secure,
+            this._partitioned,
             this._config.debug
         )
     }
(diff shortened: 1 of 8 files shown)


packages/browser/src/posthog-persistence.ts
```diff
@@ -361,6 +364,14 @@ export class PostHogPersistence {
         }
     }
 
+    set_partitioned(partitioned: boolean): void {
+        if (partitioned !== this._partitioned) {
+            this._partitioned = partitioned
+            this.remove()
+            this.save()
+        }
+    }
```
pauldambra, line 373: feels like this repeated pattern could be squashed 
but nit picking :)

pauldambra (approved):
defaults to off so LGTM

PR #2314, opened 2025-09-11
fix(surveys): load surveys in cookieless mode on consent

ordehi:
## Problem

Surveys were not loading in cookieless mode even when consent was given using the cookieless `on_reject` option

## Changes

Updates the cookieless guard clause in `surveys.loadIfEnabled()` to check `this._instance.consent.isOptedOut()` so it's skipped if the user has opted in.

Updates `opt_in_capturing` to call `loadIfEnabled()` if we're in cookieless `on_reject` mode

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/olive-areas-work.md
+++ b/.changeset/olive-areas-work.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fixed a bug that prevented surveys from loading in cookieless mode using the on_reject option. Surveys now correctly initialize when consent is given.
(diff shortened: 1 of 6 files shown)


lucasheriques (approved):
looks great, thanks for fixing. before merging though, remove the changes on `pnpm-lock.yaml`

I don't think they should be there, since you didn't change any packages on `package.json` so weird they did 🤔

PR #1055, opened 2024-03-05
feat: Refactor request logic

benjackwhite:
## Changes

Our code paths around making api requests is messy to say the least. Makes it hard to reason on, hard to make changes and includes a lot of spaghetti code that inflates the bundle size.

* Simplify everything down to a single request function that handles choosing the right available transport.

## Checklist
- [ ] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [ ] Accounted for the impact of any changes across different browsers
- [ ] Accounted for backwards compatibility of any changes (no breaking changes in posthog-js!)

PR #2318, opened 2025-09-12
feat: Add option for xhr credentials and to disable sendBeacon

robbie-c:
## Problem

This ticket: https://posthoghelp.zendesk.com/agent/tickets/37096 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/39016)

As part of trying to debug this, I noticed that we send credentials (cookies) with xhr requests. We should probably not do this. Whilst Chrome filters these cookies out, it means that sites try to send cookies from posthog.com along with xhr requests.

From looking at the capture code, I couldn't see anywhere where we used cookies.

I also asked @benjackwhite if he knew why it was added here https://github.com/PostHog/posthog-js/pull/1055/files#diff-b1a2ab03a3672e105afeaab6615ad625365cd8459d592a840d925f132af62b35R107 and he didn't

We should be safe to remove it, but just in case, let's remove it for ourselves first.

Additionally, `sendBeacon` doesn't even have an option to remove credentials, and they are sent by default. IMO, this is insane, but I didn't write the spec. I've added an option to remove sendBeacon calls, which I'll suggest this user can try out

## Changes

Add options to disable credentials on xhr and to disable sendBeacon

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/types.ts
+++ b/packages/browser/src/types.ts
@@ -992,6 +992,17 @@ export interface PostHogConfig {
      * */
     __preview_eager_load_replay?: boolean
 
+    /**
+     * Prevents posthog-js from using the `navigator.sendBeacon` API to send events.
+     * Enabling this option may hurt the reliability of sending $pageleave events
+     */
+    __preview_disable_beacon?: boolean
+
+    /**
+     * Disables sending credentials when using XHR requests.
+     */
+    __preview_disable_xhr_credentials?: boolean
+
     // ------- RETIRED CONFIGS - NO REPLACEMENT OR USAGE -------
 
     /**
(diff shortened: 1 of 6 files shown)


packages/browser/src/types.ts
```diff
@@ -992,6 +992,17 @@ export interface PostHogConfig {
      * */
     __preview_eager_load_replay?: boolean
 
+    /**
+     * Prevents posthog-js from using the `navigator.sendBeacon` API to send events.
+     * Enabling this option may hurt the reliability of sending $pageleave events
+     */
+    __preview_disable_beacon?: boolean
```
rafaeelaudibert, line 999: I don't see you using this anywhere?

rafaeelaudibert (changes requested):

packages/browser/src/types.ts
robbie-c: Good catch, greptile got me there too!

rafaeelaudibert (approved):

PR #2319, opened 2025-09-12
chore(err): move error processing to core

hpouillot:
## Problem

- Some code is duplicated between node and browser
- Preparation for react native integration

## Changes

- move error processing to core package
- make error conversion consistent between platform

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/error-tracking/error-properties-builder.ts
+++ b/packages/core/src/error-tracking/error-properties-builder.ts
@@ -71,7 +71,7 @@ export class ErrorPropertiesBuilder {
   private applyChunkIds(frames: StackFrame[], chunkIdMap?: ChunkIdMapType): StackFrame[] {
     return frames.map((frame) => ({
       ...frame,
-      chunkId: frame.filename && chunkIdMap ? chunkIdMap[frame.filename] : undefined,
+      chunk_id: frame.filename && chunkIdMap ? chunkIdMap[frame.filename] : undefined,
     }))
   }
 


packages/core/src/error-tracking/coercers/object-coercer.ts
```diff
@@ -0,0 +1,102 @@
...
+        synthetic: true,
+      }
+    }
+  }
+
+  getType(err: Record<string, unknown>): string {
+    return isEvent(err) ? err.constructor.name : 'Error'
+  }
+
+  getValue(err: object) {
```
hpouillot, line 30: I took the node version here which is more descriptive (constructor name)

packages/core/src/error-tracking/coercers/object-coercer.ts
```diff
@@ -0,0 +1,102 @@
...
+        type: this.getType(candidate),
+        value: this.getValue(candidate),
+        stack: ctx.syntheticException?.stack,
+        level: this.isSeverityLevel(candidate.level) ? candidate.level : 'error',
+        synthetic: true,
+      }
+    }
+  }
+
+  getType(err: Record<string, unknown>): string {
```
hpouillot, line 26: I took the browser version here which is more descriptive (node only returns the Error type)

packages/core/src/error-tracking/utils.ts
daibhin, line 1: Should we be removing this file (and the others) from the browser & node packages?

packages/core/src/error-tracking/parsers/winjs.ts
daibhin, line 1: Should we just remove this? Doesn't look like it's referenced anywhere

packages/core/src/error-tracking/parsers/react-native.ts
daibhin, line 1: Should this have content?

packages/core/src/error-tracking/parsers/opera.ts
daibhin, line 1: Same here... are these used?

packages/core/src/error-tracking/parsers/base.ts
```diff
@@ -0,0 +1,29 @@
...
+export const OPERA10_PRIORITY = 10
+export const OPERA11_PRIORITY = 20
+export const CHROME_PRIORITY = 30
+export const WINJS_PRIORITY = 40
+export const GECKO_PRIORITY = 50
+
+export function createFrame(filename: string, func: string, lineno?: number, colno?: number): StackFrame {
+  const frame: StackFrame = {
+    // TODO: should be a variable here
+    platform: 'web:javascript',
```
daibhin, line 14: I'm sure this is on your radar

packages/core/src/error-tracking/coercers/dom-exception-coercer.ts
```diff
@@ -0,0 +1,38 @@
+import { CoercingContext, ErrorTrackingCoercer, ExceptionLike } from '@/error-tracking/types'
+import { isBuiltin, isString } from '@/utils'
+
+export class DOMExceptionCoercer implements ErrorTrackingCoercer<DOMException> {
```
daibhin, line 4: Should this be used outside the spec?

packages/core/src/error-tracking/error-properties-builder.ts
```diff
@@ -0,0 +1,156 @@
...
+  StackLineParser,
+  ParsingContext,
+  ChunkIdMapType,
+  Mechanism,
+  ParsedException,
+} from './types'
+
+const MAX_CAUSE_RECURSION = 4
+
+export class ErrorPropertiesBuilder {
```
daibhin, line 22: I'm assuming this gets implemented in each package?

daibhin (approved):
Left a couple of questions. Very nice cleanup. I see where it's going directionally. I'm guessing you're going to add to the core in this PR and then strip out the implementations from either package

packages/core/src/error-tracking/parsers/base.ts
hpouillot: It works for now, as createFrame method is only used in browser stacktrace parsers but it's a bit risky. I will fix once we start integrating this in other packages

packages/core/src/error-tracking/parsers/react-native.ts
hpouillot: Hum forgot to remove this one, I will keep it for now and add the content when I'll integrate this into react-native package.

packages/core/src/error-tracking/utils.ts
hpouillot: Yes, I wanted to split PRs, this one if to agree on a common behavior for all platforms. I will follow-up with implementation PRs (and remove unused files)

packages/core/src/error-tracking/parsers/winjs.ts
hpouillot: I move this parser but it was not used anywhere but in general, coercers, parsers and modifiers will be passed to the ErrorPropertiesBuilder constructor inside the platform specific packages, that's why it's not used for now.

packages/core/src/error-tracking/parsers/opera.ts
hpouillot: same than winjs but this one might be useful.

packages/core/src/error-tracking/coercers/dom-exception-coercer.ts
hpouillot: same than parsers they will be imported inside patform packages.

PR #2330, opened 2025-09-17
chore(posthog-js): plug core error tracking processing

hpouillot:
## Changes

- plug error tracking processing from @posthog/core

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/browser/src/__tests__/extensions/exception-autocapture/error-conversion.test.ts
```diff
@@ -159,43 +171,22 @@ describe('Error conversion', () => {
...
+        const errorProperties: ErrorProperties = errorToProperties(
+            { event: pre as unknown as PromiseRejectionEvent },
+            { handled: false }
+        )
         expect(Object.keys(errorProperties)).toHaveLength(2)
-        expect(errorProperties.$exception_list[0].type).toEqual('UnhandledRejection')
+        expect(errorProperties.$exception_list[0].type).toEqual('PromiseRejectionEvent')
         expect(errorProperties.$exception_list[0].value).toEqual(
-            'Non-Error promise rejection captured with value: My house is on fire'
+            'PromiseRejectionEvent captured as exception with keys: isTrusted, promise, reason'
```
ablaszkiewicz, line 186: Why did we choose to display these keys instead of a `reason`? I'm not saying that's bad. Just curious.

packages/core/src/error-tracking/coercers/utils.ts
```diff
@@ -4,3 +4,30 @@ export function truncate(str: string, max: number = 0): string {
...
+
+  for (let i = keys.length; i > 0; i--) {
+    const serialized = keys.slice(0, i).join(', ')
+    if (serialized.length > maxLength) {
+      continue
+    }
+    if (i === keys.length) {
+      return serialized
+    }
+    return serialized.length <= maxLength ? serialized : `${serialized.slice(0, maxLength)}...`
```
ablaszkiewicz, line 29: I feel like `serialized.length <= maxLength` will always be true at that point because if it wasn't, then we would have `continue`d in line `24`.

All in all this for loop is quite confusing to me.

packages/browser/src/extensions/exception-autocapture/stack-trace.ts
daibhin, line 1: 😍

packages/browser/src/extensions/exception-autocapture/index.ts
```diff
@@ -139,9 +138,10 @@ export class ExceptionObserver {
         this.startIfEnabled()
     }
 
-    captureException(errorProperties: ErrorProperties) {
+    captureException(input: unknown, hint?: { handled?: boolean; syntheticException?: Error }) {
+        const errorProperties = this._instance.exceptions.buildProperties(input, hint)
         const posthogHost = this._instance.requestRouter.endpointFor('ui')
-
+        // @ts-expect-error: Do we still need this?
         errorProperties.$exception_personURL = `${posthogHost}/project/${
```
daibhin: We can probably remove this... we have the person info elsewhere so not really sure why anyone would use this

packages/browser/src/__tests__/extensions/exception-autocapture/error-conversion.test.ts
```diff
@@ -128,10 +140,10 @@ describe('Error conversion', () => {
...
         }
 
-        expect(Object.keys(errorProperties)).toHaveLength(3)
-        expect(errorProperties.$exception_list[0].type).toEqual('dom-exception')
-        expect(errorProperties.$exception_list[0].value).toEqual('oh no disaster')
-        expect(errorProperties.$exception_DOMException_code).toEqual('0')
+        expect(Object.keys(errorProperties)).toHaveLength(2)
+        expect(errorProperties.$exception_list[0].type).toEqual('DOMException')
+        expect(errorProperties.$exception_list[0].value).toEqual('dom-exception: oh no disaster')
+        // expect(errorProperties.$exception_DOMException_code).toEqual('0')
```
daibhin:
```suggestion
```

daibhin (approved):

packages/browser/src/__tests__/extensions/exception-autocapture/error-conversion.test.ts
hpouillot: It is a bit confusing but in jest env the PromiseRejectionEvent doesn't exist, it is considered as an normal event. This part is tested in playwright here: https://github.com/PostHog/posthog-js/pull/2330/files#diff-370b635963a4be88b36a0e81f47208e0d69cc973b82133dcd90ec581e16b50c9R105

packages/core/src/error-tracking/coercers/utils.ts
hpouillot: It is taken from previous version but I agree that it is a bit confusing. I will leave for now, but we can clean it later.

packages/browser/src/__tests__/extensions/exception-autocapture/error-conversion.test.ts
ablaszkiewicz: Makes sense, thanks

PR #2344, opened 2025-09-19
fix: mark renderSurvey as deprecated

lucasheriques:
## Changes

mark `renderSurvey` as deprecated, since `displaySurvey` is released and it's an improved version of it

### Libraries affected

- [x] posthog-js (web)

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -3326,11 +3326,6 @@ const add_dom_loaded_handler = function () {
 }
 
 export function init_from_snippet(): void {
-    if (assignableWindow['posthog'] && assignableWindow['posthog'].__loaded) {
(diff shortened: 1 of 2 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -3324,6 +3326,11 @@ const add_dom_loaded_handler = function () {
 }
 
 export function init_from_snippet(): void {
+    if (assignableWindow['posthog'] && assignableWindow['posthog'].__loaded) {
+        logger.warn('PostHog script loaded multiple times, skipping re-initialization')
+        return
+    }
```
marandaneto: this should not be here

lucasheriques: oops removed

marandaneto (approved):

PR #2345, opened 2025-09-19
feat: surveys matching on event properties

lucasheriques:
## Problem

Related to https://github.com/PostHog/posthog/issues/30080

This PR enables it for **web-only surveys** (i.e. using `posthog-js`). SDK-only (logic). We still need the UI to actually be able to edit surveys on the dashboard.

Though with this merged, it can already be used via the API.

## Changes

Changes the `onEvent` logic on survey event matching to also check for property matching if the survey conditions has that value. Supports the traditional match types we usually do (exact, contains, regex etc).

Tested with both e2e tests and unit tests.

Most of the lines changed are tests. The logic itself is pretty small

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-surveys-types.ts
+++ b/packages/browser/src/posthog-surveys-types.ts
@@ -199,6 +199,13 @@ export interface Survey {
             repeatedActivation?: boolean
             values: {
                 name: string
+                /** Property filters for event matching */
+                propertyFilters?: {
+                    [propertyName: string]: {
+                        values: string[]
+                        operator: PropertyMatchType
+                    }
+                }
             }[]
         } | null
         actions: {
--- a/packages/browser/src/utils/survey-event-receiver.ts
+++ b/packages/browser/src/utils/survey-event-receiver.ts
@@ -126,58 +126,61 @@ export class SurveyEventReceiver {
                     this._updateActivatedSurveys(existingActivatedSurveys)
                 }
             }
-        } else {
-            if (this._eventToSurveys.has(event)) {
-                logger.info('survey event name matched', {
-                    event,
-                    eventPayload,
-                    surveys: this._eventToSurveys.get(event),
-                })
 
-                let surveysToCheck: Survey[] = []
+            return
+        }
 
-                this._instance?.getSurveys((surveys) => {
-                    surveysToCheck = surveys.filter((survey) => this._eventToSurveys.get(event)?.includes(survey.id))
-                })
+        // if the event is not in the eventToSurveys map, nothing else to do
+        if (!this._eventToSurveys.has(event)) {
+            return
+        }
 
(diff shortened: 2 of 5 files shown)


packages/browser/src/posthog-surveys-types.ts
```diff
@@ -199,6 +199,13 @@ export interface Survey {
             repeatedActivation?: boolean
             values: {
                 name: string
+                /** Property filters for event matching */
+                propertyFilters?: {
+                    [propertyName: string]: {
+                        values: string[]
+                        operator: PropertyMatchType
```
marandaneto, line 206: i think in this way you'd be able to select a few props and always the same operator, right?
i think it'd make sense to choose different operators for different values
eg device_type operator equal Mobile and os_version higher than 10, more flexible, wdyt?

lucasheriques: not really, since this is an object

for example, in this case you shared, it would be something like this:

```ts
propertyFilters: {
	'device_type': { values: ['Mobile'], operator: 'icontains' },
	`os_version': { values: ['11', '12', '13'], operator: 'exact' }
}
```

ideally `os_version` would use number comparisons (bigger than / lower than), but it's not supported by `propertyComparisons` and I don't think it makes sense to extend it as a part of this PR, but we can create an issue and see if they will be requests for it

lucasheriques: [here's a test doing this behavior](https://github.com/PostHog/posthog-js/pull/2345/files/22109601298a3031eca9c8de9ecfdabfe85187df#diff-0b81ddf751a48d177860308e85f9e287078c0b91ca6c3ad8f6211d97fdb7115dR312-R334)

marandaneto: ah ok misunderstood the definition then, all good

packages/browser/src/utils/survey-event-receiver.ts
```diff
@@ -127,15 +126,59 @@ export class SurveyEventReceiver {
...
+            }
+
+            // if there are no property filters, it means we're only matching on event name
+            if (!eventToCheck.propertyFilters) {
+                return true
+            }
+
+            return Object.entries(eventToCheck.propertyFilters).every(([propertyName, filter]) => {
+                const eventPropertyValue = eventPayload?.properties?.[propertyName]
+                if (isUndefined(eventPropertyValue) || isUndefined(eventPropertyValue)) {
```
marandaneto: 2 times the same check

lucasheriques: fixed, was supposed to be `isNull`

marandaneto (approved):

PR #2347, opened 2025-09-20
feat: allow ignoring rageclick by css selector

pauldambra:
sometimes we want to allow autocapture on an element
but we know we don't want rageclicks

for example "previous" and "next" buttons are often clicked quickly and show up as rageclicks when they shouldn't

let's support a new default of `.ph-no-rageclick` to mirror the existing `.ph-no-capture`
and allow a custom list of selectors - since some sites will have an existing set of selectors that already make sense

when `.ph-no-rageclick` or the configured selector ignorelist is matched we don't capture a rageclick

### DIFF
--- a/packages/browser/src/autocapture-utils.ts
+++ b/packages/browser/src/autocapture-utils.ts
@@ -1,7 +1,7 @@
-import { AutocaptureConfig, Properties } from './types'
+import { AutocaptureConfig, PostHogConfig, Properties } from './types'
 import { each, entries } from './utils'
 
-import { isNullish, isString, isUndefined, isArray } from '@posthog/core'
+import { isNullish, isString, isUndefined, isArray, isBoolean } from '@posthog/core'
 import { logger } from './utils/logger'
 import { window } from './utils/globals'
 import { isDocumentFragment, isElementNode, isTag, isTextNode } from './utils/element-utils'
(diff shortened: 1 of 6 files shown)


pauldambra:
closes https://github.com/PostHog/posthog/issues/22257

packages/browser/src/autocapture-utils.ts
```diff
@@ -159,8 +151,72 @@ export function getParentElement(curEl: Element): Element | false {
...
+    if (selectorIgnoreList === false) {
+        return false
+    }
+
+    const { targetElementList } = getElementAndParentsForElement(el, false)
+    if (checkIfElementsMatchCSSSelector(targetElementList, selectorIgnoreList)) {
+        // we don't capture if we match the ignore list
+        return false
+    }
+    return true
```
TueHaulund: Matter of taste but this could all be shortened to:

`return !checkIfElementsMatchCSSSelector(targetElementList, selectorIgnoreList)`

packages/browser/src/autocapture-utils.ts
```diff
@@ -159,8 +151,72 @@ export function getParentElement(curEl: Element): Element | false {
...
+}
+
+const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false | true | undefined) => {
+    if (!window || cannotCheckForAutocapture(el)) {
+        return { parentIsUsefulElement: false, targetElementList: [] }
+    }
+
+    let parentIsUsefulElement = false
+    const targetElementList: Element[] = [el]
+    let parentNode: Element | boolean = true
```
TueHaulund: Could we move this definition inside the loop (line 201) and make it `const` ?

PR #2348, opened 2025-09-22
chore(node): node package cleaning

hpouillot:
## Problem

posthog-node requires node >= 20, but the syntax we generate targets node10.

## Changes

- distribute node package as unbundled
- switch from rollup to rslib
- remove unused dependencies (commander, mitata)
- change extension to mjs when esm
- type check tests folder
- clean imports in examples
- clean imports in exception-autocapture

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/node/package.json
```diff
@@ -53,29 +51,29 @@
...
+        "default": "./dist/entrypoints/index.edge.js"
       },
       "node": {
-        "import": "./dist/node/index.mjs",
-        "require": "./dist/node/index.cjs",
-        "default": "./dist/node/index.mjs"
+        "import": "./dist/entrypoints/index.node.mjs",
+        "require": "./dist/entrypoints/index.node.js",
+        "default": "./dist/entrypoints/index.node.js"
       },
```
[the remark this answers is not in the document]
hpouillot: nop

packages/node/package.json
```diff
@@ -9,15 +9,15 @@
...
     "test:unit": "jest",
-    "build": "rollup -c",
-    "dev": "rollup -cw",
+    "build": "rslib build",
+    "dev": "rslib build -w",
     "prepublishOnly": "pnpm lint && pnpm test:unit && pnpm build",
     "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz",
-    "gen-specs": "pnpm exec tsc && pnpm exec api-extractor run --config ./api-extractor.json --local",
-    "parse-specs": "node scripts/generate-docs.js"
+    "gen-specs": "pnpm build && pnpm exec api-extractor run --config ./api-extractor.jsonc --local",
```
hpouillot, line 20: @gewenyu99 I slightly changed the script here as the final API would come from the build task. It would also be interesting to standardise the way we generate specifications, this way we could create task dependencies inside turbo.json. Is there a github workflow to export those specifications ?

packages/node/api-extractor.json
```diff
@@ -67,7 +67,7 @@
    *
    *   "bundledPackages": [ "@my-company/*" ],
    */
-  "bundledPackages": ["@posthog/core"],
```
hpouillot, line 70: @gewenyu99 @posthog/core is not bundled anymore, is it safe to remove it here ?

packages/node/package.json
```diff
@@ -53,29 +51,29 @@
...
     ".": {
-      "types": "./dist/index.d.ts",
+      "types": "./dist/entrypoints/index.node.d.ts",
       "edge": {
-        "import": "./dist/edge/index.mjs",
-        "require": "./dist/edge/index.cjs",
-        "default": "./dist/edge/index.mjs"
+        "import": "./dist/entrypoints/index.edge.mjs",
+        "require": "./dist/entrypoints/index.edge.js",
+        "default": "./dist/entrypoints/index.edge.js"
```
daibhin, line 61: Any particular reason to move to this structure?

hpouillot: We don't bundle the package anymore so the outputs will mirror the source structure but we could probably rethink it. Maybe a "src/index.edge.ts" and "src/index.node.ts" would be clearer

packages/node/api-extractor.json
gewenyu99: 👍 I think it should be fine if it's not bundled anymore. The spec will still build, I'll investigate it if breaks anything, but if it's not bundled it shouldn't change anything..

packages/node/package.json
gewenyu99: 👍 Yeah I have an RFC on this, I want to (soon) automate all this in CI, this is a step the right way. Can I pick up the chat separately with you on this when the time comes?

https://github.com/PostHog/meta/pull/368

hpouillot: Nice, makes sense ! Sure, let me know if you need help to automate this.

gewenyu99: 100% will need help 😉 I'll ping you separately

packages/core/tsconfig.json
```diff
@@ -1,20 +1,20 @@
 {
   "extends": "@posthog-tooling/tsconfig-base",
   "compilerOptions": {
-    "rootDir": "./src",
+    "rootDir": ".",
     "outDir": "dist",
-    "module": "CommonJS",
-    "moduleResolution": "node",
+    "module": "Node16",
+    "moduleResolution": "Node16",
```
marandaneto: make sure to test the react native target with these changes, i remember having issues with that in the past

hpouillot: I just tested it on the expo example and it works for ios but not with the web target. I'm getting "PostHog: No storage available. ..." but I have the same error on main

marandaneto: https://posthog.com/docs/libraries/react-native#react-native-web-and-macos
the expo53 example is set up for mobile only since it uses the expo file provider
when i want to test on web, i usually just install the other file provider that has web support and comment out the other, so this error is ok

PR #2353, opened 2025-09-23
fix: start request queue when opting in during session

ordehi:
## Problem

When calling `posthog.opt_in_capturing()` the request queue stays paused, so events are not captured until the user reloads/loads a new page.

## Changes

Added `this._start_queue_if_opted_in()` in the workflow that runs when calling `posthog.opt_in_capturing()`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/cookieless.test.ts
+++ b/packages/browser/src/__tests__/cookieless.test.ts
@@ -311,31 +311,30 @@ describe('cookieless', () => {
             expect(posthog.sessionRecording).toBeFalsy()
         })
 
-        it('should enable request queue immediately after opt_in_capturing in cookieless mode', async () => {
+        it('should restart autocapture after opt_in_capturing in cookieless mode', async () => {
             const { posthog } = await setup({
                 cookieless_mode: 'on_reject',
-                request_batching: true,
             })
 
-            // Mock the request queue enable method
-            const mockEnable = jest.fn()
-            const originalEnable = posthog._requestQueue?.enable
-            if (posthog._requestQueue) {
-                posthog._requestQueue.enable = mockEnable
+            // Mock the autocapture startIfEnabled method
+            const mockStartIfEnabled = jest.fn()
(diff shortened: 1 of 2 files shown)


rafaeelaudibert:
Requesting a review from @robbie-c specifically since he understands cookieless better than the rest of us

packages/browser/src/__tests__/cookieless.test.ts
```diff
@@ -310,5 +310,32 @@ describe('cookieless', () => {
             expect(beforeSendMock.mock.calls[3][0].properties.$cookieless_mode).toEqual(true)
             expect(posthog.sessionRecording).toBeFalsy()
         })
+
+        it('should restart autocapture after opt_in_capturing in cookieless mode', async () => {
```
robbie-c, line 314: I'm confused about whether the bug is related to autocapture not being restarted on opt-in (which is what this is testing), or whether it's related to the request queue not starting (which is the title of this PR).

Is there a link to a ticket somewhere?

packages/browser/src/__tests__/cookieless.test.ts
```diff
@@ -310,5 +310,32 @@ describe('cookieless', () => {
...
+            }
+
+            // Clear any previous startIfEnabled calls
+            mockStartIfEnabled.mockClear()
+
+            // Opt in
+            posthog.opt_in_capturing()
+
+            // Autocapture should be restarted immediately after opt-in
+            expect(mockStartIfEnabled).toHaveBeenCalledTimes(1)
```
robbie-c, line 333: Hmm I don't think this is the right test

This is just testing an implementation detail.

Ideally we would test whether an autocapture event is sent when a simulated click happens

robbie-c (changes requested):

Issue #2290, opened 2025-09-05
Ensure all LLM providers send system prompt


carlos-marchal-ph:
### Description

Ensure all LLM providers send system prompt

PR #2354, opened 2025-09-23
fix: ensure consistent system prompts

carlos-marchal-ph:
Solves #2290, implements equivalent behaviour to [this PR for the Python SDK](https://github.com/PostHog/posthog-python/pull/326).

### DIFF
--- a/packages/ai/src/utils.ts
+++ b/packages/ai/src/utils.ts
@@ -528,3 +528,35 @@ export const sendEventToPosthog = async ({
     client.capture(event)
   }
 }
+
+export function formatOpenAIResponsesInput(input: unknown, instructions?: string | null): FormattedMessage[] {
+  const messages: FormattedMessage[] = []
+
+  if (instructions) {
+    messages.push({
+      role: 'system',
+      content: instructions,
+    })
+  }
+
+  if (Array.isArray(input)) {
+    for (const item of input) {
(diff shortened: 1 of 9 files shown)


packages/ai/src/utils.ts
```diff
@@ -528,3 +528,35 @@ export const sendEventToPosthog = async ({
...
+    })
+  }
+
+  if (Array.isArray(input)) {
+    for (const item of input) {
+      if (typeof item === 'string') {
+        messages.push({ role: 'user', content: item })
+      } else if (item && typeof item === 'object') {
+        const obj = item as Record<string, unknown>
+        const role = (obj.role as string) || 'user'
```
Radu-Raicea: Let's use a type guard instead of casting.

carlos-marchal-ph: Agreed with the converter pattern, I'm not too happy with the current state of the SDKs. I've done some work on this but it's a bit hard to budget time for it even though we know we should improve it at some point.

As for the type guard, what do you mean exactly? We are already checking for the item to be an object, the `as` cast is just required here because TS doesn't cast down on its own.

Radu-Raicea: I meant for the string type cast. We should avoid type casting and using some type guard that we define in `packages/ai/src/typeGuards.ts`. There's already a `isString` defined there.

PR #2355, opened 2025-09-23
fix: use persistence instead of localStorage on surveys

### LINKED ISSUE #1898: `ReferenceError: localStorage is not defined` with posthog.reset() and memory persistence
I'm trying out posthog in a chrome extension and would like to capture some events in the service worker, which of course means, it has no access to localStorage, sessionStorage, etc.

I've set persistence to "memory" but get an error on `posthog.reset()` on user logout that `ReferenceError: localStorage is not defined`

It seems to be coming from `posthog-surveys.ts`
```
    reset(): void {
        localStorage.removeItem('lastSeenSurveyDate')
        const surveyKeys = getSurveySeenStorageKeys()
        surveyKeys.forEach((key) => localStorage.removeItem(key))
    }
```

I initialize posthog with
```
        posthog.init(process.env.POSTHOG_API_KEY, {
            api_host: process.env.POSTHOG_HOST,
            persistence: "memory",
            disable_session_recording: true,
            autocapture: false,
            capture_pageview: false,
            capture_pageleave: false,
            person_profiles: "identified_only"
        });

```

lucasheriques:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/1898

[Related ticket](https://posthoghelp.zendesk.com/agent/tickets/38000)

## Changes

Surveys used to rely on `localStorage` directly to manage some properties we use for our functionality (surveySeen, lastSeenSurvey, and inProgressSurvey).

Now, instead, we use posthog.persistence, but maintain backwards compatibility by using localStorage if persistence is disabled. This is necessary as we can have customers that are using surveys, but might not be using persistence. Unlikely, but can happen

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-surveys.ts
```diff
@@ -57,16 +58,21 @@ export class PostHogSurveys {
...
-        localStorage.removeItem('lastSeenSurveyDate')
-        const surveyKeys = []
+        clearFromPersistenceWithLocalStorageFallback('lastSeenSurveyDate', this._instance)
+        const surveyKeys = new Set<string>()
+        for (let i = 0; i < this._instance.persistence?.props.length; i++) {
+            const key = this._instance.persistence?.props[i]
+            if (key?.startsWith(SURVEY_SEEN_PREFIX) || key?.startsWith(SURVEY_IN_PROGRESS_PREFIX)) {
+                surveyKeys.add(key)
+            }
+        }
```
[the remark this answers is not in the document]
lucasheriques: fixed, good catch

packages/browser/src/posthog-surveys.ts
```diff
@@ -57,16 +58,21 @@ export class PostHogSurveys {
...
         for (let i = 0; i < localStorage.length; i++) {
             const key = localStorage.key(i)
             if (key?.startsWith(SURVEY_SEEN_PREFIX) || key?.startsWith(SURVEY_IN_PROGRESS_PREFIX)) {
-                surveyKeys.push(key)
+                surveyKeys.add(key)
             }
         }
```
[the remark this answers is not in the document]
lucasheriques: fixed, good catch

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1034,9 +1036,9 @@ export function Questions({
     forceDisableHtml: boolean
     posthog?: PostHog
 }) {
-    // Initialize responses from localStorage or empty object
+    // Initialize responses from storage or empty object
```
marandaneto:
```suggestion
    // Initialize responses from persistence or empty object
```

packages/browser/src/utils/survey-utils.ts
```diff
@@ -39,14 +41,61 @@ export const getSurveySeenKey = (survey: Pick<Survey, 'id' | 'current_iteration'
...
-    const isSurveySeen = localStorage.getItem(getSurveySeenKey(survey))
-    // if survey is already seen, no need to set it again
+/**
+ * We used to retrieve from localStorage directly. Now, we instead rely on the persistence API, since this is the preferred way to store data.
+ * But, since we might have customers that might be using surveys with persistence disabled, we need to fallback to localStorage
+ * to maintain backwards compatibility.
+ */
+export const getFromPersistenceWithLocalStorageFallback = (key: string, posthog?: PostHog) => {
+    try {
+        if (posthog?.persistence && !posthog.persistence.isDisabled()) {
```
marandaneto: this check is duplicate across the 3 methods, i'd extract and reuse it

packages/browser/src/utils/survey-utils.ts
```diff
@@ -39,14 +41,61 @@ export const getSurveySeenKey = (survey: Pick<Survey, 'id' | 'current_iteration'
...
+ */
+export const setOnPersistenceWithLocalStorageFallback = (key: string, value: any, posthog?: PostHog) => {
+    try {
+        if (posthog?.persistence && !posthog.persistence.isDisabled()) {
+            posthog.persistence.set_property(key, value)
+        } else {
+            localStorage.setItem(key, value)
+        }
+    } catch (e) {
+        SURVEY_LOGGER.error('Error setting survey seen on persistence', e)
```
marandaneto:
```suggestion
        SURVEY_LOGGER.error('Error setting survey seen on persistence or localStorage', e)
```

packages/browser/src/utils/survey-utils.ts
```diff
@@ -39,14 +41,61 @@ export const getSurveySeenKey = (survey: Pick<Survey, 'id' | 'current_iteration'
...
+ * to maintain backwards compatibility.
+ */
+export const getFromPersistenceWithLocalStorageFallback = (key: string, posthog?: PostHog) => {
+    try {
+        if (posthog?.persistence && !posthog.persistence.isDisabled()) {
+            return posthog.persistence.get_property(key)
+        }
+        return localStorage.getItem(key)
+    } catch (e) {
+        SURVEY_LOGGER.error('Error getting property from persistence or localStorage', e)
```
marandaneto: those logs would be better if it was logged when an error happened within the persistence or localStorage happened so it'd be easier to debug issues.
right now we'd not know whats throwing, the persistence or localStorage

lucasheriques: separate try/catch blocks for all of them now

marandaneto (approved):

packages/browser/src/posthog-surveys.ts
```diff
@@ -86,6 +87,11 @@ export class PostHogSurveys {
             logger.info('Not loading surveys in cookieless mode without consent.')
             return
         }
+        if (isPersistenceEnabledWithLocalStorage(this._instance)) {
+            logger.warn(
+                'Persistence does not include localStorage, but surveys it to work properly. Please set persistence to include localStorage to avoid this warning, or set disable_surveys to true. Falling back to localStorage usage directly to maintain backwards compatibility.'
```
marandaneto: > but surveys it to work properl

this sentence is confusing

.changeset/blue-wombats-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: use persistence instead of localStorage on surveys
```
marandaneto: i think the reason should be slightly better now with the possible odd behaviour after this change (also the PR description)
maybe it should not be a patch anymore?

packages/browser/src/utils/survey-utils.ts
```diff
@@ -39,14 +42,90 @@ export const getSurveySeenKey = (survey: Pick<Survey, 'id' | 'current_iteration'
...
+        posthog?.persistence &&
+        !posthog.persistence.isDisabled() &&
+        posthog.config?.persistence?.includes('localStorage')
+    )
+}
+
+/**
+ * We used to retrieve from localStorage directly. Now, we instead rely on the persistence API, since this is the preferred way to store data.
+ * But, since we might have customers that might be using surveys with persistence disabled, we need to fallback to localStorage
+ * to maintain backwards compatibility.
```
marandaneto, line 59: i think the `persistence` config should mention that surveys will fallback to localStorage if needed

marandaneto:
> @marandaneto i actually tested more throughly, and posthog does not work on incognito well without localStorage access. as it causes the survey to show up consistently without it
> 
> so instead I made [this commit](https://github.com/PostHog/posthog-js/pull/2355/commits/50e8d6791d2f5437759f610f5f30b8e55aa1558f) to warn about it, but still fall back to localStorage. but it's behind a try/catch, so it should also avoid the exceptions users are seeing
> 
> but it's a logic change regardless. any thoughts on it before i merge?

a good moment to review/engage with https://github.com/PostHog/product-internal/pull/854 and talk to the customer support team on how to handle (and communicate) this change that might break customers
do we need this somewhere in https://posthog.com/docs/surveys/troubleshooting ?
do we need to let customers know about this change?
do we have numbers on how many customers dont init posthog with persistence? etc
i dont know the impact so i cant really tell if its ok or not
as a 1st step, i'd just try-catch the local storage usages to avoid the errors on customers as a separate PR and dont get blocked by this until you figure this out

packages/browser/src/utils/survey-utils.ts
lucasheriques: just adde

.changeset/blue-wombats-stay.md
lucasheriques: well the behavior is not changing per say - it is exactly the same, but now using posthog.persistence abstraction. because even if it's not on the correct settings, we always fallback to localStorage directly. so not a breaking change nor functionality change

but i think it's not really a bug fix, and instead a design change, changing to a minor makes sense

marandaneto:
> @marandaneto appreciate the concerns!!
> 
> however, I don't think it's needed for this case. We're still maintaining the same functionality, and just adding a warning about localStorage usage that people potentially didn't know
> 
> even if someone is using `memory/sessionStorage/cookies`, we'll still use localStorage (and warn our users about it)
> 
> also those are good questions. maybe we should also add them to the PR template as well 🤔

is there any difference between acessing `localstorage` directly or using the persistence API using the default persistence config which is `localstorage+cookie`? if there's no diff, I agree, its back compatible, if there's a difference, then its breaking the behaviour.

PR #2357, opened 2025-09-24
fix: app state logging diff performance

pauldambra:
in one scene we saw some lag while this logging was active
there didn't look to be lots of logs in the session in question
so i guess it was the perf of the diffing

wrapped that in tests and performance assertions and refactored to be ~90x faster

### DIFF
--- a/packages/browser/src/customizations/posthogReduxLogger.ts
+++ b/packages/browser/src/customizations/posthogReduxLogger.ts
@@ -1,5 +1,6 @@
-import { BucketedRateLimiter, isNullish, isObject, Logger } from '@posthog/core'
+import { BucketedRateLimiter, isNullish, isObject, isUndefined, Logger } from '@posthog/core'
 import { createLogger } from '../utils/logger'
+import { PostHog } from 'posthog-js'
 
 // types copied from redux toolkit so we can avoid taking a dependency in the library and confusing people using the SDK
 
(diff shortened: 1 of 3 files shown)


packages/browser/src/customizations/posthogReduxLogger.ts
```diff
@@ -90,68 +93,83 @@ function defaultTitleFunction(stateEvent: StateEvent): string {
...
+ * Requires that the loaded posthog instance is provided
+ * And returns the function to use as the logger
+ *
+ * e.g. const config = { logger: sessionRecordingLoggerForPostHogInstance(posthog) }
+ */
+export const sessionRecordingLoggerForPostHogInstance: (posthog: PostHog) => PostHogStateLogger =
+    (postHogInstance: PostHog) =>
+    (title: string, stateEvent: StateEvent): void => {
+        postHogInstance?.sessionRecording?.tryAddCustomEvent('app-state', { title, stateEvent })
+    }
```
pauldambra, line 114: expose this as a helper since we have replay time affordances for this specific case

packages/browser/src/customizations/posthogReduxLogger.ts
```diff
@@ -90,68 +93,83 @@ function defaultTitleFunction(stateEvent: StateEvent): string {
...
+ * NB exported for testing purposes only, not part of the public API and may change without warning
+ *
  * Returns { prevState: changedKeysOnly, nextState: changedKeysOnly }
  */
-function getChangedStateKeys<S>(prevState: S, nextState: S): { prevState?: Partial<S>; nextState?: Partial<S> } {
-    if (!isObject(prevState) || !isObject(nextState)) {
-        // we only support objects as state, but can't guarantee that's what we'll get
-        return {}
-    }
+export function getChangedStateKeys<S>(
```
pauldambra, line 122: this was very AI driven refactoring
but I got the performance and correctness tests in place first

went from >100ms to <30ms for the complex case

marandaneto (approved):
low context but looks like just fine

PR #2360, opened 2025-09-24
feat(react-native): add error tracking

hpouillot:
## Changes

- add error tracking processing to react native

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/examples/example-expo-53/pnpm-lock.yaml
+++ b/examples/example-expo-53/pnpm-lock.yaml
@@ -876,8 +876,8 @@ packages:
     engines: {node: '>=14'}
 
   '@posthog/core@file:../../target/posthog-core.tgz':
-    resolution: {integrity: sha512-sv2sck89IUXel5StH7mLlCBWpvw+Hh6JxSGDG2ZYCYlxG6V/GfCfGX8GIL/oPNnAkpd2mkFSUP7pOkkapeGHgg==, tarball: file:../../target/posthog-core.tgz}
-    version: 1.2.0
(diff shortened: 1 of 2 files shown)


daibhin (approved):

packages/react-native/src/posthog-rn.ts
```diff
@@ -800,6 +803,48 @@ export class PostHog extends PostHogCore {
...
+   *   ...
+   * })
+   * ```
+   *
+   * @param {Error} error The error to capture
+   * @param {Object} [additionalProperties] Any additional properties to add to the error event
+   * @returns {void}
+   */
+  captureException(error: Error | unknown, additionalProperties: PostHogEventProperties = {}): void {
+    const syntheticException = new Error('Synthetic Error')
```
marandaneto, line 838: why do we createa and pass a `syntheticException` if we have an `error` already?

hpouillot: because the input is unknown and it's the reponsability of the coercer (inside errorPropertiesBuilder) to use the stack from syntheticException when it makes sense. For instance this is the string coercer: https://github.com/PostHog/posthog-js/blob/main/packages/core/src/error-tracking/coercers/string-coercer.ts

marandaneto: gotcha, i'd create the synthetic exception only if needed to avoid the perf cost but that can be improved later if its a problem
if its a string i'd also do eg `Error(error)` so at least we keep the string value somewhere (if its not set anywhere else which probably is already).
thanks for the follow up, no action needed.

PR #2363, opened 2025-09-25
chore: flapping execution time tests

pauldambra:
the perf limits on the execution time tests for getChangedStateKeys were too close to the wire
let's give it some breathing room

See https://posthog.slack.com/archives/C09ADEV3AJD/p1758709244968169


packages/browser/src/__tests__/customizations/getChangeStateKeys.test.ts
pauldambra, line 1: ha, we don't need 3 million lines of snapshots 🤣

PR #2368, opened 2025-09-25
fix: recording should stop on forced idle and (conditionally) start on return

pauldambra:
if a session is forced to reset because of being idle too long without _any_ activity
then we don't have a signal to react to in the recorder
and can't do any teardown

this is most obvious if conditional recording is active, the new session will simply carry on without taking into account condition triggers

let's add an event when it happens
register for that
and we can teardown completely

i am only going to fix this in the lazy recorder since i'm trying to switch over to it 

the added benefit here is that if the page is forced to idle reset we completely teardown the recorder


packages/browser/src/sessionid.ts
```diff
@@ -36,6 +37,11 @@ export class SessionIdManager {
     // we track activity so we can end the session proactively when it has passed the idle timeout
     private _enforceIdleTimeout: ReturnType<typeof setTimeout> | undefined
 
+    private _eventEmitter: SimpleEventEmitter = new SimpleEventEmitter()
+    public on(event: 'forcedIdleReset', handler: () => void): void {
+        this._eventEmitter.on(event, handler)
+    }
```
pauldambra, line 43: adding an event emitter now, it's hard to move old events over to it as lots of the old events assume they might be called immediately on registering a handler which the event emitter doesn't do, it only fires on the next trigger

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -646,8 +648,6 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             this._endpoint = config?.endpoint
         }
 
-        this._setupSampling()
```
pauldambra, line 649: all setupSampling did was register the listener to make a sampling decision on session id change, which is now handled in the one session id callback

packages/browser/playwright/mocked/session-recording/session-recording-idle-timeout.spec.ts
pauldambra, line 1: this is entirely ai slop that doesn't work
will rewrite this after sleep

pauldambra: it's now 90% AI slop that does work

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -683,18 +683,22 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+                // a session was forced to reset due to idle timeout and lack of activity
+                this._clearConditionalRecordingPersistence()
+                this._isIdle = 'unknown'
+                this.stop()
+                // then we want a session id listener to restart the recording when a new session starts
+                this._onSessionIdListener = this._sessionManager.onSessionId((sessionId, windowId, changeReason) => {
+                    // this should first unregister itself
+                    this._onSessionIdListener?.()
+                    this._onSessionIdCallback(sessionId, windowId, changeReason)
+                })
```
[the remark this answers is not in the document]
pauldambra: i mean, this isn't right on its understanding of the logic
but you do have to look in several places in the file to prove that, so this is a reasonable refactoring for simplicity

packages/browser/src/sessionid.ts
```diff
@@ -302,7 +308,9 @@ export class SessionIdManager {
             // while this window has been idle - and the timer has not progressed - e.g. window memory frozen while hidden
             const [lastActivityTimestamp] = this._getSessionId()
             if (this._sessionHasBeenIdleTooLong(new Date().getTime(), lastActivityTimestamp)) {
+                const idleSessionId = this._sessionId
                 this.resetSessionId()
+                this._eventEmitter.emit('forcedIdleReset', { idleSessionId })
```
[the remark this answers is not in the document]
pauldambra: it's not used now greptile, but the future might use it, and the event emitter has an enforced payload so 🤷

TueHaulund (approved):

PR #2369, opened 2025-09-26
feat(flags): add option to configure evaluation environments in your posthog config

dmarticus:
## Problem

Feature flags are evaluated for all SDK instances regardless of their deployment environment, leading to unnecessary evaluations and increased costs. Teams need a way to constrain feature flag evaluation to specific environments (e.g., only evaluate production flags in production, only evaluate web flags on web platforms).

## Changes

This PR implements support for evaluation environments in the posthog-js SDK, allowing users to specify which environment tags their SDK instance should use when evaluating feature flags.

Key changes:
  - Added evaluation_environments configuration option to PostHogConfig that accepts an array of environment tags (e.g., `['production', 'web', 'checkout']`)
  - Updated the `/flags` endpoint calls to include evaluation_environments in the request payload when configured
  - Updated `getRemoteConfigPayload` method to also include evaluation environments for consistency
  - Added comprehensive tests to verify the new functionality

This implementation follows the specification outlined in:
  - PostHog documentation PR: https://github.com/PostHog/posthog.com/pull/12929
  - Related internal PRs: https://github.com/PostHog/posthog/pull/38681 and https://github.com/PostHog/posthog/pull/37719

When configured, only feature flags that have at least one matching evaluation tag will be evaluated. Feature flags with no evaluation tags will always be evaluated for backward compatibility.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-featureflags.ts
```diff
@@ -391,6 +391,11 @@ export class PostHogFeatureFlags {
             data.disable_flags = true
         }
 
+        // Add evaluation environments if configured
+        if (this._instance.config.evaluation_environments && this._instance.config.evaluation_environments.length > 0) {
```
haacked: That's a mouthful. Maybe extract this condition to a method:

```js
private _shouldIncludeEvaluationEnvironments(): boolean {
      return !!(this._instance.config.evaluation_environments?.length)
  }
```

Since it's used in a couple places.

haacked: Also, would it make sense to validate this only contains valid strings. For example:

```js
const validEnvs = this._instance.config.evaluation_environments?.filter(
       env => env && typeof env === 'string' && env.trim().length > 0
   )
```

Or even better, combine this with a console error for invalid values.

packages/browser/src/types.ts
```diff
@@ -764,6 +764,18 @@ export interface PostHogConfig {
...
+     * Evaluation environments for feature flags.
+     * When set, only feature flags that have at least one matching evaluation tag
+     * will be evaluated for this SDK instance. Feature flags with no evaluation tags
+     * will always be evaluated.
+     *
+     * Examples: ['production', 'web', 'checkout']
+     *
+     * @default undefined
+     */
+    evaluation_environments?: string[]
```
haacked: We probably don't want to allow mutations.

```suggestion
    evaluation_environments?: readonly string[]
```

PR #2372, opened 2025-09-26
feat: add nuxt support

ablaszkiewicz:
### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,126 @@
...
+      exceptionAutoCaptureEnabled: false,
+      configOverride: {},
+      debug: false,
+    },
+    nitro: {
+      exceptionAutoCaptureEnabled: false,
+      configOverride: {},
+      debug: false,
+    },
+  }),
```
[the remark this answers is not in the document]
ablaszkiewicz: It is strongly typed. Typescript will not let you not pass that variable

packages/nuxt/README.md
```diff
@@ -0,0 +1,86 @@
+# PostHog Nuxt module
+
+- handles sourcemap configuration and upload for the PostHog Error Tracking product.
+- provides posthog client and auto exception capture for Vue and Nitro,
```
daibhin:
```suggestion
- Handles sourcemap configuration and upload for the PostHog Error tracking product
- Provides PostHog client and auto exception capture for Vue and Nitro apps
```

packages/nuxt/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+Please see the main [PostHog Error Tracking docs](https://posthog.com/docs/error-tracking).
```
daibhin:
```suggestion
Please see the main [PostHog Error tracking docs](https://posthog.com/docs/error-tracking).
```

packages/nuxt/README.md
```diff
@@ -0,0 +1,86 @@
...
+
+## Usage
+
+1. Install the package using
```
daibhin:
```suggestion
1. Install the package
```

packages/nuxt/README.md
```diff
@@ -0,0 +1,86 @@
...
+    },
+    sourceMaps: {
+      enabled: true, // Enables sourcemaps generation and upload
+      envId: '2', // Environment ID, see https://app.posthog.com/settings/environment#variables
+      project: 'my-application', // (optional) Project name, defaults to git repository name
+      version: '1.0.0', // (optional) Release version, defaults to current git commit
+      privateApiKey: 'private api key', // Your personal API key. You can generate it in settings -> Personal API keys
+    },
+  },
+})
```
daibhin, line 46: Why not do something like we have for Next where we give users a function that wraps their existing config e.g.

```
// nuxt.config.ts
import { withPostHogConfig } from "@posthog/nuxt";

const nuxtConfig = {
  // Your Nuxt.js configuration here
};

export default withPostHogConfig(nuxtConfig, {
  personalApiKey: process.env.POSTHOG_PERSONAL_API_KEY!, // Personal API key used for sourcemap uploads, see https://app.posthog.com/settings/user-api-keys
  ...
});
```

That way we could make sure that the sourcemap properties were correctly set for users:
```
sourcemap: { client: 'hidden' }, // Make sure to set it (otherwise client sourcemaps will not be generated)

  nitro: {
    rollupConfig: {
      output: {
        sourcemapExcludeSources: false, // Make sure to set it (otherwise server sourcemaps will not be generated)
      },
    },
  },
```

packages/nuxt/package.json
```diff
@@ -0,0 +1,50 @@
+{
+  "name": "@posthog/nuxt",
```
daibhin, line 2: For the same reasons we call it `nextjs-config`, I think we should rename this package `nuxt-config`

packages/nuxt/src/runtime/nitro-plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+  const client = new PostHog(apiKey, {
+    host: host,
+    ...configOverride,
+  })
+
+  if (debug) {
+    client.debug(true)
+  }
+
+  nitroApp.hooks.hook('error', (error, { event }) => {
```
daibhin: Should we move the hook outside the handler and never set the handler up in the first place if we don't need it?

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,137 @@
...
+
+const VUE_OUTPUT_DIRECTORY = '.output/public'
+const NITRO_OUTPUT_DIRECTORY = '.output/server/chunks'
+
+export interface ModuleOptions {
+  host: string
+  publicApiKey: string
+  nuxt: {
+    exceptionAutoCaptureEnabled?: boolean
+    configOverride?: Partial<PostHogConfig>
```
daibhin: Why pull exception autocapture out into it's own value? What would happen if someone set `capture_exceptions` on the config to false but `exceptionAutoCaptureEnabled` to true - which value would take precedence?

daibhin: `exceptionAutoCaptureEnabled` only controls the framework handlers... does Nuxt swallow all kinds of exceptions e.g. unhandledrejections and window.onerror

packages/nuxt/package.json
```diff
@@ -0,0 +1,50 @@
...
+  "license": "MIT",
+  "type": "module",
+  "exports": {
+    ".": {
+      "types": "./dist/types.d.mts",
+      "import": "./dist/module.mjs"
+    }
+  },
+  "main": "./dist/module.mjs",
+  "typesVersions": {
```
hpouillot: a bit overkill at this stage

packages/nuxt/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@types/node": "^20.0.0",
+    "nuxt": "^4.1.2",
+    "@nuxt/devtools": "^2.6.5",
+    "@nuxt/module-builder": "^1.0.2",
+    "@nuxt/schema": "^4.1.2"
+  },
+  "peerDependencies": {
+    "posthog-node": ">=5.6.0",
+    "@posthog/core": ">=1.0.2",
+    "posthog-js": ">=1.257.2"
```
hpouillot: you should add nuxt as a peer dep

packages/nuxt/src/runtime/nitro-plugin.ts
```diff
@@ -0,0 +1,34 @@
...
+    if (!exceptionAutoCaptureEnabled) {
+      return
+    }
+
+    client.captureException(error, uuidv7(), {
```
hpouillot: I would call `await captureExceptionImmediate` in case the runtime gets destroyed

packages/nuxt/package.json
```diff
@@ -0,0 +1,50 @@
...
+      ]
+    }
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "prepare": "nuxt prepare",
+    "build": "nuxt-module-build build",
+    "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz"
```
hpouillot: would be cool to have a lint script even if it's not specific to nuxt

```suggestion
    "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz",
    "lint": "eslint src"
```

packages/nuxt/playground/nuxt.config.ts
```diff
@@ -0,0 +1,29 @@
+export default defineNuxtConfig({
+  modules: ['../src/module'],
+  devtools: { enabled: true },
+  sourcemap: { client: true },
+  posthog: {
+    host: 'http://localhost:8010',
+    publicApiKey: 'phc_VXlGk6yOu3agIn0h7lTmSOECAGWCtJonUJDAN4CexlJ',
```
hpouillot:
```suggestion
    projectApiKey: process.env.NUXT_PUBLIC_POSTHOG_PROJECT_API_KEY
```

packages/nuxt/playground/nuxt.config.ts
```diff
@@ -0,0 +1,29 @@
...
+    },
+    nitro: {
+      exceptionAutoCaptureEnabled: true,
+    },
+    sourceMaps: {
+      enabled: true,
+      version: 'V1',
+      envId: '2',
+      project: 'i-love-nuxt-1',
+      privateApiKey: 'phx_YZZHl8xzLkCWHSpVahmkggLGaS6gmSxCNmH26N0RUGZnqAs',
```
hpouillot:
```suggestion
      personalApiKey: process.env.POSTHOG_PERSONAL_API_KEY
```

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,137 @@
...
+    debug?: boolean
+  }
+  nitro: {
+    exceptionAutoCaptureEnabled?: boolean
+    configOverride?: PostHogOptions
+    debug?: boolean
+  }
+  sourceMaps:
```
hpouillot: why are there two types for sourceMaps ?

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,137 @@
...
+    nuxt.options.runtimeConfig.public.nitroPosthogClientDebug =
+      nuxt.options.runtimeConfig.public.nitroPosthogClientDebug || options.nitro.debug
+
+    if (!options.sourceMaps.enabled) {
+      return
+    }
+
+    nuxt.hooks.hook('nitro:build:public-assets', async () => {
+      try {
+        execSync(`posthog-cli sourcemap inject --directory ${VUE_OUTPUT_DIRECTORY}`)
```
hpouillot: `posthog-cli` won't be defined if the node_modules/.bin folder is not in path, you might need to move a method from nextjs-config to locate a binary

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,137 @@
...
+        `POSTHOG_CLI_TOKEN=${options.sourceMaps.privateApiKey}`,
+        `posthog-cli --host ${options.host} sourcemap upload --directory ${VUE_OUTPUT_DIRECTORY}`,
+        options.sourceMaps.version ? `--version ${options.sourceMaps.version}` : '',
+        options.sourceMaps.project ? `--project ${options.sourceMaps.project}` : '',
+      ]
+        .filter(Boolean)
+        .join(' ')
+
+      try {
+        execSync(publicUploadCmd)
```
hpouillot:
```suggestion
        execSync(publicUploadCmd, {
        	env: {
        		POSTHOG_CLI_ENV_ID: ....
        	}
        })
```

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,137 @@
...
+      }
+
+      const publicUploadCmd = [
+        `POSTHOG_CLI_ENV_ID=${options.sourceMaps.envId}`,
+        `POSTHOG_CLI_TOKEN=${options.sourceMaps.privateApiKey}`,
+        `posthog-cli --host ${options.host} sourcemap upload --directory ${VUE_OUTPUT_DIRECTORY}`,
+        options.sourceMaps.version ? `--version ${options.sourceMaps.version}` : '',
+        options.sourceMaps.project ? `--project ${options.sourceMaps.project}` : '',
+      ]
+        .filter(Boolean)
```
hpouillot: ??

packages/nuxt/package.json
ablaszkiewicz: Initially I named it `nuxt-config` but then I realized this is not just a config. This is a whole proper nuxt SDK built on top of their modules system. So this is a nuxt-native way of doing that stuff. Imo this is not a config but please leave your opinions @daibhin, @hpouillot

packages/nuxt/src/module.ts
ablaszkiewicz: Notice that one type has `enabled: true` and then some properties are required and other type has `enabled: false` and then all properties are optional. Still changed that to something more intuitive

```
interface DisabledSourcemaps {
  enabled: false
}

interface EnabledSourcemaps {
  enabled: true
  privateApiKey: string
  envId: string
  version?: string
  project?: string
}
...
sourcemaps: DisabledSourcemaps | EnabledSourcemaps | undefined
```

packages/nuxt/README.md
ablaszkiewicz: Not easily doable due to how Nuxt inferes config types dynamically based on what Nuxt modules you have installed. It does not provide access to its dynamic config type of out the box to do this kind of thing.

Spent 15 minutes trying to make it work. Maybe it is doable but I'd rather not spend more time on that.

packages/nuxt/package.json
hpouillot: In this case it makes sense to call it `@posthog/nuxt` but I would add basic features like getting the posthog instance on the client and the server

packages/nuxt/src/module.ts
hpouillot: I think nuxt swallows some exception (to be tested), but even if it does I would read the config to know if we should add a hook or not

examples/example-nuxt-module/nuxt.config.ts
```diff
@@ -0,0 +1,29 @@
+export default defineNuxtConfig({
```
hpouillot, line 2: I think we should move the playground into an examples/example-nuxt folder

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,148 @@
...
+
+export interface ModuleOptions {
+  host: string
+  publicKey: string
+  nuxt: {
+    exceptionAutoCaptureEnabled?: boolean
+    configOverride?: Partial<PostHogConfig>
+    debug?: boolean
+  }
+  nitro: {
```
hpouillot: Could we rename `nuxt` and `nitro` into `client` and `server` ?

packages/nuxt/src/module.ts
```diff
@@ -0,0 +1,148 @@
...
+    configKey: 'posthogConfig',
+    compatibility: {
+      nuxt: '>=3.7.0',
+    },
+  },
+  defaults: () => ({
+    host: 'https://us.i.posthog.com',
+    nuxt: {
+      exceptionAutoCaptureEnabled: false,
+      configOverride: {},
```
hpouillot: IMO, configOverride is a bit confusing as it is the whole config used

examples/example-nuxt-module/nuxt.config.ts
ablaszkiewicz: MOVED

packages/nuxt/src/module.ts
ablaszkiewicz: Double checked that. Every exception goes through these handlers both on frontend and backend. I manually tested unhandled exceptions, unhandled rejections, throwing inside chains, throwing inside timeouts, nextTick - everything got caught.

However I decided to disable hook-capture on frontend due to the conflict with remote config (what if we enable capture via vue's hook but then on top of that user enables auto capture in remote config - we would catch these exceptions twice as that hook does not stop error propagation)

packages/nuxt/package.json
ablaszkiewicz: Added getting ph client on the vue side. Did not add it on the nitro side as it seems that it is not a patern. Added appropriate section in docs which explains both cases.

packages/nuxt/README.md
```diff
@@ -0,0 +1,88 @@
...
+        sourcemapExcludeSources: false, // Make sure to set it (otherwise server sourcemaps will not be generated)
+      },
+    },
+  },
+
+  posthog: {
+    host: 'http://localhost:8010', // (optional) Host URL, defaults to https://us.posthog.com
+    publicApiKey: 'public api key', // Your public web snippet key. You can find it in settings
+    nuxt: {
+      exceptionAutoCaptureEnabled: true, // Enables vue runtime plugin which forwards exceptions caught via vue:error hook
```
hpouillot: we should probably update this

packages/nuxt/README.md
```diff
@@ -0,0 +1,88 @@
...
+    nitro: {
+      exceptionAutoCaptureEnabled: true, // Enables nuxt runtime plugin which forwards exceptions caught via error hook
+      configOverride?: PostHogOptions // (optional) It will be passed to the posthog-node client on init in nitro
+    },
+    sourceMaps: {
+      enabled: true, // Enables sourcemaps generation and upload
+      envId: '2', // Environment ID, see https://app.posthog.com/settings/environment#variables
+      project: 'my-application', // (optional) Project name, defaults to git repository name
+      version: '1.0.0', // (optional) Release version, defaults to current git commit
+      privateApiKey: 'private api key', // Your personal API key. You can generate it in settings -> Personal API keys
```
hpouillot: personalApiKey ?

ablaszkiewicz:
@Chippd you can get it from the URL when you're on the posthog app website.

```
https://us.posthog.com/project/<env-id>
```

PR #2373, opened 2025-09-28
fix: 2 small conditional recording fixes

pauldambra:
spotted while testing the lazy loaded recorder

only going to fix in the lazy loaded recorder as i'm trying to switch to it and don't want to muddy the testing waters

* store session id instead of boolean for sampled flag
* take full snapshot only once a minute only if not yet started
* add unit tests for lazy recording since production soak is taking so long

### DIFF
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -690,6 +690,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             })
         })
 
+        this._makeSamplingDecision(this.sessionId)
         this._receivedFlags = true
         this._startRecorder()
 
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -43,6 +43,7 @@ import {
     isNullish,
     isNumber,
     isObject,
+    isString,
     isUndefined,
 } from '@posthog/core'
 import {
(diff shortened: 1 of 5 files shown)


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -557,7 +560,10 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
     }
 
     private get _fullSnapshotIntervalMillis(): number {
-        if (this._triggerMatching.triggerStatus(this.sessionId) === TRIGGER_PENDING) {
+        if (
+            this._triggerMatching.triggerStatus(this.sessionId) === TRIGGER_PENDING &&
+            !['sampled', 'active'].includes(this.status)
+        ) {
```
pauldambra, line 566: when we're waiting on a trigger we want more frequent full snapshots

but when we have _any_ match we might already be recording

so we have to check both things

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1188,7 +1194,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         }
 
         this._instance.persistence?.register({
-            [SESSION_RECORDING_IS_SAMPLED]: shouldSample,
+            [SESSION_RECORDING_IS_SAMPLED]: shouldSample ? sessionId : false,
```
pauldambra, line 1198: we were just storing boolean true in is sampled, and every so often we manage to carry on recording in a new session
let's just store the session id

pauldambra: that way we can check if the _current_ session is sampled when deciding to record

PR #2378, opened 2025-09-30
fix: don't mangle to dollar in any build

pauldambra:
we don't want to mangle to $ ever
this updates so that the no external and es5 builds don't do that (previously they simply accepted the mangler defaults)

this won't fix every clash with jquery

since we rely on preact and chrome web vitals and at the time of writing they both mangle something to `$`

follows https://github.com/PostHog/posthog-js/pull/2266


veryayskiy (approved):
jquery is the king.

PR #2382, opened 2025-09-30
chore: add SDK Doctor test files to /examples

slshults:
Adding test files from SDK Doctor to `/examples`, requested by Rafa


rafaeelaudibert (approved):
Can you use `html` rather than `htm` as the extension? Just to keep them similar to the other ones

PR #2402, opened 2025-10-06
chore: Help getting local dev nextjs example working

robbie-c:
## Problem

The nextjs playground doesn't point to the locally built posthog-js by default

## Changes

* Point to local posthog-js
* Make pnpm-workspace.yaml empty
* Add (very redundant, we already exclude these) exclusion rules to exclude playground and examples directories from pnpm workspace (we don't need this change at all, I'm just paranoid about future refactors)
* Update comment

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/playground/README.md
+++ b/packages/browser/playground/README.md
@@ -1,5 +1,5 @@
-A collection of examples for showing how to use PostHog's JS SDK
+A collection of examples for showing how to use PostHog's JS SDK or for testing features of it.
 
-Or for testing features of it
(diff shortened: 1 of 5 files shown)


thmsobrmlr (approved):
Work like a charm ❤️

PR #2405, opened 2025-10-06
fix: lazy replay received flag tracking should be in the wrapper

pauldambra:
lifting changes out of https://github.com/PostHog/posthog-js/pull/2404

i want to make as many changes as possible before removing the lazy loaded code so i have a chance to back out of i need to

we're tracking received flag in the lazy loaded code, but not using it
and types are wrong... sessionRecording can be false but we don't represent that

### DIFF
--- a/packages/browser/src/extensions/replay/sessionrecording-wrapper.ts
+++ b/packages/browser/src/extensions/replay/sessionrecording-wrapper.ts
@@ -11,7 +11,7 @@ import {
     PostHogExtensionKind,
     window,
 } from '../../utils/globals'
-import { LAZY_LOADING, SessionRecordingStatus, TriggerType } from './triggerMatching'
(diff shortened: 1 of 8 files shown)


packages/browser/src/extensions/replay/sessionrecording-wrapper.ts
```diff
@@ -125,7 +131,9 @@ export class SessionRecordingWrapper {
             const persistence = this._instance.persistence
 
             const persistResponse = () => {
-                const sessionRecordingConfigResponse = response.sessionRecording
+                const sessionRecordingConfigResponse =
+                    response.sessionRecording === false ? undefined : response.sessionRecording
```
pauldambra, line 135: wild that we've never represented that sessionRecording can be false
it's "safe" to just treat it as undefined when false

TueHaulund (approved):

Issue #2161, opened 2024-10-25
Console logs support when using JS/TS console.log and friends


marandaneto:
### Description

Right now it only works with Native Android logcat using the `captureLog` flag.
We'd need to monkey patch the console log in JS similarly to the [Web](https://github.com/postHog/posthog-js)

iOS logs https://github.com/PostHog/posthog-ios/issues/258

marandaneto:
this can be easily done now since its added for error tracking https://github.com/PostHog/posthog-js/blob/52f646a66ec2064a94a786b68c03b4df0ca73ca2/packages/react-native/src/error-tracking/utils.ts#L35-L49

PR #2410, opened 2025-10-07
feat(react-native): add autocapture handling

hpouillot:
## Changes

- add exception autocapture
- add unhandled promise rejections autocapture

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes


.changeset/happy-terms-jam.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': patch
```
marandaneto: i think this should be a minor? its a new feat

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+    }
+  }
+
+  private autocaptureUncaughtErrors() {
+    const globalHandler = ErrorUtils.getGlobalHandler()
+    ErrorUtils.setGlobalHandler((error, isFatal) => {
+      const hint: EventHint = {
+        mechanism: {
+          type: 'onuncaughtexception',
+          handled: false,
```
marandaneto, line 116: should we use `isFatal` here? or we should check for `isFatal` when setting `$exception_level`

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+        },
+      }
+      const additionalProperties: any = {}
+      if (isFatal) {
+        additionalProperties['$exception_level'] = 'fatal' as SeverityLevel
+      }
+      this.captureException(error, additionalProperties, hint)
```
marandaneto, line 125: do we set the `error` as part of the event props?
i think the original `error` should be part of the props (and non serialized) or the hint so the users can use type check in the beforeSend callback

marandaneto, line 125: will `captureException` swallow exceptions if something goes wrong? otherwise we will throw and not call `globalHandler` which is something we should not do

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+      globalHandler?.(error, isFatal)
+    })
+  }
+
+  private autocaptureUnhandledRejections() {
+    const onUnhandledRejection = (error: unknown) => {
+      const hint: EventHint = {
+        mechanism: {
+          type: 'onunhandledrejection',
+          handled: false,
```
marandaneto, line 145: this is where a boolean isnt ideal.
iirc unhandled rejection wont crash the app
uncaught -> crashes the app
unhandled -> does not crash
manually captured -> does not crash
as an user, i want to filter exceptions by priority uncaught > unhandled > manually captured

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+        allRejections: true,
+        onUnhandled: (_: any, error: any) => onUnhandledRejection(error),
+      })
+    } else {
+      // javascript-core handling
+      const tracking = require('promise/setimmediate/rejection-tracking')
+      tracking.enable({
+        allRejections: true,
+        onUnhandled: (_: any, error: any) => onUnhandledRejection(error),
+      })
```
marandaneto: there's quite a few issues with react native and promises because of multiple versions/polyfills
we probably need to wrap within a try catch at least
https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/packages/core/src/js/integrations/reactnativeerrorhandlers.ts#L83-L84

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+        unhandledRejections: autocapture,
+      }
+    }
+    return {
+      uncaughtExceptions: !!autocapture.uncaughtExceptions,
+      unhandledRejections: !!autocapture.unhandledRejections,
+    }
+  }
+
+  private autocaptureUncaughtErrors() {
```
marandaneto, line 111: this might change if its web/mobile/macos etc
https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/packages/core/src/js/integrations/reactnativeerrorhandlers.ts#L57
we need to make sure we are not crashing on the other platforms if the signature or the handlers dont exist in certain platforms we support

packages/react-native/src/error-tracking/index.ts
marandaneto: https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/packages/core/src/js/integrations/reactnativeerrorhandlersutils.ts

packages/react-native/src/error-tracking/index.ts
```diff
@@ -7,15 +7,29 @@ import {
...
   StringCoercer,
 } from '@posthog/core/error-tracking'
 import type { PostHog } from '../posthog-rn'
 import { PostHogEventProperties } from '@posthog/core'
+import { isHermes } from '../utils'
+
+interface AutocaptureOptions {
+  uncaughtExceptions?: boolean
+  unhandledRejections?: boolean
+}
```
daibhin, line 27: Any reason not to capture console errors like we do in web?
https://github.com/PostHog/posthog-js/blob/7b4fb773d2bb1522cee183fc3c7c5701eecd7ec9/packages/browser/src/types.ts#L203-L224

packages/react-native/src/error-tracking/index.ts
hpouillot: I override `$exception_level` to fatal below, but we don't use it in the UI yet.

packages/react-native/src/error-tracking/index.ts
hpouillot: error is unknown at this point, I don't think there is a way to type check this even if we add it to the props. hint can be accessed inside the $exception_list though but we don't reexport the types for now. I'll add an ExceptionEventType to the exports.

packages/react-native/src/error-tracking/index.ts
hpouillot: we will have to change the core types,  ingestion (de)serialization and the UI. I think it makes sense but we should probably do that as a follow-up.

packages/react-native/src/error-tracking/index.ts
hpouillot: Yes, make sense. Some old react-native version requires a Promise polyfill to work, and we won't support that for now but we could add it later.

packages/react-native/src/error-tracking/index.ts
hpouillot: Not yet, but I'll add safeguards to be extra sure we don't fall into an infinite exception loops

packages/react-native/src/error-tracking/index.ts
hpouillot: No particular reason, I just saw it as a secondary step as it won't be enabled by default

.changeset/happy-terms-jam.md
hpouillot: I bumped the minor for the stacktrace parsing but I can bump it again 🚧

packages/react-native/src/error-tracking/index.ts
marandaneto: thats fine but if isFatal=false, we still set `handled=false`, so we should use isFatal for both conditions just cus semantics

packages/react-native/src/error-tracking/index.ts
marandaneto: for that to be useful, the `hint` should be passed to the beforeSend callback as well
https://github.com/getsentry/sentry-react-native/blob/0b647531c13884b6be72c6f652482d79c6d2d96f/packages/core/src/js/integrations/reactnativeerrorhandlers.ts#L170
it coule be just a additionalProperties[originalException] = error // which will be documented
so users could do eg `additionalProperties[originalException] is MyCustomException` // drop it

packages/react-native/src/error-tracking/index.ts
marandaneto: thats fine, we need to be sure that the `require` and `enable` wont throw at least in case theres something off which is very possible since we support many RN versions

packages/react-native/src/error-tracking/index.ts
marandaneto: if we do this, we could reuse for session replay as well https://github.com/PostHog/posthog-js/issues/2161

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+
+  private autocaptureUnhandledRejections() {
+    const onUnhandledRejection = (error: unknown) => {
+      const hint: EventHint = {
+        mechanism: {
+          type: 'onunhandledrejection',
+          handled: false,
+        },
+      }
+      this.captureException(error, {}, hint)
```
marandaneto, line 148: a concern that i have here is that `captureException` wont have enough time to capture the event for fatal errors
if we have the SDK as flushAt=20 default, the queue has 19, then theres an uncaught exception, the exception error is the last one to be processed, by then the process died and we lost the most important item of the queue
for uncaught errors, we should also call flush and await the promise if possible.
or we should do a sort of priority queue.
its ok to start as is but this should be a follow up since its very important for mobile error tracking

hpouillot: Yes, I was about to add this part but I think we should always flush on fatal errors even if there is no exception autocapture. I'll add another handler for this.

packages/react-native/src/error-tracking/index.ts
hpouillot: what is the oldest react-native version we support ? I can't find it in the peer deps

packages/react-native/src/error-tracking/index.ts
marandaneto: we call flush when we put the app to the backgorund as well
the problem is that those handlers (usually) are not async so you cannot await
https://github.com/PostHog/posthog-js/blob/313053f21a30970b9eb30be92fb0e7262779b66d/packages/react-native/src/posthog-rn.ts#L138-L140
the file system libs are also async, which is also another problem, if this ever becomes a problem, we should use sync APIs by default eg https://github.com/mrousavy/react-native-mmkv

packages/react-native/src/error-tracking/index.ts
marandaneto: we dont use any new APIs as of now (jus stable ones for years), so its back compatible a long way, we dont pin to any minimum version.
if we start using new APIs or APIs that have changed recently or often, we should pin a min version (eg the promise polyfill might be one of them) or we just fallback to not enabling that feature, i'd rather not pin a min. version unless needed and just do runtime checks to check if we can use X or Y

packages/react-native/src/error-tracking/index.ts
hpouillot: hum, ok but you can still have a non-fatal exception that is not handled ?

marandaneto: yes but again its the 3 state type of errors
sentry does the same https://github.com/getsentry/sentry-react-native/blob/5ee3314125fe4c01a1e998efc90708cb39b99899/packages/core/src/js/integrations/reactnativeerrorhandlers.ts#L175-L187
unhandled but does not crash and handled but does crash, and one is more important than the other

hpouillot: Ok interesting, so catching an error at the globalErrorHandler is considered handled ?

packages/react-native/src/error-tracking/index.ts
```diff
@@ -35,4 +51,70 @@ export class ErrorTracking {
...
+    }
+    return {
+      uncaughtExceptions: !!autocapture.uncaughtExceptions,
+      unhandledRejections: !!autocapture.unhandledRejections,
+    }
+  }
+
+  private autocaptureUncaughtErrors() {
+    const globalHandler = ErrorUtils.getGlobalHandler()
+    ErrorUtils.setGlobalHandler((error, isFatal) => {
```
marandaneto: for the future: expose a config to capture or not events if `__DEV__=true`

packages/react-native/src/error-tracking/index.ts
hpouillot: I added console autocapture in latest version

packages/react-native/src/error-tracking/index.ts
marandaneto: yes
isFatal=true crashes the app, isFatal = false -> RN swallows and log it out
This lib is also useful to learn about the global handlers
https://github.com/a7ul/react-native-exception-handler/

packages/react-native/src/error-tracking/index.ts
hpouillot: I will leave it like that for now, as we don't support beforeSend in react-native for now...

PR #2414, opened 2025-10-08
chore: add logger to core instance

hpouillot:
## Changes

- reuse same logger across posthog instances (needed for log capturing)

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/core/src/posthog-core-stateless.ts
```diff
@@ -138,34 +141,35 @@ export abstract class PostHogCoreStateless {
...
+    this.requestTimeout = options.requestTimeout ?? 10000 // 10 seconds
+    this.featureFlagsRequestTimeoutMs = options.featureFlagsRequestTimeoutMs ?? 3000 // 3 seconds
+    this.remoteConfigRequestTimeoutMs = options.remoteConfigRequestTimeoutMs ?? 3000 // 3 seconds
+    this.disableGeoip = options.disableGeoip ?? true
+    this.disabled = options.disabled ?? false
     this.historicalMigration = options?.historicalMigration ?? false
     // Init promise allows the derived class to block calls until it is ready
     this._initPromise = Promise.resolve()
     this._isInitialized = true
+    this._logger = createLogger('[PostHog]', this.logMsgIfDebug.bind(this))
```
marandaneto, line 172: since we bind `logMsgIfDebug`, we still only log if `debug` is enabled, right?

hpouillot: yes

.changeset/sweet-squids-grin.md
```diff
@@ -0,0 +1,7 @@
+---
+'posthog-react-native': patch
+'@posthog/core': patch
+'posthog-node': patch
```
marandaneto, line 4: i think its fine to do patches but the user does not get any benefit from it
maybe it'd make sense to merge and let it sit until a new version with a user-facing change for a proper patch/minor
no storng opinions, but imo new versions that do not add any value to the user can wait a new version (less friction and less fomo of new versions)

hpouillot: Yep, that's why I didn't put the release tag. If I don't add this changeset and a new release of posthog-node or react-native is created, they will assume `@posthog/core` will have the changes which won't be the case

PR #2415, opened 2025-10-08
fix: surveys displayed only from cache

ioannisj:
Ticket: https://posthoghelp.zendesk.com/agent/tickets/38457 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/41570)

Problem: Surveys only appear on subsequent app launches after being loaded and cached

Why? 
- `PostHogSurveyProvider` calls `getSurveys()` immediately after `ready()`, but surveys may still be loading from remote config
- On first app launch this means no cached surveys are available, and async loading hasn't completed yet. If there are surveys cached, then we were working with the old cache instead of the freshly loaded surveys

## Changes

Survey loading now works are follows: 
- Load from remote config if remote config is enabled 
- Load from flags endpoint if preload flags is enabled. If flags response is missing surveys, fallback to loading from surveys API
- Load surveys from API

`PostHogSurveyProvider` will then wait for surveys to be loaded from any of the sources above before displaying a survey

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

# Conflicts:
#	packages/react-native/src/posthog-rn.ts


packages/react-native/src/posthog-rn.ts
```diff
@@ -174,14 +177,30 @@ export class PostHog extends PostHogCore {
       this._isInitialized = true
 
       if (this._disableRemoteConfig === false) {
-        this.reloadRemoteConfigAsync()
+        this.reloadRemoteConfigAsync().then((response) => {
+          if (response) {
+            this._handleSurveysFromRemoteConfig(response)
```
ioannisj: Not really needed cause `reloadRemoteConfigAsync` already caches surveys, but wanted to be consistent with survey caching and logging in this file

packages/react-native/src/posthog-rn.ts
```diff
@@ -174,14 +177,30 @@ export class PostHog extends PostHogCore {
...
+          this._notifySurveysReady()
+        })
       } else {
         this.logMsgIfDebug(() => console.info('PostHog Debug', `Remote config is disabled.`))
+
         if (options?.preloadFeatureFlags !== false) {
           this.logMsgIfDebug(() => console.info('PostHog Debug', `Feature flags will be preloaded from Flags API.`))
-          this.reloadFeatureFlags()
+          // Preload flags (and parse surveys as well since we are calling with config=true already)
+          this._flagsAsyncWithSurveys().then(() => {
```
ioannisj: Not 100% sure if this is a correct approach. I noticed that we are calling flags with config=true already, but response type omits surveys intentionally. I didn't touch any code from js lib but I try to read surveys payload and if missing, fallback to loading from API. Not sure this is a clean approach

packages/react-native/src/posthog-rn.ts
```diff
@@ -174,14 +177,30 @@ export class PostHog extends PostHogCore {
...
+            this._notifySurveysReady()
+          })
         } else {
-          this.logMsgIfDebug(() => console.info('PostHog Debug', `preloadFeatureFlags is disabled.`))
+          this.logMsgIfDebug(() =>
+            console.info('PostHog Debug', 'preloadFeatureFlags is disabled, loading surveys from API.')
+          )
+          // Load surveys directly from API since both remote config and preloading feature flags are disabled
+          // Note: if flags are not loaded/cached then surveys will not be displayed until reloadFeatureFlags() is called, since surveys depend on internal flags
```
ioannisj, line 213: I think this comment is accurate but double check me if I'm wrong here

packages/react-native/src/surveys/PostHogSurveyProvider.tsx
```diff
@@ -83,10 +83,9 @@ export function PostHogSurveyProvider(props: PostHogSurveyProviderProps): JSX.El
 
   // Load surveys once
   useEffect(() => {
-    // TODO: for the first time, sometimes the surveys are not fetched from storage, so we need to fetch them from the API
-    // because the remote config is still being fetched from the API
     posthog
       .ready()
+      .then(() => posthog.onSurveysReady())
```
ioannisj: This is were we wait for survey to be available before continuing

marandaneto: one question, since users have to wrap their app with `PostHogProvider`, this wont drop frames of the hosting app until we wait for surveys data right?

packages/react-native/src/posthog-rn.ts
```diff
@@ -861,20 +880,139 @@ export class PostHog extends PostHogCore {
...
+    try {
+      const surveysFromApi = await super.getSurveysStateless()
+      if (surveysFromApi && surveysFromApi.length > 0) {
+        this._cacheSurveys(surveysFromApi, 'API')
+      } else {
+        this._cacheSurveys(null, 'API')
+      }
+    } catch (error) {
+      this.logMsgIfDebug(() => console.error('PostHog Debug', 'Error loading surveys from API:', error))
+      this._cacheSurveys(null, 'API (error)')
```
marandaneto: not sure we should clear the cache here
if the `getSurveysStateless` throws an error because of no connection, we should let the user still use cached surveys
we should only clear cache if we get certain types of errors eg 4xx or 5xx maybe?

packages/react-native/src/posthog-rn.ts
```diff
@@ -861,20 +880,139 @@ export class PostHog extends PostHogCore {
...
+          return
+        }
+
+        // Handle surveys if they exist in the response
+        const responseWithSurveys = flagsResponse as any
+        if (responseWithSurveys?.surveys) {
+          const surveys = responseWithSurveys.surveys
+
+          // If surveys is not an array, it means there are no surveys (its a boolean)
+          if (Array.isArray(surveys)) {
```
marandaneto: we can checl the length here as well as we do with `_loadSurveysFromAPI`

packages/react-native/src/posthog-rn.ts
```diff
@@ -861,20 +880,139 @@ export class PostHog extends PostHogCore {
...
+            this._cacheSurveys(surveys as Survey[], 'flags endpoint')
+          } else {
+            this._cacheSurveys(null, 'flags endpoint')
+          }
+        } else {
+          // If no surveys in flags response, try loading from API as fallback
+          this.logMsgIfDebug(() => console.log('PostHog Debug', 'No surveys in flags response, loading from API'))
+          await this._loadSurveysFromAPI()
```
marandaneto: isnt this `else` condition the case if `surveys=false` which means we dont have any surveys to be shown or its disabled?
if thats the case, we should not call the API anymore

packages/react-native/src/posthog-rn.ts
```diff
@@ -861,20 +880,139 @@ export class PostHog extends PostHogCore {
...
+    if (this._disableSurveys === true) {
+      this.logMsgIfDebug(() => console.log('PostHog Debug', 'Loading surveys skipped, disabled.'))
+      this._cacheSurveys(null, 'remote config (disabled)')
+      return
+    }
+
+    const surveys = response.surveys
+
+    // If surveys is not an array, it means there are no surveys (its a boolean)
+    if (Array.isArray(surveys)) {
```
marandaneto: same as https://github.com/PostHog/posthog-js/pull/2415/files#r2416102665

packages/react-native/src/posthog-rn.ts
```diff
@@ -861,20 +880,139 @@ export class PostHog extends PostHogCore {
...
+
+    this.logMsgIfDebug(() => console.log('PostHog Debug', 'No surveys found in storage'))
+    return []
+  }
+
+  /**
+   * Returns a promise that resolves when surveys are ready to be loaded.
+   * If surveys are already loaded and ready to go, returns a resolved promise instead.
+   */
+  public onSurveysReady(): Promise<void> {
```
marandaneto: does this need to be public?

ioannisj: No not really, will fix

packages/react-native/src/posthog-rn.ts
ioannisj: Since we don't control loading flags with config=true when calling `flagsAsync` (I didn't want to touch the web sdk API here), if for whatever reason in the future /flags is called without config=true, the surveys key will be missing. So added this fallback here instead

packages/react-native/src/surveys/PostHogSurveyProvider.tsx
ioannisj: Hm, no idea. I'll need to test this out actually

ioannisj: Testing this and it seems that the whole useEffect() is non-blocking. Added a delay in remote config fetch and UI remained responsive, so I think we are okay using this

packages/react-native/src/posthog-rn.ts
ioannisj: I think this is now marked as internal. Not completely sure if this is the way or if internal methods should be _underscored

packages/react-native/src/posthog-rn.ts
marandaneto: > but response type omits surveys intentionally

back then the `config=true` didnt exist, so if you think thats useful, you can add it back

packages/react-native/src/posthog-rn.ts
marandaneto: you dont have to touch the Web SDK, the core and RN SDK are isolated from the JS SDK

packages/react-native/src/posthog-rn.ts
marandaneto: `@internal` is just an annotation signal, the method is still public but might be ok
try to declare it as private or with underscore and if its ok go with one of these options, otherwise the option above is fine since we've used before

packages/react-native/src/posthog-rn.ts
```diff
@@ -965,21 +965,14 @@ export class PostHog extends PostHogCore {
...
-            this._cacheSurveys(surveys as Survey[], 'flags endpoint')
-          } else {
-            this._cacheSurveys(null, 'flags endpoint')
-          }
+        // If surveys is not an array, it means there are no surveys (its a boolean)
+        if (Array.isArray(surveys) && surveys.length > 0) {
+          this._cacheSurveys(surveys as Survey[], 'flags endpoint')
         } else {
-          // If no surveys in flags response, try loading from API as fallback
-          this.logMsgIfDebug(() => console.log('PostHog Debug', 'No surveys in flags response, loading from API'))
```
marandaneto: can we keep the logs?

packages/react-native/src/posthog-rn.ts
```diff
@@ -179,14 +182,30 @@ export class PostHog extends PostHogCore {
       this._isInitialized = true
 
       if (this._disableRemoteConfig === false) {
-        this.reloadRemoteConfigAsync()
+        this.reloadRemoteConfigAsync().then((response) => {
+          if (response) {
+            this._handleSurveysFromRemoteConfig(response)
+          }
+          this._notifySurveysReady()
+        })
```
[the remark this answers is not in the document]
marandaneto: should we do this @ioannisj ?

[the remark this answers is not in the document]
marandaneto: there are similar comments from the bot

packages/react-native/src/posthog-rn.ts
```diff
@@ -179,14 +182,30 @@ export class PostHog extends PostHogCore {
...
       } else {
-        this._logger.info('Remote config is disabled.')
+        this.logMsgIfDebug(() => console.info('PostHog Debug', `Remote config is disabled.`))
```
marandaneto: have you pulled main into your branch? there was a logging refactor, i think you should use `this._logger` instead

packages/react-native/src/posthog-rn.ts
marandaneto: can you confirm this is correct?

marandaneto (approved):
left a few comments otherwise lgtm

packages/react-native/src/posthog-rn.ts
ioannisj: Tested with a very simple survey and it still gets some internal flags. Then survey won't display until flags are manually reloaded. So afaik the comment here stands

packages/react-native/src/posthog-rn.ts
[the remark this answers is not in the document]
ioannisj: Makes sense. I would trust the bot here

PR #2416, opened 2025-10-08
chore: upgrade to jest 29

pauldambra:
what's one better than 28? 

29!


packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
rafaeelaudibert, line 1: Oh my god, this is glorious, I hate updating this file. Does it work, though? The quotes are kinda ambiguous now

PR #2417, opened 2025-10-08
feat(flags): add evaluation_environments to node and react-native SDKs

dmarticus:
## Changes

This PR implements support for evaluation environments, allowing users to specify which environment tags their SDK instance should use when evaluating feature flags.

Key changes:
  - Added evaluation_environments configuration option to the configs for Node and React-Native that accepts an array of environment tags (e.g., `['production', 'web', 'checkout']`)
  - Updated methods that call the `/flags` endpoint to include `evaluation_environments` in the request payload when configured
  - Added tests to verify the new functionality

This implementation follows the specification outlined in:
  - PostHog documentation PR: https://github.com/PostHog/posthog.com/pull/12929
  - Related internal PRs: https://github.com/PostHog/posthog/pull/38681 and https://github.com/PostHog/posthog/pull/37719

When configured, only feature flags that have at least one matching evaluation tag will be evaluated. Feature flags with no evaluation tags will always be evaluated for backward compatibility.

Basically, I'm doing what I did [here](https://github.com/PostHog/posthog-js/pull/2369), in response to a [customer request](https://posthog.slack.com/archives/C06KJQ2RH1A/p1753773309230879).

This should be perfectly safe; this field is backward compatible and won't affect existing versions of the app.

### DIFF
--- a/packages/node/src/__tests__/posthog-node.spec.ts
+++ b/packages/node/src/__tests__/posthog-node.spec.ts
@@ -1,10 +1,14 @@
-import { PostHog } from '@/entrypoints/index.node'
+import { PostHog, PostHogOptions } from '@/entrypoints/index.node'
 import { anyFlagsCall, anyLocalEvalCall, apiImplementation, isPending, wait, waitForPromises } from './utils'
 import { randomUUID } from 'crypto'
 
 jest.mock('../version', () => ({ version: '1.2.3' }))
 
 const mockedFetch = jest.spyOn(globalThis, 'fetch').mockImplementation()
+
+const posthogImmediateResolveOptions: PostHogOptions = {
+  fetchRetryCount: 0,
+}
 
 const waitForFlushTimer = async (): Promise<void> => {
   await waitForPromises()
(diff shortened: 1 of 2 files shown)


packages/core/src/types.ts
```diff
@@ -55,6 +55,17 @@ export type PostHogCoreOptions = {
...
+   * Evaluation environments for feature flags.
+   * When set, only feature flags that have at least one matching evaluation tag
+   * will be evaluated for this SDK instance. Feature flags with no evaluation tags
+   * will always be evaluated.
+   *
+   * Examples: ['production', 'web', 'mobile']
+   *
+   * @default undefined
+   */
+  evaluationEnvironments?: readonly string[]
```
marandaneto, line 68: curious why its `readonly`?

packages/react-native/src/types.ts
```diff
@@ -150,3 +150,17 @@ export interface PostHogCustomStorage {
   getItem: (key: string) => string | null | Promise<string | null>
   setItem: (key: string, value: string) => void | Promise<void>
 }
+
+export interface PostHogRNOptions {
```
marandaneto: curious why we created a new interface instead of just adding to the current `PostHogOptions` interface?

packages/core/src/types.ts
dmarticus: it's a sensitive field, since it directly determines which flags will end up evaluating on your site, and so I don't want it to be accidentally mutated without at least throwing a compiler error.

PR #2421, opened 2025-10-09
fix: chasing memory leaks in the recorder

pauldambra:
reported via tooter DM that a memory leak using posthog was cleared for them when blocking the recorder

went poking around to see if i could spot any place that might be happening

### DIFF
--- a/packages/node/src/__tests__/posthog-node.spec.ts
+++ b/packages/node/src/__tests__/posthog-node.spec.ts
@@ -25,6 +25,8 @@ const getLastBatchEvents = (): any[] | undefined => {
   return JSON.parse((call[1] as any).body as any).batch
 }
 
+jest.retryTimes(3)
+
 describe('PostHog Node.js', () => {
   let posthog: PostHog
 


packages/browser/src/extensions/replay/external/mutation-throttler.ts
```diff
@@ -94,4 +94,13 @@ export class MutationThrottler {
         }
         return event
     }
+
+    public reset() {
+        this._loggedTracker = {}
```
pauldambra, line 99: > Surely this wasn't leaking more than a couple bytes, right? Anyway, looks like an improvement!

@rafaeelaudibert yeah, this would just be a map of number to bool  so it could end up with a lot of numbers over a long session 

but i can't see it being a big deal

not calling stop potentially worse 
but yeah, really just poking with a stick 🙈

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -702,10 +703,11 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
                 this._isIdle = 'unknown'
                 this.stop()
                 // then we want a session id listener to restart the recording when a new session starts
-                const waitForSessionIdListener = this._sessionManager.onSessionId(
+                this._forceIdleSessionIdListener = this._sessionManager.onSessionId(
```
pauldambra, line 706: clearing listeners in posthog app has been pretty effective for reducing memory leaks there

so the listener and the timer here could be problematic if they end up keeping references to DOM nodes or large arrays

PR #2422, opened 2025-10-10
fix: potentially leaked timers when switching cookieless

pauldambra:
when switching cookieless modes we leave behind an instance of the session id manager
or rather we replace it, but not its timers
so first destroy those times and then replace it

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -2972,6 +2972,7 @@ export class PostHog {
                 distinct_id: COOKIELESS_SENTINEL_VALUE,
                 $device_id: null,
             })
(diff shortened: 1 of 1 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -2905,6 +2905,7 @@ export class PostHog {
             // If the user has explicitly opted out on_reject mode, then before we can start sending regular non-cookieless events
             // we need to reset the instance to ensure that there is no leaking of state or data between the cookieless and regular events
             this.reset(true)
+            this.sessionManager?.destroy()
```
pauldambra, line 2908: @robbie-c I don't think I'm changing any cookieless behaviour with this cleanup?

PR #2424, opened 2025-10-10
feat(web-vitals): apply masking to $current_url in web vitals

thmsobrmlr:
## Problem

We have a config option that allows masking senstive query params in the URL, but it wasn't applied to the URLs in the web vitals extension.

Related customer issue: https://posthoghelp.zendesk.com/agent/tickets/39027. (moved to PostHog: https://us.posthog.com/project/2/support/tickets/41673)

Related PR for masking in the heatmaps extension: https://github.com/PostHog/posthog-js/pull/2425.

## Changes

Adds support for masking URLs with the web vitals extension.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/web-vitals/index.ts
+++ b/packages/browser/src/extensions/web-vitals/index.ts
@@ -4,6 +4,9 @@ import { createLogger } from '../../utils/logger'
 import { isBoolean, isNullish, isNumber, isUndefined, isObject } from '@posthog/core'
 import { WEB_VITALS_ALLOWED_METRICS, WEB_VITALS_ENABLED_SERVER_SIDE } from '../../constants'
 import { assignableWindow, window, location } from '../../utils/globals'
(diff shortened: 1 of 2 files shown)


packages/browser/src/extensions/web-vitals/index.ts
thmsobrmlr, line 200: My main question is if we want to compare the unmasked URL here, and thus emit web vitals events when changing masked query parameters.

rafaeelaudibert (approved):
We should probably find a higher-up abstraction for this and mask this directly on the `capture` call to avoid having to add this on every code path, but this is fine for now

PR #2430, opened 2025-10-13
chore: ignore posthogs own exceptions

daibhin:
## Problem

Related to https://github.com/PostHog/post-mortems/pull/3

## Changes

We had another incident largely caused by capturing exceptions that were thrown by the PostHog SDK. Let's disable the capture of PostHog exceptions by default with a config option if customers do want to enable it.

This is a relatively rudimentary implementation that looks to see if the kaboom frame comes from the PostHog SDK

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config


packages/browser/src/types.ts
```diff
@@ -1066,6 +1066,14 @@ export interface ErrorTrackingOptions {
      * @default false
      */
     captureExtensionExceptions?: boolean
+
+    /**
+     * ADVANCED: determines whether exception caused by the PostHog SDK will be captured
```
oliverb123: Maybe `UNSTABLE`? - I'd like the ability to remove this in the future if we decide to, or change it

packages/browser/src/posthog-exceptions.ts
```diff
@@ -117,14 +127,23 @@ export class PostHogExceptions {
...
-    private _isExtensionException(properties: Properties): boolean {
-        const exceptionList = properties.$exception_list
+    private _isExtensionException(exceptionList: ErrorTracking.ExceptionList): boolean {
+        const frames = exceptionList.flatMap((e) => e.stacktrace?.frames ?? [])
+        return frames.some((f) => f.filename && f.filename.startsWith('chrome-extension://'))
+    }
 
-        if (!exceptionList || !isArray(exceptionList)) {
-            return false
+    private _isPostHogException(exceptionList: ErrorTracking.ExceptionList): boolean {
```
oliverb123, line 135: Maybe a silly question but worth double checking - this handles synthetic exceptions properly?

packages/browser/src/posthog-exceptions.ts
```diff
@@ -117,14 +127,23 @@ export class PostHogExceptions {
...
+        if (exceptionList.length > 0) {
+            const exception = exceptionList[0]
+            const frames = exception.stacktrace?.frames ?? []
+            const lastFrame = frames[frames.length - 1]
+            return lastFrame?.filename?.includes('posthog.com/static') ?? false
```
marandaneto, line 140: this will only work if people are using the script and not the npm package
this will only work if the exception is thrown in the main bundle and not by one of our extensions (otherwise the exception will be [ScriptError - no stack trace](https://github.com/PostHog/posthog/pull/39349#discussion_r2416207217)) (iirc the last 2 incidents the error was thrown by the extension)
this will only work if people are not using the script behind a proxy

imo the best approach is to swallow errors and rethrow if needed

eg
```
initSdk() {
  try {
    // do everything
  } catch (e: error) {
    // swallow, log, and rethrow if capturePostHogExceptions is enabled
  }
}
```

this would need to be done in every public method and every async method that is not awaited otherwise errors will be thrown and not handled anyway.
TLDR: always wrap the facing methods with a try catch

packages/browser/src/posthog-exceptions.ts
daibhin: Should be handled because we remove the first frame from the stacktrace if we're using the synthetic exception:
https://github.com/PostHog/posthog-js/blob/1b1d0abf46b7e2cf4a2d47fcadd95ca57489aa08/packages/core/src/error-tracking/error-properties-builder.ts#L77

(Second argument of the `stackParser` function is number of lines to skip)

packages/browser/src/posthog-exceptions.ts
daibhin: Exceptions from extensions will not always be `ScriptError`, it depends on [how the script is loaded](https://posthog.com/docs/error-tracking/common-questions#what-is-a-script-error-with-no-stack-traces). In the snippet we share we tell people to set `crossorigin="anonymous"` so that exceptions from PostHog have a stack trace.

This is an example of the type of exception from our own SDK that this would have captured: https://us.posthog.com/project/2/error_tracking/0199a9b7-2a5e-78f2-9eae-3a4427391c82?timestamp=2025-10-08T22:39:16.640000-07:00. If you dig into the properties the `filename` of the `junk_drawer` on the root frame has a value: `"https://internal-t.posthog.com/static/surveys.js?v=1.257.0"`

Totally agree though that we can take this even further and wrap methods with global try / catch methods. I'll add that in a follow up PR

PR #2434, opened 2025-10-14
fix: allow turning off debug logging

pauldambra:
i think that if you entered debug logging you couldn't turn it off

### DIFF
--- a/.changeset/shaggy-meals-kick.md
+++ b/.changeset/shaggy-meals-kick.md
@@ -2,4 +2,4 @@
 'posthog-js': patch
 ---
 
-fix: allow turning off debug mode"
+fix: allow turning off debug mode


packages/browser/src/posthog-core.ts
```diff
@@ -2515,13 +2516,20 @@ export class PostHog {
...
+                if (this.config.debug) {
+                    Config.DEBUG = true
+                    localStorage && localStorage.setItem('ph_debug', 'true')
+                    logger.info('set_config', {
+                        config,
+                        oldConfig,
+                        newConfig: { ...this.config },
+                    })
+                } else {
+                    Config.DEBUG = false
```
pauldambra, line 2535: the logger checks if `Config.DEBUG` and there's no way to set that to false if it is already true
so, even if this isn't the problem I'm looking at, it's a potential problem

PR #2435, opened 2025-10-14
feat: add option to capture pageviews on navigation history transitions

lricoy:
## Changes

Adds a new `'on_navigation'` option for the `capture_pageleave` config that captures pageleave events on every history navigation event (pushState, replaceState, popstate) in addition to window unload.

This is an extension to the `capture:_pageview: 'history_change'` option to allow people who rely on client-side navigation to have this behavior:

When `capture_pageleave: 'on_navigation'` is set:
- pageView A → (navigate) → pageLeave A + pageView B → (navigate) → pageLeave B + pageView C -> (closes tab) -> pageLeave C

The regular `true` or `if_capture_pageview` would be:
- pageView A → (navigate) → pageView B → (navigate) → pageView C -> (closes tab) -> pageLeave C

The pageleave event is captured before the new pageview during navigation and has a timestamp as well as all the other expected props because this PR only changes the trigger, everything is still handled by `PageViewManager`.

⚠️ This requires docs update on posthog.com to clarify the different `$pageleave` behavior.

### DIFF
--- a/.changeset/swift-bottles-design.md
+++ b/.changeset/swift-bottles-design.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': minor
+---
+
+Includes a option to capture pageleave events on history navigation
--- a/packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
+++ b/packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
@@ -1,4 +1,4 @@
-// Jest Snapshot v1, https://jestjs.io/docs/snapshot-testing
+// Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`config snapshot for PostHogConfig 1`] = `
 "{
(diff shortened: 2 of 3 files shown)


packages/browser/src/extensions/history-autocapture.ts
```diff
@@ -90,6 +90,16 @@ export class HistoryAutocapture {
 
             // Only capture pageview if the pathname has changed and the feature is enabled
             if (currentPathname !== this._lastPathname && this.isEnabled) {
+                // Capture pageleave for the previous page before capturing the new pageview
+                // PageViewManager will handle checking if there's a previous pageview to leave
+                if (this._instance._shouldCapturePageleaveOnNavigation()) {
+                    const pageleaveProps = this._instance.pageViewManager.doPageLeave(new Date())
+                    // Only capture pageleave if there was a previous page ($prev_pageview_id will be present)
+                    if (pageleaveProps.$prev_pageview_id) {
```
jabahamondes, line 98: When could this not be the case? (prev_pageview_id is not present)

Issue #2437, opened 2025-10-14
Unable to resolve "@posthog/core/error-tracking" from "node_modules/posthog-react-native/dist/error-tracking/index.js"


idrisssakhi:
## Bug description

Unable to resolve "@posthog/core/error-tracking" from "node_modules/posthog-react-native/dist/error-tracking/index.js"
[...]

PR #2459, opened 2025-10-20
chore: lint against subpath imports

daibhin:
## Problem

Importing from `@posthog/core/error-tracking` caused this issue: https://github.com/PostHog/posthog-js/issues/2437. Fixed by removing the subpath import in https://github.com/PostHog/posthog-js/pull/2457

Adding a linting rule to make sure this is done properly next time and no one else falls into the same trap


marandaneto:
@hpouillot @daibhin do we need this after https://github.com/PostHog/posthog-js/pull/2461 ?

daibhin:
@marandaneto no reason not to I suppose, so going to get it in

PR #2462, opened 2025-10-20
chore(llma): fix automated PRs

carlos-marchal-ph:
This should hopefully get us where we can just fully automate the process:
- Does not suggest major upgrades
- Correctly tags us using code owners file instead of deprecated reviewers field
- Automatically generates the changeset contents

There are some major upgrades for the SDKs pending too, we should double check if they are breaking changes or not and decide what to do about them (might make sense to upgrade major).

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -16,3 +16,6 @@ packages/react-native/                                 @PostHog/team-mobile
 examples/example_rn_macos/                             @PostHog/team-mobile
 packages/core/                                         @PostHog/team-mobile
 packages/example-expo-53/                              @PostHog/team-mobile
+
+# AI
+/packages/ai/package.json                              @PostHog/team-llm-analytics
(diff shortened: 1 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -16,3 +16,6 @@ packages/react-native/                                 @PostHog/team-mobile
 examples/example_rn_macos/                             @PostHog/team-mobile
 packages/core/                                         @PostHog/team-mobile
 packages/example-expo-53/                              @PostHog/team-mobile
+
+# AI
+/packages/ai/package.json                              @PostHog/team-llm-analytics
```
Radu-Raicea, line 21: Just the package.json?

carlos-marchal-ph: I can add ourselves to the entire directory but for the bot tagging us for updates this is enough

Radu-Raicea (approved):

PR #2467, opened 2025-10-20
docs: adds warning to advanced_disable_flags usage

lricoy:
## Problem

Got a Zendesk [ticket](https://posthoghelp.zendesk.com/agent/tickets/40487) that was using this flag, and I was also surprised that it disabled RemoteConfig, although it made sense afterward... So I guess we can be more explicit about this. 

## Changes

Update docstring to include a Warning.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/types.ts
+++ b/packages/browser/src/types.ts
@@ -727,10 +727,10 @@ export interface PostHogConfig {
      * and reduce resource usage on both the client and the server.
      *
      * WARNING: Disabling this will also prevent remote configuration from loading,
-     * which could mean features like web vitals, surveys, and other server-side controlled
-     * features will be disabled unless explicitly enabled via client-side config.
+     * which could mean features like web vitals, surveys, and other features configured
+     * in PostHog settings are disabled unless explicitly enabled via client-side config.
      * When setting this to true, make sure to explicitly configure any features you
-     * want to use (e.g., capture_performance, autocapture, etc.) in the init config.
+     * want to use (e.g., capture_performance, autocapture, etc.) in your SDK's init config.
      *
      * @default false
      */


packages/browser/src/types.ts
```diff
@@ -726,6 +726,12 @@ export interface PostHogConfig {
      * you may wish to turn off the call completely to avoid an extra request
      * and reduce resource usage on both the client and the server.
      *
+     * WARNING: Disabling this will also prevent remote configuration from loading,
+     * which could mean features like web vitals, surveys, and other server-side controlled
+     * features will be disabled unless explicitly enabled via client-side config.
+     * When setting this to true, make sure to explicitly configure any features you
+     * want to use (e.g., capture_performance, autocapture, etc.) in the init config.
+     *
```
lricoy, line 734: I could really use some help with the wording here. My brain is melted already 🧠 🧀 🔥

.changeset/rare-cougars-lie.md
gewenyu99, line 1: lmao

packages/browser/src/types.ts
gewenyu99:
```suggestion
     * WARNING: Disabling this will also prevent remote configuration from loading,
     * which could mean features like web vitals, surveys, and other features configured
     * in PostHog settings are disabled unless explicitly enabled via client-side config.
     * When setting this to true, make sure to explicitly configure any features you
     * want to use (e.g., capture_performance, autocapture, etc.) in your SDK's init config.
     *
```

gewenyu99: That's a pretty good attempt, I think this gets the point across

PR #2469, opened 2025-10-21
fix: don't track heatmap buffer interval while hidden

pauldambra:
we observed heatmaps trying to send the buffer every 5 seconds
even in the background
yuck

let's pause and resume on visibility change

also tears down listeners on stop for tidiness

### DIFF
--- a/packages/browser/playwright/mocked/heatmaps.spec.ts
+++ b/packages/browser/playwright/mocked/heatmaps.spec.ts
@@ -1,0 +1,124 @@
+import { expect, test } from './utils/posthog-playwright-test-base'
+import { start } from './utils/setup'
+import { pollUntilEventCaptured } from './utils/event-capture-utils'
+
+const startOptions = {
+    options: {
+        capture_heatmaps: {
+            flush_interval_milliseconds: 1000, // Short interval for testing
+        },
+    },
+    url: '/playground/cypress/index.html',
+}
+
+test.describe('Heatmaps', () => {
+    test('captures click and mousemove events', async ({ page, context }) => {
+        await start(startOptions, page, context)
+
+        // Reset to clear any initial events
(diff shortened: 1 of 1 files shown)


packages/browser/src/heatmaps.ts
```diff
@@ -53,10 +53,13 @@ export class Heatmaps {
     _initialized = false
     _mouseMoveTimeout: ReturnType<typeof setTimeout> | undefined
 
-    // TODO: Periodically flush this if no other event has taken care of it
     private _buffer: HeatmapEventBuffer
     private _flushInterval: ReturnType<typeof setInterval> | null = null
     private _deadClicksCapture: DeadClicksAutocapture | undefined
+    private _onClick_handler: ((e: Event) => void) | undefined
```
lricoy: Why the mixed cases instead of `_onClickHandler`?

packages/browser/src/heatmaps.ts
```diff
@@ -144,9 +161,40 @@ export class Heatmaps {
...
         this._initialized = true
     }
 
+    private _removeListeners(): void {
+        if (!window || !document) {
+            return
+        }
+
+        if (this._flush_handler) {
+            window.removeEventListener('beforeunload', this._flush_handler)
```
lricoy: Oh, I see, maybe it is a pattern already.

lricoy:
LGTM, just a question out of curiosity

packages/browser/src/heatmaps.ts
pauldambra: 🤣 🤣 🤣 🤣 🤣 🤣 
no, that's 100% me not carefully reading AI generated code

i owe you a drink!

lricoy: More Piña coladas next year!

PR #2484, opened 2025-10-22
chore: add hermes symbolication support for react native and error tracking

marandaneto:
## Problem

Relates to https://github.com/PostHog/posthog/pull/39998
Uses expo debug id support https://github.com/expo/expo/pull/25935 and https://github.com/expo/expo/pull/25859
https://docs.expo.dev/versions/latest/config/metro/#source-map-debug-id

TODO:
- [X] xcode build script
- [X] android build plugin
- [x] metro/expo plugin for injecting chunk ids
- [ ] add escape hatch for allow failure
- [x] RN SDK has to ship the scripts to npm as well
- [x] test posthog-cli call android
- [x] test posthog-cli call ios
- [ ] posthog-wizard injects xcode and gradle step
- [ ] docs

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -219,7 +219,7 @@ limitations under the License.
 
 ---
 
-Some files in this codebase contain code from getsentry/sentry-javascript by Software, Inc. dba Sentry.
+Some files in this codebase contain code from getsentry/sentry-javascript or getsentry/sentry-react-native by Software, Inc. dba Sentry.
 In such cases it is explicitly stated in the file header. This license only applies to the relevant code in such cases. 
 
 MIT License
@@ -232,6 +232,60 @@ the Software without restriction, including without limitation the rights to
 use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
 of the Software, and to permit persons to whom the Software is furnished to do
 so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+---
+
+Some files in this codebase contain code from facebook/metro by Meta Platforms, Inc. and affiliates.
+In such cases it is explicitly stated in the file header. This license only applies to the relevant code in such cases. 
+
+MIT License
+
+Copyright (c) Meta Platforms, Inc. and affiliates.
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
+
+---
+
+Some files in this codebase contain code from expo/expo by 650 Industries, Inc. (aka Expo).
+In such cases it is explicitly stated in the file header. This license only applies to the relevant code in such cases. 
+
+The MIT License (MIT)
+
+Copyright (c) 2015-present 650 Industries, Inc. (aka Expo)
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
 
 The above copyright notice and this permission notice shall be included in all
 copies or substantial portions of the Software.
--- a/examples/example-expo-53/README.md
+++ b/examples/example-expo-53/README.md
@@ -53,4 +53,7 @@ cd android
 set -o pipefail && xcrun xcodebuild clean build -workspace ios/exampleexpo53.xcworkspace -scheme exampleexpo53 -configuration Release -destination generic/platform=ios | xcpretty
 
 # Also: Xcode -> Product -> Archive
+
+# web
+npx expo export --clear --source-maps --platform web
 ```


examples/example-expo-53/android/app/build.gradle
```diff
@@ -81,6 +81,8 @@ def enableProguardInReleaseBuilds = (findProperty('android.enableProguardInRelea
  */
 def jscFlavor = 'io.github.react-native-community:jsc-android:2026004.+'
 
+apply from: "../../node_modules/posthog-react-native/tooling/posthog.gradle"
```
marandaneto, line 84: this has to be done via wizard or manoelly

examples/example-expo-53/ios/exampleexpo53.xcodeproj/project.pbxproj
```diff
@@ -216,7 +215,7 @@
...
...olveAppEntry')\" \"$PROJECT_ROOT\" ios absolute | tail -n 1)\"\nfi\n\nif [[ -z \"$CLI_PATH\" ]]; then\n  # Use Expo CLI\n  export CLI_PATH=\"$(\"$NODE_BINARY\" --print \"require.resolve('@expo/cli', { paths: [require.resolve('expo/package.json')] })\")\"\nfi\nif [[ -z \"$BUNDLE_COMMAND\" ]]; then\n  # Default Expo CLI command for bundling\n  export BUNDLE_COMMAND=\"export:embed\"\nfi\n\n# Source .xcode.env.updates if it exists to allow\n# SKIP_BUNDLING to be unset if needed\nif [[ -f \"$PODS_ROOT/../.xcode.env.updates\" ]]; then\n  source \"$PODS_ROOT/../.xcode.env.updates\"\nfi\n# Source local changes to allow overrides\n# if needed\nif [[ -f \"$PODS_ROOT/../.xcode.env.local\" ]]; then\n  source \"$PODS_ROOT/../.xcode.env.local\"\nfi\n\n# `\"$NODE_BINARY\" --print \"require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'\"`\n\nset -e\n\n# RN 0.69 - 0.80.0\nWITH_ENVIRONMENT=\"../node_modules/react-native/scripts/xcode/with-environment.sh\"\nPOSTHOG_XCODE=\"../node_modules/posthog-react-native/tooling/posthog-xcode.sh\"\n\nBUNDLE_REACT_NATIVE=\"/bin/sh $POSTHOG_XCODE\"\n/bin/sh -c \"$WITH_ENVIRONMENT \\\"$BUNDLE_REACT_NATIVE\\\"\"\n";
```
marandaneto: this has to be done via wizard or manoelly

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -0,0 +1,48 @@
...
+REACT_NATIVE_XCODE_DEFAULT="../node_modules/react-native/scripts/react-native-xcode.sh"
+REACT_NATIVE_XCODE="${1:-$REACT_NATIVE_XCODE_DEFAULT}"
+
+# TODO: when to call (inject): --release -- exp hermes inject --project my-app --directory $DERIVED_FILE_DIR
+# TODO: what should i set for --project?
+
+# TODO: when to call (clone): --release -- exp hermes clone --directory $DERIVED_FILE_DIR
+
+# [ -z "$SOURCEMAP_FILE" ] && export SOURCEMAP_FILE="$DERIVED_FILE_DIR/main.jsbundle.map"
+# TODO: is the CLI clever enough to pick the right file? there will be a bunch of non related files
```
marandaneto: figure out TODOs, same things on Android btw

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -0,0 +1,48 @@
...
+ARGS="--release -- exp hermes upload --directory $DERIVED_FILE_DIR"
+
+# requires posthog-cli installed
+# requires authentication (posthog-cli login) or:
+# https://github.com/PostHog/posthog/tree/master/cli#env-based-authentication
+REACT_NATIVE_XCODE_WITH_POSTHOG="posthog-cli $ARGS \"$REACT_NATIVE_XCODE\""
+
+exitCode=0
+
+# TODO: implement ALLOW_FAILURE CLI config to keep building if there are errors
```
marandaneto: implement easy CLI escape hatch otherwise the build will never succeed, eg no connection, posthog is down, etc
failing silently is also not a good option because that might mean a production release without source maps
so users could easily flip this flag to keep working temporarily

packages/react-native/tooling/posthog.gradle
```diff
@@ -0,0 +1,309 @@
...
+                        doLast {
+                            injected.execOps.exec {
+                                workingDir reactRoot
+
+                                def cliPackage = "posthog-cli"
+                                def args = [cliPackage]
+
+                                // bundleOutput and sourcemapOutput
+                                def sourcemapDir = new File(sourcemapOutput).getParent()
+                                args.addAll(["--release", "--",
```
marandaneto: afaict git info will be gotten automatically, this is cool but cannot be the only option
mobile has the concept of build variants, think of white label apps or uber driver vs uber client, paid vs free
the very same commit will generate multiple valid applications that have different code paths
so we will need to pass release info anyways
on mobile this would be the equivalent of a 'distribution'
eg an app namespace=com.posthog.app, version=1.0.0, build=1, distribution=paid
those distributions will generate different apps - consequently different sourcemaps

marandaneto: we should use the `project` here which is not the same as posthog project, so we could let users set the project or we could concatenate our own dist project, as the example above

examples/example-expo-53/metro.config.js
```diff
@@ -0,0 +1,3 @@
+const { getPostHogExpoConfig } = require('posthog-react-native/metro')
```
marandaneto, line 1: this has to be done via wizard or manoelly

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -0,0 +1,105 @@
...
+    # Fallback to searching common locations
+    export PATH="/usr/local/bin:/opt/homebrew/bin:$HOME/.cargo/bin:$HOME/.local/bin:$HOME/.posthog:$PATH"
+    PH_CLI_PATH=$(command -v posthog-cli 2>/dev/null)
+  fi
+fi
+
+if [ -z "$PH_CLI_PATH" ] || [ ! -x "$PH_CLI_PATH" ]; then
+  echo "error: posthog-cli not found"
+  exit 1
+fi
```
marandaneto, line 62: the script should be able to find the posthog-cli, and i am not 100% confident this would work, any suggestions?
i asked claude to generate this based on all the options from the download/install script

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -0,0 +1,105 @@
...
+# mimics how the file is defined in node_modules/react-native/scripts/react-native-xcode.sh (PACKAGER_SOURCEMAP_FILE)
+SOURCEMAP_PACKAGER_FILE="$CONFIGURATION_BUILD_DIR/$SOURCEMAP_NAME"
+
+# RN deletes the PACKAGER_SOURCEMAP_FILE file after execution but we need it
+# lets patch the script to comment out this part if not yet
+if grep -q '^[[:space:]]*rm.*PACKAGER_SOURCEMAP_FILE' "$REACT_NATIVE_XCODE"; then
+  echo "Patching React Native script to preserve sourcemap file..."
+  sed -i '' 's/^[[:space:]]*rm.*PACKAGER_SOURCEMAP_FILE/#&/' "$REACT_NATIVE_XCODE"
+  echo "Patched: commented out rm PACKAGER_SOURCEMAP_FILE line"
+fi
```
marandaneto, line 73: a bit intrusive but needed
we could save the file in a backup and restore after execution but i dont think this change is a big deal anyway

examples/example-expo-53/ios/exampleexpo53.xcodeproj/project.pbxproj
```diff
@@ -216,7 +215,7 @@
...
...y resolution.\n  export ENTRY_FILE=\"$(\"$NODE_BINARY\" -e \"require('expo/scripts/resolveAppEntry')\" \"$PROJECT_ROOT\" ios absolute | tail -n 1)\"\nfi\n\nif [[ -z \"$CLI_PATH\" ]]; then\n  # Use Expo CLI\n  export CLI_PATH=\"$(\"$NODE_BINARY\" --print \"require.resolve('@expo/cli', { paths: [require.resolve('expo/package.json')] })\")\"\nfi\nif [[ -z \"$BUNDLE_COMMAND\" ]]; then\n  # Default Expo CLI command for bundling\n  export BUNDLE_COMMAND=\"export:embed\"\nfi\n\n# Source .xcode.env.updates if it exists to allow\n# SKIP_BUNDLING to be unset if needed\nif [[ -f \"$PODS_ROOT/../.xcode.env.updates\" ]]; then\n  source \"$PODS_ROOT/../.xcode.env.updates\"\nfi\n# Source local changes to allow overrides\n# if needed\nif [[ -f \"$PODS_ROOT/../.xcode.env.local\" ]]; then\n  source \"$PODS_ROOT/../.xcode.env.local\"\nfi\n\n`\"$NODE_BINARY\" --print \"require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'\"`\n\nWITH_ENVIRONMENT=\"../node_modules/react-native/scripts/xcode/with-environment.sh\"\nPOSTHOG_XCODE=\"../node_modules/posthog-react-native/tooling/posthog-xcode.sh\"\n\n/bin/sh -c \"$WITH_ENVIRONMENT \\\"$POSTHOG_XCODE\\\"\"\n";
```
marandaneto, line 218: this has to be done via wizard or manoelly
users have to pretty much add this to the end of their build step (Bundle react native code and images):

```
WITH_ENVIRONMENT="../node_modules/react-native/scripts/xcode/with-environment.sh"
POSTHOG_XCODE="../node_modules/posthog-react-native/tooling/posthog-xcode.sh"

/bin/sh -c "$WITH_ENVIRONMENT \"$POSTHOG_XCODE\""
```

packages/react-native/tooling/posthog-xcode.sh
marandaneto: @oliverb123 i have not found this on the CLI, did you add it?

marandaneto:
> posthog-wizard injects xcode and gradle step

this can be done via expo plugins or wizard
if we do expo plugins, we still have to install the expo plugin
since its only 3 easy patches that you dont need to touch anymore, i will skip this to release a version asap and we circle back if needed for a smoothier installation process

packages/react-native/tooling/posthog-xcode.sh
oliverb123: Yeah, you should be able to do `posthog-cli --no-fail`

.eslintrc.cjs
```diff
@@ -93,5 +93,10 @@ module.exports = {
             },
         },
     ],
-    ignorePatterns: ['node_modules', 'dist'],
+    ignorePatterns: [
+        'node_modules',
+        'dist',
+        'packages/react-native/metro.js',
```
hpouillot: Nit: I would put this ignore patterns inside `packages/react-native/.eslintrc.cjs` and `examples/example-expo-53/eslint.config.js`

packages/react-native/src/tooling/posthogMetroSerializer.ts
```diff
@@ -0,0 +1,120 @@
...
+  debugId?: string
+}): Module[] {
+  if (!debugId) {
+    return premodules
+  }
+
+  const debugIdModuleExists = premodules.findIndex((module) => module.path === DEBUG_ID_MODULE_PATH) != -1
+  if (debugIdModuleExists) {
+    // eslint-disable-next-line no-console
+    console.warn('\n\Chunk ID module found. Skipping PostHog Chunk ID module...\n\n')
```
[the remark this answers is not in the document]
hpouillot: This comment looks right

packages/react-native/tooling/posthog-xcode.sh
hpouillot: I think there is a missing case. If users install cli as a local dep it won't resolve, you can use `npm bin` to get the closest binary folder and add it to PATH. It should something like `node_modules/.bin`

hpouillot: And you can use `npm bin -g` if it's installed globally

packages/react-native/tooling/posthog.gradle
```diff
@@ -0,0 +1,312 @@
...
+                        doFirst {
+                            injected.execOps.exec {
+                                workingDir reactRoot
+
+                                def cliPackage = "posthog-cli"
+                                def args = [cliPackage]
+
+                                args.addAll(["exp", "hermes", "clone",
+                                    "--minified-map-path", packagerSourcemapOutput, // The path to a sourcemap
+                                    "--composed-map-path", sourcemapOutput          // The path of the composed source map
```
hpouillot, line 90: if you have access to it, you can give the --project and --version here. It will create a release and associate the debugId to it

packages/react-native/tooling/posthog.gradle
```diff
@@ -0,0 +1,312 @@
...
+                        doLast {
+                            injected.execOps.exec {
+                                workingDir reactRoot
+
+                                def cliPackage = "posthog-cli"
+                                def args = [cliPackage]
+
+                                def sourcemapDir = sourcemapOutput.getParent()
+                                args.addAll(["exp", "hermes", "upload",
+                                    "--directory", sourcemapDir        // The path to a sourcemap that should be uploaded.
```
hpouillot, line 111: I added an `--include` option on the cli that support glob patterns, if needed you can specify "--include **/sourcemap.map" to target only this file

hpouillot: https://github.com/PostHog/posthog/pull/41608

packages/react-native/src/tooling/vendor/metro/metroBabelTransformer.ts
```diff
@@ -0,0 +1,64 @@
+// Vendored / modified from @facebook/metro
```
hpouillot: is this file used ? I don't see any imports

packages/react-native/src/tooling/utils.ts
```diff
@@ -0,0 +1,204 @@
...
+export function stringToUUID(str: string): string {
+  const md5sum = crypto.createHash('md5')
+  md5sum.update(str)
+  const md5Hash = md5sum.digest('hex')
+
+  // Position 16 is fixed to either 8, 9, a, or b in the uuid v4 spec (10xx in binary)
+  // RFC 4122 section 4.4
+  const v4variant = ['8', '9', 'a', 'b'][md5Hash.substring(16, 17).charCodeAt(0) % 4] as string
+
+  return `${md5Hash.substring(0, 8)}-${md5Hash.substring(8, 12)}-4${md5Hash.substring(
```
hpouillot, line 58: Nit: we could the uuid module to create a stable uuidv5 given a hash. `uuidv5("[HASH]", namespace)`

packages/react-native/package.json
```diff
@@ -5,7 +5,10 @@
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "files": [
-    "dist/"
+    "dist/",
+    "tooling/",
+    "metro.js",
```
hpouillot: If the goal is to be able to import metro.js file as `posthog-react-native/metro` we can maybe create a specific entrypoint (with exports key in package.json) for this file and put the source in tooling.

packages/react-native/package.json
```diff
@@ -5,7 +5,10 @@
...
+    "metro.d.ts"
```
hpouillot: There will be no need for this file if we do that

.eslintrc.cjs
marandaneto: I tried, did not work 😅

packages/react-native/package.json
marandaneto: Thats the goal, yes, is there any other plugin doing this so i can take a look at it? Any pointers is helpful

packages/react-native/tooling/posthog.gradle
marandaneto: I have for Android but not for Xcode so i left both out for now.
I will have to parse the info plist file from the ios dir and infer the version from there, just wanted to ship something first before adding the "nice to have" bits since figuring this out took longer than I expected

packages/react-native/package.json
hpouillot: Yes, you can have a look at the core package.json -> https://github.com/PostHog/posthog-js/blob/main/packages/core/package.json

examples/example-expo-53/export-posthog-cli.sh
```diff
@@ -0,0 +1,5 @@
+# source ./export-posthog-cli.sh
+# https://github.com/PostHog/posthog/tree/master/cli#env-based-authentication
+export POSTHOG_CLI_TOKEN=
+# ManoelTesting org
+export POSTHOG_CLI_ENV_ID=72775
```
daibhin:
```suggestion
export POSTHOG_CLI_ENV_ID=
```

packages/react-native/.eslintrc.cjs
```diff
@@ -23,5 +23,5 @@ module.exports = {
       },
     },
   ],
-  ignorePatterns: ['src/version.ts'],
+  ignorePatterns: ['src/version.ts', 'src/tooling/*', 'tooling/**'],
```
daibhin: Out of curiosity - why ignore the linting here?

packages/react-native/src/tooling/posthogMetroSerializer.ts
```diff
@@ -0,0 +1,120 @@
...
+}: {
+  graph: ReadOnlyGraph<MixedOutput>
+  premodules: Module[]
+  debugId?: string
+}): Module[] {
+  if (!debugId) {
+    return premodules
+  }
+
+  const debugIdModuleExists = premodules.findIndex((module) => module.path === DEBUG_ID_MODULE_PATH) != -1
```
daibhin: Slightly more idiomatic
```suggestion
  const debugIdModuleExists = premodules.some(module => module.path === DEBUG_ID_MODULE_PATH)
```

packages/react-native/src/tooling/posthogMetroSerializer.ts
```diff
@@ -0,0 +1,120 @@
...
+const DEBUG_ID_MODULE_PATH = '__chunkid__'
+
+const SOURCE_MAP_COMMENT = '//# sourceMappingURL='
+const DEBUG_ID_COMMENT = '//# chunkId='
+
+/**
+ * Adds PostHog Chunk ID polyfill module to the bundle.
+ */
+export function unstableBeforeAssetSerializationDebugIdPlugin({
+  premodules,
```
daibhin, line 21: total nit just to be consistent with the variable name below
```suggestion
  preModules,
```

packages/react-native/src/tooling/posthogMetroSerializer.ts
```diff
@@ -0,0 +1,120 @@
...
+ * RAM Bundles do not support custom serializers.
+ */
+export const createPostHogMetroSerializer = (customSerializer?: MetroSerializer): MetroSerializer => {
+  const serializer = customSerializer || createDefaultMetroSerializer()
+  return async function (entryPoint, preModules, graph, options) {
+    if (graph.transformOptions.hot) {
+      return serializer(entryPoint, preModules, graph, options)
+    }
+
+    const debugIdModuleExists = preModules.findIndex((module) => module.path === DEBUG_ID_MODULE_PATH) != -1
```
daibhin:
```suggestion
    const debugIdModuleExists = preModules.some((module) => module.path === DEBUG_ID_MODULE_PATH)
```

packages/react-native/src/tooling/vendor/expo/expoconfig.ts
```diff
@@ -0,0 +1,48 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
daibhin, line 21: I think we should be adding this to the [root level LICENSE file](https://github.com/PostHog/posthog-js/blob/main/LICENSE)

packages/react-native/src/tooling/vendor/metro/countLines.ts
```diff
@@ -0,0 +1,33 @@
...
+// The above copyright notice and this permission notice shall be included in all
+// copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+// SOFTWARE.
```
daibhin, line 25: Same here

packages/react-native/src/tooling/vendor/metro/utils.ts
daibhin, line 1: Not sure if there's a way around this but do we need to be considerate of package size on React Native? Or maybe all this extra vendor stuff isn't even included in the final bundle so it doesn't matter

packages/react-native/.eslintrc.cjs
marandaneto: its an adapted script (vendor or copypasta with moifications) with different linting rules, if i apply all the lints, getting fixes upstream or downstream would be just harder due to needless conflicts

packages/react-native/src/tooling/posthogMetroSerializer.ts
marandaneto: this comes from https://github.com/expo/expo/blob/080fe0db96b352ff3aeddc0e4a8c2c1d84e02f1d/packages/%40expo/metro-config/src/ExpoMetroConfig.ts#L62 so better to keep as is

packages/react-native/src/tooling/utils.ts
marandaneto: https://github.com/expo/expo/blob/080fe0db96b352ff3aeddc0e4a8c2c1d84e02f1d/packages/%40expo/metro-config/src/serializer/debugId.ts#L11-L36
expo does exactly the same so its better to keep them aligned to avoid divergencies across versions since its a stable id

packages/react-native/src/tooling/vendor/metro/utils.ts
marandaneto: i think tree shaking just works here https://docs.expo.dev/guides/tree-shaking/ (didnt check the final bundle tbh)
its just a few interfaces (and a util class) to make it easier for the final user instead of having to add another plugin that ships those classes.
if that becomes a problem, we can easily refactor later

.eslintrc.cjs
marandaneto: fixed

packages/react-native/tooling/posthog-xcode.sh
marandaneto: good catch, done

packages/react-native/src/tooling/vendor/metro/metroBabelTransformer.ts
marandaneto: left over, removed, good catch

packages/react-native/src/tooling/vendor/expo/expoconfig.ts
marandaneto: The MIT license is very permissive and only requires that the original copyright notice and license text be preserved with the code, which the comment accomplishes.
no strong opinion tho, happy to do it if you think its needed

packages/react-native/package.json
marandaneto: noice, done, thanks

packages/react-native/tooling/posthog.gradle
marandaneto: users can force a file name (BUNDLE_NAME env var) so i cannot use this, even the extension isnt mandatory but i bet people do this, will keep the --directory for now until we see its a prolem like CLI loses time opening not related files
Xcode starts a very restricted shell without the PATH and all the exported env var (eg no BUNDLE_NAME out of that context) so i cannot infer the value from there either.

packages/react-native/tooling/posthog.gradle
marandaneto: started a thread about htis explaining how it works on Xcode https://posthog.slack.com/archives/C07AA937K9A/p1763463463959149
can be a follow up PR once we have the CLI bits

packages/react-native/src/tooling/vendor/expo/expoconfig.ts
marandaneto: done

examples/example-expo-53/ios/exampleexpo53.xcodeproj/project.pbxproj
ioannisj: Should this be gated for DEBUG builds? 
Also, could we execute `with-environment.sh` within `posthog-xcode.sh`? I think people would follow this easier  and would be more readable if it's possible

packages/react-native/src/tooling/posthogMetroSerializer.ts
```diff
@@ -0,0 +1,120 @@
...
+    const modifiedPremodules = prependModule(premodules, debugIdModule)
+
+    // Run wrapped serializer
+    const serializerResult = serializer(entryPoint, modifiedPremodules, graph, options)
+    const { code: bundleCode, map: bundleMapString } = await extractSerializerResult(serializerResult)
+
+    // Add Chunk ID comment to the bundle
+    const debugId = determineDebugIdFromBundleSource(bundleCode)
+    if (!debugId) {
+      throw new Error('Chunk ID was not found in the bundle.')
```
ioannisj, line 73: Should we fail the whole build here? Trying to wrap my head around this, when do we expect determineDebugIdFromBundleSource to return null here?

marandaneto: yes, if something went wrong with the debug id injection, the user should fix (or raise an issue with us/expo)/or remove our metro plugin purposes instead of silently failing and surprised by not having human readable stack traces.
this should not happen tho, its a guardrail only to avoid surprises

PR #2500, opened 2025-10-25
fix: pause survey timer when page is hidden

pauldambra:
When the page is hidden we don't need to check if we should display a survey

we can pause and then resume when the page is visible again

(really we should wrap history and run an intersection observer so that this check is reactive instead of active but that's not _essential_


lucasheriques (approved):
ty for the fix. i do want to improve this when i have the time yeah, always felt a bit hacky

PR #2502, opened 2025-10-25
fix: rate limiter can work without a timer

pauldambra:
the bucketed rate limiter has a timer that runs to refill buckets

but since we only care about behaviour at the point we call the rate limiter
we can track the time we last called and then calculate the new token amounts at the point we're called

this stops the interval keeping a page active
and makes the rate limiting entirely reactive

- added a bunch of tests to try and make sure i didn't do something silly
- made the bucketed rate limiter apply refill based on diff between current time and last check time per bucket
- clear buckets on stop


ablaszkiewicz (approved):

packages/core/src/utils/bucketed-rate-limiter.ts
```diff
@@ -1,85 +1,67 @@
...
-      if (newTokens >= this._bucketSize) {
-        delete this._buckets[key]
-      } else {
-        this._setBucket(key, newTokens)
-      }
-    })
+    if (refillIntervals > 0) {
+      const tokensToAdd = refillIntervals * this._refillRate
+      bucket.tokens = Math.min(bucket.tokens + tokensToAdd, this._bucketSize)
+      bucket.lastAccess = bucket.lastAccess + refillIntervals * this._refillInterval
```
hpouillot, line 34: why not assigning now to lastAccess here ?

hpouillot (approved):

packages/core/src/utils/bucketed-rate-limiter.ts
pauldambra: in theory they're equivalent but i think this steps through the intervals more correctly and is more expressive of what it's doing but 🤷

PR #2503, opened 2025-10-25
feat: session recording tracks previous and next session id

pauldambra:
when watching a session it might end because a user went idle, but to the user they're just picking up after a long gap

let's track previous and next session, so we can let watchers navigate backwards and forwards in the UI

we could look up this data but it's relatively slow to query and we can just store it in the SDK instead

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -651,7 +651,9 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         }
 
         // We want to ensure the sessionManager is reset if necessary on loading the recorder
-        this._sessionManager.checkAndGetSessionAndWindowId()
+        const { sessionId, windowId } = this._sessionManager.checkAndGetSessionAndWindowId()
+        this._sessionId = sessionId
+        this._windowId = windowId
 
         if (config?.endpoint) {
             this._endpoint = config?.endpoint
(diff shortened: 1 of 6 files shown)


pauldambra:
* **#2503** <a href="https://app.graphite.dev/github/pr/PostHog/posthog-js/2503?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/PostHog/posthog-js/2503?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -741,18 +741,39 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
     }
 
     private _onSessionIdCallback: SessionIdChangedCallback = (sessionId, windowId, changeReason) => {
-        if (changeReason) {
-            this._tryAddCustomEvent('$session_id_change', { sessionId, windowId, changeReason })
+        if (!changeReason) return
```
pauldambra, line 746: as a follow-up here... i know why we'd ignore part of this for no change reason. 
but maybe we shouldn't ignore all of it 🤔

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -651,7 +651,9 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         }
 
         // We want to ensure the sessionManager is reset if necessary on loading the recorder
-        this._sessionManager.checkAndGetSessionAndWindowId()
+        const { sessionId, windowId } = this._sessionManager.checkAndGetSessionAndWindowId()
```
pauldambra, line 654: before using this check here
it was really hard to see when the session id would change without introducing some race condition

all of these code paths are synchronous so we can set the session id here on start
and then be sure that a recording that is starting will always have the session id from that check after this point

(_technically_ it should have done anyway but this makes it explicit and easier to see)

PR #2504, opened 2025-10-25
feat: add react component that tracks when its children are in view

pauldambra:
i forget where but this came up in conversation

lets us do

```
<PostHogInView>
 <MyUpSellComponent />
</PostHogInView>
```

and send an event to posthog when its viewed
e.g. to see if people are scrolling down to a particular component in a gallery of products

you can pass `trackAllChildren`​ to e.g. wrap a gallery of products and instrument each item in the gallery separately  
also adds a playground site with this implemented so you can see it working

### DIFF
--- a/packages/react/src/components/PostHogInView.tsx
+++ b/packages/react/src/components/PostHogInView.tsx
@@ -1,4 +1,4 @@
-import React, { useRef } from 'react'
+import React, { Children, useCallback, useRef } from 'react'
 import { usePostHog } from '../hooks'
 import { VisibilityAndClickTracker } from './internal/VisibilityAndClickTracker'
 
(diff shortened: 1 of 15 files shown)


pauldambra:
* **#2504** <a href="https://app.graphite.dev/github/pr/PostHog/posthog-js/2504?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.dev/github/pr/PostHog/posthog-js/2504?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/react/package.json
```diff
@@ -49,7 +49,6 @@
         "@types/react": "^17.0.0",
         "@types/testing-library__react-hooks": "^4.0.0",
         "cross-env": "^7.0.3",
-        "given2": "^2.1.7",
```
pauldambra: flyby removal of given2 which took ages to unpick from posthog-js and i thought was gone completely

packages/browser/src/__tests__/consent.test.ts
```diff
@@ -23,7 +23,7 @@ function deleteAllCookies() {
 }
 
 // periodically flakes because of unexpected console logging
-jest.retryTimes(3)
+jest.retryTimes(6)
```
pauldambra, line 26: flyby bump of this since it flaked
but that suggests this sticking plaster has very limited life time

packages/react/src/components/PostHogInView.tsx
```diff
@@ -0,0 +1,126 @@
...
+ * // With custom intersection observer options
+ * <PostHogInView
+ *   name="footer"
+ *   observerOptions={{ threshold: 0.5 }}
+ * >
+ *   <Footer />
+ * </PostHogInView>
+ * ```
+ */
+export function PostHogInView({
```
hpouillot, line 89: I am a bit confused by the name. Maybe `CaptureInView` or `CaptureViewed` would be clearer ?

pauldambra: Ah, that's fair... I was copying the `PostHogFlagged`

I like `PostHogCaptureOnViewed` maybe

hpouillot: Yes it's clearer. I am also wondering if we should prefix all our components ? It seems necessary for the `PosthogProvider` (as it provides the posthog instance) but for others it seems a bit overkill IMO. It's clearly outside the scope of this PR though, so feel free to ignore me 😅

.changeset/floppy-suns-dig.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/react': minor
```
hpouillot, line 2: You will need to bump `posthog-js` as well. posthog/react is not a dependency of posthog-js (to avoid circular deps) so it won't be automatically bumped.

pauldambra: 👍

pnpm-lock.yaml
hpouillot, line 1: the preact version changed and is causing the build to fail. I think you will need to revert changes to this file.

PR #2508, opened 2025-10-27
chore: measure bootup blocking time

pauldambra:
there is a cost to running posthog (or any other code) in the browser

and since we don't  measure that we don't know when we make it worse or better

let's try and get a measure in place we can iterate on


pauldambra:
not this

PR #2518, opened 2025-10-28
feat(llma): send number of web searches

Radu-Raicea:
## Changes

This now track web search tool calls across various providers, and sends the web search count as an event property ($ai_web_search_count) to be used in calculating the LLM costs, where applicable.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/ai/src/gemini/index.ts
+++ b/packages/ai/src/gemini/index.ts
@@ -78,7 +78,7 @@ export class WrappedModels {
             (metadata as GenerateContentResponseUsageMetadata & { thoughtsTokenCount?: number })?.thoughtsTokenCount ??
             0,
           cacheReadInputTokens: metadata?.cachedContentTokenCount ?? 0,
-          webSearchCount: hasGoogleSearchUsage(response),
+          webSearchCount: calculateGoogleWebSearchCount(response),
         },
         tools: availableTools,
       })
@@ -124,7 +124,7 @@ export class WrappedModels {
       const stream = await this.client.models.generateContentStream(geminiParams as GenerateContentParameters)
 
       for await (const chunk of stream) {
-        const chunkWebSearchCount = hasGoogleSearchUsage(chunk)
+        const chunkWebSearchCount = calculateGoogleWebSearchCount(chunk)
         if (chunkWebSearchCount > 0 && chunkWebSearchCount > (usage.webSearchCount ?? 0)) {
           usage.webSearchCount = chunkWebSearchCount
         }
@@ -349,7 +349,7 @@ export class WrappedModels {
  * Gemini bills per request that uses grounding, not per individual query.
  * Returns 1 if grounding was used, 0 otherwise.
  */
-function hasGoogleSearchUsage(response: unknown): number {
+function calculateGoogleWebSearchCount(response: unknown): number {
   if (!response || typeof response !== 'object' || !('candidates' in response)) {
     return 0
   }


packages/ai/src/gemini/index.ts
```diff
@@ -333,5 +344,83 @@ export class WrappedModels {
...
+          return false
+        })
+      }
+    }
+
+    return false
+  })
+
+  return hasGrounding ? 1 : 0
+}
```
carlos-marchal-ph, line 423: Mind-blowing that this is necessary to determine if they charge for it 🤯

packages/ai/src/gemini/index.ts
```diff
@@ -116,12 +117,18 @@ export class WrappedModels {
...
 
     try {
       const stream = await this.client.models.generateContentStream(geminiParams as GenerateContentParameters)
 
       for await (const chunk of stream) {
+        const chunkWebSearchCount = hasGoogleSearchUsage(chunk)
+        if (chunkWebSearchCount > 0 && chunkWebSearchCount > (usage.webSearchCount ?? 0)) {
+          usage.webSearchCount = chunkWebSearchCount
+        }
+
```
carlos-marchal-ph: This is a bit hard to parse. The variable seems to be a counter but the function is named like a boolean predicate. What's happening inside the loop is also a bit confusing, why are we replacing the value?

packages/ai/src/openai/index.ts
```diff
@@ -135,6 +138,11 @@ export class WrappedCompletions extends Completions {
               for await (const chunk of stream1) {
                 const choice = chunk?.choices?.[0]
 
+                const chunkWebSearchCount = calculateWebSearchCount(chunk)
+                if (chunkWebSearchCount > 0 && chunkWebSearchCount > (usage.webSearchCount ?? 0)) {
+                  usage.webSearchCount = chunkWebSearchCount
+                }
+
```
carlos-marchal-ph, line 145: Same as above

packages/ai/src/openai/index.ts
```diff
@@ -366,22 +383,31 @@ export class WrappedResponses extends Responses {
...
               }
 
               for await (const chunk of stream1) {
+                if ('response' in chunk && chunk.response) {
+                  const chunkWebSearchCount = calculateWebSearchCount(chunk.response)
+                  if (chunkWebSearchCount > 0 && chunkWebSearchCount > (usage.webSearchCount ?? 0)) {
+                    usage.webSearchCount = chunkWebSearchCount
+                  }
+                }
+
```
carlos-marchal-ph, line 410: Ditto

packages/ai/src/gemini/index.ts
Radu-Raicea: This was to prevent overcounting, since we can detect a web search on multiple chunks even though it was one web search, but `chunkWebSearchCount` is badly named, I will fix this.

Radu-Raicea: Actually, it's not even badly named, it is technically the count (1 or 0). `hasGoogleSearchUsage` is badly named, so I'll rename that instead.

carlos-marchal-ph: But can it ever be more than one?

Radu-Raicea: For Gemini specifically it's at most once per request. Even if multiple are done (if that's even possible), you get charged only for one.

PR #2519, opened 2025-10-29
feat: make rageclick detection configurable

rodrigomarcolin:
## Problem

The rageclick feature was previously hardcoded with fixed thresholds, limiting flexibility for users who needed to fine-tune detection sensitivity (e.g., for slower or faster click patterns), as mentioned in issue #2487 . 

## Changes

* Updated RageClick class to accept a RageclickConfig | boolean instead of relying on static constants.
* Extended interface RageclickConfig to accept the following optional attributes:
    * threshold_px → max distance between clicks (default 30)
    * timeout_ms → max interval between clicks (default 1000)
    * click_count → number of clicks required (default 3) 
* Updated Autocapture and Heatmaps to initialize RageClick with instance.config.rageclick.
* Extended tests in rageclick.test.ts and autocapture.test.ts to validate all new configuration behaviors and backward compatibility.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/rageclick.ts
+++ b/packages/browser/src/extensions/rageclick.ts
@@ -15,8 +15,10 @@ export default class RageClick {
     thresholdPx: number
     timeoutMs: number
     clickCount: number
+    disabled: boolean
 
     constructor(rageclickConfig: RageclickConfig | boolean) {
+        this.disabled = rageclickConfig === false
         const conf = isObject(rageclickConfig) ? rageclickConfig : {}
 
         this.thresholdPx = conf.threshold_px || DEFAULT_THRESHOLD_PX
@@ -27,6 +29,10 @@ export default class RageClick {
     }
 
     isRageClick(x: number, y: number, timestamp: number): boolean {
+        if (this.disabled) {
+            return false
+        }
+
         const lastClick = this.clicks[this.clicks.length - 1]
         if (
             lastClick &&


packages/browser/src/extensions/rageclick.ts
```diff
@@ -1,28 +1,41 @@
...
 
 export default class RageClick {
     clicks: { x: number; y: number; timestamp: number }[]
 
-    constructor() {
+    thresholdPx: number
+    timeoutMs: number
+    clickCount: number
+
+    constructor(rageclickConfig: RageclickConfig | boolean) {
```
rafaeelaudibert, line 20: I don't understand the boolean type. You're simply ignoring the value if not an object. Can we always requirrre an object here?

packages/browser/src/extensions/rageclick.ts
```diff
@@ -1,28 +1,41 @@
-// Naive rage click implementation: If mouse has not moved further than RAGE_CLICK_THRESHOLD_PX
-// over RAGE_CLICK_CLICK_COUNT clicks with max RAGE_CLICK_TIMEOUT_MS between clicks, it's
+// Naive rage click implementation: If mouse has not moved further than rageClickThresholdPx
```
rafaeelaudibert: You can remove the leading `rageClick` since these are not included in the instance variable names

packages/browser/src/extensions/rageclick.ts
```diff
@@ -1,28 +1,41 @@
...
 // counted as a rage click
 
-const RAGE_CLICK_THRESHOLD_PX = 30
-const RAGE_CLICK_TIMEOUT_MS = 1000
-const RAGE_CLICK_CLICK_COUNT = 3
+import { isObject } from '@posthog/core'
+import { RageclickConfig } from '../types'
+
+const DEFAULT_RAGE_CLICK_THRESHOLD_PX = 30
+const DEFAULT_RAGE_CLICK_TIMEOUT_MS = 1000
```
rafaeelaudibert: nit: the `RAGE_CLICK` prefix seems unneccessary

packages/browser/src/__tests__/extensions/rageclick.test.ts
```diff
@@ -73,4 +73,65 @@ describe('RageClick()', () => {
...
+    })
+
+    describe('when disabled or invalid config', () => {
+        it('handles false as input gracefully', () => {
+            instance = new RageClick(false)
+            const result = instance.isRageClick(0, 0, 0)
+            expect(result).toBe(false)
+        })
+
+        it('handles null or undefined config gracefully', () => {
```
rafaeelaudibert: This is impossible type-wise and we don't expect the SDK to support incorrect usecases. You can remove this test.

packages/browser/src/extensions/rageclick.ts
rodrigomarcolin: According to the [posthog-js documentation](https://posthog.com/docs/libraries/js/config), the rageclick config can be either a boolean or a RageclickConfig object. 

In packages/browser/src/heatmaps.ts and in packages/browser/src/autocapture.ts, the Rageclick instance now receives the rageclick config on initialization, so I figured it should accept both types to avoid type complaints.

If the user provides a boolean, I assume they want the default config. That's why it seemed right to me to also accept a boolean. It’s the approach I found to keep the change backward compatible.

pauldambra: i guess if rageclick is false we shouldn't be creating this, or we should be setting config such that we don't start rageclick detection

packages/browser/src/extensions/rageclick.ts
```diff
@@ -1,28 +1,41 @@
...
 
 export default class RageClick {
     clicks: { x: number; y: number; timestamp: number }[]
 
-    constructor() {
+    thresholdPx: number
+    timeoutMs: number
+    clickCount: number
+
+    constructor(rageclickConfig: RageclickConfig | boolean) {
```
pauldambra, line 20: Ah, yeah...  i guess we've always been handling rageClickConfig = false wrong
I think the simplest thing is we should still store that here, but if config is false then `isRageClick` is always false

other than that, this looks good to me

rodrigomarcolin: Thanks. I’ve pushed a new change where, if rageClickConfig is false, isRageClick() short-circuits to false.

packages/browser/src/extensions/rageclick.ts
rodrigomarcolin: I have updated this code and mentioned it in another thread, so I will resolve this one.

packages/browser/src/extensions/rageclick.ts
```diff
@@ -1,28 +1,47 @@
...
+    clickCount: number
+    disabled: boolean
+
+    constructor(rageclickConfig: RageclickConfig | boolean) {
+        this.disabled = rageclickConfig === false
+        const conf = isObject(rageclickConfig) ? rageclickConfig : {}
+
+        this.thresholdPx = conf.threshold_px || DEFAULT_THRESHOLD_PX
+        this.timeoutMs = conf.timeout_ms || DEFAULT_TIMEOUT_MS
+        this.clickCount = conf.click_count || DEFAULT_CLICK_COUNT
```
pauldambra, line 26: the only nit here is that you can't set any of these values to 0
given you can set any of them to 1
(and that would be silly) then i think this is OK

we're also not clamping these values to a range
but i don't think that setting them to very large or very small values has an impact other than making the analytics odd

pauldambra (approved):
LGTM

PR #949, opened 2024-01-05
feat: add opt_out_useragent_filter and $browser_type

christianvuerings:
## Description

We at Cambly would love to capture bot traffic through PostHog. Currently bot traffic is automatically filtered out and there is no way to disable the filter. https://github.com/PostHog/posthog-js/blob/958fec4df19246513264d69d52d5e36a5cf6ed52/src/posthog-core.ts#L869

## Changes

* Add `opt_out_useragent_filter` config option to allow clients to opt out of user agent filtering (e.g. bots)
* Add `$browser_type` to every capture event to be able to differentiate between bots and regular browsers. Based on [User Agent Populator](https://posthog.com/docs/cdp/user-agent-populator)

## Checklist
- [x] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [x] Accounted for the impact of any changes across different browsers

PR #2520, opened 2025-10-29
feat: add bot pageview collection behind preview flag

lricoy:
## Summary

This is an initial attempt at bot/agent analytics POC. I decided to expand a bit on https://github.com/PostHog/posthog-js/pull/949, but instead of sending those as regular $pageview events and risking messing up some of our metrics, we will use a different event while testing.

Adds support for collecting bot traffic as `$bot_pageview` events when the `__preview_send_bot_pageviews` flag is enabled.

This enables tracking bot interactions while keeping them separate from real user pageviews. 

I want to have some real data on this and cross-reference it two other data sources to figure out how much we would be losing by choosing a purely posthog-js SDK for this. The other metrics/logs 

- Backend logs for the website
- Network-level logs for LoadBalance/Cache hits

### Browser SDK
- Routes bot pageviews to `$bot_pageview` event when preview flag is enabled
- Adds `$browser_type: 'bot'` property for easy filtering
- Maintains backward compatibility (bots still dropped by default)

### Node SDK  
- Adds bot detection utilities
- Supports `__preview_send_bot_pageviews` flag
- Comprehensive test coverage

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -11,6 +11,7 @@ import {
   Logger,
   PostHogCaptureOptions,
   isPlainObject,
+  isBlockedUA,
 } from '@posthog/core'
 import {
   EventMessage,
@@ -29,7 +30,6 @@ import {
 import ErrorTracking from './extensions/error-tracking'
 import { safeSetTimeout, PostHogEventProperties } from '@posthog/core'
 import { PostHogMemoryStorage } from './storage-memory'
-import { isBlockedUA } from './utils/bot-detection'
 
 // Standard local evaluation rate limit is 600 per minute (10 per second),
 // so the fastest a poller should ever be set is 100ms.
--- a/packages/node/src/utils/bot-detection.ts
+++ b/packages/node/src/utils/bot-detection.ts
@@ -1,2 +1,0 @@
-// Re-export shared bot detection logic from @posthog/core
-export { DEFAULT_BLOCKED_UA_STRS, isBlockedUA } from '@posthog/core'


packages/node/src/utils/bot-detection.ts
```diff
@@ -0,0 +1,113 @@
+// List of blocked user agent strings that identify bots
```
lricoy, line 1: Duplicated for now to avoid messing up imports across packages as this may change.

packages/node/src/__tests__/bot-detection.spec.ts
```diff
@@ -0,0 +1,239 @@
+import { PostHog } from '@/entrypoints/index.node'
```
lricoy, line 1: Mostly duplicate, but I would rather have them while testing it than not have.

packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
```diff
@@ -579,6 +579,11 @@ exports[`config snapshot for PostHogConfig 1`] = `
     "false",
     "true"
   ],
+  "__preview_send_bot_pageviews": [
```
rafaeelaudibert: should be `capture` not `send`

packages/browser/src/posthog-core.ts
```diff
@@ -947,7 +947,12 @@ export class PostHog {
             return
         }
 
-        if (!this.config.opt_out_useragent_filter && this._is_bot()) {
+        const isBot = !this.config.opt_out_useragent_filter && this._is_bot()
```
rafaeelaudibert, line 950: Does it make sense to include the useragent filter condition in here? 🤔

packages/browser/src/posthog-core.ts
```diff
@@ -994,6 +999,14 @@ export class PostHog {
             properties: this.calculateEventProperties(event_name, properties || {}, timestamp, uuid),
         }
 
+        // Route pageviews to $bot_pageview when bot detected and preview flag enabled
+        if (event_name === '$pageview' && this.config.__preview_send_bot_pageviews && isBot) {
+            data.event = '$bot_pageview'
+            // While it's obvious that a $bot_pageview is (likely) from a bot, we explicitly set $browser_type
+            // to make it easy to filter and test bot pageviews in the product
+            data.properties.$browser_type = 'bot'
```
rafaeelaudibert, line 1007: I particularly don't see how this can be useful 🤔. Is `$browser_type` used for something else?

packages/browser/src/types.ts
rafaeelaudibert, line 1088: Is this intentional?

packages/node/src/utils/bot-detection.ts
rafaeelaudibert, line 1: Do we not have a shared folder we can import from and define this only once? We need some method here to keep these in sync.

packages/node/src/client.ts
```diff
@@ -1499,9 +1500,22 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
         return props
       })
 
+    // Handle bot pageview collection based on preview flag
+    let finalEvent = eventMessage.event
```
rafaeelaudibert: Don't really need this, you can override `eventMessage.event` like you do when overriding `eventProperties` directly below.

This is not exposed outside this function, so it's safe to be modified

.changeset/bot-pageview-collection.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog-js': minor
+'posthog-node': minor
```
hpouillot: You will need to bump minor on posthog/core as well to include user agents

packages/browser/src/posthog-core.ts
lricoy: That's what [_is_bot](https://github.com/PostHog/posthog-js/blob/ec31b822c33eac51ef92e5f0b3f19ac768cf16f9/packages/browser/src/posthog-core.ts#L3126-L3132) does, it calls `isLikelyBot` which has the UA check if we have a `navigator` or returns undefined if not.

packages/browser/src/posthog-core.ts
lricoy: Yeah, like the other comment it was already a defined behavior for `opt_out_user_agent_filter` so I would rather keep it for this config as well. 

https://github.com/PostHog/posthog-js/blob/ec31b822c33eac51ef92e5f0b3f19ac768cf16f9/packages/browser/src/posthog-core.ts#L1184-L1188

PR #2531, opened 2025-10-30
fix: apply manual overrides only when lazy loading has finished

TueHaulund:
Currently, manual overrides for Session Replay are silently ignored if they are applied before the module has finished loading.

In this PR, manual overrides are now stored in persistent storage and only applied once the module has been loaded.


packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -243,7 +243,10 @@ export class SessionRecording {
      * instead call `posthog.startSessionRecording({linked_flag: true})`
      * */
     public overrideLinkedFlag() {
-        // TODO what if this gets called before lazy loading is done
+        if (!this._lazyLoadedSessionRecording) {
+            logger.warn('overrideLinkedFlag called before recording has finished loading')
+        }
```
pauldambra, line 257: do we need a warning if we're storing the value and applying after load?

ah... someone could call it directly
so maybe we should store the override here instead of in posthog core
then it doesn't matter who calls it, it's always either stored or immediately acted on from this class

packages/browser/src/posthog-core.ts
```diff
@@ -2592,20 +2595,26 @@
             // allow the session id check to rotate session id if necessary
             this.sessionManager?.checkAndGetSessionAndWindowId()
 
-            if (overrideConfig.sampling) {
-                this.sessionRecording?.overrideSampling()
-            }
+            if (this.sessionRecording?.status === LAZY_LOADING) {
```
pauldambra: the though above means this doesn't have to change at all _i think_

PR #2536, opened 2025-10-31
feat(flags): Add real-time feature flags support

gustavohstrassburger:
Introduce a new `realtime_flags` configuration option that enables real-time feature flag updates through Server-Sent Events (SSE). When enabled, the SDK establishes a persistent connection to receive flag updates without polling.

Key changes:
- Add `realtime_flags` config option (default: false)
- Implement SSE connection management in PostHogFeatureFlags
- Auto-reconnect on connection errors with 5-second delay
- Clean shutdown of SSE connection


gustavohstrassburger:
@greptileai

PR #2541, opened 2025-11-03
chore: deprecate navigationRef autocapture

marandaneto:
## Problem

https://posthog.com/questions/error-could-not-get-the-navigation-state-is-your-component-inside-a-navigator

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/src/types.ts
+++ b/packages/react-native/src/types.ts
@@ -58,7 +58,7 @@ export type PostHogAutocaptureOptions = {
    * If you create a navigation ref with createNavigationContainerRef, you need to pass the navigation ref
    * Only used for expo-router and @react-navigation/native if captureScreens is true
    *
-   * @deprecated this is deprecated since it didn't work as expected, check ou the `captureScreens` comments for capturing screen views semi-automatically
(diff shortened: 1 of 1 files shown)


.changeset/fruity-keys-strive.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': patch
```
ioannisj, line 2: I believe this should be minor actually?

marandaneto: mm why so? its more of a fix (or even info) and not a new feat, the navigationRef didnt work as expected anyway

ioannisj: My thinking is that it changes the surface of the API, or at least the suggested use of it (comes with doc update at least). No strong feelings though so definitely not a blocker

marandaneto: thx, will keep as a patch since as i mentioned, it was not working as expected anyway, and the config is still there, so its not breaking either

PR #2544, opened 2025-11-03
feat(err): update nuxt package

hpouillot:
## Problem

- Fix: https://posthog.com/docs/error-tracking/installation/nuxt
Vercel outputs files in a different directory than nuxt default

## Changes

- No need to specify sourcemap config, we inject it automatically
- Instrument rendering errors (vue:error)
- Use process instead of inject + upload
- Call CLI on close with all assets
- Do no expose server config to clients
- Merge nuxt examples
- Add delete after upload option
- Add nuxt package to repo templates

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/examples/example-nuxt/pages/index.vue
+++ b/examples/example-nuxt/pages/index.vue
@@ -104,6 +104,6 @@ const throwNestedAsyncError = () => {
 
 // 8. Rendering error
 const redirectRenderingError = () => {
-  window.location.href = '/error'
+  window.location.href = '/page-error'
 }
 </script>


packages/nuxt/src/module.ts
```diff
@@ -47,102 +55,98 @@ export default defineNuxtModule<ModuleOptions>({
...
-        'sourcemap',
-        'upload',
-        '--directory',
-        '.output',
-        '--ignore',
-        '**/node_modules/**',
-      ]
+    const outputDir = getOutputDir(nuxt.options.nitro)
 
+    nuxt.hook('close', async () => {
```
ablaszkiewicz, line 114: Last time I was working on that, it was impossible and it already happened to us - https://posthog.slack.com/archives/C07AA937K9A/p1745422116165909?thread_ts=1745420351.451079&cid=C07AA937K9A

Basically it seems like nuxt saves the size of artifacts right after creating them and then artificially limits served files so served files have no comments appended by us (chunkIds). That is why I hooked into `nitro:build:public-assets` so we can inject source maps before nuxt saves files size.

packages/nuxt/src/runtime/vue-plugin.ts
```diff
@@ -1,31 +1,43 @@
...
+    ...posthogClientConfig,
   })
 
-  if (debug) {
+  if (posthogCommon.debug) {
     posthog.debug(true)
   }
 
+  if (autocaptureEnabled(posthogClientConfig)) {
+    nuxtApp.hook('vue:error', (error) => {
```
ablaszkiewicz: Last time I was working on that I decided to not use `vue:error` because if somebody enables auto capture on top of that, it seems that we will be capturing exceptions twice. This hook does not stop error propagation.

I thought if we just use default auto capture, we will capture all errors anyways. But to be double sure I prepared this example page which tests all kind of strange errors which can be thrown and 100% of them were caught by our built-in auto capture. It is in `example-nuxt/pages/index.vue`. Accessible after running build and then 

```
node .output/server/index.mjs
```

And then visiting `localhost:3000`

packages/nuxt/src/module.ts
ablaszkiewicz: I ran it from this branch and I think the same is happening after this change because posthog on the frontend does not seem to be initialized at all and I see error in console

```
GhAD4uOI.js:35 Uncaught SyntaxError: Unexpected end of input (at GhAD4uOI.js:35:2037)
```

This is the exact issue I ran into last time

ablaszkiewicz (changes requested):

packages/nuxt/src/module.ts
hpouillot: I added a comment for our future selves and update the process to not run on other nuxt commands

packages/nuxt/src/runtime/vue-plugin.ts
hpouillot: I added a test case for rendering error that are not caught with global listeners

examples/example-nuxt/pages/index.vue
```diff
@@ -97,4 +101,9 @@ const throwNestedAsyncError = () => {
 
   outerAsync()
 }
+
+// 8. Rendering error
+const redirectRenderingError = () => {
+  window.location.href = '/error'
```
ablaszkiewicz: Shouldn't this be `/page-error`?

PR #2547, opened 2025-11-04
fix: store recording overrides in persistent storage

TueHaulund:
This PR fixes a bug where recording overrides would not be applied if the recorder had not yet finished loading. Instead the overrides are now written to persistent storage and then applied from there when the recorder loads.

Related ticket: https://posthoghelp.zendesk.com/agent/tickets/40881 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/43376)


ksvat (approved):

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -701,6 +705,44 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             })
         })
 
+        const overrideSampling: boolean = this._instance.get_property(SESSION_RECORDING_OVERRIDE_SAMPLING) as boolean
+        if (overrideSampling) {
+            this.overrideSampling()
+
+            // Clean up the override flag after applying it
+            this._instance.persistence?.unregister(SESSION_RECORDING_OVERRIDE_SAMPLING)
+        }
```
pauldambra: nit pick we could pull a top-level function out that takes instance and key and fn

or rather bundle size is important in the SDK and this repeated action could be extracted to a method to help _a little_
(and the little helps add up)

pauldambra (approved):
nit pick to maybe help with bundling (although my intuition about bundle size doesn't always bear out) but otherwise 🚢

PR #2553, opened 2025-11-06
feat: yield to the main thread during init

pauldambra:
a user reports that posthog-js blocks the main thread more than they are comfortable

(they have tight performance constraints)

making the code faster might end up as a lot of code golf

but we can delay extension start up by yielding back to the main thread using setTimeout

to let the browser do some work while we're booting up

we could yield before init or earlier in init, but this seemed like a safe balance, since it's these extensions that will trigger lazy loading of other JS and then parse and process that code

added behind an opt in flag so we can test and then probably rollout as a new defaults option

### DIFF
--- a/playground/nextjs/src/posthog.ts
+++ b/playground/nextjs/src/posthog.ts
@@ -81,7 +81,7 @@ if (typeof window !== 'undefined') {
         __preview_remote_config: true,
         cookieless_mode: 'on_reject',
         __preview_flags_v2: true,
-        __preview_deferred_init_extensions: true, // Fixed typo!
+        __preview_deferred_init_extensions: true,
         ...configForConsent(),
     })
     // Help with debugging


pauldambra:
* **#2553** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2553?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2553?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/types.ts
```diff
@@ -662,6 +662,21 @@ export interface PostHogConfig {
...
+    /**
+     * EXPERIMENTAL: Defers initialization of non-critical extensions (autocapture, session recording, etc.)
+     * to the next event loop tick using setTimeout. This reduces main thread blocking during SDK
+     * initialization for better page load performance, while keeping critical functionality
+     * (persistence, sessions, capture) available immediately.
+     *
+     * When enabled:
+     * - Persistence, sessions, and basic capture work immediately
+     * - Extensions (autocapture, recording, heatmaps, etc.) start after yielding back to the browser
+     *
```
lricoy, line 674:
```suggestion
     * WARNING: This can lead to race conditions and events not being captured, depending on the web page initialization flow
```

LGTM since it is a preview, but we should test it thoroughly and be aware that we might not capture some events by doing that. 

I guess it can be expected since it is delaying execution, but I guess it is better to make it explicit. 

I think the most likely here would be client-side redirects or custom events being done at the start.

Another thing we could try is allowing users to choose when the extensions should be loaded. The immediate nextTick, or after the page load event, so it doesn't mess up page load metrics.

lricoy (approved):
🚢 just a note for you to consider

packages/browser/src/posthog-core.ts
```diff
@@ -537,38 +536,19 @@ export class PostHog {
...
-        if (!this.config.disable_scroll_properties) {
-            this.scrollManager.startMeasuringScrollPosition()
+        // Conditionally defer extension initialization based on config
+        if (this.config.__preview_deferred_init_extensions) {
+            // EXPERIMENTAL: Defer non-critical extension initialization to next tick
+            // This reduces main thread blocking during init
+            // while keeping critical path (persistence, sessions, capture) synchronous
+            setTimeout(() => {
+                this._initExtensions(startInCookielessMode)
+            }, 0)
```
rafaeelaudibert, line 548: I was half-expecting `.bind(this)` here to guarantee `this` is correct inside the function, but I might be being naive here and this is fine

pauldambra: one day i might understand `this` and `bind` 
also one day i may be reincarnated as an animal that does not need to

that will be a good day

will check (about bind, not about reincarnation)

pauldambra: claude says

```
The current code uses an arrow function:
  setTimeout(() => {
      this._initExtensions(startInCookielessMode)
  }, 0)

  Arrow functions lexically bind this - meaning
  this is captured from the enclosing scope
  (_init method), so it will correctly refer to
  the PostHog instance. This is one of the main
  benefits of arrow functions!

  If you were using a traditional function, you'd
   need .bind(this):
  // OLD WAY (not needed with arrow functions)
  setTimeout(function() {
      this._initExtensions(startInCookielessMode)
  }.bind(this), 0)
```

which sounds right to me.
the arrow function in setTimeout is not creating a new context so we have the right `this` when we call `initExtensions`

i will owe you one "i told you so" for when we turn this on for PH cloud

rafaeelaudibert: Damn arrow functions

packages/browser/src/types.ts
pauldambra: yep, that's a good call... i'll make it opt-in during testing and then we can keep it config'd but flip it and let people opt-out if they want blocking starting

but we should time this as well so we can quantify it for them - e.g. if it's 100ms great, if it's 1000ms not so 🙈

lricoy: Yeah, I was talking with Rafa since we're next to each other. The events I was thinking of are things we attach to dom event handlers like the history autocapture (that since it is the default pageview capture mode now should not be an extension - I will change it)

pauldambra: shall i just put it first in the list now?

that way it's the first thing we setup
either synchronously or deferred

lricoy (approved):

packages/browser/src/types.ts
lricoy: Please do if you can for this and I will check later if we can make it not an extension (I think it should be one now)

pauldambra: have moved it to the top and not deferred it into the task queue

so we might defer init 
then history is processed
then we queue deferred tasks for everything else

packages/browser/src/posthog-core.ts
```diff
@@ -670,6 +653,126 @@ export class PostHog {
         return this
     }
 
+    private _initExtensions(startInCookielessMode: boolean): void {
```
rafaeelaudibert, line 656: nit, next pr: `startInCookielessMode` should be a private getter to make this simpler. Maybe even cache it and reset on config change.

pauldambra: if we stick with this approach, there'll be at least one more change to make it a default we can change this then

is that ok?

playground/nextjs/src/posthog.ts
```diff
@@ -81,6 +81,7 @@ if (typeof window !== 'undefined') {
         __preview_remote_config: true,
         cookieless_mode: 'on_reject',
         __preview_flags_v2: true,
+        __preview_deferred_init_extensions: true, // Fixed typo!
```
rafaeelaudibert: What's this comment for?

packages/browser/src/posthog-core.ts
```diff
@@ -679,6 +782,11 @@ export class PostHog {
             return
         }
 
+        // Store config in case extensions aren't initialized yet (only needed for deferred init)
+        if (this.config.__preview_deferred_init_extensions) {
```
rafaeelaudibert, line 786: On my phone, so it's harder to see, where's this used?

packages/browser/src/posthog-core.ts
rafaeelaudibert: Absolutely

playground/nextjs/src/posthog.ts
pauldambra: ha, not 🤖 obvs

pauldambra:
```suggestion
        __preview_deferred_init_extensions: true,
```

packages/browser/src/posthog-core.ts
pauldambra: it's in case of a race between receiving remote config and running delayed init

the final task in the deferred init is to apply pending remote config if we have one

Issue #2148, opened 2024-06-06
setPersonPropertiesForFlags not called automatically when updating person properties in RN


itsramiel:
## Bug description
I am using using user properties to match surveys to a specific user. For example i setup a survey to be shown only if `survey_shown/{survey_id}` is not set as a property on the user.

The issue is that when I set the user property as I capture an event like so:
```ts
      posthog.capture('survey shown', {
        $survey_id: survey.id,
        $set: {
          [`survey_shown/${survey.id}`]: true,
        },
      });
```
and flush the events with `await posthog.flush();` the feature flags are not **immediately** correct and the survey matching is incorrect as a result. It takes some time(second or two) for the feature flags to become correct and correctly match the surveys. This leads to issues where I show the survey multiple times or I dont show a survey that was dependent on the previous survey.

## How to reproduce

1. Create an api survey and set the condition to be `survey_shown/{survey if of created survey} is not set`
2. call the following function twice while awaiting each call to ensure flushing occurred:
```ts
  const getSurveysAndMarkAsShown = async () => {
    const response = await fetch(`${HOST_URL}/api/surveys?token=${API_KEY}`);
    const json = (await response.json()) as TGetPosthogSurveysResponse;

    const featureFlags = await posthog.reloadFeatureFlagsAsync();
    if (!featureFlags) return;

    const matchingSurveys = getMatchingSurveys(json.surveys, featureFlags);
    console.log('matchingSurveys', matchingSurveys);
    matchingSurveys.forEach(survey => {
      posthog.capture('survey shown', {
        $survey_id: survey.id,
        $set: {
          [`survey_shown/${survey.id}`]: true,
        },
      });
    });

    await posthog.flush();
  };
  
  // ......

  await getSurveysAndMarkAsShown();
  await getSurveysAndMarkAsShown();
```
3. Notice that even though in the first call we are setting the user property that we are matching against and we are waiting for the events to be flushed, the same survey is returned in the subsequent call. However in a second or two the survey will no longer be matching(matching took effect)

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

## Additional context

Complete App:
```ts
import {usePostHog, PostHogProvider} from 'posthog-react-native';
import {Button, StyleSheet, View} from 'react-native';

const HOST_URL = 'your-host';
const API_KEY = 'your-api-key';

function App() {
  return (
    <PostHogProvider
      apiKey={API_KEY}
[...]

PR #2564, opened 2025-11-11
feat: Cache user and group properties for flags

dustinbyrne:
## Problem

Feature flags in the React Native SDK can return incorrect values when person or group properties haven't been processed by server-side ingestion yet. This creates a race condition where flags are evaluated without the latest property values.

This issue affects developers using feature flags that depend on person or group properties.

Resolves https://github.com/PostHog/posthog-js/issues/2148

## Changes

This PR implements person and group properties caching for feature flag evaluation, matching the behavior of the Android SDK (https://github.com/PostHog/posthog-android/pull/315) and iOS SDK (https://github.com/PostHog/posthog-ios/pull/373).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -203,7 +203,7 @@ export class PostHog extends PostHogCore {
 
       // Set default person properties for flags if enabled
       if (this._setDefaultPersonProperties) {
-        this._setDefaultPersonPropertiesForFlags()
+        this._setDefaultPersonPropertiesForFlags(false)
       }
 
       this._isInitialized = true
@@ -377,17 +377,22 @@ export class PostHog extends PostHogCore {
   reset(): void {
     super.reset()
 
-    // Restore default person properties after reset if enabled
     if (this._setDefaultPersonProperties) {
-      this._setDefaultPersonPropertiesForFlags()
+      // Reset reloads flags asyncrhonously, but doesn't wait for it.
+      // As a result, we can synchronously set the default person properties without
+      // reloading, and allow the super.reset() call to reload the flags.
+      this._setDefaultPersonPropertiesForFlags(false)
     }
   }
 
   /**
    * Helper to extract and set default person properties from app properties
+   *
    * @private
+   *
+   * @param reloadFeatureFlags Whether to reload feature flags after setting the properties. Defaults to true.
    */
-  private _setDefaultPersonPropertiesForFlags(): void {
+  private _setDefaultPersonPropertiesForFlags(reloadFeatureFlags = true): void {
     const defaultProps: Record<string, string> = {}
     const relevantKeys = [
       '$app_version',
@@ -414,7 +419,7 @@ export class PostHog extends PostHogCore {
     }
 
     if (Object.keys(defaultProps).length > 0) {
-      this.setPersonPropertiesForFlags(defaultProps)
+      this.setPersonPropertiesForFlags(defaultProps, reloadFeatureFlags)
     }
   }
 
--- a/packages/react-native/test/posthog.spec.ts
+++ b/packages/react-native/test/posthog.spec.ts
@@ -677,6 +677,41 @@ describe('PostHog React Native', () => {
         expect(cachedProps).toHaveProperty('$device_type', 'Mobile')
       })
 
+      it('should set default properties synchronously during reset without extra reload', async () => {
+        jest.spyOn(PostHog.prototype, 'getCommonEventProperties').mockReturnValue({
+          $lib: 'posthog-react-native',
+          $lib_version: '1.2.3',
+        })
+        posthog = new PostHog('test-api-key', {
+          setDefaultPersonProperties: true,
+          customAppProperties: {
+            $device_type: 'Mobile',
+            $os_name: 'iOS',
+          },
+        })
+        await posthog.ready()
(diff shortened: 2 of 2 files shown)


packages/react-native/src/posthog-rn.ts
```diff
@@ -75,6 +75,26 @@ export interface PostHogOptions extends PostHogCoreOptions {
...
+   *
+   * When enabled, the following properties are sent with every /flags request:
+   * - $app_version: App version
+   * - $app_build: App build number
+   * - $app_namespace: App bundle identifier / namespace
+   * - $os_name: Operating system name
+   * - $os_version: Operating system version
+   * - $device_type: Device type (Mobile, Desktop, Web)
+   * - $locale: User's current locale
+   * - $timezone: User's timezone
```
marandaneto: we can remove these 2 since we've discussed on android and ios, they are not evaluated and can cause confusion

packages/react-native/src/posthog-rn.ts
```diff
@@ -75,6 +75,26 @@ export interface PostHogOptions extends PostHogCoreOptions {
    * Error Tracking Configuration
    */
   errorTracking?: ErrorTrackingOptions
+
+  /**
+   * Automatically include common device and app properties in feature flag evaluation.
+   *
+   * When enabled, the following properties are sent with every /flags request:
```
marandaneto, line 82: missing $lib and $lib_version listed here

packages/react-native/src/posthog-rn.ts
```diff
@@ -742,9 +753,14 @@ export class PostHog extends PostHogCore {
...
    * @param properties The group properties to set for flag evaluation
+   * @param reloadFeatureFlags Whether to reload feature flags after setting the properties. Defaults to true.
    */
-  setGroupPropertiesForFlags(properties: Record<string, Record<string, string>>): void {
+  setGroupPropertiesForFlags(properties: Record<string, Record<string, string>>, reloadFeatureFlags = true): void {
     super.setGroupPropertiesForFlags(properties)
+
+    if (reloadFeatureFlags) {
+      this.reloadFeatureFlags()
+    }
```
dustinbyrne, line 766: Regarding existing cases where users may manually `reloadFeatureFlags` themselves after calling this method:

1. `reloadFeatureFlags` cannot be awaited
https://github.com/PostHog/posthog-js/blob/d331ab6ec2b8e1ce4e8f04623bbabcb0f7413154/packages/core/src/posthog-core.ts#L817-L818

2. There is a guard in place to prevent more than one concurrent request
https://github.com/PostHog/posthog-js/blob/d331ab6ec2b8e1ce4e8f04623bbabcb0f7413154/packages/core/src/posthog-core.ts#L453-L460

Given these two things, it should mitigate against a double reload.

dustinbyrne:
> should we add a default param to reload flags for these methods like we did on android and ios? set person and group props with reloadFeatureFlags: true by default

Good idea. I've added this in 1224066d21151353bf391e185da3766df41259f2

packages/react-native/src/posthog-rn.ts
```diff
@@ -349,6 +376,46 @@ export class PostHog extends PostHogCore {
    */
   reset(): void {
     super.reset()
+
+    // Restore default person properties after reset if enabled
+    if (this._setDefaultPersonProperties) {
+      this._setDefaultPersonPropertiesForFlags()
```
marandaneto: `reset` already reload flags by defualt
i think `_setDefaultPersonPropertiesForFlags` should take a `reloadFeatureFlags` param and pass to `setPersonPropertiesForFlags` as false so its not reloaded twice.
or `reset` takes a `reloadFeatureFlags: false` and let the `_setDefaultPersonPropertiesForFlags` reload the flags, no preference

packages/react-native/src/posthog-rn.ts
```diff
@@ -179,6 +201,11 @@ export class PostHog extends PostHogCore {
 
       this.setupBootstrap(options)
 
+      // Set default person properties for flags if enabled
+      if (this._setDefaultPersonProperties) {
+        this._setDefaultPersonPropertiesForFlags()
```
marandaneto: similar to https://github.com/PostHog/posthog-js/pull/2564/files#r2518097170 since flags will be reloaded after this method call

packages/react-native/src/posthog-rn.ts
```diff
@@ -816,6 +921,20 @@ export class PostHog extends PostHogCore {
...
+    if (properties && Object.keys(properties).length > 0) {
+      const propsToCache: Record<string, string> = {}
+      Object.keys(properties).forEach((key) => {
+        const value = properties[key]
+        if (value !== null && value !== undefined) {
+          propsToCache[key] = String(value)
+        }
+      })
+      if (Object.keys(propsToCache).length > 0) {
+        this.setPersonPropertiesForFlags(propsToCache)
```
marandaneto: similar to https://github.com/PostHog/posthog-js/pull/2564/files#r2518097170 since flags will be reloaded already during `identify` if the user wasnt identified before

PR #2565, opened 2025-11-11
feat: ignore rageclicks by click target content

pauldambra:
often we capture rageclicks because some is clicking the previous and next buttons on an image gallery (or some similar construct)

they're just noise at that point

you can set a css selector ignore list

but things should be ✨ magical ✨

this change adds a new default to not capture rageclicks on elements where the text or their aria-label contains next, previous, prev, >, <

the list of content is configurable and the feature can be disabled

for existing users they won't get this, anyone with defaults: '2025-11' or explicitly enabling it will

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -148,7 +148,7 @@ let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 &&
 const defaultsThatVaryByConfig = (
     defaults?: ConfigDefaults
 ): Pick<PostHogConfig, 'rageclick' | 'capture_pageview' | 'session_recording'> => ({
-    rageclick: defaults && defaults >= '2025-11' ? { content_ignorelist: true } : true,
+    rageclick: defaults && defaults >= '2025-11-30' ? { content_ignorelist: true } : true,
     capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true,
     session_recording: defaults && defaults >= '2025-11-30' ? { strictMinimumDuration: true } : {},
 })


pauldambra:
* **#2565** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2565?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2565?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/types.ts
```diff
@@ -270,7 +281,7 @@ export interface HeatmapConfig {
 
 export type BeforeSendFn = (cr: CaptureResult | null) => CaptureResult | null
 
-export type ConfigDefaults = '2025-05-24' | 'unset'
+export type ConfigDefaults = '2025-11' | '2025-05-24' | 'unset'
```
rafaeelaudibert: A bit yuck? But month makes sense.

We should check whether we wanna batch anything else behind this

pauldambra: oh, yeah, it's cos i have at least one more change i want to release behind a default too

i think that "this month's will change while it is this month" is reasonable enough safety

and we don't need to update the default we set in the site until we're sure we want to release it to everyone. so `2025-05-24` will still be the actual head default for now

rafaeelaudibert: Probably the new lazy init too?

rafaeelaudibert: Ah yeah, you mentioned it. Fuck Github's UI

rafaeelaudibert: Cool, that makes sense to me!

packages/browser/src/posthog-core.ts
```diff
@@ -163,7 +163,7 @@ export const defaultConfig = (defaults?: ConfigDefaults): PostHogConfig => ({
     custom_campaign_params: [],
     custom_blocked_useragents: [],
     save_referrer: true,
-    capture_pageview: defaults === '2025-05-24' ? 'history_change' : true,
+    capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true,
```
rafaeelaudibert: Maybe we need a `matchDefaults(defaults, { if: 'history_change', else: true })` kind of helper here to avoid repeating this everywhere (it's just 2 places for now, I know)

pauldambra: ok, had a quick go at something, since it's three now

packages/browser/src/posthog-core.ts
```diff
@@ -145,6 +145,14 @@ const PRIMARY_INSTANCE_NAME = 'posthog'
 // should only be true for Opera<12
 let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 && userAgent?.indexOf('Mozilla') === -1
 
+const defaultsThatVaryByConfig = (
+    defaults?: ConfigDefaults
+): Pick<PostHogConfig, 'rageclick' | 'capture_pageview' | 'session_recording'> => ({
+    rageclick: defaults && defaults >= '2025-11' ? { content_ignorelist: true } : true,
```
rafaeelaudibert: You changed to `2025-11-30` below, but missed here. Are there 3 defaults or is this one wrong?

pauldambra: it was a deliberate mistake to make sure you were paying attention

pauldambra: [side-eye-monkey](https://app.graphite.com/settings/meme-library?org=PostHog)

pauldambra:
```suggestion
    rageclick: defaults && defaults >= '2025-11-30' ? { content_ignorelist: true } : true,
```

rafaeelaudibert: `<insert paul bamboozles emoji here>`

pauldambra: everyone else gets yolo or benterprise
i get "bamboozles"
still don't know how i feel about that 🤣

rafaeelaudibert: Don't even start, I have "no hype"

PR #2567, opened 2025-11-11
feat: make deferred init default

pauldambra:
## Problem

a user reports that they have strict performance requirements and posthog was blocking too much.
in 1.290.1 we released deferred init behind a flag

for our use of posthog this made our INP measurement much less variable

## Changes

sets the default value to true for defaults 2025-11

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes


lricoy (approved):
Question: is it ok then to break the __preview* options? I always assumed so but wasn't 100% sure.

packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
```diff
@@ -242,11 +242,11 @@ exports[`config snapshot for PostHogConfig 1`] = `
   ],
   "properties_string_max_length": "number",
   "defaults": [
-    "\\"2025-11-08\\"",
+    "\\"2025-11\\"",
     "\\"2025-05-24\\"",
     "\\"unset\\""
   ],
-  "__preview_deferred_init_extensions": [
+  "deferred_init_extensions": [
```
rafaeelaudibert: Isn't `deferred_extensions_init` a better name? Or `defer_extensions_init`? You're the bri'ish tho, your call

packages/browser/src/posthog-core.ts
```diff
@@ -166,7 +166,7 @@ export const defaultConfig = (defaults?: ConfigDefaults): PostHogConfig => ({
     capture_pageview: defaults === '2025-05-24' || defaults === '2025-11' ? 'history_change' : true,
```
rafaeelaudibert: ⚠️ Gotta change this now to be `>= '2025-05-24'` (we probably need a helper to do these calculations now)

rafaeelaudibert (approved):
Would love a follow-up where we kind of write the code in a way where it reads as if `true` was the default rather than the current way where it's clearly "an after thought".

packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
pauldambra: i guess it's the extensions that's unnecessary. since the important thing is the init is deferred (the fact that what we init is extensions is secondary)

packages/browser/src/posthog-core.ts
pauldambra: yeah, two isn't too annoying, three would start to upset me :)

pauldambra:
i don't follow...
can you expand on what's wrong and what you'd prefer?

packages/browser/src/posthog-core.ts
robbie-c: as long as we use YYYY-MM-DD dates we can just use string comparisons :D

pauldambra: no way!

pauldambra: mind blown

playground/nextjs/src/posthog.ts
```diff
@@ -81,7 +81,7 @@ if (typeof window !== 'undefined') {
         __preview_remote_config: true,
         cookieless_mode: 'on_reject',
         __preview_flags_v2: true,
-        __preview_deferred_init_extensions: true,
+        deferred_init: true,
```
pauldambra:
```suggestion
```

PR #2568, opened 2025-11-12
chore: refactor all tests to use the same helpers for instant'ing posthog

pauldambra:
in another PR I was having to make changes to tonnes of files just to change setting up PostHog

but we shouldn't have lots of different ways of setting up posthog in tests

let's collapse to two

* a shared instance
* a new instance each time (but a mock)

and then it's easy to improve this over time
(since it'd be nice to remove all the untyped fake posthog's in our tests)


rafaeelaudibert (approved):
Didn't read, if tests are passing, nice

PR #2570, opened 2025-11-12
fix: autocapture rageclick detection used date.now and instead should use the event timestamp

pauldambra:
## Problem

rageclick test was flapping
because detection was using date.now for event check

## Changes

instead use event.timeStamp when present and fall back to date.now

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/seven-crews-design.md
+++ b/.changeset/seven-crews-design.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: rageclick detection should use event timestamp not current time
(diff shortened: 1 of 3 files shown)


rafaeelaudibert (approved):
nice

PR #2573, opened 2025-11-13
feat: add survey feedback button position options

adboio:
## Problem
473110514-92cc9b22-d363-4e6f-af92-c90b484a2955

Surveys with "feedback button" presentation mode display a tab/button on the right side of the page. This position cannot be changed.

Issue w/ customer +1s: https://github.com/PostHog/posthog/issues/35976

Frontend changing coming here - https://github.com/PostHog/posthog/pull/41410 - working on a better UX for why it's disabled before payment. This is safe to merge beforehand, as it defaults to "right" which is the only option available currently

## Changes

| | |
|:-------------------------:|:-------------------------:|
|Screenshot 2025-11-12 at 7 39 13 PM|Screenshot 2025-11-12 at 7 39 25 PM|
|Screenshot 2025-11-12 at 7 39 32 PM|Screenshot 2025-11-12 at 7 39 41 PM|

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -11,6 +11,7 @@ import {
     SurveyQuestionBranchingType,
     SurveyQuestionType,
     SurveySchedule,
+    SurveyTabPosition,
     SurveyType,
     SurveyWidgetType,
     SurveyWithTypeAndAppearance,
@@ -1057,6 +1058,23 @@ function getPopoverPosition(
     }
 }
 
+function getTabPositionStyles(
+    position: SurveyTabPosition = SurveyTabPosition.Right
+): React.CSSProperties {
+    switch (position) {
+        case SurveyTabPosition.Top:
+            return { top: '0', left: '50%', transform: 'translateX(-50%)' }
+        case SurveyTabPosition.Left:
+            return { top: '50%', left: '0', transform: 'rotate(90deg) translateY(-100%)', transformOrigin: 'left top' }
+        case SurveyTabPosition.Bottom: // bottom center
+            return { bottom: '0', left: '50%', transform: 'translateX(-50%)' }
+        default:
+        case SurveyTabPosition.Right:
+            // not perfectly centered vertically, to avoid a "breaking" change
+            return { top: '50%', right: '0', transform: 'rotate(-90deg) translateY(-100%)', transformOrigin: 'right top' }
+    }
+}
+
 export function SurveyPopup({
     survey,
     forceDisableHtml,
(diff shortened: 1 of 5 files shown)


packages/browser/src/extensions/surveys.tsx
```diff
@@ -1335,7 +1356,12 @@ export function FeedbackWidget({
     return (
         <Preact.Fragment>
             {survey.appearance?.widgetType === 'tab' && (
-                <button className="ph-survey-widget-tab" onClick={toggleSurvey} disabled={readOnly}>
+                <button
+                    className={`ph-survey-widget-tab ${survey.appearance?.tabPosition === SurveyTabPosition.Top ? 'widget-tab-top' : ''}`}
```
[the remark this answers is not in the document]
adboio: this is intentional - for bottom, left, and right, we want `padding-bottom` because left/right just rotate the element - this is what's in the base class :hover

we only need `padding-top` for top position, because the element is not rotated

packages/browser/src/extensions/surveys/survey.css
```diff
@@ -152,25 +152,29 @@
...
     z-index: var(--ph-survey-z-index);
-    transition: padding-bottom 0.2s ease-out;
+    transition: padding 0.2s ease-out;
     border: none;
 }
 
-.ph-survey-widget-tab:hover {
+.ph-survey-widget-tab:hover:not(.widget-tab-top) {
     padding-bottom: 16px;
 }
```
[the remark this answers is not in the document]
adboio: no, this works - i think because `bottom|left|right: 0`

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1057,6 +1058,26 @@ function getPopoverPosition(
...
+        default:
+        case SurveyTabPosition.Right:
+            // not perfectly centered vertically, to avoid a "breaking" change
+            return {
+                top: '50%',
+                right: '0',
+                transform: 'rotate(-90deg) translateY(-100%)',
+                transformOrigin: 'right top',
+            }
+    }
```
lucasheriques, line 1078: i think it's fine to fix this and make it vertically centered, we can treat this as a bugfix/improvement

lucasheriques (approved):
looks great, one comment but approving so you're good to go once fix/respond back :+1:

packages/browser/src/extensions/surveys.tsx
adboio: i was worried this might "break" existing implementations - some customer(s) in the original issue were wanting to move the button because it blocked content, so there is a chance users have structured content around the existing implementation

probably unlikely 😛  but that was my thinking process!

lucasheriques: ohhh got it, kay makes sense, let's just leave it like that then 👍  good thinking

adboio: sounds good! will wait for the tests to pass & then merge :)

PR #2578, opened 2025-11-13
feat: Add logs when overriding feature flags

rafaeelaudibert:
This method returns `undefined`, so it's hard to know whether it worked or not, let's at least add some logs when running in debug mode

Ideally we'd just return whatever was set, but we can't break backwards compatibility by changing the type here

See https://posthog.slack.com/archives/C02E3BKC78F/p1763052769755809


pauldambra (approved):
nice addition whether we do the console log switcheroo :)

rafaeelaudibert:
@pauldambra changed the approach slightly, would love a re-review

PR #2579, opened 2025-11-14
feat(nuxt): Added feature flag composables

geferon:
## Problem

Currently the Nuxt module is pretty barebones, and one major feature of Posthog imo is feature flags, so I am creating a PR to add some useful composables starting with feature flags.

## Changes

Added 4 Nuxt Composables:
- usePostHog(): posthog | undefined
- useFeatureFlagEnabled(flag: string): Ref<boolean | undefined>
- useFeatureFlagPayload(flag: string): Ref<JsonType>
- useFeatureFlagVariantKey(flag: string): Ref<string | boolean | undefined>

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/better-ghosts-stand.md
+++ b/.changeset/better-ghosts-stand.md
@@ -1,0 +1,5 @@
+---
+'@posthog/nuxt': minor
+---
+
+Added feature flag composables to the Nuxt module


packages/nuxt/src/runtime/composables/useFeatureFlagPayload.ts
```diff
@@ -0,0 +1,37 @@
...
+    const unsubscribe = posthog.onFeatureFlags?.(() => {
+      featureFlagPayload.value = posthog.getFeatureFlagPayload(flag)
+    })
+
+    return () => {
+      if (unsubscribe) {
+        unsubscribe()
+      }
+    }
+  })
```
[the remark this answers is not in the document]
geferon: Addressed

packages/nuxt/src/runtime/composables/useFeatureFlagVariantKey.ts
```diff
@@ -0,0 +1,35 @@
...
+    const unsubscribe = posthog.onFeatureFlags?.(() => {
+      featureFlagVariantKey.value = posthog.getFeatureFlag(flag)
+    })
+
+    return () => {
+      if (unsubscribe) {
+        unsubscribe()
+      }
+    }
+  })
```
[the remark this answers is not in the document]
geferon: Addressed

packages/nuxt/src/runtime/composables/useFeatureFlagEnabled.ts
```diff
@@ -0,0 +1,35 @@
...
+    const unsubscribe = posthog.onFeatureFlags?.(() => {
+      featureEnabled.value = posthog.isFeatureEnabled(flag)
+    })
+
+    return () => {
+      if (unsubscribe) {
+        unsubscribe()
+      }
+    }
+  })
```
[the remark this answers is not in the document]
geferon: Addressed

packages/nuxt/src/runtime/composables/useFeatureFlagEnabled.ts
```diff
@@ -0,0 +1,35 @@
...
+    const unsubscribe = posthog.onFeatureFlags?.(() => {
+      featureEnabled.value = posthog.isFeatureEnabled(flag)
+    })
+
+    onUnmounted(() => {
+      if (unsubscribe) {
+        unsubscribe()
+      }
+    })
+  })
```
dmarticus, line 41: I think greptile gave you some bad advice here – I don't think these lifecycle hooks should be nested, since e.g. if the component unmounts before `onMounted` runs, the cleanup won't be registered. Like greptile called it; this could also cause a memory leak.

Instead, I think a better pattern would be something like this

```
let unsubscribe: (() => void) | undefined

onMounted(() => {
  if (!posthog) return

  unsubscribe = posthog.onFeatureFlags?.(() => {
    featureEnabled.value = posthog.isFeatureEnabled(flag)
  })
})

onUnmounted(() => {
  unsubscribe?.()
})
```

this avoids the nested lifecycle hooks and still prevents the memory leak.

packages/nuxt/src/runtime/composables/useFeatureFlagEnabled.ts
```diff
@@ -0,0 +1,35 @@
...
+ * ```ts
+ * const isEnabled = useFeatureFlagEnabled('my-flag')
+ * ```
+ *
+ * @param flag - The feature flag key
+ * @returns A reactive ref containing the feature flag enabled state (boolean | undefined)
+ */
+export function useFeatureFlagEnabled(flag: string) {
+  const posthog = usePostHog()
+  const featureEnabled = ref<boolean | undefined>(posthog?.isFeatureEnabled?.(flag))
```
dmarticus, line 27: The composables initialize the ref with the current PostHog value, but on the server side, `posthog` will likely be `undefined` or not have flags loaded yet, so this will always be undefined during e.g. SSR. This is probably fine, but it might be worth documenting this behavior.

dmarticus (commented):
hi @geferon, thanks for this contribution! Generally this is solid; I left some specific feedback on this PR re: a potential memory leak issue, but once you address that (and the other specific feedback I left on this PR), I'd be happy to add this to our SDK.

Feel free to @ me when you're done addressing my feedback and I'll be sure to take another look.

packages/nuxt/src/runtime/composables/useFeatureFlagEnabled.ts
geferon: Yeah that's what I thought too but I thought perhaps it knew better... Wasn't so sure. Should be resolved.

packages/nuxt/src/runtime/composables/useFeatureFlagEnabled.ts
geferon: Documented!

geferon:
Many thanks!

dmarticus:
reimplemented this changes here: https://github.com/PostHog/posthog-js/pull/2778

PR #2585, opened 2025-11-15
feat: use yielding to main thread in more places

pauldambra:
we tested yielding processing to the main thread periodically and it worked well

reducing variability of INP on pages using that mode

let's abstract the task queue processing approach and use it in a few more places where we might be running over large arrays and cause a long task


packages/browser/jest.config.js
```diff
@@ -1,5 +1,5 @@
 module.exports = {
-    testPathIgnorePatterns: ['/node_modules/', '/cypress/', '/react/', '/test_data/', '/testcafe/'],
+    testPathIgnorePatterns: ['/node_modules/', '/cypress/', '/react/', '/test_data/', '/testcafe/', '\\.d\\.ts$'],
```
rafaeelaudibert, line 2: This doesn't look right, you're escaping the slash rather than the dot?

Shouldn't it be `\.d'\.ts$`?

packages/browser/src/posthog-core.ts
```diff
@@ -733,52 +730,39 @@ export class PostHog {
...
+                        $sdk_debug_extensions_init_method: 'deferred',
+                        $sdk_debug_extensions_init_time_ms: totalTimeMs,
+                    })
+                    logger.info(`PostHog extensions initialized (${totalTimeMs}ms)`)
+                },
+                onError: (error) => {
+                    logger.error('Error initializing extension:', error)
+                },
+            })
+            taskQueue.enqueueAll(initTasks)
```
rafaeelaudibert: Do we not need to call a method to actually start the processing?

packages/browser/src/utils/task-queue.ts
```diff
@@ -0,0 +1,210 @@
+/**
```
rafaeelaudibert: I don't exactly like how we need to use this class, but it works, so please just merge if you think this is good.

IMO the way we should consume this is `new AsyncProcessor(iterator, { timeBudgetMs: 30 }).processEach(item => doSomething(item))`

This is definitely different from your idea where each item in the array is a function and you're just calling those functions, BUT it does align more with how you're using this class (list of items that need to be processed iwth the same function everytime). It basically means to me you're trying to use the wrong abstraction for the job, hence why you had to add those 2 helper functions.

playground/remix/app/providers.tsx
```diff
@@ -8,8 +8,8 @@ export function PHProvider({ children }: { children: React.ReactNode }) {
     const [events, setEvents] = useState<CaptureResult[]>([])
 
     useEffect(() => {
-        posthog.init('phc_test_key_for_playground', {
-            api_host: '/ph-relay-xyz123',
+        posthog.init('sTMFPsFhdP1Ssg', {
```
rafaeelaudibert: Don't add our production key here, people already use our key on their apps just because we have it on our `posthog` repo , we don't need even more people using it 😥

rafaeelaudibert (approved):

packages/browser/src/utils/task-queue.ts
pauldambra: > list of items that need to be processed iwth the same function everytime

hmmm, that's not the abstraction though

this is

list of tasks that need to be processed

they could be the same task many times but they don't have to be

right now, there's no shared task queue, but we could end up with a shared task queue,  so i tried to keep it that way

(all replaceable though so 🙈)

rafaeelaudibert: But that's exactly my point: the abstraction is "list of task that needs to be processed" while the usage is "list of items that need to be processed with the same function at all times".

If the abstraction wasn't wrong we wouldn't need that `processWithYield` helper where we pass the same function at all times 😇 

btw, can we add an underscore to these functions by any chance to make sure people understand they're private? we need to make `(all replaceable though so 🙈)` true lol

pauldambra: ok, will absorb this when my brain is not offsite-tired 😆

playground/remix/app/providers.tsx
pauldambra:
```suggestion
        posthog.init('phc_test_key_for_playground', {
```

packages/browser/src/utils/task-queue.ts
pauldambra: ok, if nothing else this was wrong because every instance was its own task queue so we could still start a lot of work and block as a result  
  
so now it's a singleton queue with a high priority setting

packages/browser/src/entrypoints/array.full.es5.ts
rafaeelaudibert, line 1: Wowowow, 6% increase on bundle size for IE11. Not ideal :disappointed_relieved:

packages/browser/src/posthog-core.ts
```diff
@@ -733,52 +730,44 @@ export class PostHog {
...
+                        const totalTimeMs = Math.round(performance.now() - startTime)
+                        this.register_for_session({
+                            $sdk_debug_extensions_init_method: 'deferred',
+                            $sdk_debug_extensions_init_time_ms: totalTimeMs,
+                        })
+                        logger.info(`PostHog extensions initialized (${totalTimeMs}ms)`)
+                    },
+                    (error) => {
+                        logger.error('Error initializing extension:', error)
+                    }
```
rafaeelaudibert: Should this be on a `catch` to make it clearer?

packages/browser/src/site-apps.ts
```diff
@@ -80,16 +86,21 @@ export class SiteApps {
...
             if (!app.errored && this._bufferedInvocations.length) {
                 logger.info(`Processing ${this._bufferedInvocations.length} events for site app with id ${loader.id}`)
-                this._bufferedInvocations.forEach((globals) => app.processEvent?.(globals))
-                app.processedBuffer = true
+                void scheduler
+                    .processEach(this._bufferedInvocations, (globals) => app.processEvent?.(globals))
+                    .then(() => {
+                        app.processedBuffer = true
+                        if (Object.values(this.apps).every((app) => app.processedBuffer || app.errored)) {
+                            this._stopBuffering?.()
```
rafaeelaudibert, line 100: :warning: You're running this here and then also on line 106.  
  
Is this intended? One of them should probably go away IIUC

packages/browser/src/utils/scheduler.ts
```diff
@@ -0,0 +1,77 @@
...
+        if (items.length === 0) {
+            // eslint-disable-next-line compat/compat
+            return Promise.resolve([])
+        }
+
+        const results: R[] = new Array(items.length)
+        let completedCount = 0
+        const queue = options?.priority === 'high' ? this._highQueue : this._normalQueue
+
+        // eslint-disable-next-line compat/compat
```
rafaeelaudibert: I don't understand why we need the `compat/compat`​ rule now that we're polyfilling `Promise`​

packages/browser/src/entrypoints/array.full.es5.ts
pauldambra: ah, yeah, i didn't realise this would be so much  
  
it would be nice to start using promise in the main SDK though :see_no_evil:  
  
tbh at some point i want to cut a v2 that goes async

packages/browser/src/utils/scheduler.ts
pauldambra: it's because we still advertise ie11 in package.json because the last time i tried to take that away something in CI started failing  
  
i've just not tried agains since (and had totally forgotten to this point)  
  
but also since if we use something that isn't compat we need to know to add it to the es5 bundle (i got the warning here for using promise and so need to polyfill or use a cb instead)

pauldambra: maybe i should not use promise here  
  
i think we need to get to a place where we can cut a v2 of posthog-js  
no longer support es5 and not need to test or check  
  
can introduce a `captureAsync`​ at that point  
  
and then we can start do fun things like use indexDB

rafaeelaudibert: I'm against v2 before an SDK team comes into town - and very in favor after that

rafaeelaudibert: but yeah, all good with the Promise usage, I don't mind the polyfilling if we're ok with the package size increase for those using the ie11 version (which is probably fine!)

packages/browser/src/utils/task-queue.ts
rafaeelaudibert: nice, I like this much much better! ty

packages/browser/src/entrypoints/array.full.es5.ts
rafaeelaudibert: async async async!  
  
But this should be all right, people on IE11 don't have a good experience anyway, I'm not exactly against making it worse

packages/browser/src/utils/scheduler.ts
pauldambra: > I'm against v2 before an SDK team comes into town - and very in favor after that

arguably we need to do this for session replay  
but then we can do it properly and not side of desk

pauldambra:
Local INP testing Results Summary

  | Test        | npm (no scheduler) | local (scheduler) | Max INP Change |
  |-|-|-|-|
  | Simple page | 56ms max           | 48ms max          | -14%           |
  | Media page  | 64ms max           | 48ms max          | -25%|

packages/browser/src/__tests__/deferred-init-extensions.test.ts
```diff
@@ -39,14 +40,15 @@ describe('deferred extension initialization', () => {
...
     })
 
     describe('race condition handling', () => {
         it('should store pending remote config when it arrives before extensions initialize', async () => {
             const token = uuidv7()
             const remoteConfig: RemoteConfig = {
-                supportedCompression: ['gzip'],
-            } as RemoteConfig
+                supportedCompression: [Compression.GZipJS],
+            } as Partial<RemoteConfig> as RemoteConfig
```
rafaeelaudibert, line 51: Weird casting? Really needed?

packages/browser/src/request-queue.ts
```diff
@@ -61,18 +62,22 @@ export class RequestQueue {
             this._clearFlushTimeout()
             if (this._queue.length > 0) {
                 const requests = this._formatQueue()
-                for (const key in requests) {
-                    const req = requests[key]
-                    const now = new Date().getTime()
+                const requestEntries = Object.entries(requests)
```
rafaeelaudibert, line 65: You only use the values, so `Object.values` is all you need here

packages/browser/src/site-apps.ts
```diff
@@ -80,18 +86,28 @@ export class SiteApps {
...
     }
 
     setupSiteApp(loader: SiteAppLoader) {
         const app = this.apps[loader.id]
         const processBufferedEvents = () => {
             if (!app.errored && this._bufferedInvocations.length) {
                 logger.info(`Processing ${this._bufferedInvocations.length} events for site app with id ${loader.id}`)
-                this._bufferedInvocations.forEach((globals) => app.processEvent?.(globals))
-                app.processedBuffer = true
+                scheduler.processEach(this._bufferedInvocations, (globals) => app.processEvent?.(globals), {
```
rafaeelaudibert, line 96: You reacted with a +1 to my comment but didnt really explain whether keeping this duplicated was intended or not

packages/browser/src/utils/globals.ts
```diff
@@ -254,6 +254,41 @@ export const XMLHttpRequest =
...
 export const userAgent = navigator?.userAgent
+
+type IdleDeadline = {
+    didTimeout: boolean
+    timeRemaining: () => number
+}
+
+type IdleCallbackHandle = number | ReturnType<typeof setTimeout>
+
+export const _requestIdleCallback = (
```
rafaeelaudibert, line 265: A comment explaining the fallback and difference between `requestIdleCallback` and `setTimeout` wouldnt hurt

rafaeelaudibert (approved):
Nice, left some nits

PR #2588, opened 2025-11-17
feat: add context to node

ablaszkiewicz:
## Problem

We want contexts

## Changes

Added support for 

```
posthog.withContext(contextData, () => {
  // stuff here has this context now
}, contextOptions)
```

Context data:
- `distinctId`,
- `sessionId`,
- `tags`,
- `enableExceptionAutocapture`.

Context options:
- `fresh` - I copied python's context inheritance so you can spawn nested contexts either in `fresh` mode (discard any parent context settings) or merge mode which merges child context with the parent context.

Apart from that I added support for capturing events without distinct id.

In node's async local storage there is [enterWith](https://nodejs.org/api/async_context.html#asynclocalstorageenterwithstore) which is more flexible than [run](https://nodejs.org/api/async_context.html#asynclocalstoragerunstore-callback-args) but unfortunately it is marked as experimental whereas `run` is stable so I decided to go with that.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/__tests__/context.spec.ts
+++ b/packages/node/src/__tests__/context.spec.ts
@@ -118,9 +118,13 @@ describe('PostHog Context', () => {
 
   it('should merge contexts when fresh: false', async () => {
     posthog.withContext({ properties: { outer: 'value1', shared: 'parent' } }, () => {
-      posthog.withContext({ properties: { inner: 'value2', shared: 'child' } }, () => {
-        posthog.capture({ distinctId: 'user-5', event: 'test_event' })
-      }, { fresh: false })
+      posthog.withContext(
+        { properties: { inner: 'value2', shared: 'child' } },
+        () => {
+          posthog.capture({ distinctId: 'user-5', event: 'test_event' })
+        },
+        { fresh: false }
+      )
     })
 
     await waitForPromises()
@@ -137,9 +141,13 @@ describe('PostHog Context', () => {
 
   it('should merge sessionId from parent context', async () => {
     posthog.withContext({ sessionId: 'session-parent', properties: { level: '1' } }, () => {
-      posthog.withContext({ properties: { level: '2' } }, () => {
-        posthog.capture({ distinctId: 'user-6', event: 'test_event' })
-      }, { fresh: false })
+      posthog.withContext(
+        { properties: { level: '2' } },
+        () => {
+          posthog.capture({ distinctId: 'user-6', event: 'test_event' })
+        },
+        { fresh: false }
+      )
     })
 
     await waitForFlush()
@@ -212,25 +220,37 @@ describe('PostHog Context', () => {
     // Enter context A
     posthog.withContext({ properties: { contextA: 'valueA', level: 'A' } }, () => {
       // Enter context B (inherits from A)
-      posthog.withContext({ properties: { contextB: 'valueB', level: 'B' } }, () => {
-        // Enter context C1 (inherits from B, which has A's stuff)
-        posthog.withContext({ properties: { contextC1: 'valueC1', level: 'C1' } }, () => {
-          // Event 1: Should have A, B, and C1 context
-          posthog.capture({ distinctId: 'user-nested', event: 'event_in_C1' })
-        }, { fresh: false })
+      posthog.withContext(
+        { properties: { contextB: 'valueB', level: 'B' } },
+        () => {
+          // Enter context C1 (inherits from B, which has A's stuff)
+          posthog.withContext(
+            { properties: { contextC1: 'valueC1', level: 'C1' } },
+            () => {
+              // Event 1: Should have A, B, and C1 context
+              posthog.capture({ distinctId: 'user-nested', event: 'event_in_C1' })
+            },
+            { fresh: false }
+          )
 
-        // Exit C1 - Event 2: Should have A and B, but not C1
-        posthog.capture({ distinctId: 'user-nested', event: 'event_after_C1' })
+          // Exit C1 - Event 2: Should have A and B, but not C1
+          posthog.capture({ distinctId: 'user-nested', event: 'event_after_C1' })
 
-        // Enter context C2 (inherits from B, which still has A's stuff)
-        posthog.withContext({ properties: { contextC2: 'valueC2', level: 'C2' } }, () => {
-          // Event 3: Should have A, B, and C2 (but not C1)
-          posthog.capture({ distinctId: 'user-nested', event: 'event_in_C2' })
-        }, { fresh: false })
+          // Enter context C2 (inherits from B, which still has A's stuff)
+          posthog.withContext(
+            { properties: { contextC2: 'valueC2', level: 'C2' } },
+            () => {
+              // Event 3: Should have A, B, and C2 (but not C1)
+              posthog.capture({ distinctId: 'user-nested', event: 'event_in_C2' })
+            },
+            { fresh: false }
+          )
 
-        // Exit C2 - Event 4: Should have A and B again (no C1 or C2)
-        posthog.capture({ distinctId: 'user-nested', event: 'event_after_C2' })
-      }, { fresh: false })
+          // Exit C2 - Event 4: Should have A and B again (no C1 or C2)
+          posthog.capture({ distinctId: 'user-nested', event: 'event_after_C2' })
+        },
+        { fresh: false }
+      )
     })
 
     await waitForFlush()


packages/node/src/__tests__/extensions/exception-autocapture-globally-disabled.worker.mjs
```diff
@@ -0,0 +1,52 @@
...
+parentPort.on('message', (msg) => {
+  if (msg.action == 'throw_error') {
+    throw new Error(msg.data)
+  } else if (msg.action == 'reject_promise') {
+    Promise.resolve().then(() => {
+      throw new Error(msg.data)
+    })
+  } else if (msg.action == 'throw_with_context_disabled') {
+    posthog.withContext({ enableExceptionAutocapture: false }, () => {
+      setImmediate(() => {
```
ablaszkiewicz: There are a lot of AI comments that context will not be preserved across setImmediate() boundary but that is not true!

It is not super clear what breaks the boundary but timers certainly do not (and stuff like next tick, etc...). Promise chaining and event handlers also keep the context.

packages/node/src/types.ts
```diff
@@ -19,11 +20,14 @@ export interface SendFeatureFlagsOptions {
   flagKeys?: string[]
 }
 
-export interface EventMessage extends IdentifyMessage {
+export interface EventMessage {
+  distinctId?: string // Optional - can be provided via context
   event: string
+  properties?: Record<string | number, any>
```
daibhin: Why redeclare this if it is already a part of the `IdentifyMessage` interface?

packages/node/src/types.ts
```diff
@@ -19,11 +20,14 @@ export interface SendFeatureFlagsOptions {
...
   groups?: Record<string, string | number> // Mapping of group type to group id
   sendFeatureFlags?: boolean | SendFeatureFlagsOptions
   timestamp?: Date
+  disableGeoip?: boolean
```
daibhin: Same here

packages/node/src/extensions/error-tracking/index.ts
```diff
@@ -64,13 +64,25 @@ export default class ErrorTracking {
   }
 
   private startAutocaptureIfEnabled(): void {
-    if (this.isEnabled()) {
```
daibhin, line 67: Should we delete the `isEnabled` function?

daibhin: That function actually also included a check for `this.client.isDisabled`... shouldn't we continue to respect that? Or maybe that happens elsewhere

packages/node/src/extensions/error-tracking/index.ts
```diff
@@ -64,13 +64,25 @@ export default class ErrorTracking {
...
+    addUncaughtExceptionListener(this.onException.bind(this), this.onFatalError.bind(this))
+    addUnhandledRejectionListener(this.onException.bind(this))
   }
 
   private onException(exception: unknown, hint: CoreErrorTracking.EventHint): void {
+    const context = this.client.getContext()
+    const contextAutocaptureEnabled = context?.enableExceptionAutocapture
+
+    const shouldCapture =
+      contextAutocaptureEnabled !== undefined ? contextAutocaptureEnabled : this._exceptionAutocaptureEnabled
```
daibhin:
```suggestion
    const shouldCapture = contextAutocaptureEnabled ?? this._exceptionAutocaptureEnabled
```

packages/node/src/context.ts
```diff
@@ -0,0 +1,49 @@
...
+    return this.storage.getStore()
+  }
+
+  run<T>(context: ContextData, fn: () => T, options: ContextOptions = { fresh: true }): T {
+    const fresh = options.fresh !== false
+
+    if (fresh) {
+      return this.storage.run(context, fn)
+    } else {
+      const currentContext = this.storage.getStore() || {}
```
daibhin: Don't think this is necessary

```suggestion
      const currentContext = this.get() || {}
```

packages/node/src/extensions/error-tracking/index.ts
```diff
@@ -64,13 +64,25 @@ export default class ErrorTracking {
...
-    if (this.isEnabled()) {
-      addUncaughtExceptionListener(this.onException.bind(this), this.onFatalError.bind(this))
-      addUnhandledRejectionListener(this.onException.bind(this))
-    }
+    addUncaughtExceptionListener(this.onException.bind(this), this.onFatalError.bind(this))
+    addUnhandledRejectionListener(this.onException.bind(this))
   }
 
   private onException(exception: unknown, hint: CoreErrorTracking.EventHint): void {
+    const context = this.client.getContext()
```
daibhin: I'm sure you tested it (I will once we can `pnpm install` again) but this suggests that the context is lost when capturing an error outside of the `run`: https://nodejs.org/api/async_context.html#asynclocalstoragerunstore-callback-args.

I assume this means that the `context` here would be `undefined` given the global handler would catch the error at the root if it wasn't handled

packages/node/src/__tests__/context.spec.ts
```diff
@@ -0,0 +1,308 @@
...
+
+    expect(allEvents).toHaveLength(50)
+
+    const capturedIndices = allEvents.map((event) => event.properties.index).sort((a, b) => a - b)
+    const expectedIndices = Array.from({ length: 50 }, (_, i) => i)
+
+    expect(capturedIndices).toEqual(expectedIndices)
+  })
+
+  it('should properly inherit and restore context through nested enter/exit operations', async () => {
```
daibhin, line 219: Nice 👌

packages/node/src/types.ts
ablaszkiewicz: In identify message the `distinctId` is required. Here it is optional.

packages/node/src/context.ts
```diff
@@ -0,0 +1,49 @@
+import { AsyncLocalStorage } from 'async_hooks'
+
+export interface ContextData {
+  distinctId?: string
+  sessionId?: string
+  tags?: Record<string, any>
```
hpouillot: Really not a big fan of the "tags" name, I know that's what we use for posthog-python but it doesn't make sense to me to merge user properties with context tags...

packages/node/src/context.ts
```diff
@@ -0,0 +1,49 @@
+import { AsyncLocalStorage } from 'async_hooks'
```
hpouillot:
```suggestion
import { AsyncLocalStorage } from 'node:async_hooks'
```

hpouillot: did you test this package with edge runtime (example-cloudflare) ? This API is node specific and I think it will break on other runtimes

packages/node/src/context.ts
```diff
@@ -0,0 +1,49 @@
...
+
+export interface ContextData {
+  distinctId?: string
+  sessionId?: string
+  tags?: Record<string, any>
+  enableExceptionAutocapture?: boolean
```
hpouillot: Any reason to put this in ContextData instead of ContextOptions ? Seems strange to mix event data with posthog options.

packages/node/src/types.ts
hpouillot: You can still extend (to avoid redefining props) and override with optional properties

packages/node/src/context.ts
hpouillot: You can have a look at the `entrypoints` folder if you need examples on how to add features based on runtime

packages/node/src/types.ts
ablaszkiewicz: image

When something is required in the base type, I think you can't extend it and make it required.

packages/node/src/types.ts
ablaszkiewicz: Same as above

packages/node/src/types.ts
hpouillot: Oups, my bad, you can overrides with types but not interfaces. Up to you if you want to leave it like this or use types

packages/node/src/extensions/error-tracking/index.ts
ablaszkiewicz: Context inside uncaught exceptions/rejections handlers is preserved as long as we are in an async branch. I modified tests to make that clearer.

This will be very important header in the docs page. When running `posthog.run({}, async () => {})` you HAVE TO use `async` keyword. I will think if we can enforce it somehow tho.

<details>

<summary>Code fragment which confirms that</summary>

```
import { AsyncLocalStorage } from 'async_hooks'

class Posthog {
    private context: MyContext

    constructor() {
        this.context = new MyContext()
    }

    getContext(): MyContext {
        return this.context
    }

    run<T>(context: any, fn: () => T): T {
        return this.context.run(context, fn)
    }
}

class MyContext {
    private storage: AsyncLocalStorage<any>

    constructor() {
        this.storage = new AsyncLocalStorage<any>()
    }

    get(): any {
        return this.storage.getStore()
    }

    run<T>(context: any, fn: () => T): T {
        return this.storage.run(context, fn)
    }
}

const posthog = new Posthog()

global.process.on('uncaughtException', (err) => {
    console.log('Captured uncaught exception')
    console.log(posthog.getContext().get())
})

function main() {
    posthog.run({ somethingInContext: 'test' }, async () => {
        console.log(`Within context`, posthog.getContext().get())
        throw new Error('error')
    })
}

main()

```

</details>

packages/node/src/context.ts
ablaszkiewicz: Merged data and options into one thing. I believe it's easier to use this way.

packages/node/src/types.ts
ablaszkiewicz: Used types

packages/node/src/context.ts
ablaszkiewicz: Changed to `properties`

packages/node/src/context.ts
hpouillot: I think we should split them, it would be less confusing and eaiser to manipulate them separately later on if the options are extended. ```withContext(mergableProps, run, optionalContextOptions)```

packages/node/src/context.ts
ablaszkiewicz: Did some changes to the edge and node entrypoints to make sure it is only initialized in node. Tested with `example-cloudflare`. In fact at the beginning it threw an error that it can't import `node:async_hooks`. After changes it works:
image

packages/node/src/context.ts
ablaszkiewicz: It became mega confusing in the code when I split it and put exception auto capture in the options. It is because options is normally not accessible from the outside and it is not something that is saved within context. It is just used for context initialization. For example `fresh` is inside `ContextOptions` but it is not saved within the context itself.

On top of that if we want something to be retained within context, we have to add it to `ContextData` anyways. And we want `enableExceptionAutocapture` to be stored inside the context.

packages/node/src/extensions/error-tracking/autocapture.ts
```diff
@@ -50,13 +50,13 @@ export function addUncaughtExceptionListener(
   captureFn: (exception: Error, hint: CoreErrorTracking.EventHint) => void,
   onFatalFn: (exception: Error) => void
 ): void {
-  global.process.on('uncaughtException', makeUncaughtExceptionHandler(captureFn, onFatalFn))
+  getGlobalSafely()?.process.on('uncaughtException', makeUncaughtExceptionHandler(captureFn, onFatalFn))
```
hpouillot: you can use globalThis instead of global. it is defined on node and edge runtime but process won't be defined on edge, so this should work :
```suggestion
  globalThis.process?.on('uncaughtException', makeUncaughtExceptionHandler(captureFn, onFatalFn))
```

packages/node/src/extensions/error-tracking/index.ts
```diff
@@ -5,6 +5,7 @@ import { EventMessage, PostHogOptions } from '@/types'
 import type { Logger } from '@posthog/core'
 import { BucketedRateLimiter } from '@posthog/core'
 import { ErrorTracking as CoreErrorTracking } from '@posthog/core'
+import { log } from 'console'
```
hpouillot: 🧹

packages/node/src/extensions/error-tracking/index.ts
```diff
@@ -95,7 +108,7 @@ export default class ErrorTracking {
   }
 
   isEnabled(): boolean {
-    return !this.client.isDisabled && this._exceptionAutocaptureEnabled
+    return !this.client.isDisabled
```
hpouillot: Does it mean we will attach global handler even when user doesn't use exception autocapture ?

ablaszkiewicz: Correct.

Previously we didn't even attach listeners if client disabled exception autocapture.

Now we always attach listeners and then decide inside the `onException` method whether we should capture it. This is because now it is possible to globally disable exception capture but then enable it within some context.

packages/node/src/extensions/context/types.ts
```diff
@@ -0,0 +1,27 @@
...
+  get(): ContextData | undefined
+  run<T>(context: ContextData, fn: () => T, options?: ContextOptions): T
+}
+
+// breakdowns - good
+// it wasn't clear that the exception on the right is the effect of the list on the left
+// telegram
+//
+// backend?
+// filter out internal users?
```
hpouillot: 🧹

ablaszkiewicz: That is embarassing

packages/node/src/extensions/error-tracking/index.ts
hpouillot: it could be done as a follow-up, but I think we should not attach listeners if we don't need to

hpouillot (approved):

Issue #2157, opened 2024-10-02
Session replay support for RN Web


marandaneto:
### Description

https://posthog.com/questions/session-replay-react-native-web
https://github.com/necolas/react-native-web

slshults:
+1 https://posthoghelp.zendesk.com/agent/tickets/37340 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/40856)

matt-byers:
+1 vote for this we're going to have to implement web just to get recordings working for our app which we're building for both native and web

PR #2592, opened 2025-11-18
fix: session replay on react native web

marandaneto:
## Problem

workaround for https://github.com/PostHog/posthog-js/issues/2157

otherwise users would have to polyfill manually.

ideally users init the SDK conditionally, if mobile: uses the RN SDK, if web: uses the JS SDK

example: https://us.posthog.com/shared/6K26C6g8ljMoa0ZDOR9nJkXHchLHJg?t=4

### Libraries affected

- [ ] All of them
- [X] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/examples/example-expo-53/app/posthog.tsx
+++ b/examples/example-expo-53/app/posthog.tsx
@@ -1,5 +1,17 @@
 import PostHog from 'posthog-react-native'
 
+// If you want to use Session relay on React Native web, use the posthog-js SDK instead.
+// Example:
+//
+// import posthog from 'posthog-js'
+//
+// posthog.init(process.env.EXPO_PUBLIC_POSTHOG_PROJECT_KEY!, {
+//     host: process.env.EXPO_PUBLIC_POSTHOG_API_HOST,
+//     debug: true,
+// })
+//
+// export { posthog }
+
 export const posthog = new PostHog(process.env.EXPO_PUBLIC_POSTHOG_PROJECT_KEY!, {
     host: process.env.EXPO_PUBLIC_POSTHOG_API_HOST,
     flushAt: 1,
@@ -12,7 +24,7 @@ export const posthog = new PostHog(process.env.EXPO_PUBLIC_POSTHOG_PROJECT_KEY!,
             console: ['error', 'warn'],
         },
     },
-
+    // persistence: 'memory',
     // if using WebView, you have to disable masking for text inputs and images
     // sessionReplayConfig: {
     //   maskAllTextInputs: false,
--- a/packages/browser/src/utils/globals.ts
+++ b/packages/browser/src/utils/globals.ts
@@ -222,6 +222,14 @@ interface PostHogExtensions {
 
 const global: typeof globalThis | undefined = typeof globalThis !== 'undefined' ? globalThis : win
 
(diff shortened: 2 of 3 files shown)


examples/example-expo-53/app/posthog.tsx
```diff
@@ -12,7 +24,7 @@ export const posthog = new PostHog(process.env.EXPO_PUBLIC_POSTHOG_PROJECT_KEY!,
             console: ['error', 'warn'],
         },
     },
-
+    // persistence: 'memory',
```
ioannisj, line 27: leftover comment here?

packages/browser/src/utils/globals.ts
```diff
@@ -222,6 +222,14 @@ interface PostHogExtensions {
 
 const global: typeof globalThis | undefined = typeof globalThis !== 'undefined' ? globalThis : win
 
+// React Native polyfills for posthog-js compatibility
+if (typeof self === 'undefined') {
+    ;(global as any).self = global
+}
+if (typeof File === 'undefined') {
+    ;(global as any).File = function () {}
+}
```
ioannisj, line 231: I wonder what functionality we may be dropping with this empty stub here? I assume rrweb uses this somehow

examples/example-expo-53/app/posthog.tsx
marandaneto: no, just a handy comment to avoid reinstalling plugins https://posthog.com/docs/libraries/react-native#react-native-web-and-macos

marandaneto:
> feels like the important thing here would be the docs on _how_ to use this - but 🙌

agree, a good start would be to point out to the JS SDK tbh since its the very same thing.

packages/browser/src/utils/globals.ts
marandaneto: its us
https://github.com/PostHog/posthog-js/blob/02215b9d8ad494e72ef6db920a9feb704c046989/packages/browser/src/utils/request-utils.ts#L36
we do instanceof isFile, so this returns false on react native.


>   The feature impact:
> 
>   - Session Recording Network Capture - PostHog won't be able to
>   properly serialize FormData containing File objects when recording
>    network requests
>   - General HTTP requests - Any PostHog requests that include
>   FormData with files won't serialize correctly
> 
>   What will happen without the polyfill:
>   - If FormData contains a File object, the isFile() check crashes
>   instead of returning false
>   - The file would be converted to [object File] instead of the
>   filename
>   - Session recording network capture may miss or incorrectly record
>    file upload requests
> 
>   With the File polyfill:
>   - isFile() returns false for all objects (since nothing is
>   actually a File in React Native)
>   - Files get converted via .toString() instead of .name
>   - Session recording and requests work, but file names aren't
>   captured (graceful degradation)
> 
>   So the File polyfill ensures session recording and network 
>   requests don't crash in React Native web, with minimal
>   functionality loss.

thats my friend "Claudio" :D

PR #2593, opened 2025-11-18
feat: better integration between llma and error tracking

daibhin:
## Problem

Inspired by this thread: https://posthog.slack.com/archives/C07AA937K9A/p1759356822031919

## Changes

- Add LLMA trace_id (for starters, could add other properties too) to exceptions and set exception_id on trace events
- Add ability to Node SDK to skip previously captured exceptions (so we don't double capture)
- Remove unnecessary `isError` check given we have the error

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

### DIFF
--- a/.changeset/crisp-chairs-read.md
+++ b/.changeset/crisp-chairs-read.md
@@ -1,5 +1,0 @@
----
-'@posthog/ai': patch
----
-
-Fixes cache creation cost for Langchain with Anthropic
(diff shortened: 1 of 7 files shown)


andrewm4894 (approved):
thanks! was thinking there should be better integration between LLMA and Error tracking. We added an Errors tab recently but i think this SDK approach also opens up much deeper/native linkage that will be great for users of both products.

hpouillot (approved):

PR #2594, opened 2025-11-18
feat(node): Add experimental `FlagDefinitionCacheProvider`

dustinbyrne:
## Changes

This change introduces a new `FlagDefinitionCacheProvider` interface. This interface allows custom caching strategies to be implemented for feature flag definitions used by local evaluation.

Related to https://github.com/PostHog/posthog/issues/37584

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/yummy-sites-unite.md
+++ b/.changeset/yummy-sites-unite.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': minor
+---
+
+feat: Add `FlagDefinitionCacheProvider` interface and client configuration option


packages/node/src/extensions/feature-flags/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+   * to ensure only one instance fetches at a time in a multi-worker setup.
+   *
+   * When multiple workers share a cache, typically only one should fetch while
+   * others use cached data. Implementations can use Redis locks, database locks,
+   * or other coordination mechanisms.
+   *
+   * @returns true if this instance should fetch, false to skip and read cache
+   * @throws if coordination backend is unavailable (error will be logged, fetch continues)
+   */
+  shouldFetchFlagDefinitions(): Promise<boolean> | boolean
```
dustinbyrne, line 86: This initial interface doesn't provide any additional context to the implementation. My justification is that:
- Static information like team id can be provided via the constructor
- We already have configuration for how often we're polling to update cache. If the implementation wants to be smarter about this, they're better off persisting metadata in the cache (such as last fetch time). In a distributed service we can't guarantee things like the last time fetched.

I think there's potentially a case to be made to expose an interface to hit the flag definitions API in the future. For example, it could make `HEAD` requests and track an `ETag` or `Last-Modified` header. Or, maybe _that's_ the context we provide to this method. In any case, we're not quite ready for this and I see this as a future improvement.

packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -574,6 +630,42 @@ class FeatureFlagsPoller {
...
 
     try {
+      let shouldFetch = true
+      if (this.cacheProvider) {
+        try {
+          shouldFetch = await this.cacheProvider.shouldFetchFlagDefinitions()
+        } catch (err) {
+          this.onError?.(new Error(`Error in shouldFetchFlagDefinitions: ${err}`))
+          // Important: if `shouldFetchFlagDefinitions` throws, we
+          // default to fetching.
```
dustinbyrne, line 640: Does this seem reasonable? In the event this method throws (e.g., can't access cache), this would mean all instances resort back to polling directly from the API every interval.

packages/node/src/experimental.ts
```diff
@@ -0,0 +1,11 @@
...
+ * Experimental APIs
+ *
+ * This module exports experimental features that may change or be removed in minor versions.
+ * Use these APIs with caution and be prepared for breaking changes.
+ *
+ * @packageDocumentation
+ * @experimental
+ */
+
+export type { FlagDefinitionCacheProvider, FlagDefinitionCacheData } from './extensions/feature-flags/cache'
```
dustinbyrne, line 11: I marked the interface and config option as `@experimental` in TSDoc, explaining that it _may_ see breaking changes over minor version bumps. To be really explicit about this, I've also exposed the type under a separate import.

```ts
import type { FlagDefinitionCacheProvider } from 'posthog-node/experimental'
```

I'd like to get some real implementations working before we solidify this API.

packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -537,9 +543,59 @@ class FeatureFlagsPoller {
...
+      return false
+    }
+  }
+
   async loadFeatureFlags(forceReload = false): Promise<void> {
+    // On first load, try to initialize from cache (if a cache provider is configured)
+    if (this.cacheProvider && !this.hasAttemptedCacheLoad) {
+      this.hasAttemptedCacheLoad = true
+      await this.loadFromCache('Loaded flags from cache')
+    }
```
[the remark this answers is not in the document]
dustinbyrne: 🤨

packages/node/src/extensions/feature-flags/cache.ts
```diff
@@ -0,0 +1,114 @@
...
+ * @example
+ * ```typescript
+ * import { FlagDefinitionCacheProvider } from 'posthog-node/experimental'
+ *
+ * class RedisFlagCache implements FlagDefinitionCacheProvider {
+ *   constructor(private redis: Redis, private teamKey: string) { }
+ *
+ *   async getFlagDefinitions(): Promise<FlagDefinitionCacheData | undefined> {
+ *     const cached = await this.redis.get(`posthog:flags:${this.teamKey}`)
+ *     return cached ? JSON.parse(cached) : undefined
```
haacked, line 40: Folks will copy this. Should we have a try/catch in this example?

packages/node/src/extensions/feature-flags/feature-flags.ts
haacked: I think so.

If someone wants different behavior, there's a workaround. They could ensure this never throws (ex: with a wrapper cache provider) and instead make sure it returns `false` instead.

packages/node/src/experimental.ts
haacked: Good call.

packages/node/src/extensions/feature-flags/cache.ts
dustinbyrne: I think this is fine, if this throws it'll consider it a cache miss. If the data is malformed it's not really recoverable from here anyhow.

PR #2601, opened 2025-11-19
feat: Add an example for local eval in Cloudflare

dustinbyrne:
This PR adds an example project which demonstrates how to implement a distributed cache for PostHog feature flag definitions and local evaluation using Cloudflare KV storage. It showcases a split read/write pattern optimized for edge workers where low-latency flag evaluation is critical.

See the [README.md](https://github.com/PostHog/posthog-js/blob/84bce793e5679e2bec349e4122ca63658e51ba29/examples/example-cloudflare-kv-cache/README.md) for more details.

### DIFF
--- a/examples/example-cloudflare-kv-cache/wrangler.toml
+++ b/examples/example-cloudflare-kv-cache/wrangler.toml
@@ -1,4 +1,4 @@
-name = "example-cloudeflare-kv-cache"
+name = "example-cloudflare-kv-cache"
 main = "src/worker.ts"
 compatibility_date = "2024-11-19"
 


examples/example-cloudflare-kv-cache/README.md
```diff
@@ -0,0 +1,99 @@
+# PostHog Cloudflare KV Cache Example
```
dmarticus, line 1: Love this contribution – what do you think about adding it to the website (e.g. a tutorial for flags or something) rather than burying it in the posthog-js examples directory? users can't grep this and don't really think to look here, so while it's useful for folks at e.g. posthog who are working on this stuff, I think it would be more powerful if it were on the website.

dmarticus: I feel this way generally; like I think any time we have how-to guides with something technical, it makes sense to write the docs on the website in addition (exclusively?) to wherever they're documented in code.

dmarticus (approved):
changes look good, left a comment about potentially writing up this implementation guide on the website.

examples/example-cloudflare-kv-cache/README.md
dustinbyrne: Good point! I'll do a write up next week. 👍

PR #2602, opened 2025-11-19
ci: Fix turbo command not found

dustinbyrne:
Releases are failing, see https://github.com/PostHog/posthog-js/actions/runs/19512554077/job/55855787513#step:4:9


pauldambra (approved):
Maybe the robot is right about pnpm exec but 🤷

PR #2635, opened 2025-11-25
fix: use ph persistence instead of localStorage on surveys

alexedwardjones:
## Problem

A fix went in for [local storage usage](https://github.com/PostHog/posthog-js/issues/1898) in surveys but it was [reverted](https://github.com/PostHog/posthog-js/commit/7e568fe09e6d642a01397327b54b8aea0e836fde) due to a bug.

My understanding is that [the bug](https://github.com/PostHog/posthog/pull/39126) was related to `persistence.isDisabled is not a function`. This PR reapplies the fix with a few extra checks around the isDisabled function and falls back to `_disabled` if necessary.

## Changes

- Reapplied the persistence migration for surveys (using the shared persistence layer instead of touching localStorage directly).
- Guarded the persistence check with isFunction and a _disabled fallback so we never call a missing isDisabled.
- Added a regression test (surveys-persistence.test.ts) that simulates a “legacy” instance without isDisabled to make sure we don’t regress again.

### Libraries affected

- [ ] All of them
- [X] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code (runs for packages/browser/src/__tests__/extensions/surveys-persistence.test.ts; rest of the suite currently fails earlier because other packages are missing dev deps, so reviewers may want to re-run once the workspace installs)
- [x] Accounted for cross-platform impact (surveys only touch the browser SDK)
- [x] Backwards compatibility (still falls back to localStorage when persistence is disabled or unavailable)
- [x] Bundle size considerations (only adds a tiny helper import; no new runtime dependencies)

### If releasing new changes

- [x] Ran pnpm changeset (see .changeset/giant-lamps-cover.md or whatever file you created)
- [] Added the release label once you’re ready to publish

### DIFF
--- a/.changeset/giant-lamps-cover.md
+++ b/.changeset/giant-lamps-cover.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+use ph persistence instead of localStorage on surveys
--- a/packages/browser/playwright/mocked/surveys/core-display-logic.spec.ts
+++ b/packages/browser/playwright/mocked/surveys/core-display-logic.spec.ts
@@ -75,7 +75,8 @@ test.describe('surveys - core display logic', () => {
 
         expect(
             await page.evaluate(() => {
-                return window.localStorage.getItem('seenSurvey_123')
+                // @ts-expect-error - posthog is added to window in test setup
+                return window.posthog.persistence.get_property('seenSurvey_123')
             })
         ).toBeTruthy()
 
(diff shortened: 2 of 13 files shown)


lucasheriques (commented):
hey Alex! thanks for getting this back on. I'll review and test today and get back to you ASAP

PR #2636, opened 2025-11-25
feat(err): forward session and distinct id in express middleware

hpouillot:
## Changes

- forward session id and distinct id in express middleware

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/extensions/express.ts
+++ b/packages/node/src/extensions/express.ts
@@ -36,19 +36,21 @@ function posthogErrorHandler(posthog: PostHogBackendClient): ExpressErrorMiddlew
     const distinctId: string | undefined = req.headers['x-posthog-distinct-id'] as string | undefined
     const syntheticException = new Error('Synthetic exception')
     const hint: CoreErrorTracking.EventHint = { mechanism: { type: 'middleware', handled: false }, syntheticException }
-    const ip = req.headers['x-forwarded-for'] || req.socket.remoteAddress
-    const path = req.path // Gets the path of the request
+
     posthog.addPendingPromise(
       ErrorTracking.buildEventMessage(error, hint, distinctId, {
         $session_id: sessionId,
         $current_url: req.url,
         $request_method: req.method,
-        $request_path: path,
+        $request_path: req.path,
         $user_agent: req.headers['user-agent'],
         $response_status_code: res.statusCode,
-        $ip: ip,
-      }).then((msg) => posthog.capture(msg))
+        $ip: req.headers['x-forwarded-for'] || req?.socket?.remoteAddress,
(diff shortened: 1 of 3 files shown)


packages/node/src/extensions/express.ts
```diff
@@ -28,13 +28,24 @@ export function setupExpressErrorHandler(
...
+
+function posthogErrorHandler(posthog: PostHogBackendClient): ExpressErrorMiddleware {
+  return (error: MiddlewareError, req, res, next: (error: MiddlewareError) => void): void => {
+    const sessionId: string | undefined = req.headers['x-posthog-session-id'] as string | undefined
+    const distinctId: string | undefined = req.headers['x-posthog-distinct-id'] as string | undefined
+    const syntheticException = new Error('Synthetic exception')
+    const hint: CoreErrorTracking.EventHint = { mechanism: { type: 'middleware', handled: false }, syntheticException }
+    posthog.addPendingPromise(
+      ErrorTracking.buildEventMessage(error, hint, distinctId ?? uuidv7(), {
+        $process_person_profile: distinctId != undefined,
```
daibhin: We do this check inside the `buildEventMessage` function
```suggestion
```

packages/node/src/extensions/express.ts
```diff
@@ -28,13 +28,24 @@ export function setupExpressErrorHandler(
...
+        $session_id: sessionId,
+        $current_url: req.url,
+        method: req.method,
```
daibhin: We use the `$` in the Python SDK. We should probably try to be consistent with all this stuff
```suggestion
        $request_method: req.method,
```

packages/node/src/extensions/express.ts
```diff
@@ -28,13 +28,24 @@ export function setupExpressErrorHandler(
...
+        status_code: res.statusCode,
```
daibhin: Don't think we capture it elsewhere but maybe this should be `$status_code` to match the others?

daibhin: Worth capturing some more context like we do in Python:
https://github.com/PostHog/posthog-python/blob/494c78675d4c0e4bb3ac05acd9fa3164524839c9/posthog/integrations/django.py#L147-L163

daibhin (approved):
Nice! Totally not part of this PR but maybe we should have wrapper for non-erroring requests too so that people can instrument their whole Express app

PR #2640, opened 2025-11-25
fix: release workflow OIDC

hpouillot:
## Changes

- add "id-token: write" permissions to release workflow

### DIFF
--- a/.github/actions/release/action.yaml
+++ b/.github/actions/release/action.yaml
@@ -41,6 +41,8 @@ runs:
       shell: bash
       run: |
         pnpm publish --filter=${{ inputs.package_name }} --access public
+      env:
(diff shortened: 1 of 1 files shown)


.github/actions/release/action.yaml
rafaeelaudibert, line 1: Will this work as expected since we're calling out to the action rather than having it inline inside the workflow? My gut says this won't work and we'll need to change how this works

hpouillot: Yes, let's see if it works, we cannot set the permissions at the action level so it should inherit the one from the workflow ?

rafaeelaudibert: Maybe! Or maybe we can't use actions

PR #2643, opened 2025-11-26
fix: Only match prefixes in sensitive name regex

luke-belton:
It looks like this has been in the SDK forever, but the regex pattern we use was matching any substring for most of our sensitive looking fields (everything except `cc`). I think the intent would've been to only look at prefixes, as we were doing for `cc`.

Currently if, for example, `exp` appeared anywhere inside the `name`/`id` attribute of an element then we wouldn't capture it.

https://posthoghelp.zendesk.com/agent/tickets/42648 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/56562)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/autocapture-utils.test.ts
+++ b/packages/browser/src/__tests__/autocapture-utils.test.ts
@@ -398,34 +398,36 @@ describe(`Autocapture utility functions`, () => {
             expect(shouldCaptureElement(input)).toBe(false)
         })
 
-        it(`should not include fields with sensitive names`, () => {
-            const sensitiveNames = [
-                `cc_name`,
-                `card-num`,
-                `ccnum`,
-                `credit-card_number`,
-                `credit_card[number]`,
-                `csc num`,
-                `CVC`,
-                `Expiration`,
-                `password`,
-                `pwd`,
-                `routing`,
-                `routing-number`,
-                `security code`,
-                `seccode`,
-                `security number`,
-                `social sec`,
-                `SsN`,
-            ]
-            sensitiveNames.forEach((name) => {
-                input.name = ''
-                expect(shouldCaptureElement(input)).toBe(true)
-
-                input.name = name
-                expect(shouldCaptureElement(input)).toBe(false)
-            })
+        it.each([
+            'cc_name',
+            'card-num',
+            'ccnum',
+            'credit-card_number',
+            'credit_card[number]',
+            'csc num',
+            'CVC',
+            'Expiration',
+            'password',
+            'pwd',
+            'routing',
+            'routing-number',
+            'security code',
+            'seccode',
+            'security number',
+            'social sec',
+            'SsN',
+        ])(`should not include fields with sensitive name: %s`, (name) => {
+            input.name = name
+            expect(shouldCaptureElement(input)).toBe(false)
         })
+
+        it.each(['document-expiry', 'success_message', 'account-settings', 'compass-icon'])(
+            `should include fields with sensitive substrings that are not prefixes: %s`,
+            (name) => {
+                input.name = name
+                expect(shouldCaptureElement(input)).toBe(true)
+            }
+        )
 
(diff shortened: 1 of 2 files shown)


packages/browser/src/autocapture-utils.ts
```diff
@@ -375,7 +375,7 @@ export function shouldCaptureElement(el: Element): boolean {
     if (isString(name)) {
         // it's possible for el.name or el.id to be a DOM element if el is a form with a child input[name="name"]
         const sensitiveNameRegex =
-            /^cc|cardnum|ccnum|creditcard|csc|cvc|cvv|exp|pass|pwd|routing|seccode|securitycode|securitynum|socialsec|socsec|ssn/i
+            /^(cc|cardnum|ccnum|creditcard|csc|cvc|cvv|exp|pass|pwd|routing|seccode|securitycode|securitynum|socialsec|socsec|ssn)/i
```
pauldambra, line 378: i _think_ i understand the change here but

can you ask Claude to write parameterised tests proving the change and behaviour?

packages/browser/src/__tests__/autocapture-utils.test.ts
```diff
@@ -398,34 +398,36 @@ describe(`Autocapture utility functions`, () => {
...
+            'seccode',
+            'security number',
+            'social sec',
+            'SsN',
+        ])(`should not include fields with sensitive name: %s`, (name) => {
+            input.name = name
+            expect(shouldCaptureElement(input)).toBe(false)
+        })
 
+        it.each(['document-expiry', 'success_message', 'account-settings', 'compass-icon'])(
```
luke-belton, line 424: `document-expiry` and `compass-icon` would have failed before because of `exp` and `pass` appearing anywhere in the string.

`success_message` and `account-setting` would have passed because `cc` wasn't at the start.

packages/browser/src/autocapture-utils.ts
luke-belton: made a change - tbh I'm not totally convinced by this set of sensitive patterns anyway - there's lots of things that aren't sensitive that might start with `exp`.

Also there's part of me that thinks we shouldn't have the check here anyway:

https://github.com/PostHog/posthog-js/blob/f7d6b69e40e678d53cc9374f048e8ba9dd9c2a44/packages/browser/src/autocapture.ts#L41-L44

if a user has added the attributes to an element (even if it's a potentially sensitive element), then surely we should capture the attributes regardless?

pauldambra: it is way more preferable to miss something safe than to capture something sensitive!  
  
what we should do is let someone override this so that if they disagree then they can set their own and we add a property to say "hey, this is your regex, it's not on us"  
  
(that's how we do it in replay payload capture"

luke-belton: Having got into this a bit more, the way things are at the moment (as I understand it), we detect an element is sensitive with `shouldCaptureElement`, but that check doesn't apply in all places, so the elements would be captured regardless in `elements_chain`. I think it maybe needs some wider thought than this one small fix

luke-belton: https://github.com/PostHog/posthog-js/blob/19fc73bf2aac23bc6a9ce4e3bddd64641afec71a/src/autocapture.ts#L106

I believe this is the check we do for capturing the `elements_chain`

luke-belton: I'm going to close this PR for now - working on something over on https://github.com/PostHog/posthog-js/pull/2647 instead

PR #2645, opened 2025-11-26
chore: release workflow OIDC

hpouillot:
## Problem

- OIDC permissions might only work when npm publish is called inside the workflow, not a nested composite action
Previous attempt: https://github.com/PostHog/posthog-js/actions/runs/19681333980/job/56375850214

## Changes

- inlining

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -33,6 +33,11 @@ runs:
         registry-url: https://registry.npmjs.org
         cache: "pnpm"
 
+    # Ensure latest npm with OIDC support (npm 11.5.1+)
+    - name: Ensure modern npm (OIDC support)
+      shell: bash
+      run: npm install -g npm@11.6.4
+
     - uses: actions/cache@v4
       id: pnpm-cache
       with:


.github/actions/setup/action.yaml
```diff
@@ -33,6 +33,11 @@ runs:
         registry-url: https://registry.npmjs.org
         cache: "pnpm"
 
+    # Ensure latest npm with OIDC support (npm 11.5.1+)
+    - name: Ensure modern npm (OIDC support)
+      shell: bash
+      run: npm install -g npm@11.6.4
```
rafaeelaudibert, line 39: Do we need npm here? Isn't pnpm enough? That's a bit weird  
  
Also, can we not cache this somehow or specify the version when setting node js up?

.github/workflows/release.yml
```diff
@@ -52,14 +52,62 @@ jobs:
...
+        if: steps.check-package-version.outputs.is-new-version == 'true'
+        run: |
+          git tag -a "${{ matrix.package.name }}@${{ steps.check-package-version.outputs.committed-version }}" -m "${{ matrix.package.name }}@${{ steps.check-package-version.outputs.committed-version }}"
+
+      - name: Publish ${{ matrix.package.name }} to NPM
+        if: steps.check-package-version.outputs.is-new-version == 'true'
+        run: |
+          pnpm publish --filter=${{ matrix.package.name }} --access public
+        env:
+          NPM_CONFIG_PROVENANCE: true
```
rafaeelaudibert, line 76: Do we really need this? My understanding reading some blogposts is that this should all be automatic

.github/actions/setup/action.yaml
hpouillot: pnpm uses npm behind the scene so we want to make sure it uses a recent version (from 5 month ago)

There is no way to configure npm version inside setup-node. https://github.com/actions/setup-node

.github/workflows/release.yml
hpouillot: Not sure, It's not documented on pnpm but saw this here: https://github.com/pnpm/pnpm/issues/6435

PR #2647, opened 2025-11-26
chore: refactor sensitive data capture

luke-belton:
The logic for whether or not to capture certain elements (including sensitive data) was a bit confusing to follow, so trying to clear that up in this PR.

Originally I opened https://github.com/PostHog/posthog-js/pull/2643 but that didn't really address the core issue.

In particular there were some cases such as that identified in  [#42648](https://posthoghelp.zendesk.com/agent/tickets/42648) where we'd identify sensitive data and prevent capturing of attributes, but still capture the actual elements (including attributes) inside `$elements_chain`.

  ## Changes

  - Removes the sensitive field name regex by removing `shouldCaptureElement` (the remaining logic is now split into `isExplicitNoCapture`/`isExplicitCapture`)
  - That regex matched element names/IDs like `password`, `ccnum`, `ssn` and blocked capture
  - **Why this is safe:**
    - `isSensitiveElement` already protects ALL input/select/textarea elements by type
    - `shouldCaptureValue` still filters actual CC/SSN patterns from values
    - The regex was inconsistent (elements still appeared in `$elements_chain`) and caused false positives (e.g., blocking href capture and `data-ph-capture-attribute` attributes from `<a id="password" data-ph-capture-attribute-key="value">Reset password</a>`)
  - Explicit opt-out via `ph-no-capture`/`ph-sensitive` classes still works as expected

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/young-clocks-call.md
+++ b/.changeset/young-clocks-call.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fix autocapture incorrectly blocking capture from elements with IDs/names containing sensitive looking substrings (e.g., `<a id="password">`). Sensitive data protection is maintained through element type checks and value pattern matching, as well as `ph-no-capture` element attributes.


packages/browser/src/autocapture.ts
```diff
@@ -38,8 +37,7 @@ function limitText(length: number, text: string): string {
 }
 
 export function getAugmentPropertiesFromElement(elem: Element): Properties {
-    const shouldCaptureEl = shouldCaptureElement(elem)
-    if (!shouldCaptureEl) {
+    if (isExplicitNoCapture(elem)) {
```
luke-belton, line 40: basically, let `ph-no-capture` take precedence here. But otherwise if a user explicitly sets an attribute on an element then we should capture it as a property on the event.

packages/browser/src/autocapture-utils.ts
```diff
@@ -262,16 +266,15 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false
 }
 
 /*
- * Check whether a DOM event should be "captured" or if it may contain sensitive data
- * using a variety of heuristics.
+ * Check whether a DOM event should be "captured" using a variety of heuristics.
```
luke-belton, line 269: wasn't really checking for sensitive data

packages/browser/src/__tests__/autocapture-utils.test.ts
```diff
@@ -121,6 +120,30 @@ describe(`Autocapture utility functions`, () => {
             el.innerHTML = `Mixed "double" and 'single' quotes`
             expect(getSafeText(el)).toBe(`Mixed "double" and 'single' quotes`)
         })
+
+        it(`should collect text from sensitive elements with ph-include class`, () => {
```
luke-belton, line 124: adding tests for a couple of cases we weren't really checking

packages/browser/src/autocapture-utils.ts
```diff
@@ -338,56 +341,35 @@ export function shouldCaptureDomEvent(
...
-    const name = (el as HTMLInputElement).name || el.id || ''
-    // See https://github.com/posthog/posthog-js/issues/165
-    // Under specific circumstances a bug caused .replace to be called on a DOM element
-    // instead of a string, removing the element from the page. Ensure this issue is mitigated.
-    if (isString(name)) {
-        // it's possible for el.name or el.id to be a DOM element if el is a form with a child input[name="name"]
-        const sensitiveNameRegex =
-            /^cc|cardnum|ccnum|creditcard|csc|cvc|cvv|exp|pass|pwd|routing|seccode|securitycode|securitynum|socialsec|socsec|ssn/i
-        if (sensitiveNameRegex.test(name.replace(/[^a-zA-Z0-9]/g, ''))) {
-            return false
```
luke-belton, line 380: we have a more broader scope (i.e. safer) check for all of this in `isSensitiveElement`

packages/browser/src/autocapture-utils.ts
```diff
@@ -73,7 +73,11 @@ export function makeSafeText(s: string | null | undefined): string | null {
 export function getSafeText(el: Element): string {
     let elText = ''
 
-    if (shouldCaptureElement(el) && !isSensitiveElement(el) && el.childNodes && el.childNodes.length) {
+    if (
+        (isExplicitCapture(el) || (!isExplicitNoCapture(el) && !isSensitiveElement(el))) &&
```
luke-belton, line 77: see note below too, but the checks we removed from `shouldCaptureElement` are covered more broadly inside `isSensitiveElement`

packages/browser/src/__tests__/autocapture-utils.test.ts
```diff
@@ -348,110 +371,64 @@ describe(`Autocapture utility functions`, () => {
...
+            el.className = `test1 ph-include test2`
+            expect(isExplicitCapture(el)).toBe(true)
+        })
+
+        it(`should return false for elements without class "ph-include"`, () => {
+            const el = document!.createElement(`div`)
+            el.className = `test1 test2`
+            expect(isExplicitCapture(el)).toBe(false)
+        })
+    })
```
luke-belton, line 386: we actually don't document `ph-include` anymore as far as I can see, but keeping it around for backward compatibility

packages/browser/src/__tests__/autocapture-utils.test.ts
```diff
@@ -348,110 +371,64 @@ describe(`Autocapture utility functions`, () => {
...
-                `SsN`,
-            ]
-            sensitiveNames.forEach((name) => {
-                input.name = ''
-                expect(shouldCaptureElement(input)).toBe(true)
-
-                input.name = name
-                expect(shouldCaptureElement(input)).toBe(false)
-            })
-        })
```
luke-belton, line 428: this was in some ways redundant because we'd still capture the element with sensitive names inside `elements_chain`

packages/browser/src/__tests__/autocapture.test.ts
```diff
@@ -292,22 +302,28 @@ describe('Autocapture system', () => {
...
+        it('should not collect augment from input with class "ph-no-capture"', () => {
+            const props = getAugmentPropertiesFromElement(noCaptureInput)
 
             expect(props).toStrictEqual({})
         })
 
+        it('should collect augment from the hidden element value', () => {
+            const props = getAugmentPropertiesFromElement(hidden)
+
+            expect(props).toStrictEqual({ 'on-the-hidden': 'is on the hidden' })
```
luke-belton, line 320: if a user has added a `data-ph-capture-attribute` attribute, then we should take that as explicit opt-in that they want the attribute captured - same for tests below

packages/browser/src/__tests__/autocapture.test.ts
```diff
@@ -866,6 +882,62 @@ describe('Autocapture system', () => {
             expect(beforeSendMock.mock.calls[0][0].properties['$elements'][0]).not.toHaveProperty('attr__href')
         })
 
+        it('captures href from anchors with IDs containing sensitive substrings like password', () => {
+            const elTarget = document.createElement('span')
+            const elAnchor = document.createElement('a')
+            elAnchor.appendChild(elTarget)
+            elAnchor.setAttribute('id', 'password')
```
luke-belton: this probably would have failed before because of the checks inside `shouldCaptureElement` on 'sensitive names'

packages/browser/src/autocapture-utils.ts
pauldambra: one thing we'll need to consider here is whether we're introducing breaking behaviour

if someone is relying on old behaviour for PII avoidance we need to be sure they're still not capturing

luke-belton: it's a bit of a dilemma because I _think_ the current behavior is broken anyway, meaning that in some cases we'd still capture sensitive data even when defended elsewhere by `shouldCaptureElement`. I still need to test a bit more to confirm exactly what conditions the current 'gap' happens under. But knowing it's broken in at least some cases already, this leaves us with a couple of options:
* fix the bug that could leak sensitive data based on the previous implementation
* just remove the previous implementation and try to have one path for sensitive data identification/capture going forwards (and don't rely on fuzzy heuristics like looking for suspect sub-strings in element `id`/`name` attributes!)
* something else I'm missing...

pauldambra: the joy of a public API in an SDK :)

PR #2652, opened 2025-11-27
fix: Use deployer app token for commiting lockfile

rafaeelaudibert:


.github/workflows/label-version-bump.yml
```diff
@@ -9,7 +9,7 @@ jobs:
         name: Bump versions based on PR label
         runs-on: ubuntu-22.04
         permissions:
-            contents: write
+            contents: read
```
rafaeelaudibert, line 12: Permission for implicit `GITHUB_TOKEN` can be just for read (checkout)

PR #2661, opened 2025-11-28
feat(surveys): support actions with event properties

adboio:
## Problem

we don't support triggering surveys based on actions with event filters

## Changes

- added support for triggering surveys based on actions with event filters :smiley:
    - uses the same logic as events and cancel events, just with a bit of transformation since the shape is different

**_this is largely untested, but it's behind a feature flag, it'll be easier to test once deployed!_**

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/shaky-rats-sip.md
+++ b/.changeset/shaky-rats-sip.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': minor
+---
+
+add support for surveys triggered by actions with event property filters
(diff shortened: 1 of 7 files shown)


lucasheriques (approved):
looks good, just need to remember to add the minimum version they need on the main app too. also, before releasin

 i'd also test this on an OLDER version of the sdk with new features, to make sure it doesn't break. before actually flipping the feature flag to everyone

PR #2664, opened 2025-12-01
chore: autobump versions

hpouillot:
## Problem

- We cannot push on main
https://posthog.slack.com/archives/C09V86BQT5X/p1764256757621609

## Changes

- Create a PR to do so


hpouillot:
If we want to do that, we could use the change github action here: https://github.com/changesets/action. It will create a PR with all the changes and update the PR when there are new changes pushed to the base branch.

PR #2677, opened 2025-12-02
feat(surveys): allow customization of input text color

lucasheriques:
## Problem

we can't customize the styling for text input, which makes surveys hard to support on websites with dark mode

## Changes

allow customizing those values

relevant ticket: https://posthoghelp.zendesk.com/agent/tickets/43604 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/48182)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/happy-hands-lick.md
+++ b/.changeset/happy-hands-lick.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+chore: allow customizing text input and background for surveys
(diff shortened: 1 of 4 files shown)


adboio (approved):
in the case of sites with dark mode, how do customers usually handle that? are they passing in css variables to these inputs? just want to make sure this solution works for that

lucasheriques:
@adboio yup they just use their css variables. the customer on this ticket is using it, but the only part they can't change is the input (which is why we're adding too!)

PR #2687, opened 2025-12-04
chore: add failure alerts

hpouillot:


.github/workflows/posthog-com-upgrade.yml
```diff
@@ -82,3 +82,20 @@ jobs:
...
         run: |
           echo "PostHog.com pull request for ${{ github.event.inputs.package_name }} version ${{ github.event.inputs.package_version }} ready: ${{ steps.com-repo-pr.outputs.pull-request-url }}"
+
+      # a CDP destination is set-up to alert when we get this event
+      - name: Send failure event to PostHog
+        if: ${{ failure() }}
+        uses: PostHog/posthog-github-action@v0.1
+        with:
+          posthog-token: "${{ secrets.POSTHOG_PROJECT_API_KEY }}"
+          event: "posthog-js-github-update-workflow-failure"
```
hpouillot: I changed the event as we don't have the same properties. Can we keep the same event or will the destination fails to access some info ?

marandaneto: the event will be sent anyway https://github.com/PostHog/posthog-github-action/blob/d8898a3c51b1159a9afd879df4fdc771d81eba8c/index.js#L28
checking the cdp

pauldambra: not sure...
feels like we should have 1 event and then fix the destination
but no strong opinion, it's easy enough to duplicate the destination or whatevs

pauldambra (approved):

.github/workflows/posthog-com-upgrade.yml
hpouillot: I agree, let's keep only one event. I added the link to the destination as well if we want to add infos

marandaneto (approved):

PR #2700, opened 2025-12-05
feat: conversations widget

veryayskiy:
## Problem

Conversations widget version 1

## Changes

Adding conversations extension

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
+++ b/packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
@@ -186,15 +186,6 @@ exports[`config snapshot for PostHogConfig 1`] = `
     "false",
     "true"
   ],
-  "conversations": [
-    "undefined",
-    {
-      "position": [
-        "undefined",
-        "{ bottom?: string | undefined; right?: string | undefined; left?: string | undefined; top?: string | undefined; }"
-      ]
-    }
-  ],
   "disable_web_experiments": [
     "false",
     "true"


packages/browser/src/extensions/conversations/external/components/ConversationsWidget.tsx
```diff
@@ -0,0 +1,500 @@
...
+    }
+
+    /**
+     * Check if we need to show the identification form
+     */
+    private _needsIdentification(config: ConversationsRemoteConfig, traits: UserProvidedTraits | null): boolean {
+        // If requireEmail is not set, no identification needed
+        if (!config.requireEmail) {
+            //return false
+        }
```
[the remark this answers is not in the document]
veryayskiy: This one is temporary as i need to add a control from config and now i need to test it.

packages/browser/src/extensions/conversations/index.tsx
```diff
@@ -0,0 +1,595 @@
...
+    private _pollIntervalId: number | null = null
+    private _lastMessageTimestamp: string | null = null
+    private _isPolling: boolean = false
+    private _unsubscribeIdentifyListener: (() => void) | null = null
+    private _unreadCount: number = 0
+    // SECURITY: widget_session_id is the key for access control
+    // This is a random UUID that only this browser knows
+    private _widgetSessionId: string
+
+    constructor(config: ConversationsRemoteConfig, apiHelpers: ConversationsApiHelpers) {
```
pauldambra: you can pass in the posthog instance here when init'g the conversations class and then you have access to thinks like sending a request (i didn't de-dupe and check if you're doing significantly different things though)

```suggestion
    constructor(config: ConversationsRemoteConfig, apiHelpers: ConversationsApiHelpers, private readonly _posthog) {
```

packages/browser/src/extensions/conversations/index.tsx
```diff
@@ -0,0 +1,595 @@
...
+            const queryParams: Record<string, string> = {
+                widget_session_id: this._widgetSessionId,
+                limit: '50',
+            }
+
+            if (after) {
+                queryParams.after = after
+            }
+
+            this._api.sendRequest({
```
pauldambra: and nit-picky even if not able to re-use the main posthog one you can extract a method here :grinning:

packages/browser/src/extensions/conversations/persistence.ts
```diff
@@ -0,0 +1,267 @@
...
+     * SECURITY: This is the key for access control. Only the browser that created
+     * the widget_session_id can access tickets associated with it.
+     */
+    getOrCreateWidgetSessionId(): string {
+        // Return cached value if available
+        if (this._cachedWidgetSessionId) {
+            return this._cachedWidgetSessionId
+        }
+
+        if (typeof localStorage === 'undefined') {
```
pauldambra: they do give us persistence config when they start posthog  
cookie, localstorage, memory   
  
we could obey those and re-use them here  
  
e.g. if you've said memory only we shouldn't try to use localStorage  
  
i don't know whether you _need_ to keep persistence separate because of the semi-sensitive nature of the session id  
  
otherise you can use `this._posthog.persistence` instead of custom handling

packages/browser/src/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+
+        // Destroy the manager if it exists
+        if (this._conversationsManager) {
+            this._conversationsManager.destroy()
+            this._conversationsManager = null
+        }
+
+        // Reset state
+        this._isConversationsEnabled = undefined
+        this._remoteConfig = null
```
pauldambra: most of this could be lazy loaded, since a reset without a lazily loaded implementation is a no-op anyway

really it should be very close to `this._lazyLoadedConversations?.reset()`

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+        if (this._instance.config.disable_conversations) {
+            logger.info('Conversations disabled. Not loading.')
+            return
+        }
+        if (this._instance.config.cookieless_mode && this._instance.consent.isOptedOut()) {
+            logger.info('Not loading conversations in cookieless mode without consent.')
+            return
+        }
+
+        const phExtensions = assignableWindow?.__PosthogExtensions__
```
pauldambra, line 117: _very_ unlikely not to find this unless you boot very very early in the lifecycle  
so you can save some bytes by assuming it is there

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+        if (!this._isConversationsEnabled) {
+            logger.info('Conversations not enabled for this team.')
+            return
+        }
+
+        // Check if we have the required config
+        if (!this._remoteConfig || !this._remoteConfig.token) {
+            logger.error('Conversations enabled but missing token in remote config.')
+            return
+        }
```
pauldambra, line 136: we're doing checks like this in a couple of places now, so for byte obsession we could centralise them... although they might be different enough in each case that they don't benefit from moving (since if it ends up with three different but similar methods then it's probably longer in the minified code)

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+                } else {
+                    this._completeInitialization(phExtensions.initConversations)
+                }
+                this._isInitializing = false
+            })
+        } catch (e) {
+            this._handleLoadError('Error initializing conversations', e)
+            this._isInitializing = false
+        }
+    }
```
pauldambra, line 174: not your fault here!  
  
i feel like we have almost exactly this code in about 6 places now  
  
i bet there's space for a lazy loader to squeeze some bytes out

packages/browser/src/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+            this._handleLoadError('Error initializing conversations', e)
+            this._isInitializing = false
+        }
+    }
+
+    /**
+     * Create API helpers that wrap PostHog internal methods.
+     * These bound functions ensure the methods work correctly even after minification.
+     */
+    private _createApiHelpers(): ConversationsApiHelpers {
```
pauldambra: we _shouldn't_ need this  
  
i think maybe there's something else going on

packages/browser/src/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+            logger.error('Cannot complete initialization: remote config is null')
+            return
+        }
+
+        try {
+            // Create API helpers that wrap PostHog internal methods
+            const apiHelpers = this._createApiHelpers()
+
+            // Pass config and API helpers to the extension
+            this._conversationsManager = initConversationsFn(this._remoteConfig, apiHelpers)
```
pauldambra: from experience i'd call this `lazyConversationManager` or `lazyLoadedConversations` so it's easier for people to see what is lazy and what isn't

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -0,0 +1,243 @@
...
+        logger.error(message, error)
+        this._conversationsManager = null
+        this._isInitializing = false
+    }
+
+    /**
+     * Show the conversations widget (button and chat panel)
+     */
+    enable(): void {
+        if (!this._conversationsManager) {
```
pauldambra, line 205: we hit this problem in replay  
  
what if someone calls enable or disable while you're lazy loading :see_no_evil:  
  
we had to put a flag into storage and then check that after lazy loading so we don't miss the call  
  
(a correctness thing that can come later)

packages/browser/src/posthog-core.ts
```diff
@@ -706,6 +710,10 @@ export class PostHog {
             this.surveys.loadIfEnabled()
         })
 
+        initTasks.push(() => {
+            this.conversations.loadIfEnabled()
```
pauldambra, line 714: it's on me not doing this in the first place

i think we can do something like this to completely defer startup   
  
(right now for all of these we have anything in their constructor running synchronously and can probably get away with avoiding that)

```suggestion
            this.conversations = new PostHogConversations(this)
            this.conversations.loadIfEnabled()
```

packages/browser/src/utils/globals.ts
```diff
@@ -184,6 +186,38 @@ export interface LazyLoadedDeadClicksAutocaptureInterface {
...
 }
 
+export interface LazyLoadedConversationsInterface {
+    enable: () => void
+    disable: () => void
+    destroy: () => void
+}
+
+/**
+ * API helpers passed from the main bundle to the lazy-loaded conversations extension.
```
pauldambra: noting here as well... this shouldn;t be causing a problem so since it was, something else funky is going on :melting_face:

packages/browser/src/extensions/conversations/posthog-conversations.ts
pauldambra, line 1: silly thing for neatness  
  
we should put this in `/extensions/conversations/posthog-conversations.ts`  
  
then it can have `/extensions/conversations/external/` which has the various dependencies of the lazy loaded class  
  
and _then_ we can lint that we aren't including those dependencies in the main bundle  
  
(we aren't right now (i don't think) but easy for someone to make a mistake in future)

packages/browser/src/extensions/conversations/persistence.ts
veryayskiy: Done

packages/browser/src/extensions/conversations/index.tsx
pauldambra: i played around with this
i could only save 121 _bytes_ 
by rearranging things

so, if it is usefu then remember you could just pass in the posthog instance
but tbh the only benefit right now would be to make it more similar to other extensions which might help in future

but limited benefit right in front of us

veryayskiy: I did it as well, here is the difference:

Before:
Screenshot 2025-12-15 at 19 34 23

After:
Screenshot 2025-12-15 at 19 40 45

veryayskiy: Not much of a difference, but i'd say it's way cleaner

packages/browser/src/extensions/conversations/index.tsx
veryayskiy: I went with the one above

packages/browser/src/posthog-conversations.ts
veryayskiy: Yep, moved inside the extension

packages/browser/src/extensions/conversations/posthog-conversations.ts
veryayskiy: I think i know 2 other places where we can remove that to save some weight:
Screenshot 2025-12-15 at 19 56 35

cc @adboio @lucasheriques

veryayskiy: oh wow, removing that crashes the CI - it's screaming that phExtensions may be undefined alarm alarm alarm!

packages/browser/src/extensions/conversations/index.tsx
pauldambra: and the SDK is like dieting... you have to fight yourself everytime you think "it's just one biscuit"

packages/browser/src/extensions/conversations/posthog-conversations.ts
pauldambra: oh... yuck  
i guess _technically_ it is since it's not auto on window  
but it will always be on window  
i bet there's something we can do in globals.ts to make sure typescript knows that we have made sure it's always present  
  
(but not blocking here)

packages/browser/src/extensions/conversations/posthog-conversations.ts
veryayskiy: Good point - done

packages/browser/src/__tests__/__snapshots__/config-snapshot.test.ts.snap
```diff
@@ -182,6 +182,19 @@ exports[`config snapshot for PostHogConfig 1`] = `
       ]
     }
   ],
+  "disable_conversations": [
+    "false",
+    "true"
+  ],
+  "conversations": [
```
pauldambra: ok, going through with a nitpicking mindset :)

we could have `conversations: false`, instead of `disable_conversations: true`  
  
but arguably `disable_conversations` is more clear

packages/browser/src/__tests__/extensions/conversations/conversations-manager.test.tsx
```diff
@@ -0,0 +1,484 @@
...
+        })
+
+        it('should not include distinct_id in getMessages request for security', async () => {
+            act(() => {
+                jest.advanceTimersByTime(5000)
+            })
+
+            const calls = (mockPosthog._send_request as jest.Mock).mock.calls
+            const getMessagesCall = calls.find((call) => call[0].url.includes('/widget/messages/'))
+            expect(getMessagesCall[0].url).not.toContain('distinct_id=')
```
pauldambra, line 333: very nitpicky... we could be including the actual distinct id but not keyed by distinct_id=   
so _really_ we should test the distinct id

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -0,0 +1,616 @@
...
+    private _lastMessageTimestamp: string | null = null
+    private _isPolling: boolean = false
+    private _unsubscribeIdentifyListener: (() => void) | null = null
+    private _unreadCount: number = 0
+    // SECURITY: widget_session_id is the key for access control
+    // This is a random UUID that only this browser knows
+    private _widgetSessionId: string
+
+    constructor(
+        config: ConversationsRemoteConfig,
```
pauldambra, line 41:
```suggestion
        private readonly _config: ConversationsRemoteConfig,
```

nitpick that you can avoid having to declare and assign to a field this way  
(really very stylistic though - don't think there's a right and wrong)

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -0,0 +1,616 @@
...
+
+            // Update current ticket ID
+            if (!this._currentTicketId) {
+                this._currentTicketId = response.ticket_id
+                this._persistence.saveTicketId(response.ticket_id)
+                logger.info('New ticket created', { ticketId: response.ticket_id })
+            }
+
+            // Track message sent
+            this._posthog.capture('$conversations_message_sent', {
```
pauldambra, line 322: we should _maybe_ exclude these `$conversations_etc` events from billing  
and then set up tracking in the billing report  
so that this is free while in beta  
  
but it's not going to be _sooo_ many events so maybe OK to have it be the event cost for now

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -0,0 +1,616 @@
...
+    private _startPolling(): void {
+        if (this._pollIntervalId) {
+            return // Already polling
+        }
+
+        // Poll immediately
+        this._pollMessages()
+
+        // Set up interval
+        this._pollIntervalId = window?.setInterval(() => {
```
pauldambra, line 449: we should listen to document visibility and pause and resume this polling when the page is not visible  
(it's one of the things that stops a browser freezing a tab)  
and it saves us handling API calls for tabs nobody is on  
  
(not blocking now, but as a follow-up)

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -0,0 +1,616 @@
...
+     * Render the widget to the DOM
+     */
+    private _renderWidget(initialState: ConversationsWidgetState, initialUserTraits: UserProvidedTraits | null): void {
+        if (!document) {
+            logger.info('Conversations widget not rendered: Document not available')
+            return
+        }
+
+        // Create container if it doesn't exist
+        let container = document.getElementById(WIDGET_CONTAINER_ID) as HTMLDivElement
```
pauldambra, line 579: another potential follow-up you could register a dom loaded callback here if the document isn't ready yet, so that if there's a race this does get called  
  
(i think it's maybe unlikely we get to the point of executing this code before the DOM is loaded though so :shrug:)

packages/browser/src/extensions/conversations/external/persistence.ts
```diff
@@ -0,0 +1,257 @@
...
+
+    /**
+     * Get or create the widget session ID (random UUID for access control).
+     * This ID is generated once per browser and persists across sessions.
+     * It is NOT tied to distinct_id - it stays the same even when user identifies.
+     *
+     * SECURITY: This is the key for access control. Only the browser that created
+     * the widget_session_id can access tickets associated with it.
+     */
+    getOrCreateWidgetSessionId(): string {
```
pauldambra, line 39: silly check,... since we default to local storage in persistence  
i guess this means if you have multiple tabs open they all have the same chat messages  
i assume that's ok

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -0,0 +1,235 @@
...
+import { isNullish, isUndefined, isBoolean, isNull } from '@posthog/core'
+
+const logger = createLogger('[Conversations]')
+
+/**
+ * Extract hostname from a domain string (handles URLs and plain hostnames)
+ */
+function extractHostname(domain: string): string | null {
+    // Remove protocol if present
+    let hostname = domain.replace(/^https?:\/\//, '')
```
pauldambra, line 15: silly nit pick that this won't work for capacitor (and maybe electron)  
and that might be OK

packages/browser/src/posthog-conversations-types.ts
```diff
@@ -0,0 +1,290 @@
...
+}
+
+/**
+ * Remote configuration for conversations from the PostHog server
+ */
+export interface ConversationsRemoteConfig {
+    /**
+     * Whether conversations are enabled for this team
+     */
+    enabled: boolean
```
pauldambra, line 8: could also avoid this by just returning false or nothing when disabled  
and that is the signal that it isn't enabled  
  
not a biggie, but we're sending `enabled: true` over the wire billions and billions of times but can avoid it

packages/browser/src/posthog-conversations-types.ts
```diff
@@ -0,0 +1,290 @@
...
+
+    /**
+     * Primary color for the widget UI
+     */
+    color?: string
+
+    /**
+     * Placeholder text for the message input
+     */
+    placeholderText?: string
```
pauldambra, line 29: could also nest some of these under `customization: {}`?  
i feel like as the amount of options grows a flat list will get annoying  
but it's nit-picky

packages/browser/src/posthog-conversations-types.ts
```diff
@@ -0,0 +1,290 @@
...
+    /**
+     * Number of unread messages from the team
+     */
+    unread_count?: number
+}
+
+/**
+ * Visual state of the conversations widget
+ */
+export enum ConversationsWidgetState {
```
pauldambra: https://www.totaltypescript.com/why-i-dont-like-typescript-enums  
  
could just be `export const ConversationsWidgetState = 'open' | 'closed'`

packages/browser/src/extensions/conversations/external/persistence.ts
veryayskiy: yes, that's correct as you could accidentally close one

packages/browser/src/posthog-conversations-types.ts
veryayskiy:
```
enum PackStatus {
  Draft = 0,
  Approved = 1,
  Shipped = 2,
}

// [0, "Draft", 1, "Approved", 2, "Shipped"]
console.log(Object.keys(PackStatus));
```

WAT?!?

PR #2711, opened 2025-12-08
feat(ai): Support both Vercel AI SDK 5 and 6

andrewm4894:
## Problem

Users on Vercel AI SDK 5 (stable) cannot upgrade to @posthog/ai without also upgrading to SDK 6 (beta). Similarly, SDK 6 users cannot use @posthog/ai if it only supports SDK 5.

## Changes

- Add runtime version detection via `model.specificationVersion` (v2 for SDK 5, v3 for SDK 6)
- Support both `LanguageModelV2` (SDK 5) and `LanguageModelV3` (SDK 6) types
- Move `@ai-sdk/provider` and `ai` to peerDependencies with dual version ranges:
  - `@ai-sdk/provider`: `^2.0.0 || ^3.0.0-beta.0`
  - `ai`: `^5.0.0 || ^6.0.0-beta.0`
- Rewrite `wrapVercelLanguageModel` to wrap models directly instead of using `wrapLanguageModel` from the `ai` package (which is version-specific)
- Preserve original model type through TypeScript generics
- Export type guards `isV2Model` and `isV3Model` for external use

## How it works

The wrapper detects the model version at runtime and applies instrumentation accordingly. Both SDK versions use the same streaming and generation patterns, so a single implementation handles both.

## Release info

This is a **minor** release - no breaking changes. Existing SDK 5 users can continue without modification.

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code (both V2 and V3 models tested)
- [x] Accounted for backwards compatibility (supports both SDK 5 and 6)

### If releasing new changes

- [x] Added the "release" label to the PR

## Related

- PostHog/llm-analytics-apps#28 - Test harness updates to support both SDK versions

### DIFF
--- a/playground/nextjs/pages/product-tours.tsx
+++ b/playground/nextjs/pages/product-tours.tsx
@@ -11,7 +11,7 @@ export default function ProductTours() {
 
     useEffect(() => {
         try {
-            posthog?.productTours.getProductTours((fetchedTours: ProductTour[]) => {
+            posthog?.productTours?.getProductTours((fetchedTours: ProductTour[]) => {
                 setLoading(false)
                 setTours(fetchedTours)
                 if (fetchedTours.length > 0) {


carlos-marchal-ph (commented):
Once again, this highlights the need for us to find a way to let users bring their own SDK to not enforce a version from our end.

I'm not super sure whether releasing a new major that couples the Vercel SDK to a beta release is the best idea. Technically they could then change some API before the official release, and then we'd have to bump major again to bring it up to date.

Do we have a timeline on when they plan to do the official release?

carlos-marchal-ph:
> Now that AI v6 official has dropped does this version need to be bumped to the stable version?

@kdawgwilk [PR](https://github.com/PostHog/posthog-js/pull/2836)

PR #2713, opened 2025-12-08
fix: can we minify all the usages of object.keys

pauldambra:
## Problem

repeated code is killer for bundle
particularly if the bundler can't minify it

## Changes

i think Object.keys is used a lot and can't be minified
let's check

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/.eslintrc.cjs
+++ b/packages/browser/.eslintrc.cjs
@@ -21,6 +21,7 @@ module.exports = {
                 'no-console': 'off',
                 'no-restricted-globals': 'off',
                 'compat/compat': 'off',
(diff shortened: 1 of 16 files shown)


pauldambra:
ugh, only -0.24% for all of that 👎

PR #2723, opened 2025-12-10
feat(product tours): support 'modal' steps

adboio:
### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes


lucasheriques (approved):
good to fix the package.json to not commit your local changes

and also make it so both survreys and product tours share the url matching, since the function is the same

otherwise 🚢

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -315,6 +396,20 @@ export class ProductTourManager {
         }
 
         const step = this._activeTour.steps[this._currentStepIndex]
+
+        // Modal step (no selector) - render without a target element
+        if (!step.selector) {
+            this._captureEvent('product tour step shown', {
```
lucasheriques, line 402: good to have the events as constants in an enum, similar to how we do it for surveys

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -1,17 +1,88 @@
...
 import { addEventListener } from '../../utils'
+import { isNull } from '@posthog/core'
 
 const logger = createLogger('[Product Tours]')
 
 const document = _document as Document
+const window = _window as Window & typeof globalThis
+
+// Tour condition checking utilities (moved from utils/product-tour-utils.ts)
+function doesTourUrlMatch(tour: ProductTour): boolean {
```
lucasheriques, line 23: share this utility with surveys or use the surveys, as it's the same really  
maybe we just pass the url and the target so not pass the entire product tour

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -1,17 +1,88 @@
...
+                return regex.test(currentUrl)
+            } catch {
+                return false
+            }
+        default:
+            return false
+    }
+}
+
+function doesTourSelectorMatch(tour: ProductTour): boolean {
```
lucasheriques, line 50: survey also has a selector filter that we can prob reuse

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -1,17 +1,88 @@
...
+    }
+
+    try {
+        return !isNull(document.querySelector(conditions.selector))
+    } catch {
+        return false
+    }
+}
+
+function isTourInDateRange(tour: ProductTour): boolean {
```
lucasheriques, line 63: again can also share this with surveys 🤣 since exact same names too

PR #2725, opened 2025-12-10
fix: include initial person props in $identify when group() called first

ordehi:
Initial person properties lost when `group()` called before `identify()`

## Bug

Calling `group()` with properties before `identify()` causes initial person properties (UTM params, referrer, etc.) to be lost from the person profile.
```javascript
posthog.group('organization', 'org_123', { name: 'Acme' })
posthog.identify('user_123')
// Result: Person missing $initial_utm_source, $initial_referrer, etc.
```

**Root cause:** `group()` triggers `_calculate_set_once_properties()` which marks initial person props as "sent" via `_personProcessingSetOncePropertiesSent = true`. When `identify()` is called later, it skips sending these props because the flag is already set. However, the server ignores `$set_once` on `$groupidentify` events, so the initial props are never actually saved.

## Fix

Skip `_calculate_set_once_properties()` for `$groupidentify` events, since person properties don't apply to group events.

## Why it's safe

The server already strips `$set_once` from `$groupidentify` events ([plugin-server/src/utils/event.ts:180-189](https://github.com/PostHog/posthog/blob/master/plugin-server/src/utils/event.ts#L180-L189)):
```typescript
if (!processPerson || event.event === '$groupidentify') {
    delete event.$set
    delete event.$set_once
}
```

The SDK was sending person properties that the server discards anyway. This fix prevents the SDK from incorrectly marking them as "sent", ensuring they're included with the actual `$identify` event where they belong.

## Impact

Affects customers using `person_profiles: 'identified_only'` who call `group()` before `identify()`  a common B2B SaaS pattern where org context is set before user identification.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1141,11 +1141,11 @@ export class PostHog {
         if (setProperties) {
             data.$set = options?.$set
         }
-        // $groupidentify events don't process person $set_once on the server, so skip calculating
-        // set_once properties to avoid setting _personProcessingSetOncePropertiesSent prematurely.
-        // This ensures initial person props (UTMs, referrer) are included with subsequent $identify.
-        const setOnceProperties =
-            event_name === '$groupidentify' ? undefined : this._calculate_set_once_properties(options?.$set_once)
+        // $groupidentify doesn't process person $set_once on the server, so don't mark
+        // initial person props as sent. This ensures they're included with subsequent
+        // $identify calls.
+        const markSetOnceAsSent = event_name !== '$groupidentify'
+        const setOnceProperties = this._calculate_set_once_properties(options?.$set_once, markSetOnceAsSent)
         if (setOnceProperties) {
             data.$set_once = setOnceProperties
         }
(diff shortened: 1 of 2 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -1144,7 +1141,11 @@ export class PostHog {
         if (setProperties) {
             data.$set = options?.$set
         }
-        const setOnceProperties = this._calculate_set_once_properties(options?.$set_once)
+        // $groupidentify events don't process person $set_once on the server, so skip calculating
```
robbie-c: Do you know why they aren't processing on the server? I read the code you linked https://github.com/PostHog/posthog/blob/master/plugin-server/src/utils/event.ts#L180-L189 but that didn't really help me understand why 🙃 

It might be that we shouldn't require person processing just because someone has called `group()` in which case we should remove 

```
        if (!this._requirePersonProcessing('posthog.group')) {
            return
        }
```

from `group()` in `posthog-core.ts`

ordehi: Thank you! 

Basically, the server deletes `$set`, `$set_once`, `$unset` from these group events. As the comment says, likely a cautionary measure since groups are separate from persons to skip updating persons when not necessary (seen this pattern in person processing, we skip updating persons for specific props unless there's an explicit update for those props).

So my take is that we should stop setting the flag, but still send the props to the server, if in the future someone decides not to delete the props, they'll still be able to use them, and we fix the SDK bug that's skipping the processing when the user expects it.

1. Removed `_requirePersonProcessing` as you suggested
2. For `$groupidentify` events, we still include initial person props but don't mark them as "sent" so they'll be included with subsequent `$identify` calls

PR #2733, opened 2025-12-11
feat(surveys): add hideCancelButton to sdk (not in backend yet)

adboio:
## Problem

adds support for new `hideCancelButton` to survey appearance. not yet available in the main posthog app, but used internally for product tours

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys/components/ConfirmationMessage.tsx
+++ b/packages/browser/src/extensions/surveys/components/ConfirmationMessage.tsx
@@ -42,9 +42,11 @@ export function ConfirmationMessage({
         }
     }, [onClose])
 
+    const renderCancelButton = isPopup && appearance.hideCancelButton !== true
+
     return (
         <div className="thank-you-message" role="status" tabIndex={0} aria-atomic="true">
-            {isPopup && <Cancel onClick={() => onClose()} />}
+            {renderCancelButton && <Cancel onClick={() => onClose()} />}
             <h3 className="thank-you-message-header">{header}</h3>
             {description &&
                 renderChildrenAsTextOrHtml({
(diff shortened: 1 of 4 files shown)


packages/browser/src/extensions/surveys/components/ConfirmationMessage.tsx
```diff
@@ -44,7 +44,7 @@ export function ConfirmationMessage({
 
     return (
         <div className="thank-you-message" role="status" tabIndex={0} aria-atomic="true">
-            {isPopup && <Cancel onClick={() => onClose()} />}
+            {isPopup && <Cancel onClick={() => onClose()} appearance={appearance} />}
```
lucasheriques: actually maybe of passing a prop, we can just check if appearance.hideCancelButton here and not render it

maybe doing something like

`const renderCancelButton = isPopup && !appearance.hideCancelButton`

dont think it's worth passing the entire obj just for a conditional rendering check

lucasheriques (approved):
minor comment

packages/browser/src/extensions/surveys/components/ConfirmationMessage.tsx
adboio: ya that is fair, will do

PR #2745, opened 2025-12-12
fix: lazy-load more of product tours

adboio:
## Problem

there's product tours code in the main bundle that has no business being there

## Changes

puts that code in its place

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes


pauldambra (approved):
looks good...

i'll hassle here the same way that I did in the conversations PR

it would be cool if product tours was in

`/extensions/product-tours`

and then the dependencies its lazy loading calls into (that should never be imported outside of the lazy context)

go in 

`/extensions/product-tours/external`

and then we can lint that they aren't being imported elsewhere (which stops someone accidentally ballooning in future (not a blocker on this obvs)

PR #2751, opened 2025-12-15
refactor(nextjs-config): lazy-load webpack plugin for Turbopack builds

Arbarwings:
## Problem

When using `@posthog/nextjs-config` with Turbopack in environments like Bun, the package fails with module resolution errors because it unconditionally imports from `@posthog/webpack-plugin`, which in turn imports from `webpack`. This happens even though Turbopack builds don't need the webpack plugin at all, they use the Next.js compiler hook with the PostHog CLI directly.

## Changes

Avoid loading the webpack module when using Turbopack by:
- Adding a `/config` subpath export to `@posthog/webpack-plugin` that doesn't depend on webpack
- Importing types and `resolveConfig` from the new `/config` subpath
- Dynamically importing `PosthogWebpackPlugin` only when needed (non-Turbopack builds)

This prevents module resolution errors in environments where webpack is not available, such as when using Bun with Turbopack.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [x] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/flat-worlds-act.md
+++ b/.changeset/flat-worlds-act.md
@@ -1,0 +1,6 @@
+---
+'@posthog/nextjs-config': patch
+'@posthog/webpack-plugin': patch
+---
+
+Lazy-load webpack plugin for Turbopack builds to avoid requiring webpack when not needed.


hpouillot:
Thanks for your contribution ! Could you expand on how to reproduce the bug you are trying to fix ? If you install next package, you will get webpack as a dependency even if you use bun and turbopack

There was an issue with webpack not being listed in the peer deps which has been fixed here: https://github.com/PostHog/posthog-js/pull/2753

PR #2757, opened 2025-12-15
feat: Attempt to use environments for posthog-js deployment

rafaeelaudibert:
We wanna move our `posthog-js` - and all other SDKs eventually - to environments to make sure a maintainer is always approving the workflow. We'll monitor those from Slack.

This is an initial attempt to get this to work, let's see how this goes!

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -58,18 +58,18 @@ jobs:
             committed: ${{ steps.commit-version-bump.outputs.committed }}
         steps:
             - name: Get GitHub App token
-              id: deployer
+              id: releaser
               uses: getsentry/action-github-app-token@97c9e23528286821f97fba885c1b1123284b29cc # v2
               with:
-                  app_id: ${{ secrets.GH_APP_POSTHOG_DEPLOYER_APP_ID }}
-                  private_key: ${{ secrets.GH_APP_POSTHOG_DEPLOYER_PRIVATE_KEY }}
+                  app_id: ${{ secrets.GH_APP_POSTHOG_JS_RELEASER_APP_ID }}
+                  private_key: ${{ secrets.GH_APP_POSTHOG_JS_RELEASER_PRIVATE_KEY }} # Secrets available only inside the `NPM Release` environment, requires approval from a maintainer
 
             - name: Checkout repository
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -1,160 +1,251 @@
...
+    contents: write
+    actions: write
+    id-token: write
 
 on:
-  push:
-    branches:
-      - main
+    pull_request:
+        types: [closed]
```
Piccirello, line 10: Is this the right event? I'm assuming this means that the PR is just closed and not necessarily merged.

.github/workflows/release.yml
```diff
@@ -1,160 +1,251 @@
...
-            }
+    check-changesets:
+        name: Check for changesets
+        runs-on: ubuntu-latest
+        # Run when PR with 'release' label is merged to main, or when manually triggered
+        if: |
+            github.event_name == 'workflow_dispatch' ||
+            (github.event.pull_request.merged == true
+             && contains(github.event.pull_request.labels.*.name, 'release')
+             && github.event.pull_request.base.ref == 'main')
```
Piccirello: I'm pretty sure you can specify all of these requirements directly in `on` above.

rafaeelaudibert: I based myself off of this: https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#running-your-pull_request-workflow-when-a-pull-request-merges  
  
It seems to imply we need to check for merged here

rafaeelaudibert: The base can be moved over to the global setting :white_check_mark:

rafaeelaudibert: I can't find a way to make it work for labels, so I believe this is the way to do it

Piccirello: I thought there was a `merged` type, but you're right there's just closed and then we have to explicitly check if it was merged. And definitely right about labels, oversight on my part.

https://docs.github.com/en/webhooks/webhook-events-and-payloads

.github/workflows/release.yml
Piccirello: This is the right event. https://docs.github.com/en/webhooks/webhook-events-and-payloads

PR #2779, opened 2025-12-16
refactor: Notify Slack users via Workflow

rafaeelaudibert:
Let's notify our team on Slack so that they know there's a workflow that needs to be approved. We'll soon move this to the global `.github` repo, but for now, let's keep it here while we're testing it

### DIFF
--- a/.github/workflows/notify-slack-release-approval.yml
+++ b/.github/workflows/notify-slack-release-approval.yml
@@ -1,4 +1,6 @@
 name: 'Notify Slack - Release Approval Needed'
+permissions:
+    contents: read
 
 on:
     workflow_dispatch:


.github/workflows/notify-slack-release-approval.yml
```diff
@@ -0,0 +1,141 @@
...
+                                  "text": "Approve/Reject on GitHub",
+                                  "emoji": true
+                                },
+                                "url": "https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}",
+                                "style": "primary"
+                              }
+                            ]
+                          }
+                        ]
+                      }
```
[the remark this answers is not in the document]
rafaeelaudibert: Fixed

.github/workflows/release.yml
```diff
@@ -48,12 +48,18 @@ jobs:
...
+    notify-approval-needed:
+        name: Notify Slack - Approval Needed
+        needs: check-changesets
+        if: needs.check-changesets.outputs.has-changesets == 'true'
+        uses: ./.github/workflows/notify-slack-release-approval.yml
+
     version-bump:
         name: Bump versions and commit to main
-        needs: check-changesets
+        needs: [check-changesets, notify-approval-needed]
```
Piccirello, line 59: This job won't run if the notification fails to send, which probably isn't desirable

.github/workflows/notify-slack-release-approval.yml
```diff
@@ -0,0 +1,143 @@
...
+                  if [ -n "$SLACK_USER_GROUP_ID" ]; then
+                    echo "group_mention=<!subteam^$SLACK_USER_GROUP_ID|@group-client-libraries>" >> "$GITHUB_OUTPUT"
+                  else
+                    echo "group_mention=@group-client-libraries" >> "$GITHUB_OUTPUT"
+                  fi
+
+            - name: Send Slack notification
+              uses: slackapi/slack-github-action@b6f4640825302dc9b85bd5ffbe34dfc7a762e404 # v2.1.1
+              with:
+                  webhook: ${{ inputs.slack_webhook_url || vars.APPROVALS_CLIENT_LIBRARIES_SLACK_WEBHOOK_URL }}
```
Piccirello, line 69: Slack webhooks are sensitive since they allow sending arbitrary data with an arbitrary identity to a channel. You'll definitely want to move this to a secret. You'll also want to revoke the existing webhook since it'll now have been exposed in the Action logs.

.github/workflows/notify-slack-release-approval.yml
```diff
@@ -0,0 +1,143 @@
...
+                required: false
+                type: string
+                default: ''
+            slack_user_group_id:
+                description: 'Slack User Group ID for mentions (overrides vars.GROUP_CLIENT_LIBRARIES_SLACK_GROUP_ID)'
+                required: false
+                type: string
+                default: ''
+
+jobs:
```
Piccirello, line 20: Can you add explicit permissions to this job? It should just need `contents: read`

Piccirello (commented):
Does this repo not require PRs to be approved before merge? I think we should change that, this repo is too important.

PR #2782, opened 2025-12-16
feat(flags): Add `$feature_flag_error` property to track flag evaluation failures

haacked:
This is a port of https://github.com/PostHog/posthog-python/pull/390

## Problem

When feature flag evaluations fail (network errors, missing flags, server-side computation errors, quota limits), there's no standardized way to track and analyze these failures. This makes it difficult to debug issues and understand feature flag reliability in production.

The Python SDK already implements a `$feature_flag_error` property for this purpose. Node.js SDK users need the same observability capability.

## Changes

Adds a `$feature_flag_error` property to feature flag events in the Node.js SDK, matching the Python SDK's implementation. This property is only set when errors occur during flag evaluation.

**Error types tracked:**
- `errors_while_computing_flags` - Server returned errors while computing flags
- `flag_missing` - Requested flag not in API response  
- `quota_limited` - Feature flags billing quota exceeded
- `unknown_error` - Request failed completely (network error, timeout, etc.)

Multiple errors can occur simultaneously and are comma-joined (e.g., `"quota_limited,flag_missing"`).

**Implementation details:**
- Added `FeatureFlagError` constant object with typed error values
- Modified `getFeatureFlag()` to detect errors and set `$feature_flag_error` property
- Updated core library to preserve `quotaLimited` field in responses (was previously stripped)
- Property is only added when errors exist (no impact on success path)

> [!NOTE]
> The `FeatureFlagError` constants include `TIMEOUT`, `CONNECTION_ERROR`, and `apiError(status)` for future use, but these are not yet implemented. Specifically, HTTP 429 rate-limiting errors currently fall into `UNKNOWN_ERROR` because the core library catches HTTP errors and returns `undefined`, losing the status code. 
>
> To properly distinguish 429 errors from other failures would require modifying the core library's error handling to preserve HTTP status information, which affects the contract between core and all SDK consumers (Node, browser, React, etc.). This is deferred to a future PR to keep this change focused on parity with the Python SDK.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/types.ts
+++ b/packages/node/src/types.ts
@@ -179,31 +179,16 @@ export type PostHogFeatureFlag = {
  *   ERRORS_WHILE_COMPUTING: Server returned errorsWhileComputingFlags=true
  *   FLAG_MISSING: Requested flag not in API response
  *   QUOTA_LIMITED: Rate/quota limit exceeded
- *   TIMEOUT: Request timed out
- *   CONNECTION_ERROR: Network connectivity issue
  *   UNKNOWN_ERROR: Unexpected exceptions
- *
- * For API errors with status codes, use the apiError() method which returns
- * a string like "api_error_500".
  */
 export const FeatureFlagError = {
   ERRORS_WHILE_COMPUTING: 'errors_while_computing_flags',
   FLAG_MISSING: 'flag_missing',
   QUOTA_LIMITED: 'quota_limited',
-  TIMEOUT: 'timeout',
-  CONNECTION_ERROR: 'connection_error',
   UNKNOWN_ERROR: 'unknown_error',
-
-  /**
-   * Generate API error string with status code.
-   *
-   * @param status - HTTP status code from the API error
-   * @returns Error string like "api_error_500"
-   */
-  apiError: (status: number | string): string => `api_error_${status}`,
 } as const
 
-export type FeatureFlagErrorType = (typeof FeatureFlagError)[keyof Omit<typeof FeatureFlagError, 'apiError'>] | string
+export type FeatureFlagErrorType = (typeof FeatureFlagError)[keyof typeof FeatureFlagError] | string
 
 export interface IPostHog {
   /**


packages/node/src/types.ts
```diff
@@ -168,6 +168,43 @@ export type PostHogFeatureFlag = {
...
+  CONNECTION_ERROR: 'connection_error',
+  UNKNOWN_ERROR: 'unknown_error',
+
+  /**
+   * Generate API error string with status code.
+   *
+   * @param status - HTTP status code from the API error
+   * @returns Error string like "api_error_500"
+   */
+  apiError: (status: number | string): string => `api_error_${status}`,
```
dustinbyrne: nit - considering this also accepts a string, might be worth `toLowerCase()` if a future input is a non-numeric string?
```suggestion
  apiError: (status: number | string): string => `api_error_${status.toLowerCase()}`,
```

dustinbyrne (approved):
👍 👍

packages/node/src/types.ts
```diff
@@ -168,6 +168,43 @@ export type PostHogFeatureFlag = {
...
+ *
+ * For API errors with status codes, use the apiError() method which returns
+ * a string like "api_error_500".
+ */
+export const FeatureFlagError = {
+  ERRORS_WHILE_COMPUTING: 'errors_while_computing_flags',
+  FLAG_MISSING: 'flag_missing',
+  QUOTA_LIMITED: 'quota_limited',
+  TIMEOUT: 'timeout',
+  CONNECTION_ERROR: 'connection_error',
```
dustinbyrne: Also worth noting, these (+ `apiError`) aren't used

PR #2784, opened 2025-12-17
feat(logs): Add logs extension to posthog-js

frankh:
## Problem

People expect the logs product to "work" when they enable it and have posthog installed. We currently require a separate otel collector to be configured - lets wrap otel collector and add it to posthog-js so it works out of the box as expected

## Changes

add new item logCapture to remote config, if this is true, load the logs extension which is a thin wrapper around console.log with the otel collector

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-logs.ts
+++ b/packages/browser/src/posthog-logs.ts
@@ -9,7 +9,7 @@ export class PostHogLogs {
     private _isLoaded: boolean = false
 
     constructor(private readonly _instance: PostHog) {
-        if (this._instance.config.logs?.captureConsoleLogs) {
+        if (this._instance && this._instance.config.logs?.captureConsoleLogs) {
             this._isLogsEnabled = true
         }
     }


packages/browser/src/entrypoints/logs.ts
```diff
@@ -0,0 +1,123 @@
...
+                    logger.emit({
+                        severityText: {
+                            log: 'INFO',
+                            warn: 'WARNING',
+                            error: 'ERROR',
+                            debug: 'DEBUG',
+                            info: 'INFO',
+                        }[level],
+                        body: args.map((a) => JSON.stringify(a, errorReplacer)).join(' '),
+                        attributes: {
```
pauldambra, line 122: i don't know how much magic otel does for us

when i had a run at adding redux state logging i found it was pretty easy to be noticeably slow while processing things

it'd be good to have some tests to prove that processing very big or very complex log messages doesn't take too long

packages/browser/src/utils/globals.ts
```diff
@@ -212,6 +213,7 @@ interface PostHogExtensions {
     rrwebPlugins?: { getRecordConsolePlugin: any; getRecordNetworkPlugin?: any }
     generateSurveys?: (posthog: PostHog, isSurveysEnabled: boolean) => any | undefined
     generateProductTours?: (posthog: PostHog, isEnabled: boolean) => any | undefined
+    initializeLogs?: (posthog: PostHog) => any | undefined
```
pauldambra: i always end up regretting the shape of these hard to change things

might be nicer in 12 months time if we start with

```
logs: {
   initializeLogs?: (posthog: PostHog) => any | undefined
}
```

so that as the number of logs things grows it doesn't get messy

(ofc then in 12 months time when it's still just one thing, someone can say "which fool insisted on a nested property" but i really wish the rrweb ones were nested)

packages/browser/src/posthog-logs.ts
```diff
@@ -0,0 +1,51 @@
...
+
+    constructor(private readonly _instance: PostHog) {}
+
+    onRemoteConfig(response: RemoteConfig) {
+        // only load logs if they are enabled
+        const logCapture = response.logs?.captureConsoleLogs
+        if (isNullish(logCapture) || !logCapture) {
+            return
+        }
+        this._hasLogs = true
```
pauldambra: does `hasLogs` mean `isEnabled`?

frankh: yes, i'm not sure how i ended up with this naming

I thought i copied from surveys but that doesn't use hasSurveys so 🤷 - will rename it

packages/browser/src/__tests__/extensions/logs.test.ts
```diff
@@ -0,0 +1,599 @@
...
+
+            const initializeLogs = assignableWindow.__PosthogExtensions__.initializeLogs
+            initializeLogs(mockPostHog)
+
+            // Console methods should be wrapped (different from originals)
+            expect(assignableWindow.console.log).not.toBe(originalMethods.log)
+            expect(assignableWindow.console.info).not.toBe(originalMethods.info)
+            expect(assignableWindow.console.warn).not.toBe(originalMethods.warn)
+            expect(assignableWindow.console.error).not.toBe(originalMethods.error)
+            expect(assignableWindow.console.debug).not.toBe(originalMethods.debug)
```
pauldambra, line 176: this all looks great

it's probably worth asking the robot to copy some of the session recording or deadclicks playwright tests so we get some validation this works in the running code too

packages/browser/src/utils/globals.ts
frankh: future proofing ✅

packages/browser/src/entrypoints/logs.ts
frankh: good point, i also meant to add truncation so we don't log multi megabyte objects

rafaeelaudibert:
oh no (but it's understandable, this is all dependencies)

packages/browser/package.json
```diff
@@ -57,6 +57,11 @@
         "web-vitals": "^4.2.4"
     },
     "devDependencies": {
+        "@opentelemetry/api": "^1.9.0",
+        "@opentelemetry/api-logs": "^0.208.0",
+        "@opentelemetry/exporter-logs-otlp-http": "^0.208.0",
+        "@opentelemetry/resources": "^2.2.0",
+        "@opentelemetry/sdk-logs": "^0.208.0",
```
rafaeelaudibert: Why dev dependency? Isn't this an actual dependency?

packages/browser/src/entrypoints/logs.ts
```diff
@@ -0,0 +1,123 @@
...
+            (...args: any[]) => {
+                if (args.length > 0) {
+                    logger.emit({
+                        severityText: {
+                            log: 'INFO',
+                            warn: 'WARNING',
+                            error: 'ERROR',
+                            debug: 'DEBUG',
+                            info: 'INFO',
+                        }[level],
```
rafaeelaudibert: This object can surely be a constant that's live indexed to avoid instantiating the object every time we log

packages/browser/src/entrypoints/logs.ts
```diff
@@ -0,0 +1,123 @@
...
+                            ...attributes,
+                            ...(isObject(args[0]) ? flattenObject(args[0]) : {}),
+                        },
+                    })
+                    originalConsoleLog.apply(assignableWindow.console, args)
+                }
+            }
+
+        let originalConsoleLog = assignableWindow.console[level]
+        if ('__rrweb_original__' in assignableWindow.console[level]) {
```
rafaeelaudibert, line 137: I'm curious, what's this needed for?

packages/browser/package.json
frankh: so's rrweb right? which is under dev dependencies - I wasn't really sure but i guessed we chose to put everything in dev dependencies for some reason

packages/browser/src/entrypoints/logs.ts
frankh: if we have session replay enabled the original console log gets copied to `__rrweb_original__`

I decided to wrap this too in case some things use it (i'm not sure if anything does)

packages/browser/package.json
rafaeelaudibert: That looks like a bug too tbh. This probably works because we bundle the library for the browser and they're put together.  
  
I'd double check with Claude here, but I believe this is wrong and you should move it to `dependencies` and then a follow-up can fix the `rrweb` stuff (and some others, I see `posthog-js` here as well)

packages/browser/src/posthog-logs.ts
```diff
@@ -0,0 +1,51 @@
+import { PostHog } from './posthog-core'
+import { RemoteConfig } from './types'
+import { isNullish } from '@posthog/core'
+import { assignableWindow } from './utils/globals'
+import { createLogger } from './utils/logger'
+
+export class PostHogLogs {
+    private _isLogsEnabled?: boolean
```
rafaeelaudibert: nit: I know we don't do this on other extensions, but this could probably be assigned false by default

packages/browser/src/entrypoints/logs.ts
rafaeelaudibert: In that case, I suggest we add a comment explaining the reasoning, makes sense to me though!

packages/browser/src/types.ts
```diff
@@ -252,6 +256,11 @@ export interface RemoteConfig {
      */
     productTours?: boolean
 
+    /**
+     * Whether capturing logs is enabled
+     */
+    logCapture?: boolean
```
rafaeelaudibert: Why 2 configs? Also, is this only controllable via remote config? Why don't we have a local setting?

frankh: good catch, i renamed the setting at some point and didn't clean this up

will add as a local setting too

packages/types/src/posthog-config.ts
```diff
@@ -538,6 +538,13 @@ export interface SurveyConfig {
     autoSubmitDelay?: number
 }
 
+/**
+ * Logs configuration options
+ */
+export interface LogsConfig {
+    captureConsoleLogs?: boolean
```
rafaeelaudibert, line 545: Maybe adding a explanation of what this does, even a link to the logs docs would be good, people might learn about the product via the types :)

PR #2787, opened 2025-12-17
feat: allow customizing text color on surveys

lucasheriques:
## Problem

Survey appearance customization needs improvement to ensure consistent text contrast across different elements, particularly for rating buttons and text inputs.

## Changes

- Added support for consistent text color handling in surveys:
  - Rating button text now uses the same color as input text
  - Primary text color can be explicitly set or auto-calculated based on background
  - Active rating text color is always auto-calculated for proper contrast

Implemented the changes for both React Native and Web (posthog-js) and now began sharing more fields between them too.

### Libraries affected

- [x] posthog-js (web)
- [x] posthog-react-native

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/extensions/surveys-utils.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys-utils.test.ts
@@ -7,7 +7,7 @@ import {
     hasEvents,
     hasWaitPeriodPassed,
 } from '../../extensions/surveys/surveys-extension-utils'
-import { Survey, SurveyAppearance, SurveySchedule, SurveyType } from '../../posthog-surveys-types'
+import { Survey, SurveySchedule, SurveyType } from '../../posthog-surveys-types'
 import { SURVEY_IN_PROGRESS_PREFIX, SURVEY_SEEN_PREFIX } from '../../utils/survey-utils'
 
 describe('hasWaitPeriodPassed', () => {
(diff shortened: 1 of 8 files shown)


lucasheriques:
* **#2787** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2787?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2787?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/__tests__/extensions/surveys-utils.test.ts
lucasheriques, line 1: there are many tests here and some of them repeated, so i'm consolidating them a little

packages/core/src/types.ts
```diff
@@ -294,13 +294,18 @@ export type EvaluationReason = {
 export type SurveyAppearance = {
   // keep in sync with frontend/src/types.ts -> SurveyAppearance
   backgroundColor?: string
+  // Optional override for main survey text color. If not set, auto-calculated from backgroundColor.
+  textColor?: string
```
adboio, line 298: this will now exist on the rn sdk types, so we should still document somewhere that it doesn't work on react-native (until we update it)

adboio (approved):

packages/react-native/src/surveys/components/QuestionTypes.tsx
```diff
@@ -54,14 +54,14 @@ export function OpenTextQuestion({
...
-              color: getContrastingTextColor(appearance.inputBackground),
+              color: appearance.inputTextColor ?? getContrastingTextColor(appearance.inputBackground),
             },
           ]}
           multiline
           numberOfLines={4}
           placeholder={appearance.placeholder}
           placeholderTextColor={
-            getContrastingTextColor(appearance.inputBackground) === 'black'
+            (appearance.inputTextColor ?? getContrastingTextColor(appearance.inputBackground)) === 'black'
```
ioannisj: This would be white if inputTextColor is anything other than black right? Doesn't seem right

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -120,10 +120,14 @@ export const defaultBackgroundColor = '#eeeded' as const
...
+// textColor and inputTextColor are optional overrides (auto-calculated if not provided)
 export type SurveyAppearanceTheme = Omit<
   Required<SurveyAppearance>,
-  'widgetSelector' | 'widgetType' | 'widgetColor' | 'widgetLabel' | 'shuffleQuestions'
->
+  'widgetSelector' | 'widgetType' | 'widgetColor' | 'widgetLabel' | 'shuffleQuestions' | 'textColor' | 'inputTextColor'
+> & {
+  textColor?: string
+  inputTextColor?: string
+}
```
ioannisj, line 130: Don't really know how to read this type. Are we omitting `textColor`, `inputTextColor ` from original type and adding them back as optional properties?

marandaneto: its omitting all hardcoded values and adding back the 2 as optional

packages/core/src/types.ts
lucasheriques: the implementation is a part of the PR, not sure if you saw it before the update, can you double check?

packages/react-native/src/surveys/components/QuestionTypes.tsx
lucasheriques: | Case                                                           | Result                     |
  |----------------------------------------------------------------|----------------------------|
  | No inputTextColor, light bg → 'black' → needs 'white' contrast | dark placeholder ✓         |
  | No inputTextColor, dark bg → 'white' → needs 'black' contrast  | light placeholder ✓        |
  | inputTextColor = 'black' → needs 'white' contrast              | dark placeholder ✓         |
  | inputTextColor = '#000000' → needs 'white' contrast            | dark placeholder ✓ (FIXED) |

fixed it now, see if the table above makes sense too

packages/react-native/src/surveys/surveys-utils.ts
lucasheriques: > Don't really know how to read this type. Are we omitting textColor, inputTextColor  from original type and adding them back as optional properties?

yep! we're making most fields required (so we always have defaults to render with), but textColor and inputTextColor stay optional since they're overrides - if not set, we auto-calculate them from the background for contrast.

adboio (approved):

lucasheriques:
### Merge activity

* **Dec 18, 9:57 PM UTC**: @lucasheriques merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/2787).

PR #2789, opened 2025-12-18
fix(llma): extract model from response for OpenAI stored prompts

andrewm4894:
## Problem

When using OpenAI stored prompts, the model is defined in the OpenAI dashboard rather than passed in the API request. The PostHog AI wrapper only extracts the model from request parameters, causing generations to show up without a model and preventing cost calculations.

Fixes https://github.com/PostHog/posthog/issues/42861

## Changes

- Updated `WrappedResponses` and `WrappedCompletions` classes to extract model from response when not in params
- Added model extraction from streaming chunks for both APIs
- Applied same fix to Azure OpenAI wrapper

## Libraries affected

- [x] @posthog/ai

## Checklist

- [ ] Tests for new code

### DIFF
--- a/packages/ai/src/openai/azure.ts
+++ b/packages/ai/src/openai/azure.ts
@@ -405,7 +405,7 @@ export class WrappedResponses extends AzureOpenAI.Responses {
               await sendEventToPosthog({
                 client: this.phClient,
                 ...posthogParams,
-                model: openAIParams.model ?? modelFromResponse ?? '',
+                model: openAIParams.model ?? modelFromResponse,
                 provider: 'azure',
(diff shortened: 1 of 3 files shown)


packages/ai/src/openai/azure.ts
```diff
@@ -405,10 +424,9 @@ export class WrappedResponses extends AzureOpenAI.Responses {
               await sendEventToPosthog({
                 client: this.phClient,
                 ...posthogParams,
-                //@ts-expect-error
-                model: openAIParams.model,
+                model: openAIParams.model ?? '',
```
Radu-Raicea: Could we not send an empty string instead? Didn't check if it makes sense though.

PR #2792, opened 2025-12-18
feat(product tours): enable event and action triggers

adboio:
## Problem

product tours should be trigger-able through events or actions

## Changes

make them!

- also add preview method for use in the toolbar
- also re-enable playwright tests for survey event triggers... i _think_ i fixed the reason they were previously flakey

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/browser/src/utils/event-receiver.ts
adboio, line 1: this is basically copied from survey-event-receiver

adboio: but ... this is a little risky, so if you'd rather keep surveys as-is and only use this for product tours (for now) -- happy to do that as well

i tested it manually, but our automated testing is a bit lacking here  
  
edit: our tests are actually better than i thought lol. i think this is ok... here is also a diff of the original survey rx file with the new abstract one https://www.diffchecker.com/ZdVJNFsc/

packages/browser/playwright/mocked/surveys/event-property-filtering.spec.ts
adboio, line 1: this is before my time, but pretty sure this was flakey because we didn't wait for the api response... i had the same issue with some other tests i added recently. i'd like to keep these enabled

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -702,10 +775,40 @@ export class ProductTourManager {
...
+        logger.info(`Queueing tour ${tourId} with ${delaySeconds}s delay`)
+
+        const timeoutId = setTimeout(() => {
+            this._pendingTourTimeouts.delete(tourId)
+            logger.info(`Delay elapsed for tour ${tourId}, showing now`)
+            this.showTourById(tourId, reason)
+        }, delaySeconds * 1000)
+
+        this._pendingTourTimeouts.set(tourId, timeoutId)
+    }
```
lucasheriques, line 813: nit: might be worth adding `this.cancelPendingTour(tourId)` at the start of `queueTourWithDelay` as a defensive measure. i know `_showOrQueueTour` already checks `isTourPending` before calling this, but since `queueTourWithDelay` is public, someone could call it directly and accidentally overwrite a timeout without cancelling
the previous one

packages/browser/src/utils/product-tour-event-receiver.ts
```diff
@@ -0,0 +1,46 @@
...
+import { ProductTour } from '../posthog-product-tours-types'
+import { PostHog } from '../posthog-core'
+import { EventReceiver } from './event-receiver'
+import { createLogger } from './logger'
+import { localStore } from '../storage'
+import { TOUR_COMPLETED_KEY_PREFIX, TOUR_DISMISSED_KEY_PREFIX } from '../extensions/product-tours/constants'
+
+const logger = createLogger('[Product Tour Event Receiver]')
+
+export class ProductTourEventReceiver extends EventReceiver<ProductTour> {
```
lucasheriques, line 11: would be good to have some unit tests for ProductTourEventReceiver specifically - especially around _isItemPermanentlyIneligible since it behaves differently than the survey version (checking localStorage for completed/dismissed state)

lucasheriques (approved):

PR #2794, opened 2025-12-18
chore(ci): Replace PAT wth default github token

Piccirello:
The PAT isn't needed for these actions.


.github/workflows/label-alpha-release.yml
```diff
@@ -34,12 +34,14 @@ jobs:
     needs: check-permissions
```
hpouillot, line 34: this workflow is not used anymore, I think we can delete it

PR #2796, opened 2025-12-19
refactor: Use actions from `posthog/.github` repo

rafaeelaudibert:
We'll share this with other repos, so let's move it over to the global [`posthog/.github`](https://github.com/posthog/.github) repo


.github/workflows/release.yml
```diff
@@ -63,7 +63,7 @@ jobs:
 
             - name: Send Slack notification
               id: notify
-              uses: ./.github/actions/notify-slack-release-approval
+              uses: posthog/.github/.github/actions/notify-slack-release-approval@9b04bf3288aca2b4cd3883070858b034b4f7f334
```
Piccirello, line 67: I'm ok omitting the version from this and always using latest

Piccirello (approved):

PR #2797, opened 2025-12-20
feat(product tours): allow customization

adboio:
## Problem

- product tours can't be customized
- animations were a bit janky

## Changes

add customization, clean up components so they're less janky and animations are smooth, expose some methods to preview cards for main app

[pt-customization.mp4 <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/499bd6f4-41f0-4e97-8af1-c4dc643f4d6c.mp4)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/product-tours/components/ProductTourSurveyStepInner.tsx
+++ b/packages/browser/src/extensions/product-tours/components/ProductTourSurveyStepInner.tsx
@@ -1,12 +1,6 @@
 import { h } from 'preact'
 import { useState, useRef, useEffect } from 'preact/hooks'
-import {
-    ProductTour,
-    ProductTourStep,
-    ProductTourDismissReason,
-    ProductTourSurveyQuestion,
-} from '../../../posthog-product-tours-types'
-import { mergeAppearance } from '../product-tours-utils'
+import { ProductTourStep, ProductTourAppearance, ProductTourSurveyQuestion } from '../../../posthog-product-tours-types'
 import { cancelSVG, IconPosthogLogo } from '../../surveys/icons'
 import {
     dissatisfiedEmoji,
@@ -16,14 +10,14 @@ import {
     verySatisfiedEmoji,
 } from '../../surveys/icons'
 
-export interface ProductTourSurveyStepProps {
-    tour: ProductTour
+export interface ProductTourSurveyStepInnerProps {
     step: ProductTourStep
+    appearance?: ProductTourAppearance
     stepIndex: number
     totalSteps: number
-    onSubmit: (response: string | number | null) => void
-    onPrevious: () => void
-    onDismiss: (reason: ProductTourDismissReason) => void
+    onPrevious?: () => void
+    onSubmit?: (response: string | number | null) => void
+    onDismiss?: () => void
 }
 
 const threeScaleEmojis = [dissatisfiedEmoji, neutralEmoji, satisfiedEmoji]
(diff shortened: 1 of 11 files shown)


packages/browser/src/extensions/product-tours/product-tours-utils.ts
```diff
@@ -118,9 +119,6 @@ export function calculateTooltipPosition(targetRect: DOMRect): PositionResult {
         left = targetRect.left + targetRect.width / 2 - TOOLTIP_WIDTH / 2
     }
 
-    top = Math.max(TOOLTIP_MARGIN, Math.min(top, viewportHeight - TOOLTIP_HEIGHT_ESTIMATE - TOOLTIP_MARGIN))
-    left = Math.max(TOOLTIP_MARGIN, Math.min(left, viewportWidth - TOOLTIP_WIDTH - TOOLTIP_MARGIN))
```
lucasheriques, line 122: is this intentional? can it cause tooltips to render offscreen?

packages/browser/src/extensions/product-tours/components/ProductTourSurveyStepInner.tsx
```diff
@@ -156,99 +163,89 @@ function RatingInput({
...
+}: ProductTourSurveyStepInnerProps): h.JSX.Element {
     const [textValue, setTextValue] = useState('')
-    const appearance = mergeAppearance(tour.appearance)
     const survey = step.survey
+    const whiteLabel = appearance?.whiteLabel ?? false
     const isFirstStep = stepIndex === 0
     const isOpenText = survey?.type === 'open'
 
+    const isInteractive = !!(onPrevious || onSubmit || onDismiss)
+    const cursorStyle = isInteractive ? undefined : { cursor: 'default' }
```
lucasheriques, line 182: this code is repeated, should it be on a shared utility or no need?

lucasheriques (approved):
minor comments, LGTM if no input there :+1:

adboio:
### Merge activity

* **Dec 23, 4:13 AM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/2797).

packages/browser/src/extensions/product-tours/components/ProductTourSurveyStepInner.tsx
adboio: ah yeah totally could be - had this on automerge but i’ll clean up soon!

packages/browser/src/extensions/product-tours/product-tours-utils.ts
adboio: yeah i removed the clamping here so that the tooltips are stuck with the element, which i think is ok because we always scroll to the target before displaying. otherwise, scrolling on the page clamps the element to the viewport, so like if i scroll down a tooltip will just float at the top. this would be cool IF we had dynamic arrows like a video game quest compass haha, but i think removing clamping is better for now, til that exists

Issue #2781, opened 2025-12-16
allow changing where script tags are appended to


paperclover:
## Is your feature request related to a problem?

This code for appending script tags is too smart:
https://github.com/PostHog/posthog-js/blob/b89d0f4afbc0f5fc6d0bbc02b2fec7678a0dbf3d/packages/browser/src/entrypoints/external-scripts-loader.ts#L65-L71

It causes a hydration error because it interweaves with [this component in tanstack router](https://github.com/TanStack/router/blob/8628d0189412ccb8d3a01840aa18bac8295e18c8/packages/react-router/src/Match.tsx#L163).

## Describe the solution you'd like

- allow a configuration to always `document.body.appendChild(scriptTag)`
- allow a configuration to set the target element (so i can set it to document.body)
- allow `prepare_external_dependency_script` to return a promise so the configuration can customize how the script is mounted instead of disabling it.

## Describe alternatives you've considered

current solution is patching posthog to make the script loading simpler

## Related sub-libraries

- [x] posthog-js (web)

## Additional Notes

search keywords: hydration error, tanstack start, tanstack router, loadScript

pauldambra:
should  be fixed by https://github.com/PostHog/posthog-js/pull/2802

PR #2802, opened 2025-12-21
feat: allow adding lazy loaded scripts to head to avoid hydration errors

pauldambra:
avoid hydration errors https://github.com/PostHog/posthog-js/issues/2781

when someone is using SSR and we're lazy loading scripts we have a race between react hydrating and us altering the DOM

if we manage to alter the DOM before hydration complests then we'll cause a hydration error  
  
i don't believe we care where the script is added (since we're adding it async from code the location isn't going to affect whether it blocks) so we can add the script to the head and not have any risk of causing a hydration error

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

### DIFF
--- a/packages/browser/src/entrypoints/external-scripts-loader.ts
+++ b/packages/browser/src/entrypoints/external-scripts-loader.ts
@@ -62,12 +62,15 @@ const loadScript = (posthog: PostHog, url: string, callback: (error?: string | E
             return callback('prepare_external_dependency_script returned null')
         }
 
-        const scripts = document.querySelectorAll('body > script')
-        if (scripts.length > 0) {
-            scripts[0].parentNode?.insertBefore(scriptTag, scripts[0])
+        if (posthog.config.external_scripts_inject_target === 'head') {
+            document.head.appendChild(scriptTag)
         } else {
-            // In exceptional situations this call might load before the DOM is fully ready.
-            document.body.appendChild(scriptTag)
+            const scripts = document.querySelectorAll('body > script')
+            if (scripts.length > 0) {
+                scripts[0].parentNode?.insertBefore(scriptTag, scripts[0])
+            } else {
+                document.body.appendChild(scriptTag)
+            }
         }
     }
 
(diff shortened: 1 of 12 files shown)


packages/browser/src/entrypoints/external-scripts-loader.ts
```diff
@@ -62,13 +62,7 @@ const loadScript = (posthog: PostHog, url: string, callback: (error?: string | E
             return callback('prepare_external_dependency_script returned null')
         }
 
-        const scripts = document.querySelectorAll('body > script')
```
pauldambra, line 65: @benjackwhite to avoid me doing something silly, since you added this, do you remember why we have the two different mechanisms here?

i think it's safe to just do one since scripts added like this are async (\_i think_)

benjackwhite: pretty sure this would have been copied and pasted as part of a refactor. No idea why that matters...

rafaeelaudibert (approved):
Yeah, should be the same! This is a breaking change in a way, any chance we could hide it behind a new `defaults`or do you feel positive this is good to go?

pauldambra:
i was thinking not breaking because there should be no action on users as a result

i'll wait till the new year and deploy when i an concentrate on it

i do think it'd be nice to get the SDKs in to the mono-repo so we could validate JS SDK in the app without releasing it maybe (and then keep posthog.com for validating the html snippet works) but that's not a quick thing so 🙈

nachogarcia:
As Tanstack Start doesn't have yet an official posthog integration, this will greatly simplify current setup :) thanks!

paperclover:
@pauldambra the new year has come and gone, is there anything stopping this from being merged?

PR #2803, opened 2025-12-21
fix: alter how we add lazy loaded scripts to avoid hydration errors

pauldambra:
change how we add lazy loaded scripts so that we don't cause hydration errors
(i worry this remains a race against hydration since we are still altering the DOM 🤞)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

### DIFF
--- a/.changeset/ready-knives-walk.md
+++ b/.changeset/ready-knives-walk.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: the way we added lazy loaded files could cause hydration errors (in a race against hydration). let's avoid that


packages/browser/src/entrypoints/external-scripts-loader.ts
```diff
@@ -62,13 +62,7 @@ const loadScript = (posthog: PostHog, url: string, callback: (error?: string | E
             return callback('prepare_external_dependency_script returned null')
         }
 
-        const scripts = document.querySelectorAll('body > script')
```
pauldambra: @benjackwhite to avoid me doing something silly, since you added this, do you remember why we have the two different mechanisms here?  
  
i think it's safe to just do one since scripts added like this are async (\_i think_)

PR #2805, opened 2025-12-22
fix(ci): make Slack notifications non-blocking in release workflow

lucasheriques:
## Summary

- Fix SIGPIPE (exit 141) error in `notify-slack-release-approval` action by storing `git tag` output before piping to `head`
- Add `continue-on-error: true` to all Slack notification steps so releases proceed even if Slack is down
- Track Slack failures via `posthog-js-slack-notification-failure` PostHog event for alerting

## Context

The release workflow was failing at the "Notify Slack - Approval Needed" step due to a SIGPIPE error. When `head -n 1` gets the first line and closes the pipe, `git tag` (which is slow - external process sorting 1229 tags) may still be writing, causing SIGPIPE (exit 141).

## The fix

```bash
# Before: git tag is slow, head may close pipe while git is still writing → SIGPIPE
LATEST_TAG=$(git tag --sort=-creatordate | head -n 1)

# After: git completes fully, echo is a fast builtin that finishes before head closes
ALL_TAGS=$(git tag --sort=-creatordate)
LATEST_TAG=$(echo "$ALL_TAGS" | head -n 1)
```

## Local verification

```bash
# Original approach - SIGPIPE with pipefail enabled
$ bash -c 'set -e -o pipefail; for i in {1..10}; do (seq 1 100000 | head -n 1 > /dev/null); done'
Exit: 141

# Our fix - 50 iterations, no failures
$ bash -c 'set -e -o pipefail; for i in {1..50}; do ALL=$(git tag --sort=-creatordate); LATEST=$(echo "$ALL" | head -n 1); done; echo "passed"'
LATEST_TAG: posthog-js-lite@4.2.4
All 50 iterations passed
Exit: 0
```

## Test plan

- [ ] Trigger release workflow and verify it proceeds past Slack steps
- [ ] Verify `posthog-js-slack-notification-failure` events appear in PostHog if Slack fails

### DIFF
--- a/.github/actions/notify-slack-release-approval/action.yaml
+++ b/.github/actions/notify-slack-release-approval/action.yaml
@@ -28,7 +28,9 @@ runs:
               git fetch --tags --force
 
               # Get the most recent tag (sorted by date, not version)
-              LATEST_TAG=$(git tag --sort=-creatordate | head -n 1)
+              # Note: Store all tags first to avoid SIGPIPE (exit 141) when piping to head
+              ALL_TAGS=$(git tag --sort=-creatordate)
+              LATEST_TAG=$(echo "$ALL_TAGS" | head -n 1)
 
               if [ -z "$LATEST_TAG" ]; then
                 echo "No tags found, using initial commit"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -63,11 +63,27 @@ jobs:
 
             - name: Send Slack notification
               id: notify
+              continue-on-error: true # Don't block release if Slack notification fails
               uses: ./.github/actions/notify-slack-release-approval
               with:
                   slack_bot_token: ${{ secrets.SLACK_CLIENT_LIBRARIES_BOT_TOKEN }}
                   slack_channel_id: ${{ vars.SLACK_APPROVALS_CLIENT_LIBRARIES_CHANNEL_ID }}
                   slack_user_group_id: ${{ vars.GROUP_CLIENT_LIBRARIES_SLACK_GROUP_ID }}
+
+            - name: Send Slack failure event to PostHog
+              if: ${{ steps.notify.outcome == 'failure' }}
+              uses: PostHog/posthog-github-action@v0.1
+              with:
+                  posthog-token: '${{ secrets.POSTHOG_PROJECT_API_KEY }}'
+                  event: 'sdk-slack-notification-failure'
+                  properties: >-
+                      {
+                        "step": "notify-approval-needed",
+                        "workflow": "release",
+                        "repository": "${{ github.repository }}",
+                        "run_id": "${{ github.run_id }}",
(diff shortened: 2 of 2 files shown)


.github/actions/notify-slack-release-approval/action.yaml
```diff
@@ -28,7 +28,8 @@ runs:
               git fetch --tags --force
 
               # Get the most recent tag (sorted by date, not version)
-              LATEST_TAG=$(git tag --sort=-creatordate | head -n 1)
+              # Note: || true prevents SIGPIPE (exit 141) when head closes the pipe early
```
rafaeelaudibert: Huh, I dont understand this very well. Why would it break the pipe? 🤔 

Also, doesn't that mean we'll return `true` to `LATEST_TAG`?

.github/workflows/release.yml
```diff
@@ -87,6 +105,21 @@ jobs:
                   message: '✅ Release approved! Version bump in progress...'
                   emoji_reaction: 'white_check_mark'
 
+            - name: Send Slack failure event to PostHog
+              if: ${{ steps.slack-approved.outcome == 'failure' }}
+              uses: PostHog/posthog-github-action@v0.1
+              with:
+                  posthog-token: '${{ secrets.POSTHOG_PROJECT_API_KEY }}'
+                  event: 'posthog-js-slack-notification-failure'
```
rafaeelaudibert: Might wanna make this event name more generic since we'll add it to other SDKs as well

.github/actions/notify-slack-release-approval/action.yaml
lucasheriques: that was an older version! I've since updated to a better approach:

```
ALL_TAGS=$(git tag --sort=-creatordate)
LATEST_TAG=$(echo "$ALL_TAGS" | head -n 1)
```

why the original breaks: with git tag | head -n 1 and set -o pipefail, when head gets the first line it closes the pipe. But git tag (external process, sorting 1229 tags) is slow and may still be writing → SIGPIPE → exit 141.

why the new approach works: git tag completes fully before we pipe anything. Then echo (shell builtin) is fast enough to write to the pipe buffer before head closes it.

.github/workflows/release.yml
lucasheriques: done

.github/actions/notify-slack-release-approval/action.yaml
Piccirello: I'm also not following why this would fail tbh

Piccirello: Reminder to myself: read

> why the original breaks: with git tag | head -n 1 and set -o pipefail, when head gets the first line it closes the pipe. But git tag (external process, sorting 1229 tags) is slow and may still be writing → SIGPIPE → exit 141.

Interesting. Could be hallucinated, could be true

lucasheriques: when I tried it locally it failed, try the sample code on te PR description, i was surprised myself

PR #2809, opened 2025-12-22
fix(surveys): android keyboard glitching

adboio:
## Problem
the keyboard causes major glitching in some cases on android. customer video here: https://posthoghelp.zendesk.com/agent/tickets/44614 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/48378)

reproduction here, on android 16 with physical keyboard + stylus input modes enabled:

[Screen Recording 2025-12-22 at 12.53.23 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/5fe5cfd4-f747-42a0-abe8-2ce942217ead.mov)

## Changes

adds a new sdk param `androidKeyboardBehavior` to control the behavior for `KeyboardAvoidingView` on android

setting it to `padding` resolved the issue in my testing, but i do not want to risk a regression here
### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/src/surveys/components/SurveyModal.tsx
+++ b/packages/react-native/src/surveys/components/SurveyModal.tsx
@@ -14,10 +14,11 @@ export type SurveyModalProps = {
   appearance: SurveyAppearanceTheme
   onShow: () => void
   onClose: (submitted: boolean) => void
+  androidKeyboardBehavior?: 'padding' | 'height'
 }
 
 export function SurveyModal(props: SurveyModalProps): JSX.Element | null {
-  const { survey, appearance, onShow } = props
+  const { survey, appearance, onShow, androidKeyboardBehavior = 'height' } = props
   const [isSurveySent, setIsSurveySent] = useState(false)
   const onClose = useCallback(() => props.onClose(isSurveySent), [isSurveySent, props])
   const insets = useOptionalSafeAreaInsets()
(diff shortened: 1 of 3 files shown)


packages/react-native/src/surveys/components/SurveyModal.tsx
```diff
@@ -39,7 +40,7 @@ export function SurveyModal(props: SurveyModalProps): JSX.Element | null {
   return (
     <Modal animationType="fade" transparent onRequestClose={onClose} statusBarTranslucent={true}>
       <KeyboardAvoidingView
-        behavior={Platform.OS === 'ios' ? 'padding' : 'height'}
```
ioannisj, line 42: Looks like we were deliberately setting `height` for Android. Wonder what the original reason was but @marandaneto is OOO

ioannisj (approved):
Approved to unblock you and the customer. Would prefer though we validate with them that this will work first since it seems we've been deliberately setting `height` before

packages/react-native/src/surveys/components/SurveyModal.tsx
adboio: yeah good question, i was trying to find out but didn't get very far  
  
as far as i can tell it's always been like that, and i think the bug we are seeing is related to this https://github.com/facebook/react-native/issues/45192, but still not 100% sure  
  
i asked the customer if they'd be able to pull our branch for testing so i can hang tight on merging until i hear back from her on that

adboio:
customer tested PR branch in emulator and said the fix was working, wants us to release it so they can do more testing on real devices. merging PR!

PR #2810, opened 2025-12-22
fix(replay): add targetSessionId to avoid session rotation bugs

ksvat:
## Problem

we're still seeing some instances of session IDs not rotating properly, there seems to be some subtle timing issue that happens some of the time. 

## Changes

This aims to be more explicit about what sessionID should be used, rather than rely on this.session_id, which is updated at various points in this flow and i suspect can be momentarily out of date?

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/early-rats-follow.md
+++ b/.changeset/early-rats-follow.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Add targetSessionID to lifecycle start events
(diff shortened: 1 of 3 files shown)


pauldambra (approved):
looks good to me :rocket:

PR #2811, opened 2025-12-23
refactor: Use new semiautomatic release workflow

rafaeelaudibert:
Rather than using local actions, let's just use the ones from master - delete the local ones. Also, make sure we are using our new `posthog/create-app-token` action - it's global and a fork from the Sentry one we were using previously

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -83,10 +83,10 @@ jobs:
 
             - name: Get GitHub App token
               id: releaser
-              uses: PostHog/create-app-token@main
+              uses: actions/create-github-app-token@29824e69f54612133e76f7eaac726eef6c875baf # v2.2.1
               with:
-                  app_id: ${{ secrets.GH_APP_POSTHOG_JS_RELEASER_APP_ID }}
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -106,7 +83,7 @@ jobs:
 
             - name: Get GitHub App token
               id: releaser
-              uses: getsentry/action-github-app-token@97c9e23528286821f97fba885c1b1123284b29cc # v2
+              uses: PostHog/create-app-token@main
```
rafaeelaudibert: @Piccirello I should remove this and use the GH one, right?

Piccirello: Yes! And update the variables from `app_id` to `app-id` and `private_key` to `private-key`.

rafaeelaudibert: Done

rafaeelaudibert:
* **#2811** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2811?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2811?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2812, opened 2025-12-23
fix(surveys): queue callbacks when fetch is in progress instead of returning error

lucasheriques:
## Summary

Fixes a race condition where `onSurveysLoaded` callbacks received an empty array on first page load.

**The bug:** `generateSurveys` starts an async fetch, and when `onSurveysLoaded` callbacks tried to call `getSurveys`, it immediately returned an empty array with error "Surveys are already being loaded" due to concurrent fetch protection.

**The fix:** Instead of returning an error when a fetch is in progress, we now track the in-flight request with a Promise. Multiple callers share the same Promise and all receive the result when it completes.

```typescript
// Before: immediately returned error when fetch in progress
if (this._isFetchingSurveys) {
    return callback([], {
        isLoaded: false,
        error: 'Surveys are already being loaded',
    })
}

// After: callers subscribe to the in-flight Promise
if (this._getSurveysInFlightPromise) {
    this._getSurveysInFlightPromise.then(({ surveys, context }) => callback(surveys, context))
    return
}
```

## Test plan

- [x] Added regression test for the race condition
- [x] Added unit tests for Promise reuse behavior  
- [x] All existing survey tests pass


adboio:
* **#2812** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2812?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2812?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

adboio (approved):
looks great, i did a couple things, will merge if everything passes on CI:  
  
1\. added lint ignore to Promise (ie11)  
2\. removed `targets` and `posthog-js.tgz` files from commit

lucasheriques:
@adboio merging this now. I changed to remove the promise linter, but instead, i'm checking if the Promise primitive exists. This means this code might not work properly on IE11, but I think that's better. This also amke sure our tests are passing in the IE11 environment too

PR #2813, opened 2025-12-24
feat(flags): add updateFlags() method for injecting flags without network request

ordehi:
## Problem

When using `instrumentation-client.ts` in Next.js, PostHog initializes once per app launch. This means feature flag values don't update on subsequent client-side navigations, even if a middleware evaluates fresh flags server-side. Users need a way to inject server-evaluated flag values into the client SDK without making additional network requests. This is applicable to other single-page applications with similar initialization methods.

Related: https://github.com/PostHog/posthog/issues/43980

## Changes

Adds a new public method `posthog.updateFlags(flags, payloads?, options?)` that:

- Injects feature flag values from an external source (e.g., server-side evaluation, edge middleware)
- Fires `onFeatureFlags` callbacks so React hooks auto-update
- Works with `advanced_disable_flags: true` for server-only evaluation patterns
- Supports `{ merge: true }` option to merge with existing flags instead of replacing

This is the foundational primitive: the transport/sync layer (cookies, navigation detection, race conditions) is a separate concern requiring more careful consideration.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/featureflags.test.ts
+++ b/packages/browser/src/__tests__/featureflags.test.ts
@@ -2623,4 +2623,65 @@ describe('updateFlags', () => {
 
         expect(posthog.isFeatureEnabled('test-flag')).toBe(true)
     })
+
+    it('should not make any network requests', async () => {
+        const posthog = await createPosthogInstance()
+        const sendRequestSpy = jest.spyOn(posthog, '_send_request')
+
+        posthog.updateFlags({ 'test-flag': true })
+
+        expect(sendRequestSpy).not.toHaveBeenCalled()
+    })
+
+    it('should handle empty flags object', async () => {
+        const posthog = await createPosthogInstance()
+
+        // Set initial flags
+        posthog.updateFlags({ 'flag-a': true, 'flag-b': 'variant-1' })
+        expect(posthog.getFeatureFlag('flag-a')).toBe(true)
+
+        // Update with empty object - should clear all flags
+        posthog.updateFlags({})
+
+        expect(posthog.getFeatureFlag('flag-a')).toBe(undefined)
+        expect(posthog.getFeatureFlag('flag-b')).toBe(undefined)
+        expect(posthog.featureFlags.getFlags()).toEqual([])
+    })
+
+    it('should persist flags to storage', async () => {
+        const posthog = await createPosthogInstance()
+
+        posthog.updateFlags(
+            { 'persisted-flag': true, 'variant-flag': 'control' },
+            { 'persisted-flag': { data: 'test' } }
+        )
+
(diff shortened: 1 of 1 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -1688,6 +1689,73 @@ export class PostHog {
...
+        const mergedFlags = { ...existingFlags, ...flags }
+        const mergedPayloads = { ...existingPayloads, ...payloads }
+
+        // Convert simple flags to v4 format to avoid deprecation warning
+        const flagDetails: Record<string, FeatureFlagDetail> = {}
+        for (const [key, value] of Object.entries(mergedFlags)) {
+            const isVariant = typeof value === 'string'
+            flagDetails[key] = {
+                key,
+                enabled: isVariant ? true : !!value,
```
haacked: Minor nit and completely a matter of preference. I know `!!value` is pretty idiomatic for JS, but `Boolean(value)` just feels clearer to me about intent.

```suggestion
                enabled: isVariant ? true : Boolean(value),
```

packages/browser/src/posthog-core.ts
```diff
@@ -1688,6 +1689,73 @@ export class PostHog {
...
+                variant: isVariant ? value : undefined,
+                reason: undefined,
+                metadata: !isUndefined(mergedPayloads?.[key])
+                    ? { id: 0, version: undefined, description: undefined, payload: mergedPayloads[key] }
```
haacked: A comment here about `id: 0` might help future maintainers.

```suggestion
                    // id: 0 indicates manually injected flags (not from server evaluation)
                    ? { id: 0, version: undefined, description: undefined, payload: mergedPayloads[key] }
```

packages/browser/src/posthog-core.ts
```diff
@@ -1688,6 +1689,73 @@ export class PostHog {
...
+    updateFlags(
+        flags: Record<string, boolean | string>,
+        payloads?: Record<string, JsonType>,
+        options?: { merge?: boolean }
+    ): void {
+        // If merging, combine with existing flags
+        const existingFlags = options?.merge ? this.featureFlags.getFlagVariants() : {}
+        const existingPayloads = options?.merge ? this.featureFlags.getFlagPayloads() : {}
+        const mergedFlags = { ...existingFlags, ...flags }
+        const mergedPayloads = { ...existingPayloads, ...payloads }
```
haacked: Minor nit: If `merge: false` then these aren't technically merged flags. Perhaps `finalFlags` and `finalPayloads` would be more accurate names? Regardless, I'm fine with the name as-is too.

PR #2814, opened 2025-12-26
feat: add `$device_id` to cookies

jurajmajerik:
## Changes
Persist device ID to cookies in addition to localStorage. This fixes an issue where device ID becomes null for identified users when localStorage is cleared.

Context: https://posthoghelp.zendesk.com/agent/tickets/45826 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/50767)

### DIFF
--- a/packages/browser/src/storage.ts
+++ b/packages/browser/src/storage.ts
@@ -1,6 +1,7 @@
 import { extend } from './utils'
 import { PersistentStore, Properties } from './types'
 import {
+    DEVICE_ID,
     DISTINCT_ID,
     ENABLE_PERSON_PROCESSING,
     INITIAL_PERSON_INFO,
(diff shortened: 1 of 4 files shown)


packages/browser/src/storage.ts
```diff
@@ -259,6 +260,7 @@ export const localStore: PersistentStore = {
 // This solves issues with cookies having too much data in them causing headers too large
 // Also makes sure we don't have to send a ton of data to the server
 const COOKIE_PERSISTED_PROPERTIES = [
+    DEVICE_ID,
```
pauldambra, line 263: agree with the robot, i don't know what tests there are already but we should have something on this and assert device id is included  
  
am actually amazed that it didn't cause a failing test here - i guess they all mock posthog or have inmemory persistence so it would show actually...

pauldambra (approved):
needs a changeset iiuc  
  
(i actually don't think this will solve the ticket in question but seems a safe change)

PR #2815, opened 2025-12-26
feat: move posthog-rrweb into the JS monorepo

pauldambra:
## Problem

with posthog-rrweb
we have to publish npm packages first
then test them in posthog-js
then release that

## Changes

silly
now we've converted it to pnpm
we can move it into the mono-repo
and we only need to publish npm packages because we use them in the posthog-app

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes


.changeset/config.json
```diff
@@ -8,5 +8,19 @@
     "baseBranch": "main",
     "updateInternalDependencies": "patch",
     "bumpVersionsWithWorkspaceProtocolOnly": true,
-    "ignore": []
+    "ignore": [
```
rafaeelaudibert, line 11: Why not use changesets for it?

rafaeelaudibert: I don't think the release workflow will work well if we're not using changesets :thinking_face:

packages/browser/package.json
```diff
@@ -66,9 +66,9 @@
         "@jest/globals": "^29.7.0",
         "@playwright/test": "^1.52.0",
         "@posthog-tooling/rollup-utils": "workspace:*",
-        "@posthog/rrweb-plugin-console-record": "^0.0.34",
-        "@posthog/rrweb-record": "^0.0.34",
-        "@posthog/rrweb-types": "^0.0.34",
+        "@posthog/rrweb-plugin-console-record": "workspace:*",
+        "@posthog/rrweb-record": "workspace:*",
+        "@posthog/rrweb-types": "workspace:*",
```
rafaeelaudibert, line 72: This probably deserves a patch bump?

packages/rrweb/README.md
```diff
@@ -0,0 +1,42 @@
...
+
+No, there are packages we don't need such as the svelte player and rr-video
+So, we deleted them...
+
+## Hello internal PostHog folk
+
+We build this and publish it to NPM so that we can use it in posthog-js
+
+If you want to contribute a change back to upstream rrweb
+then you need to open a person fork and contribute from there
```
rafaeelaudibert, line 25: Maybe explain why do we need to do this versus just using the upstream? mention they're slow to take upstreams, etc, etc. etc.

.changeset/config.json
pauldambra: ah! so, we don't need (i don't think) to release all these packages because we don't use them  
  
we only ever import these in posthog for replay and for that we're using the main `rrweb` package.

so, we should get a subset of the total number of rrweb packages released and then be able to import those in the main posthog app  
  
(i'm assuming this won't work the way i expect and will need a few tries to get right)

(TL;DR rrweb publishes tonnes of packages but we don't need to release them all i believe)

ksvat (approved):
Tested locally -- added a log statement to confirm my local build was indeed using this monorepo build, saw calls to /s were getting 200, saw that we seem to be successfully ingesting/playing back the recording still 
Screenshot 2026-01-20 at 1 25 11 PM
Screenshot 2026-01-20 at 1 24 40 PM
Screenshot 2026-01-20 at 1 24 02 PM

PR #2820, opened 2025-12-29
feat(surveys): add validation types and utilities

ordehi:
## Problem

Relies on https://github.com/PostHog/posthog/pull/44076 for setting lengths in surveys, but works without it as it's back compat.

Survey open text questions with `optional: false` (required) currently accept whitespace-only responses like `"   "` as valid input. Users can bypass required field validation by entering just spaces. Additionally, there's no way to configure validation rules (like minimum length) for survey responses.

  ## Changes

CleanShot 2025-12-29 at 18 28 20

React Native:

CleanShot 2025-12-29 at 19 18 50

- **Core (`@posthog/core`)**: Added `SurveyValidationType` enum and `SurveyValidationRule` interface. Created `getValidationError()`, `getMinLengthFromRules()`, and `getMaxLengthFromRules()` utilities in new `surveys/validation.ts` module. Extended `SurveyQuestionBase` type with optional `validation` field.
- **Browser SDK (`posthog-js`)**: Updated `OpenTextQuestion` component to use validation utilities. Now trims responses before submission, uses HTML5 `setCustomValidity()` for native validation UI, and supports `minLength`/`maxLength` textarea attributes. Added Playwright E2E tests.
- **React Native SDK (`posthog-react-native`)**: Updated `OpenTextQuestion` component with validation error state, red border on error, error text display below input, and trim on submit. Added unit tests.

 **Backwards compatible**: Existing surveys without `validation` rules work unchanged but now properly reject blank responses. API surveys receive validation rules in data but enforcement is left to the user (they can import `getValidationError` from `@posthog/core`).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,70 @@
...
+          break
+
+        case SurveyValidationType.MaxLength:
+          if (rule.value !== undefined && trimmed.length > rule.value) {
+            return rule.errorMessage ?? `Please enter no more than ${rule.value} characters`
+          }
+          break
+
+        case SurveyValidationType.Email:
+          if (!EMAIL_REGEX.test(trimmed)) {
```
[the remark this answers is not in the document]
ordehi: oof, went with a simpler check

ordehi:
Been thinking about this, this PR was conflating too many concerns. The user's original issue was that users can bypass required questions by entering empty spaces, I've kept that fix and the length utilities. However, for emails we should consider adding a specific "Email" question type that uses a proper email field for privacy and full email validation.

Now, users will be able to enforce a minimum and maximum length in text type questions, which is helpful if their getting many useless answers, but if they don't set a length constraint, they'll still get the benefit of not getting empty spaces that by pass required responses.

packages/react-native/src/surveys/components/QuestionTypes.tsx
```diff
@@ -55,6 +72,7 @@ export function OpenTextQuestion({
             {
               backgroundColor: appearance.inputBackground,
               color: appearance.inputTextColor ?? getContrastingTextColor(appearance.inputBackground),
+              borderColor: showError && validationError ? '#dc3545' : appearance.borderColor,
```
adboio: i hate the thought of adding ANOTHER color option to the survey appearance settings, but should this be customizable? or maybe we should just disable the submit button and pop the error message below it, w/o changing the border?

i could be making a big deal out of nothing lol but just something to consider

packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,62 @@
+import { SurveyValidationRule, SurveyValidationType } from '../types'
+
+/**
+ * Validates a survey open text response.
+ * Returns an error message string if invalid, or empty string if valid.
+ */
+export function getValidationError(
```
adboio, line 7: nice! this looks pretty extensible for future validation rules - e.g. email, customer-supplied regex, etc

packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,62 @@
...
+        case SurveyValidationType.MaxLength:
+          if (rule.value !== undefined && trimmed.length > rule.value) {
+            return rule.errorMessage ?? `Please enter no more than ${rule.value} characters`
+          }
+          break
+      }
+    }
+  }
+
+  return '' // Valid
```
adboio: [style nit] not a huge fan of empty string representing "no error", perhaps we could have it return `false` or throw? then we can try/catch or check if the validation result is truthy later. empty string is falsey, so this still works, but explicit is better imo :)

packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,62 @@
...
+}
+
+/**
+ * Helper to extract minLength value from validation rules
+ */
+export function getMinLengthFromRules(rules: SurveyValidationRule[] | undefined): number | undefined {
```
adboio: [nit] i think these two functions could just be one function that accepts a validation type

packages/react-native/src/surveys/components/QuestionTypes.tsx
ordehi: Yeah, I was torn between just a native error or this to avoid just that. I like the idea of a simple error below the submit/next button

packages/core/src/surveys/validation.ts
ordehi: Yup, hoping to do email but I think it should be a separate question type so we also mask it and such.

packages/browser/src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -100,6 +107,39 @@ export function OpenTextQuestion({
...
+    const validationError = useMemo(() => {
+        return getValidationError(text, question.validation, question.optional)
+    }, [text, question.validation, question.optional])
+
+    // Build requirements hint message
+    const minLength = getLengthFromRules(question.validation, SurveyValidationType.MinLength)
+    const maxLength = getLengthFromRules(question.validation, SurveyValidationType.MaxLength)
+    const requirementsHint = useMemo(() => {
+        if (minLength && maxLength) {
+            return `Enter ${minLength}-${maxLength} characters`
```
adboio: `character${length > 1 ? 's' : ''}` :stuck_out_tongue:

packages/browser/src/extensions/surveys/components/QuestionTypes.tsx
```diff
@@ -100,6 +107,39 @@ export function OpenTextQuestion({
...
+            return `Maximum ${maxLength} characters`
+        }
+        return null
+    }, [minLength, maxLength])
+
+    const handleSubmit = () => {
+        if (validationError) {
+            return
+        }
+        onSubmit(text.trim())
```
adboio, line 133: [nit] maybe we can trim at the source to avoid having to call .trim in all the submit handlers -- maybe in the change handlers?

packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,56 @@
...
+    return false
+  }
+
+  // Apply validation rules (only if configured)
+  if (rules && rules.length > 0) {
+    for (const rule of rules) {
+      switch (rule.type) {
+        case SurveyValidationType.MinLength:
+          if (rule.value !== undefined && trimmed.length < rule.value) {
+            return rule.errorMessage ?? `Please enter at least ${rule.value} characters`
```
adboio, line 30: we've got similar logic for building this string in QuestionTypes, do we need it in both places / can it be extracted to a util?

packages/react-native/src/surveys/components/QuestionTypes.tsx
```diff
@@ -40,6 +43,33 @@ export function OpenTextQuestion({
...
+    () => getValidationError(text, question.validation, question.optional),
+    [text, question.validation, question.optional]
+  )
+
+  // Build requirements hint message
+  const minLength = getLengthFromRules(question.validation, SurveyValidationType.MinLength)
+  const maxLength = getLengthFromRules(question.validation, SurveyValidationType.MaxLength)
+  const requirementsHint = useMemo(() => {
+    if (minLength && maxLength) {
+      return `Enter ${minLength}-${maxLength} characters`
```
adboio: same comment on the string building here

packages/browser/src/extensions/surveys/components/QuestionTypes.tsx
ordehi: Haha, small deet. Been thinking that this is similar to a required question, which always has a 1 character requirement now, and if a user sets a 1 to 1 length on a required, no change. If they set it on an optional, the user can still skip the question without entering anything. Still, a nice catch to have, so I'll include it.

packages/browser/src/extensions/surveys/components/QuestionTypes.tsx
ordehi: I want to avoid the user getting trimmed every time they input something, causing potential flashing or even breaking input. I've seen it happen many times in other areas. Submitting to safety here. 😄

packages/core/src/surveys/validation.ts
ordehi: LinkedIn style: _agreed_

packages/core/src/surveys/validation.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Validates a survey open text response.
+ * Returns an error message string if invalid, or false if valid.
+ */
+export function getValidationError(
+  value: string,
+  rules: SurveyValidationRule[] | undefined,
+  optional: boolean | undefined
+): string | false {
```
lucasheriques, line 11: nit: i'd remove this return types and use typescript infer instead

lucasheriques (approved):
minor comment. PR looks great overall

marandaneto (approved):

PR #2823, opened 2025-12-30
feat: use native compression stream not fflate in request.ts

pauldambra:
## Problem

we ship fflate with the bundle

it's about 3.5% of what we ship

so that we can do partial compression in replay

and we compress all events before sending them over the wire (json is v. compressible)

https://caniuse.com/?search=compressionstream

CompressionStream in the browser is widely available (not ES5 but we have a separate ES5 bundle)

Fflate have a section on this

https://github.com/101arrowz/fflate#what-about-compressionstream

we want to keep compression of payloads since it reduces the amount of data in-memory and over the internal network (as opposed to just the WAN hop reduction we get with transport compression)

but we want to reduce bundle size as much as sensible

## Changes

switch to the compression stream API

so far i've confirmed i can still self-record posthog in chrome

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/playground/nextjs/bin/localdev.sh
+++ b/playground/nextjs/bin/localdev.sh
@@ -26,25 +26,53 @@ echo ""
 echo -e "${GREEN}2. Creating tarballs...${NC}"
 pnpm package
 
-# Step 3: Ensure .pnpmfile.cjs symlink exists
+# Step 3: Install into local PostHog repo if POSTHOG_REPO is set
+if [ -n "$POSTHOG_REPO" ]; then
+    echo ""
+    echo -e "${GREEN}3. Installing posthog-js into local PostHog repo...${NC}"
+
+    if [ ! -d "$POSTHOG_REPO" ]; then
+        echo -e "${YELLOW}Error: POSTHOG_REPO directory does not exist: $POSTHOG_REPO${NC}"
+        exit 1
+    fi
+
+    TARBALL="$REPO_ROOT/target/posthog-js.tgz"
+    if [ ! -f "$TARBALL" ]; then
+        echo -e "${YELLOW}Error: Tarball not found: $TARBALL${NC}"
+        exit 1
+    fi
+
+    cd "$POSTHOG_REPO"
+    pnpm -r update "posthog-js@file:$TARBALL"
+    pnpm install
+    echo ""
+    echo -e "${GREEN}Running frontend copy-scripts...${NC}"
+    cd frontend && pnpm run copy-scripts
+    echo -e "${GREEN}Installed posthog-js from $TARBALL${NC}"
+fi
+
+# Step 4: Ensure .pnpmfile.cjs symlink exists
 echo ""
-echo -e "${GREEN}3. Setting up .pnpmfile.cjs symlink...${NC}"
+echo -e "${GREEN}4. Setting up .pnpmfile.cjs symlink...${NC}"
 cd "$PLAYGROUND_DIR"
 if [ ! -e .pnpmfile.cjs ]; then
     ln -s ../.pnpmfile.cjs .pnpmfile.cjs
     echo "Created symlink to .pnpmfile.cjs"
 fi
 
-# Step 4: Install dependencies in playground
+# Step 5: Install dependencies in playground
 echo ""
-echo -e "${GREEN}4. Installing playground dependencies...${NC}"
+echo -e "${GREEN}5. Installing playground dependencies...${NC}"
 rm -rf node_modules pnpm-lock.yaml
 pnpm install
 
-# Step 5: Run dev server
+# Step 6: Run dev server
 echo ""
-echo -e "${GREEN}5. Starting dev server...${NC}"
+echo -e "${GREEN}6. Starting dev server...${NC}"
 echo -e "${YELLOW}Tip: Open localhost:3000?__posthog_debug=true to see debug logs${NC}"
+if [ -n "$POSTHOG_REPO" ]; then
+    echo -e "${YELLOW}Tip: posthog-js was also installed in $POSTHOG_REPO${NC}"
+fi
 echo ""
 
 # Pass through all environment variables and run dev


pauldambra:
* **#2823** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2823?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2823?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/__tests__/request.test.ts
```diff
@@ -20,8 +20,10 @@ jest.mock('../config', () => ({ DEBUG: false, LIB_VERSION: '1.23.45' }))
 
 const flushPromises = async () => {
     jest.useRealTimers()
-    await new Promise((res) => setTimeout(res, 0))
-    jest.useRealTimers()
+    // Need multiple flush cycles to resolve the async encoding chain
+    for (let i = 0; i < 10; i++) {
+        await new Promise((res) => setTimeout(res, 0))
+    }
```
lricoy, line 26: Since those are tests, can we use a utility for this macrotask shenanigan?

packages/browser/src/entrypoints/array.full.es5.ts
```diff
@@ -5,6 +5,7 @@
 // it doesn't include recorder which doesn't support IE11,
 // and it doesn't include "web-vitals" which doesn't support IE11
 
+import 'core-js/features/promise'
```
rafaeelaudibert, line 8: Is it even possible to have a PR from you where you aren't attempting to add promises to ES5

packages/browser/src/request.ts
```diff
@@ -5,10 +5,28 @@ import { formDataToQuery } from './utils/request-utils'
 
 import { logger } from './utils/logger'
 import { AbortController, fetch, navigator, XMLHttpRequest } from './utils/globals'
-import { gzipSync, strToU8 } from 'fflate'
 
 import { _base64Encode } from './utils/encode-utils'
 
+/* eslint-disable compat/compat */
+const supportsCompressionStream = typeof CompressionStream !== 'undefined'
```
rafaeelaudibert: Where's `CompressionStream` coming from when this runs on a browser where it isn't defined? Shouldn't this follow the approach we've always had and come from `./utils/globals` instead? This will guarantee we don't raise an error when it doesn't exist

packages/browser/src/request.ts
```diff
@@ -5,10 +5,28 @@ import { formDataToQuery } from './utils/request-utils'
...
+}
+
+const gzipCompress = async (data: string): Promise<Blob | null> => {
+    if (!supportsCompressionStream) {
+        return null
+    }
+    const stream = new Blob([data]).stream()
+    const compressedStream = stream.pipeThrough(new CompressionStream('gzip'))
+    return new Response(compressedStream).blob()
+}
```
rafaeelaudibert: This doesn't need to be `async`. Maybe it returns a `Promise` because of the `blob` interface, but it doesn't need to be declared `async`

packages/browser/src/request.ts
```diff
@@ -68,19 +86,21 @@ const encodeToDataString = (data: string | Record<string, any>): string => {
...
-            estimatedSize: blob.size,
+        const blob = await gzipCompress(jsonStringify(data))
+        if (blob) {
+            return {
+                contentType: CONTENT_TYPE_PLAIN,
+                body: blob,
+                estimatedSize: blob.size,
+            }
         }
+        // Fall through to JSON if CompressionStream not available (e.g. ES5 build)
```
rafaeelaudibert: This comment would make way more sense above `if (blob)`

packages/browser/src/request.ts
```diff
@@ -68,19 +86,21 @@ const encodeToDataString = (data: string | Record<string, any>): string => {
     return 'data=' + encodeURIComponent(typeof data === 'string' ? data : jsonStringify(data))
 }
 
-const encodePostData = ({ data, compression }: RequestWithOptions): EncodedBody | undefined => {
+const encodePostData = async ({ data, compression }: RequestWithOptions): Promise<EncodedBody | undefined> => {
     if (!data) {
         return
     }
 
     if (compression === Compression.GZipJS) {
```
rafaeelaudibert, line 93: We should probably document somewhere that if you declare `GZipJS` compression we might NOT compress it in the rare case where the browser doesn't provide a `CompressionStream` interface

packages/browser/src/request.ts
```diff
@@ -231,7 +249,7 @@ const _sendBeacon = (options: RequestWithOptions) => {
 
 const AVAILABLE_TRANSPORTS: {
     transport: RequestWithOptions['transport']
-    method: (options: RequestWithOptions) => void
+    method: (options: RequestWithOptions) => void | Promise<void>
```
rafaeelaudibert: Isn't it ALWAYS `Promise<void>` now?

packages/browser/src/request.ts
```diff
@@ -257,11 +275,15 @@ if (navigator?.sendBeacon) {
...
 // This is the entrypoint. It takes care of sanitizing the options and then calls the appropriate request method.
-export const request = (_options: RequestWithOptions) => {
+export const request = (_options: RequestWithOptions): void => {
     // Clone the options so we don't modify the original object
     const options = { ..._options }
     options.timeout = options.timeout || 60000
 
+    if (options.compression === Compression.GZipJS && !supportsCompressionStream) {
+        options.compression = undefined
+    }
```
rafaeelaudibert: Not really necessary, correct? Any benefit to unsetting this here vs. just trusting the fallback at the top?

packages/browser/src/request.ts
```diff
@@ -281,5 +303,10 @@ export const request = (_options: RequestWithOptions) => {
         throw new Error('No available transport method')
     }
 
-    transportMethod(options)
+    sendChain = sendChain
+        .then(() => transportMethod(options))
+        .catch((e) => {
+            logger.error('Request failed:', e)
+            options.callback?.({ statusCode: 0, text: String(e) })
+        })
```
rafaeelaudibert: Huh, this is a bit confusing. I don't understand JS Promises well enough but what happens if you do `.then.catch.then.catch` like you're doing here (subsequent requests will just thread the chain). Is `catch` gonna act like we're expecting?  
  
Also, why do we even need `catch` if we didn't before? I see you're calling `options?.callback` but that's a new feature. Why?

packages/browser/src/request.ts
```diff
@@ -5,10 +5,28 @@ import { formDataToQuery } from './utils/request-utils'
...
+    sendChain = Promise.resolve()
+}
+
+const gzipCompress = async (data: string): Promise<Blob | null> => {
+    if (!supportsCompressionStream) {
+        return null
+    }
+    const stream = new Blob([data]).stream()
+    const compressedStream = stream.pipeThrough(new CompressionStream('gzip'))
+    return new Response(compressedStream).blob()
```
lricoy, line 24:
```suggestion
    return new Response(compressedStream).blob().then(
    blob => new Blob([blob], { type: 'application/gzip' })
)
```

we can set the blob mime type here as well

packages/browser/src/entrypoints/array.full.es5.ts
pauldambra: es5 is so long ago now

packages/browser/src/request.ts
pauldambra: we await it when called though. so would have to change call site as well if it wasn't async, no?

packages/browser/src/request.ts
pauldambra: it caused ES5 tests to fail since they didn't know that the compression was unavailable  
  
i'll see if i can find another way around or comment why it's done :+1:

packages/browser/src/request.ts
pauldambra: because it was 1am :side-eye-monkey.meme:

pauldambra: because it was 1am :side-eye-monkey.meme:

pauldambra: side-eye-monkey

packages/browser/src/entrypoints/array.full.es5.ts
pauldambra: `es6 is ten years old `:exploding_head:

packages/browser/src/request.ts
pauldambra: actually, it's because we're too eager to set the compression on the URL IMHO... :pencil2:

pauldambra: actually, it's because we're too eager to set the compression on the URL IMHO... :pencil2:

packages/browser/src/request.ts
rafaeelaudibert: That's not how it works. You await a promise, not a call to an async function.

This works (and its equivalent to what you have here). `function` is not async, you're awaiting the promise returned by it

```
const function = () => Promise.resolve()
const answer = await function()

```

pauldambra: wait... what?! you are blowing my mental model apart here... really :eyes:

rafaeelaudibert: Wait until you learn you can `await nonAsyncFunction()` and it works just like you'd expect

rafaeelaudibert (approved):
Why are we sending the compression as a query parameter now? Are we doing anything with that value on the backend?

packages/browser/src/request.ts
```diff
@@ -257,15 +274,14 @@ if (navigator?.sendBeacon) {
...
-export const request = (_options: RequestWithOptions) => {
+export const request = (_options: RequestWithOptions): void => {
     // Clone the options so we don't modify the original object
     const options = { ..._options }
     options.timeout = options.timeout || 60000
 
     options.url = extendURLParams(options.url, {
         _: new Date().getTime().toString(),
         ver: Config.LIB_VERSION,
-        compression: options.compression,
```
pauldambra, line 268: we were always setting compression on the URL  
just now there's a fall back we can't always trust the option and let the transport method decide whether to do it or not

PR #2825, opened 2025-12-30
fix(flags): Apply exponential backoff to on-demand fetches after errors

haacked:
## Problem

If the client is configured with an invalid api token, the polling fetcher will receive an `HTTP 401 Unauthorized` response and the poller does an exponential backoff of the next requests. That's all fine and dandy.

However, on-demand fetches don’t respect backoff!

On `401`, `loadedSuccessfullyOnce` stays `false`. So every `getFeatureFlag()` call:

1. Passes the early-exit check
2. If no request is in-flight, starts a new one immediately
3. Once that 401 completes, `loadingPromise` is cleared
4. Next `getFeatureFlag()` call starts another request

## Changes

On-demand fetches triggered by `getFeatureFlag()` calls now respect exponential backoff when the server returns 401, 403, or 429 errors.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/extensions/feature-flags/feature-flags.ts
+++ b/packages/node/src/extensions/feature-flags/feature-flags.ts
@@ -634,6 +634,8 @@ class FeatureFlagsPoller {
   private beginBackoff(): void {
     this.shouldBeginExponentialBackoff = true
     this.backOffCount += 1
+    // Use the same backoff interval as the poller to avoid overwhelming
+    // the server with on-demand requests while polling is backed off.
     this.nextFetchAllowedAt = Date.now() + this.getPollingInterval()
   }
 


packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -619,6 +627,26 @@ class FeatureFlagsPoller {
...
   }
 
+  /**
+   * Enter backoff state after receiving an error response (401, 403, 429).
+   * This enables exponential backoff for the poller and blocks on-demand fetches.
+   */
+  private beginBackoff(): void {
+    this.shouldBeginExponentialBackoff = true
+    this.backOffCount += 1
+    this.nextFetchAllowedAt = Date.now() + this.getPollingInterval()
```
gustavohstrassburger, line 639: Seems odd to use the `getPollingInterval` in this method, because we are coupling the backoff with the polling interval (method `getPollingInterval` and params `this.pollingInterval`)

haacked: The coupling is intentional — when the server returns 401/403/429, we want all requests to that endpoint to back off, not just the poller. If on-demand fetches used a different (shorter) interval, they could still hammer the server while the poller is waiting.

That said, the naming could be clearer. `getPollingInterval()` does suggest it's polling-specific when it's really "the interval before the next attempt." I could rename it to something like `getBackoffInterval()` or `getNextRetryInterval()` if you think that'd help clarify the intent?

gustavohstrassburger: the only reason why I didn't suggest to rename it is because inside `getPollingInterval` it uses this.pollingInterval and it also tests:

```
    if (!this.shouldBeginExponentialBackoff) {
      return this.pollingInterval
    }
```

but since you're intentionally want to make this coupled, maybe simply renaming it would suffice (maybe not even needed though, up to you)

haacked: I'll add a clarifying comment for now.

PR #2832, opened 2025-12-30
feat(persistence): introduce custom cookie properties for localStorage+cookie persistence

rafaeelaudibert:
- Added `cookie_persisted_properties` to `PostHogConfig` to allow users to specify additional properties to be stored in cookies.
- Implemented `createLocalPlusCookieStore` to handle custom cookie properties during storage operations.
- Updated `PostHogPersistence` to utilize the new custom properties when creating the localPlusCookieStore.
- Adjusted tests to ensure correct behavior with the new persistence configuration.

### DIFF
--- a/packages/browser/src/posthog-persistence.ts
+++ b/packages/browser/src/posthog-persistence.ts
@@ -44,6 +44,16 @@ const parseName = (config: PostHogConfig): string => {
     }
 }
 
+const isArrayContentsEqual = (arr1: readonly string[], arr2: readonly string[]): boolean => {
+    if (arr1.length !== arr2.length) {
+        return false
+    }
+
+    const sortedArr1 = [...arr1].sort()
+    const sortedArr2 = [...arr2].sort()
+    return sortedArr1.every((item, index) => item === sortedArr2[index])
+}
+
 /**
  * PostHog Persistence Object
  * @constructor
@@ -333,7 +343,10 @@ export class PostHogPersistence {
         this.set_secure(config['secure_cookie'])
 
         // If the persistence type has changed, we need to migrate the data.
-        if (config.persistence !== oldConfig.persistence) {
+        if (
+            config.persistence !== oldConfig.persistence ||
+            !isArrayContentsEqual(config.cookie_persisted_properties || [], oldConfig.cookie_persisted_properties || [])
+        ) {
             const newStore = this._buildStorage(config)
             const props = this.props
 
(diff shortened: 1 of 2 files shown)


packages/browser/src/posthog-persistence.ts
```diff
@@ -99,25 +99,39 @@ export class PostHogPersistence {
...
-            store = localPlusCookieStore
         } else {
+            // UNREACHABLE CODE - `storage_type` is `never` but we need this to guarantee
+            // TS doesn't think `store` is possibly `undefined`
             store = cookieStore
         }
 
+        // If the chosen storage type is not supported,
+        // we fallback to 'localstorage+cookie' or 'cookie'
+        if (!store._is_supported()) {
```
pauldambra: i wonder if we should add a `logger.info` when we do this fallback 
how would people find out about it?
or is it not something they can do anything about i suppose

packages/browser/src/posthog-persistence.ts
```diff
@@ -99,25 +99,39 @@ export class PostHogPersistence {
...
         // We handle storage type in a case-insensitive way for backwards compatibility
         const storage_type = config['persistence'].toLowerCase() as Lowercase<PostHogConfig['persistence']>
-        if (storage_type === 'localstorage' && localStore._is_supported()) {
+        if (storage_type === 'localstorage') {
             store = localStore
-        } else if (storage_type === 'localstorage+cookie' && localPlusCookieStore._is_supported()) {
-            store = localPlusCookieStore
-        } else if (storage_type === 'sessionstorage' && sessionStore._is_supported()) {
+        } else if (storage_type === 'localstorage+cookie') {
+            const customCookieProperties = config['cookie_persisted_properties'] || []
```
pauldambra: yep, i think i agree with the robot... is there a way we can read these from config when we need them? that way someone can change the config when they need to

e.g. they have an API call that eventually is made that returns `customerIsVIP: true` and that's when they know they want to add `customerCanBuyThingsForFree` to the cookie jar

alternatively - we document that you have to know the props before posthog init

rafaeelaudibert: Yeah, _maybe_ I can make this work differently to always read from the config. If I can't get that to work in like 10min then I'll just add a comment explaining that settings these on `setConfig`doesn't have any effect

rafaeelaudibert: Ah, no, this is very easy to do, pushing commit

packages/browser/src/posthog-persistence.ts
rafaeelaudibert: Yeah, there's not much they can do but it might help debugging why PostHog might not be working for some specific customers

PR #2834, opened 2025-12-31
fix(web): Clear PageViewManager state on session rotation

ordehi:
## Problem

`PageViewManager` doesn't subscribe to session rotation events. When a user backgrounds a tab and returns after the session expires (30 min idle), the SDK still holds the old pageview timestamp. The next $pageview event calculates $prev_pageview_duration from that stale timestamp, producing impossible values (94+ hours observed) that corrupt "Average Time on Page" metrics in web analytics.

Related to the fix in #2666 

## Changes

When a browser tab is backgrounded and the session rotates (30 min idle or 24 hour max), `PageViewManager` now clears its state.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-core.ts
```diff
@@ -3274,6 +3276,7 @@ export class PostHog {
                 $device_id: null,
             })
             this.sessionManager?.destroy()
+            this.pageViewManager?.destroy()
```
jordanm-posthog, line 3279: I don't have much context here, but do you need to set the `pageViewManager` to `undefined` like we do with the `sessionManager`?

ordehi: On first glance, I thought so, `destroy()` only kills the session rotation subscription, the manager's core job (tracking pageview state, returning properties) keeps working. In cookieless mode we still want pageview tracking, just without session-aware features. Thanks for looking out!

PR #2842, opened 2026-01-05
fix: undefined constructor error

carlos-marchal-ph:
Fixes the following error with certain AI imports when using CJS:
```
TypeError: Class extends value undefined is not a constructor or null
```

[Related community question](https://posthog.com/questions/posthog-ai-incompatible-with-google-genai).

### DIFF
--- a/.changeset/smart-otters-throw.md
+++ b/.changeset/smart-otters-throw.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Fixes an error with certain AI imports when using CJS (`TypeError: Class extends value undefined is not a constructor or null`)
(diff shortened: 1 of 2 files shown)


Radu-Raicea (approved):
I would love to see more in the PR description, regarding the fix, especially for cases where the fix is not some common knowledge, like `interop`.

carlos-marchal-ph:
> I would love to see more in the PR description, regarding the fix, especially for cases where the fix is not some common knowledge, like interop.

It's the rollup equivalent [to setting `esModuleInterop` for TS](https://www.typescriptlang.org/tsconfig/#esModuleInterop). Internal ESM <-> CJS compilation nitty gritty basically.

PR #2846, opened 2026-01-05
fix(surveys): respect branching logic in URL prefill

ordehi:
## Problem

When using URL parameters to prefill survey responses, the SDK now correctly respects the survey's branching configuration. Previously, prefilled answers would always advance to the next sequential question, ignoring any skip logic configured on the question.

## Changes

- Extract getNextSurveyStep to shared utility (survey-branching.ts)
- Update calculatePrefillStartIndex to use branching logic
- Add tests for response-based, end, and specific-question branching

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/utils/survey-url-prefill.ts
+++ b/packages/browser/src/utils/survey-url-prefill.ts
@@ -1,7 +1,8 @@
-import { Survey, SurveyQuestion, SurveyQuestionType } from '../posthog-surveys-types'
+import { Survey, SurveyQuestion, SurveyQuestionBranchingType, SurveyQuestionType } from '../posthog-surveys-types'
 import { getSurveyResponseKey } from '../extensions/surveys/surveys-extension-utils'
 import { logger } from './logger'
 import { isUndefined } from '@posthog/core'
+import { getNextSurveyStep } from './survey-branching'
 
 /**
  * Extracted URL prefill parameters by question index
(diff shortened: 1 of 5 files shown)


packages/browser/src/utils/survey-url-prefill.ts
```diff
@@ -138,41 +139,57 @@ export function convertPrefillToResponses(survey: Survey, prefillParams: Prefill
...
     responses: Record<string, any>
 ): { startQuestionIndex: number; skippedResponses: Record<string, any> } {
-    let startQuestionIndex = 0
+    let currentIndex = 0
     const skippedResponses: Record<string, any> = {}
 
-    for (let i = 0; i < questions.length; i++) {
-        // stop at the first question that is not prefilled
-        if (!prefilledIndices.includes(i)) {
+    while (currentIndex < survey.questions.length) {
```
lucasheriques:
```suggestion
  const MAX_ITERATIONS = survey.questions.length + 1
  let iterations = 0
  while (currentIndex < survey.questions.length && iterations < MAX_ITERATIONS) {
```


while the BE validates for cycles, i think we could use some defensive programming here and add this so we know it has a hard stop and can't be on an infinite loop

lucasheriques (approved):
one comment but looks good!

PR #2849, opened 2026-01-06
feat(conversations): separate enable with widget enable

veryayskiy:
## Problem

We want to separate conversation initialization with conversations widget.

## Changes

`enable` now only initializes the extension and allow to use api
`widgetEnabled` initializes the widget

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/conversations/external/index.tsx
+++ b/packages/browser/src/extensions/conversations/external/index.tsx
@@ -82,11 +82,8 @@ export class ConversationsManager implements ConversationsManagerInterface {
     // SECURITY: widget_session_id is the key for access control
     // This is a random UUID that only this browser knows
     private _widgetSessionId: string
-    // Whether widget should be shown (based on widgetEnabled config)
     private _isWidgetEnabled: boolean
-    // Whether current domain is allowed for widget display
     private _isDomainAllowed: boolean
-    // Whether the widget has been rendered to the DOM
     private _isWidgetRendered: boolean = false
 
     constructor(


packages/browser/src/posthog-conversations-types.ts
```diff
@@ -4,9 +4,17 @@
...
+     * When true, the conversations API is available (posthog.conversations.*)
      */
     enabled: boolean
 
+    /**
+     * Whether the widget UI (button + chat panel) should be shown
+     * Only takes effect when enabled is true
+     * @default false
+     */
+    widgetEnabled?: boolean
```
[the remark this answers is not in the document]
veryayskiy: it's ok, as no one uses the current version yet

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -22,6 +22,52 @@ const logger = createLogger('[ConversationsManager]')
 const WIDGET_CONTAINER_ID = 'ph-conversations-widget-container'
 const POLL_INTERVAL_MS = 5000 // 5 seconds
 
+/**
+ * Extract hostname from a domain string (handles URLs and plain hostnames)
+ */
+function extractHostname(domain: string): string | null {
```
joshsny, line 28: this logic feels like its likely to exist elsewhre in posthog-js, does it, if not can we add some tests for this?

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -227,6 +287,44 @@ export class ConversationsManager implements ConversationsManagerInterface {
...
+            hasExistingTicket: !!this._currentTicketId,
+            widgetEnabled: this._isWidgetEnabled,
+            domainAllowed: this._isDomainAllowed,
+        })
+
+        // Only render widget if both widgetEnabled and domain is allowed
+        if (this._isWidgetEnabled && this._isDomainAllowed) {
+            this._initializeWidget()
+        } else {
+            logger.info('Widget not rendered', {
```
joshsny, line 298: not so familiar with posthog-js, does this log in the users browser or only if __posthog_debug=true?

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -506,8 +594,21 @@ export class ConversationsManager implements ConversationsManagerInterface {
...
      */
     enable(): void {
+        // Check domain restrictions - don't render on disallowed domains
+        if (!this._isDomainAllowed) {
+            logger.warn('Cannot enable widget: current domain is not allowed')
+            return
+        }
+
+        // If widget isn't rendered yet, render it now
+        if (!this._isWidgetRendered) {
```
joshsny, line 605: what happens if I call `enable()` twice, before `isWidgetRendered` _is set by_ `initialize()`. It looks like `_initialize()` is idempotent, so it should be fine?  
  
If so, can we make sure we have a test that checks idempotency for the initialize function?

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -518,6 +619,13 @@ export class ConversationsManager implements ConversationsManagerInterface {
         this._widgetRef?.hide()
     }
 
+    /**
+     * Check if the widget is currently visible (rendered and shown)
+     */
+    isWidgetVisible(): boolean {
+        return this._isWidgetRendered && !isNull(this._widgetRef)
```
joshsny, line 623: nit: we could pull `!isNull(this._widgetRef)` into its own function `isWidgetShown` (either public or private, depending on what you think makes sense) and leave a comment explaining what the difference is between `visible` and `shown` - and which one I want to check depending on what I am doing?

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -227,9 +179,25 @@ export class PostHogConversations {
...
     isEnabled(): boolean {
         return this._isConversationsEnabled === true
     }
+
+    /**
+     * Check if widget should be shown (based on remote config widgetEnabled)
+     * This is separate from isEnabled() - API can be enabled without showing the widget
+     */
+    isWidgetEnabled(): boolean {
+        return this._remoteConfig?.widgetEnabled === true
```
joshsny, line 194: are we 100% based on remote config now, is this always set?

packages/browser/src/extensions/conversations/external/index.tsx
veryayskiy: correct

packages/browser/src/extensions/conversations/external/index.tsx
veryayskiy: this was just moved to index from posthog-converstaions.

packages/browser/src/extensions/conversations/posthog-conversations.ts
veryayskiy: No, not 100% remote config, and not always set.
_remoteConfig can be null if:

- Remote config hasn't loaded yet
- disable_conversations is set in init config (client override)
- Server doesn't provide conversations config

The ?. operator handles this - returns false when _remoteConfig is null.
Client init config acts as a kill switch that can prevent remote config from ever loading.

PR #2851, opened 2026-01-06
feat: expose conversations api

veryayskiy:
## Problem

Developers can only use the built-in conversations widget UI.  The API methods for sending messages, getting ticket etc are not exposed. This blocks custom chat UIs or integrating conversations into existing applications.

## Changes

Expose 6 api methods to `posthog.conversations`:

- `sendMessage(message, userTraits?, newTicket?)` - Send a message (creates ticket if needed, optional newTicket to force new conversation)
- `getMessages(ticketId?, after?)` - Get messages for a ticket (defaults to current, supports pagination)
- `markAsRead(ticketId?)` - Mark messages as read (defaults to current ticket)
- `getTickets(options?)` - Get list of tickets with filtering and pagination
- `getCurrentTicketId()` - Get current active ticket ID
- `getWidgetSessionId()` - Get persistent browser session ID

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/extensions/conversations/conversations-manager.test.tsx
+++ b/packages/browser/src/__tests__/extensions/conversations/conversations-manager.test.tsx
@@ -588,17 +588,20 @@ describe('ConversationsManager', () => {
             expect(manager['_currentTicketId']).toBe('ticket-123')
         })
 
-        it('should save widget state when changed', () => {
+        it('should load saved widget state when re-rendered after hide', () => {
+            // Hide the widget
             act(() => {
-                manager.enable()
+                manager.hide()
             })
+            expect(manager.isVisible()).toBe(false)
 
-            expect(mockPosthog.capture).toHaveBeenCalledWith(
-                '$conversations_widget_state_changed',
-                expect.objectContaining({
-                    state: 'open',
-                })
-            )
+            // Re-show the widget - it should load saved state from persistence
+            act(() => {
+                manager.show()
+            })
+            expect(manager.isVisible()).toBe(true)
+            // Widget state is loaded from persistence in _initializeWidget
+            // The persistence mock returns 'closed' for loadWidgetState
         })
     })
 })


packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -590,44 +626,114 @@ export class ConversationsManager implements ConversationsManagerInterface {
     }
 
     /**
-     * Enable/show the widget (button and chat panel)
-     * If widget wasn't initially rendered (widgetEnabled was false), this will render it.
+     * Show the widget (render it to DOM).
+     * The widget respects its saved state (open/closed).
```
adboio, line 630: forgive me i have very little context on the conversations product (although i would love to learn more, i always thought posthog should build this lol)

but - should calling `show` respect the existing open/closed state?

i would think a manual call to `show` should open the widget regardless, unless "hidden" here means there is also some small "click to open the chat" button so _something_ is always made visible on the dom after calling `show`?

packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -590,44 +626,114 @@ export class ConversationsManager implements ConversationsManagerInterface {
...
+        return this._isWidgetRendered
+    }
+
+    /** Get tickets list for the current widget session */
+    async getTickets(options?: GetTicketsOptions): Promise<GetTicketsResponse> {
+        const token = this._config.token
+
+        const queryParams: Record<string, string> = {
+            widget_session_id: this._widgetSessionId,
+            limit: String(options?.limit ?? 20),
```
adboio, line 679: [nit] 20 feels like it could be low as the default, but we can always review usage data and adjust if needed :muscle:

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -153,51 +161,168 @@ export class PostHogConversations {
...
+     * const newConvo = await posthog.conversations.sendMessage('Start fresh', undefined, true)
      */
-    isWidgetEnabled(): boolean {
-        return this._remoteConfig?.widgetEnabled === true
+    async sendMessage(
+        message: string,
+        userTraits?: UserProvidedTraits,
+        newTicket?: boolean
+    ): Promise<SendMessageResponse | null> {
+        if (!this._conversationsManager) {
```
adboio, line 223: do users have a way to "wait" for conversations to be loaded, to avoid something like polling this method til it works?

same comment for the other methods that return early if conversationsManager does not exist

i think we encountered something similar in surveys recently, a customer had to poll for some time on page load before being able to interact with surveys apis

packages/browser/src/extensions/conversations/external/index.tsx
veryayskiy: show / hide - show hide the widget (round one)
open/close - it's about the chat itself

so via API we need to be able to render (show the round one) and remove

packages/browser/src/extensions/conversations/external/index.tsx
veryayskiy: good point...maybe 100 / 1000 would be better

PR #2865, opened 2026-01-09
feat(product tours): support banner rendering

adboio:
## Problem

sdk needs to render banners for product tours 😎

## Changes

adds support for rendering banners:

- always at the top of the page, with behavior options:
    - **static**: stays up there when you scroll down
    - **sticky**: stays at the top while you scroll
- click action options:
    - **tour**: trigger a product tour
    - **link**: link to something

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
+++ b/packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
@@ -1,0 +1,100 @@
+import { h, ComponentChildren } from 'preact'
+import { ProductTourStep } from '../../../posthog-product-tours-types'
+import { getStepHtml } from '../product-tours-utils'
+import { cancelSVG } from '../../surveys/icons'
+
+export interface ProductTourBannerProps {
+    step: ProductTourStep
+    onDismiss: () => void
+    onTriggerTour?: () => void
+}
+
+interface BannerWrapperProps {
+    class: string
+    children: ComponentChildren
+}
+
+interface LinkWrapperProps extends BannerWrapperProps {
+    href: string
+}
+
+interface ButtonWrapperProps extends BannerWrapperProps {
+    onClick: () => void
+}
+
+function LinkWrapper({ class: className, href, children }: LinkWrapperProps): h.JSX.Element {
+    return (
+        <a class={className} href={href} target="_blank" rel="noopener noreferrer">
+            {children}
+        </a>
+    )
+}
+
+function ButtonWrapper({ class: className, onClick, children }: ButtonWrapperProps): h.JSX.Element {
+    return (
+        <div
+            class={className}
+            role="button"
+            tabIndex={0}
+            onClick={onClick}
+            onKeyDown={(e) => {
+                if (e.key === 'Enter' || e.key === ' ') {
+                    e.preventDefault()
+                    onClick()
+                }
+            }}
+            style={{ cursor: 'pointer' }}
+        >
+            {children}
+        </div>
+    )
+}
+
+function StaticWrapper({ class: className, children }: BannerWrapperProps): h.JSX.Element {
+    return <div class={className}>{children}</div>
+}
+
+export function ProductTourBanner({ step, onDismiss, onTriggerTour }: ProductTourBannerProps): h.JSX.Element {
+    const config = step.bannerConfig ?? { behavior: 'sticky' }
+    const action = config.action
+
+    const classNames = ['ph-tour-banner', config.behavior === 'sticky' && 'ph-tour-banner--sticky']
+        .filter(Boolean)
+        .join(' ')
+
+    const content = (
+        <>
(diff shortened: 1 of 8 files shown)


packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
```diff
@@ -0,0 +1,51 @@
+import { h } from 'preact'
```
lucasheriques: is this being used?

packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
```diff
@@ -0,0 +1,51 @@
...
+        if (target?.closest('a')) {
+            e.stopPropagation()
+        }
+    }
+
+    return (
+        <div
+            class={classNames}
+            onClick={hasClickAction ? onBannerClick : undefined}
+            style={{ cursor: hasClickAction ? 'pointer' : 'default' }}
```
lucasheriques: onClick on divs is not the best idea... is there no other element we can use? why are we not using buttons here? though a button shouldn't be a parent of another button

anyway wanted to confirm. maybe ask claude if there's something batter?

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -121,6 +122,37 @@ function retrieveTourShadow(tour: ProductTour): { shadow: ShadowRoot; isNewlyCre
     }
 }
 
+function retrieveBannerShadow(tour: ProductTour): { shadow: ShadowRoot; isNewlyCreated: boolean } {
```
lucasheriques, line 125: feel like this can prob be shared between survey an product tours in the future

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -719,6 +765,45 @@ export class ProductTourManager {
...
+        const handleBannerClick = () => {
+            const action = step.bannerConfig?.action
+            if (!action || action.type === 'none') {
+                return
+            }
+
+            switch (action.type) {
+                case 'link':
+                    if (action.link) {
+                        window.open(action.link, '_blank')
```
lucasheriques: i do believe that, if someone has popups blocked, this will also fail

would be great if we could change this to be an anchor link instead of a `window.open`

lucasheriques (approved):
code looks good, but i did have a couple of comments. pls take a look

packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
adboio: hmm yeah this is kinda ugly haha, i was movin too quick - i’ll take a look at some other options

packages/browser/src/extensions/product-tours/product-tours.tsx
adboio: good call this is now at least the 3rd shadow retriever

packages/browser/src/extensions/product-tours/product-tours.tsx
adboio: ahh good looks - i’ll see what we can do here. one potential issue is that the banner content might also have anchor links but maybe that just shouldn’t be allowed… i’ll play with the UX

adboio: i used the same window.open strategy for custom links in tour step buttons, so i'll review that as well

packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
adboio: it's used for the return value of the component function  


```
h.JSX.Element
```

packages/browser/src/extensions/product-tours/components/ProductTourBanner.tsx
adboio: @lucasheriques i did a few things here:

\- banners can no longer have inline anchors, so we can simplify their rendering  
\- if banner action is link, we wrap in \<a>  
\- if banner action is open tour, we wrap in \<div> but with role and tabIndex for accessibility (to avoid nested buttons with the dismiss)  
\- also updated the product tour steps to use \<a> for the "buttons" if their action is a link

no more window.open and should be accessible now with proper elements

adboio:
### Merge activity

* **Jan 14, 6:14 AM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/2865).

PR #2869, opened 2026-01-09
chore(replay): add SDK debug logging for full snapshot timestamp

ksvat:
## Problem

we're STILL seeing sessions with missing or delayed full snapshots
this will ideally let us actually size out the issue and identify patterns (hopefully)
## Changes

add debug logging on the SDK to track first snapshot timestamp

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [ ] Tests for new code

### If releasing new changes


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -951,6 +958,16 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+
+            // Track the timestamp of the first full snapshot for this session
+            // This helps us detect session rotation issues where incremental snapshots
+            // are sent before the full snapshot
+            const currentTimestamp = this._instance.persistence?.props[SESSION_RECORDING_FIRST_FULL_SNAPSHOT_TIMESTAMP]
+            if (isUndefined(currentTimestamp)) {
+                this._instance.persistence?.register({
+                    [SESSION_RECORDING_FIRST_FULL_SNAPSHOT_TIMESTAMP]: rawEvent.timestamp,
+                })
+            }
```
pauldambra: i always have to check what it means but we have `register_once` which _i htink_ only sets it if it isn't already set so you could save some bytes and use that here  
  
(if I'm not confused about register_once :)))

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -849,6 +853,9 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             this._flushedSizeTracker.reset()
         }
 
+        // Reset first full snapshot timestamp for the new session
```
pauldambra, line 856: great example of what i was just saying where i've added complexity with sticking plasters - we have multiple places to deal with session changing when _really_ i bet we could just have one

pauldambra: not to fix now... just to publicly shame myself

PR #2871, opened 2026-01-09
fix: lastactivity timestamp can be null

pauldambra:
## Problem

lastActivityTimestamp can be null
but we check if it is 0
they are not the same value
(i am good at identifying things like that)

and so when it is null
we don't catch it as being absent

i interrogated the robot, it explains

```
  The problem: lastActivityTimestamp === 0 checks for 0, but after reset it's null. So if the first call after reset is readOnly = true:

  1. noSessionId = true → new session created
  2. But lastActivityTimestamp === 0 is false (because null !== 0)
  3. And !readOnly is false
  4. So newActivityTimestamp = lastActivityTimestamp → null
  5. Persists [null, newSessionId, startTimestamp]

  Then the second call with readOnly = false:
  1. Reads lastActivityTimestamp = null from persistence
  2. _sessionHasBeenIdleTooLong(timestamp, null) → Math.abs(timestamp - null) → Math.abs(timestamp - 0) → timestamp (a huge number like 1736459939000)
  3. timestamp > 1800000 (30 min timeout) → true → activity timeout!

```

don't you love that JS will swap in a 0 there

## Changes

check if it is 0 _and_ check if it is null

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/retry-queue.ts
+++ b/packages/browser/src/retry-queue.ts
@@ -1,6 +1,6 @@
 import { RetriableRequestWithOptions } from './types'
 
-import { isNumber, isUndefined } from '@posthog/core'
+import { isPositiveNumber, isUndefined } from '@posthog/core'
 import { logger } from './utils/logger'
 import { window } from './utils/globals'
 import { PostHog } from './posthog-core'
@@ -68,7 +68,7 @@ export class RetryQueue {
     }
 
     retriableRequest({ retriesPerformedSoFar, ...options }: RetriableRequestWithOptions): void {
-        if (isNumber(retriesPerformedSoFar) && retriesPerformedSoFar > 0) {
+        if (isPositiveNumber(retriesPerformedSoFar)) {
             options.url = extendURLParams(options.url, { retry_count: retriesPerformedSoFar })
         }
 
--- a/packages/core/src/utils/type-utils.ts
+++ b/packages/core/src/utils/type-utils.ts
@@ -62,7 +62,12 @@ export const isNullish = (x: unknown): x is null | undefined => isUndefined(x) |
 
 export const isNumber = (x: unknown): x is number => {
   // eslint-disable-next-line posthog-js/no-direct-number-check
-  return toString.call(x) == '[object Number]'
+  // x !== x is true only for NaN (ES5-compatible NaN check)
+  return toString.call(x) == '[object Number]' && x === x
+}
+
+export const isPositiveNumber = (value: unknown): value is number => {
+  return isNumber(value) && value > 0
 }
 
 export const isBoolean = (x: unknown): x is boolean => {
--- a/packages/browser/src/sessionid.ts
+++ b/packages/browser/src/sessionid.ts
@@ -7,7 +7,7 @@ import { window } from './utils/globals'
 
 import { createLogger } from './utils/logger'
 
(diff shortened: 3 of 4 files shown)


pauldambra:
* **#2871** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2871?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/2871?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/playwright/mocked/session-recording/lazy-session-recording.spec.ts
```diff
@@ -233,23 +233,44 @@ test.describe('Session recording - array.js', () => {
...
 
+        // First request may flush old session buffer
         await page.waitingForNetworkCausedBy({
             urlPatternsToWaitFor: ['**/ses/*'],
             action: async () => {
                 await page.locator('[data-cy-input]').fill('hello posthog!')
             },
         })
 
+        // Second request gets new session snapshot
```
pauldambra, line 244: previously the double reset caused this all to flush  
now we need to keep acting to force it to flush

packages/browser/src/retry-queue.ts
```diff
@@ -68,7 +68,7 @@ export class RetryQueue {
     }
 
     retriableRequest({ retriesPerformedSoFar, ...options }: RetriableRequestWithOptions): void {
-        if (isNumber(retriesPerformedSoFar) && retriesPerformedSoFar > 0) {
+        if (isPositiveNumber(retriesPerformedSoFar)) {
```
pauldambra, line 71: we had this positive check pattern in a few places so i thought i'd squash while changing

packages/core/src/utils/type-utils.ts
```diff
@@ -62,8 +62,14 @@ export const isNullish = (x: unknown): x is null | undefined => isUndefined(x) |
 
 export const isNumber = (x: unknown): x is number => {
   // eslint-disable-next-line posthog-js/no-direct-number-check
-  return toString.call(x) == '[object Number]'
+  // x !== x is true only for NaN (ES5-compatible NaN check)
```
pauldambra, line 65: is number would match for NaN  
and while that might be right for JS

humans don't expect it  
so let's exclude NaN here

packages/core/src/utils/type-utils.ts
```diff
@@ -62,8 +62,14 @@ export const isNullish = (x: unknown): x is null | undefined => isUndefined(x) |
...
+  return toString.call(x) == '[object Number]' && x === x
 }
+
+export const isPositiveNumber = (value: unknown): value is number => {
```
pauldambra, line 69: we do this in a few places so will probably squash a few bytes away doing it here

pauldambra: we did... it removed 20B from the bundle #GameChanging

PR #2883, opened 2026-01-12
feat(node): warn about experience continuity flags and add strictLocalEvaluation

ordehi:
## Problem

re: https://posthog.slack.com/archives/C09LVLPSBJA/p1767983965754459

Local eval + experience continuity is incompatible, so those flags are always fetched on the server unless `onlyEvaluateLocally: true` is set on every evaluation call. However, this is opaque and users in distributed environments may see increased costs for something easy to setup and not so easy to figure out.

## Changes

- Add warning when exp cont flags detected during local evaluation
- Warning explains these flags cause server requests, negating local eval savings
- Warning is suppressed when strictLocalEvaluation is enabled
- New strictLocalEvaluation option prevents all server fallback
- Flags that can't be evaluated locally return undefined with strict mode

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/__tests__/feature-flags.spec.ts
+++ b/packages/node/src/__tests__/feature-flags.spec.ts
@@ -5766,6 +5766,7 @@ describe('strictLocalEvaluation option', () => {
   jest.useFakeTimers()
 
   beforeEach(() => {
+    mockedFetch.mockClear()
     warnSpy = jest.spyOn(console, 'warn').mockImplementation()
   })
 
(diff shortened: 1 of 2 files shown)


haacked (approved):
Nice! No notes.

Issue #2892, opened 2026-01-13
chore: Load config from RemoteConfig endpoints by default


haacked:
The `__preview_remote_config` flag is currently `false` by default, causing posthog-js to load SDK configuration via `/flags?v=2&config=true` instead of the dedicated RemoteConfig endpoints (`/array/{token}/config`).

This creates tight coupling between config loading and the Rust flags service, where:
- Config changes risk breaking flag evaluation
- The flags service rebuilds config instead of serving it from the pre-computed hypercache
- Changes to config shape require coordinated updates across Python and Rust

Once [Phase 1 of posthog#42417](https://github.com/PostHog/posthog/issues/42417) is complete (Rust passthrough from hypercache), we should enable `__preview_remote_config` by default to:
1. Load config from `/array/{token}/config` (Python/hypercache - single source of truth)
2. Load flags from `/flags?v=2` without the `config` param (Rust - flags only)

## End State

| Endpoint | Purpose | Characteristics |
|----------|---------|-----------------|
| `/array/<token>/config` | Config (sessionRecording, surveys, heatmaps, error tracking, etc.) | Cacheable, CDN-friendly, SDK fetches on init |
| `/flags` | Flag evaluation | Dynamic, per-request evaluation, returns ONLY `{ flags, errorsWhileComputingFlags, requestId }` |

- `RemoteConfig.build_config()` becomes THE single source of truth for config
- Rust never processes Team config fields – it only evaluates flags
- Clear separation: config changes don't risk breaking flag evaluation, and vice versa

Remove `__preview_remote_config` flag entirely once rollout is complete.

## Known Complexities

- **Linked flags in config**: There's tricky coupling where config returns linked flags that needs to be handled
- **Semi-dynamic replay content**: Authorized URLs feature (90% deprecated, only deletions allowed now, no new additions) - may need forced migration
- **Multiple SDKs**: Changes need to be coordinated across various SDKs

## Future Ideal State

Serve JS with config baked in, eliminating a fetch call entirely – if someone isn't using flags, there's no fetch needed.

## Related sub-libraries

- [x] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

[...]

PR #2895, opened 2026-01-14
feat: Only use remote config

benjackwhite:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/2892

RemoteConfig (config.js) has been loaded for ages and is in use by us in production. This PR makes it the sole config loading mechanism, removing the legacy `/flags/?v=2&config=true` path and the `__preview_remote_config` gate.

### Core

- **`remote-config.ts`**: Add 5-minute visibility-aware refresh interval that pauses when the tab is backgrounded. `refresh()` only reloads flags (not the full `_onRemoteConfig` pipeline) to avoid SiteApps binding duplicate event listeners. Add `stop()` for cleanup on `reset()`.
- **`posthog-featureflags.ts`**: Remove the `flags()` method and `_flagsCalled` pattern. `_callFlagsEndpoint` always uses `/flags/?v=2` and always includes `timezone`.
- **`posthog-core.ts`**: Store `RemoteConfigLoader` instance so it can be stopped on `reset()`. Fix `process_person` config alias preservation through `_onRemoteConfig`.
- **`posthog-config.ts`**: Remove `__preview_remote_config` type.

### Tests

- Updated all test files to account for the new architecture (RemoteConfig always-on, debounced flag loading, timezone always included)
- Added tests for refresh interval, visibility-aware polling, stop/cleanup, and `advanced_disable_feature_flags_on_first_load`

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/strict-streets-sing.md
+++ b/.changeset/strict-streets-sing.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+RemoteConfig (config.js) has been loaded for ages and is in use by us in production. This PR makes it the sole config loading mechanism for posthog-js, removing the legacy /flags/?v=2&config=true path and the \_\_preview_remote_config gate.


packages/browser/src/remote-config.ts
```diff
@@ -56,26 +61,53 @@ export class RemoteConfigLoader {
...
+        if (this._refreshInterval) {
+            return
+        }
+
+        // Don't start interval in test environments to avoid issues with fake timers
+        if (typeof process !== 'undefined' && process.env.NODE_ENV === 'test') {
+            return
+        }
+
+        this._refreshInterval = setInterval(() => {
```
haacked: Currently the interval continues polling every 5 minutes even when the tab is backgrounded, which wastes bandwidth and battery (especially on mobile). Consider pausing the refresh interval when the tab is not visible following the pattern in `heatmaps.ts`:

```ts
// Setting up listeners
this._onVisibilityChange_handler = this._onVisibilityChange.bind(this)
addEventListener(document, 'visibilitychange', this._onVisibilityChange_handler)

// Handling visibility changes
private _onVisibilityChange(): void {
    // always clear the interval just in case
    if (this._flushInterval) {
        clearInterval(this._flushInterval)
    }

    this._flushInterval = shouldPoll(document)
        ? setInterval(this._flush.bind(this), this.flushIntervalMilliseconds)
        : null
}
```

That's such a useful pattern we might even consider generalizing it, but for this PR that's unnecessary.

packages/browser/src/remote-config.ts
```diff
@@ -56,26 +61,53 @@ export class RemoteConfigLoader {
...
-            logger.error('Failed to fetch remote config from PostHog.')
+    private _startRefreshInterval(): void {
+        if (this._refreshInterval) {
+            return
+        }
+
+        // Don't start interval in test environments to avoid issues with fake timers
+        if (typeof process !== 'undefined' && process.env.NODE_ENV === 'test') {
+            return
+        }
```
haacked: This pattern skips the interval entirely in tests, making the refresh behavior untestable. It'd be nice to have tests of the refresh. We could follow the pattern in `heatmaps.ts` which creates its interval normally and lets Jest's fake timers (`jest.useFakeTimers()` / `jest.advanceTimersByTime()`) control when it fires.

haacked (commented):
Excited to see this happening!

Most of my comments are non-blocking, except maybe the comment on visibility.

The AI comment about `advanced_disable_feature_flags_on_first_load` is important as that appears to be a breaking change.

packages/browser/src/__tests__/featureflags.test.ts
```diff
@@ -1159,7 +1135,7 @@ describe('featureflags', () => {
             featureFlags.reloadFeatureFlags()
             jest.runAllTimers()
             // check the request sent person properties
-            expect(instance._send_request.mock.calls[0][0].data).toEqual({
+            expect(instance._send_request.mock.calls[0][0].data).toMatchObject({
```
dustinbyrne: nit: consider leaving `toEqual` in place and adding the new timezone property instead:
```ts
timezone: expect.any(String)
```

packages/browser/src/remote-config.ts
```diff
@@ -56,26 +61,53 @@ export class RemoteConfigLoader {
...
+    refresh(): void {
+        if (this._instance._shouldDisableFlags()) {
             return
         }
 
-        if (!this._instance.config.__preview_remote_config) {
-            logger.info('__preview_remote_config is disabled. Logging config instead', config)
+        this._loadRemoteConfigJSON((config) => {
+            if (config) {
+                this._instance._onRemoteConfig(config)
```
dustinbyrne: `SiteApps` looks like it it'll bind a new event listener every time this is called. This seems worth fixing.

dustinbyrne: I'd say it's out of scope for this PR, but we'll need to ensure that `onRemoteConfig` listeners properly handle config updates.

packages/browser/src/remote-config.ts
```diff
@@ -56,36 +65,103 @@ export class RemoteConfigLoader {
...
+        // Config and flags are loaded separately: config from /array/{token}/config,
+        // flags from /flags/?v=2. Features like surveys, session recording, and product
+        // tours reference flags in their config (e.g. survey.linked_flag_key), but this
+        // is safe because those flag checks happen lazily at runtime (e.g. when deciding
+        // whether to show a survey), not during config processing. By the time a linked
+        // flag is evaluated, flags have already loaded.
+        //
+        // Even when config fails, we pass an empty object so extensions (autocapture,
+        // session recording, etc.) still initialize with their defaults.
+        this._instance._onRemoteConfig(config ?? ({} as RemoteConfig))
```
haacked, line 124: When config fetch fails, passing `{}` causes exception-autocapture and autocapture to overwrite their persisted values with `false`/`undefined`. These two extensions are missing the `'in'` guard added to dead-clicks, web-vitals, heatmaps, error-tracking, and product-tours.

Should add:
- `if (!('autocaptureExceptions' in response)) return` in exception-autocapture's `onRemoteConfig`
- `if (!('autocapture_opt_out' in response)) return` in autocapture's `onRemoteConfig`

packages/browser/src/remote-config.ts
```diff
@@ -56,36 +65,103 @@ export class RemoteConfigLoader {
...
 
-        if (!this._instance.config.__preview_remote_config) {
-            logger.info('__preview_remote_config is disabled. Logging config instead', config)
+        this._lastRefreshTimestamp = Date.now()
+        // reloadFeatureFlags() debounces internally, so rapid calls from tab
+        // switching or overlapping intervals are safe.
+        this._instance.featureFlags.reloadFeatureFlags()
+    }
+
+    private _startRefreshInterval(): void {
```
haacked, line 100: Consider simplifying the visibility handling. The pause/resume interval pattern is more complex than similar patterns in this codebase. A simpler approach keeps the interval running and checks visibility in `refresh()`:

```typescript
refresh(): void {
    if (this._instance._shouldDisableFlags() || document?.visibilityState === 'hidden') {
        return
    }
    this._lastRefreshTimestamp = Date.now()
    this._instance.featureFlags.reloadFeatureFlags()
}
```

This eliminates the handler, event listener, and timestamp tracking at the cost of a no-op timer fire every 5 minutes in background tabs.

packages/browser/src/remote-config.ts
```diff
@@ -3,10 +3,19 @@ import { RemoteConfig } from './types'
 
 import { createLogger } from './utils/logger'
 import { assignableWindow } from './utils/globals'
+import { document } from './utils/globals'
+import { addEventListener } from './utils'
 
 const logger = createLogger('[RemoteConfig]')
 
+// 5 minutes in milliseconds
+const REFRESH_INTERVAL = 5 * 60 * 1000
```
haacked, line 12: nit: Consider documenting why 5 minutes was chosen (server load, UX research, battery considerations?) and whether this should be adaptive (e.g., longer on metered connections via `navigator.connection`).

dmarticus:
> I meant to approve it. I do think the one critical issue I found should be fixed first. But otherwise this looks great!

Thanks for the review! I addressed the other feedback (exception-autocapture guard, simplified visibility handling, documented the 5-minute interval), but I intentionally left `autocapture_opt_out` without the guard.

For autocapture, the missing `autocapture_opt_out` key being treated as "enabled" is intentional. The behavior is:

1. Before remote config loads, autocapture is disabled by default (waiting for server confirmation) - see `isEnabled` at autocapture.ts:340-342
2. When remote config arrives, `autocapture_opt_out: undefined → !!undefined = false` → autocapture enabled

This is a "wait for server, then enable unless explicitly opted out" model. The SDK intentionally writes false to persistence when the key is missing, which enables autocapture after confirming the server didn't explicitly disable it. This is different from features like heatmaps where the server must explicitly send `heatmaps: true` to enable.

Added a comment to this end.

packages/browser/src/remote-config.ts
joaompneves: is it possible to make this configurable? 5 minutes seems too aggressive. It should be possible to specify a value other than 5 min, or even disable this periodic refresh at all.

dmarticus: hi @joaompneves 

good take; I made this PR to make that change: https://github.com/PostHog/posthog-js/pull/3145

Issue #2879, opened 2026-01-12
Add `$feature_flags_error` to React-Native SDK


dustinbyrne:
**Goal**: Port flag error tracking to React Native by including the `$feature_flag_error` property in `$feature_flag_called` events. This property exists to help debug flag evaluation failures.

The value of the property reflects any error(s) that occurred while retrieving remote evaluation results from the `/flags` endpoint.

Other SDKs report properties such as:
- `errors_while_computing_flags` - Server response body JSON included `errorsWhileComputingFlags=true`
- `flag_missing` - Requested flag not in API response
- `quota_limited` - Rate/quota limit exceeded
- `timeout` - Request timed out
- `connection_error` - Network connectivity issue
- `api_error_{status}` - Server error with HTTP status code (e.g., `api_error_500`)
- `unknown_error` - Unexpected exceptions

Multiple relevant errors are joined with commas (e.g., `errors_while_computing_flags,flag_missing`).

This feature has been implemented in a number of SDKs already for reference:
- Node SDK (most relevant): https://github.com/PostHog/posthog-js/pull/2782
- Python SDK: https://github.com/PostHog/posthog-python/pull/390
- Kotlin SDK: https://github.com/PostHog/posthog-android/pull/355
- DotNet SDK: https://github.com/PostHog/posthog-dotnet/pull/144
- Go SDK: https://github.com/PostHog/posthog-go/pull/142
- PHP SDK: https://github.com/PostHog/posthog-php/pull/94
- Ruby SDK: https://github.com/PostHog/posthog-ruby/pull/86

Related to https://posthoghelp.zendesk.com/agent/tickets/46332

PR #2897, opened 2026-01-14
feat(flags): Add $feature_flag_error property to track flag evaluation failures on react-native

matheus-vb:
## Problem

When feature flag evaluations fail in React Native applications, developers have no visibility into what went wrong. The `$feature_flag_called` event is captured but contains no error information, making it difficult to debug issues related to quota limits, server errors, missing flags, or network failures.

This capability already exists in other PostHog SDKs (Node, Python, Kotlin, Go) and is tracked in issue #2879.

## Changes

Adds the `$feature_flag_error` property to `$feature_flag_called` events in the React Native SDK to track flag evaluation failures. The implementation lives in `@posthog/core`.

Error types tracked:
  - `errors_while_computing_flags`: Server indicated computation failures
  - `flag_missing`: Requested flag not present in API response
  - `quota_limited`: Feature flags quota limit exceeded
  - `timeout`: Request timed out (AbortError/TimeoutError)
  - `connection_error`: Network connection failed
  - `api_error_{status}`: HTTP error with specific status code (e.g., api_error_500)
  - `unknown_error`: Unrecognized error type

Multiple errors are joined with commas (e.g., `quota_limited,flag_missing`).

Additional changes:
- Extended `PostHogFlagsStorageFormat` to persist error state across flag evaluations
- Consolidated `FeatureFlagError` constants from `posthog-node` to `@posthog/core` for reuse
- Re-exported `FeatureFlagError` from `posthog-node` to maintain backwards compatibility
- Preserved cached flags when quota limited instead of clearing them

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/core/src/posthog-core.ts
```diff
@@ -574,10 +575,20 @@ export abstract class PostHogCore extends PostHogCoreStateless {
           extraProperties,
           fetchConfig
         )
-        // Add check for quota limitation on feature flags
+
+        if (res === undefined) {
+          this.setKnownFeatureFlagDetails({
+            flags: this.getKnownFeatureFlagDetails()?.flags ?? {},
+            requestFailed: true,
```
dustinbyrne: `super.getFlags()` above returns `undefined` on error, but the information we would need to determine `connection_error`, `timeout`, `api_error_xxx` are available within that method (raised by `fetchWithRetry`).

If we change the return type, we could get that information out.

packages/core/src/posthog-core.ts
```diff
@@ -574,10 +575,20 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+          return undefined
+        }
+
         if (res?.quotaLimited?.includes(QuotaLimitedFeature.FeatureFlags)) {
-          // Unset all feature flags by setting to null
-          this.setKnownFeatureFlagDetails(null)
+          this.setKnownFeatureFlagDetails({
+            flags: {},
+            quotaLimited: res.quotaLimited,
+          })
```
dustinbyrne, line 593: I realize this is derived from the previous behavior, but I'm not actually sure why we clear flags when quota limited - the client likely still has some valid flags in cache.

```suggestion
          this.setKnownFeatureFlagDetails({
            flags: this.getKnownFeatureFlagDetails()?.flags ?? {},
            quotaLimited: res.quotaLimited,
          })
```

packages/core/src/posthog-core.ts
```diff
@@ -647,7 +664,19 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+  }
+
+  private isFeatureFlagsQuotaLimited(): boolean {
+    return this.getStoredFlagDetails()?.quotaLimited?.includes(QuotaLimitedFeature.FeatureFlags) ?? false
+  }
+
   protected getKnownFeatureFlags(): PostHogFlagsResponse['featureFlags'] | undefined {
+    if (this.isFeatureFlagsQuotaLimited()) {
+      return undefined
+    }
```
dustinbyrne: We don't need this guard when a flag is requested. If we do the right thing when we parse the quota limited response, the cache should always be valid.

dustinbyrne: This goes for the other similar guard clauses as well

packages/core/src/posthog-core.ts
dustinbyrne: This method isn't used anywhere, we should just drop it

packages/core/src/posthog-core.ts
```diff
@@ -752,12 +809,32 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+
+    const storedDetails = this.getStoredFlagDetails()
+    const hasCachedFlags = storedDetails?.flags && Object.keys(storedDetails.flags).length > 0
+    if (storedDetails?.requestFailed && !hasCachedFlags) {
+      return undefined
+    }
+
```
dustinbyrne: This is another case where we should just return the cached data. If we don't have cached data, it'll return undefined anyway.

packages/core/src/posthog-core.ts
```diff
@@ -15,6 +15,7 @@ import type {
   Survey,
   SurveyResponse,
   PostHogGroupProperties,
+  FeatureFlagErrorType,
```
dustinbyrne: This import isn't used

packages/node/src/exports.ts
```diff
@@ -1,3 +1,5 @@
 export * from './extensions/sentry-integration'
 export * from './extensions/express'
 export * from './types'
+export { FeatureFlagError } from '@posthog/core'
+export type { FeatureFlagErrorType } from '@posthog/core'
```
dustinbyrne, line 8: These two are probably worth a comment.

packages/core/src/types.ts
```diff
@@ -255,7 +255,12 @@ export type PostHogV2FlagsResponse = Omit<PostHogFlagsResponse, 'featureFlags' |
  * When we pull flags from persistence, we can normalize them to PostHogFeatureFlagDetails
  * so that we can support v1 and v2 of the API.
  */
-export type PostHogFlagsStorageFormat = Pick<PostHogFeatureFlagDetails, 'flags'>
+export type PostHogFlagsStorageFormat = Pick<PostHogFeatureFlagDetails, 'flags'> &
+  Partial<Pick<PostHogFlagsResponse, 'requestId' | 'evaluatedAt'>> & {
+    errorsWhileComputingFlags?: boolean
+    quotaLimited?: string[]
+    requestFailed?: boolean
```
dustinbyrne: If possible I'd like to see this boolean expand to cover the other request error types like `connection_error`, `timeout`, `api_error_{status}`.

dustinbyrne (changes requested):
This is looking good! I've left some comments below. Most of them are related to preserving and serving cached values. In general what we write to cache should be retrievable without much (if any) revalidation when we read. It's better for users to receive a stale flag value than to see a value change due to a transient error (billing or otherwise).

.changeset/frank-beds-search.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog-react-native': minor
+'@posthog/core': minor
+---
```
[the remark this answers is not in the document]
dustinbyrne: I don't think this is actually an issue, it's more of a refactor in `posthog-node`. I wouldn't be opposed to adding it though.

packages/core/src/posthog-core.ts
matheus-vb: Nice, agreed. I changed the behavior to preserve the cached flags upon quota limited errors. I noticed you created an issue to standardize the behavior across SDKs, but I guess this can be done here already

packages/core/src/types.ts
matheus-vb: What do you think about this? It would broaden a bit the scope of the PR. I pushed the code here but I can also move this improved error handling to a followup PR, since it ends up changing bits of the stateless client as well.

```typescript
export const FeatureFlagError = {
  ERRORS_WHILE_COMPUTING: 'errors_while_computing_flags',
  FLAG_MISSING: 'flag_missing',
  QUOTA_LIMITED: 'quota_limited',
  TIMEOUT: 'timeout',
  CONNECTION_ERROR: 'connection_error',
  /** @deprecated Use TIMEOUT, CONNECTION_ERROR, or apiError() for granular error types */
  UNKNOWN_ERROR: 'unknown_error',
  apiError: (status: number): string => `api_error_${status}`,
} as const
```

packages/core/src/posthog-core-stateless.ts
```diff
@@ -482,11 +484,29 @@ export abstract class PostHogCoreStateless {
...
+      const cause = error.error
+      // AbortError is thrown when the request times out via AbortSignal.timeout()
+      if (cause instanceof Error && cause.name === 'AbortError') {
+        return { type: 'timeout' }
+      }
+      return { type: 'connection_error' }
+    }
+
+    // Unknown error type, default to connection_error
+    return { type: 'connection_error' }
```
dustinbyrne: If we get here, presumably this is a case for `unknown_error`?

packages/core/src/posthog-core-stateless.ts
```diff
@@ -482,11 +484,29 @@ export abstract class PostHogCoreStateless {
...
+
+  private categorizeRequestError(error: unknown): FeatureFlagRequestError {
+    if (error instanceof PostHogFetchHttpError) {
+      return { type: 'api_error', statusCode: error.status }
+    }
+
+    if (error instanceof PostHogFetchNetworkError) {
+      const cause = error.error
+      // AbortError is thrown when the request times out via AbortSignal.timeout()
+      if (cause instanceof Error && cause.name === 'AbortError') {
```
dustinbyrne: I think `AbortError` might be used in older versions of Node, but we should also be checking for `TimeoutError`.
```suggestion
      if (cause instanceof Error && (cause.name === 'AbortError' || cause.name === 'TimeoutError')) {
```

```js
> x = AbortSignal.timeout(0)
> x.throwIfAborted()
Uncaught:
DOMException [TimeoutError]: The operation was aborted due to timeout
```

packages/core/src/types.ts
```diff
@@ -268,6 +273,52 @@ export type JsonType = string | number | boolean | null | { [key: string]: JsonT
...
+ *   CONNECTION_ERROR: Network connection failed
+ *   apiError: HTTP error with status code (e.g., api_error_500)
+ */
+export const FeatureFlagError = {
+  ERRORS_WHILE_COMPUTING: 'errors_while_computing_flags',
+  FLAG_MISSING: 'flag_missing',
+  QUOTA_LIMITED: 'quota_limited',
+  TIMEOUT: 'timeout',
+  CONNECTION_ERROR: 'connection_error',
+  /** @deprecated Use TIMEOUT, CONNECTION_ERROR, or apiError() for granular error types */
```
dustinbyrne: If we use this in `categorizeRequestError`, we can drop this comment

ablaszkiewicz (approved):

packages/nuxt/src/module.ts
```diff
@@ -1,6 +1,7 @@
 import { defineNuxtModule, addPlugin, createResolver, addServerPlugin, addImportsDir } from '@nuxt/kit'
 import type { PostHogConfig } from 'posthog-js'
 import type { PostHogOptions } from 'posthog-node'
+import type {} from 'nuxt/app'
```
ablaszkiewicz, line 4: Can you explain what was failing and why we need that?

matheus-vb: Sure! The Vercel posthog-nextjs-config was failing with a weird import error. There was also a TS2742 warning, TypeScript can't portably infer defineNuxtModule's return type without explicitly referencing @nuxt/schema. The empty import forces TypeScript to include the type reference ([source](https://github.com/nuxt/module-builder/issues/141#issuecomment-2078248248)). For some weird reason, with this fixed the job passes. The import gets erased at compile time, so this should only silence the warning.

PR #2911, opened 2026-01-15
fix: Fix $variables being inside single quotes and not replacing

rafaeelaudibert:
This was added recently but variables aren't being interpolated properly because they're inside single quotes.
Fix is a bit chaotic and using `jq` but it makes sense


rafaeelaudibert:
### Merge activity

* **Jan 15, 8:58 PM UTC**: @rafaeelaudibert merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/2911).

PR #2915, opened 2026-01-16
feat(llma): add time to first token reporting

richardsolomou:
## Problem

Users can see total duration, but not how quickly a model begins responding. This makes it harder to diagnose perceived slowness, compare providers/models, and optimize user experience where early tokens matter more than full completion time.

## Changes

- Added time to first token (TTFT) measurement to LLM analytics.
- TTFT is captured when the first token is received from the provider and reported alongside existing timing metrics.
- Only for streaming responses, non-streaming responses just use latency

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/ai/src/openai/azure.ts
+++ b/packages/ai/src/openai/azure.ts
@@ -148,6 +148,9 @@ export class WrappedCompletions extends AzureOpenAI.Chat.Completions {
                 // Handle tool calls
                 const deltaToolCalls = choice?.delta?.tool_calls
                 if (deltaToolCalls && Array.isArray(deltaToolCalls)) {
+                  if (firstTokenTime === undefined) {
+                    firstTokenTime = Date.now()
+                  }
                   for (const toolCall of deltaToolCalls) {
                     const index = toolCall.index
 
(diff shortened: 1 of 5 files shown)


packages/ai/src/openai/azure.ts
```diff
@@ -371,6 +378,21 @@ export class WrappedResponses extends AzureOpenAI.Responses {
...
+                  firstTokenTime === undefined &&
+                  (chunk.type === 'response.output_item.added' ||
+                    chunk.type === 'response.content_part.added' ||
+                    chunk.type === 'response.output_text.delta' ||
+                    chunk.type === 'response.reasoning_text.delta' ||
+                    chunk.type === 'response.reasoning_summary_text.delta' ||
+                    chunk.type === 'response.audio.delta' ||
+                    chunk.type === 'response.audio.transcript.delta' ||
+                    chunk.type === 'response.refusal.delta')
+                ) {
```
carlos-marchal-ph: We also do the exact same check for the vanilla OpenAI implementation, could we abstract it to a helper like `isTokenChunk`?

packages/ai/tests/vercel.test.ts
carlos-marchal-ph, line 1: We should probably add tests for V2 too right?

Issue #2160, opened 2024-06-17
Add support for `person_profiles` in react-native-posthog


itsramiel:
## Is your feature request related to a problem?
I would like to be able to use the `person_profiles` property to set as `identified_only` in my react native app

## Describe the solution you'd like
To add support for the `person_profiles` in the react native packages

## Describe alternatives you've considered
None

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-react-native

slshults:
+1 https://posthoghelp.zendesk.com/agent/tickets/24073 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/25629)

Issue #2147, opened 2024-09-11
Add `$is_identified`


marandaneto:
### Description

Android https://github.com/PostHog/posthog-android/issues/136
iOS https://github.com/PostHog/posthog-ios/issues/172

PR #2917, opened 2026-01-16
chore: person profiles rn, core and web lite

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/2160
Closes https://github.com/PostHog/posthog-js/issues/2147

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -2377,10 +2377,6 @@ export class PostHog implements PostHogInterface {
             return
         }
 
-        if (!this._requirePersonProcessing('posthog.group')) {
-            return
-        }
-
         const existingGroups = this.getGroups()
 
         // if group key changes, remove stored group properties
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -395,10 +395,6 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     options?: PostHogCaptureOptions
   ): void {
     this.wrap(() => {
-      if (!this._requirePersonProcessing('posthog.group')) {
-        return
-      }
-
       this.groups({
         [groupType]: groupKey,
       })
--- a/.changeset/seven-pigs-battle.md
+++ b/.changeset/seven-pigs-battle.md
@@ -1,5 +1,0 @@
----
-'posthog-js': patch
----
-
-fix: posthog.group won't capture a $groupidentify event if person profiles is set to never


packages/core/src/posthog-core.ts
```diff
@@ -364,6 +389,9 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     options?: PostHogCaptureOptions
   ): void {
     this.wrap(() => {
+      // Enable person processing when groups are used (if allowed by config)
+      this._requirePersonProcessing('posthog.group')
```
ioannisj: Don't we need to bail out here when false? We do on [iOS](https://github.com/PostHog/posthog-ios/blob/a01310fcb59809da6cad553359d80c3b49067892/PostHog/PostHogSDK.swift#L885)

packages/core/src/posthog-core.ts
```diff
@@ -950,4 +978,124 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+   * Returns whether the current user is identified (has a person profile).
+   *
+   * This checks:
+   * 1. If PersonMode is explicitly set to 'identified'
+   * 2. For backwards compatibility: if DistinctId differs from AnonymousId
+   *    (meaning the user was identified before the SDK was upgraded)
+   *
+   * @internal
+   */
+  protected _isIdentified(): boolean {
```
ioannisj, line 1044: I think we are missing the $is_identified event prop on capture()? [Web SDK](https://github.com/PostHog/posthog-js/blob/f0cbc0d8e4e5efc27d9595676e886d6d3d3892f4/packages/browser/src/posthog-core.ts#L1362) and [iOS](https://github.com/PostHog/posthog-ios/blob/a01310fcb59809da6cad553359d80c3b49067892/PostHog/PostHogSDK.swift#L293) include it

packages/node/example.mjs
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+/* eslint-disable no-undef */
```
marandaneto, line 2: process was not defined and linter didnt allow me to commit

packages/core/src/posthog-core.ts
marandaneto: mm good question, JS SDK does not do for groups, but does for setGroupPropertiesForFlags and groups call setGroupPropertiesForFlags

packages/core/src/posthog-core.ts
marandaneto: good point, forgot this, will add

packages/core/src/posthog-core.ts
marandaneto: https://posthog.slack.com/archives/C0368RPHLQH/p1768920312303459

ioannisj (approved):
I think it lg. Let's wait for an answer to https://github.com/PostHog/posthog-js/pull/2917#discussion_r2708670756

packages/browser/src/posthog-core.ts
```diff
@@ -2377,6 +2377,10 @@ export class PostHog implements PostHogInterface {
             return
         }
 
+        if (!this._requirePersonProcessing('posthog.group')) {
```
marandaneto: https://posthog.slack.com/archives/C0368RPHLQH/p1769180658914219?thread_ts=1768920312.303459&cid=C0368RPHLQH

packages/core/src/posthog-core.ts
marandaneto: https://github.com/PostHog/posthog-js/pull/2725#discussion_r2609598773
trying to understand the reason

marandaneto: i will rollback the change and remove the check from groups to unblock this PR and later add here or remove from the other SDKs if needed

PR #2918, opened 2026-01-16
Add `upload` option to sourcemaps config

joeriddles:
## Problem

People who want to upload source maps to PostHog in a separate CI workflow than their build process.

Source map uploading is slow and can be ran parallel in a separate workflow.

## Changes

Add `upload` option to the Next.js/Rollup/Webpack `sourcemaps` config. When false, `posthog-cli sourcemap inject` is used. When true (default), `@posthog/cli sourcemap process` is used, which injects and uploads.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [x] @posthog/rollup-plugin
- [x] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code
    - There are no existing tests for any of the modified packages
    - Default option is to upload sourcemaps

### If releasing new changes

### DIFF
--- a/packages/nextjs-config/src/utils.ts
+++ b/packages/nextjs-config/src/utils.ts
@@ -24,7 +24,6 @@ export async function processSourceMaps(posthogOptions: ResolvedPluginConfig, di
     cliOptions.push('sourcemap', 'inject')
   }
 
-  cliOptions.push('sourcemap', 'process')
   cliOptions.push('--directory', directory)
 
   if (posthogOptions.sourcemaps.project) {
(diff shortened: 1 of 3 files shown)


packages/nextjs-config/src/utils.ts
```diff
@@ -15,7 +15,15 @@ export function hasCompilerHook(): boolean {
```
hpouillot, line 15: @joeriddles Just to be sure to understand the use case, do you want to build once and upload sourcemaps in multiple environments in parallel ? What is the workflow you want to achieve ?

packages/nextjs-config/src/utils.ts
hpouillot, line 1: Do you have any errors when you run 'pnpm install' in the workspace ?

PR #2920, opened 2026-01-16
feat(flags): Add `getFeatureFlagResult` method for efficient flag + payload retrieval

haacked:
## Problem

The `posthog-node` SDK doesn't have a method that returns both the flag value AND payload in a single call. We have [one in `posthog-python`](https://github.com/PostHog/posthog-python/pull/284). We should do the same here.

### Why this matters

Currently, if you need both a flag's value and its payload, you must make two separate calls:

```typescript
const flagValue = await client.getFeatureFlag('my-flag', distinctId)
const payload = await client.getFeatureFlagPayload('my-flag', distinctId, flagValue)
```

This has several problems:

1. **Double evaluation overhead**: The flag is evaluated twice (once for value, once for payload lookup), which is wasteful especially for complex flags with dependencies
2. **Double network calls for remote evaluation**: When local evaluation isn't available, this results in two separate `/flags` requests
3. **Race conditions**: The flag value could theoretically change between the two calls
4. **Awkward API**: Users must pass `flagValue` to `getFeatureFlagPayload` to avoid yet another evaluation, which is unintuitive

With the new `getFeatureFlagResult` method, users can get everything in one call:

```typescript
const result = await client.getFeatureFlagResult('my-flag', distinctId)
if (result) {
  console.log(result.enabled)  // boolean
  console.log(result.variant)  // string | undefined  
  console.log(result.payload)  // JsonType | undefined (already parsed)
}
```

## Changes

- Add new `FeatureFlagResult` type with `key`, `enabled`, `variant`, and `payload` fields
- Add `getFeatureFlagResult` method to the PostHog client that:
  - Checks for flag overrides first
  - Tries local evaluation when personal API key is configured
  - Falls back to remote evaluation via `/decide`
  - Returns the payload already parsed (not stringified)
  - Captures `$feature_flag_called` event (respects `sendFeatureFlagEvents` option)
- Add comprehensive tests for the new method

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -619,8 +619,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
       personProperties?: Record<string, string>
       groupProperties?: Record<string, Record<string, string>>
       onlyEvaluateLocally?: boolean
-      disableGeoip?: boolean
       sendFeatureFlagEvents?: boolean
+      disableGeoip?: boolean
     } = {},
     matchValue?: FeatureFlagValue
   ): Promise<FeatureFlagResult | undefined> {


packages/node/src/client.ts
```diff
@@ -600,65 +602,47 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+    options: {
       groups?: Record<string, string>
       personProperties?: Record<string, string>
       groupProperties?: Record<string, Record<string, string>>
       onlyEvaluateLocally?: boolean
-      sendFeatureFlagEvents?: boolean
       disableGeoip?: boolean
-    }
-  ): Promise<FeatureFlagValue | undefined> {
+    } = {},
```
dustinbyrne, line 624: This is nit-picky, but I think keeping `sendFeatureFlagEvents` in `options` is cleaner. Every caller has `sendFeatureFlagEvents` in its `options` object, and so they have to destructure it out to pass it as a separate parameter. It could just be consumed here as `options.sendFeatureFlagEvents ?? true`.

haacked: Good call.

Issue #2542, opened 2025-11-03
before_send support for web lite and react native


marandaneto:
## Is your feature request related to a problem?

similar to https://github.com/PostHog/posthog-js/issues/2198
move the node bits to core and reuse it

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

PR #2931, opened 2026-01-19
chore: before_send support for web lite and react native

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/2542
Node uses a completely different structure for the message so we cant easily reuse without breaking changes
Adding beforeSend to core and web lite as a new thing
port https://posthog.com/docs/error-tracking/cutting-costs#using-the-before_send-hook

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -888,9 +888,17 @@ export abstract class PostHogCoreStateless {
       return
     }
 
+    let message: PostHogEventProperties | null = this.prepareMessage(type, _message, options)
+
+    // Allow subclasses to transform or filter the message (e.g., before_send hook)
+    message = this.processBeforeEnqueue(message)
+    if (message === null) {
+      return
+    }
+
     const data: Record<string, any> = {
       api_key: this.apiKey,
-      batch: [this.prepareMessage(type, _message, options)],
+      batch: [message],
       sent_at: currentISOTime(),
     }
 


ablaszkiewicz (approved):
Lgtm

packages/core/src/posthog-core-stateless.ts
```diff
@@ -824,14 +824,31 @@ export abstract class PostHogCoreStateless {
...
         return
       }
 
-      const message = this.prepareMessage(type, _message, options)
+      let message: PostHogEventProperties | null = this.prepareMessage(type, _message, options)
+
+      // Allow subclasses to transform or filter the message
+      message = this.processBeforeEnqueue(message)
+      if (message === null) {
+        return
```
ioannisj, line 850: should we log and warn about dropping internal events, similar to what we do on iOS and Android?

packages/core/src/posthog-core.ts
```diff
@@ -24,15 +26,16 @@ import {
   normalizeFlagsResponse,
   updateFlagValue,
 } from './featureFlagUtils'
-import { Compression, PostHogPersistedProperty } from './types'
+import { Compression, PostHogPersistedProperty, PostHogQueueItem } from './types'
```
ioannisj: PostHogQueueItem seems to be unused here?

packages/core/src/posthog-core.ts
```diff
@@ -24,15 +26,16 @@ import {
...
 import { maybeAdd, PostHogCoreStateless, QuotaLimitedFeature } from './posthog-core-stateless'
 import { uuidv7 } from './vendor/uuidv7'
-import { isPlainError } from './utils'
+import { isPlainError, safeSetTimeout } from './utils'
```
ioannisj: same, safeSetTimeout seems to be unused?

packages/core/src/posthog-core.ts
```diff
@@ -950,4 +954,81 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+
+  /**
+   * Runs the before_send hook(s) on the given capture event.
+   * If any hook returns null, the event is dropped.
+   *
+   * @param captureEvent The event to process
+   * @returns The processed event, or null if the event should be dropped
+   */
+  private _runBeforeSend(captureEvent: CaptureEvent): CaptureEvent | null {
+    const fns = Array.isArray(this._beforeSend) ? this._beforeSend : [this._beforeSend!]
```
ioannisj: I think there's a null check already on call site for _beforeSend, but let's add one here as well so it's self-contained?

packages/core/src/posthog-core.ts
```diff
@@ -950,4 +954,81 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+    let result: CaptureEvent | null = captureEvent
+
+    for (const fn of fns) {
+      result = fn(result)
```
ioannisj: Since this is external and can do anything, should we try catch and process the message unchanged in case of an error?

packages/core/src/posthog-core-stateless.ts
marandaneto: _runBeforeSend does that already

packages/core/src/posthog-core.ts
marandaneto: good catch, i made a refactor and rollback, linter didnt catch tho

packages/core/src/posthog-core.ts
marandaneto: not sure i follow, check if what is null?

packages/core/src/posthog-core.ts
marandaneto: yes, good point

packages/core/src/posthog-core.ts
ioannisj: `this._beforeSend` is optional and we are non-null asserting with `this.beforeSend!` without checking?

marandaneto: oh yes 100%, my bad, i was thinking about the Array.isArray which checks for nullability, but then it fallback to `this._beforeSend!` which could race if the global var changes within the calls anyway

PR #2934, opened 2026-01-19
ci: Notify Slack on rejected release

dustinbyrne:
## Problem

If the release deployment is rejected, nothing is reported to Slack.

## Changes

Added an additional workflow job that:
1. Runs when version-bump fails and a Slack thread exists                                                                                                                                                        
2. Checks the approvals API to detect if the failure was due to rejection                                                                                                                                        
3. Posts the rejection reason (if provided) as a reply to the original Slack thread


rafaeelaudibert (approved):
Nice! Can we get this added to `posthog-go` and `posthog-php` as well?

PR #2935, opened 2026-01-19
chore: more wrapper tests

pauldambra:
we could really use more wrapper tests
ideally to cover things we don't know about
but that's 🤯 
so let's cover things we do know about


ksvat (approved):
[cheers](https://app.graphite.com/settings/meme-library?org=PostHog)

TueHaulund (approved):
🥇

PR #2938, opened 2026-01-20
fix: track the last 6 full snapshot timestamps as a debug property

pauldambra:
we do still see some sites sending recordings without full snapshots
we believe this is related to a race condition in handling resets

let's track fullsnapshot sessoin ids by timestamp as a debug property

### DIFF
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -327,6 +327,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
     // as we make some decisions based on them without referencing LinkedFlag etc
     private _triggerMatching: TriggerStatusMatching = new PendingTriggerMatching()
     private _fullSnapshotTimer?: ReturnType<typeof setInterval>
+    private _fullSnapshotTimestamps: Array<[string, number]> = []
 
     private _windowId: string
     private _sessionId: string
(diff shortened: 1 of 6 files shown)


fasyy612 (approved):
Awesome, LGTM! 🤲

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -959,9 +960,6 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             // Full snapshots reset rrweb's node IDs, so clear any logged node tracking
             this._mutationThrottler?.reset()
 
-            // Track the timestamp of the first full snapshot for this session
-            // This helps us detect session rotation issues where incremental snapshots
-            // are sent before the full snapshot
             this._instance.persistence?.register_once(
```
pauldambra, line 966:
```suggestion
            // Track the timestamp of the first full snapshot for this session
            // This helps us detect session rotation issues where incremental snapshots
            // are sent before the full snapshot
            this._instance.persistence?.register_once(
```

playground/nextjs/bin/localdev.sh
```diff
@@ -26,6 +26,18 @@ echo ""
 echo -e "${GREEN}2. Creating tarballs...${NC}"
 pnpm package
 
+# Step 2b: If POSTHOG_REPO is set, link posthog-js to PostHog repo
```
pauldambra, line 29: means i can run [localdev.sh](http://localdev.sh) pointing at my local repo dir and it sorts out linking the package to the nextjs playground with my local posthog

e.g. I run

`NEXT_PUBLIC_POSTHOG_KEY='blah' NEXT_PUBLIC_POSTHOG_HOST='http://localhost:8010' POSTHOG_REPO=/Users/pauldambra/github/posthog ./bin/localdev.sh`

playground/nextjs/pages/index.tsx
```diff
@@ -116,6 +117,8 @@ export default function Home() {
                 <button onClick={() => setModalOpen(true)}>Open Modal</button>
             </div>
 
+            <SessionInteractions />
```
pauldambra, line 120: adds a section to the playground so we can test idle and rotation behaviour more easily

PR #2944, opened 2026-01-21
feat: allow for properties to be excluded when calling posthog.reset(…

RayKay91:
…) in React Native

## Problem

In a previous version of react-native-posthog we were able to exclude some properties from the `reset()` by passing in an array of the ones we wanted to keep. It looks like this was changed at some point, but the core functionality remains in posthog-core.ts. (If there's a different/better way to do this let me know please 🙏 !)

## Changes

Allows for the properties we'd like excluded from the `reset()` to be passed to the parent class from RN `posthog.reset()`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code - will do if there isn't a better way to do what the source code change reflects

### If releasing new changes

> [!IMPORTANT]
>
>^^^^^ I can't edit the labels!

### DIFF
--- a/.changeset/smart-falcons-tease.md
+++ b/.changeset/smart-falcons-tease.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': minor
+---
+
+Allow for properties to be excluded from reset


packages/react-native/src/posthog-rn.ts
```diff
@@ -377,10 +377,18 @@ export class PostHog extends PostHogCore {
...
+   * posthog.reset([PostHogPersistedProperty.OverrideFeatureFlags])
+   * ```
+   *
+   * @param propertiesToKeep - Optional array of persisted properties to preserve during reset
+   *
    * @public
    */
-  reset(): void {
-    super.reset()
+  reset(propertiesToKeep?: Parameters<PostHogCore['reset']>[0]): void {
```
marandaneto:
```suggestion
  reset(propertiesToKeep?: PostHogPersistedProperty[]): void {
```

better since the type is exported

packages/react-native/src/posthog-rn.ts
```diff
@@ -37,8 +37,8 @@ export interface PostHogOptions extends PostHogCoreOptions {
   persistence?: 'memory' | 'file'
   /** Allows you to provide your own implementation of the common information about your App or a function to modify the default App properties generated */
   customAppProperties?:
-    | PostHogCustomAppProperties
-    | ((properties: PostHogCustomAppProperties) => PostHogCustomAppProperties)
+  | PostHogCustomAppProperties
```
marandaneto: can you rollback the formatting, you might have a different linter version

packages/react-native/src/posthog-rn.ts
```diff
@@ -1304,8 +1312,8 @@ export class PostHog extends PostHogCore {
       if (!appBuild || !appVersion) {
         this._logger.warn(
           'PostHog could not track installation/update/open, as the build and version were not set. ' +
-            'This can happen if some dependencies are not installed correctly, or if you have provided' +
-            'customAppProperties but not included $app_build or $app_version.'
+          'This can happen if some dependencies are not installed correctly, or if you have provided' +
+          'customAppProperties but not included $app_build or $app_version.'
```
marandaneto: same https://github.com/PostHog/posthog-js/pull/2944/files#r2713794163

packages/react-native/src/posthog-rn.ts
RayKay91: amended

packages/react-native/src/posthog-rn.ts
RayKay91: rolled back formatting

packages/react-native/src/posthog-rn.ts
RayKay91: same

PR #2969, opened 2026-01-26
feat: autocapture controls

ablaszkiewicz:
# PLEASE DO NOT REVIEW IT'S BACK TO DRAFT

This PR adds support for auto capture ingestion controls fetched via remote config.

I looked at existing session replay code but:
- it serves kinda different purpose as it is heavily based on sessions whereas our ET isn't,
- it is all over the place and has super complicated dependencies so I just feel super uncomfortable changing anything there,

As a result I decided to do a dedicated, self-contained ET implementation of triggers.

# Sample rate

https://github.com/user-attachments/assets/cb020f64-9a11-4ee3-a971-7c339e3db261

# Event trigger

https://github.com/user-attachments/assets/e631986d-569f-4918-a506-023aec6163f5

# URL trigger

https://github.com/user-attachments/assets/adbb395a-95db-43e3-aa51-65422826a6f9

# Flag trigger

https://github.com/user-attachments/assets/9d0ec90f-255a-45e8-a703-5c22f5e390ab

# COMBINED

https://github.com/user-attachments/assets/7039ec50-3105-42da-8acb-17f17f5a21f7


packages/browser/src/types.ts
```diff
@@ -429,11 +432,25 @@ export type ErrorEventArgs = [
...
-export interface SessionRecordingUrlTrigger {
+export interface UrlTrigger {
     url: string
     matching: 'regex'
 }
 
+export interface ErrorTrackingAutoCaptureControls {
+    id?: string | null
+    library: string
+    matchType: 'any' | 'all'
```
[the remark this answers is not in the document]
ablaszkiewicz: Yeah for now it is not used

packages/browser/src/extensions/exception-autocapture/controls/triggers/sample-trigger.ts
```diff
@@ -0,0 +1,21 @@
...
+    init(context: DeciderContext): void {
+        this._context = context
+        this._sampleRate = context.config?.sampleRate ?? null
+    }
+
+    shouldCapture(): boolean | null {
+        if (isNull(this._sampleRate)) {
+            return null
+        }
+        return Math.random() < this._sampleRate
```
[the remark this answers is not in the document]
ablaszkiewicz: Works like intended

packages/browser/src/extensions/exception-autocapture/controls/autocaptureDecider.ts
```diff
@@ -0,0 +1,66 @@
...
+    constructor(posthog: PostHog) {
+        this._posthog = posthog
+    }
+
+    init(remoteConfig: RemoteConfig): void {
+        const autoCaptureConfig = remoteConfig.errorTracking?.autoCaptureControls?.web
+        const context = this._buildContext(autoCaptureConfig)
+
+        this._deciders.length = 0
+        this._deciders.push(new URLDecider(), new EventDecider(), new FlagDecider(), new SampleDecider())
```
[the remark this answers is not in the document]
ablaszkiewicz: As intended

packages/browser/src/utils/policyMatching.ts
pauldambra, line 1: can we call this specificationMatching or triggerMatching :)

packages/browser/src/extensions/exception-autocapture/controls/autocaptureDecider.ts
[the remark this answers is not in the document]
pauldambra: agree init must be idempotent, this will inevitably be called more than once

packages/browser/src/extensions/exception-autocapture/controls/autocaptureDecider.ts
```diff
@@ -0,0 +1,66 @@
...
+
+        for (const decider of this._deciders) {
+            decider.init(context)
+        }
+    }
+
+    shouldCapture(): boolean {
```
pauldambra, line 64: i wonder how much more of this file could be lazy loaded... is exception code lazy loaded already?

ablaszkiewicz: exception code is not lazy loaded

packages/browser/src/types.ts
```diff
@@ -242,6 +242,9 @@ export interface RemoteConfig {
         autocaptureExceptions?: boolean
         captureExtensionExceptions?: boolean
         suppressionRules?: ErrorTrackingSuppressionRule[]
+        autoCaptureControls?: {
```
pauldambra, line 245: is this autocapture or triggering?
we already have autocapture and it has a relatively defined meaning

should we call this something else?

ablaszkiewicz: This triggers autocapture

packages/browser/src/extensions/replay/external/triggerMatching.ts
pauldambra, line 1: iirc this code is implicitly lazy loaded i don't think that changes here but we should be careful of backwards compatability in changing it

ablaszkiewicz: How do I verify whether it broke the lazy load?

packages/browser/src/extensions/exception-autocapture/controls/autocaptureDecider.ts
pauldambra: shame... could it be?

packages/browser/src/extensions/replay/external/triggerMatching.ts
pauldambra: i need to get this in https://github.com/PostHog/posthog-js/pull/3003

packages/browser/src/types.ts
pauldambra: well, surely not... autocapture is a specific feature in autocapture.ts  
  
what does the SDK do while waiting for these controls over remote config  
what does it do when it receives the remote config  
  
in replay we have `trigger_pending` as a state so we can distinguish whether the config was received (cos users will 100% ask about it) and we write that to debug properties on events  
  
there these external controls are whether to trigger recording  
  
here surely these external controls are whether to trigger exception capture

pauldambra: (if we just give everything the same name it will end up super hard to change things, and every change in the SDK is basically supported forever so you're stuck with it. these are hard to change decisions)  
  
e.g. i really wish all our config was either enable or disable instead of the mix we have

PR #2979, opened 2026-01-27
feat: rollup changes and hono+vite+cloudflare example

ablaszkiewicz:
## Problem

- some changes in rollup plugin required for hono+vite+cloudflare setup to work,
- hono+vite+cloudflare example

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [x] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

### If releasing new changes

### DIFF
--- a/examples/example-cloudflare-hono-vite/src/index.ts
+++ b/examples/example-cloudflare-hono-vite/src/index.ts
@@ -11,7 +11,7 @@ const posthog = new PostHog(
 app.get('/', async (c) => {
 
   const error = new Error('test from cloudflare')
-  posthog.captureExceptionImmediate(error, 'cloudflare-user-id')
+  await posthog.captureExceptionImmediate(error, 'cloudflare-user-id')
 
   return c.json({ success: true, message: 'Exception captured!' })
 })
--- a/packages/rollup-plugin/src/index.ts
+++ b/packages/rollup-plugin/src/index.ts
@@ -68,7 +68,7 @@ export default function posthogRollupPlugin(userOptions: PostHogRollupPluginOpti
 
                 for (const fileName in bundle) {
                     const chunk = bundle[fileName]
-                    const isJsFile = /\.(m?c?js)$/.test(fileName)
+                    const isJsFile = /\.(js|mjs|cjs)$/.test(fileName)
                     if (chunk.type === 'chunk' && isJsFile) {
                         const chunkPath = path.resolve(...basePaths, fileName)
                         chunks[chunkPath] = chunk


packages/rollup-plugin/src/index.ts
```diff
@@ -68,13 +68,19 @@ export default function posthogRollupPlugin(userOptions: PostHogRollupPluginOpti
...
-                    if (chunk.type === 'chunk') {
+                    const isJsFile = /\.(m?c?js)$/.test(fileName)
+                    if (chunk.type === 'chunk' && isJsFile) {
                         const chunkPath = path.resolve(...basePaths, fileName)
                         chunks[chunkPath] = chunk
                         args.push('--file', chunkPath)
                     }
                 }
 
+                if (Object.keys(chunks).length === 0) {
```
ablaszkiewicz, line 79: We have to do this because for example hono+vite have this setup where there are multiple stages of the build and some stages may not output any files. 

We then get an error from the CLI that we didn't provide any files

packages/rollup-plugin/src/index.ts
```diff
@@ -68,13 +68,19 @@ export default function posthogRollupPlugin(userOptions: PostHogRollupPluginOpti
 
                 for (const fileName in bundle) {
                     const chunk = bundle[fileName]
-                    if (chunk.type === 'chunk') {
+                    const isJsFile = /\.(m?c?js)$/.test(fileName)
+                    if (chunk.type === 'chunk' && isJsFile) {
```
ablaszkiewicz, line 72: We have to do this because hono+vite setup outputs strange non-js map files which point to non-existent files.

This is an example file `dist/client/__cloudflare_fallback_entry__.map`
```
{"version":3,"file":"__cloudflare_fallback_entry__","sources":[],"sourcesContent":[],"names":[],"mappings":""}
```

This is strange but this is the default behavior and everything works fine so I guess we're good with that check

examples/example-cloudflare-hono-vite/src/index.ts
```diff
@@ -0,0 +1,21 @@
+import { Hono } from 'hono'
+import { PostHog } from 'posthog-node'
+
+const app = new Hono()
+
+app.get('/', async (c) => {
+  const posthog = new PostHog(
```
hpouillot: Can we move the posthog instance outside the endpoint as a singleton ? Also we should call captureExceptionImmediate, this way there is no need for flush or shutdown

examples/example-cloudflare-hono-vite/src/index.ts
```diff
@@ -0,0 +1,19 @@
...
+
+const posthog = new PostHog(
+  process.env.POSTHOG_PROJECT_API_KEY!,
+  { host: process.env.POSTHOG_API_HOST! }
+)
+
+app.get('/', async (c) => {
+
+  const error = new Error('test from cloudflare')
+  posthog.captureExceptionImmediate(error, 'cloudflare-user-id')
```
hpouillot:
```suggestion
  await posthog.captureExceptionImmediate(error, 'cloudflare-user-id')
```

PR #2982, opened 2026-01-28
feat(llma): add prompt management

Radu-Raicea:
## Changes

This adds the prompt management functions to the SDK.

The flow is to:
1. Fetch a prompt template
2. Compile the prompt with variables
3. Use it in your LLM calls

There's SDK side caching for the prompt. There's also a an optional fallback prompt. You can technically achieve 100% availability by using both.

Example:
```typescript
import { Prompts, OpenAI } from '@posthog/ai'

const posthog = new PostHog('phc_xxx', {
  personalApiKey: 'phx_xxx',
})

const prompts = new Prompts({ posthog })
const openai = new OpenAI({ apiKey: 'sk-xxx', posthog })

const template = await prompts.get('support-system-prompt', {
  fallback: 'You are a helpful assistant.',
})

const systemPrompt = prompts.compile(template, {
  company: 'Acme Corp',
  tier: 'premium',
})

const response = await openai.chat.completions.create({
  model: 'gpt-4',
  messages: [
    { role: 'system', content: systemPrompt },
    { role: 'user', content: 'How do I reset my password?' },
  ],
  posthogDistinctId: 'user-123',
})
```

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/ai/src/prompts.ts
+++ b/packages/ai/src/prompts.ts
@@ -1,3 +1,5 @@
+/// <reference lib="dom" />
+
 import type { PostHog } from 'posthog-node'
 import type { CachedPrompt, GetPromptOptions, PromptApiResponse, PromptVariables, PromptsDirectOptions } from './types'
 


packages/ai/src/index.ts
```diff
@@ -4,10 +4,12 @@ import { wrapVercelLanguageModel } from './vercel/middleware'
 import PostHogAnthropic from './anthropic'
 import PostHogGoogleGenAI from './gemini'
 import { LangChainCallbackHandler } from './langchain/callbacks'
+import { Prompts } from './prompts'
```
andrewm4894, line 7: general one, just wondering - do or would we want to pass any special `$ai_prompt_*` type properties through to llm analytics events we end up with? 

assuming we tryign to be as clean and minimal as possible here to begin but just wondering

andrewm4894: also wondering about if we want any error tracking integration or anything to measure somehow prompt failure events in some way. but again assuming we being as minimal as possible here to start as simple as possible.

Radu-Raicea: > general one, just wondering - do or would we want to pass any special $ai_prompt_* type properties through to llm analytics events we end up with?

This is actually an interesting point, but it's hard to integrate it _natively_. We don't know what they'll be doing with the compiled prompt. However, We could guide them to pass a property like `$ai_prompt_key` to LLM calls/generation events, with the value being the prompt key name, and if they do, we display things nicely in the LLMA dashboard (with Experiments support in the future). Same thing with `$ai_prompt_version` when we add the concept of versions.

Radu-Raicea: > also wondering about if we want any error tracking integration or anything to measure somehow prompt failure events in some way. but again assuming we being as minimal as possible here to start as simple as possible.

Noting it down, it's a good idea.

PR #2997, opened 2026-01-29
chore(product tours): export utils for use in main app

adboio:
### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/entrypoints/product-tours-preview.es.ts
+++ b/packages/browser/src/entrypoints/product-tours-preview.es.ts
@@ -1,1 +1,2 @@
 export { renderProductTourPreview } from '../extensions/product-tours/preview'
+export { addProductTourCSSVariablesToElement } from '../extensions/product-tours/product-tours-utils'
--- a/.changeset/dark-flies-bow.md
+++ b/.changeset/dark-flies-bow.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+exporting some utils


packages/browser/src/entrypoints/product-tours-preview.es.ts
```diff
@@ -1 +1,3 @@
 export { renderProductTourPreview } from '../extensions/product-tours/preview'
+export { addProductTourCSSVariablesToElement } from '../extensions/product-tours/product-tours-utils'
```
pauldambra, line 2: just a note that this becomes part of the public API of the SDK once you export it  
  
we maybe can't think why someone would use it but that doesn't mean they won't  
  
it's fine to expose it, but changing it becomes "make sure it's not a breaking change" territory  
  
you might not want to stand that cost remembering that a little duplication is cheaper than the wrong abstraction

adboio: for sure - willing to take that tradeoff here  
  
i'm reworking the product tour content editor to be more wysiwyg-like so the editor is much closer to exactly what the end-user will see  
  
we can't make breaking changes once exported, but we'll never drift on the content styles & accidentally do/show something unexpected to customers or their users - this ensures the preview / editor / end-tour all use the exact same styles  
  
also just realized i don't actaully need the DEFAULT_PRODUCT_TOUR_APPEARANCE exported, will remove that but keep the other function

pauldambra: i wonder if this will keep growing and it's worth taking the hit now and having a bundle from the main app (like the toolbar) that the SDK can load. i.e. we'd have product-tours.js like we have toolbar.js  
  
can't quite put my finger on why the dependency in this diretion feels wrong to me.   
i'm not ticking request changes, cos if someone thinks i'm being overly cautious then they can approve and i'll get back in my box :)

adboio: hmm that's a fair point, maybe so - i'm working on the main app side of this now, lemme think on it while i do

PR #2999, opened 2026-01-30
fix: fetch the actual config when user has stale config

fasyy612:
## Problem

We kept seeing replays that does not match the recording conditions (url match trigger etc), seeing their replays the actual config seems to be not loaded at all for only some users. Perhaps stale cached config was used at startup and new triggers from fresh decide responses were never applied to the running recorder.

## Changes

Reload the config so the user always gets proper configurations for replay settings

### DIFF
--- a/packages/browser/src/extensions/replay/session-recording.ts
+++ b/packages/browser/src/extensions/replay/session-recording.ts
@@ -7,7 +7,13 @@ import {
     SESSION_RECORDING_REMOTE_CONFIG,
 } from '../../constants'
 import { PostHog } from '../../posthog-core'
-import { Properties, RemoteConfig, SessionRecordingPersistedConfig, SessionStartReason } from '../../types'
+import {
+    Properties,
+    RemoteConfig,
+    SessionRecordingPersistedConfig,
+    SessionRecordingRemoteConfig,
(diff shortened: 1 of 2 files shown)


packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -71,6 +71,9 @@ export class SessionRecording {
 
     startIfEnabledOrStop(startReason?: SessionStartReason) {
         if (this._isRecordingEnabled && this._lazyLoadedSessionRecording?.isStarted) {
+            // Recording is already running - update config in case new remote config arrived
```
pauldambra: could we not do this when the config arrives?  
  
or rather I _thought_ we did do this when the config arrives  
  
i think the flow should be

1. start with cached config if there is any or wait
2. recieve new config
3. stop and restart with the new config if it is different

i'd trace through from `onRemoteConfig`

fasyy612: sounds good, I was just thinking maybe some users have a stale config somehow (cache or similar) that prevents them from getting the actual remote config 🤔 I'll try tracing first as well!

fasyy612:
will open a fresh one

PR #3003, opened 2026-01-31
fix: run playwright tests with the new extension files and old core to check for breaking changes

pauldambra:
we can make a change in one of the lazy loaded files 
when it is released people with an earlier core posthog
will lazy load the new version
this is an easy way for us to introduce bugs

let's run the playwright test with array.js from the currently published at the time of the PR with the new lazy loaded files to expose this

tested this by taking commit c8d3fbe5e3a4944596d7a34086484affd94d7329 which introduced a breaking change bug and testing it against 1.334.1 the previous version at the time and seeing it fail

### DIFF
--- a/packages/browser/playwright/compat-skips.ts
+++ b/packages/browser/playwright/compat-skips.ts
@@ -11,13 +11,7 @@ import { satisfies } from 'compare-versions'
  *
  * Format: Map of version ranges to test patterns with required reasons
  *
- * Example:
- *   '<1.335.0': [
- *     {
- *       test: 'web_vitals_attribution: true includes attribution data',
- *       reason: 'web_vitals_attribution option added in #2953',
- *     },
- *   ]
+
  */
 export const compatSkips: { range: string; test: string; reason: string }[] = [
     {


.github/workflows/library-ci.yml
```diff
@@ -57,6 +57,154 @@ jobs:
...
+
+            ```typescript
+            export const compatSkips = [
+                {
+                    range: '<1.335.0',
+                    test: 'web_vitals_attribution: true includes attribution data',
+                    reason: 'web_vitals_attribution option added in #2953',
+                },
+            ]
+            ```
```
rafaeelaudibert, line 176: Can we maybe make it fail until we explain the reason?

rafaeelaudibert: Also, this is tricky, because we don't know what version we'll have once our code is deployed. Thoughts?

packages/browser/playwright/compat-skips.ts
```diff
@@ -0,0 +1,42 @@
...
+ * Format: Map of version ranges to test patterns with required reasons
+ *
+ * Example:
+ *   '<1.335.0': [
+ *     {
+ *       test: 'web_vitals_attribution: true includes attribution data',
+ *       reason: 'web_vitals_attribution option added in #2953',
+ *     },
+ *   ]
+ */
```
rafaeelaudibert: Do we even need the example if we have one down there? he

packages/browser/playwright/global-setup-compat.ts
```diff
@@ -0,0 +1,48 @@
...
+            : `https://registry.npmjs.org/posthog-js/${targetVersion}`
+    const registryResponse = await fetch(registryUrl)
+    if (!registryResponse.ok) {
+        throw new Error(`Failed to fetch NPM registry: ${registryResponse.status}`)
+    }
+    const packageInfo = await registryResponse.json()
+    const version = packageInfo.version
+
+    // eslint-disable-next-line no-console
+    console.log(`Compat tests using posthog-js@${version} from NPM`)
```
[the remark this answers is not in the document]
rafaeelaudibert: We are _reading_ from there Graphite, so what you're saying isn't right

rafaeelaudibert (approved):
It makes sense! Left some comments on how we could make this even better

.github/workflows/library-ci.yml
pauldambra: yep, it should fail until there's a skip for it  
  
if you add a new feature the test for that _has_ to fail in the old version even though it's not a breaking change.

packages/browser/playwright/compat-skips.ts
pauldambra: :shrug:someone could delete them all in future but then, someone could always change anything

packages/browser/playwright/compat-skips.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Format: Map of version ranges to test patterns with required reasons
+ *
+ * Example:
+ *   '<1.335.0': [
+ *     {
+ *       test: 'web_vitals_attribution: true includes attribution data',
+ *       reason: 'web_vitals_attribution option added in #2953',
+ *     },
+ *   ]
```
pauldambra:
```suggestion
```

.github/workflows/library-ci.yml
pauldambra: i'm gonna suck it and see since there's already a PR i want to validate with this

PR #3004, opened 2026-01-31
feat: add multi-language survey support with automatic language detection and fallback

Basit-Balogun10:
## Problem

Teams running global products need to collect feedback from users who speak different languages. Currently, surveys only support one language per survey, forcing teams to either create duplicate surveys for each language (maintenance nightmare, fragmented analytics) or use English-only surveys (low response rates from non-English speakers).

This implements the final piece of the surveys translation epic, enabling one survey to automatically display in multiple languages based on the user's language preference.

Related:
- Backend PR: https://github.com/PostHog/posthog/pull/45096 (now https://github.com/PostHog/posthog/pull/47934)
- Frontend PR: https://github.com/PostHog/posthog/pull/45281
- Issue: https://github.com/PostHog/posthog/issues/42154

## Changes

Added multi-language survey support to posthog-js with automatic language detection and intelligent fallback:

**Core functionality:**
- Detects user language from person properties (`$stored_person_properties.language`)
- Applies translations with fallback logic: exact match → base language (e.g., `fr-CA` → `fr`) → default
- Merges translated fields on top of default survey/question fields
- Tracks `$survey_language` in survey events for analytics

**Translation scope:**
- Survey-level: name, description, thank you message (header, description, close button)
- Question-level: question text, description, button text, link URLs, rating labels, multiple choice options

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys.tsx
+++ b/packages/browser/src/extensions/surveys.tsx
@@ -69,6 +69,7 @@ import {
     calculatePrefillStartIndex,
 } from '../utils/survey-url-prefill'
 import { getNextSurveyStep } from '../utils/survey-branching'
+import { applySurveyTranslationForUser } from '../utils/survey-translations'
 
 // Re-export for surveys-preview entrypoint
 export { getNextSurveyStep }
@@ -180,25 +181,30 @@ export class SurveyManager {
     }
 
     public handlePopoverSurvey = (surveyParam: Survey, options?: DisplaySurveyPopoverOptions): void => {
+        const { survey: translatedSurvey, language: surveyLanguage } = applySurveyTranslationForUser(
+            surveyParam,
+            this._posthog
+        )
+
         // apply overrides for position / selector (needed for thumb surveys)
         const survey =
             options?.position || options?.selector
                 ? {
-                      ...surveyParam,
+                      ...translatedSurvey,
                       appearance: {
-                          ...surveyParam.appearance,
+                          ...translatedSurvey.appearance,
                           ...(options.position && { position: options.position }),
                           ...(options.selector && { widgetSelector: options.selector }),
                       },
                   }
-                : surveyParam
+                : translatedSurvey
 
         this._clearSurveyTimeout(survey.id)
 
         const { properties, initialResponses } = options ?? {}
         const hasPrefillData = initialResponses && Object.keys(initialResponses).length > 0
         const isSurveyCompleted = hasPrefillData
-            ? this._handleInitialResponses(survey, initialResponses, properties)
+            ? this._handleInitialResponses(survey, initialResponses, properties, surveyLanguage)
             : false
 
         // if the survey is done (from prefill) and there is no thank-you, we can break early
@@ -232,6 +238,7 @@ export class SurveyManager {
             style: positionStyle,
             isSurveyCompleted: isSurveyCompleted,
             skipShownEvent: options?.skipShownEvent,
+            surveyLanguage: surveyLanguage,
         }
 
         if (delaySeconds <= 0) {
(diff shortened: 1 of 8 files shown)


packages/browser/src/utils/survey-translations.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    return null
+}
+
+/**
+ * Normalizes a language code to lowercase for consistent matching
+ * @param languageCode - The language code to normalize (e.g., 'FR', 'en-US')
+ * @returns Normalized language code (e.g., 'fr', 'en-us')
+ */
+function normalizeLanguageCode(languageCode: string): string {
```
adboio: i know it didn't exist when you started haha - but we added a bunch of this logic for product tours, can we re-use it?

packages/browser/src/utils/survey-translations.ts
```diff
@@ -0,0 +1,220 @@
...
+import { createLogger } from './logger'
+
+const logger = createLogger('[SurveyTranslations]')
+
+/**
+ * Detects the user's language from person properties
+ * @param instance - PostHog instance to retrieve person properties
+ * @returns The detected language code (e.g., 'fr', 'es', 'en-US') or null if not found
+ */
+export function detectUserLanguage(instance: PostHog): string | null {
```
adboio, line 24: tours will use browser lanugage, with a fallback to `config.override_display_language`

let's match that :)

https://posthog.com/docs/product-tours/localization

Basit-Balogun10:
Awesomeeee! Saw that it got merged too, thanksss!!!

The frontend? I don't get the question though, could you clarify? We had the PR over here https://github.com/PostHog/posthog/pull/45281 and just saw that it's now closed @adboio

Will address your review comments on this PR tomorrow, btw!

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1011,6 +1040,7 @@ export function usePopupVisibility(
                     [SurveyEventProperties.SURVEY_ID]: survey.id,
                     [SurveyEventProperties.SURVEY_ITERATION]: survey.current_iteration,
                     [SurveyEventProperties.SURVEY_ITERATION_START_DATE]: survey.current_iteration_start_date,
+                    ...(surveyLanguage && { $survey_language: surveyLanguage }),
```
adboio: nice! i don't think i did this for tours, i'll have to update that

nit: no need to add this conditionally, it's ok to pass `undefined`

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1265,6 +1332,7 @@ export function Questions({
                 isSurveyCompleted,
                 posthog,
                 properties,
+                surveyLanguage,
```
adboio, line 1339: we do a lot of passing this language around, i wonder if it could be stored as a property on `SurveyManager` directly?

although that could be trick, in case the user's language changes mid-survey...

product tours does not record which language was displayed (we should), but we do resolve the translation per-step which partially solves for "user changed languages"

https://github.com/PostHog/posthog-js/blob/e90364f3504086356fe68069f4c37e89c043a7f7/packages/browser/src/extensions/product-tours/product-tours.tsx#L231

maybe we can somehow attach a listener to `navigator.language` (see [getBrowserLanguage](https://github.com/PostHog/posthog-js/blob/e90364f3504086356fe68069f4c37e89c043a7f7/packages/browser/src/utils/event-utils.ts#L174), used for tours) and the posthog config, so we can re-render and update the language reference if it changes?

adboio: looks like there is a widely-supported `languagechange` event https://developer.mozilla.org/en-US/docs/Web/API/Window/languagechange_event

but there is no (easy) way to listen to config changes right now

feel free to ignore this as out of scope :laughing:

packages/browser/src/utils/survey-translations.ts
adboio: specifically we can re-use [getBrowserLanguage](https://github.com/PostHog/posthog-js/blob/e90364f3504086356fe68069f4c37e89c043a7f7/packages/browser/src/utils/event-utils.ts#L174) and use `config.override_display_language` as the override

https://github.com/PostHog/posthog-js/blob/e90364f3504086356fe68069f4c37e89c043a7f7/packages/browser/src/extensions/product-tours/product-tours.tsx#L239

packages/browser/src/utils/survey-translations.ts
adboio: we can prob rip some of this out to a helper https://github.com/PostHog/posthog-js/blob/e90364f3504086356fe68069f4c37e89c043a7f7/packages/browser/src/extensions/product-tours/product-tours-utils.ts#L295

packages/browser/src/extensions/surveys.tsx
Basit-Balogun10: Thanks for the suggestion! Storing language on SurveyManager and listening to languagechange/config changes would be a great enhancement, but yea we can pass this onto another pr/iteration where we add config change events to posthog core (and then we can re-render when language changes mid-session)

Guess we can just keep the current approach (passing surveyLanguage through functions) for this PR since it's simpler and works well. We can revisit dynamic language change handling in a future PR once we have a config change event system.

Added a TODO comment in the code to help us track this as a potential future enhancement (will take a shot at it once this is merged)

packages/browser/src/extensions/surveys.tsx
Basit-Balogun10: Makes sense!

packages/browser/src/utils/survey-translations.ts
Basit-Balogun10: Cool!

Basit-Balogun10: Btw, I also added person properties language as a third fallback (after config and browser language) to allow users to programmatically set language via `posthog.identify`. Thought this might be useful for user preferences from database or server-side language detection (for customers using the node sdk? essentially gives customer more optoins too)

PR #3006, opened 2026-02-01
feat(analytics-platform): Add function setTestUser

robbie-c:
## Problem

We want to make it easier for people to use cohorts for internal test filters, as from a query performance perspective, checking presence of a user in a small cohort is basically free, whereas checking the host / email column for every row can double the number of bytes read on disk of every query.

The problem is that it's very useful to be able to specify a hostname for test users! The problem with this is that it doesn't work with cohorts, because event properties can't be used in a cohort definition. This PR effectively changes this into a person property so that it can be used with cohorts

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -155,13 +155,13 @@ const defaultsThatVaryByConfig = (
     defaults?: ConfigDefaults
 ): Pick<
     PostHogConfig,
-    'rageclick' | 'capture_pageview' | 'session_recording' | 'external_scripts_inject_target' | 'testuser_hostname'
+    'rageclick' | 'capture_pageview' | 'session_recording' | 'external_scripts_inject_target' | 'test_user_hostname'
 > => ({
     rageclick: defaults && defaults >= '2025-11-30' ? { content_ignorelist: true } : true,
     capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true,
     session_recording: defaults && defaults >= '2025-11-30' ? { strictMinimumDuration: true } : {},
     external_scripts_inject_target: defaults && defaults >= '2026-01-30' ? 'head' : 'body',
-    testuser_hostname: defaults && defaults >= '2026-01-30' ? /^(localhost|127\.0\.0\.1)$/ : undefined,
+    test_user_hostname: defaults && defaults >= '2026-01-30' ? /^(localhost|127\.0\.0\.1)$/ : undefined,
 })
 
 // NOTE: Remember to update `types.ts` when changing a default value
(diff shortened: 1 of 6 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -3134,6 +3148,34 @@ export class PostHog implements PostHogInterface {
         this.setPersonProperties({}, {})
     }
 
+    /**
+     * Marks the current user as a test user by setting the `$test_user` person property to `true`.
+     * This also enables person processing for the current user.
+     *
+     * This is useful for using in a cohort your internal/test filters for your posthog org.
+     * @see https://posthog.com/tutorials/filter-internal-users
+     * Create a cohort with `$test_user` IS SET, and set your internal test filters to be NOT IN that cohort.
```
rafaeelaudibert, line 3157: Can we guarantee this cohort always exists AND is the default for new accounts? This will make people's lives much easier.

packages/browser/src/posthog-core.ts
```diff
@@ -153,11 +153,15 @@ let ENQUEUE_REQUESTS = !SUPPORTS_REQUEST && userAgent?.indexOf('MSIE') === -1 &&
...
-): Pick<PostHogConfig, 'rageclick' | 'capture_pageview' | 'session_recording' | 'external_scripts_inject_target'> => ({
+): Pick<
+    PostHogConfig,
+    'rageclick' | 'capture_pageview' | 'session_recording' | 'external_scripts_inject_target' | 'testuser_hostname'
+> => ({
     rageclick: defaults && defaults >= '2025-11-30' ? { content_ignorelist: true } : true,
     capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true,
     session_recording: defaults && defaults >= '2025-11-30' ? { strictMinimumDuration: true } : {},
     external_scripts_inject_target: defaults && defaults >= '2026-01-30' ? 'head' : 'body',
+    testuser_hostname: defaults && defaults >= '2026-01-30' ? /^(localhost|127\.0\.0\.1)$/ : undefined,
```
rafaeelaudibert: Why is this `testuser_hostname` and not `test_user_hostname`? Feels so weird

packages/browser/src/posthog-core.ts
robbie-c: I'm working on that here https://github.com/PostHog/posthog/pull/45899, but it's more complicated than you might think (cohorts aren't supported by hog functions), it might need to wait until behavioural cohorts are done

packages/browser/src/posthog-core.ts
robbie-c: I can change it :D

rafaeelaudibert (approved):
Looks good, but after reviewing it 3 times I now wonder if it should be `test_users`plural? I dont know, up to you!

PR #3007, opened 2026-02-01
fix(product tours): add element precision, only mark tours as shown if first element was found

adboio:
## Problem

two issues:

1. the element inference is too good at finding _something_, even when it's a "bad" something, which means tours are sometimes showing when they shouldn't. this also makes tour building feel wonky. see https://posthog.slack.com/archives/C07QD3LT8U9/p1769930669658259
2. we mark tours as shown before we know if the first step successfully rendered, meaning tour attempts could be "wasted" - see https://posthog.slack.com/archives/C07QD3LT8U9/p1769893805876789?thread_ts=1769725713.960099&cid=C07QD3LT8U9

## Changes

- adds `precision` param to inferred selectors. this is a number 0-1 that controls how many groups we check during element lookup. a future main app PR will put this in a slider for users, but for now, we default to 1, which means we'll only check the most specific group
- reorder the logic on tour start so we don't mark tours as shown until we actually know the first step successfully rendered

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/product-tours/element-inference.ts
+++ b/packages/browser/src/extensions/product-tours/element-inference.ts
@@ -87,6 +87,7 @@ export interface InferredSelector {
     autoData: string
     text: string | null
     excludeText?: boolean
+    precision?: number
 }
 
 function getElementText(element: HTMLElement): string | null {
@@ -159,7 +160,7 @@ export function findElement(selector: InferredSelector): HTMLElement | null {
             logger.error('Invalid autoData structure:', autoData)
             return null
         }
-        const { text, excludeText } = selector
+        const { text, excludeText, precision = 1 } = selector
 
         // excludeText -> user setting, usually if the target element
         // has dynamic/localized text
@@ -174,10 +175,15 @@ export function findElement(selector: InferredSelector): HTMLElement | null {
             return null
         }
 
+        // precision controls how many groups to search
+        // 1 = strict (only most specific group), 0 = loose (all groups)
+        const maxGroups = Math.max(1, Math.ceil((1 - precision) * groups.length))
+
         const visibilityCache = new WeakMap<HTMLElement, boolean>()
 
         // try each selector group, starting w/ most specific (lowest cardinality)
-        for (const group of groups) {
+        for (let i = 0; i < maxGroups; i++) {
+            const group = groups[i]
             const votes = new Map<HTMLElement, number>()
             let winner: HTMLElement | null = null
             let maxVotes = 0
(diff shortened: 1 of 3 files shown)


packages/browser/src/extensions/product-tours/element-inference.ts
```diff
@@ -174,10 +175,15 @@ export function findElement(selector: InferredSelector): HTMLElement | null {
             return null
         }
 
+        // precision controls how many groups to search
+        // 1 = strict (only most specific group), 0 = loose (all groups)
+        const maxGroups = Math.max(1, Math.ceil((1 - precision) * groups.length))
```
rafaeelaudibert, line 180: This feels a bit arcane and hard to explain. What if you had a linear value here rather than a proportional number where you basically let people choose how many selectors we try before giving up? Default to 10 or something.

rafaeelaudibert (commented):
Hmmm, the `precision` argument is a bit too arcane. Can we make it something easier to understand?

packages/browser/src/extensions/product-tours/element-inference.ts
adboio: i'm using a proportional number here because we don't know how many groups we'll have for a given element, and we don't know what their cardinalities will be, so a fixed number slider would be tricky - one element might have 10 groups, another might have 5, then the user is choosing 1-5 vs. 1-10 etc etc

although other options could be....

- assume a large number of groups >N just means any groups N+ are worthless, which is probably fair, then at least the user's slider is always capped at 1-(min(N, group count))
- maybe put the control on max cardinality instead of group count? the problem still exists that the lowest cardinality and range of cardinalities are dynamic, but that might be easier to explain

adboio: the idea with the current implementation was to show a slider without any numbers, just something like "loosest -> strictest"

if we moved to a cardinality control, for example, we could probably still present a slider with (min cardinality - max cardinality) using the same language, but with more details for more technical users, maybe even showing them some details from our inference data like the selectors and cardinality

lucasheriques (approved):

adboio:
### Merge activity

* **Feb 3, 11:15 PM UTC**: @adboio merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/3007).

PR #3009, opened 2026-02-02
feat(core,react-native): add setPersonProperties method

marandaneto:
## Description

Implements `setPersonProperties` support (`$set` event) for the core and React Native SDKs, matching the browser SDK implementation.

This method allows setting person properties directly without needing to call `identify()`, which is useful for updating user properties on already identified users.

Related to: https://github.com/PostHog/posthog-flutter/issues/276

### Core SDK (`@posthog/core`)
- Added `setPersonProperties(userPropertiesToSet?, userPropertiesToSetOnce?)` method
- Sends a `$set` event with `$set` and `$set_once` properties
- Updates person properties for feature flag evaluation
- Respects `personProfiles` setting (`identified_only`, `always`, `never`)

### React Native SDK (`posthog-react-native`)
- Added `setPersonProperties(userPropertiesToSet?, userPropertiesToSetOnce?, reloadFeatureFlags?)` method
- Includes `reloadFeatureFlags` parameter (defaults to `true`) consistent with `setPersonPropertiesForFlags`

### Tests
- Added comprehensive tests for all `personProfiles` modes
- Tests for `$set`, `$set_once`, and combined properties
- Tests for person processing behavior

## Usage

```javascript
// Set user properties (reloads feature flags by default in RN)
posthog.setPersonProperties({
    email: 'user@example.com',
    plan: 'premium'
})

// Set properties with $set_once
posthog.setPersonProperties(
    { name: 'Max Hedgehog' },  // $set properties
    { initial_url: '/blog' }   // $set_once properties
)

// React Native: Set properties without reloading feature flags
posthog.setPersonProperties({ plan: 'premium' }, undefined, false)
```

## Checklist
- [x] Changes are covered by tests
- [x] Code follows the project's style guidelines (linting passes)
- [x] Documentation (JSDoc) added for new methods

### DIFF
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -30,7 +30,7 @@ import {
 import { Compression, FeatureFlagError, PostHogPersistedProperty } from './types'
 import { maybeAdd, PostHogCoreStateless, QuotaLimitedFeature } from './posthog-core-stateless'
 import { uuidv7 } from './vendor/uuidv7'
-import { isEmptyObject, isPlainError } from './utils'
+import { isEmptyObject, isNullish, isPlainError } from './utils'
 
 export abstract class PostHogCore extends PostHogCoreStateless {
   // options
@@ -1208,7 +1208,9 @@ export abstract class PostHogCore extends PostHogCoreStateless {
     reloadFeatureFlags = true
   ): void {
     this.wrap(() => {
-      if (isEmptyObject(userPropertiesToSet) && isEmptyObject(userPropertiesToSetOnce)) {
+      const isSetEmpty = isNullish(userPropertiesToSet) || isEmptyObject(userPropertiesToSet)
+      const isSetOnceEmpty = isNullish(userPropertiesToSetOnce) || isEmptyObject(userPropertiesToSetOnce)
+      if (isSetEmpty && isSetOnceEmpty) {
         return
       }
 


packages/core/src/posthog-core.ts
```diff
@@ -1162,6 +1166,67 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+  setPersonProperties(
+    userPropertiesToSet?: { [key: string]: string },
+    userPropertiesToSetOnce?: { [key: string]: string },
+    reloadFeatureFlags = true
+  ): void {
+    this.wrap(() => {
+      if (!userPropertiesToSet && !userPropertiesToSetOnce) {
+        return
+      }
+
```
ioannisj, line 1216: I think we should move the check for empty props and return early here without capturing $set? Or is this intentional to allow creating a profile other than identify?

packages/core/src/posthog-core.ts
```diff
@@ -1162,6 +1166,67 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+   * @public
+   *
+   * @param userPropertiesToSet - Optional: An object of properties to store about the user.
+   *   These properties will overwrite any existing values for the same keys.
+   * @param userPropertiesToSetOnce - Optional: An object of properties to store about the user.
+   *   If a property is previously set, this does not override that value.
+   * @param reloadFeatureFlags - Whether to reload feature flags after setting the properties. Defaults to true.
+   */
+  setPersonProperties(
+    userPropertiesToSet?: { [key: string]: string },
```
ioannisj: shouldn't this be [PostHogEventProperties](https://github.com/PostHog/posthog-js/blob/fdde2bfaf3ef2a7c61c8d9b39b5c2cb6db0574c8/packages/core/src/types.ts#L184) (`[key: string]: JsonType` )

marandaneto: yes for JsonType, no for PostHogEventProperties since its not an event, you can check eg setPersonPropertiesForFlags
JS (browser package) calls the PostHogEventProperties type as Properties so its generic and can be used for event props and person props, but renaming this would be a breaking change

packages/core/src/posthog-core.ts
marandaneto: now i also check if they are empty

packages/core/src/posthog-core.ts
```diff
@@ -1162,6 +1166,67 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+      }
+
+      if (!this._requirePersonProcessing('posthog.setPersonProperties')) {
+        return
+      }
+
+      // Update person properties for feature flags evaluation
+      // Merge setOnce first, then set to allow overwriting
+      const mergedProperties = { ...(userPropertiesToSetOnce || {}), ...(userPropertiesToSet || {}) }
+      if (Object.keys(mergedProperties).length > 0) {
```
ioannisj: https://github.com/PostHog/posthog-js/pull/3009/changes#r2759069112 no need for this check?

marandaneto: oh yeah good catch, since we check above now

PR #3017, opened 2026-02-03
feat: Feedback recordings

luke-belton:
## Problem

In support it's often useful context when customers share screen recordings of their issue - but we then have to cross reference with session replay.

## Changes

Build the whole workflow in PostHog on top of session recording - add audio recording support to allow users to share feedback when they need support from PostHog.

This is an initial experimental/alpha release.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt

## Checklist

- [ ] Tests for new code

### If releasing new changes


pauldambra:
ok. we're down to +1.5 increase
would be good to figure out if we can get that down further

PR #3018, opened 2026-02-03
fix: Bundle size + config naming for feedback recordings

luke-belton:
## Problem

https://github.com/PostHog/posthog-js/pull/2571 had two issues:
* it increased bundle size a lot because the module I was doing the lazy loading from still had heavy imports
* the new config option wasn't sufficiently clear this feature was experimental

## Changes

* Rename the new config option `disable_feedback_recordings` => `_experimental_disable_feedback_recordings` - still defaults to `true` (off)
* Use pattern from dead clicks module to lazy load the heavier dependencies for this module

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/entrypoints/feedback-recording.ts
+++ b/packages/browser/src/entrypoints/feedback-recording.ts
@@ -35,7 +35,6 @@ class LazyLoadedFeedbackRecording implements LazyLoadedFeedbackRecordingInterfac
             },
             onCancel: () => {
                 this._resetState()
-                logger.info('Feedback recording UI cancelled')
                 onCancel()
             },
         })
--- a/packages/browser/src/extensions/feedback-recording.ts
+++ b/packages/browser/src/extensions/feedback-recording.ts
@@ -36,6 +36,7 @@ export class FeedbackRecording {
     public cleanup(): void {
         if (this._lazyLoadedFeedbackRecording) {
             this._lazyLoadedFeedbackRecording.cleanup()
+            this._lazyLoadedFeedbackRecording = null
         }
 
         this._isUIActive = false


packages/browser/src/posthog-feedback-recording.ts
pauldambra, line 1: so for lazy loading take a look at how deadclicks does it

you want an `entrypoint` that has all the code that will only be loaded if configured

and then in `extensions` you have code that decides whether to load the new js file 

so you'll end up with a `feedback-recording.js` being published and then on config loading that or not

pauldambra: ping here - if you need advice/example let me know!

pauldambra: @luke-belton

luke-belton: thank you! still on my todo list to keep looking at this when I find some quieter time away from tickets

luke-belton: I think my current implementation after an initial set of changes is somewhere between the deadclicks and survey approach for this (probably more like surveys!)

pauldambra: surveys is one of the earlier implementations so i'd copy deadclicks  
would be good to have these all be the same so that we make fewer mistakes with it :)

luke-belton: I think this is hopefully getting to a better place now. Bundle sizes have come back down to close(r) to what was there before. Not for this PR of course, but do you think the SDK is at a point where we could/should be abstracting some of the shared logic around extension loading?

Issue #2163, opened 2025-05-22
Capture `$set` event when identify is called with the same distinct id and different user props


marandaneto:
### Description

https://github.com/PostHog/posthog-js/blob/c8c747a53c4768686542899631ebbe440ab39760/src/posthog-core.ts#L1598

Only capture $set event if props have changed, see [this](https://github.com/PostHog/posthog-android/issues/251)

PR #3023, opened 2026-02-04
feat: only capture $set events if the user properties have changed

marandaneto:
## Problem

closes https://github.com/PostHog/posthog-js/issues/2163

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/utils/string-utils.ts
+++ b/packages/core/src/utils/string-utils.ts
@@ -25,6 +25,27 @@ export function isDistinctIdStringLike(value: string): boolean {
 }
 
 /**
+ * Recursively sorts all keys in an object and its nested objects/arrays.
+ * Used to ensure deterministic JSON serialization regardless of object construction order.
+ */
+function deepSortKeys(value: JsonType): JsonType {
+  if (value === null || typeof value !== 'object') {
+    return value
+  }
+
+  if (Array.isArray(value)) {
+    return value.map(deepSortKeys)
+  }
+
+  return Object.keys(value)
+    .sort()
+    .reduce((acc: { [key: string]: JsonType }, key) => {
+      acc[key] = deepSortKeys(value[key])
+      return acc
+    }, {})
+}
+
+/**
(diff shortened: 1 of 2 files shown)


packages/core/src/utils/string-utils.ts
```diff
@@ -21,3 +23,26 @@ export const stripLeadingDollar = function (s: string): string {
...
+  distinct_id: string,
+  userPropertiesToSet?: { [key: string]: JsonType },
+  userPropertiesToSetOnce?: { [key: string]: JsonType }
+): string {
+  const sortKeys = (obj?: { [key: string]: JsonType }): { [key: string]: JsonType } | undefined =>
+    obj
+      ? Object.keys(obj)
+          .sort()
+          .reduce((acc: { [key: string]: JsonType }, key) => ((acc[key] = obj[key]), acc), {})
+      : undefined
```
ioannisj: This is still top-level sorting right? 

```
getPersonPropertiesHash('123', { address: { city: 'NYC', zip: '10001' } }) 
getPersonPropertiesHash('123', { address: { zip: '10001', city: 'NYC' } })
 
```

would likely produce a different hash? We can add a test to check

marandaneto: yes, JS SDK does not even sort it

marandaneto: lemme check if i can fix this without much pain

marandaneto: done

ioannisj: Faster than I could comment :)

PR #3028, opened 2026-02-04
fix(core): queue pending feature flags reload instead of dropping requests

marandaneto:
## Problem

Relates: https://github.com/PostHog/posthog-ios/issues/456
Similar fix: https://github.com/PostHog/posthog-android/pull/407

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/__tests__/posthog.featureflags.spec.ts
+++ b/packages/core/src/__tests__/posthog.featureflags.spec.ts
@@ -973,14 +973,16 @@ describe('PostHog Feature Flags v4', () => {
         expect(posthog.getFeatureFlagPayload('feature-1')).toEqual(null)
       })
 
-      it('should emit debug message when quota limited', async () => {
-        const warnSpy = jest.spyOn(console, 'warn')
-        posthog.debug(true)
+      it('should emit featureflags event with quotaLimited when quota limited', async () => {
+        const featureFlagsHandler = jest.fn()
+        posthog.on('featureflags', featureFlagsHandler)
+
         await posthog.reloadFeatureFlagsAsync()
 
-        expect(warnSpy).toHaveBeenCalledWith(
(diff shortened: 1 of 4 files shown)


packages/core/src/posthog-core.ts
```diff
@@ -516,7 +528,17 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
     await this._initPromise
     if (this._flagsResponsePromise) {
-      return this._flagsResponsePromise
+      // Queue the reload request instead of dropping it
+      // This ensures that requests with $anon_distinct_id (from identify()) are not lost
+      this._logger.info('Feature flags are being loaded already, queuing reload.')
+      // Resolve any existing pending promise with the in-flight request's result to avoid hanging promises
+      if (this._pendingFlagsRequest) {
+        this._flagsResponsePromise.then(this._pendingFlagsRequest.resolve).catch(this._pendingFlagsRequest.reject)
+      }
```
ioannisj, line 537: I understand that this is best effort (since new flags will be replaced anw with the fresh API call), but should we be also passing a value to caller to let them know that the response is stale (and being refreshed)? Maybe they'd want to ignore it

marandaneto: i'd say this is ok since this is the user calling reload multiple times, we wouldnt know if its stale directly from the sdk anyway, we just know there will be another request going on

PR #3031, opened 2026-02-04
fix(react-native): prevent duplicate events with transactional storage updates

marandaneto:
## Problem

This change ensures that events are only considered 'sent' after the queue
changes have been fully persisted to async storage. Previously, the SDK would
fire-and-forget storage writes, which could lead to duplicate events if the
app crashed after successfully sending events to the server but before the
storage write completed.

Changes:
- Add flushStorage() hook in core that subclasses can override to wait for
  storage operations to complete
- Track pending async persist operations in PostHogRNStorage with waitForPersist()
- Make persistQueueChange() in _flush() async and await flushStorage() after
  each queue update
- React Native SDK overrides flushStorage() to wait for pending storage writes

This prevents the scenario where:
1. Events are sent to server successfully
2. App crashes before async storage write completes
3. On restart, events are still in queue and get sent again

Now the flush only completes after storage is confirmed written, ensuring
crash-safe event delivery without duplicates.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/lemon-jars-refuse.md
+++ b/.changeset/lemon-jars-refuse.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+fix: prevent duplicate events with transactional storage (async) updates


packages/react-native/src/storage.ts
```diff
@@ -29,13 +30,35 @@ export class PostHogRNStorage {
     }
   }
 
+  /**
+   * Waits for all pending storage persist operations to complete.
+   * This ensures data has been written to the underlying storage before proceeding.
+   */
+  async waitForPersist(): Promise<void> {
+    while (this._pendingPersistCount > 0) {
+      await new Promise((resolve) => setTimeout(resolve, 1))
```
ioannisj: Can we find a better way to do this rather than polling? Maybe promised-based approach when incrementing/decrementing `_pendingPersistCount`?

packages/core/src/posthog-core-stateless.ts
```diff
@@ -862,6 +862,18 @@ export abstract class PostHogCoreStateless {
...
 
+  /**
+   * Hook that allows subclasses to wait for storage operations to complete.
+   * This is called after queue changes are persisted during flush to ensure
+   * data is safely written to storage before considering events as sent.
+   *
+   * Override this in implementations with async storage (e.g., React Native)
+   * to prevent duplicate events on app crash/restart scenarios.
+   */
+  protected async flushStorage(): Promise<void> {
```
ioannisj, line 873: Should we document that this implementation should not throw? We are not wrapping the call in a try-catch I think so either we try-catch or document here

ioannisj (approved):
Couple of comments from me. This eliminates the window of duplicates but mobile has way too many edge cases that we won't be able to handle

packages/react-native/src/storage.ts
marandaneto: it was a promise approach https://github.com/PostHog/posthog-js/pull/3031/changes/bf9ce13cd13a8d5d25bb27fbb034cee958f127ec
but that would lead to memory leaks
whats your concern with the current approach?

packages/core/src/posthog-core-stateless.ts
marandaneto: this is our sdk impl, i dont see how this could throw if one is no op and the other one is just waiting the promises to resolve?

ioannisj: true, mainly for future uses if any

packages/react-native/src/storage.ts
ioannisj: just a nit, seems cpu intensive to me, even though I don't expect _pendingPersistCount to grow significantly.

marandaneto: i could increase the timeout to 10 at least
this is a message loop btw, the CPU wont be awaiting until the timeout is over, its gonna do some other stuff, async/stuff works this way

marandaneto: technically this should be always 1 cus its single threaded, unless the system decides some promises are more important than others and run out of order which is unlikely, i think

marandaneto: https://github.com/PostHog/posthog-js/pull/3031/changes/3c5444c513d484acd814983f1e6f3a70f6de269c

packages/core/src/posthog-core-stateless.ts
marandaneto: https://github.com/PostHog/posthog-js/pull/3031/changes/d7f80ed4bfd0e15657f03fddb7bbaa180525ea97

PR #3032, opened 2026-02-04
feat(flags): add featureFlagsReloading event for tracking flag reload state

dmarticus:
## Problem

When users call `posthog.group()` to switch group context, feature flags are reloaded from the server. The `useFeatureFlagVariantKey` hook (and similar) updates when the new flags arrive, but there's no way to know when flags are *loading* vs *loaded*. This causes UI flicker when the flag value changes, rather than allowing users to show a proper loading state during the transition.

## Changes

Added a new `featureFlagsReloading` event that fires when feature flags begin reloading (e.g., after `group()`, `identify()`, or `reloadFeatureFlags()` calls). This pairs with the existing `onFeatureFlags` callback to give users full control over loading states:

```js
// Fires when flags start reloading
posthog.on('featureFlagsReloading', () => {
  setIsLoading(true)
})

// Fires when flags are done loading
posthog.onFeatureFlags(() => {
  setIsLoading(false)
})
```

Changes:
- Added `featureFlagsReloading` event emission in `reloadFeatureFlags()`
- Exposed the event via `posthog.on('featureFlagsLoading', callback)`
- Added internal `_emit()` method for cross-module event emission
- Added tests for the new event

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1936,7 +1936,7 @@ export class PostHog implements PostHogInterface {
      *
      * Available events:
      * - `eventCaptured`: Emitted immediately before trying to send an event
-     * - `featureFlagsLoading`: Emitted when feature flags are being reloaded (e.g. after `identify()`, `group()`, or `reloadFeatureFlags()`)
+     * - `featureFlagsReloading`: Emitted when feature flags are being reloaded (e.g. after `identify()`, `group()`, or `reloadFeatureFlags()`)
      *
      * {@label Capture}
      *
@@ -1949,8 +1949,8 @@ export class PostHog implements PostHogInterface {
      *
      * @example
      * ```js
-     * // Track when feature flags are loading to show a loading state
-     * posthog.on('featureFlagsLoading', () => {
+     * // Track when feature flags are reloading to show a loading state
+     * posthog.on('featureFlagsReloading', () => {
      *   console.log('Feature flags are being reloaded...')
      * })
      * ```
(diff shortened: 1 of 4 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -1947,13 +1947,21 @@ export class PostHog implements PostHogInterface {
...
+     * ```
+     *
      * @public
      *
      * @param {String} event The event to listen for.
      * @param {Function} cb The callback function to call when the event is emitted.
      * @returns {Function} A function that can be called to unsubscribe the listener.
      */
-    on(event: 'eventCaptured', cb: (...args: any[]) => void): () => void {
+    on(event: 'eventCaptured' | 'featureFlagsLoading', cb: (...args: any[]) => void): () => void {
```
gustavohstrassburger: Seems odd that this was only necessary for your event, are we just missing proper typing/docs for other events here?

packages/browser/src/posthog-featureflags.ts
```diff
@@ -394,6 +394,9 @@ export class PostHogFeatureFlags {
             return
         }
 
+        // Emit event so consumers know flags are being reloaded
```
gustavohstrassburger, line 397: We are calling the event `featureFlagsLoading`, but this only emits when we reload flags, so at the first load this will not be called, right?

If that is the case, I suggest changing the name of the event, to avoid the confusion. `Is this called when the flags are loaded in the first place?`

dmarticus: oh yeah maybe call it `featureFlagReloading`

packages/browser/src/posthog-core.ts
dmarticus: these are the only two unsubscriber events we have afaik; from the docs:
```
     * Unlike  `onFeatureFlags` and `onSessionId` these are not called when the
     * listener is registered, the first callback will be the next event
     * _after_ registering a listener
```

PR #3034, opened 2026-02-05
feat(surveys): add shown tracking to useThumbSurvey

adboio:
## Problem
the `useThumbSurvey` never tracks `survey shown` events if there is no follow-up.

we might want to know when these surveys were "shown" in the UI.

particularly for LLMA, we want to be able to say "user saw the survey, but did not answer."

## Changes
adds automatic `survey shown` tracking on mount for `useThumbSurvey`, with an option to `disableAutoShownTracking` and optionally manually call `trackShown` instead

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react/src/hooks/useThumbSurvey.ts
+++ b/packages/react/src/hooks/useThumbSurvey.ts
@@ -1,6 +1,6 @@
-import { useState, useCallback, useRef, useMemo, type RefCallback } from 'react'
+import { useState, useCallback, useRef, useMemo, type RefCallback, useEffect } from 'react'
 import { usePostHog } from './usePostHog'
-import { DisplaySurveyType, SurveyPosition } from 'posthog-js'
+import { DisplaySurveyType, SurveyEventName, SurveyEventProperties, SurveyPosition } from 'posthog-js'
 
 export interface UseThumbSurveyOptions {
     /** ID of the target PostHog survey */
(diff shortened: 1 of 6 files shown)


Radu-Raicea (approved):

packages/react/src/hooks/useThumbSurvey.ts
```diff
@@ -89,9 +96,29 @@ export function useThumbSurvey({
...
+    const shownRef = useRef(false)
+    const respondedRef = useRef(false)
+
+    const trackShown = useCallback(() => {
+        if (shownRef.current || !posthog) return
+        shownRef.current = true
+        posthog.capture(SurveyEventName.SHOWN, {
+            [SurveyEventProperties.SURVEY_ID]: surveyId,
+            sessionRecordingUrl: posthog.get_session_replay_url?.(),
+            ...properties,
```
lucasheriques, line 108: i think we're missing a couple of othjer survey releated properties here no? can you double check the `usePopupVisiblity` hook?

lucasheriques (approved):
one comment but looks good overall

packages/react/src/hooks/useThumbSurvey.ts
adboio: yeah we generally include survey name and iteration as well here, but we do not have the survey object at this point

but maybe we can attempt to fetch the survey on mount and then we'd have more data ... we could then also maybe swap from index-based question submissions to question IDs ?

i'll follow up with that!

PR #3039, opened 2026-02-05
fix(err): fix console errors capturing

hpouillot:
## Problem

- console errors (with strings) on not captured anymore because the stack associated contains SDK frames

Fix: https://posthoghelp.zendesk.com/agent/tickets/49104 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/52080)

## Changes

- Skip 1 more frame
- Add playwright test for console errors
- Mock ingestion endpoint during playwright e2e tests

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/entrypoints/exception-autocapture.ts
+++ b/packages/browser/src/entrypoints/exception-autocapture.ts
@@ -68,6 +68,7 @@ const wrapConsoleError = (captureFn: (props: ErrorTracking.ErrorProperties) => v
         const errorProperties = errorPropertiesBuilder.buildFromUnknown(error || event, {
             mechanism: { handled: false },
             syntheticException: new Error('PostHog syntheticException'),
+            skipFirstLines: 2,
         })
         captureFn(errorProperties)
         return originalConsoleError?.(...args)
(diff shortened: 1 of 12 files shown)


ablaszkiewicz (approved):

packages/browser/src/entrypoints/exception-autocapture.ts
paras-fundwave: when objects are passed in args, this turns them into `[object]`

not sure if this needs to be changed, but saw this while debugging so flagging it

hpouillot: Good catch, I updated the code to directly use the object if there is only one argument

PR #3056, opened 2026-02-07
feat(node): Add /edge subpath export for non-standard edge runtimes

richardsolomou:
## Problem

Runtimes like Convex run on V8 isolates (similar to Cloudflare Workers) but don't declare `edge`, `edge-light`, or `workerd` export conditions. This causes `posthog-node`'s export resolution to fall through to the Node.js build, which fails because Node.js APIs aren't available.

Users in these environments have no way to opt into the edge build without patching resolution.

## Changes

Adds a `./edge` subpath export to `posthog-node`'s package.json, allowing consumers to explicitly import the edge build:

```typescript
import { PostHog } from 'posthog-node/edge'
```

This bypasses condition-based resolution entirely and unconditionally provides the edge-compatible build.

### Libraries affected

- [x] posthog-node

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/bright-edge-export.md
+++ b/.changeset/bright-edge-export.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+feat: Add /edge subpath export for non-standard edge runtimes


pauldambra (approved):
LGTM

andrewm4894 (approved):

PR #3057, opened 2026-02-07
feat(nuxt): add enabled option in module config

robinsimonklein:
## Problem

Developers using the Nuxt module often need to disable PostHog in certain environments (testing, local development, staging). Currently, the only way to do this is to conditionally exclude the module from `nuxt.config.ts`, which can lead to runtime errors if code tries to use PostHog composables when the module isn't loaded.

## Changes

 Added an optional `enabled` configuration option to the Nuxt module
 
### Behavior
- When enabled: false: PostHog client is not initialized, all composables return undefined, no events are sent
- When enabled: true or undefined: Normal behavior (backward compatible)

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code (didn't find tests for @posthog/nuxt 👀 )

### If releasing new changes

### DIFF
--- a/packages/nuxt/src/runtime/nitro-plugin.ts
+++ b/packages/nuxt/src/runtime/nitro-plugin.ts
@@ -11,12 +11,8 @@ export default defineNitroPlugin((nitroApp) => {
   const posthogServerConfig = runtimeConfig.posthogServerConfig as PostHogServerConfig
   const debug = posthogCommon.debug as boolean
 
-  // Skip initialization if PostHog is explicitly disabled
-  if (posthogCommon.enabled === false) {
-    return
-  }
-
   const client = new PostHog(posthogCommon.publicKey, {
+    disabled: posthogCommon.enabled === false,
     host: posthogCommon.host,
     ...posthogServerConfig,
   })


ablaszkiewicz:
@robinsimonklein - thanks for the contribution. It makes sense to add an option to disable it. I just think that making posthog `undefined` might cause some serious bugs. Imagine your code does this:
```
posthog.capture(...); // <-- execution will stop at this line with "Cannot read properties of undefined (reading 'capture')" 
someCrucialLogic();
```

I've checked and:
- in Node SDK (this would be nitro plugin) we allow passing `disabled: true` into posthog options,
- in browser SDK (this would be vue plugin) we do not have an explicit disable option but we expose `opt_out_capturing_by_default` - I am checking with the team whether this is the correct option because I am not sure. I'll update this thread once I figure this out.

I'm happy to merge this once we solve this :)

robinsimonklein:
@ablaszkiewicz Good catch on the `disabled` option in the Node SDK! I just pushed the change.

  For the browser SDK, I suspect `opt_out_capturing_by_default` is more for GDPR compliance, but let's wait for the
  team's input.

  My initial thinking was that optional chaining (`posthog?.capture()`) wouldn't be a major issue since `enabled`
  defaults to `true` (no breaking changes). But you're right that it's not ideal for the DX.

PR #3065, opened 2026-02-09
feat(react-native): Add useFeatureFlagResult hook

dustinbyrne:
## Problem

React Native SDK users who need multiple pieces of feature flag information (enabled state, variant key, payload) must call two or three separate hooks (`useFeatureFlagEnabled`, `useFeatureFlagVariantKey`, `useFeatureFlagPayload`), each of which independently subscribes to flag updates. This is verbose and means two/three separate re-renders when flags load.

## Changes

Add a new `useFeatureFlagResult` hook that returns a single `FeatureFlagResult` object containing `key`, `enabled`, `variant`, and `payload` in one call. This is a thin React wrapper around the `posthog.getFeatureFlagResult()` method.

- Subscribes to `onFeatureFlags` for reactive updates, with proper cleanup on unmount
- Accepts an optional `client` parameter to override the context client, matching the pattern used by existing hooks
- Exported from `posthog-react-native`

Tests cover boolean flags, multivariate flags, payloads, missing/undefined flags, re-rendering on flag updates, unsubscribe on unmount, and custom client injection.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/test/useFeatureFlagResult.spec.ts
+++ b/packages/react-native/test/useFeatureFlagResult.spec.ts
@@ -120,10 +120,11 @@ describe('useFeatureFlagResult', () => {
     expect(mockPostHog.getFeatureFlagResult).not.toHaveBeenCalled()
   })
 
-  it('should throw when no client is provided via context or prop', () => {
-    jest.spyOn(console, 'error').mockImplementation(() => {})
-    expect(() => renderHook(() => useFeatureFlagResult('test-flag'))).toThrow(
-      'useFeatureFlagResult requires a PostHog client provided as an argument or via context.'
+  it('should log an error when no client is provided via context or prop', () => {
+    const consoleErrorSpy = jest.spyOn(console, 'error').mockImplementation(() => {})
+    renderHook(() => useFeatureFlagResult('flag'))
+    expect(consoleErrorSpy).toHaveBeenCalledWith(
+      expect.stringContaining('useFeatureFlagResult was called without a PostHog client')
     )
   })
 


packages/react-native/src/hooks/useFeatureFlagResult.ts
```diff
@@ -0,0 +1,20 @@
...
+  const posthog = client || contextClient
+
+  const [result, setResult] = useState<FeatureFlagResult | undefined>(posthog.getFeatureFlagResult(flag))
+
+  useEffect(() => {
+    setResult(posthog.getFeatureFlagResult(flag))
+    return posthog.onFeatureFlags(() => {
+      setResult(posthog.getFeatureFlagResult(flag))
+    })
+  }, [posthog, flag])
```
marandaneto:
```suggestion
  const posthog = client || contextClient

  const [result, setResult] = useState<FeatureFlagResult | undefined>(posthog?.getFeatureFlagResult(flag))

  useEffect(() => {
    setResult(posthog?.getFeatureFlagResult(flag))
    return posthog?.onFeatureFlags(() => {
      setResult(posthog?.getFeatureFlagResult(flag))
    })
  }, [posthog, flag])
```

iirc `usePostHog()` can return undefined if people dont use the provider

ioannisj: existing useFeatureFlag() doesn't seem to do this though. If that's the case should we update that as well?

dustinbyrne: This seems really easy to overlook given the return type doesn't indicate it can be undefined. I've opened another PR to address that: https://github.com/PostHog/posthog-js/pull/3071

marandaneto: yeah i think i saw an issue about this already

PR #3066, opened 2026-02-09
fix: cache invalidation was _too_ aggressive

pauldambra:
in https://github.com/PostHog/posthog-js/pull/3051 we added cache invalidation because we observed cached config being applied for too long

however, if someone has code that does not write the cache timestamp 

then we would check for it
not find it 
and clear their config
but they would _never_ write the cache_timestamp

in that case we should instead write a cache timestamp
so that we can use the config in the case where the parent recorder doesn't know about the cache timestamp

### DIFF
--- a/.changeset/stale-sheep-make.md
+++ b/.changeset/stale-sheep-make.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: remote config cache invalidation was too aggressive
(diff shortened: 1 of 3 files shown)


pauldambra:
* **#3066** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3066?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3066?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3072, opened 2026-02-11
chore(posthog-js): add identity verification

adboio:
## Problem

the initial use-case here is we want users to be able to inject person properties into product tours step content

the current flow is:

- sdk hits tours api
- tour api returns all active tours, with precomputed html content

new proposed flow is:

- sdk hits tours api with distinct ID
- server fetches person properties
- server renders templated tour content with properties
- tour api returns all active tours, with rendered html content that maybe includes person properties

the security concern is that this effectively makes all person properties used in product tours public, if you have the site api key + distinct ID.

see https://posthog.slack.com/archives/C08SCAZ52S3/p1770777242381329

## Changes

introduces a new identity verification concept. the flow will be:

- we provide customers with a secret
- customers use secret to compute HMAC of the user's ID
- customer pass HMAC through `identify`, or new method `setIdentityVerification`
- we can include HMAC in requests to posthog, e.g. the request to the tours API
- in the future, this could be used to validate other sorts of requests if needed :)

main app PR to follow

this PR doesn't actually change any behavior, because no api endpoints (yet) do anything with the verification key

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/gentle-grapes-kiss.md
+++ b/.changeset/gentle-grapes-kiss.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+introduces identity verification concept for some posthog requests
--- a/packages/browser/src/__tests__/posthog-core.identify.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.identify.test.ts
@@ -1,4 +1,4 @@
-import { USER_STATE } from '../constants'
+import { IDENTITY_VERIFICATION_SIGNATURE, USER_STATE } from '../constants'
 import { PostHog } from '../posthog-core'
 import { assignableWindow } from '../utils/globals'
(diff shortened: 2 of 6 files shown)


benjackwhite (changes requested):
just making sure this doesn't get missed - replied internally but i'm vetoing this idea / solution until we've agreed more deeply on the way to go given the security risks

PR #3077, opened 2026-02-11
feat: add @posthog/convex package

richardsolomou:
## Problem

There's no official PostHog integration for [Convex](https://convex.dev) backends. Users who build on Convex currently have no straightforward way to capture analytics events, identify users, manage groups, or evaluate feature flags from their Convex mutations and actions.

## Changes

Adds the `@posthog/convex` package — a Convex component that provides:

- **Event capture** — track events from mutations and actions
- **User identification** — identify and alias users
- **Group analytics** — manage group properties
- **Feature flags** — evaluate feature flags server-side

The package was previously developed in a standalone repo (`posthog-convex`) and is now moved into the posthog-js monorepo. A second commit switches the test runner from vitest to jest to align with the rest of the monorepo.

Includes a full example app, comprehensive tests, and documentation.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -44,7 +44,9 @@
   },
   "devDependencies": {
     "@babel/cli": "^7.19.3",
+    "@babel/preset-env": "catalog:",
     "@babel/preset-react": "catalog:",
+    "@babel/preset-typescript": "catalog:",
     "@babel/plugin-transform-class-properties": "^7.27.1",
     "@babel/plugin-transform-private-methods": "^7.27.1",
     "@babel/plugin-transform-private-property-in-object": "^7.27.1",
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -23,6 +23,8 @@
   },
   "devDependencies": {
     "@ai-sdk/provider": "^3.0.8",
+    "@babel/preset-env": "catalog:",
+    "@babel/preset-typescript": "catalog:",
     "@posthog-tooling/rollup-utils": "workspace:*",
     "@posthog-tooling/tsconfig-base": "workspace:*",
     "@types/jest": "catalog:",
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -40,7 +40,9 @@
         }
     },
     "devDependencies": {
+        "@babel/preset-env": "catalog:",
         "@babel/preset-react": "catalog:",
+        "@babel/preset-typescript": "catalog:",
         "@posthog-tooling/rollup-utils": "workspace:*",
         "@rollup/plugin-inject": "^4.0.2",
         "@rollup/plugin-replace": "^2.3.4",
--- a/packages/web/package.json
+++ b/packages/web/package.json
@@ -29,6 +29,8 @@
     "@posthog/core": "workspace:*"
   },
   "devDependencies": {
+    "@babel/preset-env": "catalog:",
+    "@babel/preset-typescript": "catalog:",
     "@posthog-tooling/tsconfig-base": "workspace:*",
     "@posthog-tooling/rollup-utils": "workspace:*",
     "@types/jest": "catalog:",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -212,6 +212,12 @@ importers:
       '@ai-sdk/provider':
         specifier: ^3.0.8
         version: 3.0.8
+      '@babel/preset-env':
+        specifier: 'catalog:'
+        version: 7.28.5(@babel/core@7.28.5)
+      '@babel/preset-typescript':
+        specifier: 'catalog:'
+        version: 7.28.5(@babel/core@7.28.5)
       '@posthog-tooling/rollup-utils':
         specifier: workspace:*
         version: link:../../tooling/rollup-utils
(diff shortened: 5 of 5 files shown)


.changeset/cold-dogs-allow.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/convex': minor
```
marandaneto, line 2: will that be 0.1.0?

packages/convex/example/convex/_generated/api.d.ts
```diff
@@ -0,0 +1,53 @@
+/* eslint-disable */
```
marandaneto: i'd move the example folder to https://github.com/PostHog/posthog-js/tree/main/examples

packages/react-native/package.json
```diff
@@ -46,6 +46,7 @@
     "@babel/plugin-transform-class-properties": "^7.27.1",
     "@babel/plugin-transform-private-methods": "^7.27.1",
     "@babel/plugin-transform-private-property-in-object": "^7.27.1",
+    "@babel/preset-react": "catalog:",
```
marandaneto: is that needed?

marandaneto:
0 exp. with convex so maybe someone else could take a look?

.changeset/cold-dogs-allow.md
richardsolomou: Yep. Plan is to keep it at 0.x until the API is solidified. Will promote to 1.0.0 once the API is stable.

packages/react-native/package.json
richardsolomou: Temporary fix for a broken test from a different PR. Reverted in 12f6e923!

packages/convex/example/convex/_generated/api.d.ts
richardsolomou: Done! ✅

packages/react-native/package.json
richardsolomou: Turns out it's necessary after all. The react-native package uses @babel/preset-react in its babel config but never declares it as a dependency, so it basically worked on main branch through hoisting. This package's addition shifted the dependency tree due to all the new deps, which broke the hoisting.

packages/convex/README.md
```diff
@@ -0,0 +1,253 @@
...
+The official [PostHog](https://posthog.com) component for [Convex](https://convex.dev). Capture events, identify users, manage groups, and evaluate feature flags — all from your mutations and actions.
+
+Found a bug? Feature request? [File it here](https://github.com/PostHog/posthog-js/issues).
+
+## 🚀 Quick Start
+
+Install the package:
+
+```sh
+npm install @posthog/convex
```
rafaeelaudibert: Suggest using `pnpm` since that's what we use here

packages/convex/README.md
```diff
@@ -0,0 +1,253 @@
...
+  distinctId: "user_123",
+  alias: "anonymous_456",
+});
+```
+
+All of the above methods schedule the PostHog API call asynchronously via `ctx.scheduler.runAfter`, so they return immediately without blocking your mutation or action.
+
+## 🚩 Feature Flags
+
+Feature flag methods evaluate flags by calling the PostHog API and returning the result. They require an **action** context (they use `ctx.runAction` internally).
```
rafaeelaudibert, line 150: Doesn't this mean this will be terribly slow since we'll always trigger a new action?

packages/convex/src/client/_generated/_ignore.ts
rafaeelaudibert, line 1: Use `.gitignore` instead of this

packages/convex/src/component/lib.test.ts
rafaeelaudibert, line 1: Can this be removed?

packages/convex/src/component/lib.ts
```diff
@@ -0,0 +1,259 @@
...
+    return result ?? null
+  },
+})
+
+export const isFeatureEnabled = action({
+  args: featureFlagArgs,
+  handler: async (_ctx, args) => {
+    const client = createClient(args.apiKey, args.host)
+    const result = await client.isFeatureEnabled(args.key, args.distinctId, featureFlagOptions(args))
+    await client.shutdown()
```
rafaeelaudibert, line 165: Does it really make sense to always call shutdown here? Is there something we can do in Convex where this keeps running in the background and we're just calling out to it rather than starting a new client and killing it every single time?

packages/convex/src/component/lib.ts
```diff
@@ -0,0 +1,259 @@
...
+
+export const getFeatureFlag = action({
+  args: featureFlagArgs,
+  handler: async (_ctx, args): Promise<FeatureFlagValue | null> => {
+    const client = createClient(args.apiKey, args.host)
+    const result = await client.getFeatureFlag(args.key, args.distinctId, featureFlagOptions(args))
+    await client.shutdown()
+    return result ?? null
+  },
+})
```
rafaeelaudibert, line 158: I wouldn't consider this integration complete until we're running local evaluation in Convex. This means we'd keep polling feature flags config from the backend and storing them locally in a new table managed by us. We'd then have a cron that keeps polling them. And then calling out to `getFeatureFlag` simply uses the flags that are stored in the database rather than querying the backend every single time - which will be very slow.

rafaeelaudibert: Of course, we don't need that for v0.1.0 but we'll need that for v1.0.0

PR #3079, opened 2026-02-11
chore(sdk): update z-index heirarchy for tours > surveys > conversations

adboio:
## Problem

interactions between tours, surveys, and conversations are not intentional / unexpected

## Changes

as a first pass at not breaking things, this PR simply updates z-index values to support a hierarchy of tours (2147483646, tooltips +1) > surveys (2147483645) > conversations (2147483644)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/surveys/survey.css
+++ b/packages/browser/src/extensions/surveys/survey.css
@@ -5,7 +5,7 @@
         'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
     --ph-survey-box-padding: 20px 24px;
     --ph-survey-max-width: 300px;
-    --ph-survey-z-index: 2147483647;
+    --ph-survey-z-index: 2147483645; /* Z_INDEX_SURVEYS from constants.ts */
     --ph-survey-border-color: #dcdcdc;
     --ph-survey-border-bottom: 1.5px solid var(--ph-survey-border-color);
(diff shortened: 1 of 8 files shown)


packages/browser/src/extensions/surveys/survey.css
```diff
@@ -5,7 +5,7 @@
         'Apple Color Emoji', 'Segoe UI Emoji', 'Segoe UI Symbol';
     --ph-survey-box-padding: 20px 24px;
     --ph-survey-max-width: 300px;
-    --ph-survey-z-index: 2147483647;
+    --ph-survey-z-index: 2147483645;
```
lucasheriques: can we have those numbers on constants so we don't need to remember? then we can add a comment in the const itself, so we know why are they like this

adboio: haha yes excellent point

veryayskiy (approved):
🚀

lucasheriques (approved):

PR #3083, opened 2026-02-12
refactor(ai): remove static Buffer imports for edge compatibility

richardsolomou:
## Problem

`@posthog/ai` uses `import { Buffer } from 'buffer'` in two files, which prevents it from running in edge runtimes (Cloudflare Workers, Vercel Edge Functions, Deno Deploy, Convex, etc.). Since `posthog-node` already supports edge via `posthog-node/edge`, the AI package should be edge-compatible too.

This will be especially useful when using `withTracing` for the Vercel AI SDK in `@convex-dev/agent` to integrate PostHog LLM Analytics with Convex AI Agent.

## Changes

Remove all static `Buffer` imports and replace with universal alternatives:

- **Binary-to-base64 conversion** (`utils.ts`): Runtime `typeof Buffer` check — uses the fast native `Buffer` path in Node.js, falls back to `btoa` loop in edge runtimes. `instanceof Uint8Array` catches both `Uint8Array` and `Buffer` (since `Buffer extends Uint8Array`).
- **UTF-8 string sanitization** (`utils.ts`): `Buffer.from(str, 'utf8').toString('utf8')` replaced with `TextEncoder`/`TextDecoder` roundtrip — same lone-surrogate sanitization behavior, available in all runtimes.
- **Byte length check** (`middleware.ts`): `Buffer.byteLength()` replaced with `TextEncoder().encode().byteLength` — standard WinterCG API.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/ai/tests/tools.test.ts
+++ b/packages/ai/tests/tools.test.ts
@@ -341,5 +341,32 @@ describe('Tool Handling Tests', () => {
         },
       ])
     })
+
+    it('should convert Uint8Array inline data to base64', () => {
+      const bytes = new Uint8Array([72, 101, 108, 108, 111]) // "Hello"
+      const response = {
+        candidates: [
+          {
+            content: {
+              parts: [{ inlineData: { mimeType: 'audio/pcm', data: bytes } }],
+            },
+          },
+        ],
+      }
+
+      const formatted = formatResponseGemini(response)
+      expect(formatted).toEqual([
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'audio',
+              mime_type: 'audio/pcm',
+              data: 'SGVsbG8=', // base64 of "Hello"
+            },
+          ],
+        },
+      ])
+    })
   })
 })


packages/ai/src/utils.ts
```diff
@@ -279,9 +278,17 @@ export const formatResponseGemini = (response: any): FormattedMessage[] => {
             const mimeType = part.inlineData.mimeType || 'audio/pcm'
             let data = part.inlineData.data
 
-            // Handle binary data (Buffer/Uint8Array -> base64)
-            if (data instanceof Uint8Array || Buffer.isBuffer(data)) {
-              data = Buffer.from(data).toString('base64')
+            // Handle binary data (Uint8Array/Buffer -> base64)
```
andrewm4894, line 281: do we have tests for each path in the tree here? just want to sense check this one

```
The new edge-runtime path in formatResponseGemini converts Uint8Array payloads via btoa when Buffer is unavailable, but this branch is not exercised by current tests (unit coverage still shows lines 283-290 uncovered). Please add a test that runs with Buffer unavailable and asserts the exact base64/sanitization output for inlineData to avoid edge-only regressions.
```

andrewm4894: not blocking but just a comment

PR #3084, opened 2026-02-12
feat: react native error boundary

ablaszkiewicz:
## Problem

People expect us to provide `PostHogErrorBoundary` in react-native code. 

Another problem is that you are unable to run example app following the `README.md`.

## Changes

I copied our react error boundary, adjusted it and made sure that it works. There are some gotchas:
- in dev mode react will propagate errors to the global handler even if they were swallowed by some error boundary. This is expected,
- when you have both console capture enabled and error boundary, exception will be captured twice. That is because react by default logs errors to console. Sentry has that limitation and they have it documented so we will do the same for now.

## Why dedup can't work

I've tried what we did in node - we attach a property to an error object and dedup based on that. It is not possible because stuff passed to console is an already preformatted string so best we could do is message based dedup which is far from ideal and it's just gambling. Leaving this as it is and will add note in docs.

## Tests

When running tests, I disabled console capture in posthog error tracking settings. I used this:
```
<PostHogErrorBoundary fallback={ErrorFallback} additionalProperties={{ attachedFromPostHogErrorBoundary: true }}>
```

### WEB

https://github.com/user-attachments/assets/f17aa3bd-74b2-4ee5-a963-e91b34297de7

### IOS

On IOS we see that error overlay screen. From what I read this is the default behavior in react native ios dev builds. In production builds it doesn't happen.

https://github.com/user-attachments/assets/cc64dfd9-efd3-43a5-ae16-5d227d9e5fa7

### ANDROID

Same stuff on android with the overlay screen

https://github.com/user-attachments/assets/86eaeb51-9968-4646-bdcf-3bc053b404d3

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

### DIFF
--- a/examples/example-expo-53/app/posthog.tsx
+++ b/examples/example-expo-53/app/posthog.tsx
@@ -24,7 +24,7 @@ export const posthog = new PostHog(process.env.EXPO_PUBLIC_POSTHOG_PROJECT_API_K
             console: ['error', 'warn'],
         },
     },
-    persistence: 'memory',
+    // persistence: 'memory',
     // if using WebView, you have to disable masking for text inputs and images
     // sessionReplayConfig: {
     //   maskAllTextInputs: false,
--- a/examples/example-expo-53/pnpm-lock.yaml
+++ b/examples/example-expo-53/pnpm-lock.yaml
@@ -3140,7 +3140,7 @@ packages:
       react-native: '*'
 
   posthog-react-native@file:../../target/posthog-react-native.tgz:
-    resolution: {integrity: sha512-667Jtue0VPDCP++bto1r8NEq5XR2c4o05GZAD7Bo7UaepgdsLIhcyQrM9fveB4NTelqzUvQQmpDYRx+V7ukilA==, tarball: file:../../target/posthog-react-native.tgz}
+    resolution: {integrity: sha512-zg7JKLPu8M6cw+8xcKMevOjrY/lpm+uWQPwrUXI7Rmkbbok1OdFOhvbgupx4ahSbfUZ8nzxJgQ9nWlKHtVMV1w==, tarball: file:../../target/posthog-react-native.tgz}
     version: 4.32.0
     peerDependencies:
       '@react-native-async-storage/async-storage': '>=1.0.0'


examples/example-expo-53/package.json
```diff
@@ -36,8 +36,8 @@
         "expo-web-browser": "~14.2.0",
         "posthog-react-native": "*",
         "posthog-react-native-session-replay": "^1.2.0",
-        "react": "19.2.1",
-        "react-dom": "19.2.1",
+        "react": "19.0.0",
```
ablaszkiewicz, line 39: I had to to this. Otherwise I got an error when using `npx expo run:ios` indicating that there is a mismatch between `react-native-renderer` and `react` versions. 

There is no `react-native-renderer` dependency in `package.json` + I didn't find it in pnpm lock so I decided to downgrade react.

It works fine after that change

packages/react-native/src/PostHogErrorBoundary.tsx
```diff
@@ -0,0 +1,83 @@
...
+  componentDidCatch(error: unknown, errorInfo: React.ErrorInfo): void {
+    const { additionalProperties } = this.props
+    let currentProperties
+    if (isFunction(additionalProperties)) {
+      currentProperties = additionalProperties(error)
+    } else if (typeof additionalProperties === 'object') {
+      currentProperties = additionalProperties
+    }
+    const { client } = this.context
+    client.captureException(error, currentProperties)
```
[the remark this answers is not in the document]
ablaszkiewicz: Almost all following comments are about a lack of that exceptionEvent handling. I wanted to start with the simplest possible implementation and having that exceptionEvent propagation sounds like a huge overkill.

examples/example-expo-53/package.json
marandaneto: just example so its ok

packages/react-native/src/PostHogErrorBoundary.tsx
```diff
@@ -0,0 +1,83 @@
...
+const isFunction = (f: any): f is (...args: any[]) => any => typeof f === 'function'
+
+export const __POSTHOG_ERROR_MESSAGES = {
+  INVALID_FALLBACK:
+    '[PostHog][PostHogErrorBoundary] Invalid fallback prop, provide a valid React element or a function that returns a valid React element.',
+}
+
+export class PostHogErrorBoundary extends React.Component<PostHogErrorBoundaryProps, PostHogErrorBoundaryState> {
+  static contextType = PostHogContext
+  context!: React.ContextType<typeof PostHogContext>
```
marandaneto, line 31: can context be undefined here?

packages/react-native/src/PostHogErrorBoundary.tsx
marandaneto: i think client can be undefined if not using the posthogprovider, so we should check for that
context here https://github.com/PostHog/posthog-js/pull/3071

packages/react-native/src/PostHogErrorBoundary.tsx
```diff
@@ -0,0 +1,83 @@
...
+          error: state.error,
+          componentStack: state.componentStack,
+        }) as React.ReactNode)
+      : fallback
+
+    if (React.isValidElement(element)) {
+      return element as React.ReactElement
+    }
+    //eslint-disable-next-line no-console
+    console.warn(__POSTHOG_ERROR_MESSAGES.INVALID_FALLBACK)
```
marandaneto: i think we should remove this otherwise we'd capture an error from this as well

packages/react-native/src/PostHogErrorBoundary.tsx
ablaszkiewicz: Done

packages/react-native/src/PostHogErrorBoundary.tsx
ablaszkiewicz: Done

packages/react-native/src/PostHogErrorBoundary.tsx
ablaszkiewicz: Great catch

packages/react-native/src/PostHogErrorBoundary.tsx
```diff
@@ -0,0 +1,83 @@
...
+    })
+  }
+
+  public render(): React.ReactNode {
+    const { children, fallback } = this.props
+    const state = this.state
+
+    if (state.componentStack == null) {
+      return isFunction(children) ? children() : children
+    }
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

PR #3086, opened 2026-02-12
feat: add remote config support for error tracking and capture perfor…

marandaneto:
## Problem

Port of PostHog/posthog-android#415 for the JS SDK.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -1054,7 +1054,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
   async reloadFeatureFlagsAsync(
     sendAnonDistinctId?: boolean
   ): Promise<PostHogFlagsResponse['featureFlags'] | undefined> {
-    return (await this.flagsAsync({ sendAnonDistinctId: sendAnonDistinctId ?? true }))?.featureFlags
+    return (await this.flagsAsync({ sendAnonDistinctId: sendAnonDistinctId }))?.featureFlags
   }
 
   onFeatureFlags(cb: (flags: PostHogFlagsResponse['featureFlags']) => void): () => void {


packages/react-native/src/posthog-rn.ts
```diff
@@ -1249,8 +1276,8 @@ export class PostHog extends PostHogCore {
       maskAllTextInputs = true,
       maskAllImages = true,
       maskAllSandboxedViews = true,
-      captureLog = true,
-      captureNetworkTelemetry = true,
+      captureLog: localCaptureLog = true,
+      captureNetworkTelemetry: localCaptureNetworkTelemetry = true,
```
marandaneto, line 1284: right now these 2 arent so important because they are also fetched by the native sdks (android and ios) but once we also capture local logs directly on react native with monkeypatching and make our own fetch wrapper so people could use it, we'd have remote config right away

packages/core/src/posthog-core.ts
```diff
@@ -716,6 +743,18 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+          // fetchConfig=true means the response contains remote config fields (errorTracking, capturePerformance, etc.).
+          // When called from _remoteConfigAsync -> reloadFeatureFlags, fetchConfig defaults to true,
+          // so this fires once. _remoteConfigAsync skips its own onRemoteConfig call in that case.
+          if (fetchConfig) {
+            try {
+              this.onRemoteConfig(res)
+            } catch (e) {
+              this._logger.error('Error in onRemoteConfig callback:', e)
+            }
+          }
```
dustinbyrne: If flags are quota limited, we won't get this far

marandaneto: good catch, will fix

.gitignore
```diff
@@ -46,3 +46,6 @@ target
 next-env.d.ts
 
 CLAUDE.md
+
+# Ignore Ralph's local configuration directory
+.ralph/
```
ioannisj, line 51: TIL

.changeset/wild-guests-buy.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog-react-native': minor
+'@posthog/core': minor
+---
+
+support remote config for error tracking, session replay capture performance and capture logs
```
ioannisj: Should we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) here?

i’ve noticed some react-native changelogs use type prefixes like `feat:` and `fix:`, while others don’t. Would probably be a good idea to be consistent here

packages/core/src/types.ts
```diff
@@ -225,6 +225,28 @@ export type PostHogRemoteConfig = {
...
+  /**
+   * Error tracking remote config.
+   * Either a boolean (false = disabled) or a map with configuration.
+   * When a map, `autocaptureExceptions` (boolean) controls whether automatic exception capture is enabled remotely.
+   */
+  errorTracking?:
+    | boolean
+    | {
+        [key: string]: JsonType
+      }
```
ioannisj, line 238: nit: this pattern is kind of common in remote config (bool or object), would it be more readable if we use a shared type alias?

packages/core/src/posthog-core.ts
```diff
@@ -716,6 +743,18 @@ export abstract class PostHogCore extends PostHogCoreStateless {
           // Mark that we hit the /flags endpoint so we can capture this in the $feature_flag_called event
           this.setPersistedProperty(PostHogPersistedProperty.FlagsEndpointWasHit, true)
           this.cacheSessionReplay('flags', res)
+
+          // Notify onRemoteConfig when the flags response includes config data.
+          // fetchConfig=true means the response contains remote config fields (errorTracking, capturePerformance, etc.).
+          // When called from _remoteConfigAsync -> reloadFeatureFlags, fetchConfig defaults to true,
+          // so this fires once. _remoteConfigAsync skips its own onRemoteConfig call in that case.
+          if (fetchConfig) {
```
ioannisj: this is a default true param, so for user-triggered `reloadFeatureFlags()` this will also load remote config and, theoretically install/uninstall integrations. I think we should only do this on the first load

packages/react-native/src/posthog-rn.ts
```diff
@@ -281,6 +292,19 @@ export class PostHog extends PostHogCore {
     await this._initPromise
   }
 
+  /**
+   * Called when remote config has been loaded (from either the remote config endpoint or the flags endpoint).
+   * Gates error tracking autocapture based on the remote config response.
+   *
+   * Session replay config (consoleLogRecordingEnabled, capturePerformance.network_timing) is already
+   * cached via PostHogPersistedProperty.RemoteConfig and applied at startup in startSessionReplay().
+   *
```
ioannisj, line 301: Why not react to fresh remote config values for session replay as well? Are we relying on native logic to handle this instead? Otherwise, session replay will need an additional app launch to sync changes (1st to load and cache, 2nd to apply from cached remote config)

.changeset/wild-guests-buy.md
marandaneto: yep, feat makes sense

packages/core/src/types.ts
marandaneto: yes, but then i'd need to refactor half of the code, so i kept as it is
theres also the downside of more types, bigger bundle size

packages/core/src/posthog-core.ts
marandaneto: mm let me check, that wasnt intentional

packages/react-native/src/posthog-rn.ts
marandaneto: > Why not react to fresh remote config values for session replay as well? Are we relying on native logic to handle this instead

yes

PR #3090, opened 2026-02-13
feat(product tours): enable image preload on tour fetch

adboio:
## Problem

product tour steps with images don't load their images until the step actually appears, meaning there's a visual shift/flash when they do load

## Changes

adds image preloading to tours on pageload, plus an option to disable image preload per tour

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/product-tours/product-tours-utils.ts
+++ b/packages/browser/src/extensions/product-tours/product-tours-utils.ts
@@ -1,6 +1,7 @@
 import DOMPurify from 'dompurify'
 
 import {
+    JSONContent,
     ProductTourAppearance,
     ProductTourSelectorError,
     ProductTourStep,
@@ -281,6 +282,20 @@ function escapeHtml(text: string): string {
     const div = document.createElement('div')
     div.textContent = text
     return div.innerHTML
+}
+
+export function getStepImageUrls(step: ProductTourStep): string[] {
+    const urls: string[] = []
+    function walk(node: JSONContent) {
+        if (node.type === 'image' && node.attrs?.src) {
+            urls.push(node.attrs.src)
+        }
+        node.content?.forEach(walk)
+    }
+    if (step.content) {
+        walk(step.content)
+    }
+    return urls
 }
 
 export function getStepHtml(step: ProductTourStep): string {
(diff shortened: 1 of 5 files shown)


packages/browser/src/extensions/product-tours/product-tours-utils.ts
```diff
@@ -283,6 +284,20 @@ function escapeHtml(text: string): string {
...
+        if (node.type === 'image' && node.attrs?.src) {
+            urls.push(node.attrs.src)
+        }
+        node.content?.forEach(walk)
+    }
+    if (step.content) {
+        walk(step.content)
+    }
+    return urls
+}
```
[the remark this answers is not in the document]
adboio: content html is always precomputed from content - the old "content-only" method is deprecated and will be removed

packages/browser/src/extensions/product-tours/product-tours.tsx
```diff
@@ -194,12 +195,26 @@ export class ProductTourManager {
...
 
+    private _preloadTourImages(tours: ProductTour[]): void {
+        const urls = tours
+            .filter((tour) => !tour.disable_image_preload)
+            .flatMap((tour) => tour.steps.flatMap(getStepImageUrls))
+
+        for (const url of urls) {
+            if (!this._preloadedImageUrls.has(url)) {
+                this._preloadedImageUrls.add(url)
+                new Image().src = url
```
rafaeelaudibert, line 213: Is this the most modern way to do this? Why not preload hints in the head?

adboio: hmm we could, but with this approach we also get access to `onload` / `onerror` which lets us actually wait for the images to load

obv not using that in this PR 🤣 but that's the ideal state

this change is just a minor improvement to mitigate the issue, and still really relies on the tour being configured with a delay so the images actually have time to load (without a delay, if a tour is set to show on pageload, we would basically start preload -> 1ms -> start tour... with a delay it's preload -> queue tour -> wait tour delay -> start tour)

need a bit more time to consider how we'd wait for images with error / timeout handling etc

rafaeelaudibert (approved):

PR #3093, opened 2026-02-13
fix(replay): retry session recording start when remote config is stale

TueHaulund:
## Summary

- When `LazyLoadedSessionRecording.start()` is called but the persisted remote config has expired (5-minute TTL), the recorder permanently fails to start for the entire session — rrweb never captures any DOM events
- This was discovered from a customer session where the SDK was re-initialized 18 times (SPA navigation pattern) and the recorder never started despite recording being enabled
- Adds a freshness pre-check in `_onScriptLoaded` that requests a fresh config via `RemoteConfigLoader` before calling `start()`, with a one-shot guard to prevent infinite retry loops
- Extracts `RECORDING_REMOTE_CONFIG_TTL_MS` so the TTL constant is shared between `session-recording.ts` and `lazy-loaded-session-recorder.ts`

## Test plan

- [x] New tests: stale config triggers refresh, retry only fires once, recording starts after fresh config arrives
- [ ] Verify with the affected customer's SPA re-init pattern


pauldambra (approved):
wild!  
bad assumption about the order of events from me :sweat:

PR #3094, opened 2026-02-13
feat(replay): add $snapshot_max_depth_exceeded debug property

TueHaulund:
## Summary
- Add $snapshot_max_depth_exceeded debug property to all $snapshot events
- Defaults to false, latches to true once DOM depth limit is hit during serialization
- Resets on session change so it does not persist across sessions
- Uses wasMaxDepthReached() and resetMaxDepthState() from @posthog/rrweb-record (re-exported through the dependency chain from @posthog/rrweb-snapshot in rrweb 0.0.42 to avoid bundle size increase)
- Bumps all @posthog/rrweb-* deps to 0.0.42

Related: PostHog/posthog-rrweb#130 (max depth limit), PostHog/posthog-rrweb#132 (re-export through chain)

## Test plan
- [x] Test for false on all snapshot captures
- [x] Test for true when depth limit is hit
- [x] Test for reset on session change
- [x] All 88 test suites passing

### DIFF
--- a/packages/browser/src/entrypoints/posthog-recorder.ts
+++ b/packages/browser/src/entrypoints/posthog-recorder.ts
@@ -6,7 +6,12 @@ import { LazyLoadedSessionRecording } from '../extensions/replay/external/lazy-l
 
 assignableWindow.__PosthogExtensions__ = assignableWindow.__PosthogExtensions__ || {}
 assignableWindow.__PosthogExtensions__.rrwebPlugins = { getRecordConsolePlugin, getRecordNetworkPlugin }
-assignableWindow.__PosthogExtensions__.rrweb = { record: rrwebRecord, version: 'v2', wasMaxDepthReached, resetMaxDepthState }
+assignableWindow.__PosthogExtensions__.rrweb = {
+    record: rrwebRecord,
+    version: 'v2',
+    wasMaxDepthReached,
+    resetMaxDepthState,
+}
 assignableWindow.__PosthogExtensions__.initSessionRecording = (ph) => new LazyLoadedSessionRecording(ph)
 
 // we used to put all of these items directly on window, and now we put it on __PosthogExtensions__


.changeset/snapshot-max-depth.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+feat(replay): add $snapshot_max_depth_exceeded debug property and bump rrweb to 0.0.41 with DOM depth limit
```
pauldambra: i guess 42

```suggestion
feat(replay): add $snapshot_max_depth_exceeded debug property and bump rrweb to 0.0.42 with DOM depth limit
```

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1183,6 +1192,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
                     $window_id: snapshotBuffer.windowId,
                     $lib: 'web',
                     $lib_version: Config.LIB_VERSION,
+                    $snapshot_max_depth_exceeded: this._maxDepthExceeded,
```
pauldambra: we drop all unknown properties from snapshot_events iirc
(they're manually mapped in capture)

should this go in the generic debug properties?

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1,4 +1,5 @@
 import type { recordOptions, rrwebRecord as rrwebRecordType } from '../types/rrweb'
+import { wasMaxDepthReached, resetMaxDepthState } from '@posthog/rrweb-record'
```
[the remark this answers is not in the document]
pauldambra: 🤔 are not able to treeshake this module maybe?

are we pulling all of rrweb-record into lazy recorder here?
maybe we need to only grab these things in the entrypoint (where we already import record) and then reexport them there?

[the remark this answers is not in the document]
pauldambra: Image

still getting this

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -142,8 +142,12 @@ const newQueuedEvent = (rrwebMethod: () => void): QueuedRRWebEvent => ({
     attempt: 1,
 })
 
+function getRRWeb() {
```
pauldambra, line 145: surprised the linter didn't complain this has no explicit return type

pauldambra (approved):
didn't run it 🙈 
but this looks correct

PR #3103, opened 2026-02-16
feat(llma): add OTEL manual capture pipeline for AI SDK spans

richardsolomou:
## Problem

We need a manual OpenTelemetry capture path for Vercel AI SDK telemetry in `@posthog/ai`, so teams using `experimental_telemetry` can emit PostHog AI events without relying on model wrappers.

## Changes

- Added OTEL span processing pipeline for `@posthog/ai`:
  - `PostHogSpanProcessor`
  - `createPostHogSpanProcessor`
  - `captureSpan`
- Added default AI SDK OTEL mapper (`aiSdkSpanMapper`) for doGenerate/doStream/doEmbed spans.
- Added OTEL mapper registry and shared OTEL mapper/types interfaces.
- Mapped AI SDK span attributes to PostHog AI generation/embedding payloads:
  - model/provider, input/output, token usage, latency, TTFT, tool definitions, and telemetry metadata.
- Added support in mapper for richer output extraction from telemetry attributes (including media/file-like response shapes where present).
- Exported OTEL entrypoints from package index.
- Added docs section for OTEL + AI SDK usage and custom mappers.
- Added unit tests for OTEL capture/mapping behavior (`tests/otel.test.ts`).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/ai/src/otel/mappers/aiSdk.ts
+++ b/packages/ai/src/otel/mappers/aiSdk.ts
@@ -6,6 +6,15 @@ import type { PostHogSpanMapper, PostHogSpanMapperResult, UsageData } from '../t
 const OTEL_STATUS_ERROR = 2
 const AI_TELEMETRY_METADATA_PREFIX = 'ai.telemetry.metadata.'
 
+type InstrumentationInfo = {
+  version?: unknown
+}
+
+type ReadableSpanWithInstrumentation = ReadableSpan & {
+  instrumentationScope?: InstrumentationInfo
+  instrumentationLibrary?: InstrumentationInfo
+}
+
 function parseJsonValue<T>(value: unknown): T | null {
   if (value === undefined || value === null) {
     return null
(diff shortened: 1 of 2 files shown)


packages/ai/src/otel/processor.ts
```diff
@@ -0,0 +1,34 @@
...
+  constructor(
+    private readonly phClient: PostHog,
+    private readonly options: PostHogTelemetryOptions = {}
+  ) {}
+
+  onStart(_span: Span, _parentContext: Context): void {
+    // no-op
+  }
+
+  onEnd(span: ReadableSpan): void {
```
andrewm4894, line 19: just want to double check this bot comment - eg anything to think about there for potential of losing any in flight stuff? 

```
onEnd calls captureSpan (async) but doesn't track the promise. This means:
  - forceFlush() and shutdown() are no-ops that resolve immediately, even if spans are still being processed. The OTEL SDK calls forceFlush before shutdown — if the process exits, in-flight spans will be silently dropped.
  - Under high throughput, many concurrent promises could pile up with no backpressure.

Consider tracking pending promises and actually awaiting them in forceFlush()
```

packages/ai/src/otel/mappers/aiSdk.ts
```diff
@@ -0,0 +1,777 @@
...
+function buildPosthogProperties(attributes: Record<string, unknown>, operationId: string): Record<string, unknown> {
+  const telemetryMetadata = extractAiSdkTelemetryMetadata(attributes)
+  const finishReasons = toStringArray(parseJsonValue(attributes['gen_ai.response.finish_reasons']))
+  const finishReason = toStringValue(attributes['ai.response.finishReason']) || finishReasons[0]
+  const toolChoice = parseJsonValue(attributes['ai.prompt.toolChoice']) ?? attributes['ai.prompt.toolChoice']
+
+  return {
+    ...telemetryMetadata,
+    $ai_framework: 'vercel',
+    $ai_framework_version: '6',
```
andrewm4894: should this be hardcoded to 6? i _think_ there is a way to get this back out of vercel sdk no? iirc we already doing it in vercel

andrewm4894: @richardsolomou this one almost a blocker i think, expecially if we can get the correct version from the ai sdk itself?

andrewm4894: eg i think the `experimental_telemetry` is around since v4? i might be wrong on that but worth checking

PR #3105, opened 2026-02-17
feat(product tours): support tour content localization

adboio:
## Problem

product tours only support one language

closes https://github.com/PostHog/posthog/issues/47968#issue-3940215543

## Changes

- adds sdk support for tour content translations
- adds a new config field `override_display_language`, which i expect we can also use for surveys, and more in the future

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/types/src/posthog-config.ts
+++ b/packages/types/src/posthog-config.ts
@@ -1390,6 +1390,16 @@ export interface PostHogConfig {
      */
     internal_or_test_user_hostname?: string | RegExp | null
 
+    /**
+     * Display language override for Product Tours.
+     *
+     * Must be a valid BCP 47 language code.
+     *
+     * In the future this will be used for Surveys and other products that
+     * deliver in-app experiences to end-users.
+     */
+    override_display_language?: string | null
+
     // ------- PREVIEW CONFIGS -------
 
     /**
--- a/.changeset/two-ideas-rule.md
+++ b/.changeset/two-ideas-rule.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+add support for product tours localization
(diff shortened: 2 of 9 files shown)


packages/types/src/posthog-config.ts
```diff
@@ -1390,6 +1390,16 @@ export interface PostHogConfig {
      */
     internal_or_test_user_hostname?: string | RegExp | null
 
+    /**
+     * Client (user) language. Must be a valid BCP 47 language code.
```
pauldambra: we can detect language in the browser
but we want to be able to override it?

(not always detecting is a good idea since i often have my phone in italian and some products then switch on every device i have, very annoying)

client is super overloaded as a tool. is it the browser, the person, the SDK instance

and we have `client (user)` in the comment to show us it isn't clear enough

but the comment is helping us here though

packages/types/src/posthog-config.ts
```diff
@@ -1390,6 +1390,16 @@ export interface PostHogConfig {
...
+     *
+     * Used as a language override for Product Tours content localization.
+     *
+     * In the future this will be used for Surveys and other products that
+     * deliver in-app experiences to end-users.
+     */
+    client_language?: string | null
```
pauldambra:
```suggestion
    override_display_language?: string | null
```

i _think_ that's a better name?

packages/types/src/posthog-config.ts
adboio: heard on the naming, happy to change that! the fact that i had to write “(user)” to explain myself should’ve stopped me sooner 🤣

and yeah override has been requested by a few customers for similar reasons - we default to browser lang unless this is provided

packages/types/src/posthog-config.ts
adboio: yep i like this - still generic so we can use it for surveys too 🫡

marandaneto (approved):

PR #3107, opened 2026-02-17
feat: warning on capture('$exception')

ablaszkiewicz:
# NODE

```
async function main2() {
    console.log("Capturing via capture()")
    client.capture({
        event: '$exception'
    })

    console.log("Capturing via captureException()")
    client.captureException(new Error("ewtgeowrintgpiormgrg"), 'asd');


    console.log("Capturing via auto capture")
    throw new Error("auto capture error")
}
```

image

# BROWSER

image

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -398,7 +398,7 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     }
     if (props.event === '$exception' && !props._originatedFromCaptureException) {
       this._logger.warn(
-        'Capturing a `$exception` event via `posthog.captureImmediate(\'$exception\')` is unreliable because it does not attach required metadata. Use `posthog.captureExceptionImmediate(error)` instead, which attaches this metadata by default.'
+        "Capturing a `$exception` event via `posthog.captureImmediate('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureExceptionImmediate(error)` instead, which attaches this metadata by default."
       )
     }
     return this.addPendingPromise(


packages/node/src/client.ts
```diff
@@ -327,6 +327,11 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     if (typeof props === 'string') {
       this._logger.warn('Called capture() with a string as the first argument when an object was expected.')
     }
+    if (props.event === '$exception' && !props._originatedFromCaptureException) {
+      this._logger.warn(
+        "Using `posthog.capture('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureException()` instead, which attaches required metadata automatically."
```
hpouillot:
```suggestion
        "Using `posthog.capture('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureException(error)` instead, which attaches required metadata automatically."
```

packages/node/src/client.ts
```diff
@@ -391,6 +396,11 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     if (typeof props === 'string') {
       this._logger.warn('Called captureImmediate() with a string as the first argument when an object was expected.')
     }
+    if (props.event === '$exception' && !props._originatedFromCaptureException) {
+      this._logger.warn(
+        'Capturing a `$exception` event via `posthog.captureImmediate()` is unreliable because it does not attach required metadata. Use `posthog.captureExceptionImmediate()` instead, which attaches this metadata by default.'
```
hpouillot:
```suggestion
        'Capturing a `$exception` event via `posthog.captureImmediate('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureExceptionImmediate(error)` instead, which attaches this metadata by default.'
```

packages/browser/src/posthog-core.ts
```diff
@@ -1142,6 +1142,12 @@ export class PostHog implements PostHogInterface {
             delete properties?.$current_url
         }
 
+        if (event_name === '$exception' && !options?._originatedFromCaptureException) {
+            logger.warn(
+                "Using `posthog.capture('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureException()` instead, which attaches required metadata automatically."
```
hpouillot:
```suggestion
                "Using `posthog.capture('$exception')` is unreliable because it does not attach required metadata. Use `posthog.captureException(error)` instead, which attaches required metadata automatically."
```

PR #3121, opened 2026-02-19
feat: Ability to retrieve your previous tickets

veryayskiy:
This PR allows users to recover their previous tickets when switching devices or browsers.

# Problem
Right now, if you open a product in another browser or computer, you won't see your previous tickets.

Also, the `widget_session_id` was being regenerated on every page refresh despite being stored in localStorage. Root cause: PostHog's core persistence layer silently drops data when the cookie+localStorage merge fails on large entries, so persistence.props came back empty while the raw localStorage still had the value.

# Solution
Users can retrieve previous conversations via email by:

1. clicking "Retrieve conversations"
2. entering their email
3. and following a restore link.

Moved all conversation data out of PostHog's core persistence blob into its own dedicated ph_conv_<token> localStorage key (same pattern as toolbar/surveys extensions). Includes a one-time migration from the legacy persistence keys. This eliminates the silent data loss entirely.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/conversations/external/components/ConversationsWidget.tsx
+++ b/packages/browser/src/extensions/conversations/external/components/ConversationsWidget.tsx
@@ -639,7 +639,6 @@ export class ConversationsWidget extends Component<WidgetProps, WidgetState> {
         return (
             <div style={styles.widget}>
                 <div style={windowStyle}>
-                    {/* Header */}
                     <div style={styles.header}>
                         <div style={showBackButton ? styles.headerWithBack : styles.headerTitle}>
                             {showBackButton && this._renderBackButton(styles)}


packages/browser/src/extensions/conversations/external/persistence.ts
```diff
@@ -72,186 +87,194 @@ export class ConversationsPersistence {
...
     /**
-     * Load user-provided traits from persistence
+     * One-time migration: copy conversations data from PostHog's main
+     * persistence blob into the dedicated localStorage key, then remove
+     * the old keys from PostHog persistence so they stop bloating it.
      */
-    loadUserTraits(): UserProvidedTraits | null {
-        if (!this._isPersistenceAvailable()) {
-            return null
+    private _migrateFromLegacyPersistence(): void {
```
marandaneto, line 181: do we actually need to do this since its still not a public beta or anything? so its ok to break in order to have a cleaner code?

veryayskiy: Yes, but the moment we deploy on posthog i'll follow up with the cleaning.

Issue #2854, opened 2026-01-07
posthog-node shows "[PostHog] Timed out while shutting down PostHog" messages in dev


ajsharp:
## Bug description

* Calling `shutdown()` creats lots of "Timed out while shutting down Posthog" messages in our dev and CI logs
* We don't see any of these in prod, but we see them in dev and CI

## How to reproduce

```typescript
export async function getBootstrapData(
  flagKeys: string[]
): Promise<PostHogBootstrapData> {
  const distinctId = await getDistinctIdFromCookies();

  // If no flags to fetch, return early with just the distinctId
  if (flagKeys.length === 0) {
    return {
      distinctId,
      featureFlags: {},
      featureFlagPayloads: {},
    };
  }

  const posthog = new PostHog(config.apiKey, {
    host: "https://app.posthog.com",
    // Disable automatic flushing in serverless - we flush manually
    flushAt: 1,
    flushInterval: 0,
  });

  try {
    const result = await posthog.getAllFlagsAndPayloads(distinctId, { flagKeys });

    await posthog.shutdown();

    return {
      distinctId,
      featureFlags: result.featureFlags ?? {},
      featureFlagPayloads: (result.featureFlagPayloads ?? {}) as Record<string, JsonType>,
    };
  } catch (error) {
    console.error("Failed to fetch bootstrap data:", error);
    await posthog.shutdown();
    // Return just the distinctId on error - PostHog will fetch flags client-side
    return {
      distinctId,
      featureFlags: {},
      featureFlagPayloads: {},
    };
  }
}
```

## Related sub-libraries

- [x] posthog-node

## Additional context

Versions: 
- nextjs 15.5.9
- posthog-node 5.18.0

marandaneto:
Provided posthog-node shutdown repro no longer reproduces

Issue #2250, opened 2025-08-26
Flickering when rendering feature flags in Next.JS


mhornbacher:
## Bug description

In various bits of documentation (see context below) it is indicated that bootstrapping would avoid flickering of the page. However in practice this turns out not to be true when using the `useFeatureFlagX` flags. This is due to the client being created within a `useEffect` due to it not being serializable (it has active functionality).

## How to reproduce

1. Create a NextJS application with PostHog
2. Create a "use-client" component that uses any of the `useFeatureFlag` hooks
3. Render it and *load that page first.*

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Additional context

> "showing components without flickering,"
> https://posthog.com/tutorials/nextjs-bootstrap-flags

> "Both approaches avoid flicker"
> https://posthog.com/docs/libraries/next-js#integration

marandaneto:
#3122

Issue #2095, opened 2025-07-17
Opt-Capturing persistence cookie is being ignored


julek-kal:
Hi, I'm using posthog-js with a Vue/Nuxt setup, and I believe there's an issue with how opt_out_capturing_persistence_type is handled during initialization:

```ts
posthog.init('<ph_project_api_key>', {
    opt_out_capturing_persistence_type: 'cookie',
});
```
When I call `opt_out_capturing()`, it partially works — the value is set, but it still gets saved in localStorage instead of a cookie.

From what I can tell (though I'm not a JavaScript expert — just trying to implement cross-subdomain opt-out in existing website), this might be caused by the following logic in consent.ts:

```ts
    private get _storage() {
        if (!this._persistentStore) {
            const persistenceType = this._config.opt_out_capturing_persistence_type
            this._persistentStore = persistenceType === 'localStorage' ? localStore : cookieStore
            const otherStorage = persistenceType === 'localStorage' ? cookieStore : localStore

            if (otherStorage._get(this._storageKey)) {
                if (!this._persistentStore._get(this._storageKey)) {
                    // This indicates we have moved to a new storage format so we migrate the value over
                    this.optInOut(otherStorage._get(this._storageKey) === '1')
                }

                otherStorage._remove(this._storageKey, this._config.cross_subdomain_cookie)
            }
        }

        return this._persistentStore
    }
```
It looks like this getter is called before `posthog.init()`, which leads to this issue — since `this._config.opt_out_capturing_persistence_type` is undefined at that point, the default becomes localStorage.

I've reproduced this in `posthog-js/packages/browser/playground/nextjs`: after setting `opt_out_capturing_persistence_type` to `cookie` in `posthog.ts`, the opt-out flag still ends up in localStorage.

marandaneto:
#3360

Issue #1975, opened 2025-05-23
dynamicIO nextjs 15.3.0+


followbl:
has anyone been able to test out posthog in nextjs `15.3.0+` with 

```ts
  experimental: {
    ppr: true,
    dynamicOnHover: true,
    dynamicIO: true,
    reactCompiler: true,
}
```
with `dynamicIO` enabled

only way I could get posthog to work was this

```tsx
"use client";

import { Suspense, useEffect } from "react";
import dynamic from "next/dynamic";
import posthog from "posthog-js";

import { env } from "~/config/global/env.mjs";
import { isProd } from "~/lib/utils";

const PostHogProviderBase = dynamic(
  () => import("posthog-js/react").then((mod) => mod.PostHogProvider),
  {
    ssr: false,
    loading: () => null,
  },
);

export function PostHogProvider({ children }: { children: React.ReactNode }) {
  useEffect(() => {
    if (isProd) {
      posthog.init(env.NEXT_PUBLIC_POSTHOG_KEY, {
        api_host: env.NEXT_PUBLIC_POSTHOG_HOST,
        capture_pageview: false, // Disable automatic pageview capture, as we capture manually
        capture_pageleave: true,
      });
    }
  }, []);

  return (
    <Suspense>
      <PostHogProviderBase client={posthog}>{children}</PostHogProviderBase>
    </Suspense>
  );
}
```

believe me when I say I tried everything. Posthog uses new Date() and or Math.random() which aren't allowed in pre-rendering.

It recommends to opt out to use `await connection()` from `next/server` but that did not work here...

I know it's early, but I would imagine this will be coming down the pipeline in the near future, would be great to get some clarity on what the recommended loading pattern for 'ph' is with `dynamicIO`

marandaneto:
Addressed by first-class @posthog/next support from PR #3122

PR #3122, opened 2026-02-19
feat(next): Add first class support for Next.js

dustinbyrne:
## Problem

Setting up PostHog in a Next.js app currently requires ~100 lines of boilerplate spread across 4+ files, and it's still easy to get wrong. The documented patterns lead to real, persistent problems that users hit constantly:

**Timing issues** : The current `useEffect`-based initialization means posthog-js isn't ready when child components render. Consent banners, flag hooks, and `get_session_id()` all return `undefined` on first render ([#1925](https://github.com/PostHog/posthog-js/issues/1925), [#2250](https://github.com/PostHog/posthog-js/issues/2250), [#714](https://github.com/PostHog/posthog-js/issues/714)).

**Client/server identity mismatch** : There's no built-in way to share identity between server and client. Users manually create a posthog-node client, call `shutdown()` (which produces timeout warnings — [#2854](https://github.com/PostHog/posthog-js/issues/2854)), and wire bootstrap data through by hand. Even then, it's easy to configure flags that flicker on navigation ([#2250](https://github.com/PostHog/posthog-js/issues/2250), [#43980](https://github.com/PostHog/posthog-js/issues/3980)).

**Consent/storage confusion** : Server-side code has no awareness of opt-out state, so it evaluates flags and captures events for users who have opted out. With `useEffect`-based initialization, consent checks via `usePostHog` hooks run before the client is initialized, reading consent state from the wrong storage type ([#2095](https://github.com/PostHog/posthog-js/issues/2095)).

**Framework compatibility** : Each Next.js release brings new breakage: hydration errors from injected `<script>` tags ([#1645](https://github.com/PostHog/posthog-js/issues/1645)), `ssr: false` no longer allowed in server components ([#26016](https://github.com/PostHog/posthog/issues/26016)), Turbopack HMR failures ([#1585](https://github.com/PostHog/posthog-js/issues/1585)), `dynamicIO` incompatibility ([#1975](https://github.com/PostHog/posthog-js/issues/1975)), and PostHog's fetch headers breaking Next.js caching ([#1529](https://github.com/PostHog/posthog-js/issues/1529)).

**Missing proxy story**: No first-class way to route SDK traffic through the app's own domain, which is the PostHog/posthog#1 recommended setup for ad-blocker resilience.

## Changes

Introduces `@posthog/next`, a new first-class Next.js integration package that addresses common configuration patterns.

### Architecture

**Middleware** runs on every request and:
- Seeds a PostHog identity cookie with a UUIDv7 anonymous ID on first visit, so server and client share the same `distinct_id` from the very first render
- Optionally proxies SDK API calls (`/ingest/*` → PostHog ingest) for ad-blocker resilience
- Respects consent state and skips cookie seeding when the user has opted out
- Composes with existing middleware via the `response` option

**PostHogProvider** (a React Server Component) reads the identity cookie and:
- When `bootstrapFlags` is enabled, evaluates flags server-side via `posthog-node` and passes results to the client as bootstrap data. Hooks return real values immediately, no flicker
- Skips flag evaluation entirely when the user has opted out
- Is **static-safe by default**. It does not call `cookies()` unless `bootstrapFlags` is enabled, so it works with static generation, ISR, and PPR

**Client initialization** is eager (during render, not in a `useEffect`), so child components and hooks see a fully configured posthog-js instance immediately. This eliminates the entire class of "undefined on first render" bugs and race conditions.

**Server utilities** (`getPostHog()`) read the same identity cookie and scope the `posthog-node` client per-request via `enterContext()`, so server-side events and flag evaluations are attributed to the correct user without manual wiring.

**Automatic flushing on Vercel**: The server client automatically detects `@vercel/functions.waitUntil` at module load and passes it to `posthog-node`, ensuring captured events and flag evaluations are flushed before serverless functions terminate, no manual `shutdown()` or `flush()` calls needed.

**Pages Router** support via `@posthog/next/pages` with `PostHogProvider`, `PostHogPageView`, and `getServerSidePostHog()`.

### Entry points

| Import path | Environment | Purpose |
|---|---|---|
| `@posthog/next` | Client + Server | Provider, PageView, hooks |
| `@posthog/next/pages` | Client + Server | Pages Router support |

### What's included

- Full App Router example with auth, consent banner, server flags, SSR bootstrap, and middleware proxy
- Unit tests for all components, cookie parsing, middleware, and server utilities
- README and detailed USAGE.md documentation

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/next/package.json
```diff
@@ -0,0 +1,91 @@
...
+            "edge-light": "./dist/index.edge.js",
+            "edge": "./dist/index.edge.js",
+            "worker": "./dist/index.edge.js",
+            "browser": "./dist/index.client.js",
+            "react-server": "./dist/index.js",
+            "default": "./dist/index.client.js"
+        },
+        "./pages": {
+            "types": "./dist/pages.d.ts",
+            "default": "./dist/pages.js"
```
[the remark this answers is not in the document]
dustinbyrne: Stale PR description in this case, resolved.

.changeset/free-candies-cut.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/next': minor
```
marandaneto, line 2: since its a new package, you'll need to release a 0.0.1 by hand and then configure OIDC via npm, lemme know if you need any help

packages/next/src/app/PostHogProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+import { cookies } from 'next/headers'
+import type { PostHogOptions } from 'posthog-node'
+import { getOrCreateNodeClient } from '../server/nodeClientCache'
+import { NEXTJS_CLIENT_DEFAULTS, resolveApiKey } from '../shared/config'
+import { readPostHogCookie, isOptedOut } from '../shared/cookie'
+
+type AllFlagsOptions = {
+    groups?: Record<string, string>
+    personProperties?: Record<string, string>
+    groupProperties?: Record<string, Record<string, string>>
```
marandaneto, line 14: arent they <string, any> instead?

packages/next/src/app/PostHogProvider.tsx
```diff
@@ -0,0 +1,136 @@
...
+        ...NEXTJS_CLIENT_DEFAULTS,
+        ...clientOptions,
+        ...(host ? { api_host: host } : {}),
+    }
+
+    let bootstrap: BootstrapConfig | undefined
+
+    if (bootstrapFlags) {
+        try {
+            bootstrap = await evaluateFlags(apiKey, resolvedOptions, bootstrapFlags, serverOptions)
```
marandaneto, line 96: evaluating flags here right away sync. might slowdown the page start, should this be async instead?

packages/next/src/client/ClientPostHogProvider.tsx
```diff
@@ -0,0 +1,55 @@
...
+ * We initialize posthog-js eagerly during render (client-side only) rather than
+ * deferring to a useEffect. React fires effects bottom-up, so child useEffects
+ * (e.g. a consent banner) would access the posthog instance before the parent
+ * provider's useEffect calls init(). By initializing during render and passing
+ * the `client` prop, we guarantee the instance is fully configured before any
+ * child code runs.
+ */
+export function ClientPostHogProvider({ apiKey, options, bootstrap, children }: ClientPostHogProviderProps) {
+    if (!apiKey) {
+        throw new Error('[PostHog Next.js] apiKey is required')
```
marandaneto: I'm not a big fan of throwing errors, a misconfiguration in the SDK (e.g., user forgot to set an env. var.) and their deploy will break all their users, much better if we just degrade safelly like log out and noop

packages/next/src/pages/PostHogPageView.tsx
```diff
@@ -0,0 +1,41 @@
...
+ *   )
+ * }
+ * ```
+ */
+export function PostHogPageView() {
+    const router = useRouter()
+    const posthog = usePostHog()
+
+    useEffect(() => {
+        if (!posthog || !router.isReady) {
```
marandaneto, line 33: can we actually delay capturing the event here if router.isReady isnt ready (until its ready)? otherwise we will just lose this event unless use effect will be triggered again once its ready, not sure if thats the case

packages/next/src/app/PostHogProvider.tsx
dustinbyrne: Yeah, it should be - or at least `Record<string, JsonType>`

This doesn't match the type in `posthog-node` though, so I left it as `<string, string>`. It's at least a backwards compatible change to widen that type afterwards. I'll follow up with a change to `posthog-node` to fix the type.

marandaneto: Ah ok, node is wrong then, it should be jsontype all the way so it will be back compatible right cus string is a valid jsontype

packages/next/src/client/ClientPostHogProvider.tsx
dustinbyrne: oof I've been making an effort to avoid this - fixed in https://github.com/PostHog/posthog-js/pull/3122/commits/f7d0266aeea63b8cb1f9ae8a381f84e70556483c

packages/next/src/pages/PostHogPageView.tsx
dustinbyrne: It's in the dependencies of the useEffect, so it'll be re-run should `router.isReady` be false initially but become true. I've added a test to validate this behavior in https://github.com/PostHog/posthog-js/pull/3122/commits/ef0bef6330bc7d3bcbcd57ab661fdc6941753073

packages/next/src/app/PostHogProvider.tsx
dustinbyrne: It is async. This _does_ force the page to render dynamically, so it's behind an opt-in `bootstrapFlags` toggle:
```tsx
<PostHogProvider bootstrapFlags>
  ...
</PostHogProvider>
```

Without `bootstrapFlags`, we won't resolve flags server-side in order to bootstrap the client. It's useful in that it'll prevent flickering as flags load or change on the client, but it can't be used on a static page.

marandaneto: Gotcha, i meant sync because we await the call

marandaneto (approved):
noice!

mhornbacher:
Thank you for this! 🙏

PR #3131, opened 2026-02-23
fix: anonymous users in support

veryayskiy:
## Problem

In PostHog instance we have 2 instances, one by the posthog-js and another one by the toolbar. Because of that, for us, sometimes the toolbar instance got initialized first and mess with distinct id for conversations

## Changes

Do not initialize the conversations for the toolbar instance. Only for the posthog one.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/olive-roses-leave.md
+++ b/.changeset/olive-roses-leave.md
@@ -1,5 +1,5 @@
 ---
-'posthog-js': minor
+'posthog-js': patch
 ---
 
 Fix issue with toolbar interferring distinct id in conversations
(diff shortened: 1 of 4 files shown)


.changeset/olive-roses-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': minor
```
marandaneto: patch instead?

packages/browser/src/extensions/conversations/posthog-conversations.ts
```diff
@@ -74,6 +74,12 @@ export class PostHogConversations {
         if (this._instance.config.disable_conversations) {
             return
         }
+        // The toolbar's internal PostHog instance must not own the conversations
+        // manager — its distinct_id is always anonymous and would be sent instead
+        // of the identified user's ID.
+        if (this._instance.config.name === 'ph_toolbar_internal') {
```
marandaneto: lets make this a const or even an util method like isToolbarInstance(instance), your call

PR #3132, opened 2026-02-23
feat: Support .register (super properties) in Nodejs

benjackwhite:
## Problem

Noticed that the expected super properties from [PostHog repo implementation](https://github.com/PostHog/posthog/blob/master/nodejs/src/utils/posthog.ts#L21) just don't show up at all. Turns out this never worked.

It makes "sense" in the context that register is meant to be a stateful concept. But it also makes sense for the SDK to support as a global concept (we have `posthoganalytics.super_properties` in python for example).

## Changes
* Adds test coverage
* Fixes the issue

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/tender-mice-call.md
+++ b/.changeset/tender-mice-call.md
@@ -2,4 +2,4 @@
 'posthog-node': minor
 ---
 
-Fixes an issue where super properties added with `.register` were not included in all subsequent events
+Officially adds support for super properties added with `.register` (it was available as a function but did not actually work previously)


.changeset/tender-mice-call.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-node': minor
```
marandaneto, line 2: should that be a patch instead? method register existed already

benjackwhite: I actually couldn't decide what makes sense here:
1. Major - makes sense as its been broken so long that anyone using it might not realise and suddenly have a big breakage when we fix it
2. Minor - its undocumented so this is adding a new "feature" from that perspective
3. Patch - it fixes something that at least in code was always there


🤷 no strong feelings - can swap to patch

marandaneto: Yeah, I'd lean towards 2 or 3 since it's a back-compatible feat/bugfix.
PR is approved so your call, thx for fixing!

PR #3134, opened 2026-02-23
feat(react-native): support session replay sampleRate config

marandaneto:
## Problem

React Native session replay did not support `sessionRecording.sampleRate` from remote config (as added in posthog-android#437), and local sample rate handling needed clear precedence and validation.

depends on https://github.com/PostHog/posthog-android/pull/437/ and iOS WIP

## Changes

- Added `sampleRate?: number` to `PostHogSessionReplayConfig`.
- Added `getRemoteConfigNumber(...)` to parse numeric/string remote config values (e.g. `"0.5"`).
- Updated `initializeSessionReplayNative` to:
  - include `sampleRate` in `sdkReplayConfig` passed to native session replay init,
  - enforce local `sessionReplayConfig.sampleRate` precedence over remote config,
  - validate sample rate is a finite number between `0` and `1`,
  - ignore invalid values and log warnings.
- Added unit tests for numeric remote config parsing in `remote-config.spec.ts`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/replay/session-recording.ts
+++ b/packages/browser/src/extensions/replay/session-recording.ts
@@ -11,7 +11,7 @@ import { RemoteConfigLoader } from '../../remote-config'
 import { Properties, RemoteConfig, SessionRecordingPersistedConfig, SessionStartReason } from '../../types'
 import { type eventWithTime } from './types/rrweb-types'
 
-import { isNullish, isUndefined, isValidSampleRate } from '@posthog/core'
+import { isNullish, isNumber, isUndefined, isValidSampleRate } from '@posthog/core'
 import { createLogger } from '../../utils/logger'
 import {
     assignableWindow,
@@ -152,7 +152,7 @@ export class SessionRecording {
         if (isNullish(rate)) {
             return null
         }
-        const parsed = typeof rate === 'number' ? rate : parseFloat(rate as string)
+        const parsed = isNumber(rate) ? rate : parseFloat(rate as string)
         if (!isValidSampleRate(parsed)) {
             logger.warn(`${source} must be between 0 and 1. Ignoring invalid value:`, rate)
             return null


pnpm-workspace.yaml
```diff
@@ -47,6 +47,7 @@ minimumReleaseAgeExclude:
   - "@posthog/rrweb-snapshot@0.0.42"
   - "@posthog/rrdom@0.0.42"
   - "@posthog/cli"
+  - "posthog-react-native-session-replay"
```
marandaneto, line 50: @PostHog/team-security is that ok? otherwise we cannot ever bump our own dependency right away

packages/react-native/src/utils.ts
```diff
@@ -69,3 +69,42 @@ export function getRemoteConfigBool(
...
+ * Remote config values may be either numbers or numeric strings.
+ *
+ * @param field The remote config field (e.g. `response.sessionRecording`)
+ * @param key The key to read (e.g. `'sampleRate'`)
+ */
+export function getRemoteConfigNumber(
+  field: boolean | { [key: string]: JsonType } | undefined,
+  key: string
+): number | undefined {
+  if (field == null || typeof field === 'boolean' || typeof field !== 'object') {
```
dustinbyrne: Minor nit: `typeof field === 'boolean'` is redundant

pnpm-workspace.yaml
Piccirello: Yes definitely, that's alright. If we get owned again we'll just suffer the consequences 💀

packages/react-native/src/utils.ts
marandaneto: good catch

PR #3142, opened 2026-02-24
feat(flags): Add feature_flag_cache_ttl_ms config to prevent stale flag values

dmarticus:
## Summary

When the `/flags` request fails (e.g., due to ad blockers or network issues), old cached feature flag values were being used indefinitely, causing users to see disabled variants. This was reported by a customer: 
https://posthoghelp.zendesk.com/agent/tickets/50494 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/53669)

This PR adds a configurable TTL (`feature_flag_cache_ttl_ms`) that:
- Makes `getFeatureFlag()` return `undefined` instead of stale cached values
- Prevents stale `$feature/` properties from being attached to events  
- Stores an evaluation timestamp in persistence for TTL checks

**Usage:**
```javascript
posthog.init('your-api-key', {
  feature_flag_cache_ttl_ms: 86400000, // 1 day
})
```

When enabled:
- If the cache is older than the TTL, `getFeatureFlag()` returns `undefined` with a warning
- Stale `$feature/` properties are not attached to events via `persistence.properties()`
- This prevents outdated flag values from being used when the `/flags` request fails

By default (0 or undefined), the existing behavior is preserved – cached values never expire.

Ships with docs here: https://github.com/PostHog/posthog.com/pull/15295

## Test plan

- [x] Added unit tests for `getFeatureFlag()` returning undefined when cache is stale
- [x] Added unit tests for `persistence.properties()` skipping `$feature/` properties when cache is stale
- [x] Added tests for edge cases: fresh cache, no TTL configured, TTL=0, missing evaluated_at
- [x] Verified all existing tests pass


packages/browser/src/posthog-persistence.ts
```diff
@@ -137,14 +138,32 @@ export class PostHogPersistence {
...
+     * @param ttl Optional TTL override (uses config value if not provided)
+     * @internal
+     */
+    _isFeatureFlagCacheStale(ttl?: number): boolean {
+        const effectiveTtl = ttl ?? this._config.feature_flag_cache_ttl_ms
+        if (!effectiveTtl || effectiveTtl <= 0) {
+            return false
+        }
+        const evaluatedAt = this.props[PERSISTENCE_FEATURE_FLAG_EVALUATED_AT]
+        return !evaluatedAt || Date.now() - evaluatedAt > effectiveTtl
```
gustavohstrassburger: So, if we don't have an evaluatedAt, are we considering the cache stale? is this intended?

gustavohstrassburger: Also, evaluatedAt needs to be stored as timestamp for this to work, not as an ISO string or some other format. Haven't checked this, but just want to make sure we're doing the right thing here.

gustavohstrassburger (approved):
No blockers, just take a look at my comments, I believe your approach is right, just wanted to double check those points with you.

My only "concern" with this PR/approach is that since this won't be the new default, it will mostly be a tool if we get new complaints. 
I get it that changing this to be the default could be somewhat dangerous, but I was just thinking about this while reading the PR (haven't come up with a solution for this though).

dustinbyrne (approved):
Looks good to me!

> **Usage:**
> ```ts
> posthog.init('your-api-key', {
>  feature_flag_cache_ttl_ms: 3600000, // 1 hour
> })
> ```

If this makes its way into documentation, I'd recommend something longer like a day or even a week. A one hour TTL seems likely to occur given an idle tab, or in something with soft routing like React where the client may only be initialized once on initial load and not again on navigation.

packages/browser/src/posthog-persistence.ts
dmarticus: > So, if we don't have an evaluatedAt, are we considering the cache stale? is this intended?

Yes, intentional. If `evaluatedAt` is missing (e.g. SDK upgrade where old cached flags don't have it), we conservatively treat the cache as stale so the next `/flags` response will set a fresh timestamp. Added a comment in the code to make this clearer, plus a `typeof` guard (see below).

dmarticus: > Also, evaluatedAt needs to be stored as timestamp for this to work, not as an ISO string or some other format. Haven't checked this, but just want to make sure we're doing the right thing here.

Confirmed — it's stored as a numeric timestamp (epoch ms). The `FlagsResponse` type has `evaluatedAt?: number` and the server sends it as a number. I've now also added a `typeof evaluatedAt \!== 'number'` runtime guard so if it ever arrives as a string, we treat the cache as stale rather than doing broken arithmetic. Added a test for this case too (738048ba).

PR #3146, opened 2026-02-25
fix(react): PostHogFeature should render fallback when flag doesn't exist

gustavohstrassburger:
## Problem

The PostHogFeature component had incorrect behavior when a feature flag doesn't exist and no match is specified. Instead of rendering the fallback, it would incorrectly render the children content.

This bug occurred because the component only checked if the variant matched without first verifying if the variant exists at all.

## Changes

- Added guard to check if variant exists before evaluating match conditions
- Fixed logic to properly render fallback when feature flag doesn't exist
- Added comprehensive test cases for undefined flag scenarios

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/small-areas-melt.md
+++ b/.changeset/small-areas-melt.md
@@ -1,0 +1,5 @@
+---
+'@posthog/react': patch
+---
+
+Fix PostHogFeature component to properly handle undefined flags and false values


packages/react/src/components/PostHogFeature.tsx
```diff
@@ -31,21 +31,23 @@ export function PostHogFeature({
...
-                trackInteraction={shouldTrackInteraction}
-                trackView={shouldTrackView}
-                onInteract={() => captureFeatureInteraction({ flag, posthog, flagVariant: variant })}
-                onView={() => captureFeatureView({ flag, posthog, flagVariant: variant })}
-                {...props}
-            >
-                {childNode}
-            </VisibilityAndClickTrackers>
-        )
+    if (variant !== undefined) {
```
marandaneto: https://posthog.com/docs/libraries/js/usage#enriched-flag-analytics
anyone can pass any `flag` to PostHogFeature
how do we know if the flag actually exists?
it could that `payload` and `variant` are undefined here

marandaneto: ah your last commit checks for variant

PR #3154, opened 2026-02-26
fix(browser): Improve tablet detection when Chrome sends desktop-like UA

slshults:
## Problem

[A customer reported](https://posthoghelp.zendesk.com/agent/tickets/51304) that tablet devices are being classified as Desktop in `$device_type`. 

Chrome on Android tablets defaults to "request desktop site" mode, sending a UA string like: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/145.0.0.0 Safari/537.36`
(I confirmed that string with my own Pixel tablet) 

That string is indistinguishable from actual desktop Linux, and the existing `detectDeviceType()` function relies solely on UA string parsing, so it falls through to the "Desktop" default for these devices.

## Changes

When UA-based detection falls through to "Desktop", we now check supplementary browser signals in `getEventProperties()`:

- **`navigator.userAgentData.platform`** (Client Hints API) — reports the true platform (`"Android"`) even when Chrome is sending a desktop UA string
- **`navigator.maxTouchPoints`** — confirms the device has touch capability
- **Screen dimensions** — `Math.min(screen.width, screen.height) >= 600` differentiates tablets from phones

This check only runs when all three conditions are met (UA defaulted to Desktop + platform is Android + touch points > 0), so existing correct classifications are never overridden. The supplementary logic lives in `event-utils.ts` rather than the core `user-agent-utils.ts` to keep the core util pure and minimize blast radius.

Also added `platform` to the existing `NavigatorUAData` type interface.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/utils/event-utils.ts
+++ b/packages/browser/src/utils/event-utils.ts
@@ -283,7 +283,8 @@ export function getEventProperties(
         const screenWidth = window?.screen?.width ?? 0
         const screenHeight = window?.screen?.height ?? 0
         const shortSide = Math.min(screenWidth, screenHeight)
-        deviceType = shortSide >= 600 ? 'Tablet' : 'Mobile'
+        const shortSideDp = shortSide / (window?.devicePixelRatio ?? 1)
+        deviceType = shortSideDp >= 600 ? 'Tablet' : 'Mobile'
     }
 
     return extend(
--- a/packages/browser/src/__tests__/utils/event-utils.test.ts
+++ b/packages/browser/src/__tests__/utils/event-utils.test.ts
@@ -6,6 +6,7 @@ import {
     getTimezoneOffset,
 } from '../../utils/event-utils'
 import * as globals from '../../utils/globals'
+import { isUndefined } from '@posthog/core'
 
 describe(`event-utils`, () => {
     describe('properties', () => {
@@ -64,6 +65,7 @@ describe(`event-utils`, () => {
         const originalMaxTouchPoints = Object.getOwnPropertyDescriptor(window.navigator, 'maxTouchPoints')
         const originalScreenWidth = Object.getOwnPropertyDescriptor(window.screen, 'width')
         const originalScreenHeight = Object.getOwnPropertyDescriptor(window.screen, 'height')
+        const originalDevicePixelRatio = Object.getOwnPropertyDescriptor(window, 'devicePixelRatio')
 
         function mockNavigator(userAgentData: any, maxTouchPoints: number) {
             Object.defineProperty(window.navigator, 'userAgentData', {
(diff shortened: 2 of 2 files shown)


packages/browser/src/utils/event-utils.ts
```diff
@@ -273,13 +273,26 @@ export function getEventProperties(
...
+    // Chrome on Android tablets defaults to "request desktop site" mode, sending
+    // a desktop-like UA (e.g. "X11; Linux x86_64"). The UA-based detectDeviceType()
+    // falls through to "Desktop". We use the Client Hints API and touch capability
+    // to catch this case — the browser reports the true platform even when the UA lies.
+    let deviceType = detectDeviceType(userAgent)
+    if (deviceType === 'Desktop' && navigator?.userAgentData?.platform === 'Android' && navigator?.maxTouchPoints > 0) {
+        const screenWidth = window?.screen?.width ?? 0
+        const screenHeight = window?.screen?.height ?? 0
+        const shortSide = Math.min(screenWidth, screenHeight)
+        deviceType = shortSide >= 600 ? 'Tablet' : 'Mobile'
```
marandaneto:
```suggestion
        const shortSideDp = shortSide / (window?.devicePixelRatio ?? 1)
        deviceType = shortSideDp >= 600 ? 'Tablet' : 'Mobile'
```

This is consistent with Android’s dp-based tablet threshold logic, even though you're using pixels, not dp.
It is a reasonable heuristic, but not perfect across device pixel ratios.
If you wanted to be more accurate, you could normalize by devicePixelRatio:

Ps: I didnt test the code above

slshults: Oh, yes. Nice, thanks! Normalized and ready for re-review.

Regarding testing: I ran full the posthog-js unit test suite (94 suites, 3405 tests) before the initial commit, and again after the DPR normalization tweak. No failures. 🙂

packages/browser/src/utils/event-utils.ts
```diff
@@ -273,13 +273,27 @@ export function getEventProperties(
...
+    // falls through to "Desktop". We use the Client Hints API and touch capability
+    // to catch this case — the browser reports the true platform even when the UA lies.
+    let deviceType = detectDeviceType(userAgent)
+    if (deviceType === 'Desktop' && navigator?.userAgentData?.platform === 'Android' && navigator?.maxTouchPoints > 0) {
+        const screenWidth = window?.screen?.width ?? 0
+        const screenHeight = window?.screen?.height ?? 0
+        const shortSide = Math.min(screenWidth, screenHeight)
+        const shortSideDp = shortSide / (window?.devicePixelRatio ?? 1)
+        deviceType = shortSideDp >= 600 ? 'Tablet' : 'Mobile'
+    }
```
lricoy, line 288: Small nit, but I think you could keep this inside `detectDeviceType` as it is already what it is doing, just an extra case. This would mean all the logic would be kept there.

marandaneto: yep +1 here
if detectDeviceType is called in more places, is it? they wouldnt benefit from this fix

marandaneto: posthog-js/packages/react-native/src/native-deps.tsx
packages/browser/src/extensions/utils/matcher-utils.ts
surveys, product tours and react native web wont apply this fix @slshults 
better to move this code within the detectDeviceType method

slshults: Oh, yes, thanks! 
https://github.com/PostHog/posthog-js/pull/3170

PR #2234, opened 2025-08-21
fix: expo-43 and new expo-file-system APIs with back compatibility support

marandaneto:
## Problem

Expo 54 and expo-file-system 19 changed its APIs but added a legacy package
Closes https://github.com/PostHog/posthog-js/issues/2229

## Changes

Add back compatibility and new API support for expo-file-system

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Checklist

- [X] Tests for new code
- [X] Accounted for the impact of any changes across different platforms
- [X] Accounted for backwards compatibility of any changes (no breaking changes!)
- [ ] Took care not to unnecessarily increase the bundle size

### If releasing new changes

- [X] Ran `pnpm changeset` to generate a changeset file
- [X] Added the "release" label to the PR to indicate we're publishing new versions for the affected packages

PR #3156, opened 2026-02-26
fix(react-native): expo-file-system detection broken on Expo SDK 54 stable

### LINKED ISSUE #3151: posthog-react-native: expo-file-system detection broken on Expo SDK 54 stable
## Bug description

`posthog-react-native@4.36.0` throws `Error: Method writeAsStringAsync imported from "expo-file-system" is deprecated` on every storage write when used with Expo SDK 54 (stable).

This is distinct from #2229 (SDK 54 beta), which was fixed in #2234. The fix checks whether `readAsStringAsync` exists on the `expo-file-system` module to decide between the legacy and new File API paths:

```javascript
// native-deps.js (simplified)
if (expoFileSystemLegacy.readAsStringAsync) {
  return buildLegacyStorage(_filesystem); // uses writeAsStringAsync
}
// else: use new File(uri) / file.write(value)
```

In SDK 54 **beta**, the legacy methods were completely removed (`undefined`), so this check correctly fell through to the new API.

In SDK 54 **stable**, the legacy methods are still exported as functions — they just throw a deprecation error when called ([source](https://github.com/expo/expo/blob/931706e623011b51a21e3493608e27519dc0d64c/packages/expo-file-system/src/legacyWarnings.ts#L51-L57)). So the existence check passes, PostHog enters the legacy path, and `writeAsStringAsync` throws.

The `require('expo-file-system/legacy')` fallback (Step 1 in the detection logic) also doesn't resolve at runtime — the package ships `legacy.ts` but no `legacy.js` at the root, and PostHog's compiled JS can't resolve the TypeScript file.

## How to reproduce

1. Create an Expo SDK 54 app (stable, not beta)
2. Install `posthog-react-native@4.36.0`
3. Set up PostHogProvider with default options (no `customStorage`)
4. Run the app — errors appear on every `capture()`, `identify()`, or any operation that persists to storage

## Workaround

Pass a custom storage provider to bypass the detection logic entirely:

```tsx
import { MMKV } from 'react-native-mmkv';

const storage = new MMKV({ id: 'posthog' });

<PostHogProvider
  apiKey="..."
  options={{
    customStorage: {
      getItem: (key) => storage.getString(key) ?? null,
      setItem: (key, value) => storage.set(key, value),
    },
  }}
>
```

## Suggested fix

The detection should test whether the function actually works, not just whether it exists. For example:

```javascript
try {
  // Try calling readAsStringAsync with a non-existent path
  // If it throws with the deprecation message, fall through to new API
  await _filesystem.readAsStringAsync('__posthog_probe__');
} catch (e) {
  if (e.message?.includes('deprecated')) {
    // Functions exist but throw — use new File API
    return buildNewFileStorage(_filesystem);
  }
  // Other error (e.g., file not found) — legacy API works, use it
  return buildLegacyStorage(_filesystem);
}
```

Or more simply, check `_filesystem.Paths` to detect the new API:

```javascript
if (_filesystem.Paths && _filesystem.File) {
  return buildNewFileStorage(_filesystem);
}
if (_filesystem.readAsStringAsync) {
  return buildLegacyStorage(_filesystem);
}
```

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Additional context

- #2229 — original SDK 54 beta report (fixed in #2234)
- #2453 — same bug report, closed as "already fixed" but the fix doesn't handle stable SDK 54 behavior
- `expo-file-system` v19.0.21 (Expo SDK 54 stable)
- Confirmed on `posthog-react-native` 4.36.0 (latest as of Feb 2026)

marandaneto:
## Problem

`posthog-react-native` throws `Error: Method writeAsStringAsync imported from "expo-file-system" is deprecated` on every storage write when used with Expo SDK 54 stable.

The previous fix (#2234) checked whether `readAsStringAsync` exists to decide between legacy and new File API paths. This worked for SDK 54 **beta** (where legacy methods were completely removed), but breaks on SDK 54 **stable** where legacy methods are still exported as functions — they just throw a deprecation error when called.

Additionally, the new File API usage was incorrect — `Paths.document.info().uri` is not valid per the [Expo docs](https://docs.expo.dev/versions/latest/sdk/filesystem/). `Paths.document` is a `Directory` instance and the `File` constructor accepts `(Directory, ...strings)` for path joining.

Closes #3151

## Changes

**Reordered detection logic** in `buildOptimisiticAsyncStorage()`:

1. **New API first**: Check for `Paths`/`File` on the main `expo-file-system` module — if present (SDK 54+), use the new File API
2. **Legacy fallback**: Only then check for `readAsStringAsync` via the legacy subpath (`expo-file-system/legacy`, SDK 55+) or the main module (SDK ≤53)

This is safe because the deprecated throwing stubs and the new `Paths`/`File` API were introduced in the same version (expo-file-system v19). By checking for `Paths`/`File` first and returning early, we ensure the legacy fallback only runs on SDK ≤53 where `readAsStringAsync` is a real working function.

**Fixed File API usage**: Changed from `new File(Paths.document.info().uri + key)` to `new File(Paths.document, key)` per the [Expo filesystem docs](https://docs.expo.dev/versions/latest/sdk/filesystem/).

**Detection matrix across all Expo versions:**

| SDK | `Paths`/`File` | `readAsStringAsync` | Result |
|---|---|---|---|
| ≤ 53 | ❌ absent | ✅ real function | → Legacy `buildLegacyStorage()` |
| 54 beta | ✅ present | ❌ absent | → New File API |
| 54 stable | ✅ present | ⚠️ exists but throws | → New File API |
| 55+ | ✅ present | ⚠️ exists but throws (working version in `/legacy` subpath) | → New File API |

### Libraries affected

- [x] posthog-react-native

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/react-native/src/native-deps.tsx
+++ b/packages/react-native/src/native-deps.tsx
@@ -149,7 +149,7 @@ const buildLegacyStorage = (filesystem: any): PostHogCustomStorage => {
   }
 }
 
-export const buildOptimisiticAsyncStorage = (): PostHogCustomStorage | undefined => {
+export const buildOptimisticAsyncStorage = (): PostHogCustomStorage | undefined => {
   // On web platform during SSR (no window), skip file storage
   // The caller will fall back to memory storage
   if (isWeb() && typeof (GLOBAL_OBJ as any)?.window === 'undefined') {
--- a/packages/react-native/test/storage-expo55.spec.ts
+++ b/packages/react-native/test/storage-expo55.spec.ts
@@ -47,7 +47,7 @@ jest.mock('react-native', () => ({
   Platform: { OS: 'ios' },
 }))
 
-import { buildOptimisiticAsyncStorage } from '../src/native-deps'
+import { buildOptimisticAsyncStorage } from '../src/native-deps'
 import { OptionalExpoFileSystem } from '../src/optional/OptionalExpoFileSystem'
 
 describe('Expo SDK 55 - prefers new File API over working legacy subpath', () => {
@@ -59,7 +59,7 @@ describe('Expo SDK 55 - prefers new File API over working legacy subpath', () =>
   })
 
   it('should use new File/Paths API even when legacy subpath is available', () => {
-    const storage = buildOptimisiticAsyncStorage()
+    const storage = buildOptimisticAsyncStorage()
     expect(storage).toBeDefined()
 
     // Verify it uses new API
@@ -73,7 +73,7 @@ describe('Expo SDK 55 - prefers new File API over working legacy subpath', () =>
   })
 
   it('should read using new File API, not legacy', async () => {
-    const storage = buildOptimisiticAsyncStorage()!
+    const storage = buildOptimisticAsyncStorage()!
 
     const result = await storage.getItem('test-key')
     expect(result).toBe('stored-value')
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -18,7 +18,7 @@ import {
 } from '@posthog/core'
 import { PostHogRNStorage, PostHogRNSyncMemoryStorage } from './storage'
 import { version } from './version'
-import { buildOptimisiticAsyncStorage, getAppProperties } from './native-deps'
+import { buildOptimisticAsyncStorage, getAppProperties } from './native-deps'
 import {
   PostHogAutocaptureOptions,
   PostHogCustomAppProperties,
@@ -183,7 +183,7 @@ export class PostHog extends PostHogCore {
 
     let theStorage: PostHogCustomStorage | undefined
     if (this._persistence === 'file') {
-      theStorage = options?.customStorage ?? buildOptimisiticAsyncStorage()
+      theStorage = options?.customStorage ?? buildOptimisticAsyncStorage()
     }
 
     if (theStorage) {
--- a/packages/react-native/test/storage-expo54-stable.spec.ts
+++ b/packages/react-native/test/storage-expo54-stable.spec.ts
@@ -43,7 +43,7 @@ jest.mock('react-native', () => ({
   Platform: { OS: 'ios' },
 }))
 
-import { buildOptimisiticAsyncStorage } from '../src/native-deps'
+import { buildOptimisticAsyncStorage } from '../src/native-deps'
 import { OptionalExpoFileSystem } from '../src/optional/OptionalExpoFileSystem'
 
 describe('Expo SDK 54 stable - new File API detection', () => {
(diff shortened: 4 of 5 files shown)


packages/react-native/test/storage-expo55.spec.ts
```diff
@@ -0,0 +1,83 @@
...
+
+jest.mock('../src/optional/OptionalAsyncStorage', () => ({
+  OptionalAsyncStorage: undefined,
+}))
+
+jest.mock('react-native', () => ({
+  Platform: { OS: 'ios' },
+}))
+
+import { buildOptimisiticAsyncStorage } from '../src/native-deps'
```
ioannisj: typo here?

packages/react-native/src/native-deps.tsx
```diff
@@ -160,44 +160,50 @@ export const buildOptimisiticAsyncStorage = (): PostHogCustomStorage | undefined
...
+    const legacyModule = (OptionalExpoFileSystemLegacy || OptionalExpoFileSystem) as any
     try {
-      const expoFileSystemLegacy = filesystem as any
-      // identify legacy APIs with older versions (expo <= 53 and expo-file-system <= 18)
-      if (expoFileSystemLegacy.readAsStringAsync) {
-        return buildLegacyStorage(filesystem)
+      if (legacyModule && typeof legacyModule.readAsStringAsync === 'function') {
+        return buildLegacyStorage(legacyModule)
       }
     } catch (e) {}
```
ioannisj, line 206: We are completely silent here, should we log something?

ioannisj (approved):
Did not go into great details but changes seem reasonable.

packages/react-native/src/native-deps.tsx
marandaneto: mm why? its a valid path/supported path using older versions

packages/react-native/test/storage-expo55.spec.ts
marandaneto: why? its a valid method

marandaneto: or do you mean 'optimistic'?

ioannisj: Yeah had to read it twice. It's "optimisitic" right?😛

marandaneto: yep, but that was just an import, the method was already with a typo from day 1

marandaneto: good catch!

PR #3163, opened 2026-02-26
feat(node): Add timestamp tracking for debugging local feature flag evaluation

gustavohstrassburger:
## Problem

Local feature flag evaluation in posthog-node was missing the timestamp tracking that's available in remote evaluation, making it difficult to debug timing-related issues when troubleshooting flag behavior.

When debugging cache TTL problems or investigating timing between flag definition loading and evaluation, developers had no visibility into:
- When flag definitions were loaded from the server
- When individual flags were evaluated for comparison with remote evaluation behavior

## Changes

Added simple timestamp tracking to improve debugging and observability for local feature flag evaluation:

- Track when flag definitions are loaded from server (`flagDefinitionsLoadedAt`)
- Include timestamps in `$feature_flag_called` events for locally evaluated flags
- Provide equivalent timing information to what remote evaluation already includes

The implementation uses a straightforward approach with inline `Date.now()` calls based on code review feedback, avoiding complex data structures while providing the necessary debugging information.

This ensures locally evaluated flags include the same timestamp data as remotely evaluated flags in analytics events, making it easier to debug timing-related issues and maintain consistency between evaluation methods.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -821,18 +821,16 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
           locally_evaluated: flagWasLocallyEvaluated,
           [`$feature/${key}`]: response,
           $feature_flag_request_id: requestId,
-          $feature_flag_evaluated_at: evaluatedAt,
+          $feature_flag_evaluated_at: flagWasLocallyEvaluated ? Date.now() : evaluatedAt,
         }
 
-        // Add local evaluation timestamps when flag was evaluated locally
+        // Add local evaluation definition load timestamp
         if (flagWasLocallyEvaluated && this.featureFlagsPoller) {
           const flagDefinitionsLoadedAt = this.featureFlagsPoller.getFlagDefinitionsLoadedAt()
 
           if (flagDefinitionsLoadedAt !== undefined) {
             properties.$feature_flag_definitions_loaded_at = flagDefinitionsLoadedAt
           }
-          // For local evaluation, evaluation happens immediately when you call the method
-          properties.$feature_flag_evaluated_at = Date.now()
         }
 
         if (featureFlagError) {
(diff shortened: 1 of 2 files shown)


dmarticus (commented):
Okay, this is a valid problem and I'm glad you're tackling it. I am not sure it's exactly the right way to go about it though; here's what I'm thinking:

1. The `evaluationTimestampCache` design is awkward. The context is created fresh per-request, so the "cache" is set during evaluation and read moments later in the same call. It's not really caching anything
across requests - it's just a roundabout way to pass `Date.now()` from the evaluation point to the event emission.
2. Over-engineering? For local evaluation, the flag is computed synchronously. The evaluation timestamp is essentially "now." Why not just capture `Date.now()` directly when building the event properties, rather than storing it in a context object first?
3. The `flagDefinitionsLoadedAt` part makes sense - knowing when your cached definitions were fetched is useful for cache TTL management and debugging staleness issues.
4. The per-flag timestamp seems less useful - For remote evaluation, `evaluatedAt` comes from the server and tells you when the server made the decision (which could differ from request time due to network latency). For local evaluation, evaluation happens immediately when you call the method - the timestamp is always "right now."

A simpler approach might be:
- Keep `flagDefinitionsLoadedAt`
- Drop the `evaluationTimestampCache` complexity and just use Date.now() inline when setting `$feature_flag_evaluated_at` for locally evaluated flags (unless there's a specific use case where you evaluate a flag, then later send the event and need to preserve the original evaluation time – but that doesn't seem like the common pattern here)

PR #3165, opened 2026-02-26
feat(browser): Tree shake additional extensions

dustinbyrne:
## Problem

Six classes are always constructed in the PostHog constructor regardless of whether they're needed: Surveys, Toolbar, PostHogExceptions, Conversations, Logs, and WebExperiments. This means they can't be tree-shaken out of slim/custom bundles.

This is PR 2/3
1. https://github.com/PostHog/posthog-js/pull/3164
2. https://github.com/PostHog/posthog-js/pull/3165 **<<< you are here**
3. https://github.com/PostHog/posthog-js/pull/3166

## Changes

Makes all six classes conditionally constructed via `__extensionClasses`, matching the pattern already used for autocapture, session recording, heatmaps, etc.

- **Imports changed to type-only**: Hard imports of 6 classes replaced with `import type` in posthog-core.ts. Classes are only referenced through the `__extensionClasses` config.
- **Properties made optional**: `surveys?`, `toolbar?`, `exceptions?`, `conversations?`, `logs?`, `experiments?` — all use optional chaining at call sites (~15 locations).
- **Graceful degradation**: Survey methods (`getSurveys`, `getActiveMatchingSurveys`, `canRenderSurvey`, etc.) call the error callback when the surveys module isn't loaded. `captureException` returns early. `loadToolbar` returns false.
- **`_onRemoteConfig` fully unified**: Removes the remaining hardcoded dispatch calls — everything now goes through `this.extensions.forEach(ext => ext.onRemoteConfig?.(config))`.
- **6 classes** now `implement Extension` with appropriate `initialize()` methods.
- **Survey enums → const objects**: All TypeScript `enum` declarations in `posthog-surveys-types.ts` converted to `as const` objects with type aliases. Enums compile to IIFEs that bundlers can't tree-shake; const objects are plain property access.
- **Extension bundles updated**: New exports `SurveysExtensions`, `ToolbarExtensions`, `ExperimentsExtensions`, `ConversationsExtensions`, `LogsExtensions`, and `ErrorTrackingExtensions` now includes `exceptions`.

## Bundle size

| | module.js | module.slim.js | module.js (gz) | module.slim.js (gz) |
|---|---|---|---|---|
| **main** | 177.1 kB | 142.2 kB | 56.8 kB | 46.3 kB |
| **this PR** | 177.5 kB (+0.2%) | 107.0 kB (-24.8%) | 56.9 kB (+0.3%) | 36.4 kB (-21.4%) |

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/types.ts
+++ b/packages/browser/src/types.ts
@@ -18,6 +18,12 @@ import type { SessionRecording } from './extensions/replay/session-recording'
 import type { Heatmaps } from './heatmaps'
 import type { PostHogProductTours } from './posthog-product-tours'
 import type { SiteApps } from './site-apps'
+import type { PostHogSurveys } from './posthog-surveys'
+import type { Toolbar } from './extensions/toolbar'
+import type { PostHogExceptions } from './posthog-exceptions'
+import type { WebExperiments } from './web-experiments'
+import type { PostHogConversations } from './extensions/conversations/posthog-conversations'
+import type { PostHogLogs } from './posthog-logs'
 
 // ============================================================================
 // Re-export public types from @posthog/types
@@ -141,6 +147,12 @@ export type PostHogConfig = Omit<BasePostHogConfig, 'loaded'> & {
         webVitalsAutocapture?: ExtensionConstructor<WebVitalsAutocapture>
         exceptionObserver?: ExtensionConstructor<ExceptionObserver>
         deadClicksAutocapture?: ExtensionConstructor<DeadClicksAutocapture>
+        exceptions?: ExtensionConstructor<PostHogExceptions>
+        surveys?: ExtensionConstructor<PostHogSurveys>
+        toolbar?: ExtensionConstructor<Toolbar>
+        experiments?: ExtensionConstructor<WebExperiments>
+        conversations?: ExtensionConstructor<PostHogConversations>
+        logs?: ExtensionConstructor<PostHogLogs>
     }
 }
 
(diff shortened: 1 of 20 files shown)


marandaneto (approved):

packages/browser/src/types.ts
```diff
@@ -25,6 +25,22 @@ import type { WebExperiments } from './web-experiments'
...
+ * Augmented by the slim bundle entry point to mark tree-shakeable extensions
+ * as optional. Empty by default (full bundle), meaning extensions are guaranteed.
+ */
+export interface TreeShakeableConfig {}
+
+/**
+ * For the full bundle (default), resolves to T (extensions guaranteed present).
+ * For the slim bundle (augmented with { optional: true }), resolves to T | undefined.
+ */
+export type TreeShakeable<T> = 'optional' extends keyof TreeShakeableConfig ? T | undefined : T
```
dustinbyrne, line 42: I noticed our Vercel build was failing because it expected `posthog.surveys` to always be defined. In the full bundle it always is defined, but the type checker still requires a null check. Having thought about it some more, it doesn't seem reasonable for us to be making a breaking change in the type checking.

I came up with this solution. This allows us to mark properties as always defined in a full bundle, and optionally defined in a slim bundle. It defaults to being defined, so there's no risk of breaking backwards compatibility with types.

lucasheriques (approved):
looks good on the surveys side :pray:

PR #3172, opened 2026-02-27
fix(browser): Send gzipped request body as ArrayBuffer

dustinbyrne:
## Problem

Playwright integration tests are failing quite often on Safari.

After some debugging, I discovered that some event ingestion requests were returning 400 responses, stating the body was empty.

As it turns out, WebKit sends an empty request body via `fetch` when the body contains a `Blob`. I was able to confirm this behavior in a separate minimal reproduction. This likely [affects other Safari users](https://us.posthog.com/project/2/logs?serviceNames=%5B%22contour%22%5D&filterGroup=%7B%22type%22%3A%22AND%22%2C%22values%22%3A%5B%7B%22type%22%3A%22AND%22%2C%22values%22%3A%5B%7B%22key%22%3A%22response_code%22%2C%22value%22%3A%5B%22400%22%5D%2C%22operator%22%3A%22exact%22%2C%22type%22%3A%22log_attribute%22%7D%5D%7D%5D%7D) as well.

Integration tests on Safari were only passing if the events could be sent before remote config was received, enabling compression (and thus sending blobs on fetch). The Playwright test mocked both`/array/<token>/config` and `/flags` as a source of remote config but not `/array/<token>/config.js`, which we've recently switched to. When we switched over, the integration tests began hitting the real service, getting a full remote config, and enabling compression.

## Changes

We now send an ArrayBuffer instead of a Blob.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -1,6 +1,7 @@
 /* eslint-disable compat/compat */
 /// <reference lib="dom" />
 
+import { TextDecoder } from 'util'
 import { extendURLParams, request } from '../request'
 import { Compression, RequestWithOptions } from '../types'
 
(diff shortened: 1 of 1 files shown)


pauldambra (commented):
vercel failure seems unrelated  


```
> playground-nextjs@0.1.0 build /vercel/path0/playground/nextjs
> next build --no-lint
 ⚠ Linting is disabled.
   ▲ Next.js 15.5.7
   Checking validity of types ...
   Creating an optimized production build ...
Failed to compile.
./pages/index.tsx
Module not found: Can't resolve 'posthog-js/lib/src/customizations'
https://nextjs.org/docs/messages/module-not-found
> Build failed because of webpack errors
 ELIFECYCLE  Command failed with exit code 1.
Error: Command "pnpm build" exited with 1
```

PR #3174, opened 2026-02-28
fix(replay): detect and report rrweb silent initialization failure

TueHaulund:
## Summary

rrweb's `record()` wraps its entire body in a try-catch that silently swallows errors (via `console.warn`) and returns `undefined`. The SDK was not checking this return value, so it would continue to report `recording_status: active` while rrweb never actually initialized — resulting in zero recording data being captured or flushed.

Likely triggers include canvas recording with cross-origin tainted canvases (`SecurityError` on `toDataURL()`), strict CSP policies, or complex DOM structures that cause rrweb to throw during initialization.

## Changes

- **`types/rrweb.ts`**: Fixed `rrwebRecord` return type to `(() => void) | undefined` to match the actual upstream rrweb behavior — previously declared as `() => void` which meant TypeScript wouldn't warn about the missing check
- **`triggerMatching.ts`**: Added `RRWEB_ERROR` status constant and `rrwebError` property to `RecordingTriggersStatus` — all three matcher functions (`nullMatch`, `anyMatch`, `allMatch`) check it first
- **`lazy-loaded-session-recorder.ts`**: Added `_rrwebError` flag, set when `rrwebRecord()` returns falsy in `_startRecorder()`, cleared on successful start (supports retries), passed through the status matcher
- **`sessionRecordingStatus.test.ts`**: Tests that `RRWEB_ERROR` is returned by both matchers, and that it takes precedence over sampling and active triggers
- **`lazy-sessionrecording.test.ts`**: Tests that `started === false` and `status === 'rrweb_error'` when `rrwebRecord()` returns `undefined`, and that recording recovers when rrweb starts successfully on retry

## Test plan

- [x] All 12 replay test suites pass (410 tests)
- [ ] Manual test with canvas recording on a page with cross-origin tainted canvas
- [ ] Verify `$recording_status` property shows `rrweb_error` in debug events when rrweb fails to initialize

### DIFF
--- a/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
+++ b/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
@@ -814,8 +814,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         this._makeSamplingDecision(this.sessionId)
         this._startRecorder()
 
-        if (!this.isStarted) {
-            this._rrwebError = true
+        if (this._rrwebError) {
             return
         }
 
@@ -1557,6 +1556,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             $sdk_debug_replay_flushed_size: this._flushedSizeTracker?.currentTrackedSize,
             $sdk_debug_replay_full_snapshots: this._fullSnapshotTimestamps,
             $snapshot_max_depth_exceeded: this._maxDepthExceeded,
+            $sdk_debug_replay_rrweb_error: this._rrwebError,
         }
     }
 


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -813,6 +814,11 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         this._makeSamplingDecision(this.sessionId)
         this._startRecorder()
 
+        if (!this.isStarted) {
+            this._rrwebError = true
+            return
+        }
```
pauldambra: am i being confused... `this._rrwebError` is already being set in `this._startRecorder` could we just check it here?
```suggestion
        if (this._rrwebError) {
            return
        }
```

??

packages/browser/src/extensions/replay/types/rrweb.ts
```diff
@@ -104,7 +104,7 @@ export type recordOptions = {
 
 // Replication of `record` from inside `@rrweb/record`
 export type rrwebRecord = {
-    (options: recordOptions): () => void
+    (options: recordOptions): (() => void) | undefined
```
pauldambra, line 107: i wish i'd finished moving rrweb into this monorepo... we should really make this explicit 
`undefined` to mean error is so vague 🙈

PR #3206, opened 2026-03-05
feat(ai): support versioned prompts

Radu-Raicea:
## Summary
- add a `version` option to `Prompts.get()` so callers can request a specific published prompt version
- keep latest and exact-version prompt cache entries isolated, and clear all cached variants for a prompt name
- add prompt SDK tests for versioned fetches, cache isolation, cache clearing, and version-specific 404 errors

### DIFF
--- a/packages/ai/src/prompts.ts
+++ b/packages/ai/src/prompts.ts
@@ -173,9 +173,15 @@ export class Prompts {
    */
   clearCache(name?: string): void {
     if (name !== undefined) {
-      const prefix = `${name}::`
+      const latestKey = this.getCacheKey(name)
+      const versionPrefix = `${name}::version:`
       for (const key of this.cache.keys()) {
-        if (key.startsWith(prefix)) {
+        if (key === latestKey) {
+          this.cache.delete(key)
+          continue
+        }
+
+        if (key.startsWith(versionPrefix) && /^\d+$/.test(key.slice(versionPrefix.length))) {
           this.cache.delete(key)
         }
       }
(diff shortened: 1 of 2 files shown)


packages/ai/src/prompts.ts
```diff
@@ -75,20 +80,31 @@ export class Prompts {
     }
   }
 
+  private getCacheKey(name: string, version?: number): string {
+    return version === undefined ? `${name}::latest` : `${name}::version:${version}`
+  }
```
carlos-marchal-ph, line 85: I would document that we enforce cache naming in the backend that prevents using `::`:

```suggestion

  // :: is restricted from use in cache names in the backend, so no chance of accidental collision
  private getCacheKey(name: string, version?: number): string {
    return version === undefined ? `${name}::latest` : `${name}::version:${version}`
  }
```

Still a bit brittle since we could change that in the backend without realising this code exists, the ideal solution would be to use a `Map<string, Map<number | undefined, CachedPrompt>>` as the cache, mimicking the Python approach of a tuple as a key.

packages/ai/src/prompts.ts
```diff
@@ -153,17 +169,28 @@ export class Prompts {
...
   clearCache(name?: string): void {
     if (name !== undefined) {
-      this.cache.delete(name)
+      const latestKey = this.getCacheKey(name)
+      const versionPrefix = `${name}::version:`
+      for (const key of this.cache.keys()) {
+        if (key === latestKey) {
+          this.cache.delete(key)
+          continue
+        }
```
carlos-marchal-ph, line 182: The approach described in my previous comment would also make clearing all versions for a prompt a one liner without string manipulation.

packages/ai/src/prompts.ts
```diff
@@ -153,17 +169,28 @@ export class Prompts {
   /**
    * Clear the cache for a specific prompt or all prompts
    *
-   * @param name - Optional prompt name to clear. If not provided, clears all cached prompts.
+   * @param name - Optional prompt name to clear. If provided, clears all cached versions for that prompt.
    */
   clearCache(name?: string): void {
```
carlos-marchal-ph, line 174: Shouldn't this be `clearCache(name?: string, version?: number): void` for parity with Python's API?

PR #3216, opened 2026-03-10
chore(node): Add edge runtime test coverage

dustinbyrne:
## Problem

The Node SDK exports an edge client, but we do not regularly validate tests in an edge runtime.

## Changes

- Add `test:edge-compat` script to run Node tests under `@edge-runtime/jest-environment`
- Add edge-compat tests to CI                                        
- Replace `AbortSignal.timeout()` in `fetchWithRetry` with manual `AbortController` + `clearTimeout` because the old approach leaked timers in edge runtimes where `setTimeout` returns numeric handles that can't be `unref()`'d
- Same fix for `_shutdown`'s `Promise.race` timeout
- Add test verifying the AbortController abort fires on slow requests

There is a patch release included in this PR but it's mostly cleanup to ensure we don't leak timers.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/core
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/good-ties-double.md
+++ b/.changeset/good-ties-double.md
@@ -1,0 +1,5 @@
+---
+'@posthog/core': patch
+---
+
+fix: Clean up leaked timers in edge runtimes


.changeset/good-ties-double.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/core': patch
```
marandaneto, line 2: since node depends on @posthog/core, dont we need a node patch as well?

dustinbyrne: I believe it will automatically patch bump dependencies - will double check

PR #3222, opened 2026-03-10
fix(node): Stop sending config=true on server-side /flags requests

haacked:
## Problem

Server-side SDKs (posthog-node) send `config=true` on every `/flags` request, causing the server to include session recording settings, autocapture config, surveys, and other client-side-only configuration in the response. The node SDK never reads any of these config fields — it only uses flag-related fields (`flags`, `featureFlags`, `featureFlagPayloads`, `errorsWhileComputingFlags`, `quotaLimited`, `requestId`). This is wasted bandwidth on both client and server.

From production access logs, ~34% of `/flags` requests currently include `config=true`, with a significant portion coming from posthog-node.

## Changes

Change the default value of `fetchConfig` in `PostHogCoreStateless.getFlags()` from `true` to `false`.

This is safe because:
- **Client-side SDKs are unaffected** — the browser SDK (`PostHogCore._flagsAsync`) and React Native SDK both pass `fetchConfig` explicitly to `super.getFlags()`, never relying on the default.
- **Server-side SDKs get the correct behavior** — `getFeatureFlagDetailsStateless()` (the stateless path used by posthog-node) calls `this.getFlags()` without passing `fetchConfig`, so it was the only consumer relying on the default.

### Libraries affected

- [x] posthog-node

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/rich-hats-rescue.md
+++ b/.changeset/rich-hats-rescue.md
@@ -1,5 +1,6 @@
 ---
 'posthog-node': patch
+'@posthog/core': patch
 ---
 
-Set config=false to request only the necessary data
+Omit the config query parameter by default to request only the necessary data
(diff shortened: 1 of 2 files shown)


dmarticus (approved):
ooops, must've missed this

PR #3223, opened 2026-03-11
chore(ci): sign release commits

Piccirello:
This is a prerequisite to enforcing commit signing in this repo.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -206,7 +206,7 @@ jobs:
             - name: Checkout repository
               uses: actions/checkout@v4
               with:
-                  ref: main
+                  ref: ${{ needs.version-bump.outputs.commit-hash }}
                   fetch-depth: 0
 
             - name: Setup environment


.github/workflows/release.yml
```diff
@@ -67,9 +67,9 @@ jobs:
         if: always() && needs.check-changesets.outputs.has-changesets == 'true'
         environment: 'NPM Release' # This will require an approval from a maintainer, they are notified in Slack above
         permissions:
-            contents: write
+            contents: read
```
Piccirello, line 70: This job performs all writes using the app token, so the `GITHUB_TOKEN` doesn't need `write` perms.

.github/workflows/release.yml
```diff
@@ -262,15 +249,6 @@ jobs:
               env:
                   NPM_CONFIG_PROVENANCE: true
```
rafaeelaudibert: <p class='graphite__hidden'><i>[Re: lines +234 to +250]</i></p>
What happens if publishing fails? Previously we'd not push the tag to github, but we do now. Will tagging again - if we retry deploying - work or will it fail?
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3223?utm_source=unchanged-line-comment">Graphite</a>.</p>

.github/workflows/release.yml
```diff
@@ -250,10 +234,13 @@ jobs:
...
               if: steps.check-package-version.outputs.is-new-version == 'true'
               env:
+                  GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   PACKAGE_NAME: ${{ matrix.package.name }}
                   COMMITTED_VERSION: ${{ steps.check-package-version.outputs.committed-version }}
               run: |
-                  git tag -a "$PACKAGE_NAME@$COMMITTED_VERSION" -m "$PACKAGE_NAME@$COMMITTED_VERSION"
+                  gh api "repos/${{ github.repository }}/git/refs" \
+                    -f "ref=refs/tags/$PACKAGE_NAME@$COMMITTED_VERSION" \
+                    -f "sha=$(git rev-parse HEAD)"
```
Piccirello, line 250: I initially intended to create the release using the GitHub App token, that way it would be attributed to the app. But the necessary secrets are only available in the `release` environment, which this doesn't run it. Adding this step to the environment would require a second approval.

Ultimately I think this is ok. The goal is to ensure all commits/releases are signed, and not necessarily to ensure they're attributed to the GitHub app.

.github/workflows/release.yml
Piccirello: That's a great spot. Pushing a tag that already exists results in a 422 and a non-zero exit code. I'll swap the order of these so that the tag push occurs after the `pnpm publish`. This way the new process is functionally equivalent to the previous process.

.github/workflows/release.yml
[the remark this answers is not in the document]
Piccirello: Good find, I think this is a legit race condition.

@rafaeelaudibert I've fixed this in 633028634b45e322af929eaa6185eab9d2325d4e but would love to get your take before merging. If valid, we likely have this same race in our other release workflows.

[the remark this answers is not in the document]
rafaeelaudibert: Hmm, it's possible, yeah. I wonder if there's something better akin to `event.commit-sha` we could use? This way we know we're always pulling from the point in time this job/event was triggered

PR #3228, opened 2026-03-11
fix(replay): restart recorder when session rotates externally while idle

TueHaulund:
## Summary

- When the recorder is idle and an analytics event (e.g. `$pageleave`, `$exception`) triggers session rotation via `checkAndGetSessionAndWindowId` in posthog-core, the recorder failed to restart under the new session
- `_onSessionIdCallback` relied on `_updateWindowAndSessionIds` to detect the change, but that method returns early when `_isIdle` is true
- Result: analytics events go to the new session, recording data stays under the old session → "Recording not found"
- Fix: detect idle state in `_onSessionIdCallback` and directly call `stop()/start()` to restart under the new session

## Context

Investigated via support ticket #48896 95/100 "missing" recording sessions had recordings under a different session ID. 91/95 were activity timeout rotations (>30min idle gap).

## Test plan

- [x] Added regression test: `restarts recorder when session rotates externally while idle`
- [x] All 188 lazy-sessionrecording tests pass
- [x] All 420 replay tests pass
- [x] All 61 sessionid tests pass
- [ ] Runtime verification on EXAMIND platform (breakpoints at idle guard + callback)

### DIFF
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -3466,6 +3466,8 @@ describe('Lazy SessionRecording', () => {
                 },
                 lastActivityTimestamp: expect.any(Number),
             })
+
+            expect(tryAddCustomEvent).toHaveBeenCalledTimes(3)
         })
 
         it('emits session linking events on session past maximum length', () => {
@@ -3509,6 +3511,8 @@ describe('Lazy SessionRecording', () => {
                 },
                 lastActivityTimestamp: expect.any(Number),
             })
+
+            expect(tryAddCustomEvent).toHaveBeenCalledTimes(3)
         })
 
         it('includes flushed_size with actual data size in session ending event', () => {
@@ -3586,6 +3590,8 @@ describe('Lazy SessionRecording', () => {
                     sessionPastMaximumLength: false,
                 },
             })
+
+            expect(tryAddCustomEvent).toHaveBeenCalledTimes(1)
         })
 
         it('always emits $session_id_change event regardless of change reason', () => {


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -920,10 +920,20 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
 
         this._clearConditionalRecordingPersistence()
 
-        // Note: We don't call stop()/start() here because _updateWindowAndSessionIds
-        // already handles the restart. This callback fires synchronously during
-        // checkAndGetSessionAndWindowId(), so _updateWindowAndSessionIds will detect
-        // the session change and handle the restart after this callback returns.
+        if (this._isIdle === true) {
+            // When idle, _updateWindowAndSessionIds returns early at the _isIdle guard
```
pauldambra: but then `_updateWindowAndSessionIds` is called on rrweb events

so the sequence i'd expect is

* recorder goes idle
* posthog event fires
* and so session id changes
* recorder comes back from idle
* fires an rrweb event
* hits _updateWindowAndSessionIds
* we pick up new session id and restart

i wonder why not

which i guess is a long way of saying i wonder if we can handle this in one place to avoid having to think about several

pauldambra: session id changed while idle
is a good reason to at least stop the recorder though

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -3424,8 +3466,6 @@ describe('Lazy SessionRecording', () => {
                 },
                 lastActivityTimestamp: expect.any(Number),
             })
-
-            expect(tryAddCustomEvent).toHaveBeenCalledTimes(3)
```
pauldambra, line 3428: how come we lose these assertions?
(it's not like me  not to have made these assertions concrete tbh i'd have expected to assert "we call these N custom events in any order")

packages/browser/playwright/mocked/session-recording/session-rotation-scenarios.spec.ts
pauldambra, line 1: are we _certain_ the recorder is going idle and so we're exercising the new restart?
(i'd have to change the test to be sure, so i'm lazily asking how we know we're hitting our new restart? if we comment out the change does the test fail to confirm the bug?)

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
fasyy612: The previous broader consolidated approach needed this removed I think, but the current targeted change can have this back so I'll restore these 👀

packages/browser/playwright/mocked/session-recording/session-rotation-scenarios.spec.ts
fasyy612: good point 🤔 I'll try locally commenting it out to see if it's making any difference.

fasyy612: `restarts recorder when session rotates externally while idle` unit test fails locally without the change 📝

Issue #2225, opened 2025-08-19
`captureImmediate` resolves before network request finishes


Narixius:
## Bug description

`captureImmediate` in the PostHog Node/Edge build fails to await or return the `_capture` promise, causing it to resolve before the underlying network request to PostHog has completed.

In Node.js this usually goes unnoticed (the process stays alive long enough for the request to complete), but in Cloudflare Workers / Next.js (via OpenNext), this causes events to be dropped since the runtime ends execution as soon as the returned promise resolves.

The code in `captureImmediate` method:

https://github.com/PostHog/posthog-js/blob/7183893a7f89fbaaab5e65d9e2e6986cf9d95bf2/packages/node/src/client.ts#L266-L273

This discards the promise. It should instead be:
```ts
.then(async additionalProperties => {
  await _capture({
    ...additionalProperties,
    ...properties,
    $groups: groups
  })
})

// or even

.then(additionalProperties => {
  return _capture({
    ...additionalProperties,
    ...properties,
    $groups: groups
  });
});

```

I’m calling `captureImmediate` inside `waitUntil` to keep the promise alive after responding to the user, but the issue is not related to `waitUntil`. Even a plain `await posthog.captureImmediate(...)` does not wait for the network request, because the `_capture` promise is dropped from the chain. As a result, the request may complete prematurely or be dropped entirely in environments that exit once the top-level promise resolves (e.g. Cloudflare Workers, Vercel Edge).

## How to reproduce

1. Deploy a Next.js app with OpenNext on Cloudflare Workers.
2. Call posthog.captureImmediate(...) inside an API route.
3. Wrap it with `ctx.waitUntil(posthog.captureImmediate(...))` to keep the Worker alive.
4. Observe that the Worker completes, but the event is not delivered, because the fetch is never awaited.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config

## Additional context
- Environment: Cloudflare Workers via OpenNext
- This issue only affects runtimes where completion of the returned promise is required to guarantee delivery (`waitUntil`).
- Fix is a one-line change: return/await _capture (PR is on the way 😄 ).

PR #3238, opened 2026-03-16
fix(node): captureExceptionImmediate does not return a promise to await

hamiltont:
## Problem

We hit an issue in production where server-side exceptions appeared in Vercel logs but never arrived in PostHog Error Tracking. After digging in, something smells wrong in `captureExceptionImmediate` -- there's no `return` statement on the promise. This is the same class of bug as #2225, which fixed a missing `return` inside `captureImmediate` (if this PR is accepted, it would be wise to do a quick scan of any other `*Immediate` functions and see if there are similar bugs in other function variations)

Without return, `await captureExceptionImmediate(...)` resolves immediately, so when the serverless function context exits, the event is lost (Vercel, Cloudflare Workers). PostHog docs recommend `await posthog.captureExceptionImmediate(...)` e.g. 
  * [NextJS Example Code](https://github.com/PostHog/posthog-js/blob/main/examples/example-nextjs/src/app/actions.ts)
  * [captureExceptionImmediate documentation](https://github.com/PostHog/posthog-js/blob/main/packages/node/src/client.ts#L1941)

Caveat: This was diagnosed with AI assistance, so I could be wrong about something If the missing `return` is intentional and there's a mechanism we've missed, I'd appreciate knowing why.

## Changes

```typescript
// before
async captureExceptionImmediate(error, distinctId?, additionalProperties?): Promise<void> {
  if (!ErrorTracking.isPreviouslyCapturedError(error)) {
    const syntheticException = new Error('PostHog syntheticException')
    this.addPendingPromise(   // ← no return; is this intentional?
      ErrorTracking.buildEventMessage(...).then((msg) => this.captureImmediate(msg))
    )
  }
}


// after
async captureExceptionImmediate(error, distinctId?, additionalProperties?): Promise<void> {
  if (!ErrorTracking.isPreviouslyCapturedError(error)) {
    const syntheticException = new Error('PostHog syntheticException')
    return this.addPendingPromise(
      ErrorTracking.buildEventMessage(...).then((msg) => this.captureImmediate(msg))
    )
  }
}
```

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -1970,7 +1970,7 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
   ): Promise<void> {
     if (!ErrorTracking.isPreviouslyCapturedError(error)) {
       const syntheticException = new Error('PostHog syntheticException')
-      this.addPendingPromise(
+      return this.addPendingPromise(
         ErrorTracking.buildEventMessage(error, { syntheticException }, distinctId, additionalProperties).then((msg) =>
           this.captureImmediate(msg)
         )


hpouillot (approved):
Nice, thanks for your contribution ! I reapplied the fix here, looks like there is an issue with the CI: https://github.com/PostHog/posthog-js/pull/3243

hamiltont:
@hpouillot I think this is the same type of bug (missing return means a caller cannot await) also in client.ts. Found in 5.34.7

Should I file a new issue, or is it minor enough that you could just confirm and add a return statement :-) 

Note: I stand by that "it would be wise to do a quick scan of any other *Immediate functions and see if there are similar bugs in other function variations" - now that we have found three different bugs of the same type in this codebase, i'd suspect there are a few other lurking about

```
  /**
   * Identify a user and set their properties immediately (synchronously).
   *
   * @example
   * ```ts
   * // Basic immediate identify
   * await client.identifyImmediate({
   *   distinctId: 'user_123',
   *   properties: {
   *     name: 'John Doe',
   *     email: 'john@example.com'
   *   }
   * })
   * ```
   *
   * {@label Identification}
   *
   * @param data - The identify data containing distinctId and properties
   * @returns Promise that resolves when the identify is processed
   */
  async identifyImmediate({ distinctId, properties = {}, disableGeoip }: IdentifyMessage): Promise<void> {
    // Catch properties passed as $set and move them to the top level
    const { $set, $set_once, $anon_distinct_id, ...rest } = properties
    // if no $set is provided we assume all rest properties are $set
    const setProps = $set || rest
    const setOnceProps = $set_once || {}
    const eventProperties = {
      $set: setProps,
      $set_once: setOnceProps,
      $anon_distinct_id: $anon_distinct_id ?? undefined,
    }
    super.identifyStatelessImmediate(distinctId, eventProperties, { disableGeoip })
  }
  ```

hpouillot:
@hamiltont Thanks for reporting this again ! It is fixed in the latest node release, no other immediate methods have this problem.

hamiltont:
@hpouillot no problem, AI discovered it when I upgraded to the newest and removed my work around, I asked it to sanity check that I got everything right. It randomly read the whole client file and pulled this other one out 🤗

PR #3240, opened 2026-03-16
feat: add _overrideSDKInfo method for wrapper SDKs to override $lib and $lib_version

marandaneto:
## Problem

Wrapper SDKs (e.g. posthog-flutter, posthog-react-native) that use the browser JS SDK underneath currently have no way to override the `$lib` and `$lib_version` event properties. These are hardcoded to `'web'` and the JS SDK version respectively, making it impossible to attribute events to the correct wrapper SDK.

## Changes

- Added `LIB_NAME` to the internal `Config` object (alongside existing `LIB_VERSION`)
- Replaced hardcoded `$lib: 'web'` with `$lib: Config.LIB_NAME` in `event-utils.ts` and `lazy-loaded-session-recorder.ts`
- Added `_overrideSDKInfo(sdkName, sdkVersion)` method on the PostHog class, marked as `@internal`
- Wrapper SDKs can call `posthog._overrideSDKInfo('posthog-flutter', '5.0.0')` after init to override both values
- Added `Config.JS_SDK_VERSION` to preserve the original JS SDK version for the `ver` HTTP request query parameter, ensuring server-side debugging/compatibility is unaffected by wrapper overrides

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/terser-mangled-names.json
+++ b/packages/browser/terser-mangled-names.json
@@ -355,6 +355,7 @@
         "_options",
         "_originalUserConfig",
         "_originatedFromCaptureException",
+        "_overrideSDKInfo",
         "_override_warning",
         "_pageHideListener",
         "_pageViewFallBack",


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1224,7 +1224,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
                     $snapshot_data: snapshotBuffer.data,
                     $session_id: snapshotBuffer.sessionId,
                     $window_id: snapshotBuffer.windowId,
-                    $lib: 'web',
```
marandaneto, line 1227: @PostHog/team-replay this will attribute recordings to flutter instead of web for flutter web apps cc @fivestarspicy 
recordings also have a source property: web|mobile we know if those flutter recordings are web or mobile

packages/browser/src/posthog-core.ts
```diff
@@ -2905,6 +2905,18 @@ export class PostHog implements PostHogInterface {
...
+     * @internal
+     * Allows wrapper SDKs (e.g. posthog-flutter, posthog-react-native) to override the
+     * `$lib` and `$lib_version` properties sent with every event.
+     *
+     * This is not a public API and may change without notice.
+     */
+    _overrideSDKInfo(sdkName: string, sdkVersion: string): void {
+        Config.LIB_NAME = sdkName
+        Config.LIB_VERSION = sdkVersion
+    }
```
marandaneto, line 2918: i prefer that over public fields in the posthog config file

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
fivestarspicy: thx manoel

PR #3242, opened 2026-03-16
feat(browser): Add support for pre-loaded remote-config

dustinbyrne:
## Problem

When using the upcoming snippet v2, the server provides version-pinned SDK assets via remote config. The browser SDK needs to support loading extensions (recorder, toolbar, etc.) from these versioned CDN paths instead of the default `/static/` route.

## Changes

- Added `sdkVersion` field to `RemoteConfig` type with `requested`, `resolved`, and optional `scriptBaseUrl` properties
- Added `_scriptBaseUrl` property to PostHog core, set from `config.sdkVersion.scriptBaseUrl` during remote config processing
- Updated external-scripts-loader to use the versioned base URL when `_scriptBaseUrl` is set, falling back to the existing `/static/` path for v1 snippet behavior
- `remote-config` scripts always load from the token-specific path regardless of `_scriptBaseUrl`
- Added unit tests for all new code paths

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/entrypoints/external-scripts-loader.ts
+++ b/packages/browser/src/entrypoints/external-scripts-loader.ts
@@ -109,6 +109,9 @@ assignableWindow.__PosthogExtensions__.loadExternalDependency = (
     // Default: load from /static/ via request router (snippet v1)
     let scriptUrlToLoad = `/static/${kind}.js` + `?v=${posthog.version}`
     if (kind === 'toolbar') {
+        // toolbar.js has a 24-hour CDN TTL but contains a rotating token valid for
+        // only 5 minutes. Bust the cache on a 5-minute boundary so the browser always
+        // fetches a fresh copy with a valid token.
         const fiveMinutesInMillis = 5 * 60 * 1000
         const timestampToNearestFiveMinutes = Math.floor(Date.now() / fiveMinutesInMillis) * fiveMinutesInMillis
         scriptUrlToLoad = `${scriptUrlToLoad}&t=${timestampToNearestFiveMinutes}`


packages/browser/src/entrypoints/external-scripts-loader.ts
```diff
@@ -90,23 +90,30 @@ assignableWindow.__PosthogExtensions__.loadExternalDependency = (
...
-        scriptUrlToLoad = `/array/${posthog.config.token}/config.js`
+        const url = posthog.requestRouter.endpointFor('assets', `/array/${posthog.config.token}/config.js`)
+        loadScript(posthog, url, callback)
+        return
+    }
+
+    // When the server provides a resolved SDK version (snippet v2),
+    // load from the version-prefixed path via the request router.
+    // No cache-busting needed — the version in the URL is the cache key.
+    if (posthog._resolvedSdkVersion) {
```
marandaneto, line 103: should we check for the `kind` here as well?
also, should we check for _resolvedSdkVersion even before the condition above?
i wonder how this works, is loadExternalDependency called multiple times with the same kind?

packages/browser/src/posthog-core.ts
```diff
@@ -613,6 +614,13 @@ export class PostHog implements PostHogInterface {
             this.sessionPropsManager = new SessionPropsManager(this, this.sessionManager, this.persistence)
         }
 
+        // Read resolved SDK version from pre-loaded config (snippet v2) before extensions
+        // initialize, so loadExternalDependency uses the versioned asset path
+        const preloadedConfig = assignableWindow._POSTHOG_REMOTE_CONFIG?.[this.config.token]?.config
+        if (preloadedConfig?.sdkVersion?.resolved) {
+            this._resolvedSdkVersion = preloadedConfig.sdkVersion.resolved
```
ioannisj: Probably a naive question here but is remote config guaranteed to be preloaded? if not, I don't see where we assign `_resolvedSdkVersion` elsewhere

dustinbyrne: It's served in the same document as array.js in the v2 web snippet. The preloaded remote config is executed and bound to the window before PostHog is initialized.

packages/browser/src/entrypoints/external-scripts-loader.ts
dustinbyrne: `kind` should be defined and reference an asset available in the version directory. The one outlier is `toolbar.js`, which is built and served from the Django app.

My decision (for now) is to build toolbar.js separately in CI and include it as a versioned asset.

We'd have better control over the versioning of the toolbar if it were a package in this repo. I think that's a reasonable future change worth making.

I don't believe it's ever called more than once for the same asset kind. If it is, it's likely in error. Though, we'd not be affecting this behavior either way.

The only thing before this is the special case for loading remote config. I think this will be unused but I'll circle back and double check.

packages/browser/src/posthog-core.ts
```diff
@@ -613,6 +614,13 @@ export class PostHog implements PostHogInterface {
...
+        }
```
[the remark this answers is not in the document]
dustinbyrne: I decided to drop this. If we need it, it can always be added back in later.

PR #3245, opened 2026-03-17
fix(flags): handle plain array and object forms in `overrideFeatureFlags`

dmarticus:
### Problem

`overrideFeatureFlags` in the browser SDK accepted plain arrays and plain objects in its type signature (`OverrideFeatureFlagsOptions`), but the implementation only handled two cases:
- `false` → clear overrides
- `{ flags: ..., payloads: ... }` → the explicit form

Passing a plain array like `['beta-feature']` or a plain object like `{'beta-feature': 'variant'}` would silently do nothing — the code checked for `'flags' in overrideOptions` which was `false` for these forms, so it fell through without registering any overrides.

This was reported by a customer who followed the (incorrect) docs and got no errors but no overrides either.

### Root cause

When `overrideFeatureFlags` was introduced in #1697, it only implemented the explicit `{ flags: ... }` form internally. The deprecated `override()` method papered over this by wrapping its arguments in `{ flags: ... }` before forwarding, so the old API kept working. But the type union and JSDoc both advertised the plain forms as valid for the new method — they just never worked.

When `overrideFeatureFlags` was later added to the node SDK in #2840, the contributor implemented all the forms the types promised (arrays, plain objects, and explicit), so the two SDKs diverged.

### Fix

Add the missing array and plain-object handling to the browser implementation, matching what the node SDK already does:

- **Plain array** (`['flag-a', 'flag-b']`): converts to `{ 'flag-a': true, 'flag-b': true }` and registers
- **Plain object** (`{'flag-a': 'variant'}`): registers directly as flag overrides (fallback after the `flags`/`payloads` check)

Also fixes the JSDoc in both `posthog-featureflags.ts` and `posthog.ts` to accurately document all supported forms.

### Changes

- `packages/browser/src/posthog-featureflags.ts` — Add array and plain-object branches to `overrideFeatureFlags`, fix JSDoc
- `packages/types/src/posthog.ts` — Fix JSDoc examples
- `packages/browser/src/__tests__/featureflags.test.ts` — Add 3 tests for the new shorthand forms

### DIFF
--- a/.changeset/honest-rats-raise.md
+++ b/.changeset/honest-rats-raise.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/types': patch
+---
+
+handle plain array and object forms in overrideFeatureFlags


packages/browser/src/posthog-featureflags.ts
```diff
@@ -61,6 +61,15 @@ const PERSISTENCE_FEATURE_FLAG_PAYLOADS = '$feature_flag_payloads'
 const PERSISTENCE_OVERRIDE_FEATURE_FLAG_PAYLOADS = '$override_feature_flag_payloads'
 const PERSISTENCE_FEATURE_FLAG_REQUEST_ID = '$feature_flag_request_id'
 
+/** Converts an array of flag names to a Record where each flag is set to true. */
+const arrayToFlagsRecord = (flags: string[]): Record<string, string | boolean> => {
```
gustavohstrassburger: The return type always has true in the values of the object, so `string | boolean` is actually broader than it could be.

packages/browser/src/posthog-featureflags.ts
```diff
@@ -1013,6 +1030,16 @@ export class PostHogFeatureFlags implements Extension {
...
+        if (overrideOptions && typeof overrideOptions === 'object') {
+            this._instance.persistence.register({
+                [PERSISTENCE_OVERRIDE_FEATURE_FLAGS]: overrideOptions as Record<string, string | boolean>,
+            })
+            this._fireFeatureFlagsCallbacks()
+
+            return forceDebugLogger.info('Flag overrides set', { flags: overrideOptions })
+        }
+
         this._fireFeatureFlagsCallbacks()
```
gustavohstrassburger: In any scenario, is it expected to get here?

If not, we should probably add a log here.

And maybe we shouldn't even be calling `_fireFeatureFlagsCallbacks` (I wouldn't change that, but this seems like an error).

PR #3247, opened 2026-03-17
fix(browser): prevent silent identity switch during bootstrap and auto-identify anonymous users

dmarticus:
## Problem

When a server-side rendered page bootstraps feature flags for an identified user, but the browser has a stale anonymous identity in local persistence from a previous session, there's a mismatch between the identity used for flag evaluation on the server vs. client. This can cause:

1. Inconsistent feature flag behavior between server and client renders
2. Duplicate `$feature_flag_called` events for the same user
3. Analytics data attributed to the wrong identity

Additionally, the existing bootstrap logic would silently overwrite an already-identified user's `distinct_id` with a different bootstrap ID, causing an untracked identity switch that corrupts analytics.

Context: https://posthog.slack.com/archives/C07Q2U4BH4L/p1773674684145869

## Changes

**Auto-identify anonymous users on bootstrap**: When bootstrapping with `isIdentifiedID: true` and the local persistence has an anonymous user with a different `distinct_id`, we now automatically call `identify()` to properly merge the anonymous user into the identified user. This ensures the `$identify` event is fired with the correct `$anon_distinct_id`, enabling proper identity resolution on the server.

**Prevent silent identity switches**: When an already-identified user would be overwritten by a different bootstrap ID, we now log a warning and preserve the existing identity instead of silently switching. Users who need to switch identities should call `reset()` before reinitializing.

**Test improvements**: Added comprehensive test coverage for the new behavior, including verification of the `$identify` event payload and edge cases like `isIdentifiedID: undefined`.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/posthog-core-also.test.ts
+++ b/packages/browser/src/__tests__/posthog-core-also.test.ts
@@ -3,6 +3,7 @@ import { mockLogger } from './helpers/mock-logger'
 import * as globals from '../utils/globals'
 import { document, window } from '../utils/globals'
 import { uuidv7 } from '../uuidv7'
+import { isUndefined } from '@posthog/core'
 import { ENABLE_PERSON_PROCESSING, USER_STATE } from '../constants'
 import { createPosthogInstance, defaultPostHog } from './helpers/posthog-instance'
 import { PostHogConfig, RemoteConfig } from '../types'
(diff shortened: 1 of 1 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -633,13 +633,49 @@ export class PostHog implements PostHogInterface {
...
+            const existingDistinctId = this.get_distinct_id()
+            const existingUserState = this.persistence.get_property(USER_STATE)
+
+            if (
+                config.bootstrap.isIdentifiedID &&
+                existingDistinctId != null &&
+                existingDistinctId !== bootstrapDistinctId &&
+                existingUserState === 'anonymous'
+            ) {
+                // The server bootstrapped flags for an identified user, but local persistence
```
dustinbyrne: nit picky
```suggestion
                // The server bootstrapped identity for an identified user, but local persistence
```

PR #3257, opened 2026-03-19
fix: pin release workflow to trigger commit SHA instead of latest main

marandaneto:
## Problem

The release workflow checks out `ref: main` (latest HEAD) in both the `check-changesets` and `version-bump` jobs, rather than the commit that actually triggered the workflow. Since the workflow requires manual approval via the `NPM Release` environment gate, there can be a delay between trigger and approval. Any commits merged to `main` during that window get silently included in the release — meaning what's published may not match the commit that was intended to be released.

## Changes

- **Pin checkouts to trigger commit**: Both `check-changesets` and `version-bump` now check out `github.event.pull_request.merge_commit_sha` (or `github.sha` for `workflow_dispatch`) instead of `ref: main`.
- **Release branch pattern**: The `version-bump` job creates a `release/<run_id>` branch from the pinned trigger commit. The version bump and lockfile update are committed to this branch instead of directly to `main`.
- **Publish from release branch**: The `publish` job (already correctly using the `commit-hash` output) now publishes code from the release branch commit. GitHub releases target the specific commit hash instead of `main`.
- **New `merge-release-branch` job**: After all packages publish successfully, the release branch is merged back to `main` (so version bumps and changelog updates land on main) and then deleted. Uses the app token to handle branch protection.
- **Notification dependency update**: `notify-released` now depends on `merge-release-branch` so the success Slack message only fires after everything is complete.

context: https://posthog.slack.com/archives/C0ACMS2BNTU/p1773857143048669

### Libraries affected

- [ ] All of them

No library code changes — this is a CI workflow fix only.

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.github/workflows/release-merge.yml
+++ b/.github/workflows/release-merge.yml
@@ -45,7 +45,7 @@ jobs:
                     -f commit_message="chore: merge $BRANCH_NAME into main [skip ci]"
 
             - name: Delete release branch
-              if: always() && steps.check-branch.outputs.exists == 'true'
+              if: success() && steps.check-branch.outputs.exists == 'true'
               env:
                   GH_TOKEN: ${{ github.token }}
                   BRANCH_NAME: ${{ steps.check-branch.outputs.branch-name }}


.github/workflows/release.yml
```diff
@@ -333,11 +342,46 @@ jobs:
...
+                  fetch-depth: 0
+                  token: ${{ steps.releaser.outputs.token }}
+
+            - name: Merge release branch to main
+              run: |
+                  git config user.name "github-actions[bot]"
+                  git config user.email "github-actions[bot]@users.noreply.github.com"
+                  git fetch origin ${{ needs.version-bump.outputs.release-branch }}
+                  git merge origin/${{ needs.version-bump.outputs.release-branch }} --no-edit
+                  git push origin main
```
Piccirello: Merging this way will result in unsigned commits, which are now blocked in this repo. You'll need to either use the `gh` cli or use the `planetscale/ghcommit-action` action used elsewhere in this file.

marandaneto: https://github.com/PostHog/posthog-js/pull/3257/changes/992b52bd87ad7cb23c982ff7cfa5aa30e1518818

.github/workflows/release.yml
```diff
@@ -333,11 +342,42 @@ jobs:
                   message: '❌ Failed to release `${{ matrix.package.name }}@v${{ steps.check-package-version.outputs.committed-version }}`! <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|View logs>'
                   emoji_reaction: 'x'
 
+    merge-release-branch:
+        name: Merge release branch back to main
+        needs: [version-bump, publish]
+        runs-on: ubuntu-latest
+        if: always() && needs.publish.result == 'success' && needs.version-bump.outputs.release-branch != ''
+        environment: 'NPM Release' # Re-uses the already-approved environment from version-bump (no second approval needed)
```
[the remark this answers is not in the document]
marandaneto: have to figure this out, we dont want to approve twice

marandaneto:
@Piccirello will wait for your review since this changed a bit

.github/workflows/release-merge.yml
```diff
@@ -0,0 +1,81 @@
...
+            - name: Merge release branch to main via API
+              if: steps.check-branch.outputs.exists == 'true'
+              env:
+                  GH_TOKEN: ${{ github.token }}
+                  BRANCH_NAME: ${{ steps.check-branch.outputs.branch-name }}
+              run: |
+                  gh api "repos/${{ github.repository }}/merges" \
+                    -f base="main" \
+                    -f head="$BRANCH_NAME" \
+                    -f commit_message="chore: merge $BRANCH_NAME into main [skip ci]"
```
Piccirello, line 45: I believe this will be blocked due to the lack of a PR. You'll need to use a GitHub app to work around this. Which maybe means a second GitHub app per repo? 😬

.github/workflows/release-merge.yml
```diff
@@ -0,0 +1,81 @@
...
+    merge-release-branch:
+        name: Merge release branch back to main
+        runs-on: ubuntu-latest
+        if: github.event.workflow_run.conclusion == 'success'
+        steps:
+            - name: Check if release branch exists
+              id: check-branch
+              env:
+                  GH_TOKEN: ${{ github.token }}
+                  BRANCH_NAME: release/${{ github.event.workflow_run.id }}
```
Piccirello, line 26: If we move forward with this I'll add an additional ruleset to lock down `release/*` branches, otherwise we're opening ourselves up to some other attacks, like pushing directly to a release branch without a PR.

Piccirello (commented):
I like this idea a lot in theory, but now things are getting even more complicated.

.github/workflows/release-merge.yml
marandaneto: uff, i think indeed this is getting way too complicated

.github/workflows/release-merge.yml
marandaneto: good point

PR #3260, opened 2026-03-19
fix: Instead of throwing, log as an error if the recorder isn't yet fully available

kyleswank:
## Problem

We have a customer with an error popping up in their error tracker related to this thrown error. @pauldambra believes it is due to ad blockers. I wasn't able to fully recreate it but instead of throwing an error we can log here with the early return and hopefully not break anything terribly. 

## Changes

Changes a `throw` to a `logger.error`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.github/workflows/check-posthog-major-version.yml
+++ b/.github/workflows/check-posthog-major-version.yml
@@ -24,6 +24,9 @@ jobs:
         
       - name: Check for major version bumps
         id: check-major
+        env:
+          # Avoid wrapping labels JSON in shell single quotes; apostrophes in label text break the script.
+          PR_LABELS_JSON: ${{ toJSON(github.event.pull_request.labels) }}
         run: |
           # Get only added changeset files in this PR
           changesets=$(git diff --name-only --diff-filter=A origin/${{ github.base_ref }}...HEAD | grep '^\.changeset/.*\.md$' | grep -v 'README.md' || true)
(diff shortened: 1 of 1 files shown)


packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -269,7 +269,8 @@ export class SessionRecording implements Extension {
 
     private _onScriptLoaded(startReason?: SessionStartReason) {
         if (!assignableWindow.__PosthogExtensions__?.initSessionRecording) {
-            throw Error('Called on script loaded before session recording is available')
+            logger.error('Called on script loaded before session recording is available')
```
pauldambra:
```suggestion
            logger.error('Called on script loaded before session recording is available. This can be caused by adblockers')
            this.instance.registerProperty('sdk_debug_recording_script_not_loaded': true)
```

that instance.registeFoo is from memory so definitely wrong...
probably want to register for session instead of permantently

kyleswank: Used register_for_session with a new property: $sdk_debug_recording_script_not_loaded

I also have a PR queued up for posthog to add it to the taxonomy.py

marandaneto: oh damn just saw this
i opened https://github.com/PostHog/posthog-js/pull/3266 but happy to keep this one

marandaneto: @kyleswank also add a changeset entry and the release label so we can publish a new version

kyleswank: @marandaneto changeset added, I just used yours, thanks for that! Also added the release label.

kyleswank: @marandaneto how do I fix these two tests that are failing? The version one doesn't make much sense to me and the playwright tests, as far as I can tell, aren't related to the code change.

pauldambra:
ok, ci green, ready when you are

PR #3261, opened 2026-03-19
feat(node): add versioned defaults and strictCapture option

lricoy:
## Problem

We have seen some cases of people using the posthog-js method signature for capturing events like `capture(event: string, properties?: PostHogEventProperties, options?: PostHogCaptureOptions): void` instead of `capture(props: EventMessage): void` this would bring up a warning log `Called capture() with a string as the first argument when an object was expected.` but many undesired behavior as well.

## Changes

Add a new `defaults` option to avoid a breaking change and a new `strictCapture` config that when combined with the new default config option will make the code path throw instead of just warning.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/node/examples/strict-capture-test.mjs
+++ b/packages/node/examples/strict-capture-test.mjs
@@ -1,4 +1,5 @@
 #!/usr/bin/env node
+/* eslint-env node */
 /**
  * Strict Capture Demo
  *
@@ -8,13 +9,17 @@
  *
  * Usage:
  *   node examples/strict-capture-test.mjs
+ *   POSTHOG_PROJECT_API_KEY=phc_... node examples/strict-capture-test.mjs
  */
 
 const { PostHog } = await import('../dist/entrypoints/index.node.mjs')
 
+const API_KEY = process.env.POSTHOG_PROJECT_API_KEY || 'fake-key'
+const HOST = process.env.POSTHOG_HOST || 'http://localhost:8000'
+
 // --- Strict mode (via versioned defaults) ---
-const strict = new PostHog('fake-key', {
-  host: 'http://localhost:8000',
+const strict = new PostHog(API_KEY, {
+  host: HOST,
   defaults: '2026-03-19',
 })
 
@@ -30,8 +35,8 @@ try {
 }
 
 // --- Legacy mode (no defaults) ---
-const legacy = new PostHog('fake-key', {
-  host: 'http://localhost:8000',
+const legacy = new PostHog(API_KEY, {
+  host: HOST,
 })
 legacy.debug(true)
 


packages/node/src/client.ts
```diff
@@ -431,7 +440,11 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
    */
   capture(props: EventMessage): void {
     if (typeof props === 'string') {
-      this._logger.warn('Called capture() with a string as the first argument when an object was expected.')
+      const msg = 'Called capture() with a string as the first argument when an object was expected.'
+      if (this.options.strictCapture) {
+        throw new TypeError(msg)
```
marandaneto, line 445: IMO SDKs should never throw, SDKs should degrade gracefully and warn/log out if something is wrong (which is whats happening here).
Curious why people called with the wrong params tho, though. Was it not up-to-date documentation? LLMs guessing public APIs?

pauldambra: really silly question... why not accept both?
we have a situation where we'd like the SDKs to be similar but they're not
LLMs are gonna make the mistake
just have them all accept all the options 🤷

marandaneto: all fine aligning the behaviour.
my comment is about not crashing peoples apps in production, i prefer a degrade gracefully approach/log out rather than being the one to blame, this is a best practices approach for sdks.
i'd be ok to throw if its not a production build.

pauldambra: yep, agree!

lricoy: Sounds good, but would we still change the behavior with the new `strictCapture` or. `defaults` config so it is not a breaking change, or do you two think we can change the behavior in a non-production build?

For context, I've seen this on 3 occasions that I recall:

- LLM hallucinated or chose the wrong SDK options
- No TS, so the function signature wasn't checked
- Used on a file that ran on both server/frontend with a custom bundling/wrapper step, so the wrong SDK was being used

marandaneto: i think its always ok if we throw so agents or humans would know they have to fix something during dev/debug builds
i think its not ok if its a release/production build

a very simple example

sdk throws if no api token is set, makes a lot of sense in theory
the api token is set from build vars, ci, etc
well that build var, ci or something was unset for reasons, bug happens, permissions, etc
you deployed a new version anyway
now all of your users cant use your app because the sdk left the app in a broken state

this is obv the worst case scenario

lricoy: Yeah, makes sense, I am on board. Thanks @marandaneto !

Issue #3252, opened 2026-03-18
@posthog/nextjs-config requires projectId during next typegen when sourcemaps.enabled is false


jakeleventhal:
## Bug description

`@posthog/nextjs-config` now throws during config evaluation if `projectId` is missing, even when source map handling is explicitly disabled.

That breaks non-build commands like `next typegen`, where the only goal is route type generation and no PostHog source map upload should run.

## Reproduction

This reproduces with `@posthog/nextjs-config@1.8.21` and `next@16.1.7`.

`package.json`

```json
{
  "name": "posthog-nextjs-config-repro",
  "private": true,
  "scripts": {
    "typegen": "next typegen"
  },
  "dependencies": {
    "@posthog/nextjs-config": "1.8.21",
    "next": "16.1.7",
    "react": "19.2.0",
    "react-dom": "19.2.0"
  }
}
```

`next.config.mjs`

```js
import { withPostHogConfig } from "@posthog/nextjs-config";

export default withPostHogConfig(
  {},
  {
    personalApiKey: "test-key",
    sourcemaps: {
      enabled: false
    }
  }
);
```

`app/page.js`

```js
export default function Page() {
  return "hello";
}
```

Then run:

```bash
bun install
bun run typegen
```

## Actual behavior

`next typegen` fails immediately while loading `next.config.mjs`:

```text
$ next typegen
Unhandled Rejection: Error: projectId is required (envId is deprecated)
    at <unknown> (next.config.mjs:3:16)
```

## Expected behavior

If `sourcemaps.enabled` is `false`, the config should not require `projectId`/`envId` up front.

In that case I would expect commands like `next typegen` to proceed normally, since no source map processing or upload should happen.

## Regression

The same repro passes on `@posthog/nextjs-config@1.8.15`:

```text
$ next typegen
Generating route types...
✓ Types generated successfully
```

From what I can tell, this comes from the dependency bump to `@posthog/webpack-plugin@1.2.27`, where `resolveConfig` now throws if neither `projectId` nor `envId` is provided before checking whether sourcemaps are enabled.

It looks like the validation should either:

1. Only happen when source map processing is actually enabled, or
2. Be deferred until the plugin is about to do source map work.

PR #3265, opened 2026-03-20
chore(et): clean config building

hpouillot:
## Problem

- Code is duplicated
- Validation is done on resolution causing issue when sourcemaps upload is disabled

Fix: https://github.com/PostHog/posthog-js/issues/3252

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [x] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [x] @posthog/rollup-plugin
- [x] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/funky-eggs-serve.md
+++ b/.changeset/funky-eggs-serve.md
@@ -1,0 +1,8 @@
+---
+'@posthog/webpack-plugin': patch
+'@posthog/nextjs-config': patch
+'@posthog/rollup-plugin': patch
+'@posthog/core': patch
+---
+
+fix sourcemap upload with stdin, clean config


packages/core/src/process/config.ts
```diff
@@ -0,0 +1,86 @@
...
+    /** @deprecated Use releaseVersion instead */
+    version?: string
+    releaseVersion?: string
+    deleteAfterUpload?: boolean
+    batchSize?: number
+  }
+}
+
+export interface ResolvedPluginConfig extends Omit<PluginConfig, 'envId' | 'projectId'> {
+  projectId?: string
```
cat-ph, line 27: thought (more like nit): curious whether we should not have this optional (since it's the "resolved" config, and I think we're "technically" also changing the exposed types, although it's more "permissive"), or have a union type with sourcemaps.enabled: false and projectId?: string or sourcemaps.enabled: true and projectId: string (but that might cause.. other issues 😬)

hpouillot: I agree I think an union would be cleaner, maybe we can update this when we remove the deprecated keys. The `ResolvedPluginConfig` is only used internally for now (users should only use the `PluginConfig` interface) so it should be fine, but let's keep it in mind.

PR #3268, opened 2026-03-21
fix for ssr flags for posthog/next

firstarcprime:
## Problem

@dustinbyrne 

There is a hydration error with @posthog/next using SSR with Nextjs 16 App Router. The docs (USAGE.md and README.md) were followed exactly for setting up (minus how I had to use proxy.ts instead of middleware.ts as the latter is deprecated). 

After investigating, I found that the flags that were fetched by the PosthogProvider were not being passed to the ClientPosthogProvider during SSR; they seem to only be passed along when posthogJS is initialised. From what I could debug, this leads to the feature flags being undefined on the server and then defined on the client, which flickers the UI. 

I performed this change as a patch and it finally started working as I expected; the UI would load with the correct feature flag on both the server and the client. I'll admit that I'm not a good enough engineer to know exactly how hacky this change is, so I was hoping somebody with more experience could take a look. This is my first contributing PR, so apologies for mistakes/inconveniences.

## Changes

I removed the window check and added a call to updateFlags if the user enables bootstrapping on the PosthogProvider. This seemed to resolve the issue, though I have no idea if there will be any hidden consequences.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/next/src/client/ClientPostHogProvider.tsx
+++ b/packages/next/src/client/ClientPostHogProvider.tsx
@@ -4,6 +4,7 @@ import React from 'react'
 import posthogJs from 'posthog-js'
 import { PostHogProvider as ReactPostHogProvider } from 'posthog-js/react'
 import type { BootstrapConfig, PostHogConfig } from 'posthog-js'
+import { usePathname } from 'next/navigation'
 
 export type { BootstrapConfig }
 
@@ -39,17 +40,14 @@ export function ClientPostHogProvider({ apiKey, options, bootstrap, children }: 
     }
 
     const mergedOptions = bootstrap ? { ...options, bootstrap: { ...options?.bootstrap, ...bootstrap } } : options
-
     // Initialize eagerly during render on the client so that child effects
     // see a fully configured posthog instance. The `__loaded` guard prevents
     // double-init (e.g. React StrictMode).
     if (!posthogJs.__loaded) {
         posthogJs.init(apiKey, mergedOptions)
-    }
-
-    if (bootstrap?.featureFlags) {
+    } else if (bootstrap?.featureFlags) {
         // If bootstrapping, update feature flags from the server
-        posthogJs.updateFlags(bootstrap.featureFlags)
+        posthogJs.updateFlags(bootstrap.featureFlags, bootstrap.featureFlagPayloads)
     }
 
     return <ReactPostHogProvider client={posthogJs}>{children}</ReactPostHogProvider>


packages/next/src/client/ClientPostHogProvider.tsx
```diff
@@ -43,9 +43,14 @@ export function ClientPostHogProvider({ apiKey, options, bootstrap, children }:
...
     // double-init (e.g. React StrictMode).
-    if (typeof window !== 'undefined' && !posthogJs.__loaded) {
+    if (!posthogJs.__loaded) {
         posthogJs.init(apiKey, mergedOptions)
     }
 
+    if (bootstrap?.featureFlags) {
+        // If bootstrapping, update feature flags from the server
+        posthogJs.updateFlags(bootstrap.featureFlags)
+    }
```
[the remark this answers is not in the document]
dustinbyrne: I don't think we need the useRef here, but we should only either `init` or `updateFlags`. if/else seems reasonable.

dustinbyrne, line 51: Also consider payloads
```suggestion
    if (!posthogJs.__loaded) {
        posthogJs.init(apiKey, mergedOptions)
    } else if (bootstrap?.featureFlags) {
        // If bootstrapping, update feature flags from the server
        posthogJs.updateFlags(bootstrap.featureFlags, bootstrap.featureFlagPayloads
    }
```

firstarcprime: A configurable option sounds great; personally, I think always rendering on the server is a sensible default, as the flicker can be a massive deal for user experience. Users may not notice a small increase in load time, but they'll definitely notice the flicker, and it could make the website owners seem like they're running a visibly buggy site. Though, off the top of my head, I'm not sure how bootstrapping only on first page load could be achieved, and whether this could reintroduce the flicker.

In response to the useEffect you mentioned, I've actually tried testing this and I think it might cause another issue; I get a NextJs error saying that it can't update one of my components at the same time as ClientPosthogProvider.

Thanks for the correction on the "else if" and the flag payloads.

firstarcprime:
Added the if/else and payloads

dustinbyrne:
Looking at this again with fresh eyes, the core issue appears to be that the server-side never writes `config.bootstrap.featureFlags`, while our hooks have explicit cases for these bootstrapped feature flags. This    
causes the server to return the flag value as undefined, while the client is able to resolve the actual value.

This PR fixes the issue by allowing the server-side to call into posthog-js. While this works, ideally we should avoid calling the browser SDK from the server.

I've put together a separate fix for this [here](https://github.com/PostHog/posthog-js/pull/3376) and cited you as a co-author. I'm going to close this PR, but feel free to re-open if this solves a separate issue.

PR #3270, opened 2026-03-22
fix(react-native): prevent set -e from aborting xcode script when npm is not in PATH

cat-ph:
## Problem

Xcode Archive builds fail when `posthog-xcode.sh` cannot find `posthog-cli`. The script tries `npm prefix -g` and `npm root` to locate the CLI, but Xcode runs build phases with a stripped `PATH` that excludes nvm, fnm, and Homebrew (Apple Silicon) directories. When `npm` is not in that PATH, both commands exit with code 127. Because `set -e` is active throughout the script, this causes an immediate abort before the fallback block (which extends `PATH` with common locations like `/opt/homebrew/bin`) is ever reached.

Reported by users on monorepo setups using Expo CNG with `@posthog/cli` installed globally via nvm or Homebrew on Apple Silicon.

## Changes

Added `|| true` to the three command substitutions used to locate the CLI:

- `npm prefix -g` - global npm prefix lookup
- `npm root` - local npm root lookup
- `command -v posthog-cli` - final fallback search

This prevents `set -e` from aborting the script on a missing `npm` and allows the PATH-extended fallback to run as intended.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### DIFF
--- a/.changeset/loud-streets-raise.md
+++ b/.changeset/loud-streets-raise.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+prevent xcode build abort when npm is not in PATH


packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -36,18 +36,18 @@ if [ -f "$HOME/.posthog/posthog-cli" ]; then
...
     PH_CLI_PATH="$NPM_GLOBAL_PREFIX/bin/posthog-cli"
   else
     # Check if installed as local dependency
-    NPM_LOCAL_ROOT=$(npm root 2>/dev/null)
+    NPM_LOCAL_ROOT=$(npm root 2>/dev/null || true)
     if [ -n "$NPM_LOCAL_ROOT" ] && [ -f "$NPM_LOCAL_ROOT/.bin/posthog-cli" ]; then
       PH_CLI_PATH="$NPM_LOCAL_ROOT/.bin/posthog-cli"
     else
       # Fallback to searching common locations
       export PATH="/usr/local/bin:/opt/homebrew/bin:$HOME/.cargo/bin:$HOME/.local/bin:$HOME/.posthog:$PATH"
```
cat-ph, line 49: thought: might still have cases where we don't catch this (i.e. other specific node version managers), but at least this version prevents a direct exit when npm isn't found and gives this fallback path a chance

PR #3281, opened 2026-03-24
feat(replay): account for V2 trigger groups

ksvat:
## Problem

Introduce V2 trigger functionality to sdk

#### Implements V2 trigger groups for session recording with per-group sampling, per-group minimum duration checking, and per-group conditions.  
  
Introduction of Strategy Pattern

The implementation introduces a RecordingStrategy interface with two concrete implementations:

1. V1RecordingStrategy

- Supports V1 trigger matching 

1. V2RecordingStrategy (New Implementation)

- Handles multiple independent trigger groups with union behavior
- Each group has its own:
    - Conditions (events/URLs/flags with any or all matching)
    - Sample rate (always required, 0-1)
    - Minimum duration (optional, 0-30000ms)
- Uses the **lowest minimum duration** from all matched groups
- Makes per-group sampling decisions stored in persistence

##### Strategy Methods

```
  interface RecordingStrategy {
      // Core status resolution - determines whether to record
      getStatus(context: RecordingStrategyContext): SessionRecordingStatus

      // Make sampling decisions for the current session
      makeSamplingDecisions(sessionId: string): void

      // Check URL triggers on navigation
      checkUrlTriggers(sessionId: string, onPause, onResume, onActivate): void

      // Setup event trigger listeners
      setupEventTriggerListeners(onEvent, sessionId, onActivate): void

      // Update session properties with active trigger info
      updateActiveTriggers(): void

      // Check if any triggers are pending (waiting)
      hasPendingTriggers(sessionId: string): boolean

      // Get the minimum duration requirement
      getMinimumDuration(sessionId: string): number | null
  }
```

New Logging

- All matched groups (regardless of sampling) are tracked in:
```$sdk_debug_replay_matched_recording_trigger_groups: [`` ``{ id: '1234', name: 'Error Tracking', matched: true, sampled: true },`` ``{ id: 'premium', name: 'Premium Users', matched: true, sampled: false }`` ``]```

Slight Performance Optimization

- Added _hasCompletedInitialFlush flag:
- Before flush: Check all triggers on every event (allows later groups with lower minDuration to reduce wait time)
- After flush: Stop checking triggers (we're already recording)

### tests

1. Session properties registration - Verifies tracking of matched/sampled groups
2. URL blocklist priority - Confirms blocklist still blocks V2 triggers
3. Union behavior - Multiple groups triggering and being tracked
4. Empty conditions - Groups with no conditions trigger immediately

### Backward Compatibility

V1 behavior is unchanged

- V2 only activates when version: 2 is present in remote config
- V1 continues to work exactly as before when version !== 2
- No breaking changes to existing configurations

[Screen Recording 2026-03-24 at 9.33.26 PM.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/4b239ca4-0849-431a-af37-01f5bf89cdf2.mov)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### CODE REVIEW

File: packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts [hunk 1]
```diff
@@ -395,12 +404,37 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
     }
 
     private get _minimumDuration(): number | null {
```
pauldambra, line 412: 🤖 _from paul's review patterns:_

i count at least 5 places where the code checks `config?.version === 2 && ...` with slightly different formulations — some check `config.triggerGroups`, some check `this._triggerGroupMatchers.length`.

i wonder if there's a single `private get _isV2(): boolean` that captures this? right now if someone adds a new V2 path they have to know which flavour of the check to copy, and if they get it subtly wrong it'll be one of those bugs where V2 is "mostly working" but breaks on one specific path. not blocking, just feels like coupling-through-convention.

packages/browser/src/extensions/replay/external/triggerMatching.ts
```diff
@@ -406,6 +425,84 @@ export class EventTriggerMatching implements TriggerStatusMatching {
...
+            // Empty conditions = trigger immediately on session start
+            this._combinedMatching = new AlwaysActivatedTriggerMatching()
+        } else {
+            // Convert group config to the format expected by the individual matchers
+            const config = {
+                urlTriggers: group.conditions.urls,
+                eventTriggers: group.conditions.events,
+                linkedFlag: group.conditions.flag,
+                urlBlocklist: [], // groups don't have blocklist
+            } as any
```
pauldambra: 🤖 _from paul's review patterns:_

the `as any` cast here means if someone renames `urlTriggers` to `urlPatterns` or whatever, the compiler won't tell you this broke.

i wonder if it's worth making a narrow type that represents just the fields the individual matchers actually need? or even just `as Pick<SessionRecordingRemoteConfig, 'urlTriggers' | 'eventTriggers' | ...>`? that way the compiler is still in the loop.

not blocking — the tests would catch it — but i'm slightly allergic to `as any` in code that's this important to get right.

---

🤖 _from c2 wiki / XP lens:_

the deeper issue is that `URLTriggerMatching`, `EventTriggerMatching`, and `LinkedFlagMatching` each accept the full remote config bag and reach in for the 2 fields they care about — that's Feature Envy. if each matcher accepted just its own slice (`EventTriggerMatching.onConfig({ eventTriggers: string[] })`), `TriggerGroupMatching` could pass the right data without a cast. bigger change, not necessarily for this PR, but it's the direction that removes the `as any` properly.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1139,18 +1217,38 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
     }
 
     get status(): SessionRecordingStatus {
```
pauldambra, line 1161: 🤖 _from paul's review patterns:_

the `status` getter casts `this._statusMatcher as (triggersStatus: RecordingTriggersStatusV2) => ...` — nothing prevents someone from setting `_statusMatcher = anyMatchSessionRecordingStatus` (V1) then calling it with V2 args. the cast silences the compiler.

i wonder if the V2 path could just call `triggerGroupsMatchSessionRecordingStatus` directly? V2 always uses the same function. that'd remove the need for the cast entirely in the V2 branch.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1543,6 +1647,130 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            this._triggerGroupMatchers.push(matcher)
+        }
+
+        // Setup event trigger listeners for all groups
+        this._removeEventTriggerCaptureHook?.()
+        this._removeEventTriggerCaptureHook = this._instance.on('eventCaptured', (event: CaptureResult) => {
+            // Performance optimization: Stop checking triggers after initial buffer flush
+            // This allows all trigger groups to be evaluated during the buffering period (for minDuration calculation)
+            // but stops wasting CPU after we've already sent data to the server
+            if (this._hasCompletedInitialFlush) {
```
pauldambra: 🤖 _from paul's review patterns:_

love the idea of stopping trigger checks after the first flush. but if this goes wrong (stops checking too early), how would we know?

silly question — would it be worth a debug log line when the hook gets removed? something like `logger.info('V2 trigger checking stopped after initial flush')`. cheap insurance for debugging in production.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1543,6 +1647,130 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+                logger.error('Could not activate event trigger for trigger groups', e)
+            }
+        })
+    }
+
+    /**
+     * V2: Make sampling decisions for all trigger groups
+     * Implements union behavior: evaluate ALL groups, if ANY sample rate hits → record
+     */
+    private _makeTriggerGroupSamplingDecisions(sessionId: string): void {
```
pauldambra: 🤖 _from paul's review patterns:_

i'm lazily asking... are group IDs always safe for use as persistence key suffixes in `$posthog_sr_group_sampling_<groupID>`? if someone creates a group with an ID containing dots or brackets, does that mess with the persistence layer? probably fine if IDs are server-generated UUIDs, but if they're user-defined strings then 😬

packages/browser/src/extensions/replay/external/triggerMatching.ts
```diff
@@ -513,3 +610,76 @@ export function allMatchSessionRecordingStatus(triggersStatus: RecordingTriggers
...
+    let anyGroupSampled = false
+
+    for (const matcher of groupMatchers) {
+        const groupStatus = matcher.triggerStatus(triggersStatus.sessionId)
+
+        if (groupStatus === TRIGGER_ACTIVATED) {
+            anyGroupActivated = true
+            // Check if this group's sample rate hit
+            const groupId = matcher.group.id
+            if (samplingResults.get(groupId) === true) {
```
pauldambra: 🤖 _from c2 wiki / XP lens:_

what's the intended meaning when a group is `TRIGGER_ACTIVATED` but `samplingResults` has no entry? the code returns `ACTIVE`, and the edge case test covers it, but the _intent_ is ambiguous: does `ACTIVE` mean "we're recording but not because of sampling" or "something went wrong and we're recording as a fallback"?

if it truly shouldn't happen, a log warning would surface the bug. if it _can_ happen (race between trigger activation and sampling decision), the behavior should be expressed in the function name or structure rather than relying on the comment.

packages/browser/src/types.ts
```diff
@@ -488,6 +498,30 @@ export interface SessionRecordingUrlTrigger {
...
+        events?: string[]
+        urls?: SessionRecordingUrlTrigger[]
+        flag?: string | FlagVariant
+    }
+}
+
+/**
+ * V2 Trigger Groups configuration from remote config
+ */
+export interface SessionRecordingTriggerGroupsConfig {
```
pauldambra: 🤖 _from c2 wiki / XP lens:_

`SessionRecordingTriggerGroupsConfig` is defined here but not imported or used anywhere. Rule 4 of Simple Design: no superfluous parts. If you need it later, you'll write it then and it'll match what you actually need. YAGNI says delete it.

packages/browser/src/__tests__/extensions/replay/triggerGroups.test.ts
```diff
@@ -0,0 +1,362 @@
...
+                },
+            }
+
+            const matcher = new TriggerGroupMatching(fakePostHog, group, () => {})
+            expect(matcher.group.minDurationMs).toBe(0)
+        })
+    })
+
+    describe('triggerGroupsMatchSessionRecordingStatus - Basic States', () => {
+        const createMockMatcher = (
```
pauldambra: 🤖 _from both paul's review patterns and c2 wiki:_

`createMockMatcher` is defined three times across the describe blocks in this file with minor variations (one adds a `sampleRate` param). OnceAndOnlyOnce — a single version with optional overrides would deduplicate:

```typescript
const createMockMatcher = (
    id: string,
    triggerStatus: TriggerStatus,
    overrides?: Partial<{ sampleRate: number }>
): TriggerGroupMatching => ({
    group: { id, name: `Group ${id}`, sampleRate: overrides?.sampleRate ?? 1.0, conditions: { matchType: 'any' } },
    triggerStatus: () => triggerStatus,
    stop: () => {},
} as unknown as TriggerGroupMatching)
```

also — these tests are excellent candidates for parameterized tests (`test.each`). each test is: given these inputs, expect this status. that's a table, not a list of test functions 🚢

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -775,31 +814,48 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
-        if (config?.triggerMatchType === 'any') {
-            this._statusMatcher = anyMatchSessionRecordingStatus
-            this._triggerMatching = new OrTriggerMatching([this._eventTriggerMatching, this._urlTriggerMatching])
+        // Check if this is V2 trigger groups config
+        const isV2 = config?.version === 2 && config?.triggerGroups && config.triggerGroups.length > 0
+
+        if (isV2) {
+            // V2: Setup trigger group matchers
+            this._statusMatcher = triggerGroupsMatchSessionRecordingStatus
+            this._setupTriggerGroups(config.triggerGroups!)
```
pauldambra: 🤖 _from c2 wiki / XP lens:_

the V1/V2 branching in `_onRemoteConfig`, `status`, `_minimumDuration`, `_flushBuffer`, and URL trigger checking is a Shotgun Surgery smell — one concept (V2 trigger groups) is smeared across many methods.

ComposedMethod / Strategy pattern suggests: if every method asks "am I V1 or V2?", you likely have two strategies wanting to be separate objects. extracting V2 behavior into something like a `TriggerGroupRecordingStrategy` that owns the matchers, sampling results, flush flag, and min-duration logic would mean the V1/V2 branch exists in exactly one place. `LazyLoadedSessionRecording` is already a Large Class — every future V2 feature will add another if/else to every method.

not a "you must" — it works. but worth considering before V3 arrives 😄

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -4142,4 +4142,405 @@ describe('Lazy SessionRecording', () => {
...
+                })
+            )
+
+            expect(sessionRecording.status).toBe('buffering')
+
+            // Trigger the event
+            simpleEventEmitter.emit('eventCaptured', { event: 'test_event' })
+
+            // Should be ACTIVE (matched but sampled out), not SAMPLED
+            expect(sessionRecording.status).toBe('active')
```
TueHaulund: Hmm, why is the session recording status ACTIVE if the session was not sampled?

packages/browser/src/extensions/replay/external/triggerMatching.ts
```diff
@@ -406,6 +449,101 @@ export class EventTriggerMatching implements TriggerStatusMatching {
...
+                urlTriggers: group.conditions.urls || [],
+                eventTriggers: group.conditions.events || [],
+                linkedFlag: group.conditions.flag || null,
+                urlBlocklist: [], // groups don't have blocklist
+            }
+
+            // onConfig methods only access the properties we're providing, so safe to cast
+            this._urlTriggerMatching.onConfig(config as ReplayConfigType)
+            this._eventTriggerMatching.onConfig(config as ReplayConfigType)
+            this._linkedFlagMatching.onConfig(config as ReplayConfigType, onFlagStarted)
```
TueHaulund: I think the type is wrong here - shouldn't it be `as TriggerMatchingConfig` in all 3 occurrences?

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1014,12 +1079,30 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         }
 
         // Check if the URL matches any trigger patterns
+        // V1: Check global URL triggers
         this._urlTriggerMatching.checkUrlTriggerConditions(
             () => this._pauseRecording(),
             () => this._resumeRecording(),
             (triggerType, matchDetail) => this._activateTrigger(triggerType, matchDetail),
             this.sessionId
         )
```
TueHaulund, line 1044: So does this mean we run _both_ the V1 and V2 triggers if V2 is enabled? Is that intentional?

packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -0,0 +1,557 @@
...
+    }
+
+    updateActiveTriggers(): void {
+        const recordingGroups: Array<{ id: string; name: string; matched: boolean; sampled: boolean }> = []
+
+        for (const matcher of this._triggerGroupMatchers) {
+            const group = matcher.group
+            const groupId = group.id
+            const triggerStatus = matcher.triggerStatus(
+                this._instance.sessionManager!.checkAndGetSessionAndWindowId().sessionId
```
TueHaulund: Instead of the non-null assertion on the sessionManager here you could just pass sessionId as a parameter or get it from the strategy context

packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -0,0 +1,557 @@
...
+
+        this._linkedFlagMatching.onConfig(config, (flag, variant) => {
+            this._reportStarted('linked_flag_matched', { flag, variant })
+        })
+    }
+
+    getStatus(context: RecordingStrategyContext): SessionRecordingStatus {
+        const config = context.remoteConfig
+        const matchFunction =
+            config?.triggerMatchType === 'any' ? anyMatchSessionRecordingStatus : allMatchSessionRecordingStatus
```
TueHaulund: I think we should determine the matchFunction once and just store it instead of re-determining it on every call. If remoteConfig changes for whatever reason things could go wrong if the matchFunction changes as a result.

PR #3285, opened 2026-03-26
fix(identify): reject string "undefined" and "null" in identify() and be more clear when rejecting

pauldambra:
## Summary

- When someone does `` posthog.identify(`${undefinedVar}`) ``, JS coerces `undefined` to the string `"undefined"`, which passes all existing validation and silently identifies the user as literal `"undefined"`
- `_validateIdentifyId` now rejects `"undefined"`, `"null"`, and whitespace-only strings in addition to the existing checks
- All invalid distinct ID rejections now use `logger.critical` (always visible in console) instead of `logger.error` (debug-only)
- No config changes, no defaults gating — this is a logging-only change, not a behavioural one. We already rejected these values, we just now also reject the string coerced versions

## Test plan

- [x] Parameterized test confirms `undefined`, `null`, `""`, `"   "`, `false`, `"undefined"`, `"null"` are all rejected with critical log
- [x] All existing identify tests pass
- [x] Build passes

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -2290,6 +2290,26 @@ export class PostHog implements PostHogInterface {
         )
     }
 
+    private _validateIdentifyId(id: string | undefined): id is string {
+        if (!id || isEmptyString(id)) {
+            logger.critical('Unique user id has not been set in posthog.identify')
+            return false
+        }
+        if (id === COOKIELESS_SENTINEL_VALUE) {
+            logger.critical(
+                `The string "${id}" was set in posthog.identify which indicates an error. This ID is only used as a sentinel value.`
+            )
+            return false
+        }
+        if (isDistinctIdStringLike(id) || ['undefined', 'null'].includes(id.toLowerCase())) {
+            logger.critical(
+                `The string "${id}" was set in posthog.identify which indicates an error. This ID should be unique to the user and not a hardcoded string.`
+            )
+            return false
+        }
+        return true
+    }
+
     /**
      * Associates a user with a unique identifier instead of an auto-generated ID.
      * Learn more about [identifying users](/docs/product-analytics/identify)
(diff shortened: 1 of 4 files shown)


packages/types/src/posthog-config.ts
```diff
@@ -1016,11 +1014,21 @@ export interface PostHogConfig {
...
+     * - `'2026-04-01'`: Defaults from '2026-01-30' plus strict_identify defaults to true (throws on invalid distinct IDs)
      *
      * @default 'unset'
      */
     defaults: ConfigDefaults
 
+    /**
+     * When true, posthog.identify() throws an error if called with an invalid distinct_id
+     * (e.g. undefined, null, empty string, or the string "undefined").
+     * When false, invalid calls log a console error and return without identifying.
```
marandaneto: i think SDKs should never throw, swallow users input and log is the best you can do to be a good citizen
throwing might 'break' people's apps, which is worse than just not getting that user identified.

marandaneto: > We never want our SDKs to be the cause of customer breakage

It's a SDK philosophy that i learned and agree with

https://github.com/PostHog/posthog-js/pull/3261 @lricoy added similar pattern which i dont think is the best approach
Curious about what other @PostHog/team-client-libraries folks think

pauldambra: no, fair... i was being overly keen

packages/browser/src/posthog-core.ts
```diff
@@ -2286,10 +2276,23 @@ export class PostHog implements PostHogInterface {
...
             this.surveys?.canRenderSurveyAsync(surveyId, forceReload) ??
             // eslint-disable-next-line compat/compat
-            Promise.resolve({ visible: false, disabledReason: SURVEYS_NOT_AVAILABLE })
+            Promise.resolve({ visible: false, disabledReason: 'Surveys module not available' })
         )
     }
 
+    private _validateIdentifyId(id: string | undefined): string | undefined {
+        if (!id || !id.trim()) {
+            return 'Unique user id has not been set in posthog.identify'
```
marandaneto: lets extract 'Unique user id has not been set in posthog.identify to a const since its used 2 times, saves a few bytes

packages/browser/src/posthog-core.ts
```diff
@@ -2342,22 +2345,9 @@ export class PostHog implements PostHogInterface {
...
-                `The string "${new_distinct_id}" was set in posthog.identify which indicates an error. This ID should be unique to the user and not a hardcoded string.`
-            )
-            return
-        }
-        if (new_distinct_id === COOKIELESS_SENTINEL_VALUE) {
-            logger.critical(
-                `The string "${COOKIELESS_SENTINEL_VALUE}" was set in posthog.identify which indicates an error. This ID is only used as a sentinel value.`
-            )
+        const identifyError = this._validateIdentifyId(new_distinct_id)
+        if (identifyError || !new_distinct_id) {
```
marandaneto:
```suggestion
        if (identifyError) {
```

we already check this in the _validateIdentifyId method

packages/browser/src/posthog-core.ts
```diff
@@ -2331,6 +2331,26 @@ export class PostHog implements PostHogInterface {
      *  it will be overwritten by the value in userPropertiesToSet.
      * @param {Object} [userPropertiesToSetOnce] Optional: An associative array of properties to store about the user. If property is previously set, this does not override that value.
      */
+    private _validateIdentifyId(id: string | undefined): id is string {
```
dustinbyrne: This method was placed in between `identify` and its docstring. We should move it.

pauldambra: ooooof :shame-conga:

pauldambra: baby-circle

PR #3286, opened 2026-03-26
perf(browser): use async CompressionStream for gzip to avoid blocking main thread

marandaneto:
## Problem

Users running PostHog in performance-constrained environments (e.g., real-time browser games in iframes) experience significant frame drops caused by `gzipSync` from fflate blocking the main thread for **300ms+** during request compression. This is especially impactful with session recordings + canvas recording, where compressed payloads are sent frequently.

Related: https://posthoghelp.zendesk.com/agent/tickets/43990 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/55566)
but also https://github.com/PostHog/posthog-js/pull/2823 and some other PRs

## Changes

Uses the native [CompressionStream API](https://developer.mozilla.org/en-US/docs/Web/API/CompressionStream) for gzip compression when available on `fetch`/`XHR` transports. This is async and non-blocking, handled by the browser's built-in compression engine.

**How it works:**
- The `request()` entry point checks if `CompressionStream` is available (safely imported from `globals.ts`)
- For `fetch`/`XHR` with gzip compression, it pre-encodes the body asynchronously via `gzipCompress` from `@posthog/core`, storing the result in `_encodedBody`
- `encodePostData()` checks for `_encodedBody` first, skipping the synchronous fflate path entirely
- If async compression fails, falls back to the existing synchronous fflate path via `.catch()`
- `sendBeacon` always uses the synchronous fflate path (required — runs during page unload)

**Fallback behavior:**
- Browsers **with** `CompressionStream` (Chrome 80+, Firefox 113+, Safari 16.4+): async native compression ✅
- Browsers **without** `CompressionStream`: existing `gzipSync` from fflate (unchanged behavior)
- `sendBeacon` transport: always uses sync fflate (required for `beforeunload`)

**Files changed:**
- `packages/browser/src/request.ts` — async pre-encoding with `CompressionStream` via `gzipCompress` from `@posthog/core`, fallback to sync fflate
- `packages/browser/src/utils/globals.ts` — export `CompressionStream` with safe `global?.` access pattern
- `packages/core/src/index.ts` — export `gzipCompress` for reuse

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/core

## Checklist

### If releasing new changes

### DIFF
--- a/.changeset/async-gzip-compression.md
+++ b/.changeset/async-gzip-compression.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/core': patch
+---
+
+Use async native CompressionStream for gzip compression to avoid blocking the main thread
--- a/packages/browser/src/request.ts
+++ b/packages/browser/src/request.ts
@@ -8,6 +8,11 @@ import { AbortController, fetch, navigator, XMLHttpRequest } from './utils/globa
 import { gzipSync, strToU8 } from 'fflate'
 
 import { _base64Encode } from './utils/encode-utils'
+import { isGzipSupported } from '@posthog/core'
+
+interface RequestWithEncodedBody extends RequestWithOptions {
+    _encodedBody?: EncodedBody
+}
 
 // eslint-disable-next-line compat/compat
 export const SUPPORTS_REQUEST = !!XMLHttpRequest || !!fetch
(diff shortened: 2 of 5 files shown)


dustinbyrne (approved):
Nice, [96.19%](https://caniuse.com/?search=CompressionStream) availability on the `CompressionStream` API too. With this, we should pull `fflate` out of the bundle and dynamically load it only if necessary. Most users will no longer need it.

PR #3295, opened 2026-03-28
feat: add disableFeatureFlagReload option to identify()

dmarticus:
## Summary

- Adds a `disableFeatureFlagReload` option to `identify()` across browser, core, and React Native SDKs
- When `identify()` changes the distinct_id, the SDK unconditionally reloads feature flags. The `/flags` request fires before the person merge triggered by `$identify` has propagated server-side, which can cause the server to re-bucket the user under a different variant. This option lets callers suppress the automatic reload and let flags refresh on the next session when identities are fully merged.
- New `IdentifyOptions` type in `@posthog/types` and `PostHogIdentifyOptions` in the core package

### Usage

```typescript
// Browser SDK
posthog.identify('authenticated-user-id', userProperties, undefined, {
  disableFeatureFlagReload: true,
})

// React Native SDK
posthog.identify('authenticated-user-id', userProperties, {
  disableFeatureFlagReload: true,
})
```

## Test plan

- [x] Added browser SDK tests for `disableFeatureFlagReload: true` and `false`
- [x] Added core SDK tests verifying no `/flags` request when option is set
- [x] All existing identify tests continue to pass
- [x] React Native tests pass (inherits core behavior via `super.identify()`)
- [x] TypeScript compiles cleanly across all three packages

### DIFF
--- a/packages/browser/src/__tests__/posthog-core.identify.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.identify.test.ts
@@ -305,6 +305,21 @@ describe('identify()', () => {
 
             expect(instance.unregister).toHaveBeenCalledWith('$flag_call_reported')
         })
+
+        it('does not reload feature flags when disableFeatureFlagReload is true', () => {
+            instance.identify('a-new-id', undefined, undefined, { disableFeatureFlagReload: true })
+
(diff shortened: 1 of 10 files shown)


marandaneto:
On my phone now and didnt review the code yet but i remember some sdks cache person props during the identify call and reload flags with all new/cached props so the reload flags should be evaluated correctly already.

PR #3296, opened 2026-03-30
fix(core): fix captureException ReferenceError in React Native for missing Event global

### LINKED ISSUE #3294: [React Native] - Error tracking isn't working out of the box
## Bug description

While using the `posthog-react-native` library, trying to track exceptions using the `captureException` or using the `autocapture` flag results in an error, and exceptions aren't being tracked and are not shown in the dashboard. I'm triggering a simple error to double-check to see them in the dashboard. 

I'm using library version: `^4.37.6`.

## How to reproduce

1.   Follow the installation guide and setup for exceptions at: [PostHog React Native Exceptions Tracking](https://posthog.com/docs/error-tracking/installation/react-native)
2. Enable `postHog.debug(true)` to review logs
3.  Trigger a test error in any part of the application 
4. The exception isn't showing in the dashboard
## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [X] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

Here I add a simple code for testing this out:
```
  try {
      throw new Error('Test handled error for PostHog')
    } catch (error) {
      postHog.captureException(error)
    }
```
Image

What I'm seeing is an error because one of these checks inside the library does `error instanceof Event`, and because Event doesn't exist globally in React Native as it does in a browser. 

Hope this helps to debug the error

marandaneto:
## Problem

Calling `captureException` in React Native crashes with:

> ReferenceError: Property 'Event' doesn't exist

This affects all React Native users using error tracking. The root cause is that `isEvent()` in `@posthog/core` references the `Event` global directly as a value (`!isUndefined(Event)`), but `Event` is a Web API that does not exist in React Native's Hermes/JSC JavaScript engine. Unlike `typeof Event` which safely returns `"undefined"`, passing `Event` as a function argument throws a `ReferenceError` before the function even executes.

Closes #3294

### Bug fix
- **`packages/core/src/utils/type-utils.ts`**: Changed `!isUndefined(Event)` to `typeof Event !== 'undefined'` in `isEvent()`. The `typeof` operator safely checks for undeclared globals without throwing.

### Prevention — new ESLint rule
- **`tooling/eslint-plugin-posthog-js/no-unsafe-web-global.js`**: New `posthog-js/no-unsafe-web-global` ESLint rule that flags direct value-level references to 30+ Web API globals (`Event`, `ErrorEvent`, `MutationObserver`, `XMLHttpRequest`, etc.) that don't exist in React Native.
  - **Catches**: `const x = Event`, `new Event()`, `candidate instanceof Event`, `isUndefined(Event)`
  - **Allows**: `typeof Event !== 'undefined'`, guarded patterns (`typeof Event !== 'undefined' && isInstanceOf(candidate, Event)`), type annotations (`: Event`, `is Event`)
- **`.eslintrc.cjs`**: Rule is off by default, enabled only for `packages/core/src/**` (excluding test files) — the shared package that must work in both browser and React Native.
- **Tests added** in `custom-eslint-rules.test.js` covering valid and invalid patterns.

### Libraries affected

- [x] @posthog/core (patch) — fix for the `isEvent()` crash
- [x] posthog-react-native (patch) — picks up the core fix

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/tooling/eslint-plugin-posthog-js/no-unsafe-web-global.js
+++ b/tooling/eslint-plugin-posthog-js/no-unsafe-web-global.js
@@ -145,10 +145,6 @@ module.exports = {
                     return
                 }
 
-                // Allow: as expressions (x as ErrorEvent) — the identifier is in a type position
-                if (parent.type === 'TSAsExpression' && parent.typeAnnotation === node) {
-                    return
-                }
 
                 // Allow: import specifiers
                 if (


tooling/eslint-plugin-posthog-js/custom-eslint-rules.test.js
```diff
@@ -132,6 +133,61 @@ ruleTester.run('no-direct-boolean-check', noDirectBooleanCheck, {
...
 
+const tsRuleTester = new RuleTester({
+    parser: require.resolve('@typescript-eslint/parser'),
+    parserOptions: {
+        ecmaVersion: 2015,
+        sourceType: 'module',
+    },
+})
+
+tsRuleTester.run('no-unsafe-web-global', noUnsafeWebGlobal, {
```
hpouillot, line 144: 🔥

PR #3297, opened 2026-03-30
fix(browser): wrap sendBeacon body in Blob to ensure Content-Type header is set

### LINKED ISSUE #3262: POST object expects Content-Type multipart/form-data
## Bug description

Getting the above error in prod but it's working fine on dev. This is happening with ingest/flags/?v=2, /ingest/e/?ip=0&_=1773988066866&ver=1.360.0&compression=gzip-js and all other posthog related APIs
## How to reproduce

1. This is happening with 1.360.0 version in vite app

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

marandaneto:
## Problem

Fixes #3262. Users with reverse proxies, WAFs, or CDNs in production that enforce Content-Type headers were getting "POST object expects Content-Type multipart/form-data" errors on all PostHog API requests sent via `sendBeacon`.

Since commit 2e46959e (PR #3172, "Send gzipped request body as ArrayBuffer"), gzip-compressed bodies changed from `Blob` to `ArrayBuffer`. While this fixed a WebKit issue with `fetch`, it inadvertently broke the `sendBeacon` transport:

- `sendBeacon(url, Blob)` → browser uses the Blob's `type` as `Content-Type` ✅
- `sendBeacon(url, ArrayBuffer)` → browser sends **no Content-Type header** ❌

This explains why the issue only manifests in production (where proxies enforce headers) but not in development.

## Changes

- **`packages/browser/src/request.ts`**: In `_sendBeacon()`, always wrap the body in a `Blob` with the correct `contentType`. Previously only string bodies were wrapped; now ArrayBuffer bodies are also wrapped, ensuring `Content-Type: text/plain` is set for gzip-compressed requests. Also added an early return guard when `body` is undefined to avoid creating a Blob with literal `"undefined"` content.
- **`packages/browser/src/__tests__/request.test.ts`**: Updated the sendBeacon gzip test to expect a `Blob` (with `type: 'text/plain'`) instead of a raw `ArrayBuffer`.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/request.ts
+++ b/packages/browser/src/request.ts
@@ -255,10 +255,13 @@ const _sendBeacon = (options: RequestWithOptions) => {
 
     try {
         const { contentType, body } = encodePostData(options) ?? {}
+        if (!body) {
+            return
+        }
         // sendBeacon requires a Blob to set the Content-Type header correctly.
         // Without wrapping, ArrayBuffer bodies are sent with no Content-Type,
         // which can cause issues with proxies/WAFs that require it.
-        const sendBeaconBody = body instanceof Blob ? body : new Blob([body as BlobPart], { type: contentType })
+        const sendBeaconBody = body instanceof Blob ? body : new Blob([body], { type: contentType })
         navigator!.sendBeacon!(url, sendBeaconBody)
     } catch {
         // send beacon is a best-effort, fire-and-forget mechanism on page unload,


packages/browser/src/request.ts
```diff
@@ -255,8 +255,13 @@ const _sendBeacon = (options: RequestWithOptions) => {
...
         const { contentType, body } = encodePostData(options) ?? {}
-        // sendBeacon requires a blob so we convert it
-        const sendBeaconBody = typeof body === 'string' ? new Blob([body], { type: contentType }) : body
+        if (!body) {
+            return
+        }
+        // sendBeacon requires a Blob to set the Content-Type header correctly.
+        // Without wrapping, ArrayBuffer bodies are sent with no Content-Type,
+        // which can cause issues with proxies/WAFs that require it.
+        const sendBeaconBody = body instanceof Blob ? body : new Blob([body], { type: contentType })
```
marandaneto, line 264: does not interfer with https://github.com/PostHog/posthog-js/pull/3172 cus this is sendBeacon transport only

PR #3302, opened 2026-03-30
fix(flags): preserve $set_once semantics in local flag evaluation cache

dmarticus:
## Summary

- **Bug**: When `identify()` was called with `$set_once` properties, the local cache used for feature flag evaluation (`/flags`) always overwrote existing values, ignoring `$set_once` "skip if exists" semantics. Customers calling `identify()` on every app open with `$set_once` values had the cached value overwritten to today's value, which then overrode the correct (older) database value during flag evaluation via `/flags`.
- **Fix**: `setPersonPropertiesForFlags()` now accepts a `propertiesSetOnce` parameter that only writes keys not already present in the cache, mirroring server-side `$set_once` behavior. Applied across **browser**, **core**, **react-native**, and **types** packages.

## Detail

The root cause was in three places:

1. `posthog-core.ts` — Both `identify()` and `setPersonProperties()` merged `$set_once` and `$set` into a single flat object via spread: `{ ...setOnce, ...set }`, losing all `$set_once` semantics before passing to the flag cache.

2. `posthog-featureflags.ts` (browser) / `posthog-core.ts` (core) — `setPersonPropertiesForFlags()` always overwrote, with no option to skip existing keys.

3. The `/flags` request sends these cached properties as `person_properties`, which the server uses as overrides for flag evaluation — so the wrong (latest) cached value took precedence over the correct database value.

### Example scenario (before fix)
```js
// Day 1
posthog.identify('user1', {}, { $set_once: { initial_plan: 'free' } })
// Cache: { initial_plan: 'free' } ✓
// Server $set_once: 'free' ✓

// Day 90
posthog.identify('user1', {}, { $set_once: { initial_plan: 'pro' } })
// Server $set_once: still 'free' ✓ (correctly skipped)
// Cache: 'pro' ✗ (OVERWRITTEN — sent to /flags, overrides DB value)
```

### Packages changed

| Package | Change |
|---------|--------|
| `posthog-js` (browser) | `setPersonPropertiesForFlags()` and `identify()` / `setPersonProperties()` pass `$set_once` separately |
| `@posthog/core` | Same fix in the shared core implementation |
| `posthog-react-native` | `identify()` extracts `$set_once` before `super.identify()` (which deletes it from the object) and passes it through |
| `@posthog/types` | `PostHogInterface.setPersonPropertiesForFlags` signature updated with optional `propertiesSetOnce` |

### RN subtlety

Core's `identify()` does `delete properties?.$set_once` before processing. The RN override calls `super.identify()` first, so it must extract `$set_once` *before* the super call or the value is lost.

## Test plan

- [x] Browser: `$set_once` properties skip existing keys in cache
- [x] Browser: `$set` properties still overwrite even when `$set_once` would not
- [x] Browser: `$set_once` properties included in `/flags` request payload
- [x] Browser: all 3523 unit tests passing
- [x] Core: `$set_once` properties skip existing keys in cache
- [x] Core: `$set` overrides `$set_once` for same key
- [x] Core: all 381 tests passing
- [x] RN: `$set_once` cached with set-once semantics from `identify()`
- [x] RN: subsequent `identify()` does not overwrite existing `$set_once` keys
- [x] RN: all 60 tests passing

### DIFF
--- a/.changeset/strong-wasps-hide.md
+++ b/.changeset/strong-wasps-hide.md
@@ -1,8 +1,0 @@
----
-'posthog-js': patch
-'posthog-react-native': patch
-'@posthog/core': patch
-'@posthog/types': patch
----
-
-fix(flags): preserve $set_once semantics in local flag evaluation cache


packages/browser/src/posthog-core.ts
```diff
@@ -2606,8 +2603,12 @@ export class PostHog implements PostHogInterface {
      * @param {Object} properties The properties to override.
      * @param {Boolean} [reloadFeatureFlags] Whether to reload feature flags.
      */
-    setPersonPropertiesForFlags(properties: Properties, reloadFeatureFlags = true): void {
-        this.featureFlags?.setPersonPropertiesForFlags(properties, reloadFeatureFlags)
+    setPersonPropertiesForFlags(
```
marandaneto: the same applies to node, js-lite, rn, should we do the change in this PR?

marandaneto: most SDKs only get the properties
eg https://github.com/PostHog/posthog-ios/blob/4d35bc645cc8e8aa1d651048a2b2e4e386483487/PostHog/PostHogSDK.swift#L1109-L1133
and now we have properties and propertiesSetOnce
should we instead do something similar to the identify method (core package)?

```
      const userPropsOnce = properties?.$set_once
      delete properties?.$set_once

      // if no $set is provided we assume all properties are $set
      const userProps = properties?.$set || properties
```

and then we tell people to pass $set and $set_once if they want both?

the other options is to deprecate properties and add userPropertiesToSet, and userPropertiesToSetOnce

i think properties and propertiesSetOnce is a bit confusing since its all about the user props, isnt it?

dmarticus: > the same applies to node, js-lite, rn, should we do the change in this PR?

Good call — since they are all in this repo I will do the change across node, js-lite, and rn in this PR.

dmarticus: > most SDKs only get the properties eg https://github.com/PostHog/posthog-ios/blob/4d35bc645cc8e8aa1d651048a2b2e4e386483487/PostHog/PostHogSDK.swift#L1109-L1133 and now we have properties and propertiesSetOnce should we instead do something similar to the identify method (core package)?
> 
> ```
>       const userPropsOnce = properties?.$set_once
>       delete properties?.$set_once
> 
>       // if no $set is provided we assume all properties are $set
>       const userProps = properties?.$set || properties
> ```
> 
> and then we tell people to pass $set and $set_once if they want both?
> 
> the other options is to deprecate properties and add userPropertiesToSet, and userPropertiesToSetOnce
> 
> i think properties and propertiesSetOnce is a bit confusing since its all about the user props, isnt it?



Agreed, the separate `propertiesSetOnce` param is awkward. Going with Option A — keep a single `properties` arg on the public API and parse `$set` / `$set_once` keys internally (like `identify()` already does). No breaking change, and it is consistent with how users already pass these keys elsewhere.

Option B (deprecate + add two explicit params) is cleaner but would be a breaking change to a public method, so not worth it here.

dustinbyrne: I'm still seeing `propertiesSetOnce` params. Do you still plan on dropping it?

dmarticus: that's my bad – forgot to commit!

dmarticus: @dustinbyrne you mind taking another look?

.changeset/strong-wasps-hide.md
```diff
@@ -0,0 +1,8 @@
+---
+'posthog-js': patch
+'posthog-react-native': patch
+'@posthog/core': patch
+'@posthog/types': patch
```
dustinbyrne: Types aren't changing any more. We should drop this before merging.

dmarticus: oh yeah, gotta regenerate the changeset.

PR #3307, opened 2026-03-31
chore: upload posthog-js dist to S3 on release

dustinbyrne:
## Problem

Customers will be able to pin their posthog-js version (e.g. `"1"`). When a request hits `us-assets.i.posthog.com/array/{team_token}/array.js`, Django resolves the configured version and serves the JS artifact from S3. The infra side (buckets + OIDC roles) is in [posthog-cloud-infra#7255](https://github.com/PostHog/posthog-cloud-infra/pull/7255) — this PR adds the write path from the release workflow.

## Changes

**New file: `.github/scripts/upload-posthog-js-s3.sh`**
Reusable script that:
1. Uploads `packages/browser/dist/*.js` to `s3://{bucket}/{version}/` with immutable cache headers
2. Downloads the current `versions.json` manifest (or starts with `[]`)
3. Appends `{"version": "x.y.z", "timestamp": "..."}` if not already present
4. Uploads the updated `versions.json`

Safety measures:
- Semver regex validation on `VERSION` — blocks path traversal and shell metacharacters
- `VERSION` read from env var (not a shell argument) to avoid injection via `${{ }}` interpolation
- Distinguishes "versions.json doesn't exist" from real S3 errors to prevent silently dropping previous entries
- Uses `mktemp -d` for temp files

**Modified: `.github/workflows/release.yml`**
New `upload-s3` job that runs after `publish`:
- Uses `S3 Upload` GitHub environment (matches OIDC subject from infra PR)
- Checks if posthog-js was released via `check-package-version`
- Configures OIDC credentials for US (`us-east-1`) then EU (`eu-central-1`)
- Calls the upload script for each region
- Posts to Slack thread on failure
- `notify-released` now also waits on `upload-s3`

The Celery task `sync_js_snippet_manifest` (runs every 5 min) will pick up the updated `versions.json`, compute pointer maps, validate artifacts, update Redis, and purge CDN.

### Setup needed before first use
I'll need help with this:
1. Create **"S3 Upload"** environment in GitHub repo settings _without_ required reviewers
	- The S3 release is still gated by the NPM release required reviewers. Adding additional required reviewers would add another approval step to the same process.
3. Add environment variables:
   - `AWS_S3_UPLOAD_ROLE_ARN_US` — ARN of the US upload role
   - `AWS_S3_UPLOAD_ROLE_ARN_EU` — ARN of the EU upload role

### Libraries affected

- [ ] All of them

## Checklist

### DIFF
--- a/.github/scripts/upload-posthog-js-s3.sh
+++ b/.github/scripts/upload-posthog-js-s3.sh
@@ -58,9 +58,10 @@ else
         '. + [{"version": $v, "timestamp": $ts}]' "$TMPWORKDIR/versions.json" > "$TMPWORKDIR/versions_updated.json"
 
     # Validate the updated manifest before uploading: must be a non-empty JSON array
-    # where every entry has .version and .timestamp strings.
-    if ! jq -e 'if type != "array" then error
-        elif length == 0 then error
+    # where every entry has .version and .timestamp strings, and length is exactly original + 1.
+    EXPECTED_LENGTH=$(( $(jq 'length' "$TMPWORKDIR/versions.json") + 1 ))
+    if ! jq -e --argjson expected "$EXPECTED_LENGTH" 'if type != "array" then error
+        elif length != $expected then error
         elif any(.[]; (.version | type) != "string" or (.timestamp | type) != "string") then error
         else true end' "$TMPWORKDIR/versions_updated.json" > /dev/null 2>&1; then
         echo "ERROR: versions_updated.json failed validation — aborting upload" >&2


.github/workflows/release.yml
```diff
@@ -334,11 +334,107 @@ jobs:
...
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v6
+              with:
+                  ref: ${{ needs.version-bump.outputs.commit-hash }}
+                  fetch-depth: 0
+
+            - name: Check posthog-js version
+              id: check-version
+              uses: PostHog/check-package-version@v2.1.0
```
marandaneto:
```suggestion
              uses: PostHog/check-package-version@v2
```

.github/scripts/upload-posthog-js-s3.sh
```diff
@@ -0,0 +1,62 @@
...
+    echo '[]' > "$TMPDIR/versions.json"
+fi
+
+if jq -e --arg v "$VERSION" '.[] | select(.version == $v)' "$TMPDIR/versions.json" > /dev/null 2>&1; then
+    echo "Version $VERSION already in versions.json, skipping"
+else
+    jq --arg v "$VERSION" --arg ts "$(date -u +%Y-%m-%dT%H:%M:%SZ)" \
+        '. + [{"version": $v, "timestamp": $ts}]' "$TMPDIR/versions.json" > "$TMPDIR/versions_updated.json"
+    aws s3 cp "$TMPDIR/versions_updated.json" "s3://$BUCKET/versions.json" \
+        --content-type "application/json"
```
Piccirello, line 73: This seems like the riskiest bit to me, since one bad change will break the entire file. I'm assuming using jq to modify the file means it will only ever generate valid json. Is there any additional check we can run here to verify the contents match what we expect before publishing?

.github/workflows/release.yml
```diff
@@ -334,11 +334,107 @@ jobs:
                   message: '❌ Failed to release `${{ matrix.package.name }}@v${{ steps.check-package-version.outputs.committed-version }}`! <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|View logs>'
                   emoji_reaction: 'x'
 
+    build-s3-artifacts:
```
Piccirello, line 337: The aws script is racey, so verified this workflow already limits concurrency to 1 👍

.github/workflows/release.yml
```diff
@@ -334,11 +334,107 @@ jobs:
...
+            contents: read
+        outputs:
+            is-new-version: ${{ steps.check-version.outputs.is-new-version }}
+            committed-version: ${{ steps.check-version.outputs.committed-version }}
+        steps:
+            - name: Checkout repository
+              uses: actions/checkout@v6
+              with:
+                  ref: ${{ needs.version-bump.outputs.commit-hash }}
+                  fetch-depth: 0
```
Piccirello: Is the full repo history actually needed or can we use a fetch-depth of 1?

.github/scripts/upload-posthog-js-s3.sh
dustinbyrne: I've added some additional verification's using `jq`:
- It's a JSON array                                                                                                                                                                                                
- The length of the array is exactly original + 1                                                                                                                                 
- Every entry is of the expected schema

This is a good point though. Breaking this file would mean breaking version synchronization. I'll make a note to add some server-side observability for this case.

.github/workflows/release.yml
dustinbyrne: Good catch, changed to 1.

Issue #2623, opened 2025-11-24
Feature flag value changes after identify, `Flag persistence` is enabled


O4epegb:
## Bug description

Feature flag value changes after identify, `Flag persistence` is enabled

## How to reproduce

Demo: https://stackblitz.com/edit/vitejs-vite-61subwxi?file=index.html

Steps:

- visit the page, it will display in what branch current anon user is.
- click `identify` to identify as `test user` and to see in what branch you are - `control` or `show`
- you might see the issue right away if there is a mismatch, if not - continue with next step
- If you didn't encounter the issue - now open new incognito page or press `reset` button until initial anonymous user gets into a different brach (e.g. if the `test user` is in `control` then reset until you initial user lands into `show`)
- press identify
- you will see that feature flag has changed, even though it's a persistent flag with `person_profiles: 'always'` etc

Quote from docs:

```
Feature flag values are calculated based on a user's properties. Since it's possible for a user to have different properties before and after login, they may receive different feature flag values before and after logging in.

By enabling the option to persist feature flags across authentication, you ensure that the flag value remains the same.
```

As far as I understand, this situation should not happen, at least with fresh session, if the fresh anon user saw `show` it should not change to `control` after identify. Otherwise I don't see a point of persistence flags at all, if they don't work in this case then I don't quite understand when they work at all, maybe I'm still missing some crucial info though 😅 

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

Project info

```
Session: https://us.posthog.com/project/sTMFPsFhdP1Ssg/replay/0197201d-88ae-76a0-9c20-7acc32fb6a27?t=2598
Admin: http://go/adminOrgEU/0197201e-8a1e-0000-1b08-3d3706945667 (project ID 69186)
```

justinluttrell:
I'm able to reproduce the same issue, despite enabling "Flag persistence" and setting `person_profiles: 'always'`. The feature flags still change after `identify()` is called.

[...]

Issue #1756, opened 2025-02-21
[BUG][FEATURE FLAGS] Feature flag changes value even if "Flag persistence" enabled


Garkavenko:
#### Issue Summary
I am experiencing unexpected behavior with feature flags and experiments in PostHog. Even though my feature flag persists across authentication events, its value changes after calling `posthog.identify`.

#### Steps to Reproduce
I have created a simple repository to reproduce the issue: [posthog-feature-flags-bug-reproduce](https://github.com/Garkavenko/posthog-feature-flags-bug-reproduce).

To run the reproduction:
```sh
node index.js
```

The setup is a basic Express server that serves an HTML page. The logic is as follows:
- Initialize the PostHog instance.
- Subscribe to `onFeatureFlags` events.
- Render all feature flag values on the screen.
- Provide an "Identify user" button that triggers `posthog.identify` on click.

#### Expected Behavior
When opening the site in an incognito window, waiting for the first flag value to render, and then clicking "Identify user," the feature flag should remain consistent across events.

#### Observed Behavior
Once in approximately 7–10 attempts, the flag value changes between "control" and "test" after calling `posthog.identify`.

#### Steps to Catch the Bug
1. Modify the `id` and `email` inside the HTML to be unique.
2. Open a new **incognito session** (not just a new tab).
3. Wait for the first `onFeatureFlags` event (flag value renders on the screen).
4. Click on the "Identify user" button.
5. Observe if the flag value changes.

#### Screenshots
**Before `identify` Call:** 
![Image](https://github.com/user-attachments/assets/7636bf90-8b57-491e-98f9-a2dfa9aae177)
**After `identify` Call:**  
![Image](https://github.com/user-attachments/assets/6a17d7e8-14cb-4c9b-92ec-e2cd060140d5)

**PostHog Account View:**  
![Image](https://github.com/user-attachments/assets/935dde5e-1c01-4bfd-b9e0-c2fa2991b050)
![Image](https://github.com/user-attachments/assets/302f7c5c-46e8-4928-969c-8c9057088780)

 #### Additional Information
This inconsistency affects experiment allocation and may lead to unpredictable user experiences. Any insights into why `posthog.identify` affects the feature flag value would be helpful. Let me know if more details are needed!
Project ID: 23437

Garkavenko:
Can anyone help with that?

haacked:
[...]

PR #3340, opened 2026-04-06
feat(react-native): add device bucketing support

dmarticus:
## Summary

- Adds `$device_id` support to the React Native SDK so feature flags can bucket on a stable device identifier instead of `distinct_id`
- The device ID persists across `identify()` and `reset()` calls, preventing variant flipping during identity transitions (anonymous → authenticated, logout → re-login)
- Sends `$device_id` in `/flags` request payload, matching the browser SDK's existing behavior

## Problem

Users running experiments targeting anonymous users see variant assignments flip when `identify()` or `reset()` changes the `distinct_id`. The hash changes → different bucket → different variant for the same physical device. This was reported by multiple customers (Going app team) and tracked in #2623 and #1756.

## Changes

- **`packages/core/src/types.ts`**: Add `DeviceId` to `PostHogPersistedProperty` enum
- **`packages/core/src/posthog-core.ts`**: Include `$device_id` in `/flags` request `extraProperties` (no-op for SDKs that don't set it)
- **`packages/react-native/src/posthog-rn.ts`**:
  - Initialize `$device_id` from anonymous ID on first SDK init
  - Preserve across `reset()` by default (opt-in `resetDeviceId` param to regenerate)
  - Expose `getDeviceId()` with lazy-init for upgrades from older SDK versions
- **`packages/react-native/test/posthog.spec.ts`**: 8 new tests covering init, persistence, identify/reset preservation, flag request payload, and upgrade path

## Test plan

- [x] All 8 new device bucketing tests pass
- [x] All 67 existing RN tests pass (no regressions)
- [x] All 381 core tests pass (no regressions)
- [x] Manual verification with a flag configured for device bucketing on the server side

### DIFF
--- a/packages/browser/playwright/mocked/retry-queue.spec.ts
+++ b/packages/browser/playwright/mocked/retry-queue.spec.ts
@@ -9,6 +9,7 @@ const startOptions = {
 
 test.describe('retry queue', () => {
     test('retries failed capture requests and stops after success', async ({ page, context }) => {
+        test.setTimeout(90000)
         const captureRequests: Request[] = []
         let errorResponseCount = 0
         const maxErrorResponses = 3
@@ -77,9 +78,10 @@ test.describe('retry queue', () => {
         await expect(async () => {
             expect(captureRequests.length).toBe(requestCountAfterSuccess)
         }).toPass({ timeout: 5000 })
-    }, 90000)
+    })
 
     test('stops retrying after 10 attempts', async ({ page, context }) => {
+        test.setTimeout(60000)
         const captureRequests: Request[] = []
 
         // Mock the capture endpoint to always fail
@@ -122,7 +124,7 @@ test.describe('retry queue', () => {
         // Verify counts are incrementing
         expect(retryCountMatches).toContain(1)
         expect(retryCountMatches).toContain(2)
-    }, 60000)
+    })
 
     test('immediately retries when coming back online', async ({ page, context }) => {
         const captureRequests: Request[] = []


.changeset/afraid-mammals-train.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': minor
```
marandaneto: looks like its dupe by .changeset/device-bucketing-rn.md

packages/react-native/src/posthog-rn.ts
```diff
@@ -218,6 +218,12 @@ export class PostHog extends PostHogCore {
 
       this.setupBootstrap(options)
 
+      // Initialize device_id if not already set. This provides a stable identifier
+      // for device-level feature flag bucketing that survives identify() and reset().
+      if (!this.getPersistedProperty(PostHogPersistedProperty.DeviceId)) {
+        this.setPersistedProperty(PostHogPersistedProperty.DeviceId, this.getAnonymousId())
```
marandaneto: why do we reuse the this.getAnonymousId()? should we generate a deviceId instead (getDeviceId)?

packages/react-native/src/posthog-rn.ts
```diff
@@ -428,11 +434,17 @@ export class PostHog extends PostHogCore {
...
    *
    * @param propertiesToKeep - Optional array of persisted properties to preserve during reset
+   * @param resetDeviceId - If true, generates a new device ID. Defaults to false,
+   *   preserving the device ID for stable feature flag bucketing across identity changes.
    *
    * @public
    */
-  reset(propertiesToKeep?: PostHogPersistedProperty[]): void {
-    super.reset(propertiesToKeep)
+  reset(propertiesToKeep?: PostHogPersistedProperty[], resetDeviceId?: boolean): void {
```
marandaneto: Instead of having `resetDeviceId`, should we actually have the `propertiesToKeep` with a default value of `PostHogPersistedProperty.DeviceId`?
feels like 2 params for the same feature

packages/react-native/src/posthog-rn.ts
```diff
@@ -744,6 +756,27 @@ export class PostHog extends PostHogCore {
     return super.getDistinctId()
   }
 
+  /**
+   * Returns the stable device identifier used for device-level feature flag bucketing.
+   * This ID persists across identify() and reset() calls, only changing on
+   * reset(undefined, true) or a fresh app install.
```
marandaneto: also changes if the user deletes the cached data (manually) or if the OS decides to clean the OS cache data

packages/react-native/src/posthog-rn.ts
```diff
@@ -744,6 +756,27 @@ export class PostHog extends PostHogCore {
...
+   *
+   * @returns The device ID, or an empty string if not yet initialized
+   */
+  getDeviceId(): string {
+    const deviceId = this.getPersistedProperty<string>(PostHogPersistedProperty.DeviceId)
+    if (!deviceId) {
+      // Lazy init for upgrades: existing installs won't have a device_id yet
+      const anonId = this.getAnonymousId()
```
marandaneto, line 795: we prob should check if anonId isnt empty

packages/react-native/src/posthog-rn.ts
marandaneto: we prob should check if deviceId isnt empty

.changeset/afraid-mammals-train.md
dmarticus: oh whoops, I thought I canceled one of the changeset operations.

packages/react-native/src/posthog-rn.ts
dmarticus: We use `getAnonymousId()` as a *seed* value — it gives us a stable UUID that already exists at init time without generating a new one. Once set, the device ID is fully independent from the anonymous ID: it persists across `identify()` and `reset()` calls even as the anonymous ID changes.

The latest commit adds empty-string guards on both the init path and `getDeviceId()` so we never persist an empty value.

packages/react-native/src/posthog-rn.ts
dmarticus: Good call — removed `resetDeviceId` entirely. `reset()` now defaults `propertiesToKeep` to include `DeviceId` (along with the lifecycle properties), so callers only need to pass `propertiesToKeep` if they explicitly want to clear it. Single param, no redundancy.

packages/react-native/src/posthog-rn.ts
dmarticus: Good point — updated the JSDoc to mention that the device ID changes if the user clears the app cache or if the OS reclaims storage. Also added a lazy-init path in `getDeviceId()` so that if the persisted value is missing (cache cleared, OS cleanup, or app upgrade from before this feature), we regenerate from the current anonymous ID.

packages/react-native/src/posthog-rn.ts
dmarticus: Added — both the init path and `getDeviceId()` now check that `anonId` is truthy before persisting it.

packages/react-native/src/posthog-rn.ts
dmarticus: Added empty-string guard in the init block as well — we only persist `deviceId` if `anonId` is non-empty.

marandaneto (approved):

PR #3342, opened 2026-04-07
feat(replay): property filtering on triggers

ksvat:
## Problem

we want to allow more flexibility and allow for "WHERE" clauses, or basically just property filtering (ie, EVENT where path=foo)

## Changes  
  
NOTE: 

I implemented "group level" properties but dont plan on exposing it in the UI for a while because theres some funkiness to it.   
  
Logic flow:   
 Event captured  
→ For each trigger group:  
    → Does event name match any event trigger in this group?   
        → Yes → Check group-level property filters (AND logic across all filters)  
              → Pass → Check per-event property filters for the matched event trigger  
                  → Pass → Activate this group's trigger (persisted per-group)  
                  → Fail → Skip activation  
              → Fail → Skip activation  
        → No → Skip     

matchTriggerPropertyFilters does the actual comparison. It:

- Routes each filter to the right property source based on filter.type ('event' or 'person')
- ANDs all filters together (every filter must match, ie "path contains x" and "path doesnt contain y" must match all)
- ORs multiple values within a single filter (any value can match, ie "path contains x, y, z" can match on any of those)
- Handles array-valued properties by normalizing both sides to string arrays
- Delegates to the existing propertyComparisons operators (exact, icontains, regex, gt, lt, etc.)

:star2::warning: Person properties come from STORED_PERSON_PROPERTIES_KEY in persistence, They're not fetched on demand, so there's a timing gap on initial page load before they're available, and could possibly be "stale" later

see screen recording for testing locally. Conditions: 

see via screen recording that recording does NOT start when i visit trigger flag (768) but DOES start when i visit Wizard flag (not contains 768) [trigger_conditons_2_conditions.mov <span class="graphite__hidden">(uploaded via Graphite)</span> ](https://app.graphite.com/user-attachments/video/7d691b5f-df88-47f7-808f-eb3e8e02abda.mov)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/extensions/replay/external/recording-strategies.ts
+++ b/packages/browser/src/extensions/replay/external/recording-strategies.ts
@@ -13,6 +13,7 @@ import {
     SESSION_RECORDING_TRIGGER_V2_GROUP_EVENT_PREFIX,
     SESSION_RECORDING_TRIGGER_V2_GROUP_URL_PREFIX,
     SESSION_RECORDING_TRIGGER_V2_GROUP_SAMPLING_PREFIX,
+    STORED_PERSON_PROPERTIES_KEY,
 } from '../../../constants'
 import {
     EventTriggerMatching,
@@ -33,6 +34,7 @@ import {
 import { sampleOnProperty } from '../../sampling'
 import { isBoolean, isNull, isNullish, isNumber } from '@posthog/core'
 import { createLogger } from '../../../utils/logger'
+import { matchTriggerPropertyFilters } from '../../../utils/property-utils'
 
 const logger = createLogger('[SessionRecording]')
 
@@ -370,9 +372,12 @@ export class V2TriggerGroupStrategy implements RecordingStrategy {
                 onPause,
                 onResume,
                 (triggerType) => {
-                    // Use per-group activation instead of global V1 _activateTrigger
+                    // Check group-level property filters before activating
+                    if (!this._checkGroupLevelProperties(matcher, undefined)) {
+                        return
+                    }
+
                     matcher.activateTrigger(triggerType, sessionId)
-                    // Update session properties after activation
                     this.updateActiveTriggers(sessionId)
                 },
                 sessionId
@@ -403,6 +408,31 @@ export class V2TriggerGroupStrategy implements RecordingStrategy {
                     matcher.checkEventTriggerConditions(
                         event.event,
                         (triggerType) => {
+                            // Check group-level property filters
+                            if (!this._checkGroupLevelProperties(matcher, event.properties)) {
+                                return
+                            }
+
+                            // Check per-event property filters.
+                            // A group may contain multiple entries for the same event name with
+                            // different property filters — e.g. "purchase amount > 100" OR
+                            // "purchase by a VIP". Treat same-name entries as a disjunction:
+                            // activate if *any* matching clause's filters pass. An entry with no
+                            // properties is treated as an unconditional match (matching V1).
+                            const matchedTriggers = (matcher.group.conditions.events || []).filter(
+                                (t) => t.name === event.event
+                            )
+                            const personProperties = this._instance.get_property(STORED_PERSON_PROPERTIES_KEY)
+                            const anyMatched = matchedTriggers.some(
+                                (t) =>
+                                    !t.properties ||
+                                    t.properties.length === 0 ||
+                                    matchTriggerPropertyFilters(t.properties, event.properties, personProperties)
+                            )
+                            if (!anyMatched) {
+                                return
+                            }
+
                             matcher.activateTrigger(triggerType, sessionId)
                             this.updateActiveTriggers(sessionId)
                         },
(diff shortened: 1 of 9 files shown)


ksvat:
* **#3365** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3365?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3342** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3342?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3342?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3281** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3281?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -370,9 +372,12 @@ export class V2TriggerGroupStrategy implements RecordingStrategy {
                 onPause,
                 onResume,
                 (triggerType) => {
-                    // Use per-group activation instead of global V1 _activateTrigger
+                    // Check group-level property filters before activating
+                    if (!this._checkGroupLevelProperties(matcher, undefined)) {
```
TueHaulund, line 376: **URL/flag activation passes `undefined` as event properties** — any group that combines `conditions.urls` or `conditions.flag` with a group-level `properties` entry of `type: 'event'` will now fail to activate.

Inside `matchTriggerPropertyFilters`, an `event`-type filter reads `undefined?.[filter.key]`, gets `undefined`, short-circuits to `false`, and the whole `.every()` returns `false`. The URL/flag match is discarded and the session stays buffering forever. There's no event in scope on this code path, so event-type filters aren't just absent — they're categorically inapplicable.

Suggested fix: restrict `_checkGroupLevelProperties` on this path to `type: 'person'` filters and treat event-type filters as non-applicable (pass). Alternatively, validate upstream in config parsing that event-type group filters are incompatible with URL/flag-only trigger groups. Either way, a regression test that combines a URL trigger with a group-level event filter would lock this in.

packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -403,6 +408,27 @@ export class V2TriggerGroupStrategy implements RecordingStrategy {
...
                         event.event,
                         (triggerType) => {
+                            // Check group-level property filters
+                            if (!this._checkGroupLevelProperties(matcher, event.properties)) {
+                                return
+                            }
+
+                            // Check per-event property filters
+                            const eventTriggers = matcher.group.conditions.events || []
+                            const matchedTrigger = eventTriggers.find((t) => t.name === event.event)
```
TueHaulund: **`.find()` silently drops same-name event clauses.** A trigger group expressing "purchase with amount > 100 OR purchase by a VIP" by listing `purchase` twice with different property filters will only ever evaluate the first clause — subsequent same-name entries are unreachable.

The previous implementation flattened all event names into a list (`eventNames` in `TriggerGroupMatching`), so same-name duplicates were equivalent and the property-less match in `checkEventTriggerConditions` still fires here. But the per-event property check now makes them semantically distinct, and `.find` turns what users would naturally write as OR across clauses into "first clause wins".

Suggested shape (adjust naming to match surrounding style):

```ts
const matchedTriggers = eventTriggers.filter((t) => t.name === event.event)
const anyMatched = matchedTriggers.some(
    (t) =>
        !t.properties ||
        t.properties.length === 0 ||
        matchTriggerPropertyFilters(
            t.properties,
            event.properties,
            this._instance.get_property(STORED_PERSON_PROPERTIES_KEY)
        )
)
if (!anyMatched) {
    return
}
```

Worth a regression test with two same-name event entries — one passing, one failing — asserting the group still activates.

packages/browser/src/extensions/replay/external/recording-strategies.ts
ksvat: I am restricting this on the FE rn -- you cant have an event filter w URL trigger, and i plan to have warnings/fail to save if thats the case.   
  
IDK i def dont feel like "pass" is the correct behavior... if i'm somehow able to add a filter/get around the FE restrictions, i expect it to further filter my results (and a fallback of failing is safer) vs auto-passing   
  
for example if a user somehow got their state to "url regex = foo, event.counry = US" ... those conditions will never match (bc theres no event) so it should fail IMO rather than just passing everything with URL = foo

ksvat:
but there is a changeset.......  .changeset/early-numbers-mate.md

PR #3351, opened 2026-04-07
fix(node): send $device_id as top-level field in /flags request for device bucketing

### LINKED ISSUE #3318: posthog-node: $device_id not sent as top-level field in /flags request, breaking device bucketing on remote evaluation
## Bug description

`posthog-node` sends `$device_id` only inside `person_properties` when calling `/flags`, but the feature-flags service expects it as a **top-level field** in the request body. This causes all device-bucketed flags to return `false` during server-side remote evaluation, even when `$device_id` is correctly passed via `personProperties` per [the docs](https://posthog.com/docs/feature-flags/device-bucketing).

The bug is in [`getFlags()` in `posthog-core-stateless.ts`](https://github.com/PostHog/posthog-js/blob/main/packages/core/src/posthog-core-stateless.ts#L459-L478) — `requestData` is constructed without extracting `$device_id` from `personProperties` into a top-level field:

```ts
const requestData: Record<string, any> = {
    token: this.apiKey,
    distinct_id: distinctId,
    groups,
    person_properties: personProperties,  // $device_id is buried here
    group_properties: groupProperties,
    ...extraPayload,
}
```

The browser SDK sends `$device_id` as a top-level field, which is why client-side device bucketing works. Local evaluation also works correctly (`getBucketingValueForFlag` reads `$device_id` from `personProperties`). Only server-side remote evaluation is broken.

The Rust feature-flags service logs: `Flag configured for device_id bucketing but no device_id provided, returning false`

Tested on self-hosted PostHog with `posthog-node` v5.26.1 and v5.28.9 (latest).

## How to reproduce

1. Create a feature flag with "Match by" set to **Device** (device bucketing)
2. Evaluate it server-side with `posthog-node`:
   ```js
   const value = await posthog.getFeatureFlag('my-device-flag', distinctId, {
     personProperties: { $device_id: deviceId },
   })
   // value is always false for device-bucketed flags
   ```
3. Confirm with raw curl — top-level `$device_id` works, `person_properties` only does not:
   ```bash
   # Works — returns correct variant
   curl -X POST "https://your-host/flags/?v=2" -H "Content-Type: application/json" \
     -d '{"token":"...","distinct_id":"ABC","$device_id":"device1","person_properties":{"$device_id":"device1"}}'

   # Broken — device-bucketed flags return false / out_of_rollout_bound
   curl -X POST "https://your-host/flags/?v=2" -H "Content-Type: application/json" \
     -d '{"token":"...","distinct_id":"ABC","person_properties":{"$device_id":"device1"}}'
   ```

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

**Workaround:** Use local evaluation (`personalApiKey` option) which correctly handles device bucketing, or call `/flags` directly with `$device_id` as a top-level field in the request body.

dmarticus:
## Summary

- Fixes #3318: `posthog-node` sends `$device_id` only inside `person_properties` when calling `/flags`, but the feature flags service expects it as a **top-level field** for device-based bucketing. This caused all device-bucketed flags to return `false` during server-side remote evaluation.
- Extracts `$device_id` from `personProperties` and adds it as a top-level field in the `/flags` request payload in `posthog-core-stateless.ts`, matching the behavior of the browser SDK.

## Test plan

- [x] Added test: `$device_id` is sent as a top-level field when present in `personProperties`
- [x] Added test: `$device_id` is not included as a top-level field when absent from `personProperties`
- [x] All existing core tests pass (384 tests)

### DIFF
--- a/.changeset/fix-device-id-flags.md
+++ b/.changeset/fix-device-id-flags.md
@@ -1,5 +1,6 @@
 ---
 '@posthog/core': patch
+'posthog-node': patch
 ---
 
 Send $device_id as a top-level field in /flags requests so the feature flags service can use it for device-based bucketing during remote evaluation


.changeset/fix-device-id-flags.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/core': patch
```
marandaneto, line 2: if you just publish core, customers wont see an effect until a new version of the web, node, rn etc are also published, is that the intention?

dmarticus: ah, shoot, was being too clever with the changeset – didn't realize changes to core wouldn't propagate. I'll update – thank you!

PR #3362, opened 2026-04-09
fix(cookieless): start in cookieless mode when opt_out_capturing_by_default is set

### LINKED ISSUE #2841: opt_out_capturing_by_default and cookieless_mode
## Bug description

Hello,
I think that there is something wrong with **opt_out_capturing_by_default** and **cookieless_mode**

## How to reproduce

Configure posthog init as below:
`posthog.init("xxx", { api_host: 'xxx', ui_host: 'https://eu.posthog.com', cookieless_mode: "on_reject", opt_out_capturing_by_default: true, capture_exceptions: true, debug: true })`

posthog-js is not sending any event until `posthog.opt_out_capturing();` is called

Is it the expected behaviour? I'd expect that it immediately sends data in cookieless mode, because basically opt_out_capturing_by_default should behave like posthog.opt_out_capturing(); but immediately since first call.

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

sampennington:
## Problem

Closes #2841.

If you set `cookieless_mode: 'on_reject'` together with `opt_out_capturing_by_default: true`, the SDK drops every event instead of falling back to cookieless. The customer's workaround was to call `opt_out_capturing()` manually.

The two cookieless gates (`is_capturing()` and the `$cookieless_mode` property tag) both checked `isExplicitlyOptedOut()`, which is only true for `DENIED`. But `opt_out_capturing_by_default` leaves consent at `PENDING` — so the checks fall through and capture gets blocked entirely.

## Changes

- New `isRejected()` on `ConsentManager`
- Swapped two sites in `posthog-core.ts` over to `isRejected()`: `is_capturing()` and the `$cookieless_mode` tag in `calculateEventProperties`. 

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/consent.ts
```diff
@@ -57,6 +57,13 @@ export class ConsentManager {
         return this.consent === ConsentStatus.DENIED
     }
 
+    public isRejected() {
+        return (
+            this.consent === ConsentStatus.DENIED ||
+            (this.consent === ConsentStatus.PENDING && this._config.opt_out_capturing_by_default === true)
```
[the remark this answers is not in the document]
sampennington: Fixed in `a6cb1f0e`

sampennington:
Hey @robbie-c this [ticket](https://posthoghelp.zendesk.com/agent/tickets/54631) was trying to be sent to you last week, but I had a go at it.
Client libraries team were not sure on the intended behaviour, but pretty sure this is what it should do. Can you confirm? 😄

packages/browser/src/posthog-core.ts
```diff
@@ -3520,6 +3473,12 @@ export class PostHog implements PostHogInterface {
                     this.sessionRecording,
                     new SessionRecordingClass(this) as SessionRecording
                 )
+                // Replay the cached remote config so the new recorder picks up server-side
+                // settings (enable flag, endpoint, sampling) that arrived while we were
+                // still in cookieless mode and sessionRecording didn't yet exist.
+                if (this._lastRemoteConfig) {
+                    this.sessionRecording?.onRemoteConfig?.(this._lastRemoteConfig)
```
marandaneto, line 3478: method `onRemoteConfig` gets called automatically when extensions are loaded, so we dont need to do that i guess?

packages/browser/src/consent.ts
```diff
@@ -57,6 +57,13 @@ export class ConsentManager {
         return this.consent === ConsentStatus.DENIED
     }
 
+    public isRejected() {
```
marandaneto, line 56: `isOptedOut` should call this method then since it does quite the same thing and we can improve a bit of bundle size by reducing the code

packages/browser/src/posthog-core.ts
sampennington: A bunch of session replay playwright tests would fail without this, and this seemed to be the fix. I'll
double check what that was again 👀

packages/browser/src/posthog-core.ts
```diff
@@ -1388,7 +1388,7 @@ export class PostHog implements PostHogInterface {
 
         if (
             this.config.cookieless_mode == COOKIELESS_ALWAYS ||
-            (this.config.cookieless_mode == COOKIELESS_ON_REJECT && this.consent.isExplicitlyOptedOut())
+            (this.config.cookieless_mode == COOKIELESS_ON_REJECT && this.consent.isRejected())
```
marandaneto: this if does something very similar to L948-952, can we extract and reuse?

packages/browser/src/posthog-core.ts
sampennington: Yes so these fail:
```
  1) [chromium] › reproduces customer issue: cookieless_mode on_reject + opt_out_capturing_by_default
     Test timeout of 30000ms exceeded.
     (hangs on waitingForNetworkCausedBy({ urlPatternsToWaitFor: ['**/*recorder.js*'] }))

  2) [chromium] › session recording auto-starts after opt_in_capturing without explicit startSessionRecording call
     Test timeout of 30000ms exceeded.
     (same — recorder.js never loads, so the network-wait times out)
```

Reason being:
We skip creating sessionRecording at init when startInCookielessMode is true (L773), so it isn't in _extensions when onRemoteConfig fans out. When we create it later during opt-in, it has no persisted remote config to read, so _isRecordingEnabled is false and recording never starts. The replay is what gives the new recorder its config.

packages/browser/src/consent.ts
sampennington: Done ✅

packages/browser/src/posthog-core.ts
marandaneto: mm then we have to fix the test instead? it does not look like the correct approach having the remote config state cached in a flag just for making the tests happy

sampennington: I don't think thats whats happening here, without this the session replay wouldn't work I thought but I can seee if there's a nicer way to do it

marandaneto (approved):

packages/browser/src/posthog-core.ts
marandaneto: not a blocker then, feel free to merge it if this fixes the main issue

PR #3378, opened 2026-04-13
fix: disable native async gzip after NotReadableError

marandaneto:
## Problem

Some browser SDK users are seeing `NotReadableError` from the native async gzip path when sending compressed requests. When that happens, the request should still go through and we should stop retrying the broken native compression path for the rest of the page lifetime.

## Changes

- add a `rethrow` option to `gzipCompress` in `@posthog/core`, with docs explaining the Safari 16 `NotReadableError` use case
- move native gzip `NotReadableError` detection into `@posthog/core` so browser code can reuse it
- update `packages/browser/src/request.ts` to reuse `gzipCompress(..., { rethrow: true })`
- detect `NotReadableError`, swallow it, retry the current request without compression, and disable native async gzip for subsequent requests in the same page session
- keep the existing synchronous gzip-js fallback for other compression failures
- add a changeset for `posthog-js`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -560,4 +560,126 @@ describe('request', () => {
             )
         })
     })
+
+    describe('native async gzip retry flow', () => {
+        let isolatedRequestModule: any
+        let isolatedCompression: typeof Compression
+        let mockedIsolatedFetch: jest.Mock
+        let mockedIsolatedGzipCompress: jest.Mock
+
+        beforeEach(async () => {
+            jest.resetModules()
+            jest.clearAllMocks()
+            jest.useFakeTimers()
+            jest.setSystemTime(now)
+
+            mockedIsolatedFetch = jest.fn(() =>
+                Promise.resolve({
+                    status: 200,
+                    text: () => Promise.resolve('{ "a": 1 }'),
+                })
+            )
+            mockedIsolatedGzipCompress = jest.fn()
+
+            jest.doMock('../utils/globals', () => ({
+                ...jest.requireActual('../utils/globals'),
+                fetch: mockedIsolatedFetch,
+                XMLHttpRequest: jest.fn(),
+                navigator: {
+                    sendBeacon: jest.fn(),
+                },
+                CompressionStream: jest.fn(),
+            }))
+
+            jest.doMock('@posthog/core', () => ({
+                ...jest.requireActual('@posthog/core'),
+                gzipCompress: mockedIsolatedGzipCompress,
+            }))
+
+            isolatedRequestModule = await import('../request')
+            isolatedCompression = (await import('../types')).Compression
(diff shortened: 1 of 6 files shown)


packages/browser/src/__tests__/request.test.ts
```diff
@@ -560,4 +560,105 @@ describe('request', () => {
...
+                method: 'POST',
+                compression: isolatedCompression.GZipJS,
+            })
+
+            await flushPromises()
+
+            expect(mockedIsolatedGzipCompress).toHaveBeenCalledTimes(1)
+            expect(mockedIsolatedFetch).toHaveBeenCalledTimes(1)
+            expect(mockedIsolatedFetch.mock.calls[0][0]).toContain('&compression=gzip-js')
+            expect(mockedIsolatedFetch.mock.calls[0][1].body).toBe('{"foo":"bar"}')
```
ioannisj: Since this is the uncompressed path, shouldn't we be stripping `compression=gzip-js` as well?

marandaneto: retries after error, so the request should have it

marandaneto: also gzip native async != sync gzip with fflate (fallback)

ioannisj: Make sense for the original request yes, but shouldn't the retry request remove that query param?

marandaneto: the fallback is also gzip but with fflate so why would we remove it?

ioannisj: But the assertion above is that body is plain text not ArrayBuffer. Just trying understand and make sure that this would not eventually fail server side

marandaneto: i think i understood now, lemme check
the url is constructed before failing so the parameter is stale, the behaviour is correct tho

marandaneto: fixed by reconstructing the url instead

marandaneto: normal browser:
1st native async gzip then 2nd gzip with fflate then 3rd uncompressed request

safari browser with this bug:
1st native async gzip error then 2nd uncompressed request

ioannisj (approved):
LG, waiting for tests to pass

PR #3379, opened 2026-04-13
fix(browser): merge partial /flags response with bootstrapped flags

dmarticus:
## Summary

- When `advanced_only_evaluate_survey_feature_flags: true` is set, the `/flags` response now **merges** with existing flags instead of overwriting them
- Bootstrapped non-survey flags (e.g., session recording linked flags) are preserved after the survey-only `/flags` response arrives

## Root Cause

`parseFlagsResponse` has two code paths: a **merge** path (when `errorsWhileComputingFlags` is true) and an **overwrite** path (default). When `advanced_only_evaluate_survey_feature_flags` is enabled, the server intentionally returns only survey-related flags — but since this isn't an error condition, `errorsWhileComputingFlags` is false. The SDK treats this intentionally-partial response as a complete flag set and overwrites everything in persistence, wiping all bootstrapped non-survey flags.

This means any customer using `bootstrap.featureFlags` together with `advanced_only_evaluate_survey_feature_flags: true` loses their bootstrapped flags as soon as the `/flags` response lands — breaking session recording linked flags, custom feature flags, and anything else that was bootstrapped.

## Fix

Added a `partialResponse` option that threads from `_callFlagsEndpoint` → `receivedFeatureFlags` → `parseFlagsResponse`. When `advanced_only_evaluate_survey_feature_flags` is true, the response is treated as partial and merged with existing flags (new values take precedence for overlapping keys, but non-survey flags are preserved).

## Test plan

- [x] New unit tests: partial response merges with existing flags
- [x] New unit tests: overlapping keys in partial response update correctly
- [x] New unit tests: without `partialResponse`, overwrite behavior is unchanged

---

### DIFF
--- a/.changeset/fix-partial-flags-overwrite-bootstrap.md
+++ b/.changeset/fix-partial-flags-overwrite-bootstrap.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fix bootstrapped feature flags being overwritten by partial /flags response when `advanced_only_evaluate_survey_feature_flags` is enabled
--- a/packages/browser/src/__tests__/featureflags.test.ts
+++ b/packages/browser/src/__tests__/featureflags.test.ts
@@ -2888,124 +2888,84 @@ describe('parseFlagsResponse', () => {
     })
 
     describe('partialResponse option', () => {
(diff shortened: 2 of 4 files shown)


haacked (approved):
LGTM!

PR #3386, opened 2026-04-14
fix(browser): gate versioned lazy bundle paths behind preview flag

dustinbyrne:
## Problem

We want to support semver-qualified external dependency bundle paths like `/static/1.370.0/recorder.js`, but we need to ship the SDK change without breaking existing deployments that still only route the legacy `/static/recorder.js?v=1.370.0` form.

We also want preview-domain testing to work for the broader `/static/*` surface, not just exact semver asset paths.

## Changes

- keep the legacy external dependency asset URLs as the default behavior
- add a temporary internal `__preview_external_dependency_versioned_paths` flag to opt into `/static/${posthog.version}/${kind}.js`
- allow that preview flag to be a string asset-host override for any `/static/*` asset path while leaving non-static asset paths unchanged
- keep non-versioned asset paths on the normal host unless an explicit string override is configured
- add Jest coverage for both the default and preview-flagged asset path behavior
- add config typing for the preview flag and update the changeset

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


marandaneto (approved):

packages/browser/src/posthog-core.ts
```diff
@@ -391,7 +390,6 @@ export class PostHog implements PostHogInterface {
     __request_queue: QueuedRequestWithOptions[]
     _pendingRemoteConfig?: RemoteConfig
     _remoteConfigLoader?: RemoteConfigLoader
-    _resolvedSdkVersion?: string
```
turnipdabeets, line 394: Just confirming: after this revert, is sdkVersion.resolved in the remote config now unused on the client? It looks like the backend might still be populating it.

dustinbyrne: the route which populates it is now unused - I plan on circling back to delete all that

turnipdabeets (approved):

PR #3387, opened 2026-04-14
feat(ai): return prompt metadata from Prompts.get()

carlos-marchal-ph:
## Problem

`Prompts.get()` returns only the prompt string, discarding `name` and `version` from the API response. Users want to track `$ai_prompt_name` / `$ai_prompt_version` on AI generation events without a separate HTTP call.

## Changes

- Add `withMetadata` option to `get()`. When `true`, returns a `PromptResult` discriminated union with `source` (`api` | `cache` | `stale_cache` | `code_fallback`), `name`, and `version`.
- Deprecate calling `get()` without `withMetadata` (warns once per instance). Plain-string return will be removed in next major.
- Internal cache now stores metadata alongside the prompt text.
- Validate `name` and `version` fields in API response.

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code

### DIFF
--- a/.changeset/prompt-metadata.md
+++ b/.changeset/prompt-metadata.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': minor
+---
+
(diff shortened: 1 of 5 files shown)


andrewm4894 (approved):
do we need a changeset ?

PR #3389, opened 2026-04-14
feat: add exception steps buffering and public API

hpouillot:
## Problem

Error tracking currently captures `$exception` events but does not support attaching pre-error breadcrumb context from the browser SDK. We need a way to record exception steps and include them on the next captured exception in a format compatible with the backend/UI `$exception_steps` handling.

## Changes

- Added shared exception-steps support in `@posthog/core`:
  - `EXCEPTION_STEP_INTERNAL_FIELDS` constants
  - `ExceptionStepsBuffer` (FIFO queue with `max_queue_size`)
  - config resolution helpers and defaults (`enabled=true`, `max_queue_size=20`, `max_bytes=16384`)
  - reserved key stripping for `$message/$timestamp/$type/$level`
  - byte-budget filtering for attached steps (keeps steps closest to the exception)
- Added browser SDK API:
  - `posthog.addExceptionStep(message, properties?)`
- Wired exception steps into browser exception capture flow:
  - attaches buffered steps to `$exception_steps` when not manually provided
  - manual `$exception_steps` override is respected
  - buffer is cleared after a capture attempt
  - buffer is preserved when exceptions are dropped by suppression rules
- Added config support in `@posthog/types`:
  - `ExceptionStepsConfig`
  - `capture_exceptions.exception_steps`
- Updated browser exception autocapture runtime config typing cleanup.
- Added/updated tests in core and browser for queue behavior, byte limits, overrides, clearing/retention semantics, and key conflict handling.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-core.ts
```diff
@@ -3215,6 +3216,26 @@ export class PostHog implements PostHogInterface {
...
+     *
+     * {@label Error tracking}
+     *
+     * @public
+     *
+     * @example
+     * ```js
+     * posthog.addExceptionStep('Checkout button clicked', {
+     *   type: 'ui.interaction',
+     *   level: 'info',
```
marandaneto: are those special properties that are highlighted in the PostHog UI? if so, how do users know about them since the `Properties` is very generic

packages/browser/src/posthog-exceptions.ts
```diff
@@ -70,34 +79,133 @@ export class PostHogExceptions implements Extension {
...
+                if (!this._captureExtensionExceptions && this._isExtensionException(exceptionList)) {
+                    logger.info('Skipping exception capture because it was thrown by an extension')
+                    return
+                }
+
+                if (
+                    !this._instance.config.error_tracking.__capturePostHogExceptions &&
+                    this._isPostHogException(exceptionList)
+                ) {
+                    logger.info('Skipping exception capture because it was thrown by the PostHog SDK')
```
marandaneto, line 133: should we add a addExceptionStep for this? so users know the previous exceptions were dropped due to this reason

packages/browser/src/posthog-exceptions.ts
```diff
@@ -70,34 +79,133 @@ export class PostHogExceptions implements Extension {
...
+
+            this._exceptionStepsBuffer.add({
+                [ErrorTracking.EXCEPTION_STEP_INTERNAL_FIELDS.MESSAGE]: message,
+                [ErrorTracking.EXCEPTION_STEP_INTERNAL_FIELDS.TIMESTAMP]: new Date().toISOString(),
+                ...(type ? { [ErrorTracking.EXCEPTION_STEP_INTERNAL_FIELDS.TYPE]: type } : {}),
+                ...(level ? { [ErrorTracking.EXCEPTION_STEP_INTERNAL_FIELDS.LEVEL]: level } : {}),
+                ...sanitizedProperties,
+            })
+        } catch (error) {
+            logger.warn('Failed to add exception step. Ignoring breadcrumb.', error)
```
marandaneto: i think those should be `error` as for every other catch clause

packages/browser/src/posthog-exceptions.ts
```diff
@@ -70,34 +79,133 @@ export class PostHogExceptions implements Extension {
...
+                $exception_steps: exceptionSteps,
+            }
+        } catch (error) {
+            logger.warn('Failed to read buffered exception steps. Capturing exception without steps.', error)
+            return properties
+        }
+    }
+
+    private _coerceExceptionStepProperties(properties?: Properties): Record<string, unknown> {
+        if (!properties || typeof properties !== 'object' || isArray(properties)) {
```
marandaneto: i think we have an util class for isObject or something

packages/core/src/error-tracking/exception-steps.ts
```diff
@@ -0,0 +1,258 @@
...
+export type ResolvedExceptionStepsConfig = {
+  enabled: boolean
+  max_queue_size: number
+  max_bytes: number
+}
+
+export const DEFAULT_EXCEPTION_STEPS_CONFIG: ResolvedExceptionStepsConfig = {
+  enabled: true,
+  max_queue_size: 20,
+  max_bytes: 32768,
```
marandaneto: we should add a human readable comment/jsdoc here eg 32768 is ~32kb

packages/core/src/error-tracking/exception-steps.ts
```diff
@@ -0,0 +1,258 @@
...
+      if (totalBytes + bytes > maxBytes) {
+        break
+      }
+
+      attachableSteps.push(serializedStep.step)
+      totalBytes += bytes
+    }
+
+    return attachableSteps.reverse()
+  }
```
marandaneto: i think we should keep a global var with the current value of totalBytes so we dont need to iterate the whole list every time, otherwise its a performance hit
reset when needed, append when needed based on the last addition

packages/core/src/error-tracking/exception-steps.ts
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    while (this._steps.length > this._config.max_queue_size) {
+      this._steps.shift()
+    }
+  }
+}
+
+function normalizePositiveInteger(input: number | undefined, fallback: number): number {
+  if (typeof input !== 'number' || !isFiniteNumber(input)) {
```
marandaneto: i think we have an util for isNumber

packages/core/src/error-tracking/exception-steps.ts
```diff
@@ -0,0 +1,258 @@
...
+    const parsed = JSON.parse(json)
+    if (!parsed || typeof parsed !== 'object' || Array.isArray(parsed)) {
+      return undefined
+    }
+
+    const parsedStep = parsed as Record<string, unknown>
+    const message = parsedStep[EXCEPTION_STEP_INTERNAL_FIELDS.MESSAGE]
+    const timestamp = parsedStep[EXCEPTION_STEP_INTERNAL_FIELDS.TIMESTAMP]
+
+    if (typeof message !== 'string' || message.trim().length === 0) {
```
marandaneto: i think we have an utils for isString

packages/types/src/posthog-config.ts
```diff
@@ -211,27 +211,55 @@ export interface DeadClickCandidate {
...
+     * @default 20
+     */
+    max_queue_size?: number
+
+    /**
+     * The maximum UTF-8 byte budget used when attaching `$exception_steps` to an exception event.
+     *
+     * @default 32768
+     */
+    max_bytes?: number
```
marandaneto, line 233: we should mention its an `or` thing, `max_queue_size` or `max_bytes` right?

packages/types/src/posthog-config.ts
```diff
@@ -211,27 +211,55 @@ export interface DeadClickCandidate {
     absoluteDelayMs?: number
 }
 
+export type ExceptionStepsConfig = {
```
marandaneto, line 219: already defined in packages/core/src/error-tracking/exception-steps.ts
can we reuse it?

marandaneto:
@ioannisj since i know you walted to talk/implement this

packages/types/src/posthog-config.ts
```diff
@@ -211,27 +211,55 @@ export interface DeadClickCandidate {
...
      *
      * @default false
      */
-    capture_console_errors: boolean
+    capture_console_errors?: boolean
+
+    /**
+     * Controls buffering and payload limits for exception steps added via `addExceptionStep`.
+     */
+    exception_steps?: ExceptionStepsConfig
```
ablaszkiewicz: Why is it a part of `ExceptionAutoCaptureConfig`?

packages/browser/src/posthog-exceptions.ts
```diff
@@ -70,34 +79,133 @@ export class PostHogExceptions implements Extension {
...
             }
 
-            if (
-                !this._instance.config.error_tracking.__capturePostHogExceptions &&
-                this._isPostHogException(exceptionList)
-            ) {
-                logger.info('Skipping exception capture because it was thrown by the PostHog SDK')
+            const propertiesForExceptionCapture =
+                this._exceptionStepsConfig.enabled && isNullish(properties.$exception_steps)
+                    ? this._addBufferedExceptionSteps(properties, this._exceptionStepsConfig.max_bytes)
```
ablaszkiewicz: `_addBufferedExceptionSteps` accepts max bytes to then pass it to `_exceptionStepsBuffer.getAttachable` which defines a default for this variable.

I think we can get rid of this argument at all from all these methods? Buffer is already instantiated with the same config.

ablaszkiewicz (approved):

ablaszkiewicz (approved):
Remember about changeset and label. Great change.

packages/browser/src/posthog-core.ts
cat-ph: yeah good point - do we want to make it more "obvious" that we treat type and level differently?

packages/types/src/posthog-config.ts
cat-ph: ➕

packages/browser/src/posthog-exceptions.ts
```diff
@@ -70,34 +79,133 @@ export class PostHogExceptions implements Extension {
...
+                return this._instance.capture('$exception', propertiesForExceptionCapture, {
+                    _noTruncate: true,
+                    _batchKey: 'exceptionEvent',
+                    _originatedFromCaptureException: true,
+                })
+            } catch (error) {
+                logger.warn('Failed to capture exception event. Dropping this exception.', error)
                 return
+            } finally {
+                this._exceptionStepsBuffer.clear()
```
cat-ph, line 157: question: wondering if there's a case where we want to still keep the buffer? i.e. a capture doesn't actually send something (before_send filtering it maybe or some other way) and then we "lose" the steps here? 🤔

packages/browser/src/posthog-exceptions.ts
hpouillot: Interesting, I didn't think about using exception steps inside the posthog SDK but yes it makes sense. I will do a follow up PR to add more, I think there are lot of interesting steps we could add.

packages/core/src/error-tracking/exception-steps.ts
hpouillot: We cannot do that since the order is reversed but we can add a cache of step size to speed up computation when exception occurs. We could also use it to prune based on max bytes and not only max size.

packages/types/src/posthog-config.ts
hpouillot: I am wondering if we should keep the max_queue_size, specially if we start pruning on max_bytes.

cat-ph (approved):

packages/types/src/posthog-config.ts
marandaneto: makes sense to me, less config the better

packages/core/src/error-tracking/exception-steps.ts
marandaneto: yep, also to not keep counting if you already know the max byte is reached

packages/types/src/posthog-config.ts
hpouillot: Good catch ! I think agent got lost because config because exception autocapture config is not part of the error tracking config. I move the current exception step config inside error tracking and will fix the autocapture config as a follow-up.

packages/browser/src/posthog-core.ts
marandaneto: an option is to make type and level its own params in the addStep method
or make a new object like:

```
StepParams = {
 type // default to something,
 level // default to something,
 parameters // extra params
}
```

then people know they can/need to, the current way i believe no one will set type and level unless they read the source code

packages/types/src/posthog-config.ts
hpouillot: There is no dependencies between the two, so hard to share. I think this question raises another one, should we make posthog core depends on posthog types or do we reaaaaaally need posthog type ? I mean we could just embed the types in posthog core and do `import type`

marandaneto: imo core depending on types makes sense as long as they match
if they are different things, not really
if we can reuse the types, its less duplication and serving fewer bytes
not a blocker for this PR tho since its how it is already

packages/browser/src/posthog-core.ts
hpouillot: To be honest, I am not sure they are needed. We don't expose a filtering on the UI, so they will be just shown as muted text in the UI for now. I would rather keep the api simple : 
```
addExceptionStep("TEXT", additional params)
```

If we want to add them later I think we can add an optional third argument

packages/browser/src/posthog-exceptions.ts
hpouillot: Good catch and I agree, flushing the buffer when exceptions are dropped is not ideal but I don't think we have an afterSend hook. Capturing at the extension level is fire and forget for now.

packages/browser/src/posthog-core.ts
hpouillot: I removed them

packages/core/src/error-tracking/exception-steps.ts
hpouillot: Should be better now https://github.com/PostHog/posthog-js/pull/3389/commits/ea23af80f2235d567573e937ea74c36ab2134f93

packages/browser/src/posthog-exceptions.ts
hpouillot: Fixed with the changes in https://github.com/PostHog/posthog-js/commit/ea23af80f2235d567573e937ea74c36ab2134f93

packages/browser/src/posthog-exceptions.ts
hpouillot: My bad, capture returns undefined when beforeSend returns null. It's fixed now

hpouillot:
Thanks guys for the thorough review on an half-baked PR 🔥

marandaneto (approved):

PR #3404, opened 2026-04-16
ci(release): publish posthog-js assets to semver and major S3 paths

dustinbyrne:
## Problem

We want the `posthog-js` browser asset release to follow a simple S3-first model that is easy to reason about and easy to serve directly from asset hostnames.

The desired release shape is:

- publish immutable assets to `/static/<semver>/...`
- publish the same assets to `/static/<major>/...`
- publish compatibility assets to `/static/...`

That gives us exact-version pinning, major-version aliases, and the existing compatibility path with a much simpler release flow.

## Changes

- replace the shell-based S3 upload script with typed release tooling under `tooling/release`
- publish browser dist assets to three S3 prefixes:
  - `/static/<semver>/...`
  - `/static/<major>/...`
  - `/static/...`
- simplify `.github/workflows/release.yml` so the CDN asset publish path just builds artifacts and uploads them to those S3 prefixes
- keep the sparse checkout fix so release tooling installs include the root `patches/` directory

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


tooling/release/src/upload-posthog-js-s3.ts
```diff
@@ -0,0 +1,229 @@
...
+
+    const majorAliasKeys = await uploadReleaseAssets(bucket, uploadPlans.majorAlias, 'major-version alias assets')
+    await tagS3Keys(bucket, majorAliasKeys)
+    await verifyUploadedAssets(bucket, majorAliasKeys, 'major-version alias assets')
+
+    const compatibilityKeys = await uploadReleaseAssets(
+        bucket,
+        uploadPlans.compatibility,
+        'top-level compatibility assets'
+    )
```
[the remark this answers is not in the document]
dustinbyrne: versions.json is no more

tooling/release/src/s3.ts
```diff
@@ -0,0 +1,78 @@
...
+        cacheControl?: string
+    } = {}
+): Promise<void> {
+    await getS3Client().send(
+        new PutObjectCommand({
+            Bucket: bucket,
+            Key: key,
+            Body: fs.createReadStream(filePath),
+            ContentType: options.contentType,
+            CacheControl: options.cacheControl,
```
turnipdabeets, line 63:
```suggestion
            CacheControl: options.cacheControl,
            Tagging: 'public=true',     
```
Adding Tagging: 'public=true' to the PutObjectCommand in putS3ObjectFromFile folds the upload and the tag into a single S3 call. This might lets you drop tagS3ObjectPublic, the tagS3Keys helper, and the three await tagS3Keys passes in uploadPostHogJsS3 so Cloudflare can serve it immediately with no second round-trip.

dustinbyrne: good catch, the separate tagging step was an artifact of the shell script - fixed in https://github.com/PostHog/posthog-js/pull/3404/commits/f627e4095a6c243d5a0d9fcbf4ff4367667409bc

turnipdabeets (approved):

PR #3425, opened 2026-04-20
fix(browser): classify SDK persistence keys explicitly

marandaneto:
## Problem

The browser SDK currently mixes user super properties and SDK-owned persistence state in the same persistence store, then relies on an exclusion list to avoid leaking internal state into captured events. That made it easy to accidentally hide properties that should stay event-visible, as happened in the regression fixed by #3419.

While landing that fix, both the Playwright and TestCafe real-ingestion browser jobs also showed flaky event API results in CI: some runs returned duplicate ingested events, some surfaced partially-ingested event payloads before the expected `$pageview`/`$autocapture` shape was visible, and the browser jobs were timing out before we had exhausted a long enough polling window for production ingestion to stabilize.

## Changes

- replace the runtime `PERSISTENCE_RESERVED_PROPERTIES` filtering path with an explicit `PERSISTENCE_KEY_POLICY` for SDK-owned persistence keys
- classify SDK persistence keys as `event`, `hidden`, or `derived`, while keeping unknown keys event-visible for backwards-compatible `posthog.register(...)` behavior
- move derived feature-flag event-property transformation into a small shared `persistence-key-transforms.ts` module and reference it from the policy
- add regression tests for event-visible SDK keys like `$feature_flag_payloads` and `$surveys_activated`
- add a stricter completeness test that scans browser persistence write/remove sites, requires constant-backed keys, and enforces policy coverage for exact and prefix-based SDK keys
- centralize `PostHogPersistence` key mutation behind `_setProp` / `_deleteProp` and add an AST mutation-boundary test so wrappers/refactors still have to flow through a small checked sink surface
- replace remaining raw SDK persistence/session keys at scanned write sites with shared constants
- update `packages/browser/terser-mangled-names.json` for the new mangled browser property names so the consistency check passes
- de-dupe ingested Playwright integration events by `uuid` before running the real-ingestion assertions so duplicate API rows do not fail the job
- keep polling Playwright real-ingestion results until the actual assertion passes, rather than stopping once any target count is visible, so the job waits for the expected unique event IDs and payload shape to appear
- extend the Playwright integration workflow/job timeout to 30 minutes so CI only fails after the longer ingestion polling window is exhausted
- de-dupe ingested TestCafe IE11 events by `uuid` before counting/asserting them, so the real-ingestion checks operate on unique events
- keep polling TestCafe real-ingestion results until the actual assertion passes, rather than stopping once any target count is visible, so duplicate rows and partially-ingested payloads do not fail the job early
- align the TestCafe workflow timeout with the ingestion checker deadline so CI does not kill the event verification step prematurely

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/persistence-key-policy.ts
```diff
@@ -0,0 +1,167 @@
...
+    WEB_VITALS_ENABLED_SERVER_SIDE,
+} from './constants'
+
+export type PersistenceKeyExposure = 'event' | 'hidden' | 'derived'
+
+interface PersistenceKeyPolicyEntry {
+    exposure: PersistenceKeyExposure
+}
+
+export const PERSISTENCE_KEY_POLICY: Record<string, PersistenceKeyPolicyEntry> = {
```
marandaneto, line 80: i'd need help to validate this, whats supposed to be set in events, and what not

turnipdabeets:
Really like the direction here — splitting "SDK-owned" from "event-visible" and making the classification explicit is the right call, and the AST test is a nice forcing function.                                     
                                                                                                                                                                                                                         
  One suggestion (not a blocker) with how the guarantee is enforced, though:         
                                                                                                                                                                                                                         
  The AST test catches unclassified keys by pattern-matching a hardcoded list of method names (register, set_property, register_for_session, _register_single, etc.). That works today, but it means the "new internal   
  state must be intentionally classified" promise only holds for writes that go through one of those recognized call sites.                                                                                              
                                                                                                                                                                                                                         
  If someone later adds a new wrapper method on persistence, writes through a helper in a different file, or refactors how persistence is accessed, the scanner quietly stops covering those paths — and we're back to   
  the same silent-failure mode that caused the original $feature_flag_payloads bug, just with a different shape.
                                                                                                                                                                                                                         
  Suggestion: move the enforcement from the test to the persistence class itself. Today, register, register_once, set_property, and set_event_timer each write to this.props[...] directly. If we funnel them through a  
  single private _setProp(key, value) helper, we can add a dev-only check there that warns when a $- or ph_-prefixed key has no policy entry:
                                                                                                                                                                                                                         
 ```
 private _setProp(key, value) {                                  
      if (__DEV__ && (key.startsWith('$') || key.startsWith('ph_')) && !getPersistenceKeyPolicy(key)) {
          logger.warn(`SDK persistence key "${key}" has no policy entry`)                                                                                                                                                
      }
      this.props[key] = value                                                                                                                                                                                            
  }                                                                                                                                                                                                                      
```
  That refactor makes the guarantee shape-independent: any future code path that ends up in persistence gets checked, regardless of what method or helper it came through. User super-props         
  (posthog.register({ plan: 'pro' })) don't trip the check because they don't start with $ or ph_, so backwards compat is preserved.
                                                                                                                                                                                                                         
  Keep the AST test for the "no raw string literals at write sites" rule — that's still useful at review time — but the runtime guard becomes the load-bearing coverage and the hardcoded method list in the test can go 
  away.

File: packages/browser/src/persistence-key-policy.ts [hunk 2]
```diff
@@ -0,0 +1,167 @@
...
+    STORED_GROUP_PROPERTIES_KEY,
+    STORED_PERSON_PROPERTIES_KEY,
+    SURVEYS,
+    SURVEYS_ACTIVATED,
+    USER_STATE,
+    WEB_VITALS_ALLOWED_METRICS,
+    WEB_VITALS_ENABLED_SERVER_SIDE,
+} from './constants'
+
+export type PersistenceKeyExposure = 'event' | 'hidden' | 'derived'
```
ioannisj, line 73: Wonder if for the `derived` case, we also attach/require a transformer function right inside the policy itself. My understanding is that we use this to transform `$enabled_feature_flags` to `$feature/*` so keeping it in the policy is single source of truth + clearer intent?

marandaneto:
> Suggestion: move the enforcement from the test to the persistence class itself.

good point, will do

> we can add a dev-only check there that warns when a $- or ph-prefixed key has no policy entry

not all keys start with those, eg __alias and more, so we cant rely on this

i prefer the AST instead of runtime checks, so we guard against even before releasing it
edit: i improved the AST tests a bit and now we should catch most of the new APIs/refactoring when touching persistence

packages/browser/src/persistence-key-policy.ts
marandaneto: good point, lemme look into it

marandaneto: done

packages/browser/src/persistence-key-policy.ts
marandaneto: | Source | Checked property | Exposure | Expected by policy | Actual usage in non-test source | Example non-test files |
|---|---|---|---|---|---|
| AUTOCAPTURE_DISABLED_SERVER_SIDE | `$autocapture_disabled_server_side` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:535`<br>`posthog/hogql/ai.py:1313`<br>`posthog/management/commands/generate_dlq_events.py:159` |
| EXCEPTION_CAPTURE_ENABLED_SERVER_SIDE | `$exception_capture_enabled_server_side` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:827`<br>`posthog/hogql/ai.py:1574`<br>`posthog/management/commands/generate_dlq_events.py:160` |
| WEB_VITALS_ENABLED_SERVER_SIDE | `$web_vitals_enabled_server_side` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1705`<br>`posthog/hogql/ai.py:2397`<br>`posthog/management/commands/generate_dlq_events.py:157` |
| DEAD_CLICKS_ENABLED_SERVER_SIDE | `$dead_clicks_enabled_server_side` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1814`<br>`posthog/hogql/ai.py:2496`<br>`posthog/management/commands/generate_dlq_events.py:161` |
| PRODUCT_TOURS_ENABLED_SERVER_SIDE | `$product_tours_enabled_server_side` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:396` |
| WEB_VITALS_ALLOWED_METRICS | `$web_vitals_allowed_metrics` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1733`<br>`posthog/hogql/ai.py:2425`<br>`posthog/management/commands/generate_dlq_events.py:194` |
| SESSION_RECORDING_REMOTE_CONFIG | `$session_recording_remote_config` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:414` |
| SESSION_RECORDING_IS_SAMPLED | `$session_is_sampled` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:710`<br>`posthog/hogql/ai.py:1452` |
| SESSION_RECORDING_URL_TRIGGER_ACTIVATED_SESSION | `$session_recording_url_trigger_activated_session` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:918`<br>`posthog/hogql/ai.py:1646` |
| SESSION_RECORDING_FIRST_FULL_SNAPSHOT_TIMESTAMP | `$debug_first_full_snapshot_timestamp` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:390` |
| PERSISTENCE_ACTIVE_FEATURE_FLAGS | `$active_feature_flags` | hidden | not used | used | `ee/hogai/chat_agent/sql/prompts.py:263`<br>`posthog/taxonomy/taxonomy.py:1412`<br>`frontend/src/lib/components/PropertyFilters/utils.ts:596` |
| PERSISTENCE_FEATURE_FLAG_REQUEST_ID | `$feature_flag_request_id` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1442`<br>`posthog/hogql/ai.py:2134`<br>`posthog/management/commands/generate_dlq_events.py:165` |
| PERSISTENCE_OVERRIDE_FEATURE_FLAGS | `$override_feature_flags` | hidden | not used | used | `frontend/src/toolbar/flags/flagsToolbarLogic.ts:247` |
| STORED_PERSON_PROPERTIES_KEY | `$stored_person_properties` | hidden | not used | used | `posthog/cdp/templates/helpers.py:233`<br>`frontend/src/layout/FeaturePreviews/featurePreviewsLogic.tsx:150`<br>`posthog/cdp/site_functions.py:123` |
| SURVEYS | `$surveys` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1795`<br>`posthog/hogql/ai.py:2487` |
| CONVERSATIONS_LEGACY_WIDGET_SESSION_ID | `$conversations_widget_session_id` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1295` |
| CONVERSATIONS_LEGACY_TICKET_ID | `$conversations_ticket_id` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1291` |
| PERSISTENCE_FEATURE_FLAG_EVALUATED_AT | `$feature_flag_evaluated_at` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1447`<br>`posthog/hogql/ai.py:2139` |
| INITIAL_PERSON_INFO | `$initial_person_info` | hidden | not used | used | `posthog/taxonomy/taxonomy.py:1690`<br>`posthog/hogql/ai.py:2382` |
| SDK_DEBUG_REPLAY_MATCHED_RECORDING_TRIGGER_GROUPS | `$sdk_debug_replay_matched_recording_trigger_groups` | event | used | not used | — |
| SDK_DEBUG_REPLAY_TRIGGER_GROUPS_COUNT | `$sdk_debug_replay_trigger_groups_count` | event | used | not used | — |

packages/browser/src/persistence-key-transforms.ts
```diff
@@ -0,0 +1,16 @@
+import type { Properties } from './types'
+
+import { isObject } from '@posthog/core'
+
+export const transformEnabledFeatureFlagsToEventProperties = (value: Properties | undefined): Properties => {
+    if (!isObject(value)) {
+        return {}
+    }
```
dustinbyrne, line 8: Just a thought - it'd be nice to be able to abstract the whole "if the feature flag cache is stale we skip this" logic into the transformer and out of persistence property construction.

Perhaps with dependencies it's a bit tricky (or not worth) to remove completely.

packages/browser/src/persistence-key-policy.ts
```diff
@@ -0,0 +1,169 @@
...
+    SURVEYS,
+    SURVEYS_ACTIVATED,
+    USER_STATE,
+    WEB_VITALS_ALLOWED_METRICS,
+    WEB_VITALS_ENABLED_SERVER_SIDE,
+} from './constants'
+import { transformEnabledFeatureFlagsToEventProperties } from './persistence-key-transforms'
+import type { Properties, Property } from './types'
+
+export type PersistenceKeyExposure = 'event' | 'hidden' | 'derived'
```
dustinbyrne, line 73: Seems worth a comment here explaining what each option is

packages/browser/src/persistence-key-transforms.ts
marandaneto: agree, i tried, i ran into some cyclic issues and gave up so i could at least i have a PoC working
i can give it another try

haacked:
Overall, I think this is a good approach to make things more explicit.

It looks like `SDK_DEBUG_*` events are always emitted. Should those only be emitted when debug = true? Or should there be a new setting to emit them?

marandaneto:
> Overall, I think this is a good approach to make things more explicit.
> 
> It looks like `SDK_DEBUG_*` events are always emitted. Should those only be emitted when debug = true? Or should there be a new setting to emit them?

those come from @PostHog/team-replay 
we'd need to ask them, but i believe this is for snapshots only and not events (otherwise its a bug?)
or @pauldambra 
probably [here](https://github.com/PostHog/posthog-js/blob/d7b296f6ac9a12069548e9713dd6567efc2f582d/packages/browser/src/posthog-core.ts#L1432-L1439)

pauldambra:
we purposefully write `SDK_DEBUG_*` properties onto all events so that we can investigate things  
it's _mostly_ useful for replay but not only  
i'd be open to letting people opt out but they make support much easier

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -824,18 +825,34 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
-        this._checkOverride(SESSION_RECORDING_OVERRIDE_SAMPLING, () => {
-            this.overrideSampling()
-        })
-        this._checkOverride(SESSION_RECORDING_OVERRIDE_LINKED_FLAG, () => {
-            this.overrideLinkedFlag()
-        })
-        this._checkOverride(SESSION_RECORDING_OVERRIDE_EVENT_TRIGGER, () => {
-            this.overrideTrigger('event')
-        })
-        this._checkOverride(SESSION_RECORDING_OVERRIDE_URL_TRIGGER, () => {
```
pauldambra, line 836: oh, i'd be super uncomfortable changing these  
they protect against invalid states in the recorder  
we've frozen people's sites where we've got in a loop  
it would be way safer to leave them alone!

marandaneto:
> we purposefully write `SDK_DEBUG_*` properties onto all events so that we can investigate things it's _mostly_ useful for replay but not only i'd be open to letting people opt out but they make support much easier

what if this is only for `$snapshot` events, is that enough?

packages/browser/src/persistence-key-policy.ts
pauldambra: things like `$autocapture_disabled_server_side` are pretty small  
(if this is about storage)  
and explicitly there to make our lives easier  
we don't use them a lot but when you need them you really need them

pauldambra (dismissed):
i have only ever clicked request changes a handful of times in my life :)  
  
i'd be _super_ uncomfortable with changing the replay persistence properties  
  
(but i only have half a brain here and i might be missing what's happening)

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: mm its just a refactor, nothing changed really
unless i missed something that you can see?

packages/browser/src/persistence-key-policy.ts
marandaneto: i dont want to remove anything that is useful
i just dont know all of them so its hard to tell if they are intentionally part of events or leaked as some other config
i asked AI to check 1 by 1 matching this repo and the posthog repo (eg part of the taxonomy, frontend, filters, insights etc)

pauldambra: they session replay ones to do with triggers are essential code-flow stuff  
\_and_ serve as debug signals  
  
the others are all debug signals   
given i've never touched conversations, flags, and surveys code personally so I wouldn't know about those

dustinbyrne: `$feature_flag_request_id`, `$feature_flag_evaluated_at`, `$override_feature_flags` are used for flags troubleshooting/support. Likely useful to keep.

Not sure about `$active_feature_flags`, but I believe @haacked was looking into this recently.

cc @PostHog/team-feature-flags

haacked: Yeah, I've been looking into `$active_feature_flags`. Unfortunately, some customers do have insights that reference that field. I think it was sort of a hacky experiment approach rather than using our built-in approach, but I can't be sure.

`$session_recording_remote_config` has a lot of data. Should that still be sent? Also `$surveys_activated`? I may have been a bit overzealous in https://github.com/PostHog/posthog-js/pull/3392 when I removed them.

haacked: `$feature_flag_request_id`, `$feature_flag_evaluated_at`, `$override_feature_flags` <- These fields don't take up a lot of data anyways.

pauldambra:
> what if this is only for $snapshot events, is that enough?

nope they go through a totally different ingestion pipeline
it has to be on event events :)

marandaneto:
ok so i think the safest bet here is to keep all properties as is, and then remove them (by changing the sdk prop policy) one by one once you know this is leaking to events and its not used/it was not intentional
so this PR can at least guard against leaking new properties to all events that werent intentional

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: We added AST tests that enforce:

 - persistence writes/removes use constant-backed keys
 - policy coverage stays explicit
 - persistence mutation sites remain easy to audit

 If _checkOverride does unregister(key) internally, the AST scan only sees a method parameter named key, not the
 actual constant like:

 - SESSION_RECORDING_OVERRIDE_SAMPLING
 - SESSION_RECORDING_OVERRIDE_LINKED_FLAG

 So the key classification becomes opaque to the test

marandaneto:
> ok so i think the safest bet here is to keep all properties as is, and then remove them (by changing the sdk prop policy) one by one once you know this is leaking to events and its not used/it was not intentional so this PR can at least guard against leaking new properties to all events that werent intentional

Double-checked: current policy matches legacy behavior.

 ### Result

 Compared against origin/main’s old PERSISTENCE_RESERVED_PROPERTIES:

 - Exact keys: no mismatches
 - Prefix keys: no mismatches
 - Derived behavior: still matches legacy behavior for ENABLED_FEATURE_FLAGS
     - stored as $enabled_feature_flags
     - emitted as $feature/<flag>
     - stale-cache gating still preserved

packages/browser/src/persistence-key-policy.ts
marandaneto: see https://github.com/PostHog/posthog-js/pull/3425#issuecomment-4294327075

dustinbyrne (approved):
nice!

packages/browser/src/posthog-persistence.ts
```diff
@@ -160,17 +160,19 @@ export class PostHogPersistence {
...
-                if (!this._isFeatureFlagCacheStale()) {
-                    const keys = Object.keys(v)
-                    for (let i = 0; i < keys.length; i++) {
-                        p[`$feature/${keys[i]}`] = v[keys[i]]
-                    }
+            const policy = getPersistenceKeyPolicy(k)
+
+            if (policy?.exposure === 'derived') {
+                if (k === ENABLED_FEATURE_FLAGS && this._isFeatureFlagCacheStale()) {
+                    return
```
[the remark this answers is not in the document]
marandaneto: https://github.com/PostHog/posthog-js/pull/3425#discussion_r3111963624

packages/browser/src/persistence-key-policy.ts
```diff
@@ -0,0 +1,178 @@
...
+    [SDK_DEBUG_REPLAY_TRIGGER_GROUPS_COUNT]: { exposure: 'event' },
+    [SDK_DEBUG_REPLAY_URL_TRIGGER_STATUS]: { exposure: 'event' },
+    [SESSION_RECORDING_START_REASON]: { exposure: 'event' },
+}
+
+const PERSISTENCE_KEY_PREFIX_POLICY: Array<[string, PersistenceKeyPolicyEntry]> = [
+    [SESSION_RECORDING_TRIGGER_V2_GROUP_EVENT_PREFIX, { exposure: 'event' }],
+    [SESSION_RECORDING_TRIGGER_V2_GROUP_URL_PREFIX, { exposure: 'event' }],
+    [SESSION_RECORDING_TRIGGER_V2_GROUP_SAMPLING_PREFIX, { exposure: 'event' }],
+]
```
ksvat, line 163: these are just for tracking internal v2 trigger state (debug v2 trigger info tracked in SDK_DEBUG_REPLAY_MATCHED_RECORDING_TRIGGER_GROUPS) so these can be hidden?  
  
the future support-hero in me wants every piece of data available, but this also cooould get noisy (if there are 5 trigger groups, thered be 3 prefix * 5 trigger group properties for this)

marandaneto: changed them to `hidden`

packages/browser/playwright/fixtures/ingestion.ts
```diff
@@ -27,6 +27,23 @@ export const testIngestion = testPostHog.extend<{}, { ingestion: IngestionPage }
     ],
 })
 
+const dedupeEventsByUuid = (events: CaptureResult[]): CaptureResult[] => {
```
marandaneto, line 30: this is fixing CI flakiness, should be another PR but I cant get this to green on this PR rn

.github/workflows/testcafe.yml
```diff
@@ -63,6 +63,6 @@ jobs:
 
       - name: Check ${{ matrix.name }} events
         if: steps.is-affected.outputs.is-affected == 'true'
-        timeout-minutes: 10
+        timeout-minutes: 30
```
marandaneto, line 66: because of ingestion lag, lets align with the general e2e tests that are 30min as well

PR #3426, opened 2026-04-20
fix: trim whitespace from API keys and host config

### LINKED ISSUE #3424: SDK should trim or validate API key — trailing whitespace/newline causes silent event loss
Hey there, frustrating user error here. I was setting up a new project on posthog, had an existing project functioning fine, went back and forth a lot with Claude, and the PostHog AI agent, turns out the issue was because my environment vars had a newline character on the end. This worked on localhost, but didn't work in production with Vercel.

<details><summary>Brief timeline of me debugging</summary>

## System

- **Stack**: Next.js 16.2.3, posthog-js 1.367.0, @posthog/next 0.1.0
- **Deployment**: Vercel (Fluid Compute, sfo1 region)
- **PostHog**: Cloud US (us.i.posthog.com)
- **Proxy path**: /ingest/\* → us.i.posthog.com (manual fetch in src/proxy.ts)

## Timeline

- PostHog captures localhost events but not (appsite).com production events — both return 200
- Investigated proxy setup: rewrites in next.config.ts forward to PostHog
- Discovered next.config.ts rewrites strip POST bodies on Vercel — events arrive empty
- Switched to NextResponse.rewrite() in proxy.ts (middleware) — same result
  - Installed @posthog/next official package — uses NextResponse.rewrite() internally, same problem
- Replaced with manual fetch() proxy that reads request.arrayBuffer() and forwards body explicitly
- Production requests now fail with ERR_CONTENT_DECODING_FAILED — new error
  - fetch() auto-decompresses gzip responses but forwards the original
    content-encoding: gzip header — browser tries to decompress already-decompressed data
- Stripped content-encoding, content-length, transfer-encoding headers from proxied response — decoding errors resolved
- Added debug: true to PostHog init, pushed to production
  - Console shows PostHog initializing correctly, sending events, getting 200 back — but still no events in dashboard
- Inspected POST request payload in DevTools Network tab
  - properties.token field shows "phc\_...2mS\n" — trailing newline in the API key
  - Ran vercel env pull --environment=production | od -c — confirmed literal \n embedded in NEXT_PUBLIC_POSTHOG_KEY Vercel env var
- PostHog silently accepts the malformed token (200) but never processes the events
  - Filed bead ba-c5q2 documenting the discovery chain
  - Filed bead ba-eurp for build-time key format validation
  - Filed PostHog upstream issue #3424 requesting token trim or validation warning

</details> 

## Problem

When `NEXT_PUBLIC_POSTHOG_KEY` contains a trailing newline (e.g. `phc_abc123\n`), the SDK initializes successfully, sends events, and receives HTTP 200 responses — but **no events ever appear in the PostHog dashboard**. There is zero indication anything is wrong.

This is easy to trigger accidentally when pasting env vars in hosting platforms like Vercel, which don't strip trailing whitespace from values.

## Discovery chain

We spent hours debugging this. The only way we found it was:

1. Enabled `debug: true` in PostHog init
2. Inspected the POST request body in browser DevTools Network tab
3. Noticed `"token": "phc_...2mS\n"` — the trailing `\n` was visible in the JSON payload
4. Confirmed via `vercel env pull --environment=production | od -c`

## Expected behavior

Either:

1. **Trim the API key** — `apiKey.trim()` before use (preferred, zero friction)
2. **Warn at init time** — if the key doesn't match `/^phc_[a-zA-Z0-9]+$/`, log a clear warning: `"PostHog API key contains unexpected characters (whitespace/newline). This will cause events to be silently dropped."`

Option 1 is a one-line fix and would prevent this entire class of bugs. The `@posthog/next` package already warns if the key doesn't start with `phc_` — extending that to catch whitespace would be valuable.

## Environment

- `posthog-js`: 1.367.0
- `@posthog/next`: 0.1.0
- Next.js 16.2.3 on Vercel
- PostHog Cloud US (`us.i.posthog.com`)

## Reproduction

```js
posthog.init("phc_validKey\n", {
  api_host: "/ingest",
});
// Events send, return 200, never appear in dashboard
```

marandaneto:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/3424

Whitespace-padded config values were preserved in a couple of SDK entry points inside this monorepo. That meant callers could accidentally initialize `@posthog/core` with an API key or host containing spaces/newlines, or configure `@posthog/plugin-utils` with a whitespace-polluted `personalApiKey`/`host`, and the SDK would use those raw values.

## Changes
- trim `apiKey` and `host` in `@posthog/core` before storing them
- default a blank `host` back to `https://us.i.posthog.com` after trimming
- trim `personalApiKey` and `host` in `@posthog/plugin-utils`
- reject whitespace-only `personalApiKey` the same way we already reject empty values when sourcemaps are enabled
- add regression tests
- add a patch changeset for `@posthog/core` and `@posthog/plugin-utils`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [x] @posthog/convex
- [x] @posthog/next
- [x] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


turnipdabeets (approved):

packages/ai/src/otel/exporter.ts
```diff
@@ -43,14 +53,15 @@ export interface PostHogTraceExporterOptions {
  */
 export class PostHogTraceExporter extends OTLPTraceExporter {
   constructor(options: PostHogTraceExporterOptions) {
-    if (!options.apiKey) {
+    const apiKey = normalizeApiKey(options.apiKey)
```
hpouillot, line 57: this will throw a trim is not a function on undefined api keys

cat-ph (approved):
🚀

packages/plugin-utils/src/config.ts
```diff
@@ -48,7 +59,8 @@ export interface ResolveConfigOptions {
 
 export function resolveConfig(options: PluginConfig, resolveOptions?: ResolveConfigOptions): ResolvedPluginConfig {
     const projectId = options.projectId ?? options.envId
-    const host = options.host ?? 'https://us.i.posthog.com'
+    const personalApiKey = normalizeApiKey(options.personalApiKey)
```
hpouillot, line 62: same here, I think we want to keep the error message explicit since the api key could be missing and it should not be required unless sourcemap upload is enabled.

packages/ai/src/otel/exporter.ts
marandaneto: this is defined as non optional https://github.com/PostHog/posthog-js/blob/c7b470f9eb6ec02ad16d43723039162fc15868e5/packages/ai/src/otel/exporter.ts#L21

packages/plugin-utils/src/config.ts
marandaneto: same here https://github.com/PostHog/posthog-js/blob/c7b470f9eb6ec02ad16d43723039162fc15868e5/packages/plugin-utils/src/config.ts#L17

packages/ai/src/otel/exporter.ts
marandaneto: with TS is a bit harder to pass something null/unfedined, but much easier in TS, indeed

packages/plugin-utils/src/config.ts
hpouillot: Yes, we need to update the type to be more specific. apiKey is only required when sourcemaps upload is enabled, there is a runtime check for this. 

Usually users have this in their config and just remove the env variable in dev which leads to undefined values.
```apiKey: process.env.MY_KEY!```

marandaneto: done

hpouillot (approved):

trillium:
Thanks all, having this would have saved me a few hours of troubleshooting and I appreciate the speed this was addressed!

Issue #3427, opened 2026-04-20
Cannot find module `...` in @posthog/next with vitest.


IcaroG:
## Bug description

When running tests with vitest with @posthog/next being used, this error is being thrown:

```
Error: Cannot find module '.../@posthog/next/dist/client/PostHogPageView'
imported from .../@posthog/next/dist/index.client.js
```

## How to reproduce

https://codesandbox.io/p/github/IcaroG/posthog-next-vitest-repro/main?import=true

Run `pnpm install` and `pnpm test`

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

Disclaimer: the minimal reproducible repository was created with AI, but I've manually tested it. There are more findings in the README.

dustinbyrne:
Hey @IcaroG, thanks for the issue and the clear reproduction. I've opened a [PR](https://github.com/PostHog/posthog-js/pull/3428) to resolve this.

If you give it a try on a larger project let me know how it goes!

PR #3428, opened 2026-04-20
fix: make @posthog/next importable in Node ESM environments

dustinbyrne:
## Problem

`@posthog/next` could not be imported from Vitest and other Node ESM environments because the published package mixed native ESM output with specifiers that only work when a bundler rewrites them. This showed up in #3427 and in a fresh packed-tarball repro.

## Changes

- make `@posthog/next` publish runtime-safe ESM specifiers by adding explicit `.js` extensions to internal relative imports/exports
- switch runtime imports to ESM-safe entrypoints where needed (`next/navigation.js`, `next/router.js`, `next/headers.js`, `next/server.js`)
- switch React hooks/context imports from `posthog-js/react` to `@posthog/react`
- update the Next package tests/Jest config to work with the new runtime-safe source imports
- add a publish-time guard that fails the build if emitted `dist` files regress to unsafe ESM specifiers
- add a changeset for `@posthog/next`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/next/scripts/check-relative-esm-specifiers.mjs
```diff
@@ -0,0 +1,59 @@
...
+const distDir = new URL('../dist/', import.meta.url)
+const allowedExtensions = /\.(?:js|mjs|cjs|json)$/
+const forbiddenBareSpecifiers = new Set([
+    'next/navigation',
+    'next/router',
+    'next/headers',
+    'next/server',
+    'posthog-js/react',
+])
+const specifierPattern = /\b(?:import|export)\s+(?:[^'"\n]*?\sfrom\s+)?['"]([^'"]+)['"]|\bimport\(\s*['"]([^'"]+)['"]\s*\)/g
```
dustinbyrne, line 14: I don't love that we're adding this script instead of build-time checks but:
- it _does_ work as a validation that we're native ESM compatible
- switching compiler options to change the module to `nodenext` also requires we change module resolution and that led to deeper issues with resolving `posthog-js`

So, I suppose it's fine for now. I'd like to circle back and address this properly, but I don't think we need to delay this fix.

marandaneto (approved):

PR #3431, opened 2026-04-21
fix(ai): read tool-call input field in vercel non-streaming output

### LINKED ISSUE #3011: Tool call arguments missing in Vercel AI SDK v6 non-streaming output
## Bug description

When using `withTracing` with the Vercel AI SDK v6 and `generateText` (non-streaming), tool call arguments are logged as empty `"{}"` instead of the actual arguments. This is because `mapVercelOutput` looks for `item.args` or `item.arguments`, but Vercel AI SDK v6 uses `item.input` for tool call arguments.

### Environment

- `@posthog/ai`: 7.8.2
- `ai` (Vercel AI SDK): 6.0.57
- Node.js: 22.x

```typescript
import { generateText, tool } from "ai";
import { createOpenAI } from "@ai-sdk/openai";
import { z } from "zod";

const phClient = new PostHog("phc_xxx", { host: "https://eu.i.posthog.com" });
const openai = createOpenAI();

const myTool = tool({
  description: "A test tool",
  parameters: z.object({
    message: z.string(),
  }),
  execute: async (input) => input,
});

await generateText({
  model: withTracing(openai("gpt-4o"), phClient, {}),
  tools: { myTool },
  prompt: "Call myTool with message 'hello world'",
});
```

### Expected behavior

PostHog event should contain the tool call with arguments:
```json
{
  "type": "tool-call",
  "id": "call_xxx",
  "function": {
    "name": "myTool",
    "arguments": "{\"message\":\"hello world\"}"
  }
}
```

### Actual behavior

PostHog output event contains empty arguments:
```json
{
  "type": "tool-call",
  "id": "call_xxx",
  "function": {
    "name": "myTool",
    "arguments": "{}"
  }
}
```

### Root cause

In `src/vercel/middleware.ts`, the `mapVercelOutput` function (around line 208) looks for the wrong property:

```javascript
if (item.type === 'tool-call') {
  return {
    type: 'tool-call',
    id: item.toolCallId,
    function: {
      name: item.toolName,
      arguments: item.args || JSON.stringify(item.arguments || {})  // ❌ Wrong properties
    }
  };
}
```

Vercel AI SDK v6 uses `item.input` for tool call arguments, not `item.args` or `item.arguments`, see [migration guide](https://ai-sdk.dev/docs/migration-guides/migration-guide-5-0#tool-property-changes-argsresult--inputoutput).

**Note**: The streaming path in `doStream` correctly uses `toolCall.input` (line 583), and `mapVercelPrompt` also correctly uses `c.input` (line 139). Only `mapVercelOutput` has this bug.

### Suggested fix

```javascript
if (item.type === 'tool-call') {
  return {
    type: 'tool-call',
    id: item.toolCallId,
    function: {
      name: item.toolName,
      arguments: item.input || item.args || JSON.stringify(item.arguments || {})  // ✅ Check item.input first
    }
  };
}
```

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

richardsolomou:
## Problem

When using `withTracing` with Vercel AI SDK v6 and `generateText` (non-streaming), tool call arguments were logged as empty `"{}"` in `$ai_output_choices` instead of the actual arguments. Affects anyone tracking tool-calling agents with `@posthog/ai` + `ai@^6`.

Root cause: `mapVercelOutput` in `packages/ai/src/vercel/middleware.ts` looked for `item.args` / `item.arguments`, but AI SDK v5+ emits tool-call content parts with `item.input` (a stringified JSON, per `LanguageModelV{2,3}ToolCall` in `@ai-sdk/provider`). The streaming path and `mapVercelPrompt` already read `input` correctly — only the non-streaming output mapping was wrong.

Fixes #3011.

## Changes

- `mapVercelOutput` now reads `item.input` first, falling back to `item.args` / `item.arguments` for older AI SDK versions. When `input` is an object (non-conforming providers), it is `JSON.stringify`d to produce the string shape `$ai_output_choices` expects.
- Added a parameterised test covering both string-input (per spec) and object-input (defensive) cases for V3 non-streaming tool calls.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

---

### DIFF
--- a/packages/ai/src/vercel/middleware.ts
+++ b/packages/ai/src/vercel/middleware.ts
@@ -200,7 +200,8 @@ const mapVercelOutput = (result: LanguageModelContent[]): PostHogInput[] => {
       return { type: 'text', text: truncate(item.text) }
     }
     if (item.type === 'tool-call') {
-      const rawArgs = (item as any).input ?? (item as any).args ?? (item as any).arguments ?? {}
+      const toolCall = item as { input?: unknown; args?: unknown; arguments?: unknown }
+      const rawArgs = toolCall.input ?? toolCall.args ?? toolCall.arguments ?? {}
       return {
         type: 'tool-call',
         id: item.toolCallId,
--- a/packages/ai/src/vercel/middleware.ts
+++ b/packages/ai/src/vercel/middleware.ts
@@ -200,12 +200,13 @@ const mapVercelOutput = (result: LanguageModelContent[]): PostHogInput[] => {
       return { type: 'text', text: truncate(item.text) }
     }
     if (item.type === 'tool-call') {
+      const rawArgs = (item as any).input ?? (item as any).args ?? (item as any).arguments ?? {}
       return {
         type: 'tool-call',
         id: item.toolCallId,
         function: {
           name: item.toolName,
-          arguments: (item as any).args || JSON.stringify((item as any).arguments || {}),
+          arguments: typeof rawArgs === 'string' ? rawArgs : JSON.stringify(rawArgs),
         },
       }
     }
--- a/packages/ai/tests/vercel.test.ts
+++ b/packages/ai/tests/vercel.test.ts
@@ -373,6 +373,59 @@ describe('Vercel AI SDK - Dual Version Support', () => {
         },
       ])
     })
+
+    it.each([
+      ['string input (per spec)', '{"message":"hello world"}'],
+      ['object input (defensive)', { message: 'hello world' }],
+    ])('should handle non-streaming tool calls with %s (AI SDK v6)', async (_label, input) => {
+      const baseModel: LanguageModelV3 = {
+        specificationVersion: 'v3' as const,
+        provider: 'openai',
+        modelId: 'gpt-4o',
+        supportedUrls: {},
+        doGenerate: jest.fn().mockResolvedValue({
+          content: [
+            {
+              type: 'tool-call',
+              toolCallId: 'call_123',
+              toolName: 'myTool',
+              input,
+            },
+          ],
+          usage: v3TokenUsage(10, 5),
+          response: { modelId: 'gpt-4o' },
+          providerMetadata: {},
+          finishReason: { unified: 'tool-calls' as const, raw: undefined },
+          warnings: [],
+        }),
+        doStream: jest.fn(),
+      }
+
+      const model = withTracing(baseModel, mockPostHogClient, {
+        posthogDistinctId: 'test-user',
+        posthogTraceId: 'test-v3-tool-input',
+      })
+
+      await model.doGenerate({
+        prompt: [{ role: 'user' as const, content: [{ type: 'text' as const, text: 'Call myTool' }] }],
+      } as any)
+
+      expect(mockPostHogClient.capture).toHaveBeenCalledTimes(1)
+      const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
+
+      expect(captureCall[0].properties.$ai_output_choices).toEqual([
+        {
+          role: 'assistant',
+          content: [
+            {
+              type: 'tool-call',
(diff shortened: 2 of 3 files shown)


packages/ai/src/vercel/middleware.ts
```diff
@@ -200,12 +200,13 @@ const mapVercelOutput = (result: LanguageModelContent[]): PostHogInput[] => {
       return { type: 'text', text: truncate(item.text) }
     }
     if (item.type === 'tool-call') {
+      const rawArgs = (item as any).input ?? (item as any).args ?? (item as any).arguments ?? {}
```
carlos-marchal-ph: The three `as any` casts could collapse into one narrow assertion, since `item` is already narrowed to the tool-call variants in this branch and V3 already types `input` as `unknown`:

```ts
const toolCall = item as { input?: unknown; args?: unknown; arguments?: unknown }
const rawArgs = toolCall.input ?? toolCall.args ?? toolCall.arguments ?? {}
```

packages/ai/tests/vercel.test.ts
```diff
@@ -373,6 +373,59 @@ describe('Vercel AI SDK - Dual Version Support', () => {
         },
       ])
     })
+
+    it.each([
+      ['string input (per spec)', '{"message":"hello world"}'],
```
carlos-marchal-ph: The `args` and `arguments` fallbacks have no test coverage, so if they get dropped in a future refactor the regression would be silent. Worth a third `it.each` row feeding something like `args: '{"a":1}'` on the mock content part?

packages/ai/tests/vercel.test.ts
```diff
@@ -373,6 +373,59 @@ describe('Vercel AI SDK - Dual Version Support', () => {
...
+      }
+
+      const model = withTracing(baseModel, mockPostHogClient, {
+        posthogDistinctId: 'test-user',
+        posthogTraceId: 'test-v3-tool-input',
+      })
+
+      await model.doGenerate({
+        prompt: [{ role: 'user' as const, content: [{ type: 'text' as const, text: 'Call myTool' }] }],
+      } as any)
```
carlos-marchal-ph: Could this use `LanguageModelV3CallOptions` instead of `as any`? Would keep the test honest to the public type surface if the shape changes upstream.

carlos-marchal-ph (approved):

packages/ai/src/vercel/middleware.ts
richardsolomou: Done in d4736d58 — collapsed into a single `{ input?: unknown; args?: unknown; arguments?: unknown }` assertion.

packages/ai/tests/vercel.test.ts
richardsolomou: Good catch — added rows for `args: "<string>"` and `arguments: { ... }` in d4736d58, so both fallback paths are now covered.

packages/ai/tests/vercel.test.ts
richardsolomou: Switched to `LanguageModelV3CallOptions` in d4736d58.

PR #3432, opened 2026-04-21
feat: support addTracingHeaders in react-native sdk

### LINKED ISSUE #3196: Using __add_tracing_headers with React Native SDK doesnt add the sessions to each request.
## Bug description

Using https://posthog.com/docs/llm-analytics/link-session-replay#implementation with React Native SDK doesnt add the sessions to each request.

## How to reproduce

```js
<PostHogProvider apiKey="...." options={{
                        host: "https://eu.posthog.com",
                        __add_tracing_headers: ['localhost', 'localhost:8000', '127.0.0.1', '127.0.0.1:8000'],
...
```

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

richardsolomou:
## Problem

Fixes #3196.

`__add_tracing_headers` is documented for linking LLM traces to session replays (see https://posthog.com/docs/llm-analytics/link-session-replay), but the option was silently ignored in the React Native SDK — the feature was only implemented in the browser SDK, which patches `window.fetch`/`window.XMLHttpRequest`. Users passing the option in `PostHogProvider` options saw no session headers on outgoing requests.

Per review consensus on this PR, the `__` prefix on the browser option was a legacy internal-naming convention. Since the option is a public API surface, this PR also renames it to `addTracingHeaders` on the browser SDK and types package (keeping `__add_tracing_headers` there as a deprecated alias), and lifts both the option and the patch implementation into `@posthog/core` so any SDK extending it inherits them automatically.

## Changes

**`@posthog/core`**
- Add `addTracingHeaders?: string[]` to `PostHogCoreOptions` — any SDK extending it (posthog-react-native today, posthog-web next) picks up the config field for free.
- Add `packages/core/src/tracing-headers.ts` exporting `patchFetchForTracingHeaders(client, hostnames)` and a minimal `TracingHeadersClient` interface (`{ getDistinctId(): string, getSessionId(): string }`) so any client satisfying that contract can use it.
- The wrapped `globalThis.fetch` is tagged with a non-enumerable marker so re-invocations unwrap the previous patch before rewrapping — preventing patches from stacking across HMR, tests, or multiple PostHog instances.

**React Native**
- Wire up `patchFetchForTracingHeaders` from `@posthog/core` during `initAfterStorage` when `addTracingHeaders` is set.
- No separate option declaration on `PostHogOptions` — it's inherited from `PostHogCoreOptions`.
- Caller-provided headers and init fields are preserved (plain object, `Headers` instance, or `Request` input); IDs are read per-request so identify/session rotation are respected.
- `X-POSTHOG-WINDOW-ID` is browser-specific and skipped.

**Browser + types**
- Add `addTracingHeaders?: string[]` to `PostHogConfig`; `__add_tracing_headers` marked `@deprecated`.
- `TracingHeaders` extension reads `addTracingHeaders` first, falls back to `__add_tracing_headers` so existing browser users keep working.

**Tests (14 in `packages/core/src/__tests__/tracing-headers.spec.ts`)**
- Covers: matching/non-matching hostnames across string / URL / Request inputs, plain-object headers, `Headers`-instance `init.headers`, `Request`-input header preservation, dynamic id reads, invalid URLs, #3196 regression (the reporter's exact hostname list), fetch-rejection propagation, `restore()` behavior, missing-fetch no-op, and idempotent re-initialisation.
- Types snapshot updated to include `addTracingHeaders`.

### Caveats (matching browser SDK behavior)

- Port-qualified entries like `localhost:8000` never match because `URL.hostname` strips ports; the bare `localhost` / `127.0.0.1` entries in the reporter's list still match. Same as the browser implementation.
- XHR is not patched. The browser's XHR patch doesn't actually apply headers (creates a `Request` and discards it), and modern LLM/HTTP clients in RN use `fetch`.
- Idempotency only unwraps our own immediate predecessor. If another library wraps `fetch` between two PostHog inits, their wrapper has no marker, so we treat it as the "original" and the earlier PostHog patch stays live underneath. Considered out of scope — a comment in `tracing-headers.ts` documents the assumption.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

Also bumps `@posthog/core` (minor) for the new `patchFetchForTracingHeaders` export and `addTracingHeaders` option on `PostHogCoreOptions`.

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (two files: one for the rename, one for the RN feature)

---


packages/react-native/src/posthog-rn.ts
```diff
@@ -292,6 +312,10 @@ export class PostHog extends PostHogCore {
       void this.persistAppVersion()
 
       void this.startSessionReplay(options, cachedRemoteConfig ?? undefined)
+
+      if (options?.__add_tracing_headers && options.__add_tracing_headers.length > 0) {
+        patchFetchForTracingHeaders(this, options.__add_tracing_headers)
```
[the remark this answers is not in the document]
richardsolomou: Addressed in d6278d0c, though taking a different approach: I made `patchFetchForTracingHeaders` idempotent via a non-enumerable marker on the wrapped function, rather than storing a restore on the instance.

One nit on the original claim: the leaked patch wouldn't actually "inject headers onto unintended requests" — the hostname filter ensures only the configured hosts ever get headers. The real risk is patches stacking across HMR/test re-inits, which the marker-based unwrap-before-rewrap now handles. Latest instance wins, no stacking.

packages/react-native/test/tracing-headers.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+  })
+
+  it('does not add tracing headers for non-matching hostnames', async () => {
+    restore = patchFetchForTracingHeaders(makeInstance('d-1', 's-1'), ['api.example.com'])
+
+    await globalAny.fetch('https://other.example.com/thing')
+
+    const [, init] = mockFetch.mock.calls[0]
+    expect(init).toBeUndefined()
+  })
```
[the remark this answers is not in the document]
richardsolomou: Done in d6278d0c — folded the symmetric matching/non-matching hostname cases into an `it.each` table.

packages/react-native/src/posthog-rn.ts
```diff
@@ -105,6 +106,25 @@ export interface PostHogOptions extends PostHogCoreOptions {
...
+   * to requests whose hostname matches one of the entries in the list.
+   *
+   * @example
+   * ```ts
+   * new PostHog('<key>', {
+   *   __add_tracing_headers: ['api.example.com'],
+   * })
+   * ```
+   */
+  __add_tracing_headers?: string[]
```
marandaneto: why does it start with a `__` prefix?

packages/core/src/__tests__/tracing-headers.spec.ts
```diff
@@ -0,0 +1,156 @@
+import { patchFetchForTracingHeaders } from '../src/tracing-headers'
+import type { PostHog } from '../src/posthog-rn'
+
+describe('patchFetchForTracingHeaders', () => {
```
marandaneto, line 3: have you tested this works on android and ios running on emulators/real devices?
just wondering if this workas as intended since on react native the requests are delegated to the native layer anyway

packages/react-native/src/posthog-rn.ts
richardsolomou: Matching the browser SDK convention where the same option is already `__add_tracing_headers` (see `posthog-config.ts:1555`). The name stays identical so users copying the config from the [link-session-replay docs](https://posthog.com/docs/llm-analytics/link-session-replay) don't have to know they're on RN vs. web. Happy to rename it for the RN surface if you'd prefer, but I'd rather diverge only if there's a reason to.

packages/core/src/__tests__/tracing-headers.spec.ts
richardsolomou: Nope, only Jest, I don't have an emulator set up on this machine atm. But the native-layer concern shouldn't matter here: RN's `global.fetch` is a JS-side polyfill (`whatwg-fetch`) that builds headers in JS before handing off to the native networking module. Patching `globalThis.fetch` modifies `init.headers` before the native layer sees the request, so the native code just ships whatever headers it was given.

packages/react-native/src/posthog-rn.ts
marandaneto: its not very common to do this, usually config named with __ means internal/private, so i wonder if that was the reason to name __add_tracing_headers for JS (i think it was a hackathon project), since we want to make this public and users should set it, i dont see why it'd have a prefix with __

packages/core/src/__tests__/tracing-headers.spec.ts
marandaneto: yeah i dont think its gonna be an issue, but i'd validate to make 100% sure
@ioannisj or @turnipdabeets could help here

packages/react-native/src/tracing-headers.ts
```diff
@@ -0,0 +1,75 @@
...
+const parseHostname = (url: string): string | undefined => {
+  try {
+    return new URL(url).hostname
+  } catch {
+    return undefined
+  }
+}
+
+const shouldAddHeaders = (url: string, hostnames: string[]): boolean => {
+  if (!isArray(hostnames)) {
```
carlos-marchal-ph: The `isArray` check is dead given the TypeScript type, since the public option is typed as `string[] | undefined` and posthog-rn.ts already filters empty. The browser SDK keeps it for a legacy boolean case, but there's no legacy here. Any reason to keep it?

packages/react-native/src/tracing-headers.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  // If we (or a previous PostHog instance) already patched fetch, unwrap so the latest
+  // instance's hostname list and session/distinct ids take effect and patches don't stack.
+  const originalFetch: FetchFn = currentFetch[PATCH_MARKER]?.original ?? currentFetch
+
+  const wrappedFetch: PatchedFetch = async function (input, init) {
+    try {
+      const urlString =
+        typeof input === 'string' ? input : input instanceof URL ? input.toString() : (input as Request)?.url
```
carlos-marchal-ph: Can we narrow with `instanceof Request` instead of the `as Request` cast?

```suggestion
      const urlString =
        typeof input === 'string'
          ? input
          : input instanceof URL
            ? input.toString()
            : input instanceof Request
              ? input.url
              : undefined
```

packages/react-native/src/tracing-headers.ts
```diff
@@ -0,0 +1,75 @@
...
+        const headers = new Headers(init?.headers ?? (input instanceof Request ? input.headers : undefined))
+        const distinctId = instance.getDistinctId()
+        const sessionId = instance.getSessionId()
+        if (distinctId) {
+          headers.set(DISTINCT_ID_HEADER, distinctId)
+        }
+        if (sessionId) {
+          headers.set(SESSION_ID_HEADER, sessionId)
+        }
+        init = { ...(init ?? {}), headers }
```
carlos-marchal-ph: Reassigning the `init` parameter reads a bit oddly, would you mind binding to a new `const initWithHeaders` and passing that to `originalFetch`?

packages/core/src/tracing-headers.ts
```diff
@@ -0,0 +1,75 @@
...
+export const patchFetchForTracingHeaders = (instance: PostHog, hostnames: string[]): (() => void) => {
+  const globalAny = globalThis as unknown as { fetch?: PatchedFetch }
+  const currentFetch = globalAny.fetch
+  if (!isFunction(currentFetch)) {
+    return () => {}
+  }
+
+  // If we (or a previous PostHog instance) already patched fetch, unwrap so the latest
+  // instance's hostname list and session/distinct ids take effect and patches don't stack.
+  const originalFetch: FetchFn = currentFetch[PATCH_MARKER]?.original ?? currentFetch
```
carlos-marchal-ph, line 62: Worth a short comment here noting the assumption: we only unwrap our own immediate predecessor. If another library wraps fetch between two PostHog inits, their wrapper has no `PATCH_MARKER` and we treat it as the 'original', so the earlier PostHog patch stays live in the chain and headers get written twice. Probably out of scope to fix, but would save a future reader the puzzlement.

packages/react-native/test/tracing-headers.spec.ts
```diff
@@ -0,0 +1,156 @@
+import { patchFetchForTracingHeaders } from '../src/tracing-headers'
+import type { PostHog } from '../src/posthog-rn'
+
+describe('patchFetchForTracingHeaders', () => {
+  const globalAny = globalThis as any
```
carlos-marchal-ph: Can we avoid `any` here?

```suggestion
  const globalAny = globalThis as unknown as { fetch: typeof fetch }
```

packages/core/src/__tests__/tracing-headers.spec.ts
turnipdabeets: @richardsolomou happy to pair and validate. This might a good opportunity to help each other since I'm new to this as well.

packages/react-native/src/posthog-rn.ts
carlos-marchal-ph: Agreed, I'd rather just deprecate the `__` from the browser SDK and start accepting it unprefixed if it's meant to be a public interface.

packages/core/src/__tests__/tracing-headers.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    expect(secondHeaders.get('X-POSTHOG-SESSION-ID')).toBe('s-2')
+  })
+
+  it('ignores invalid URLs without throwing', async () => {
+    restore = patchFetchForTracingHeaders(makeInstance('d-1', 's-1'), ['api.example.com'])
+
+    await expect(globalAny.fetch('not a url')).resolves.toEqual({ status: 200 })
+    const [, init] = mockFetch.mock.calls[0]
+    expect(init).toBeUndefined()
+  })
```
carlos-marchal-ph, line 138: The `try/catch` in the wrapper is scoped correctly, but we don't test that the wrapper rejects transparently if the underlying fetch rejects. A `mockFetch.mockRejectedValueOnce(err)` plus `expect(...).rejects.toEqual(err)` would confirm that.

packages/react-native/test/tracing-headers.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    const [, init] = mockFetch.mock.calls[0]
+    expect(init.method).toBe('POST')
+    expect(init.body).toBe('{}')
+    const headers = init.headers as Headers
+    expect(headers.get('Content-Type')).toBe('application/json')
+    expect(headers.get('Authorization')).toBe('Bearer abc')
+    expect(headers.get('X-POSTHOG-DISTINCT-ID')).toBe('d-1')
+    expect(headers.get('X-POSTHOG-SESSION-ID')).toBe('s-1')
+  })
```
carlos-marchal-ph: One more parameterised case covering `URL` input and another covering `init.headers` as a `Headers` instance would lock in those branches for free.

packages/react-native/test/tracing-headers.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  it.each([
+    { label: 'adds headers for a matching hostname', url: 'https://api.example.com/thing', expectHeaders: true },
+    {
+      label: 'does not add headers for a non-matching hostname',
+      url: 'https://other.example.com/thing',
+      expectHeaders: false,
+    },
+  ])('$label', async ({ url, expectHeaders }) => {
```
carlos-marchal-ph: The implementation has dedicated branches for `input` being a `Request` (both for reading the URL and for falling back to `input.headers` when `init.headers` is absent), but nothing exercises that path. Worth adding a case like `fetch(new Request('https://api.example.com/thing', { headers: { 'X-Caller': 'c' } }))` asserting both the caller header and the tracing headers end up on the outgoing init.

carlos-marchal-ph (approved):
Looks good overall, React Native can be very tricky though so ideally someone who has it set up on the client libraries side could lend you a hand :)

packages/react-native/src/posthog-rn.ts
marandaneto: __add_tracing_headers would still need to be an alias to avoid breaking changes
i have no idea who added this and why with the __
can we git blame and ask?
I know @dustinbyrne started to work on this to make 'distributed tracing' just work, so maybe that's something he'd already?

dustinbyrne: agree, I wouldn't worry too much about maintaining browser conventions here because we need to update the browser SDK as well.

Prefer something like `addTracingHeaders`, no need to maintain the snake case either

richardsolomou: Renamed to `addTracingHeaders` on both the RN and browser SDKs in e7f32be2. `__add_tracing_headers` kept as a deprecated alias on both so existing users keep working — tried the new name first and fall back. Updated the types snapshot too. Happy to split the browser-side rename into its own PR if that's cleaner, but since it's literally three lines of fallback I folded it in here to avoid the churn.

packages/react-native/src/tracing-headers.ts
richardsolomou: Good catch, dropped in e7f32be2.

packages/react-native/src/tracing-headers.ts
richardsolomou: Done in e7f32be2.

packages/react-native/src/tracing-headers.ts
richardsolomou: Agreed, bound to `const initWithHeaders` in e7f32be2.

packages/core/src/tracing-headers.ts
richardsolomou: Added a comment in e7f32be2 explaining the limitation.

packages/react-native/test/tracing-headers.spec.ts
richardsolomou: Done in e7f32be2 — replaced with `globalThis as unknown as { fetch: typeof fetch | undefined }` (kept `| undefined` so the "fetch not available" test can still assign `undefined`).

packages/core/src/__tests__/tracing-headers.spec.ts
richardsolomou: Added in e7f32be2. `propagates rejections from the underlying fetch transparently` uses `mockRejectedValueOnce` and asserts `.rejects.toBe(err)`.

packages/react-native/test/tracing-headers.spec.ts
richardsolomou: Done in e7f32be2. Folded URL-input and Request-input into the existing `it.each`, and added a dedicated test for `init.headers` being a `Headers` instance (kept it as a standalone `it` because asserting header preservation differs from the matches/doesn't-match shape of the `it.each`).

packages/react-native/test/tracing-headers.spec.ts
richardsolomou: Covered in e7f32be2 with two tests: the `it.each` case `Request input, matching hostname` exercises the URL-extraction branch, and `preserves caller-provided headers when fetch is called with a Request` exercises the `input.headers` fallback (asserts both `X-Caller` and the tracing headers land on the outgoing init).

packages/react-native/src/posthog-rn.ts
```diff
@@ -105,6 +106,30 @@ export interface PostHogOptions extends PostHogCoreOptions {
...
+   *   addTracingHeaders: ['api.example.com'],
+   * })
+   * ```
+   */
+  addTracingHeaders?: string[]
+
+  /**
+   * @deprecated Use {@link addTracingHeaders} instead. Kept for backwards compatibility.
+   */
+  __add_tracing_headers?: string[]
```
marandaneto: since RN has its own config, should we just not add this at all?

packages/react-native/src/tracing-headers.ts
```diff
@@ -0,0 +1,84 @@
+import { isFunction } from '@posthog/core'
```
marandaneto: if this whole fine isnt RN specific, and this can be reused for js-lite and node, we could move to the core package

.changeset/rn-add-tracing-headers.md
```diff
@@ -0,0 +1,9 @@
+---
+"posthog-react-native": minor
+"posthog-js": patch
+"@posthog/types": minor
+---
+
+feat: rename `__add_tracing_headers` to `addTracingHeaders` (the `__` prefix was a legacy naming convention for internal/experimental options — the option is now public). `__add_tracing_headers` continues to work as a deprecated alias.
+
+feat(react-native): support `addTracingHeaders` (and the deprecated `__add_tracing_headers`) to inject `X-POSTHOG-DISTINCT-ID` and `X-POSTHOG-SESSION-ID` headers on outgoing `fetch` requests for linking LLM traces and session replays.
```
marandaneto: not a blocker but if you want you could split the changeset into multiple files
so the types and JS are in one file with the rename
the other file is the react native with adding support for something new

right now the RN changelog will contain a rename of a property that didnt exist before which might be confusing

marandaneto (approved):
left a few suggestions otherwise all good
assuming you validated this with @turnipdabeets LGTM

packages/core/src/__tests__/tracing-headers.spec.ts
richardsolomou: @turnipdabeets Thanks, I'll take you up on that. Let me know when's good — I'll ping you on Slack to coordinate.

marandaneto:
Thanks @richardsolomou o/

packages/react-native/src/posthog-rn.ts
richardsolomou: Good point, dropped in 02c45662.

packages/react-native/src/tracing-headers.ts
richardsolomou: You're right, moved to `@posthog/core` in 02c45662 (with its tests). Decoupled from the RN `PostHog` type via a minimal `TracingHeadersClient` interface (`{ getDistinctId, getSessionId }`) so anything that satisfies that contract can use it — posthog-web could pick this up next. Not touching the browser SDK's separate impl in this PR (it patches XHR too and goes via `assignableWindow`), but that could be a follow-up to consolidate.

.changeset/rn-add-tracing-headers.md
richardsolomou: Split in 02c45662 — `rn-add-tracing-headers.md` now only contains the RN feature entry, and `rename-add-tracing-headers.md` covers the browser/types/core rename.

packages/react-native/src/posthog-rn.ts
```diff
@@ -105,6 +106,25 @@ export interface PostHogOptions extends PostHogCoreOptions {
...
+   * to requests whose hostname matches one of the entries in the list.
+   *
+   * @example
+   * ```ts
+   * new PostHog('<key>', {
+   *   addTracingHeaders: ['api.example.com'],
+   * })
+   * ```
+   */
+  addTracingHeaders?: string[]
```
marandaneto: if you intend to reuse this across node/js-lite you could move this to packages/core/src/types.ts (not a blocker)

richardsolomou: Good catch! Done in 7269ee22 — `addTracingHeaders` now lives on `PostHogCoreOptions` in `types.ts`, so any SDK extending it (RN today, posthog-web next) picks up the option alongside the `patchFetchForTracingHeaders` helper for free. Dropped the duplicate declaration from `PostHogOptions` in RN.

turnipdabeets (approved):

PR #3433, opened 2026-04-21
fix(ai): adjust V3 cache tokens for Claude on any host

### LINKED ISSUE #3022: Cost calculation of aisdk from @posthog/ai is not cache aware

Cost calculation in sendEventToPosthog
Costs are calculated as inputCostUSD = inputCost * inputTokens and outputCostUSD = outputCost * outputTokens without subtracting cache

Example:
Image

The displayed cost is $0.0875765
But it should be 

```python
prompt_total = 21561
output_total = 139
cache_read = 17399
cache_write = 4157

# pricing reference: https://models.dev/?search=claude-sonnet-4-5%4020250929
price_input_per_1m = 3.0
price_output_per_1m = 5.0
price_read_per_1m = 0.3
price_write_per_1m = 3.75

# Assumption: Prompt total includes cache read and write tokens.
# Remainder is charged at standard input price.
standard_input_tokens = prompt_total - (cache_read + cache_write)

cost_standard = (standard_input_tokens / 1_000_000) * price_input_per_1m
cost_read = (cache_read / 1_000_000) * price_read_per_1m
cost_write = (cache_write / 1_000_000) * price_write_per_1m
cost_output = (output_total / 1_000_000) * price_output_per_1m

total_cost = cost_standard + cost_read + cost_write + cost_output

print(f"{standard_input_tokens=}")
print(f"{cost_standard=}")
print(f"{cost_read=}")
print(f"{cost_write=}")
print(f"{cost_output=}")
print(f"{total_cost=}")
```

```text
standard_input_tokens=5
cost_standard=1.5000000000000002e-05
cost_read=0.0052197
cost_write=0.01558875
cost_output=0.000695
total_cost=0.02151845
```

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

richardsolomou:
## Problem

Closes #3022.

For Vercel AI SDK LanguageModel V3, `adjustAnthropicV3CacheTokens` only fired when `model.provider` contained `"anthropic"`. With `@ai-sdk/amazon-bedrock`, the provider is `"amazon-bedrock"`, so Bedrock-hosted Claude skipped the adjustment entirely. The server's exclusive cache accounting (auto-detected from `claude` in the model name) then double-charged cache tokens — @gauravp-cubic reported 5-10x cost overstatement at typical cache hit rates, up to 9.6x at 95%+ hit ratios.

Second bug: `extractAdditionalTokenValues` only read `providerMetadata.anthropic.cacheCreationInputTokens`. Bedrock populates cache-write via the V3-standardized `usage.inputTokens.cacheWrite` instead, so `$ai_cache_creation_input_tokens` was silently dropped on every Bedrock-Claude event.

## Changes

- Broaden Anthropic detection: match on `provider.includes("anthropic")` OR `modelId` containing `claude`/`anthropic`. This covers direct Anthropic, Amazon Bedrock, Google Vertex, OpenRouter, and any other Claude host — mirroring the server's own `claude`-based heuristic for `$ai_cache_reporting_exclusive`.
- Add `extractCacheWriteTokens` helper that reads V3's standardized `usage.inputTokens.cacheWrite`, and fall back to it in `extractAdditionalTokenValues` when `providerMetadata.anthropic` is absent. Wired through both `doGenerate` and `doStream`.

Verified against the exact numbers in the issue:
- OP's 21561/17399/4157 → `$ai_input_tokens = 5`, `$ai_cache_creation_input_tokens = 4157`.
- @gauravp-cubic's row 1 (22145/21405) → `$ai_input_tokens = 740`, cost `$0.01440` (matches their "Correct" column exactly, down from the reported `$0.12143`).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

---

### DIFF
--- a/packages/ai/src/vercel/middleware.ts
+++ b/packages/ai/src/vercel/middleware.ts
@@ -301,41 +301,6 @@ const extractWebSearchCount = (providerMetadata: unknown, usage: any): number =>
   })
 }
 
-// Extract additional token values from provider metadata
-const extractAdditionalTokenValues = (providerMetadata: unknown): Record<string, any> => {
-  if (
-    providerMetadata &&
-    typeof providerMetadata === 'object' &&
-    'anthropic' in providerMetadata &&
-    providerMetadata.anthropic &&
-    typeof providerMetadata.anthropic === 'object' &&
-    'cacheCreationInputTokens' in providerMetadata.anthropic
-  ) {
-    return {
-      cacheCreationInputTokens: providerMetadata.anthropic.cacheCreationInputTokens,
-    }
-  }
-  return {}
-}
-
-// For Anthropic providers in V3, inputTokens.total is the sum of all tokens (uncached + cache read + cache write).
-// Our cost calculation expects inputTokens to be only the uncached portion for Anthropic.
-// This helper subtracts cache tokens from inputTokens for Anthropic V3 models.
-const adjustAnthropicV3CacheTokens = (
-  model: LanguageModel,
-  provider: string,
-  usage: { inputTokens?: number; cacheReadInputTokens?: unknown; cacheCreationInputTokens?: unknown }
-): void => {
-  if (isV3Model(model) && provider.toLowerCase().includes('anthropic')) {
-    const cacheReadTokens = (usage.cacheReadInputTokens as number) || 0
-    const cacheWriteTokens = (usage.cacheCreationInputTokens as number) || 0
-    const cacheTokens = cacheReadTokens + cacheWriteTokens
-    if (usage.inputTokens && cacheTokens > 0) {
-      usage.inputTokens = Math.max(usage.inputTokens - cacheTokens, 0)
-    }
-  }
-}
-
 // Helper to extract numeric token value from V2 (number) or V3 (object with .total) usage formats
 const extractTokenCount = (value: unknown): number | undefined => {
   if (typeof value === 'number') {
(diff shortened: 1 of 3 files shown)


packages/ai/src/vercel/middleware.ts
```diff
@@ -388,6 +353,75 @@ const extractCacheReadTokens = (usage: Record<string, unknown>): unknown => {
...
+    }
+  }
+  return {}
+}
+
+// Detects Anthropic Claude regardless of host (direct Anthropic, Amazon Bedrock, Google Vertex, etc.).
+// The server applies exclusive cache token accounting based on the model name, so any Claude model
+// needs its V3 input tokens adjusted to exclude cache tokens — not just those routed through a
+// provider whose name contains "anthropic".
+const isAnthropicClaudeModel = (model: LanguageModel, provider: string): boolean => {
```
carlos-marchal-ph: `isAnthropicClaudeModel` reads from `model.modelId`, but the id sent to the server (and used by its exclusive cache accounting heuristic) honours `posthogModelOverride` and `result.response?.modelId`. So if someone overrides the id to hide or expose 'claude', the client and server will disagree on whether to adjust. Could we pass the captured modelId string in here instead of the raw model? That would also let you drop the `LanguageModel` dependency from this helper's signature.

packages/ai/src/vercel/middleware.ts
```diff
@@ -388,6 +353,75 @@ const extractCacheReadTokens = (usage: Record<string, unknown>): unknown => {
...
+    'cacheCreationInputTokens' in providerMetadata.anthropic
+  ) {
+    return {
+      cacheCreationInputTokens: providerMetadata.anthropic.cacheCreationInputTokens,
+    }
+  }
+  if (usage && typeof usage === 'object') {
+    const cacheWrite = extractCacheWriteTokens(usage as Record<string, unknown>)
+    if (cacheWrite !== undefined) {
+      return { cacheCreationInputTokens: cacheWrite }
```
carlos-marchal-ph, line 391: This makes `$ai_cache_creation_input_tokens` appear as 0 on events where previously the field was omitted (cacheWrite of 0 with no `providerMetadata.anthropic`). Probably fine and arguably more informative, but it is a shape change downstream consumers will see. If you want to be conservative, treat 0 as absent in this fallback branch.

packages/ai/src/vercel/middleware.ts
```diff
@@ -388,6 +353,75 @@ const extractCacheReadTokens = (usage: Record<string, unknown>): unknown => {
...
+  ) {
+    return (usage.inputTokens as { cacheWrite: unknown }).cacheWrite
+  }
+  return undefined
+}
+
+// Extract additional token values from provider metadata, with a V3 standardized fallback
+// (e.g. Amazon Bedrock exposes cache write tokens via usage.inputTokens.cacheWrite rather
+// than providerMetadata.anthropic.cacheCreationInputTokens).
+const extractAdditionalTokenValues = (providerMetadata: unknown, usage?: unknown): Record<string, any> => {
```
carlos-marchal-ph: Both call sites pass `usage`. Worth making it required in the signature so the contract is clearer?

packages/ai/tests/vercel.test.ts
```diff
@@ -797,6 +797,153 @@ describe('Vercel AI SDK - Dual Version Support', () => {
...
+      await flushPromises()
+
+      expect(mockPostHogClient.capture).toHaveBeenCalledTimes(1)
+      const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
+
+      // 1120 - 1000 - 20 = 100
+      expect(captureCall[0].properties['$ai_input_tokens']).toBe(100)
+      expect(captureCall[0].properties['$ai_cache_read_input_tokens']).toBe(1000)
+      expect(captureCall[0].properties['$ai_cache_creation_input_tokens']).toBe(20)
+    })
```
carlos-marchal-ph: Agreeing with greptile here. These three share all their scaffolding and differ only in the usage object and the expected properties. An `it.each` table would match the pattern used elsewhere in this file (see the 'reasoning in %s streaming' test) and would make it trivial to add a Vertex or OpenRouter row later. Extending `createMockAnthropicV3Model` to optionally omit `providerMetadata.anthropic` would work too.

packages/ai/tests/vercel.test.ts
```diff
@@ -797,6 +797,153 @@ describe('Vercel AI SDK - Dual Version Support', () => {
...
+        model: model,
+        prompt: 'Test Bedrock Claude cache read',
+      })
+
+      expect(mockPostHogClient.capture).toHaveBeenCalledTimes(1)
+      const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
+
+      // 22145 - 21405 - 0 = 740
+      expect(captureCall[0].properties['$ai_input_tokens']).toBe(740)
+      expect(captureCall[0].properties['$ai_cache_read_input_tokens']).toBe(21405)
```
carlos-marchal-ph: This test's title says it covers reading cacheWrite from V3 usage, but it does not assert what `$ai_cache_creation_input_tokens` actually ends up as. Given the shape change in the fallback (0 vs undefined), worth pinning it down with `.toBe(0)` or `.toBeUndefined()`.

packages/ai/tests/vercel.test.ts
```diff
@@ -797,6 +797,153 @@ describe('Vercel AI SDK - Dual Version Support', () => {
...
+
+    it('should adjust cache tokens and read cacheWrite from V3 usage for Bedrock Claude', async () => {
+      // Simulates @ai-sdk/amazon-bedrock hosting Claude: provider is "amazon-bedrock",
+      // no providerMetadata.anthropic is set, and cacheWrite is exposed via usage.inputTokens.cacheWrite.
+      // Previously this path was skipped, causing cacheRead/cacheWrite tokens to be double-counted
+      // by the server's exclusive cache accounting (triggered by the "claude" model name).
+      const baseModel: LanguageModelV3 = {
+        specificationVersion: 'v3' as const,
+        provider: 'amazon-bedrock',
+        modelId: 'global.anthropic.claude-opus-4-6-v1',
```
carlos-marchal-ph, line 811: All three Bedrock ids here contain both 'anthropic' and 'claude'. Vertex ids look like `claude-sonnet-4-5@20250929`, only 'claude'. A test covering that would verify the `|anthropic` arm of the regex.

carlos-marchal-ph (approved):

packages/ai/src/vercel/middleware.ts
richardsolomou: Good catch — refactored in 38a68d4c. `isAnthropicClaudeModel` now takes the resolved modelId string (honouring `posthogModelOverride` and `response.modelId`) so client and server see the same id. Also drops the `LanguageModel` dependency from the signature like you suggested.

packages/ai/src/vercel/middleware.ts
richardsolomou: Went conservative per your suggestion — fallback in 38a68d4c only returns when `cacheWrite > 0`, so events that previously omitted `$ai_cache_creation_input_tokens` still do.

packages/ai/src/vercel/middleware.ts
richardsolomou: Done in 38a68d4c.

packages/ai/tests/vercel.test.ts
richardsolomou: Collapsed into an `it.each` table in 38a68d4c. Rows: Bedrock cacheRead-only, Bedrock cacheRead+Write via V3 location, and Vertex-style modelId. Adding a new host is now a one-line row.

packages/ai/tests/vercel.test.ts
richardsolomou: Pinned in 38a68d4c — the `it.each` table now asserts `$ai_cache_creation_input_tokens` explicitly, using `.toBeUndefined()` for the cacheWrite=0 row and `.toBe(n)` for the populated rows. This also locks in the conservative shape-preservation for the 0 case.

packages/ai/tests/vercel.test.ts
richardsolomou: Added a Vertex row with `claude-sonnet-4-5@20250929` in 38a68d4c — covers the `claude` arm of the regex without relying on the `anthropic` substring.

PR #3437, opened 2026-04-21
fix(core): log invalid api keys and disable client

### LINKED ISSUE #3436: [@posthog/nuxt] Server crashes with empty API key in SPA mode (ssr: false)
## Bug description

The `@posthog/nuxt` module registers a Nitro server plugin (`nitro-plugin.js`) that calls `new PostHog(publicKey, ...)` unconditionally. When `publicKey` is empty or not set in the server runtime environment, this throws:

```
Error: You must pass your PostHog project's api key.
```

This happens even when Nuxt is configured with `ssr: false` (SPA mode), where the Nitro server only serves the static SPA shell and has no use for server-side PostHog.

## Steps to reproduce

1. Create a Nuxt 4 app with `ssr: false`
2. Install `@posthog/nuxt` and configure it with `NUXT_PUBLIC_POSTHOG_PROJECT_TOKEN` env var
3. Build and deploy — the env var is available at build time (via `--dotenv .env.prod`) but not in the server runtime environment
4. Server crashes on startup

## Expected behavior

The Nitro plugin should gracefully skip initialization when `publicKey` is empty/missing, or not register at all in SPA mode where it serves no purpose.

## Current workaround

Remove the Nitro plugin via a Nuxt hook:

```ts
// nuxt.config.ts
hooks: {
  "nitro:config"(nitroConfig) {
    nitroConfig.plugins = nitroConfig.plugins?.filter(
      (p) => typeof p === "string" && !p.includes("@posthog/nuxt"),
    );
  },
},
```

## Environment

- `@posthog/nuxt`: 1.x (latest)
- Nuxt: 4.3.1
- `ssr: false` (SPA mode)
- Node.js: v22

## Suggested fix

Either:
1. Guard the server plugin: `if (!publicKey) return;` at the top of `nitro-plugin.js`
2. Don't register `addServerPlugin()` when `ssr: false` is detected in `module.setup()`
3. Both

dustinbyrne:
## Problem

Invalid or blank API keys in `@posthog/core` currently throw during `PostHogCoreStateless` construction.

Generally, we want to avoid throwing exceptions from within the SDK, so invalid API keys are better logged as an error.

Fixes https://github.com/PostHog/posthog-js/issues/3436

## Changes

- replaced the constructor assertion in `posthog-core-stateless` with a missing-key check that logs an error and disables the client
- kept a single `_logger` instantiation in the constructor
- updated init tests to assert missing/empty/non-string API keys now log and disable the client instead of throwing
- added a changeset for `@posthog/core`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -24,7 +24,6 @@ import {
 } from './types'
 import {
   allSettled,
-  assert,
   currentISOTime,
   PromiseQueue,
   removeTrailingSlash,
@@ -147,8 +146,12 @@ export abstract class PostHogCoreStateless {
   constructor(apiKey: string, options: PostHogCoreOptions = {}) {
     const normalizedApiKey = typeof apiKey === 'string' ? apiKey.trim() : ''
     const normalizedHost = typeof options.host === 'string' ? options.host.trim() : ''
+    const missingApiKey = !normalizedApiKey
 
-    assert(normalizedApiKey, "You must pass your PostHog project's api key.")
+    this._logger = createLogger('[PostHog]', this.logMsgIfDebug.bind(this))
+    if (missingApiKey) {
+      this._logger.error("You must pass your PostHog project's api key. The client will be disabled.")
+    }
 
     this.apiKey = normalizedApiKey
     this.host = removeTrailingSlash(normalizedHost || 'https://us.i.posthog.com')
@@ -170,12 +173,11 @@ export abstract class PostHogCoreStateless {
     this.featureFlagsRequestTimeoutMs = options.featureFlagsRequestTimeoutMs ?? 3000 // 3 seconds
     this.remoteConfigRequestTimeoutMs = options.remoteConfigRequestTimeoutMs ?? 3000 // 3 seconds
     this.disableGeoip = options.disableGeoip ?? true
-    this.disabled = options.disabled ?? false
+    this.disabled = (options.disabled ?? false) || missingApiKey
     this.historicalMigration = options?.historicalMigration ?? false
     // Init promise allows the derived class to block calls until it is ready
     this._initPromise = Promise.resolve()
     this._isInitialized = true
-    this._logger = createLogger('[PostHog]', this.logMsgIfDebug.bind(this))
     // Support both evaluationContexts (new) and evaluationEnvironments (deprecated)
     this.evaluationContexts = options?.evaluationContexts ?? options?.evaluationEnvironments
     if (options?.evaluationEnvironments && !options?.evaluationContexts) {
--- a/.changeset/fresh-worms-fix.md
+++ b/.changeset/fresh-worms-fix.md
@@ -1,0 +1,5 @@
+---
+'@posthog/core': patch
+---
+
+Log invalid API keys as errors and disable the core client instead of throwing.
--- a/.changeset/fresh-worms-fix.md
+++ b/.changeset/fresh-worms-fix.md
@@ -2,4 +2,4 @@
 '@posthog/core': patch
 ---
 
-Log invalid API keys as errors and disable the core client instead of throwing.
+Disable the core client instead of throwing when the API key is missing, blank, or invalid.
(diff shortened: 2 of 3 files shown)


marandaneto (approved):

packages/core/src/posthog-core-stateless.ts
```diff
@@ -147,8 +146,12 @@ export abstract class PostHogCoreStateless {
   constructor(apiKey: string, options: PostHogCoreOptions = {}) {
     const normalizedApiKey = typeof apiKey === 'string' ? apiKey.trim() : ''
     const normalizedHost = typeof options.host === 'string' ? options.host.trim() : ''
+    const missingApiKey = !normalizedApiKey
 
-    assert(normalizedApiKey, "You must pass your PostHog project's api key.")
+    this._logger = createLogger('[PostHog]', this.logMsgIfDebug.bind(this))
+    if (missingApiKey) {
```
marandaneto, line 152: was that caused by https://github.com/PostHog/posthog-js/pull/3426 ?
because the SDK already had `assert(apiKey, "You must pass your PostHog project's api key.")` before so it should be the same behaviour as before

marandaneto: i know some other SDKs still throw if apiKey is missing, but just logging an error instead of throwing leads to some other errors so its not a one line change

dustinbyrne: Yeah, see https://github.com/PostHog/posthog-js/issues/3436

`assert` throws, so it seems like it caught some people unexpectedly. I'm just removing the exception here.

marandaneto: i saw the issue, just wondering how this is an issue since the `assert` already existed?

marandaneto: so i am just wondering if this file is the right place to be fixed

dustinbyrne: You're right, that assertion would have been throwing regardless of https://github.com/PostHog/posthog-js/pull/3426, so this is actually unrelated. I jumped to conclusions here. If we'd prefer to keep the assertion, I can close this.

So this is a longer-lived issue where we're still constructing a Node client in Nuxt even if SSR is disabled. I'll take a look into that.

marandaneto: I think your approach here makes sense, regardless. I don't know if the SDK will throw somewhere without the API key, prob not, since you did this `this. disabled = (options.disabled ?? false) || missingApiKey.`
I didn't check or bother to change all SDKs to do something similar because it'd be a bigger change for now

dustinbyrne: Okay, and thinking some more about this:
> So this is a longer-lived issue where we're still constructing a Node client in Nuxt even if SSR is disabled. I'll take a look into that.

I don't think this is something we'd want to change in the Nuxt integration. No SSR != no server-side logic, so not creating a client would be incorrect. The server-side shouldn't be throwing just because it doesn't make use of the node client.

Second guessed myself, but I'm re-opening this 😄

packages/core/src/posthog-core-stateless.ts
```diff
@@ -147,8 +146,12 @@ export abstract class PostHogCoreStateless {
...
+      this._logger.error("You must pass your PostHog project's api key. The client will be disabled.")
```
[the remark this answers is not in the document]
dustinbyrne: I edited the stale PR description which likely brought on this line of thinking. I think this is fine. It silently won't work if debug logging is disabled. IMO this is better than logging critical/error in cases where the API is expected not to be set.

PR #3447, opened 2026-04-22
feat(node): add `getFeatureFlags` for bulk flag evaluation with events

dmarticus:
## TL;DR

Add a new `getFeatureFlags(keys, distinctId, options)` method to the Node SDK for evaluating a known subset of feature flags in one bulk pass while emitting `$feature_flag_called` events. Also adds a `sendFeatureFlagEvents` option to `getAllFlags` and `getAllFlagsAndPayloads` for opt-in per-flag event emission.

The idea here is that we have some customers who want to evaluate a bunch of flags on the server side as part of like an initial page load for their app, but they (a) don't want to call `getAllFlags` because they only want to evaluate _some_ of them (plus `getAllFlags` doesn't emit `$feature_flag_called` events for all flags, which I'm patching here too), and (b) they don't want to do like 10 individual `getFeatureFlag` calls (10 network requests! dumb!) as part of loading one page. Honestly this is so reasonable I can't believe we didn't have something like it already.

I'm gonna add this to the rest of the server-side SDKs too, I think people will really like this. Prioritizing Node first as a proving ground + the specific customer who asked about this is using Node.

## Changes

- **New `getFeatureFlags` method**: Evaluates a subset of feature flags in bulk while still emitting `$feature_flag_called` events per resolved flag
- **Stuck with the hybrid evaluation strategy**: Locally-evaluated flags reuse the poller's cached definitions; keys that can't be resolved locally fall through to a single remote `/flags` call with `flag_keys_to_evaluate`
- **Stuck with event deduplication**: Uses the existing `distinctIdHasSentFlagCalls` cache so the single-flag and bulk paths share one source of truth
- **New `sendFeatureFlagEvents` option**: Added to both `getAllFlags` and `getAllFlagsAndPayloads` methods for opt-in per-flag event emission. Defaults to off, obvi, but people may like this.
- **Buncha tests**: Added 362 lines of tests covering remote evaluation, local-only evaluation, hybrid scenarios, event deduplication, and edge cases

### Libraries affected

- [x] posthog-node

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility (new method and optional parameter, no breaking changes)
- [x] Changeset generated

---

### DIFF
--- a/.changeset/node-bulk-flag-events.md
+++ b/.changeset/node-bulk-flag-events.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': minor
+---
+
+Add `getFeatureFlags(keys, distinctId, options)` for evaluating a known subset of feature flags in one bulk pass while still emitting `$feature_flag_called` events per resolved flag. Also adds a `sendFeatureFlagEvents` option to the existing `getAllFlags` and `getAllFlagsAndPayloads` methods for opt-in per-flag event emission. Locally-evaluated flags reuse the poller's cached definitions; any keys that can't be resolved locally fall through to a single remote `/flags` call with `flag_keys_to_evaluate`. Event dedup uses the existing `distinctIdHasSentFlagCalls` cache so the single-flag and bulk paths share one source of truth.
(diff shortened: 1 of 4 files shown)


dmarticus:
closing in favor of a more principled approach, see https://posthog.slack.com/archives/C07Q2U4BH4L/p1777054243923129

PR #3453, opened 2026-04-23
refactor(logs): lift OTLP utilities to @posthog/core

turnipdabeets:
Relocates browser's OTLP serialization helpers from                                                                                                                                                                
  `packages/browser/src/logs-utils.ts` into `packages/core/src/logs/logs-utils.ts`                                                                                                                                       
  so the upcoming React Native logs feature consumes the same builders without
  duplication. Browser imports them from
  @posthog/core; the local browser file and its test are deleted.                                                                                                                                                    
                                               
  User-visible behavior changes for browser:                                                                                                                                                                         
                                               
  1. NaN and plus or minus Infinity attribute values now serialize as                                                                                                                                                
     {stringValue: NaN or Infinity or -Infinity} instead of being silently
     lost during JSON encoding. JSON.stringify turns non-finite floats into                                                                                                                                          
     null, which the browser SDK was unknowingly emitting on the wire — a                                                                                                                                            
     silent data-loss bug. proto3 JSON mapping (which OTLP HTTP rides)
     requires the literal strings; this change preserves the value regardless                                                                                                                                        
     of what downstream parsers do with null.  
                                                                                                                                                                                                                     
  2. Browser now populates scope.version with Config.LIB_VERSION. Server-side,                                                                                                                                       
     the instrumentation_scope column changes from "posthog-js@" to
     "posthog-js@<semver>" — a telemetry improvement the server already knew                                                                                                                                         
     how to use. Risk surface: queries filtering on the literal empty-version                                                                                                                                        
     pattern stop matching new logs.                                                                                                                                                                                 
                                                                                                                                                                                                                     
  Type changes:                                                                                                                                                                                                      
  - @posthog/types LogSdkContext gains optional screenName and appState
    fields so a single context shape works for both web (currentUrl) and                                                                                                                                             
    mobile (screenName + appState) callers. Existing fields unchanged.
  - @posthog/core gains a workspace dependency on @posthog/types so the                                                                                                                                              
    lifted helpers can consume the shared log type vocabulary.                                                                                                                                                       
  - @posthog/react-native adds @posthog/types as a workspace dep in
    preparation for the upcoming logs module.                                                                                                                                                                        
                                                 
  Mechanical refactor for browser:                                                                                                                                                                                   
  buildOtlpLogsPayload previously hardcoded Config.LIB_NAME inside the
  function; the lifted version takes scopeName and scopeVersion as required
  parameters. Browser passes Config.LIB_NAME and Config.LIB_VERSION. 

## Problem

<!-- Who are we building for, what are their needs, why is this important? -->

## Changes

<!-- What is changed and what information would be useful to a reviewer? -->

### Libraries affected

<!-- Please mark which libraries will require a version bump. -->

- [ ] All of them
- [X] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

<!-- For more details check RELEASING.md -->


frankh (approved):

.changeset/lift-otlp-utils-to-core.md
```diff
@@ -0,0 +1,11 @@
+---
+'posthog-js': patch
+'@posthog/core': patch
```
marandaneto, line 4: i think types should also be released? looks like its also changed

packages/react-native/src/logs/types.ts
```diff
@@ -0,0 +1,62 @@
...
+
+// Mobile defaults exported so the PostHogLogs implementation in later commits
+// can apply them in one place.
+export const DEFAULT_FLUSH_INTERVAL_MS = 10000
+export const DEFAULT_RATE_CAP_WINDOW_MS = 10000
+export const DEFAULT_MAX_BUFFER_SIZE = 100
+export const DEFAULT_MAX_LOGS_PER_INTERVAL = 500
+export const DEFAULT_MAX_BATCH_RECORDS_PER_POST = 50
+export const DEFAULT_BACKGROUND_FLUSH_BUDGET_MS = 25000
+export const DEFAULT_TERMINATION_FLUSH_BUDGET_MS = 2000
```
marandaneto, line 61: Should those live in the core or types then? or is it RN specific?

turnipdabeets: These are mobile defaults, so RN specific

PR #3457, opened 2026-04-23
feat(node): use relative file paths in error tracking stack frames

hpouillot:
## Problem

The Node.js SDK sends absolute file paths in error tracking stack frames (e.g., `/Users/alice/myapp/src/index.js`). This causes:
- **Poor fingerprinting in Cymbal**: the same error on different machines produces different fingerprints because absolute paths differ
- **Privacy concerns**: absolute paths leak filesystem structure (usernames, deployment paths)

Cymbal does NOT use `filename` for source map resolution (it uses `chunk_id` + `lineno`/`colno`), so making it relative is safe for symbolication.

## Changes

Adds a `createRelativePathModifier()` that runs as the **last** frame modifier, after module and context-lines modifiers have already used the absolute paths they need. For each frame with an absolute `filename`:
- Stores the original in `abs_path`
- Converts `filename` to a relative path using `process.cwd()` as base

Skips `node:`, `data:`, and already-relative paths. No `fs` dependency so it works in edge environments.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/breezy-pumas-build.md
+++ b/.changeset/breezy-pumas-build.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': minor
+---
+
+send relative file paths in frames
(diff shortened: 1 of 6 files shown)


ablaszkiewicz (approved):
Very good change!!!

cat-ph (approved):
🚀

PR #3461, opened 2026-04-23
chore(ci): trigger release workflow from pushes to main

dustinbyrne:
## Problem

The `Release` workflow is gated by the `NPM Release` environment, which is only allowed to run from `main`.

Today the workflow triggers from `pull_request.closed`, so the run reaches `version-bump` on a `refs/pull/*/merge` ref instead of `refs/heads/main`. That breaks environment protection and causes the deployment request to be rejected even when the PR was merged into `main`.

## Changes

- switch the release workflow trigger from `pull_request.closed` to `push` on `main`
- keep `workflow_dispatch` for manual releases
- limit automatic push-triggered releases to merges that touch `.changeset/**`, so non-release pushes do not queue behind the release concurrency group
- add a small `resolve-release-context` job that:
  - skips automated `[version bump]` commits to avoid recursive release runs
  - resolves the PR associated with the pushed commit
  - only continues when a merged PR to `main` currently has the `release` label
- keep the existing batching behavior where additional `main` changes can accumulate while `version-bump` is awaiting approval

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,8 @@ permissions:
 on:
     push:
         branches: [main]
+        paths:
+            - '.changeset/**'
     workflow_dispatch:
 
 # Concurrency control: only one release process can run at a time
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -4,8 +4,7 @@ permissions:
     contents: read
 
 on:
-    pull_request:
-        types: [closed]
+    push:
         branches: [main]
     workflow_dispatch:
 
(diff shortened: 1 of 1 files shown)


dustinbyrne:
I'll roll this out to the other repositories if we're good with this strategy

feliperalmeida (approved):

.github/workflows/release.yml
```diff
@@ -4,9 +4,10 @@ permissions:
     contents: read
 
 on:
-    pull_request:
-        types: [closed]
+    push:
```
marandaneto, line 7: i think [pull_request_target](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#pull_request_target) would work as well but im unsure if its safe
i recall reading stuff about that one, iirc it was the vector attack for shai hulud

.github/workflows/release.yml
```diff
@@ -16,15 +17,69 @@ concurrency:
...
+                  CURRENT_REF: ${{ github.ref }}
+                  PUSH_SHA: ${{ github.sha }}
+                  PUSH_HEAD_COMMIT_MESSAGE: ${{ github.event.head_commit.message }}
+              run: |
+                  set -euo pipefail
+
+                  SHOULD_RELEASE=false
+
+                  if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
+                      if [ "$CURRENT_REF" != "refs/heads/main" ]; then
```
marandaneto, line 43: do we actually need this? if this workflow only runs on push for main

.github/workflows/release.yml
```diff
@@ -16,15 +17,69 @@ concurrency:
...
+                      else
+                          echo "::notice::Skipping release because push $PUSH_SHA is not associated with a merged PR to main that has the 'release' label."
+                      fi
+                      echo "should-release=false" >> "$GITHUB_OUTPUT"
+                      exit 0
+                  fi
+
+                  RELEASE_PRS=$(echo "$PRS_JSON" | jq -r '[.[] | select(.merged_at != null and .base.ref == "main" and (.labels | map(.name) | index("release"))) | "#\(.number)"] | join(", ")')
+                  echo "✓ Release requested by push $PUSH_SHA via $RELEASE_PRS"
+                  echo "should-release=true" >> "$GITHUB_OUTPUT"
```
marandaneto, line 76: i wonder if we actually need the release label
we already have changesets and check-changesets
the downside is if we want to merge something and not release right away, we dont set the label
we could simply not create the changeset as well
this would simplify the release job
curious about your opinion

dustinbyrne: Agree, this seems like a good approach. If you don't want to merge right away, better to put it on a separate branch otherwise it'll go out with an unrelated release anyhow.

.github/workflows/release.yml
dustinbyrne: The idea was that it ensured it only ran the workflow from main, but I don't think it's necessary

.github/workflows/release.yml
dustinbyrne: Yeah, I explicitly tried to stay away from `pull_request_target` for this reason

PR #3462, opened 2026-04-23
feat(logs): storage

turnipdabeets:
## Problem

Adds the internal `PostHogLogs` module that captures OTLP log records into a                                                                                                                                     
  dedicated AsyncStorage-backed queue on RN. The module is fully private —
  `posthog.captureLog()` is not yet exposed.                                                                                                                                                                                            
                                              
  No customer-facing behavior change in this PR. Want to confirm the storage approach, adding new storage blob for logs.                                                                                                                                                                                                      
                                                                                                                                                                                                                     
  ## Architecture: storage                                                                                                                                                      
                                                                                                                                                                                                                     
  Logs writes get physical isolation from the main RN storage file (prevents                                                                                                                                       
  write-amplification during log bursts and contamination across pipelines)                                                                                                                                          
  while keeping the standard `getPersistedProperty` / `setPersistedProperty`                                                                                                                                       
  API that every other feature uses.                                                                                                                                                                                 
                                                                                                                                                                                                                   
  - `PostHogPersistedProperty.LogsQueue` added to `@posthog/core`'s enum.                                                                                                                                            
  - RN's `getPersistedProperty` / `setPersistedProperty` overrides route this                                                                                                                                        
    specific key to a dedicated `PostHogRNStorage` instance backed by                                                                                                                                                
    `.posthog-rn-logs.json`; all other keys stay in `.posthog-rn.json` as                                                                                                                                            
    before.                                                                                                                                                                                                        
  - `reset()` preserves both `Queue` and `LogsQueue` so in-flight records                                                                                                                                            
    survive identity changes.                                                                                                                                                                                        
                                                                                                                                                                                                                     
  One AsyncStorage backend (customer-provided or auto-detected), two                                                                                                                                                 
  AsyncStorage keys, two `PostHogRNStorage` instances vs sharing the same file. 

Next up:
wires the HTTP send path (/i/v1/logs POST via retriable(), flush timer + size trigger, 5xx/network retry + 4xx drop, queue clear on success) plus the minimal resource attributes, beforeSend, and remote-config before this becomes a public-facing API. 

## Changes

<!-- What is changed and what information would be useful to a reviewer? -->

### Libraries affected

<!-- Please mark which libraries will require a version bump. -->

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

<!-- For more details check RELEASING.md -->


packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+ * before preload completes are held in `_pendingCaptures` and drained into the
+ * persisted queue once storage is ready.
+ */
+export class PostHogLogs {
+  private _localEnabled: boolean
+  private _storageReady: boolean = false
+  private _pendingCaptures: BufferedLogEntry[] = []
+  // Serializes concurrent flush calls — next flush awaits the previous one
+  // before running. No-op today (flushStorage is idempotent) but forward-
+  // compat for 2b where flushLogs will also POST to the server, at which
```
marandaneto: what does 2b mean here?

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+      this._logsStorage.preloadPromise
+        .then(() => this._onStorageReady())
+        .catch((err) => {
+          this._logger.error('Logs storage preload failed:', err)
+          // Preload has permanently failed. Flip the flag and proceed with an
+          // empty in-memory cache — reads will return undefined (as if the
+          // queue were empty), and captures will overwrite with an empty-
+          // baseline queue. We lose visibility into pre-existing persisted
+          // records but keep collecting new ones, which is better than
+          // permanently dropping every new capture into the pending buffer.
```
marandaneto: is that what we do for the events queue as well?

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+    if (!this._localEnabled || this._instance.optedOut) {
+      return
+    }
+    if (!options?.body) {
+      return
+    }
+
+    const sdkContext = this._getSdkContext()
+    const record = buildOtlpLogRecord(options, sdkContext)
+    const maxBufferSize = this._config?.maxBufferSize ?? DEFAULT_MAX_BUFFER_SIZE
```
marandaneto: sdkContext and maxBufferSize can probably be a global assigned in the ctor so we dont have to compute every single tine, they are static values right

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    if (!this._storageReady) {
+      // Cold-start window: hold in memory, drain when preload finishes.
+      // Bound by the same maxBufferSize so a log storm during init can't OOM.
+      if (this._pendingCaptures.length >= maxBufferSize) {
+        this._pendingCaptures.shift()
+        this._logger.info('Logs pending-buffer is full during storage preload, dropping oldest record.')
+      }
+      this._pendingCaptures.push({ record })
```
marandaneto: is that what we do for events? if the app crashes while data is held in memory, we lose this event

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    for (const entry of this._pendingCaptures) {
+      if (queue.length >= maxBufferSize) {
+        queue.shift()
+      }
+      queue.push(entry)
+    }
+    this._pendingCaptures = []
+
+    this._instance.setPersistedProperty(PostHogPersistedProperty.LogsQueue, queue)
```
marandaneto: `captureLog` does something similar here, should we extract and reuse?

packages/react-native/src/logs/index.ts
marandaneto: looks like `_getSdkContext` is dynamic, i'd rename this method tho
i thought sdk context is sdk name and sdk version
i'd call this like _getSessionContext or _getDynamicContext

packages/react-native/src/posthog-rn.ts
```diff
@@ -111,6 +117,7 @@ export interface PostHogOptions extends PostHogCoreOptions {
 export class PostHog extends PostHogCore {
   private _persistence: PostHogOptions['persistence']
   private _storage: PostHogRNStorage
+  private _logsStorage: PostHogRNStorage
```
marandaneto, line 123: i think we should rename `_storage` now to `_eventsStorage`

packages/react-native/src/storage.ts
```diff
@@ -10,12 +11,14 @@ export class PostHogRNStorage {
   memoryCache: PostHogStorageContents = {}
   storage: PostHogCustomStorage
   preloadPromise: Promise<void> | undefined
+  private _storageKey: string
   private _pendingPromises: Set<Promise<void>> = new Set()
 
-  constructor(storage: PostHogCustomStorage) {
+  constructor(storage: PostHogCustomStorage, storageKey: string) {
```
marandaneto: we just need to make sure that we are not breaking the public API since users can set their own `customStorage`, otherwise we'd need a major bump

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+ * shutdown `waitForPersist()`. Data access always goes through `_instance`.
+ *
+ * Preload race: during the cold-start window the underlying storage's
+ * `memoryCache` is empty until its async preload resolves. If captureLog ran
+ * the read-mutate-write path during that window it would read an empty queue
+ * and overwrite any persisted-from-previous-session records. Captures arriving
+ * before preload completes are held in `_pendingCaptures` and drained into the
+ * persisted queue once storage is ready.
+ */
+export class PostHogLogs {
```
marandaneto: could this be reused by js-lite and node? or is it very react native specific, if so why?
the `PostHog` type could be imported from core as well since `PostHog` inherits from `PostHogCore`

packages/react-native/src/posthog-rn.ts
```diff
@@ -196,10 +193,12 @@ export class PostHog extends PostHogCore {
     }
 
     if (theStorage) {
-      this._storage = new PostHogRNStorage(theStorage)
+      this._storage = new PostHogRNStorage(theStorage, POSTHOG_MAIN_STORAGE_KEY)
+      this._logsStorage = new PostHogRNStorage(theStorage, POSTHOG_LOGS_STORAGE_KEY)
       storagePromise = this._storage.preloadPromise
```
ioannisj: Shouldn't we gate on both _storage and _logsStorage with .all here?

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+        .then(() => this._onStorageReady())
+        .catch((err) => {
+          this._logger.error('Logs storage preload failed:', err)
+          // Preload has permanently failed. Flip the flag and proceed with an
+          // empty in-memory cache — reads will return undefined (as if the
+          // queue were empty), and captures will overwrite with an empty-
+          // baseline queue. We lose visibility into pre-existing persisted
+          // records but keep collecting new ones, which is better than
+          // permanently dropping every new capture into the pending buffer.
+          this._onStorageReady()
```
ioannisj: Can _onStorageReady() throw? (esp on custom storage impl) If so, error here could be from previous call to _onStorageReady

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,173 @@
...
+
+    const queue =
+      this._instance.getPersistedProperty<BufferedLogEntry[]>(PostHogPersistedProperty.LogsQueue) ?? []
+
+    if (queue.length >= maxBufferSize) {
+      queue.shift()
+      this._logger.info('Logs queue is full, dropping oldest record.')
+    }
+
+    queue.push({ record })
```
ioannisj: I think this is repeated in a few place, we can maybe extract it into a helper? (L82, L152 from what I see)

packages/react-native/src/posthog-rn.ts
ioannisj: Now logs queue is preserved as well right?

packages/react-native/src/logs/index.ts
turnipdabeets: will remove, it's just what I had queued up with Claude next

packages/react-native/src/logs/index.ts
turnipdabeets: I can swap _instance: PostHog → PostHogCore, but it looks like ErrorTracking uses PostHog as well - Should I change them both? Is there an example of a file we're reusing for js-lite and node? The rest is pretty generic except Cold-start preload handling because RN async storage has a hydration window and RN's shutdown-drain abstraction; Node uses beforeExit, browser uses pagehide/beacon

packages/react-native/src/posthog-rn.ts
turnipdabeets: yes, will update

packages/react-native/src/logs/index.ts
turnipdabeets: renamed

packages/react-native/src/logs/index.ts
turnipdabeets: Extracted

packages/react-native/src/storage.ts
turnipdabeets: This constructor isn't a public API, right? The `PostHog.customStorage` is still intact and we defined the separation inside using the customStorage but with the internal file names. Did a refactor to hopefully make it more clear.

packages/react-native/src/logs/index.ts
turnipdabeets: collapsed into _enqueue()

packages/react-native/src/logs/index.ts
turnipdabeets: refactored to match events

packages/react-native/src/logs/index.ts
marandaneto: > it looks like ErrorTracking uses PostHog as well - Should I change them both?

if you think its not breaking anything and make the error tracking code reusable (as core) instead of RN specific, sure, and then you could move that part to the core package (this should be a follow up PR and not a blocker here)

> Is there an example of a file we're reusing for js-lite and node?

the whole core package basically, as long as it makes sense, we should not duplicate across all packages if they do the same thing, have the same features

i'd not move everything to core if its a specific feature for that package eg (a node APM feature wont make sense in RN and JS-lite despite being pure RS)

marandaneto: in this case, i believe js-lite and node would benefit from this class so it'd make sense to live in the core package imo

packages/react-native/src/logs/index.ts
```diff
@@ -0,0 +1,144 @@
...
+  private async _doFlushStorage(): Promise<void> {
+    await this._logsStorage.waitForPersist()
+  }
+
+  private _enqueue(entry: BufferedLogEntry): void {
+    // Read opt-in state inside the init-guarded body so we see fully-loaded
+    // storage, not pre-preload defaults (matches events' `wrap()` callers).
+    if (this._instance.optedOut) {
+      return
+    }
```
marandaneto: i think this should live in the `captureLog` method and return early, otherwise we are wasting cycles by constructing the whole log and discarding after

packages/react-native/src/posthog-rn.ts
```diff
@@ -119,6 +127,7 @@ export class PostHog extends PostHogCore {
   private _disableSurveys: boolean
   private _disableRemoteConfig: boolean
   private _errorTracking: ErrorTracking
+  private _logs: PostHogLogs
```
marandaneto, line 132: adding logs config in the `PostHogOptions` comes after i guess?

packages/react-native/src/posthog-rn.ts
```diff
@@ -208,6 +213,31 @@ export class PostHog extends PostHogCore {
       })
     }
 
+    this._logs = new PostHogLogs(this, undefined, this._logger, this._logsStorage)
```
marandaneto: I wonder if this should be constructed after `initAfterStorage` or do you want to be able to log stuff while the SDK isn't ready yet?

packages/react-native/src/logs/index.ts
turnipdabeets: this got removed in a refactor

packages/react-native/src/posthog-rn.ts
turnipdabeets: Yes, intentional. The idea is captureLog should work from the moment new PostHog() returns, same contract as events capture() and ErrorTracking.captureException()

packages/react-native/src/logs/index.ts
turnipdabeets: good call, moved

packages/react-native/src/posthog-rn.ts
turnipdabeets: Opted for `Promise.allSettled` so that one doesn't block the other. Maybe be useful to have logs if events fail for example.

packages/react-native/src/posthog-rn.ts
turnipdabeets: renamed

packages/react-native/src/posthog-rn.ts
marandaneto: make sure to use `packages/core/src/utils/index.ts` allSettled utils instead of `Promise.allSettled`
actually a linter would be good for this as well
if you search for allSettled usages you see a comment where it tells to not use it due to patching issues

packages/react-native/src/logs/types.ts
```diff
@@ -35,27 +42,25 @@ export interface PostHogLogsConfig {
...
+  flushIntervalMs?: number
+  rateCapWindowMs?: number // separate from flushIntervalMs so flush cadence does not move the rate-cap window
+  maxBufferSize?: number
+  maxLogsPerInterval?: number
+  maxBatchRecordsPerPost?: number // keeps each POST under the 2 MB server cap
 
   // Shutdown — separate budgets because foreground→background and app-terminate
   // have different OS-imposed windows.
-  backgroundFlushBudgetMs?: number // default: 25000 (under iOS beginBackgroundTask ~30s)
-  terminationFlushBudgetMs?: number // default: 2000 (reset / app-terminate path)
```
marandaneto, line 47: why removing all default comments?

packages/react-native/src/logs/index.ts
turnipdabeets: Moved to Core. Node extends PostHogCoreStateless, needed to required passing a `getContext` closure so node can pass a distinct id and session id as well.

packages/react-native/src/posthog-rn.ts
turnipdabeets: Ah good call out, fixed.

packages/react-native/src/logs/types.ts
turnipdabeets: I don't think this file exists anymore, it got moved to core and so I removed comments that are RN specific. SDKs will pass in their defaults.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -194,7 +192,15 @@ export abstract class PostHogCoreStateless {
...
-  protected wrap(fn: () => void): void {
+  /**
+   * Schedule fn to run once the SDK has finished initializing. If init has
+   * been rejected or the SDK is disabled, fn is silently dropped. Used by
+   * the SDK's own capture/identify/alias/etc. paths, and by composition
+   * modules (e.g. PostHogLogs) that need init-gated work without subclassing.
+   *
+   * @internal
+   */
+  public wrap(fn: () => void): void {
```
marandaneto: i think the caller could call wrap and then captureLog instead of making this public, right? but only when you add the captureLog into this class as well
if you make it public, its just harder to make it non public anymore
the `internal` helps but if we can avoid even better

marandaneto (approved):
approving to unblock, left a couple of comments otherwise lgtm

.eslintrc.cjs
```diff
@@ -105,6 +105,26 @@ module.exports = {
                 'posthog-js/no-unsafe-web-global': 'error',
             },
         },
+        {
+            files: [
```
marandaneto, line 109: nice

packages/core/src/posthog-core-stateless.ts
turnipdabeets: reverted public and called handles wrap now

packages/react-native/src/posthog-rn.ts
turnipdabeets: yep!

PR #3476, opened 2026-04-24
feat(node): add evaluateFlags() API for single-call flag evaluation

dmarticus:
## Problem

Phase 1 + Phase 2 of the Server SDK Feature Flag Evaluations RFC for `posthog-node`. Companion to the Python SDK PR ([PostHog/posthog-python#539](https://github.com/PostHog/posthog-python/pull/539)).

Today every flag check fires its own `/flags` request, and `capture({ sendFeatureFlags: true })` silently fires yet another on every captured event. The flag values on a captured event can diverge from the ones the code actually branched on when person/group properties differ between calls. `sendFeatureFlags` also attaches every evaluated flag to every event, which bloats properties on high-volume events.

### New API (Phase 1)

`posthog.evaluateFlags(distinctId, ...)` returns a `FeatureFlagEvaluations` snapshot:

```ts
const flags = await posthog.evaluateFlags(distinctId, { personProperties: { plan: 'enterprise' } })
if (flags.isEnabled('new-dashboard')) {
  renderNewDashboard()
}
posthog.capture({ distinctId, event: 'page_viewed', flags })
```

A single `/flags` request powers both branching and event enrichment. `isEnabled()` and `getFlag()` fire `$feature_flag_called` events (deduped through the existing cache) with the full metadata — `$feature_flag_id`, `$feature_flag_version`, `$feature_flag_reason`, `$feature_flag_request_id` — so experiment exposure tracking keeps working.

### Two layers of scoping

- **Network-level (`flagKeys` option):** scopes the underlying `/flags` request itself.

  ```ts
  const flags = await posthog.evaluateFlags(distinctId, { flagKeys: ['new-dashboard', 'checkout-flow'] })
  ```

- **Event-level (filter helpers):** narrow which flags get attached to a captured event without re-fetching.

  ```ts
  posthog.capture({ ..., flags: flags.onlyAccessed() })           // only flags the developer checked
  posthog.capture({ ..., flags: flags.only(['new-dashboard']) })  // specific keys
  ```

  `onlyAccessed()` honors its name — if nothing has been accessed, it returns an empty snapshot (no fallback to all flags). `only([...])` warns and drops unknown keys; the warning is silenceable via the `featureFlagsLogWarnings: false` SDK option.

### Slices are for capture, not branching

`onlyAccessed()` / `only([...])` return filtered snapshots intended for `capture()`. Calling `isEnabled()` / `getFlag()` on a slice for a key that was filtered out is a no-op (no event fires) — the flag wasn't actually missing, it was excluded from the slice.

### Granular `$feature_flag_error` reporting

Response-level errors (`errors_while_computing_flags`, `quota_limited`) are propagated into `$feature_flag_called` events from the snapshot. A missing flag during a quota-limited response now reports `quota_limited,flag_missing` instead of just `flag_missing`, matching the single-flag path's granularity.

### Exception captures carry flag context

`captureException()` and `captureExceptionImmediate()` accept an optional `flags` argument so `$exception` events carry the same flag context as the rest of your request's events:

```ts
const flags = await posthog.evaluateFlags(distinctId)
try {
  riskyThing()
} catch (e) {
  posthog.captureException(e, distinctId, undefined, undefined, flags)
}
```

### `flags` vs `sendFeatureFlags` precedence

When both are passed, `flags` always wins and we log a warning so the precedence isn't surprising:

> `[PostHog] Both flags and sendFeatureFlags were passed to capture(); using flags and ignoring sendFeatureFlags.`

### Deprecation warnings (Phase 2)

The legacy single-flag surface keeps working but now emits a deduped `[PostHog] ... is deprecated` console warning the first time it's used:

- `getFeatureFlag()`
- `isFeatureEnabled()`
- `getFeatureFlagPayload()`
- `capture({ sendFeatureFlags })` (only when truthy)

`isFeatureEnabled` is restructured to call `_getFeatureFlagResult` directly instead of routing through `getFeatureFlag`, so a single user-level call emits exactly one warning instead of cascading two. The dedup is process-wide via `emitDeprecationWarningOnce`, matching Python's `warnings.warn` default-dedup behavior. Phase 3 (removal in next major) ships separately.

### Local evaluation

Transparent. When the poller resolves a flag, the snapshot carries `locally_evaluated: true` and reason `"Evaluated locally"`, matching what `getFeatureFlag()` emits today.

### Backwards compatibility

**No breaking changes.** All existing call paths return the same values they did before — the only behavior changes are:
1. The new deprecation console warnings (deduped per-process)
2. `onlyAccessed()` no longer falls back to all flags when nothing was accessed (empty input → empty output)

### Internals

`_getFeatureFlagResult` was refactored: the dedup + capture portion is extracted into `_captureFlagCalledEventIfNeeded`, which is shared between the single-flag path and the new `FeatureFlagEvaluations` object. Both paths now dedupe identically. `evaluateFlags` uses `getFeatureFlagDetailsStateless` (the rich-detail endpoint) rather than the bare values endpoint so the snapshot carries full per-flag metadata.

## Tests

`packages/node/src/__tests__/evaluate-flags.spec.ts` — 32 tests covering remote evaluation, local evaluation, filtering helpers, capture integration, `flagKeys` round-trip, empty-distinctId safety, error-granularity propagation, deprecation warning emission (with no-cascade verification), and `captureException`/`captureExceptionImmediate` flag forwarding.

Full Node SDK suite: zero regressions on this branch (the same pre-existing failures from `main` remain). `pnpm lint` clean.

---


packages/node/src/feature-flag-evaluations.ts
```diff
@@ -0,0 +1,335 @@
...
+  _getDistinctId(): string {
+    return this._distinctId
+  }
+
+  /**
+   * @internal
+   */
+  _getGroups(): Record<string, string | number> | undefined {
+    return this._groups
+  }
```
dustinbyrne: unused public methods

packages/node/src/client.ts
dustinbyrne, line 1133: We should JSDoc `@deprecated` these too

agent visibility and all

dustinbyrne (approved):
nice

packages/node/src/feature-flag-evaluations.ts
dmarticus: Good catch — dropped both `_getDistinctId()` / `_getGroups()` in 67b3a0263. Verified no callers across the repo before removing.

packages/node/src/client.ts
dmarticus: Done in 67b3a0263 — added JSDoc `@deprecated` tags to `getFeatureFlag`, `isFeatureEnabled`, and `getFeatureFlagPayload` on both the `PostHogBackendClient` impl (client.ts) and the `IPostHog` interface (types.ts). The runtime `console.warn` we already had handles users running code; the JSDoc tag adds IDE strike-through + agent-tooling visibility so code agents reading the public surface see the deprecation immediately, before any call is even made.

PR #3480, opened 2026-04-26
feat(logs): flush RN logs queue to /i/v1/logs with batching

turnipdabeets:
## Problem

`posthog-react-native` has no first-party way to ship records to PostHog's logs product (`/i/v1/logs`). Customers who want structured logs have to roll their own OTLP exporter, storage, lifecycle handling, and rate cap. Storage scaffold landed in #3462; this PR adds the public API and the flush pipeline.

## Changes

Adds manual log capture for React Native. Pipeline lives in a shared `PostHogLogs` class in `@posthog/core` (so the JS SDK can later reuse it); RN-specific behavior (mobile flush cadence, AppState transitions, OS-imposed shutdown windows, dedicated logs storage file, `os.*` resource attrs) is wired in `posthog-react-native`.

### Public API (RN)

- `posthog.captureLog({ body, level?, attributes?, trace_id?, span_id?, trace_flags? })`
- `posthog.logger.{trace,debug,info,warn,error,fatal}(body, attributes?)` — lazy + memoized
- `posthog.flushLogs(): Promise<void>` — events still flush via `posthog.flush()`; `shutdown()` drains both
- `new PostHog(key, { logs: PostHogLogsConfig })` — `serviceName` / `serviceVersion` / `environment` / `resourceAttributes`, buffering knobs (`flushIntervalMs`, `rateCapWindowMs`, `maxBufferSize`, `maxLogsPerInterval`, `maxBatchRecordsPerPost`), lifecycle budgets (`backgroundFlushBudgetMs`, `terminationFlushBudgetMs`), `beforeSend`
- Re-exports from `@posthog/core`: `BeforeSendLogFn`, `CaptureLogOptions`, `CaptureLogger`, `LogAttributes`, `LogAttributeValue`, `LogSeverityLevel`, `PostHogLogsConfig`

### Pipeline

`captureLog` → `beforeSend` → tumbling-window rate cap → OTLP record build → persisted queue (dedicated `.posthog-rn-logs.json`) → batched gzipped POST to `/i/v1/logs`.

- **413** halves `maxBatchRecordsPerPost` and retries the same records; ramps back up by 1 per healthy send. Drops a single oversized record at size 1 with an explicit warn.
- **Concurrent `flush()`** serialized via a single in-flight promise — head of queue cannot be sent twice.
- **Flush triggers**: timer, buffer-at-capacity, AppState change, manual `flushLogs()`, `shutdown()`. Background and termination flushes are time-bounded so a slow network cannot outlive the OS-imposed window.
- **Per-record context** tagged at capture time, not flush: `distinctId`, `sessionId`, `screenName`, `appState`, `activeFeatureFlags`.
- **Storage**: `LogsQueue` routes to a dedicated `_logsStorage` file, separate from the events queue. Preserved across `reset()`. `waitForPersist()` runs between batches so a crash after a successful HTTP send but before the queue-advance hits disk cannot replay records.

### Gating — important to note

**Manual capture is unconditional.** Calling `captureLog` / `logger.*` ships records, matching the events pipeline's manual `capture()` shape. Only blockers: `optedOut`, missing/empty `body`, and missing API key. The server cannot remotely block manual capture — same as events.

**The wire field `response.logs.captureConsoleLogs` is browser-only.** It gates the JS SDK's `console.*` autocapture extension (per the project setting copy: *"Automatically capture browser console logs"*). RN does not read this field today. When console autocapture lands on RN as a follow-up, that PR will introduce a local opt-in via the same field for the autocapture path specifically — matching `<PostHogProvider autocapture>` and `captureAppLifecycleEvents` for events. Manual capture will remain unconditional.

Earlier iterations of this PR had a local `logs.captureConsoleLogs: true` opt-in flag plus a tri-state remote kill switch. Both removed:
- The local flag was misleading — the name promised console interception but autocapture isn't part of this PR.
- The remote kill switch conflated two concerns. The UI toggle is named for browser autocapture; using it as a kill switch for RN manual capture meant a customer flipping off web autocapture would also silently kill their RN manual logs, with no UI path to opt out of that side effect.
- Both can be re-introduced cleanly in the autocapture follow-up where they actually fit the wire-field semantic.

> Naming: `posthog.captureLog()` (this PR) is unrelated to the pre-existing `sessionReplayConfig.captureLog` boolean which controls session replay's `console.*` recording.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

Plus `@posthog/core` (minor bump in the same changeset).

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (`.changeset/rn-manual-log-capture.md`)

### Test plan

**Automated**
- `@posthog/core` `src/logs/index.spec.ts` — 54/54 passing.
- `posthog-react-native` `posthog.spec.ts` — 94/94 passing.

**Manual — Android, example-expo-53, Project 381971** (HogQL + logcat + on-disk file inspection)

<details>
<summary>Phase-by-phase breakdown — 9 phases ✅, 1 skipped (kill switch removed)</summary>

| # | What | Result |
|---|---|---|
| 1 | Per-level taps + Mixed + manual `Flush now`; default-INFO when no level provided | ✅ |
| 2 | OTLP resource attrs (`service.*`, `os.*`, `telemetry.sdk.*`); user `resourceAttributes` overrides defaults | ✅ |
| 3 | `screen.name` via `posthog.screen()`, `app.state=foreground`, `posthogDistinctId`, `sessionId` tagged at capture time (`feature_flags` covered by unit tests; project has no flags) | ✅ |
| 4 | `identify('alice')` / `reset()` / `optOut()` / `optIn()` — verified via distinctId tagging across captures | ✅ |
| 5 | `Flood 600`: rate cap held (max_idx=499, single warn line), 400 buffer evictions logged, records spanning full rate-cap window arrived | ✅ |
| 6 | Tap 3 → home button → records arrive before the 10s timer would fire (`flushWithTimeout(backgroundFlushBudgetMs)` drained inside the bg-task window); AppState log itself tags `app.state=background` correctly | ✅ |
| 7 | _Kill switch removed from this PR — skipped_ | n/a |
| 8 | Airplane mode → 5 captures → `length=5` on disk → network back → `Flush now` → all 5 land with original capture timestamps; network failure classified as `retry-later` (records retained) | ✅ |
| 9 | Empty/no body silently dropped; `beforeSend → null` logs and drops; `beforeSend → throw` logs error but record still arrives (chain continues with original input) | ✅ |
| 10 | 10 captures → `am force-stop` → relaunch → records replay (durability ✅). Reveals **at-least-once delivery**: kill timing between server ACK and `_persistQueueAdvance` disk-write produces duplicates on relaunch — same trade-off as events pipeline | ✅ with caveat |

</details>

Crash-safety: at-least-once delivery, matching events. Records durable across hard kills. Specific kill window (server ACK → disk-advance) can produce duplicates on next flush. Acceptable trade — tombstone-first / at-most-once would lose records on crash.

Earlier iOS sim dogfood (same project) verified the same paths plus 413 retry/ramp-up: temporary `_sendLogsBatch` wrapper forced `too-large` on first call → `Received 413 when sending logs batch of size 20, reducing batch size to 10` → two HTTP 200s with halved batch + linear ramp-up → 20 records delivered, no duplicates.

### DIFF
--- a/packages/core/src/logs/index.ts
+++ b/packages/core/src/logs/index.ts
@@ -248,13 +248,18 @@ export class PostHogLogs {
    * Mirrors the web SDK's resource attribute layout
    * (`packages/browser/src/posthog-logs.ts:163`). `service.name` is always
    * present; `environment` / `serviceVersion` only appear when configured;
-   * `resourceAttributes` spreads last so user keys win.
+   * `telemetry.sdk.*` is OTLP-standard and identifies which client emitted
+   * the batch (most logs backends index on it for SDK-version dashboards
+   * and bug-correlation). `resourceAttributes` spreads last so user keys
+   * win on any conflict.
    */
   private _buildResourceAttributes(): Record<string, LogAttributeValue> {
     return {
       'service.name': this._config.serviceName || 'unknown_service',
       ...(this._config.environment && { 'deployment.environment': this._config.environment }),
       ...(this._config.serviceVersion && { 'service.version': this._config.serviceVersion }),
+      'telemetry.sdk.name': this._instance.getLibraryId(),
+      'telemetry.sdk.version': this._instance.getLibraryVersion(),
       ...this._config.resourceAttributes,
     }
   }
--- a/packages/core/src/logs/index.ts
+++ b/packages/core/src/logs/index.ts
@@ -63,9 +63,23 @@ export class PostHogLogs {
    * Manual `captureLog` and `logger.*` are otherwise unconditional (matching
    * the browser SDK's manual path). The kill switch is the one mobile-only
    * lever — useful for shutting down a runaway log flood remotely without
-   * shipping an app update. An undefined value never overwrites a previous
-   * explicit decision so a remote response that omits the key can't silently
-   * re-enable a feature the server explicitly disabled.
+   * shipping an app update.
+   *
+   * Best-effort, eventually consistent: between SDK init and the first
+   * remote-config response, `_remoteEnabled` is `undefined` and capture is
+   * allowed. Records emitted in that window will enqueue and flush even if
+   * the customer's project has the switch on. Hosts that cache the previous
+   * remote-config response (RN does, see `posthog-rn.ts`) cover second-and-
+   * later launches; the very first launch and any post-`reset()` of the
+   * cached response are uncovered.
+   *
+   * Monotonically biased toward "allow": `setRemoteEnabled(undefined)` is a
+   * no-op and never overwrites a previous explicit decision. If the server
+   * has flipped the kill switch on (`false`) and later wants to release it,
+   * it must respond with an explicit `true` — removing the key from the
+   * response leaves the kill switch in place. This prevents a flaky or
+   * partially-populated response from silently re-enabling a feature the
+   * server explicitly disabled.
    */
   setRemoteEnabled(enabled: boolean | undefined): void {
     if (enabled !== undefined) {
@@ -322,10 +336,15 @@ export class PostHogLogs {
   }
 
   private _enqueue(entry: BufferedLogEntry): void {
-    // Re-check: optedOut can flip between captureLog and here — preload may
-    // have hydrated the real persisted value, or optIn/optOut may have fired
-    // while this fn was deferred.
+    // Re-check both gates: state can flip between captureLog and here.
+    // optedOut: preload may have hydrated the real persisted value, or
+    //   optIn/optOut may have fired while this fn was deferred.
+    // _isCaptureEnabled: a remote-config response that flipped the kill
+    //   switch may have landed during the same init-promise tick window.
     if (this._instance.optedOut) {
+      return
+    }
+    if (!this._isCaptureEnabled()) {
       return
     }
 
--- a/packages/core/src/logs/index.ts
+++ b/packages/core/src/logs/index.ts
@@ -7,9 +7,17 @@ import type { BeforeSendLogFn, BufferedLogEntry, CaptureLogOptions, ResolvedPost
 
 export class PostHogLogs {
   private _localEnabled: boolean
+  // Tri-state remote flag:
+  //   - undefined → remote hasn't spoken yet; local config decides.
+  //   - true      → remote opt-in; capture allowed even if local is off.
+  //   - false     → remote kill-switch; capture forbidden even if local is on.
+  // Updated via `setRemoteEnabled()` from the host SDK's `onRemoteConfig`.
+  private _remoteEnabled?: boolean
   private _maxBufferSize: number
   private _flushIntervalMs: number
-  // Mutable — halved on 413 to shrink the next POST.
+  // Mutable — halved on 413 to shrink the next POST, and ramped back up by
+  // one record after each successful send so a one-off oversized payload
+  // (e.g. a giant stack trace) doesn't permanently degrade throughput.
   private _maxBatchRecordsPerPost: number
   private _flushTimer?: ReturnType<typeof safeSetTimeout>
   // Serializes concurrent flushes — the second caller awaits the first rather
@@ -39,7 +47,7 @@ export class PostHogLogs {
     // wires this to its dedicated `_logsStorage.waitForPersist()`.
     private readonly _waitForStoragePersist: () => Promise<void> = () => Promise.resolve()
   ) {
-    this._localEnabled = _config.enabled !== false
+    this._localEnabled = _config.captureConsoleLogs === true
     this._maxBufferSize = _config.maxBufferSize
     this._flushIntervalMs = _config.flushIntervalMs
     this._maxBatchRecordsPerPost = _config.maxBatchRecordsPerPost
@@ -47,8 +55,35 @@ export class PostHogLogs {
     this._maxLogsPerInterval = _config.maxLogsPerInterval
   }
 
+  /**
+   * The host SDK calls this from its `onRemoteConfig` handler with whatever
+   * `response.logs.captureConsoleLogs` resolves to:
+   *   - `true`  — server-side opt-in; capture allowed even if local is off.
+   *   - `false` — server-side kill-switch; capture forbidden even if local
+   *               is on.
+   *   - `undefined` — no opinion; local config decides.
+   */
+  setRemoteEnabled(enabled: boolean | undefined): void {
+    if (enabled !== undefined) {
+      this._remoteEnabled = enabled
+    }
+  }
+
+  /**
+   * Effective enabled state for the hot path. Either side can opt in; remote
+   * `false` is a hard kill-switch that overrides local. This matches the JS
+   * SDK except for the explicit kill-switch — browser only treats remote as
+   * an additional on-switch, not an off-switch.
+   */
+  private _isCaptureEnabled(): boolean {
+    if (this._remoteEnabled === false) {
+      return false
+    }
+    return this._localEnabled || this._remoteEnabled === true
+  }
+
   captureLog(options: CaptureLogOptions): void {
-    if (!this._localEnabled) {
+    if (!this._isCaptureEnabled()) {
       return
     }
     if (this._instance.optedOut) {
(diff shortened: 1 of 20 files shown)


ioannisj:
Taking a look in a bit 🙏

packages/core/src/logs/index.ts
```diff
@@ -1,26 +1,101 @@
...
+
+  // Fixed-window rate cap. Tumbling (not sliding) for cheap arithmetic on the
+  // hot path. Window rolls the first time `captureLog` fires after the window
+  // expires — no background timer needed. `_droppedWarned` keeps the log noise
+  // to one line per window regardless of how many records got dropped.
+  private _rateCapWindowMs: number
+  private _maxLogsPerInterval?: number
+  private _intervalWindowStart = 0
+  private _intervalLogCount = 0
+  private _droppedWarned = false
```
marandaneto, line 34: do we need all of this? We don't seem to have this for the batch endpoint
i'd avoid extra complexity if the other approach is simpler and works well

packages/core/src/logs/index.ts
```diff
@@ -1,26 +1,101 @@
...
+   * The host SDK calls this from its `onRemoteConfig` handler with whatever
+   * `response.logs.captureConsoleLogs` resolves to:
+   *   - `false` — server-side kill-switch; capture forbidden even if local
+   *               opted in.
+   *   - `true`  — server has no objection; effective state still requires a
+   *               local opt-in.
+   *   - `undefined` — no opinion; local config decides.
+   *
+   * AND semantics with local: this mirrors the RN session-replay
+   * `consoleLogRecordingEnabled` gate (`packages/react-native/src/posthog-rn.ts:1813-1825`)
```
marandaneto: i'd avoid references of file names and line numbers since this gets outdated very fast
if you think this is important for the reader, add a github permlink

packages/core/src/logs/index.ts
```diff
@@ -30,15 +105,234 @@ export class PostHogLogs {
       return
     }
 
+    // Ordering: beforeSend → rate cap → OTLP build. beforeSend runs first so
+    // user-dropped records don't consume the per-interval budget (documented
+    // in `logs/types.ts`). Matches the events pipeline's `before_send`
+    // semantics in `packages/core/src/posthog-core.ts:1478-1499`.
+    const filtered = this._runBeforeSend(options)
```
marandaneto, line 66: do we also run before send for logs on the web?

marandaneto:
I like code comments, but this PR has a lot and they're quite large. This tends to go outdated and most people wont read it since they are again, huge
i dont know if this is a bad thing now since agents do the work
@ioannisj wdyt?

packages/core/src/logs/index.ts
turnipdabeets: ah agreed, I tried to removed some of these but missed others. I think I'll have to add this to memory

packages/core/src/logs/index.ts
turnipdabeets: No, browser doesn't use a before_send for logs, but RN and browser use it for events. Do you think we should follow the browser on this?

marandaneto: or run `before_send` for the browser sdk
i guess users will want a way to filter out logs? check with the logs team, maybe theres a reason not to

packages/core/src/logs/index.ts
turnipdabeets: This is mirroring the approach in browser for auto capture `packages/browser/src/posthog-logs.ts`. Auto capture can produce huge volume by logging from dependencies. But I don't have the full auto capture in this PR for logs in RN yet.

packages/core/src/logs/index.ts
turnipdabeets: @PostHog/logs was there a reason to leave out send_before on logs?

packages/core/src/logs/index.ts
marandaneto: i think RN wont have this problem, Node might have it, so i guess its worth keeping them

packages/react-native/src/posthog-rn.ts
```diff
@@ -42,6 +45,46 @@ import { ErrorTracking, ErrorTrackingOptions } from './error-tracking'
...
+ * are treated as foreground — the app is still running JS, just not the
+ * primary scene. 'unknown' returns undefined so the attribute is omitted
+ * rather than guessed.
+ */
+/**
+ * Extracts the logs gate from a remote config response. Accepts the same
+ * `boolean | { captureConsoleLogs?: boolean }` shape the server may return,
+ * and surfaces the tri-state expected by `PostHogLogs.setRemoteEnabled`:
+ *   - `undefined`              → no opinion; leave local in charge
+ *   - `true`                   → server-side opt-in (forces on)
```
ioannisj: I think actual implementation here differs since when true, we respect the local config

packages/core/src/logs/index.ts
ioannisj: Yeah lets keep even if this complicates implementation a bit

packages/core/src/logs/types.ts
```diff
@@ -29,11 +29,26 @@ export interface BufferedLogEntry {
...
+ * transformed) record to keep it, or `null` to drop it. Configure as a single
+ * function or an array (chain of filters, evaluated left-to-right).
+ */
+export type BeforeSendLogFn = (record: CaptureLogOptions) => CaptureLogOptions | null
+
 // Public configuration for the logs module. Per-SDK defaults diverge (mobile
 // cellular radio cost, browser tab suspension, node process lifecycle).
 export interface PostHogLogsConfig {
-  // Master switch. Default: true when a config object is provided.
-  enabled?: boolean
```
ioannisj, line 36: This was not public api yet right?

packages/core/src/logs/index.ts
ioannisj: I'd keep it, especially for logs that can be very chatty

ioannisj (approved):

ioannisj:
> I like code comments, but this PR has a lot and they're quite large. This tends to go outdated and most people wont read it since they are again, huge i dont know if this is a bad thing now since agents do the work @ioannisj wdyt?

Left some comments but overall I think it's solid. On the comments, yeah made it harder for me to follow the code, but could be helpful in keeping agents in line. Since I think this will be a blueprint implementation for the other SDKs, let's see how much it helps

packages/core/src/logs/types.ts
turnipdabeets: It's not used yet. It was introduced under RN in https://github.com/PostHog/posthog-js/pull/3453 but moved to core here.

packages/react-native/src/posthog-rn.ts
turnipdabeets: updated

turnipdabeets: FYI I also removed the auto capture bool in this PR to make it manual capture only so this follows events where remote config isn't needed on manual logs

packages/core/src/logs/types.ts
```diff
@@ -54,13 +54,23 @@ export interface PostHogLogsConfig {
...
+// `rateCapWindowMs` is always resolved so the rate-cap arithmetic doesn't
+// branch at the hot path. `maxLogsPerInterval` stays optional — `undefined`
+// means unbounded.
 export interface ResolvedPostHogLogsConfig extends PostHogLogsConfig {
   maxBufferSize: number
+  flushIntervalMs: number
+  maxBatchRecordsPerPost: number
+  rateCapWindowMs: number
+  backgroundFlushBudgetMs: number
+  terminationFlushBudgetMs: number
```
marandaneto, line 190: are these public APIs?
should we start with them private and make them public if needed? i dont think its clear what they do if you dont read the source code, so maybe thats a bit confusing

marandaneto (approved):

packages/core/src/logs/types.ts
turnipdabeets: moved 2 to be private

PR #3482, opened 2026-04-27
fix: fallback when survey svg icons unavailable

marandaneto:
## Problem

React Native apps can use survey UI without having `react-native-svg` installed or fully linked. The SDK imported `react-native-svg` directly for survey icons, which could either crash module resolution or render React Native's `Unimplemented` placeholder when the JS package existed but native SVG components were unavailable.

## Changes

- Adds an optional `react-native-svg` loader that only enables SVG icons when the native SVG view managers needed by the survey icons are registered.
- Updates survey icons to use a type-only `SvgProps` import and fall back to simple text labels when SVG is unavailable.
- Marks `react-native-svg` as an optional peer dependency.
- Adds a changeset for `posthog-react-native`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/react-native/package.json
```diff
@@ -130,6 +130,9 @@
     },
     "react-native-safe-area-context": {
       "optional": true
+    },
+    "react-native-svg": {
```
marandaneto, line 134: this was missing

packages/react-native/src/surveys/icons.tsx
```diff
@@ -1,61 +1,201 @@
...
+  )
+}
+
+export const SatisfiedEmoji = (props: SvgProps): JSX.Element =>
+  renderSvgIcon({
+    props,
+    height: '48',
+    viewBox: '0 -960 960 960',
+    width: '48',
+    fallback: ':)',
```
marandaneto, line 92: if SVG lib isnt available at runtime, then we fallback to ASCII emojis

ioannisj: We could use unicode emojis here?

ioannisj (approved):
LG, minor comment + fix lint error on CancelSVG?

packages/react-native/src/surveys/icons.tsx
marandaneto: tried and didnt work either (depends on device/simulator having an usable emoji font - my emualtor rendered as '?'), this is a fallback anyway, our installation steps require the svg lib so its less concerning (better than a crash or not showing at all)

PR #3492, opened 2026-04-27
feat(surveys): support translations on react native

lucasheriques:
## Summary
- add survey translation support to `posthog-react-native`
- move the shared survey translation model and pure translation logic into `@posthog/core`
- keep browser using the shared core logic and remove the old browser-only language matcher
- include `$survey_language` in React Native survey shown/sent/dismissed events
- add shared survey response aliases in `@posthog/core` and reuse them across browser and React Native survey event code

## Why
The React Native SDK lives in the same repo and consumes the same survey payload shape, so keeping translation matching and translation application browser-local was unnecessary duplication. Moving the pure logic into `@posthog/core` gives both SDKs a single implementation while leaving platform-specific language detection at the SDK layer.

This also follows up on the survey event typing drift between browser and React Native by centralizing the shared survey response value/response map aliases in core instead of repeating inline record types.

## Testing
- `pnpm --dir packages/core test:unit -- translations.spec.ts --runInBand`
- `pnpm --dir packages/core build`
- `pnpm --dir packages/browser exec jest src/__tests__/utils/survey-translations.test.ts src/__tests__/extensions/surveys.test.ts --runInBand`
- `pnpm --dir packages/browser build`
- `pnpm --dir packages/react-native exec jest -c jest.config.js test/surveyTranslations.spec.ts test/sendSurveyEvent.spec.ts --runInBand`
- `pnpm --dir packages/react-native build`


packages/react-native/src/posthog-rn.ts
```diff
@@ -115,6 +115,13 @@ export interface PostHogOptions extends PostHogCoreOptions {
    * @default true
    */
   setDefaultPersonProperties?: boolean
+
+  /**
+   * Overrides the language used when rendering translated survey copy.
+   * When unset, the SDK falls back to the persisted person property `language`
+   * and then the device locale.
+   */
+  override_display_language?: string | null
```
turnipdabeets: would be nice to keep camelCase here

packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -413,7 +413,7 @@ interface SendSurveyEventArgs {
     surveyLanguage?: string | null
 }
 
-const getSurveyResponseValue = (responses: Record<string, string | number | string[] | null>, questionId?: string) => {
+const getSurveyResponseValue = (responses: SurveyResponses, questionId?: string) => {
```
turnipdabeets: can we move helps getSurveyResponseValue, buildSurveyResponseProperties, surveyHasResponses, and getSurveyInteractionProperty into core now?

packages/core/src/surveys/translations.ts
```diff
@@ -0,0 +1,244 @@
...
+): string | null {
+  if (!translations || !targetLanguage) {
+    return null
+  }
+
+  const normalizedTarget = normalizeLanguageCode(targetLanguage)
+
+  const exactMatch = Object.keys(translations).find((key) => normalizeLanguageCode(key) === normalizedTarget)
+  if (exactMatch) {
+    logger.info(`Found exact translation match: ${exactMatch}`)
```
turnipdabeets: if it catches every match should this be a debug log instead?

packages/core/src/surveys/translations.ts
```diff
@@ -0,0 +1,244 @@
...
+}
+
+type TranslatableSurvey<TQuestion extends TranslatableSurveyQuestion = TranslatableSurveyQuestion> = {
+  name: string
+  translations?: Record<string, SurveyTranslation>
+  appearance?: TranslatableSurveyAppearance | null
+  questions: TQuestion[]
+}
+
+function mergeQuestionTranslation<TQuestion extends TranslatableSurveyQuestion>(
```
turnipdabeets, line 129: would be nice to have a unit test for this

packages/react-native/src/surveys/components/Surveys.tsx
```diff
@@ -15,6 +16,8 @@ import { LinkQuestion, MultipleChoiceQuestion, OpenTextQuestion, RatingQuestion
 import { PostHog } from '../../posthog-rn'
 import { usePostHog } from '../../hooks/usePostHog'
 
+const SURVEY_LANGUAGE_PROPERTY = '$survey_language'
```
turnipdabeets: nit: `$survey_language` is spread before the response props in sendSurveyShownEvent / sendSurveyEvent but after buildSurveyResponseProperties(...) in dismissedSurveyEvent. Since it's survey metadata (like `$survey_id` / `$survey_iteration`), can we put it next to those in all three handlers for consistency?

turnipdabeets (commented):
LGTM just a few comments

packages/react-native/src/posthog-rn.ts
lucasheriques: fixed!

packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
lucasheriques: moved, makes sense

packages/core/src/surveys/translations.ts
lucasheriques: agree, changed to debug too

packages/core/src/surveys/translations.ts
lucasheriques: added!

packages/react-native/src/surveys/components/Surveys.tsx
lucasheriques: done too

lucasheriques:
@turnipdabeets thanks for all the helpful comments! addressed all of them. missed your approval to merge it though

this will come soon to the iOS/Android SDKs too, soon 🙌

turnipdabeets (approved):

PR #3493, opened 2026-04-27
perf(rn): coalesce per-tick storage writes

turnipdabeets:
## Summary

Separating this PR since it also affect events:

Each `PostHogRNStorage` mutation (`setItem`, `removeItem`, `clear`) currently `JSON.stringify`s the full `memoryCache` and writes it to AsyncStorage on every call. A sync burst of N captures is N writes of progressively larger blobs — cumulative O(n²) bytes, plus an AsyncStorage round-trip per capture. Native iOS/Android avoid this by using one file per event; RN's single-blob shape (inherited from the JS storage abstraction) is where the cost lives.

This PR adds **tick-level write coalescing**: a mutation updates `memoryCache` synchronously and arms a single `setTimeout(0)` that fires `persist()` once per tick with the final state. Multiple mutations within the same tick collapse into one disk write. `persist()` body unchanged — same `JSON.stringify`, same `_pendingPromises` tracking, same error handling.

`waitForPersist()` drains any scheduled (not yet fired) write **synchronously** before awaiting in-flight async writes. This preserves the events-flush durability contract documented at `posthog-core-stateless.ts:1131` (*"Wait for storage to complete to prevent duplicate events on app crash"*) — `flushStorage` callers continue to get the same guarantees.

Helps both events and logs equally — they share the `PostHogRNStorage` class.

## Why this matters

Events were already paying O(n²) write cost per sync burst, but real apps don't capture events at sync-burst speed. Logs are structurally more burst-prone (especially once auto capture, `captureConsoleLogs` ships on RN — every `console.log` becomes a capture, including third-party library output). This PR brings RN's per-capture I/O cost shape to parity with native iOS/Android.

## Defensive call-site additions

Two small defensive changes in `posthog-rn.ts` to keep events durable under coalescing:

1. **AppState handler**: drain `_eventsStorage` alongside `_logsStorage`. The AppState transition is itself a macrotask whose ordering vs. our scheduled `setTimeout(0)` is undefined; without this, a capture landing in the same tick as a foreground→background transition could lose its scheduled write to OS suspension.

2. **`_shutdown` override**: drain both storages after `super._shutdown` completes. The events shutdown loop only awaits `flushStorage` when it successfully advances the queue; `setPersistedProperty` calls for `distinctId`, `sessionId`, `deviceId`, feature flag overrides, etc. wouldn't otherwise be drained on process exit.

## Events-safety audit

Pre-implementation audit walked every storage call site to confirm correctness contracts are preserved:

- ✅ `_flush()` queue-advance contract (`posthog-core-stateless.ts:1108-1198`) — preserved by `waitForPersist`'s synchronous drain.
- ✅ Cold-start preload — read-side, untouched.
- ✅ Bypass-`memoryCache` reads — none. Only `populateMemoryCache` reads AsyncStorage directly.
- ✅ Multi-key sync writes — coalesce into one final-state write. Strictly fewer writes, identical observable end state.
- ✅ Concurrent `persist()` — `_pendingPromises` set tracks all in-flight writes regardless of who triggered them.
- ⚠️ AppState background — gap closed by adding `_eventsStorage.waitForPersist()` (above).
- ⚠️ Process exit — gap closed by adding both drains to `_shutdown` (above).

## Test changes

- **`storage.spec.ts`**: 4 new tests covering coalescing behavior, scheduled-write draining, post-failure scheduling recovery; 2 existing tests updated to `await waitForPersist()` before asserting disk state (they previously relied on per-call sync-persist).
- **`posthog.spec.ts`**: 6 sync-init session-id tests and 1 async-init `beforeEach` updated to `await rnStorage.waitForPersist()` after seeding storage; 1 feature-flag override test does the same. These tests previously relied on undocumented per-call sync-persist behavior — the new pattern explicitly waits.

## Verification

- All 307 RN tests pass.
- All 509 core tests pass.
- Lint clean.
- TS build clean.

## Test plan

- [x] `pnpm test` (RN package)
- [x] `pnpm test` (core package)
- [x] `pnpm exec eslint src test`
- [x] `pnpm turbo build --filter=posthog-react-native`
- [ ] Manual simulator smoke test in `examples/example-expo-53`:
  - capture event → background app → foreground → confirm event reaches PostHog UI
  - capture event → force-quit → relaunch → confirm event sent on next session

## Rollback

Single-commit, three files. `git revert` is safe — no schema/API changes, no migration.

## Something we should also do

- **Smart-erase eviction** (level-priority eviction in `core/logs/index.ts`) — separate logs-only fix for the *Error+stack evicted by Flood-600* regression. Will land in a follow-up PR targeting the same base branch.

### DIFF
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -1,4 +1,4 @@
-import { AppState, Dimensions, Linking, Platform } from 'react-native'
+import { AppState, type AppStateStatus, Dimensions, Linking, Platform } from 'react-native'
 
 import {
   CaptureLogOptions,
@@ -44,6 +44,23 @@ import { OptionalReactNativeSessionReplay } from './optional/OptionalSessionRepl
 import { ErrorTracking, ErrorTrackingOptions } from './error-tracking'
 
 export { PostHogPersistedProperty }
+
+/**
+ * Collapses RN's broader AppState status set into the OTLP `app.state`
+ * enum (foreground|background). 'inactive' (iOS transition) and 'extension'
+ * are treated as foreground — the app is still running JS, just not the
+ * primary scene. 'unknown' returns undefined so the attribute is omitted
+ * rather than guessed.
+ */
+function mapAppStateForLogs(state: AppStateStatus | undefined): 'foreground' | 'background' | undefined {
+  if (state === 'background') {
+    return 'background'
+  }
+  if (!state || state === 'unknown') {
+    return undefined
+  }
+  return 'foreground'
+}
 
 export interface PostHogOptions extends PostHogCoreOptions {
   /**
@@ -150,6 +167,11 @@ export class PostHog extends PostHogCore {
   private _disableRemoteConfig: boolean
   private _errorTracking: ErrorTracking
   private _logs: PostHogLogs
+  // Cached, foreground/background view of the app's lifecycle. Read on the
+  // log-capture hot path (per record) so we tag every log with whether it
+  // happened in foreground or background. Updated by the AppState listener
+  // and seeded from `AppState.currentState` at construction.
+  private _currentAppState?: 'foreground' | 'background'
   private _surveysReadyPromise: Promise<void> | null = null
   private _surveysReady: boolean = false
   private _setDefaultPersonProperties: boolean
@@ -241,14 +263,29 @@ export class PostHog extends PostHogCore {
       this._logsStorage = createLogsMemoryStorage()
     }
 
+    // Seed from sync `AppState.currentState` so the very first capture (which
+    // can happen before any 'change' event fires) is already tagged. Maps
+    // RN's broader status set into the OTLP `app.state` enum's
+    // foreground/background dichotomy.
+    this._currentAppState = mapAppStateForLogs(AppState.currentState)
+
     this._logs = new PostHogLogs(
       this,
       resolveLogsConfig(options?.logs),
       this._logger,
-      () => ({
-        distinctId: this.getDistinctId() || undefined,
-        sessionId: this.getSessionId() || undefined,
-      }),
+      () => {
+        // Pulled at capture time so each tag reflects state at the moment
+        // the log was fired, not at flush.
+        const flags = this.getFeatureFlags()
+        const flagKeys = flags ? Object.keys(flags) : undefined
+        return {
+          distinctId: this.getDistinctId() || undefined,
+          sessionId: this.getSessionId() || undefined,
+          screenName: (this.sessionProps?.$screen_name as string | undefined) || undefined,
+          appState: this._currentAppState,
+          activeFeatureFlags: flagKeys && flagKeys.length > 0 ? flagKeys : undefined,
+        }
+      },
       (fn) => this.wrap(fn),
       // Block between batches on the logs-storage disk write so a crash can't
       // replay an already-sent batch. Events do the equivalent via
@@ -261,6 +298,14 @@ export class PostHog extends PostHogCore {
       // ignore unknown state (usually initial state, the app might not be ready yet)
       if (state === 'unknown') {
         return
+      }
+
+      // Update before kicking off the flush — captures that race the flush
+      // (e.g. fired in a `componentWillUnmount` triggered by backgrounding)
+      // should already see the new state.
+      const mapped = mapAppStateForLogs(state)
+      if (mapped) {
+        this._currentAppState = mapped
       }
 
       void this.flush().catch(async (err) => {
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -1,6 +1,8 @@
 import { AppState, Dimensions, Linking, Platform } from 'react-native'
 
 import {
+  CaptureLogOptions,
+  CaptureLogger,
   JsonType,
   PostHogCaptureOptions,
   PostHogCore,
@@ -9,6 +11,7 @@ import {
   PostHogFetchOptions,
   PostHogFetchResponse,
   PostHogLogs,
+  PostHogLogsConfig,
   PostHogPersistedProperty,
   PostHogRemoteConfig,
   Survey,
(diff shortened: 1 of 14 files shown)


packages/react-native/src/posthog-rn.ts
```diff
@@ -319,8 +319,16 @@ export class PostHog extends PostHogCore {
...
+      // Drain pending storage writes to disk before the OS may suspend
+      // the process. waitForPersist swallows errors internally and drains
+      // any tick-coalesced scheduled write synchronously before awaiting
+      // in-flight async writes. Both sides are needed: writes are
+      // tick-coalesced in `storage.ts:schedulePersist`, and the AppState
+      // transition is itself a macrotask whose ordering vs. our scheduled
+      // setTimeout(0) callback is undefined — without this, a capture
+      // landing in the same tick as a foreground→background transition
+      // could lose its scheduled write to OS suspension.
+      void this._eventsStorage.waitForPersist()
```
ioannisj, line 331: I may be misreading this or failing to completely understand the flow, but are we flushing and draining storage items on disk in parallel here? this could eventually lead to duplicate sends? And since now writes are coalesced, the prob of more than one event being duplicated is higher?

ioannisj (commented):
One concern is that now that we are batching writes, when persist() throws the effect could be bigger? Previously a persist() could fail but would naturally be retried on next event/write. Now it feels we are failing to write more event, but also retrying less frequently because of coalescing.

packages/react-native/src/posthog-rn.ts
turnipdabeets: RN needs to write to both a memoryCache and AsyncStorage. Previously we did this after every setItem. Now if a flood of events/logs happen at the same time we can schedule one disk write for all those changes in a short time. `setTimeout(0)` allows us to write "after the current synchronous code finishes, on the next macrotask."

> we flushing and draining storage items on disk in parallel here

Yes, in parallel. Duplicates would happen if network and disk got out of sync. They don't, because `_flush()` waits for the disk update to land before resolving (await this.flushStorage()). The parallel `waitForPersist()` doesn't send to the network or touch the queue, so it can't put them out of sync.

> Previously a persist() could fail but would naturally be retried on next event/write. Now it feels we are failing to write more event, but also retrying less frequently because of coalescing.

Every `persist()` writes the full memoryCache snapshot, so a failed write self-heals on the next successful one — same as pre-PR, just one sub-ms macrotask later. AppState/_shutdown force fresh persists at lifecycle boundaries so failures can't sit indefinitely (see test: keeps scheduling writes after a previous persist fails).

One real-but-tiny tradeoff: the in-memory→disk-initiated window is one macrotask wider for any storage mutation that isn't immediately followed by a flush. So a hard crash within that sub-ms window could lose the in-flight tick's mutations. But the window is small compared to the native-I/O latency the write was already exposed to.

marandaneto:
is this a current perf issue? Did anyone report a slowdown, or how do we identify the perf bottleneck?

packages/react-native/src/posthog-rn.ts
ioannisj: Okay got it, thanx for the detailed explanation. I think I had this backwards

ioannisj (commented):
Will probably want @dustinbyrne eyes here as well

packages/react-native/src/storage.ts
```diff
@@ -74,22 +94,58 @@ export class PostHogRNStorage {
...
+    this._persistTimer = setTimeout(() => {
+      // Clear handle before persist() so a sync throw can't leave the
+      // scheduler stuck. Async errors are caught inside persist() itself.
+      this._persistTimer = undefined
+      try {
+        this.persist()
+      } catch (err) {
+        console.warn('PostHog storage scheduled persist threw:', err)
+      }
+    }, 0)
```
dustinbyrne, line 114: Out of curiosity, is the main motivation here that RN log capture will make storage writes much more bursty than typical event capture?

Just trying to understand whether this is a specific mitigation for logs/console autocapture, versus a more general fix for RN queue storage write amplification?

packages/react-native/src/posthog-rn.ts
```diff
@@ -489,7 +497,20 @@ export class PostHog extends PostHogCore {
...
+      // distinctId, sessionId, deviceId, feature flag overrides, etc.
+      // only get a scheduled setTimeout(0) write; without an explicit
+      // drain here, those writes could miss disk if the process exits
+      // before the next macrotask. Runs in `finally` so a timed-out
+      // shutdown (super._shutdown rejects after shutdownTimeoutMs) still
+      // flushes pending writes before the rejection propagates.
+      // waitForPersist swallows errors internally and never throws.
+      await Promise.all([this._eventsStorage.waitForPersist(), this._logsStorage.waitForPersist()])
+    }
   }
```
dustinbyrne, line 514: Should we be taking `shutdownTimeoutMs` into account for these `waitForPersist` calls?

If one of the `shutdown` calls times out, we've already exceeded our "shutdown SLA". Maybe we race with the remaining time?

packages/react-native/src/storage.ts
turnipdabeets: This motivation is for RN logs specifically, and exactly that it'll be more bursty. This came up when testing on example app with a large burst of logs (Cc @marandaneto just from testing no report of slowdown). The assumption is auto capturing logs can be bursty.

turnipdabeets:
going to close this for now until I focus on auto capture for RN

whatsrupp:
I was doing some sentry profiling today and 

just letting you know this would probably be really good to fix from a consumer perspective!

This absolutely hammers low end Android for us when there's a lot of events fired simultaneously, can happen on navigation etc
(We're a react native - expo app)

Triggers a lot of JS work because of the full serialisation and then a lot of IO for the write to disk

We've changed to using a debounced MMKV and I think I'm going to probably try and add your changes around the the JSON.stringify into a patch

I'm not sure if there's a risk to that and don't really have context on why you do so much writing/ serialising but thought it would be good to share

Thanks!

turnipdabeets:
> just letting you know this would probably be really good to fix from a consumer perspective!


Thanks for this feedback. I've opened a new PR to address: https://github.com/PostHog/posthog-js/pull/3701

PR #3451, opened 2026-04-23
feat(posthog-ai): Support generic AI generation tracing beyond Vercel…

Danish2op:
## Description

This PR introduces `captureAiGeneration` to the `@posthog/ai` package, exposing a lower-level generic tracking function that bypasses the need for the Vercel AI SDK wrapper (`withTracing`). 

This allows capturing `$ai_generation` tracing events natively when using alternative AI calling mechanisms, such as:
- Direct `env.AI.run()` calls (e.g., Cloudflare Workers AI bindings)
- TanStack AI adapters (`@cloudflare/tanstack-ai`)
- Any non-Vercel AI SDK provider

The newly exposed function seamlessly delegates to the internal event senders (`sendEventToPosthog` and `sendEventWithErrorToPosthog`), ensuring the exact same `$ai_generation` event schema, properties, and analytical behavior as `withTracing`, but completely decoupled from a specific AI SDK.

Fixes #3444 

## Changes Made
- Added `captureAiGeneration` in `packages/ai/src/capture.ts`.
- Exported `captureAiGeneration` and its options interface (`CaptureAiGenerationOptions`) in `packages/ai/src/index.ts`.
- Added comprehensive unit tests in `packages/ai/tests/capture.test.ts` to ensure payload formatting, error handling, auto-generated trace IDs, and privacy mode work flawlessly.

## Usage

Consumers can now trace generations easily:

```typescript
import { captureAiGeneration } from '@posthog/ai'

await captureAiGeneration(posthogClient, {
  distinctId: 'user-123',
  traceId: 'trace-abc', // automatically generated if omitted
  provider: 'cloudflare-workers-ai',
  model: '@cf/zai-org/glm-4.7-flash',
  input: messages,
  output: responseText,
  usage: { inputTokens: 1200, outputTokens: 450 },
  latency: 2.3, // seconds
  properties: { feature: 'transcript-toc', jobId: '...' },
})

PR #3499, opened 2026-04-29
feat(posthog-ai): add captureAiGeneration as canonical capture primitive

### LINKED ISSUE #3444: feat(posthog-ai): Support generic AI generation tracing beyond Vercel AI SDK
## Problem

`@posthog/ai` currently only supports tracing via `withTracing()`, which wraps a Vercel AI SDK `LanguageModel` instance. There is no way to capture `$ai_generation` events when using other AI calling mechanisms:

- Direct `env.AI.run()` calls (Cloudflare Workers AI binding)
- TanStack AI adapters (`@cloudflare/tanstack-ai`)
- Any non-Vercel AI SDK provider

## Use case

We use Cloudflare Workers AI with reasoning models (GLM-4.7-flash, Gemma-4). The `workers-ai-provider` for Vercel AI SDK has a bug where `reasoning_effort` and `chat_template_kwargs` are silently dropped (https://github.com/cloudflare/ai/issues/501). To work around this, we need to call `env.AI.run()` directly — but then we lose PostHog AI tracing.

## Proposal

Expose a lower-level function that captures `$ai_generation` events without requiring a Vercel AI SDK `LanguageModel` wrapper. Something like:

```typescript
import { captureAiGeneration } from '@posthog/ai'

await captureAiGeneration(posthogClient, {
  distinctId: 'user-123',
  traceId: 'trace-abc',
  provider: 'cloudflare-workers-ai',
  model: '@cf/zai-org/glm-4.7-flash',
  input: messages,
  output: responseText,
  usage: { inputTokens: 1200, outputTokens: 450 },
  latency: 2.3, // seconds
  properties: { feature: 'transcript-toc', jobId: '...' },
})
```

This would emit the same `$ai_generation` event with the same `$ai_*` properties that `withTracing` produces, making it visible in the LLM Analytics dashboard — but decoupled from any specific AI SDK.

## Context

- TanStack AI is a growing alternative to Vercel AI SDK (especially in the TanStack Start ecosystem)
- Cloudflare Workers AI's direct binding is sometimes necessary when SDK adapters have gaps
- The existing `withTracing` wrapper covers the Vercel AI SDK path well, but the ecosystem is fragmenting

richardsolomou:
## Problem

Closes #3444. `@posthog/ai` had no way to emit `$ai_generation` events for LLM calls that don't go through one of the wrapped clients — direct Cloudflare Workers AI bindings, TanStack AI adapters, or custom HTTP clients all flew under the radar. PR #3451 took a first pass at this but layered a third capture surface on top of the existing two internal helpers, leaving us with public, semi-public, and private layers all mostly doing the same thing — and the public one missing model parameters, cost overrides, and a clean `error` story.

## Changes

- **New public primitive** `captureAiGeneration(client, options)` exported from `@posthog/ai`. It replaces the old internal `sendEventToPosthog` / `sendEventWithErrorToPosthog`. The options object mirrors the `$ai_generation` event shape directly: explicit `modelParameters`, `properties`, `groups`, `privacyMode`, `costOverride`, `modelOverride`, `providerOverride`, etc. — no more `params: <vendor body>` quirk, no more `MonitoringParams` cast.
- **Every wrapper migrated** to call `captureAiGeneration` (`withTracing`, OpenAI, Azure, Anthropic, GoogleGenAI). External callers and SDK wrappers now funnel through the same primitive, so events from `captureAiGeneration` are indistinguishable from those produced by `withTracing`.
- **Drops the double-`JSON.stringify(error)` bug** that several wrapper call sites hit by stringifying before passing to `sendEventWithErrorToPosthog` (which stringified again).
- **Caller-supplied `httpStatus` is preserved** even when `error` is present (e.g. rate-limit responses).
- README updated with a Cloudflare Workers AI example.

### Behavior notes for reviewers

- **Exception autocapture is now uniform across every error path.** Some non-streaming `Promise.then(success, failure)` paths in `openai/index.ts` and `openai/azure.ts` previously called `sendEventToPosthog({ error: JSON.stringify(error) })` and skipped autocapture. Now consistent with the streaming paths they sit alongside. Only material on clients with `enableExceptionAutocapture: true`.
- **Truthy error check preserved.** `null`/`0`/`false`/`''` skip the error path (matches the old `if (error)` semantics).

### Usage

```typescript
import { captureAiGeneration } from '@posthog/ai'

const phClient = new PostHog('<KEY>', { host: 'https://us.i.posthog.com' })

await captureAiGeneration(phClient, {
  distinctId: 'user_123',
  traceId: 'trace_abc',
  provider: 'cloudflare-workers-ai',
  model: '@cf/zai-org/glm-4.7-flash',
  input: messages,
  output: result.response,
  modelParameters: { reasoning_effort: 'high' },
  usage: { inputTokens: 1200, outputTokens: 450 },
  latency: 2.3,
  properties: { feature: 'transcript-toc' },
})
```

- 16 new tests in `packages/ai/tests/captureAiGeneration.test.ts` covering the canonical event shape, traceId / distinctId fallback, `captureImmediate`, privacy-mode redaction, the three error/httpStatus cases (parameterised), exception autocapture, model/provider overrides (with and without), cost overrides, embedding events, `$ai_tokens_source: passthrough` detection, the falsy-error skip, and the `!client.capture` early return.
- Existing wrapper test suites still pass against the migrated call sites (344 tests pass overall; the one pre-existing `tests/openai.test.ts` "Stream interrupted" crash reproduces on `main` and is unrelated to this PR).
- Lint clean.

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes!) — `sendEventToPosthog` and `sendEventWithErrorToPosthog` were not part of the public package surface (not exported from `index.ts`), so removing them is internal cleanup.

### If releasing new changes

- [x] Added the "release" label to the PR

---


packages/ai/src/captureAiGeneration.ts
```diff
@@ -0,0 +1,185 @@
...
+ * arbitrary clients (Cloudflare Workers AI, custom HTTP, etc.) and get the
+ * same events the SDK wrappers produce.
+ *
+ * When `error` is set, the event is captured as an error and the enriched
+ * error is returned so the caller can re-throw it.
+ */
+export const captureAiGeneration = async (
+  client: PostHog,
+  options: CaptureAiGenerationOptions
+): Promise<unknown | void> => {
```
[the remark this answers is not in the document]
richardsolomou: Good catch — fixed in d4b01713.

packages/ai/tests/captureAiGeneration.test.ts
```diff
@@ -0,0 +1,208 @@
...
+    const client = buildClient()
+
+    await captureAiGeneration(client, {
+      ...baseRequiredOptions,
+      error: new Error('no status field'),
+      httpStatus: 429,
+    })
+
+    expect(lastCaptureProperties(client).$ai_http_status).toBe(429)
+  })
```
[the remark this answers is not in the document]
richardsolomou: Done in d4b01713 — the error/httpStatus block is now a single `it.each` (and gained a missing 500-fallback case), and the model/provider resolution test was parameterised to also pin the no-override path.

packages/ai/src/captureAiGeneration.ts
```diff
@@ -0,0 +1,182 @@
...
+  /** Wall-clock latency in seconds. */
+  latency?: number
+  /** Time from request start to the first streamed token, in seconds. */
+  timeToFirstToken?: number
+
+  usage?: TokenUsage
+
+  /** Extra event properties merged into the captured event. */
+  properties?: Record<string, unknown>
+  groups?: Record<string, any>
```
carlos-marchal-ph: Can we use `unknown` here instead of `any`? `properties` two lines up already uses `unknown` and the rest of the option shape would line up.

packages/ai/src/captureAiGeneration.ts
```diff
@@ -0,0 +1,182 @@
...
+ * This is the canonical primitive that every `@posthog/ai` wrapper
+ * (`withTracing`, `OpenAI`, `Anthropic`, `GoogleGenAI`, …) funnels through, so
+ * external code can use it directly to instrument LLM calls made through
+ * arbitrary clients (Cloudflare Workers AI, custom HTTP, etc.) and get the
+ * same events the SDK wrappers produce.
+ *
+ * When `error` is set, the event is captured as an error and the enriched
+ * error is returned so the caller can re-throw it.
+ */
+export const captureAiGeneration = async (client: PostHog, options: CaptureAiGenerationOptions): Promise<unknown> => {
```
carlos-marchal-ph: The returned `enrichedError` is the same reference as `options.error`. The enrichment is the in-place mutation, not the return value, so callers who do `throw originalError` get the same flag set as callers who do `throw await captureAiGeneration(...)`.

Also worth noting that returning `unknown` from a function which could possible be undefined or an error is a bit of an anti-pattern. I'd rather have this just be `Promise<void>` and re-throw the error, or if we want to give users more flexibility, enrich the error in place or expose the error enrichment function as a public tool and make it idempotent by marking an enriched error with a symbol.

packages/ai/src/captureAiGeneration.ts
```diff
@@ -0,0 +1,182 @@
...
+  const safeOutput = sanitizeValues(options.output)
+
+  let httpStatus = options.httpStatus
+  let errorData: Record<string, unknown> = {}
+  let enrichedError: unknown
+  if (options.error) {
+    if (httpStatus === undefined) {
+      httpStatus =
+        options.error && typeof options.error === 'object' && 'status' in options.error
+          ? ((options.error as { status?: number }).status ?? 500)
```
carlos-marchal-ph: Could drop the `as { status?: number }` cast by adding a `typeof options.error.status === 'number'` check after the `'status' in` guard.

packages/ai/tests/captureAiGeneration.test.ts
```diff
@@ -0,0 +1,247 @@
...
+      $ai_output_tokens: 15,
+      $ai_latency: 1.5,
+      $ai_time_to_first_token: 0.4,
+      $ai_trace_id: 'trace-abc',
+      $ai_base_url: 'https://api.example.com',
+      $ai_stop_reason: 'stop',
+      $ai_tokens_source: 'sdk',
+      feature: 'transcript-toc',
+    })
+    expect(event.properties.$ai_tools).toHaveLength(1)
```
carlos-marchal-ph: Worth tightening this to `toEqual` so the array shape passes through unchanged, not just length.

packages/ai/src/vercel/middleware.ts
```diff
@@ -525,19 +524,24 @@ export const wrapVercelLanguageModel = <T extends LanguageModel>(
...
             usage,
             stopReason: finishReasonStr,
             tools: availableTools,
+            properties: mergedOptions.posthogProperties,
+            groups: mergedOptions.posthogGroups,
+            privacyMode: mergedOptions.posthogPrivacyMode,
+            modelOverride: mergedOptions.posthogModelOverride,
+            providerOverride: mergedOptions.posthogProviderOverride,
+            costOverride: mergedOptions.posthogCostOverride,
             captureImmediate: mergedOptions.posthogCaptureImmediate,
```
carlos-marchal-ph: These seven posthog* fields are duplicated across all four call sites in this file. Could hoist a baseOptions const at the top of the wrapper and spread it at each site.

carlos-marchal-ph (approved):
This is 🔥 love to see it

packages/ai/src/captureAiGeneration.ts
richardsolomou: Yep, you're right — restored to `Record<string, unknown>` in b9f8ae2e. The `EventMessage.groups` assignment doesn't actually need a cast (`unknown` is assignable to `any`), so the option type and the assignment now line up.

packages/ai/src/captureAiGeneration.ts
richardsolomou: Great point — switched to `Promise<void>` in b9f8ae2e. Wrappers now `throw error` directly (the in-place mutation of `__posthog_previously_captured_error` happens regardless of whether the caller throws the returned reference or the original). Added a test pinning the in-place mutation contract so it doesn't silently regress.

packages/ai/src/captureAiGeneration.ts
richardsolomou: Done in b9f8ae2e — replaced the cast with a `typeof options.error.status === 'number'` guard.

packages/ai/tests/captureAiGeneration.test.ts
richardsolomou: Tightened to `toEqual` in b9f8ae2e.

packages/ai/src/vercel/middleware.ts
richardsolomou: Done in b9f8ae2e — hoisted a `baseOptions` const next to `mergedOptions`, all four call sites now `...baseOptions` it.

PR #3500, opened 2026-04-29
feat(replay): mask the parameter in replays on request

fasyy612:
## Problem

Session replay captures full page URLs (rrweb meta events, fallback pageview, $url_changed) without honouring the SDK's existing mask_personal_data_properties / custom_personal_data_properties config, so query-string values like auth tokens show in recordings.

## Changes

`_maskUrl()` in `lazy-loaded-session-recorder.ts` now runs `maskQueryParams()` against that config before any user-supplied network mask function, plus one new test in `lazy-sessionrecording.test.ts`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -547,23 +548,28 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+                name: maskedUrl,
             } as any)
             // CapturedNetworkRequest uses 'name' for URL, but also check 'url' for compatibility
             return result?.name ?? (result as any)?.url
         }
 
         if (userSessionRecordingOptions.maskNetworkRequestFn) {
             const result = userSessionRecordingOptions.maskNetworkRequestFn({
-                url,
+                url: maskedUrl,
```
TueHaulund: Should we consider passing in the unmasked URL here?

The network request masking functions that customers are using now might break if we suddenly start feeding it masked URLs.

Then we should apply the URL masking to the returned URL before returning, I suppose...

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -547,23 +548,28 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+        const paramsToMask = this._instance.config.mask_personal_data_properties
+            ? [...PERSONAL_DATA_CAMPAIGN_PARAMS, ...(this._instance.config.custom_personal_data_properties || [])]
+            : []
+        const maskedUrl = maskQueryParams(url, paramsToMask, MASKED)
+
         // userSessionRecordingOptions.maskNetworkRequestFn is deprecated, fallback to it
         if (userSessionRecordingOptions.maskCapturedNetworkRequestFn) {
             const result = userSessionRecordingOptions.maskCapturedNetworkRequestFn({
-                name: url,
+                name: maskedUrl,
```
TueHaulund: Same thing here...

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
fasyy612: good point, I'll pass the unmasked url for `567` and `559`.

fasyy612:
Turns out its already implemented via `maskCapturedNetworkRequestFn`

PR #3508, opened 2026-04-30
fix(core): consume fetch response body to prevent CF Workers warnings

### LINKED ISSUE #3173: `_flush()` never consumes fetch response body, causing Cloudflare Workers cross-request promise warnings
## Bug description

`posthog-node` (via `@posthog/core`) triggers Cloudflare Workers cross-request promise resolution warnings because `_flush()` never consumes the fetch response body.

In `packages/core/src/posthog-core-stateless.ts`, `_flush()` calls `fetchWithRetry()` but discards the returned `Response` without reading the body (~line 1309):

```ts
await this.fetchWithRetry(url, fetchOptions, retryOptions)
// ↑ Response discarded — body never consumed
```

`fetchWithRetry()` (~line 1367) returns the raw response without consuming it either:

```ts
res = await this.fetch(url, { signal: ..., ...options })
if (!isNoCors && (res.status < 200 || res.status >= 400)) {
  throw new PostHogFetchHttpError(res, reqByteLength)
}
return res // body still unconsumed
```

In Cloudflare Workers, an unconsumed response body forces the runtime to clean up the underlying connection later, potentially in a different request's context, which triggers:

```
"Warning: A promise was resolved or rejected from a different request context than the one it was created in.
However, the creating request has already been completed or canceled. Continuations for that request are
unlikely to run safely and have been canceled."
```

This is the same class of bug the Sentry SDK had and fixed in [getsentry/sentry-javascript#18534](https://github.com/getsentry/sentry-javascript/issues/18534) / [PR #18545](https://github.com/getsentry/sentry-javascript/pull/18545).

**Suggested fix**: consume or cancel the response body in `_flush()`:

```ts
// Option A: consume the body
const res = await this.fetchWithRetry(url, fetchOptions, retryOptions)
await res.text().catch(() => {})

// Option B: cancel the body stream
const res = await this.fetchWithRetry(url, fetchOptions, retryOptions)
await res.body?.cancel().catch(() => {})
```

**Workaround** — pass a custom `fetch` that eagerly consumes response bodies:

```ts
const cfSafeFetch: typeof fetch = async (input, init) => {
  const response = await fetch(input, init)
  const body = await response.text().catch(() => '')
  return new Response(body, {
    status: response.status,
    statusText: response.statusText,
    headers: response.headers,
  })
}

new PostHog(apiKey, {
  host,
  flushAt: 1,
  flushInterval: 0,
  fetch: cfSafeFetch,
})
```

## How to reproduce

1. Deploy a Cloudflare Worker that uses `posthog-node` with `flushAt: 1, flushInterval: 0` (recommended config per [PostHog CF Workers docs](https://posthog.com/docs/libraries/cloudflare-workers))
2. Call `capture()` inside a request handler (e.g. via `executionContext.waitUntil(...)` for non-blocking analytics)
3. Send concurrent requests to the worker — the runtime warns about cross-request promise resolution from the unconsumed response bodies left by `_flush()`

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

- `posthog-node@5.24.11` / `@posthog/core@1.20.1`
- Cloudflare Workers with `compatibility_date: "2026-01-12"` and `nodejs_compat` compatibility flag
- The warning is non-fatal but the runtime **cancels continuations** for the leaked promises, so post-flush error handling may silently not run
- High-traffic workers with analytics middleware on every request generate a high volume of these warnings
- The root cause is in `packages/core/src/posthog-core-stateless.ts`, so it may affect any runtime that enforces response body consumption (not just CF Workers)

KuaaMU:
## Summary

Consume the response body returned by `fetchWithRetry()` in both `_flush()` and `sendImmediate()` by calling `response.body?.cancel()`. This prevents cross-request promise resolution warnings in runtimes like Cloudflare Workers that enforce response body consumption.

Fixes #3173

## Problem

In `packages/core/src/posthog-core-stateless.ts`, both `_flush()` (line ~1153) and `sendImmediate()` (line ~978) call `fetchWithRetry()` but discard the returned `Response` without reading the body:

```ts
await this.fetchWithRetry(url, fetchOptions, retryOptions)
// ↑ Response discarded — body never consumed
```

In Cloudflare Workers, an unconsumed response body forces the runtime to clean up the underlying connection later, potentially in a different request's context, which triggers:

> Warning: A promise was resolved or rejected from a different request context than the one it was created in. However, the creating request has already been completed or canceled. Continuations for that request are unlikely to run safely and have been canceled.

This is the same class of bug the Sentry SDK had and fixed in [getsentry/sentry-javascript#18534](https://github.com/getsentry/sentry-javascript/issues/18534) / [PR #18545](https://github.com/getsentry/sentry-javascript/pull/18545).

## Fix

Consume or cancel the response body after `fetchWithRetry()` returns:

```ts
const response = await this.fetchWithRetry(url, fetchOptions, retryOptions)
await response.body?.cancel().catch(() => {})
```

Uses `response.body?.cancel()` (Option B from the issue) which is the most lightweight approach — it tells the runtime to discard the body without reading it into memory.

### Locations fixed

1. `_flush()` — the main batch flush path called on interval/queue full
2. `sendImmediate()` — the synchronous send path used when `flushAt: 1`

### Why not the other call sites?

The other `fetchWithRetry()` callers (remote config, feature flags, surveys) all consume the response via `.json()`, so they don't have this issue.

## Testing

- The change is minimal (10 lines added, 2 removed) and follows the exact pattern suggested in the issue
- No behavioral change — the response body was already being discarded; we're just doing it explicitly and cleanly


KuaaMU:
Thanks for the review @greptile-apps! Fixed all 3 issues:

1. Added `body?: ReadableStream<Uint8Array> | null` to `PostHogFetchResponse` type
2. Fixed broken optional-chain on line 982: `response.body?.cancel().catch()` → `response.body?.cancel()?.catch()`
3. Same fix on line 1161

The  was called on `undefined` when `body` is `null` (e.g. 204 No Content), causing a `TypeError` inside the silent error path.

turnipdabeets (approved):

turnipdabeets (approved):
LGTM, might be good to add a unit test asserting cancel() is invoked

KuaaMU:
Thanks for the review @greptile-apps @chatgpt-codex-connector! You're right about the optional chain issue.

When `response.body` is `null`, `response.body?.cancel()` evaluates to `undefined`, and calling `.catch()` on `undefined` throws a `TypeError`.

The fix: move the `?.` to also cover `.catch()`:
```ts
// Before (broken):
await response.body?.cancel()?.catch(() => {})

// After (correct):
await response.body?.cancel()?.catch(() => {})
```

Wait — looking more carefully, the current code is `response.body?.cancel()?.catch(() => {})` which already has `?.` before `.catch()`. Let me verify this is actually correct with the current diff...

Actually, I see the diff shows `await response.body?.cancel()?.catch(() => {})` — the `?.` before `.catch()` should short-circuit when `cancel()` returns `undefined`. But I'll double-check and push a fix if needed.

marandaneto:
closing by https://github.com/PostHog/posthog-js/pull/3516

marandaneto:
closing because commits arent signed, i couldnt force push, decided to open a new PR instead

PR #3510, opened 2026-04-30
chore: move posthog-rrweb into the monorepo

arnohillen:
## Summary

Imports the [`posthog-rrweb`](https://github.com/PostHog/posthog-rrweb) fork (currently published as `v0.0.60`) into `packages/rrweb/` so the rrweb packages release from this repo via the existing changesets pipeline. Collapses the current two-step publish flow (`posthog-rrweb` → npm → `posthog-js`) into a single monorepo release, picking up where #2815 left off.

### What's moved

- Snapshot taken from `posthog-rrweb@main` (latest commit on top of the `0.0.60` release).
- `rrvideo`, `rrweb-player`, and `web-extension` are intentionally **dropped** — `posthog-js` doesn't consume them.
- All other rrweb subpackages live under `packages/rrweb/` (top-level subpackages) and `packages/rrweb/plugins/` (plugins).

### What gets published from here

The same set of packages we already publish today, identified by inspecting `npm view` of each upstream package:

- `@posthog/rrweb`
- `@posthog/rrweb-types`
- `@posthog/rrweb-utils`
- `@posthog/rrdom`
- `@posthog/rrweb-snapshot`
- `@posthog/rrweb-record`
- `@posthog/rrweb-plugin-console-record`

Internal-only packages (`rrdom-nodejs`, `@posthog/rrweb-replay`, `@posthog/rrweb-all`, `@posthog/rrweb-packer`, the experimental canvas-webrtc / sequential-id / console-replay plugins) are listed under `.changeset/config.json` `ignore` so they don't get versioned or pushed to npm.

### Consumer wiring

`packages/browser` (a.k.a. `posthog-js`) now resolves the three rrweb packages it bundles via `workspace:*`:

- `@posthog/rrweb-record`
- `@posthog/rrweb-plugin-console-record`
- `@posthog/rrweb-types`

They remain in `devDependencies` since they're bundled into the recorder, not exposed at runtime.

### Patches removed

The two pnpm patches that lived under `patches/` are no longer needed and have been deleted, along with the `pnpm.patchedDependencies` block in the root `package.json`:

- `patches/@rrweb__record@2.0.0-alpha.17.patch` — added PostHog-specific source modifications: Angular Zone unpatched-prototype lookup (`angularZoneUnpatchedAlternative`), defensive `<link>` href handling (`hrefFrom` / `findStylesheet`), and a small `getUntaintedPrototype` fallback. These now live as real TypeScript in `packages/rrweb/utils/src/index.ts` and `packages/rrweb/rrweb-snapshot/src/snapshot.ts`.
- `patches/@rrweb__rrweb-plugin-console-record@2.0.0-alpha.17.patch` — pure dead-code-elimination hack that stripped ~7000 lines of `postcss` + `picocolors` from the upstream prebuilt `dist/`. With the in-tree fork building from source via `vite.config.default.ts`, those deps are simply not bundled, so there's nothing to strip.

`pnpm patch` also doesn't apply to `workspace:*` deps anyway, so retaining the patches would have been a no-op.

### Tooling / config changes

- **`pnpm-workspace.yaml`**: adds `packages/rrweb/*` and `packages/rrweb/plugins/*` globs and a small catalog for `vite`, `vite-plugin-dts`, `esbuild-plugin-umd-wrapper`.
- **`turbo.json`**: picks up `vite.config.{ts,js}` as build inputs and adds `prepublish` / `check-types` tasks used by the rrweb subpackages.
- **Root `package.json`**: hoists `esbuild`, `esbuild-plugin-umd-wrapper`, and `rollup-plugin-visualizer` into devDependencies so vite builds resolve correctly across all rrweb subpackages.
- **`.npmrc`**: a few `trust-policy-exclude[]` entries for transitive deps brought in by the rrweb tree that lost provenance attestations (`@typescript-eslint/parser@5.62.0`, `rxjs@7.8.2`, `semver@5.7.2`, `vite@5.4.21`, etc.). Without these, `pnpm install` fails under our `trust-policy: no-downgrade` setting.
- **`packages/rrweb/.eslintrc.js`**: scoped overrides that disable upstream-only rules and `eslint-plugin-posthog-js` rules that don't apply to the rrweb tree.
- **`.gitignore`**: ignore vite timestamp files and `*-bundle-analysis.html` (see CI fixes below).

### CI / release changes

- **`.github/workflows/release.yml`**: extends the publish matrix to the seven publishable `@posthog/rrweb-*` packages.
- **`.github/workflows/library-ci.yml`**: adds an `rrweb` job (vitest + puppeteer under `xvfb-run`) that only runs when `packages/rrweb/**` changes. Failed snapshot diffs upload as artifacts.

### Post-push CI fixes

The local test plan passed cleanly, but GitHub Actions surfaced three issues on first push that needed follow-up commits:

1. **Prettier v3 async `format()` in rrweb test utils** (`packages/rrweb/rrweb/test/utils.ts`)
   - The rrweb package's `test/utils.ts` imports `format` from `prettier` and uses it synchronously inside `stringifyDomSnapshot`. The standalone `posthog-rrweb` repo pins prettier 2.x; in this monorepo the rrweb package resolves the workspace root's prettier 3.7.4, where `format()` returns a `Promise<string>` instead of a `string`.
   - Symptom: 5 failed `replayer.test.ts` snapshots showing `[object Promise]` instead of formatted HTML.
   - Fix: made `stringifyDomSnapshot` async, used `Promise.all` over the asset map, and awaited the call in `assertDomSnapshot`.

2. **Stop tracking `*-bundle-analysis.html`** (root `.gitignore` + delete 12 generated files under `packages/rrweb/`)
   - `rollup-plugin-visualizer` writes `<name>-bundle-analysis.html` per build, embedding absolute build-time paths (`/Users/<dev>/...` locally vs `/home/runner/work/...` in CI).
   - Symptom: the "Write mangled property names" job (which runs a build then `git diff --exit-code`) failed because each build mutates these files.
   - Fix: added `*-bundle-analysis.html` to `.gitignore` and removed all 12 tracked copies under `packages/rrweb/`.

3. **`compliance/{browser,node}/Dockerfile`**
   - **Stale `COPY patches`**: Both had `COPY patches ./patches`, but this PR deletes `patches/` entirely. On a fresh CI checkout the directory doesn't exist, so the COPY failed: `failed to compute cache key: ... "/patches": not found`. Fix: removed `COPY patches ./patches` from both.
   - **Missing rrweb workspace deps in image**: `compliance/browser/Dockerfile` didn't copy `packages/rrweb`, but `posthog-js` now depends on `@posthog/rrweb-record`, `@posthog/rrweb-plugin-console-record`, and `@posthog/rrweb-types` as `workspace:*`. Fix: added `COPY packages/rrweb ./packages/rrweb` (and `turbo.json` to the existing root-config copy).
   - **Workspace deps weren't being built**: the Dockerfile previously ran `pnpm --filter posthog-js build` directly, which only invokes posthog-js's own build — the rrweb workspace deps' `dist/` (where their `main`/`module`/`types` resolve) was never populated, so TypeScript blew up with `error TS2307: Cannot find module '@posthog/rrweb-record'`.
   - **Nested-turbo race when fixing the above naïvely**: switching to `pnpm turbo run build --filter=posthog-js` exposed a separate issue — each rrweb package's `build` script is itself `pnpm turbo run prepublish`, so the outer turbo running `@posthog/rrweb-record:build` and `@posthog/rrweb-plugin-console-record:build` in parallel spawned two nested turbos that raced on the shared `@posthog/rrweb-utils:prepublish` task and one of them crashed in `vite-plugin-dts` mid-write.
   - **`@posthog/core` regression hidden by docker cache**: switching the previous step to `pnpm turbo run prepublish` (instead of `build`) skipped `@posthog/core` and `@posthog/types` (those ship a `build` script, not `prepublish`), so the next CI run failed with `Cannot find module '@posthog/core'`. My local re-test missed it because docker cached the earlier `pnpm --filter @posthog/core build` layer; `--no-cache` reproduces the failure.
   - **Final fix**: explicit ordering that respects each package's actual script — `pnpm --filter @posthog/types build` → `pnpm --filter @posthog/core build` → `pnpm turbo run prepublish` (handles all rrweb packages with proper dep order and no nested turbo) → `pnpm --filter posthog-js build`. Verified with `docker build --no-cache -f compliance/browser/Dockerfile .`.

### Behavior changes

None intended. Verified locally:

- `packages/browser/dist/recorder.js` (and the other recorder bundles) are **byte-identical** to the same artifacts built from `main`.
- `posthog-js` and rrweb test suites pass.
- `es-check` still passes on the bundled output.
- The mangled-properties allowlist is unchanged.

### Versioning

The accompanying changeset (`.changeset/move-posthog-rrweb-into-monorepo.md`) bumps each of the seven publishable packages by a `patch` so the first monorepo release lands cleanly after `v0.0.60`.

## Test plan

CI is the source of truth, but locally I ran:

- [x] `pnpm install` clean (after wiping `node_modules`)
- [x] `pnpm turbo run build` across the whole graph (skipping the pre-existing `@posthog/convex` clean issue on `main`)
- [x] `pnpm turbo run check-types` across the rrweb tree + `__tests__` in `packages/browser`
- [x] `pnpm lint` on the whole monorepo
- [x] `pnpm turbo --filter=posthog-js test:unit` and `test:functional`
- [x] `pnpm turbo --filter=@posthog/rrweb-snapshot --filter=@posthog/rrdom --filter=@posthog/rrweb-replay --filter=@posthog/rrweb-all --filter=@posthog/rrweb-packer test`
- [x] `@posthog/rrweb` puppeteer suite via `xvfb-run` with Chrome cached at `~/.cache/puppeteer` (re-verified all 37 `replayer.test.ts` cases pass after the prettier-async fix)
- [x] Bundle equivalence vs `main` worktree: `packages/browser/dist/recorder*.js` byte-for-byte identical
- [x] `es-check es5` / `es-check es6` on the new bundles
- [x] Mangled-properties output unchanged (re-verified after the `*-bundle-analysis.html` gitignore fix)
- [x] `pnpm changeset status` / `pnpm changeset version` dry-run: only the seven publishable packages bump
- [x] `release.yml` matrix includes all and only those seven packages
- [x] Tarball install in `playground/nextjs` (the `[ReferenceError: document is not defined]` SSR error in `replay-examples/animations` is pre-existing on `main`, not a regression)
- [x] `docker build -f compliance/browser/Dockerfile .` builds end-to-end after the Dockerfile fix

### Things to watch on this PR

- The `rrweb` CI job only runs on changes under `packages/rrweb/**`, so this PR is the first time it'll run end-to-end on Actions runners.
- Codeql / dependency review: this introduces a large vendor tree. If anything reappears from the #2815 review, I'll address in follow-ups.
- After merge, the next changesets release will publish v0.0.61 of the seven `@posthog/rrweb-*` packages from this repo for the first time. Worth a sanity check on npm dist-tags after the release job runs.

Refs: PostHog/posthog-rrweb, supersedes #2815.


packages/rrweb/rrweb-snapshot/package.json
```diff
@@ -0,0 +1,85 @@
...
+        "test:watch": "vitest watch",
+        "retest:update": "vitest run --update",
+        "test:update": "pnpm build && vitest run --update",
+        "bench": "vite build && vitest bench",
+        "dev": "vite build --watch",
+        "build": "pnpm turbo prepublish",
+        "check-types": "tsc --noEmit",
+        "prepublish": "pnpm check-types && vite build",
+        "lint": "pnpm eslint src"
+    },
```
TueHaulund: Quick consistency nit: the scripts and indentation here drift from the sibling vendored packages (`packages/rrweb/replay/package.json`, `packages/rrweb/all/package.json`):

| | this file | replay / all |
|---|---|---|
| `build` | `pnpm turbo prepublish` | `pnpm turbo run prepublish` |
| `test` | `pnpm build && vitest run` | `vitest run` |
| indentation | 4 spaces | 2 spaces |

Both forms of the build command work, but `pnpm turbo run prepublish` is what the rest of the monorepo uses. Forcing a build before every test also makes `pnpm test` here noticeably slower than in the siblings.

Probably worth aligning all three — looks like a porting artifact from the upstream rrweb repo, easy to lose track of later. Not a blocker.

TueHaulund (approved):

PR #3511, opened 2026-05-01
fix(replay): detect session rotation while _isIdle is unknown

fasyy612:
## Problem

When a session rotates before any rrweb-interactive event fires, the recorder misses the rotation. The new session's initial Meta + FullSnapshot end up written to the prior session's blob, leaving the new session with orphan incrementals and no renderable base frame, the player sits on "Buffering..." until the next periodic snapshot (~5 minutes later).

`_updateWindowAndSessionIds` bailed early on truthy `_isIdle`, which catches both `true` (confirmed idle) and the initial `'unknown'` state. Sessions whose `_isIdle` never flips from `'unknown'` to `false` (no rrweb-interactive event yet) silently miss session-change detection.

## Changes

Change the predicate from `if (this._isIdle)` to `if (this._isIdle === true)`. This mirrors the existing `=== true` predicate in `onRRwebEmit`, whose comment already calls out: _"we don't want to return early if idle is 'unknown'"_. The confirmed-idle path is unchanged, the `'unknown'` state now correctly proceeds to the session-change check.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/fine-pears-care.md
+++ b/.changeset/fine-pears-care.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+detect session rotation while \_isIdle is unknown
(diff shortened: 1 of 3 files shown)


ksvat (approved):
make sure to run a changeset  
  
also truthiness strikes again RIPPPPP

PR #3517, opened 2026-05-04
fix(llma): correct Gemini image inlineData mistyped as audio

carlos-marchal-ph:
## Problem

A customer reported that an LLM trace for a Gemini image-preview model was rendering the **output** as an audio player instead of an image. The captured payload at `properties.$ai_output_choices[0].content[0]` looked like:

```json
{ "data": "/9j/4AAQ…", "mime_type": "image/jpeg", "type": "audio" }
```

Note `type: "audio"` even though `mime_type` is `image/jpeg`. Root cause: `formatResponseGemini` in `packages/ai/src/utils.ts` hard-coded `type: 'audio'` for any Gemini `inlineData` part, regardless of mime type, and defaulted the mime to `'audio/pcm'`. The PostHog frontend then trusted that label and routed the part through the audio renderer.

⚠️ **note:** this change affects an internal only path since multimodal support is not rolled out to users. the user facing fix was to suggest to them to disable the private env var that enables this. no

## Changes

- `packages/ai/src/utils.ts` — `formatResponseGemini` now dispatches on mime type:
  - `image/*` → `{ type: 'image', inline_data: { mime_type, data } }`
  - `audio/*` → existing `{ type: 'audio', mime_type, data }` shape (only when actually audio)
  - everything else (`application/pdf`, `video/*`, missing mime, …) → `{ type: 'document', inline_data: { mime_type, data } }`
  - Default mime fallback changed from `'audio/pcm'` to `'application/octet-stream'` so unknown types route to `document`, never silently become audio.
- `packages/ai/src/gemini/index.ts` — applied the same audio/image/document dispatch in `WrappedModels.formatPartsAsContentBlocks` for parity, and dropped a now-unnecessary `as FormattedContentItem` cast.
- `packages/ai/src/types.ts` — aligned `FormattedImageContent` and `FormattedDocumentContent` with the snake-case `inline_data: { mime_type, data }` shape we already emit; kept Anthropic-style `source` as optional for forward compat.
- `packages/ai/tests/tools.test.ts` — added regression coverage for `image/jpeg`, `application/pdf`, and missing-mime cases. The existing `audio/pcm` test still passes unchanged.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

Not releasing as a standalone version, can be bundled in the next release

---


packages/ai/src/utils.ts
```diff
@@ -275,6 +281,19 @@ export const formatResponseOpenAI = (response: any): FormattedMessage[] => {
...
+  mimeType: string,
+  data: string
+): FormattedAudioContent | FormattedImageContent | FormattedDocumentContent => {
+  if (mimeType.startsWith('audio/')) {
+    return { type: 'audio', mime_type: mimeType, data }
+  }
+  if (mimeType.startsWith('image/')) {
+    return { type: 'image', inline_data: { mime_type: mimeType, data } }
+  }
+  return { type: 'document', inline_data: { mime_type: mimeType, data } }
```
richardsolomou, line 294: Potentially dumb question on the fall-through: a `video/mp4` part will end up as `{ type: 'document', inline_data: { mime_type: 'video/mp4', data } }`. The PR description calls this out as intentional, but I'd want to double-check the frontend's "document" renderer actually inspects `mime_type` and does something reasonable for video, otherwise we've just swapped one mis-rendering (audio renderer for images) for another (document/PDF renderer for video). If the renderer doesn't disambiguate, a separate `'video'` branch here would be cheap to add and keeps the dispatch honest.

packages/ai/src/gemini/index.ts
```diff
@@ -313,16 +314,8 @@ export class WrappedModels {
...
-
-        blocks.push({
-          type: contentType,
-          inline_data: {
-            data: inlineData.data,
-            mime_type: mimeType,
-          },
-        } as FormattedContentItem)
+        const mimeType = inlineData.mimeType || inlineData.mime_type || 'application/octet-stream'
+        blocks.push(buildInlineDataBlock(mimeType, inlineData.data))
```
richardsolomou, line 318: Heads up that this also changes the input-formatting path, which the PR description doesn't mention. Audio `inlineData` here used to come out as `{ type: 'document', inline_data: { mime_type, data } }` and will now be `{ type: 'audio', mime_type, data }` - different label, different shape. Probably the right call, but worth a line in the description so anyone rendering `$ai_input` knows to expect it. Also potentially worth a check to see if anything else depended on `inlineData` that might need changing.

richardsolomou (approved):
LGTM, just a couple things, but needs a changeset too!

packages/ai/src/gemini/index.ts
carlos-marchal-ph: That is actually the correct call and in line with other code paths

packages/ai/src/utils.ts
carlos-marchal-ph: TBH since multimodal support hasn't landed in prod, we can cross that bridge when we get there. The first version was built for audio and images only, and we are still pending on having a sane way to process large blobs to actually be able to roll it out 😞

PR #3535, opened 2026-05-05
chore: remove unused deprecate-rrweb-packages script

arnohillen:
## Summary

- Removes `scripts/deprecate-rrweb-packages.sh`, a one-off helper added in #3532 to mark the `@posthog/rrweb*` npm packages as deprecated. It's no longer needed and can live in git history if we ever want to reference it again.

## Test plan

- [ ] N/A — repo-only deletion of an unused script


TueHaulund (approved):

marandaneto:
@PostHog/team-replay should we merge this?

PR #3539, opened 2026-05-07
feat(llma): unify provider sanitisation under a generic binary content redactor

### LINKED ISSUE #2134: maximum event size exceeded using @posthog/ai when passing inline images
## Bug description

I am using the posthog client to proxy my gemini/openAI requests. When sending inline images to the API, the posthog request fails. Fortunatelly it does not fail the whole request, only the posthog request (so i still get the answer back from the API). 
 Example error:
```typescript
Error while flushing PostHog: message=HTTP error while fetching PostHog: status=413, reqByteLength=1252075, response body=maximum event size exceeded: Event rejected by kafka during send PostHogFetchHttpError: HTTP error while fetching PostHog: status=413, reqByteLength=1252075
    at retriable (XXXnode_modules/posthog-core/src/index.ts:1123:17)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async retriable (XXXnode_modules/posthog-core/src/utils.ts:38:19)
    at async PostHog.fetchWithRetry (XXXnode_modules/posthog-core/src/index.ts:1106:12)
    at async PostHog._flush (XXXnode_modules/posthog-core/src/index.ts:1046:9) {
  response: Response {
    status: 413,
    statusText: 'Payload Too Large',
    headers: Headers {
      date: 'Wed, 02 Jul 2025 17:41:58 GMT',
      'content-type': 'text/plain; charset=utf-8',
      'transfer-encoding': 'chunked',
      connection: 'keep-alive',
      vary: 'origin, access-control-request-method, access-control-request-headers, Accept-Encoding',
      'access-control-allow-credentials': 'true',
      'x-envoy-upstream-service-time': '21',
      'content-encoding': 'gzip',
      server: 'envoy',
      'strict-transport-security': 'max-age=31536000; includeSubDomains'
    },
    body: ReadableStream { locked: true, state: 'closed', supportsBYOB: true },
    bodyUsed: true,
    ok: false,
    redirected: false,
    type: 'basic',
    url: 'https://eu.i.posthog.com/batch/'
  },
  reqByteLength: 1252075
}
```
Example request:
```typescript
const posthogClient = new PostHog(POSTHOG_PUBLIC_KEY, { host: "https://eu.i.posthog.com" });
const geminiClient = new OpenAI({
  apiKey: GEMINI_API_KEY,
  baseURL: "https://generativelanguage.googleapis.com/v1beta/openai",
  posthog: posthogClient,
});
let base64Image = "xxxx"; // use valid image here. In theory the gemini API support up to 20MB of inline images
const messages: any = [
  {
    role: "user",
    content: [
      {
        type: "text",
        text: "Describe the image",
      },
      { type: "image_url", image_url: { url: `data:image/jpeg;base64,${base64Image}` } },
    ],
  },
];
const response = await this.geminiClient.chat.completions.create({
  model: "models/gemini-2.5-flash-lite-preview-06-17",
  messages: messages,
  max_tokens: 100,
});
```
More details about passing images can be found here: https://ai.google.dev/gemini-api/docs/image-understanding#inline-image

Note: I guess this will also happen for audio and video passing. Maybe even for passing documents.
## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [ ] posthog-node
- [x] posthog-ai
- [ ] posthog-react-native
- [ ] posthog-nextjs-config

carlos-marchal-ph:
## Problem

`sanitization.ts` carried six near-duplicate provider walkers — one per shape we encounter (`image_url`, `inlineData`, `source.data`, etc.). Every new provider needed another walker and every new content shape needed each walker patched, so coverage gaps were inevitable. The placeholder also lied: every redaction came out as `[base64 image redacted]`, even for audio, video, or PDF data.

Several users were already encountering cases where redaction was not happening, causing events to fail due to size limits.

## Changes

New `sanitization/` module that consolidates the redaction logic:

- `Base64Recognizer` — single recognizer for data URLs (with media-type capture) and raw base64. Uses a confidence-prefix alphabet check on `minLength` chars, with a `startsWith('data:')` fast-reject up front.
- `MediaTypeContext` — derives the most specific MIME from sibling/parent metadata or the field name. Inference ladder is split into public methods (`inferFromSiblingMime`, `inferFromSiblingFormat`, `inferFromParentType`, `inferFromKey`).
- `BinaryContentRedactor` — generic walker. Redacts data URLs, raw base64 (length floor 64 with strong context, 1024 without), and `Uint8Array`/`Buffer` values. Cycle-safe via a `WeakSet` it owns. Skips redaction when `_INTERNAL_LLMA_MULTIMODAL` is set.

Provider sanitizers (`sanitizeOpenAI`, `sanitizeAnthropic`, `sanitizeGemini`, `sanitizeLangChain`, `sanitizeOpenAIResponse`, `redactBase64DataUrl`) collapse into thin wrappers over a shared redactor instance — call sites in `openai/`, `anthropic/`, `gemini/`, `langchain/`, `vercel/`, `utils.ts` are unchanged.

The entire original test suite of these sanitisers has been kept almost intact and still passing. The only minor modifications are to accommodate the new placeholders with more information, and to adapt to an issue that's been fixed which meant short alphanumeric strings, like single words, could be accidentally redacted.

Placeholders now carry the actual content type:
- `[base64 image/jpeg redacted]`, `[base64 image/png redacted]`, `[base64 video/mp4 redacted]` for data URLs
- `[base64 application/pdf redacted]`, `[base64 audio/L16;codec=pcm;rate=24000 redacted]` from sibling MIME hints
- `[base64 audio redacted]` when only the family is known
- `[base64 redacted]` for bare-string entry points with no parent context

Commits are split for review:
1. `feat: implement generic binary content redactor` — adds the new classes + 37 unit tests in isolation, no callers changed.
2. `feat: switch sanitisiation to generic redactor` — rewires `sanitization.ts` to delegate, updates legacy test placeholder assertions to the precise outputs.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Built collaboratively with Claude Code over multiple iterations. Notable design calls during the session:

- The recognizer keeps a length floor (64 strong / 1024 weak); whitespace stripping was rejected because most punctuation-light sentences would otherwise look like base64.
- `BinaryContentRedactor` owns its `WeakSet` and resets it per `redact()` call rather than threading `visited` through `walk()`.
- `MediaTypeContext.forChild` was removed; callers use the constructor directly.
- The placeholder format intentionally diverges from the legacy `[base64 image redacted]` so audio/video/PDF data is no longer mislabelled "image".


packages/ai/src/sanitization/binary_content_redactor.ts
```diff
@@ -0,0 +1,67 @@
...
+    if (typeof value === 'string') return this.redactString(value, ctx)
+    if (typeof value !== 'object') return value
+
+    // Buffer extends Uint8Array, so this branch catches both.
+    if (typeof Uint8Array !== 'undefined' && value instanceof Uint8Array) {
+      return this.placeholderFor(ctx.inferMediaType())
+    }
+
+    if (this.visited.has(value)) return null
+    this.visited.add(value)
```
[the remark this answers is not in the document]
carlos-marchal-ph: We are not expecting shared refs at all, this is a safety valve so we don't go into an infinite loop.

packages/ai/src/sanitization/binary_content_redactor.ts
```diff
@@ -0,0 +1,67 @@
+import { Base64Recognizer } from './base64_recognizer'
+import { MediaTypeContext } from './media_type_context'
+
+const STRONG_CONTEXT_MIN_LENGTH = 64
```
Radu-Raicea, line 4: Why the minimum length? Could we have smaller but still base64?

packages/ai/src/sanitization.ts
```diff
@@ -1,270 +1,15 @@
...
-    return data.map(processGeminiItem)
-  }
-
-  return processGeminiItem(data)
-}
-
-export const sanitizeLangChain = (data: unknown): unknown => {
-  return processMessages(data, sanitizeLangChainImage)
-}
+export const sanitizeOpenAI = (data: unknown): unknown => redactor.redact(data)
```
Radu-Raicea, line 11: Is this being done for streaming OpenAI too?

packages/ai/src/sanitization/binary_content_redactor.ts
carlos-marchal-ph: It's very unlikely to have such small base64 files and if we did it wouldn't be so bad to ingest them. On the flip side without this, we could have some fields that look like or are base64 that we want to keep, like IDs or encoded JSON payloads, or even alphanumeric strings without spaces, that we might end up redacting by accident.

packages/ai/src/sanitization.ts
carlos-marchal-ph: Yes, I didn't change any callsites for these functions, just their internal implementation.

Radu-Raicea (approved):

packages/ai/src/sanitization.ts
carlos-marchal-ph: Here it is [in chat completions](https://github.com/PostHog/posthog-js/blob/main/packages/ai/src/openai/index.ts#L265-L266) and here it is [in the responses API ](https://github.com/PostHog/posthog-js/blob/main/packages/ai/src/openai/index.ts#L477-L480), both for streaming and in master.

We don't touch these files on this PR, does this still look off?

PR #3542, opened 2026-05-07
fix(replay): preserve <style> textContent on var-shorthand CSS

TueHaulund:
## Problem

A user reported that recordings of a Chakra UI v3 + Panda CSS app rendered
the login form with no internal padding — labels, inputs, and buttons flush
against the card edge — even though their actual UI was correctly padded.
Most CSS (colors, fonts, border-radius, button styling) was intact in
replay; only certain layout shorthands dropped out.

Inspecting their export, exactly one rule in the FullSnapshot looked
broken:

```css
.css-1xow64h {
  padding-top: ;
  padding-right: ;
  padding-left: ;
  padding-bottom: var(--chakra-spacing-12);
  ...
}
```

This is the form card class (`p={8}` + `pb={12}` overrides). The original
Chakra/Panda CSS combines a `var()`-bearing shorthand with a longhand
override:

```css
.css-1xow64h { padding: var(--chakra-spacing-8); padding-bottom: var(--chakra-spacing-12); ... }
```

Per the [CSS Custom Properties spec](https://drafts.csswg.org/css-variables/#variables-in-shorthands),
when a shorthand contains `var()`, its longhand sub-properties are stored
with empty token lists, so `CSSStyleRule.cssText` emits
`padding-top: ; padding-right: ; padding-left: ;`. Verified in real
Chrome — the output matches the export byte-for-byte. The shorthand
itself isn't recoverable via `getPropertyValue('padding')` either after a
longhand override (Chrome returns `""`), so there's no smart-serialization
workaround inside the rule walker. The original `<style>` textContent is
the only place the canonical CSS exists.

The current `serializeTextNode` discards textContent and replaces it with
`stringifyStylesheet(parent.sheet)` whenever the `<style>` has a single
text-node child, so the corruption silently strips layout from the
replayed page.

This affects any CSS-in-JS framework that combines `var()` with shorthands
followed by a longhand override — Chakra UI v3, Panda CSS, Emotion, etc.
The same class of bug is tracked upstream as
[rrweb-io/rrweb#1667](https://github.com/rrweb-io/rrweb/issues/1667)
(reported for the `background` shorthand; same root cause and CSSOM
limitation).

## Changes

- Add `hasEmptyShorthandLonghand(css)` to `utils.ts`: detects the empty
  `prop: ;` artifact pattern, ignoring custom properties (`--foo: ;` is
  legitimately allowed).
- In `serializeTextNode`, when `stringifyStylesheet(parent.sheet)` would
  emit that pattern, fall back to the original `<style>` textContent
  rather than the corrupted serialization.
- Tests:
  - Parameterized unit tests for the detector (corrupted vs. clean
    inputs, custom properties excluded).
  - Integration test using the existing `<style>` test pattern, with
    `cssText` mocked via `Object.defineProperty` to simulate the browser
    behavior (jsdom does not reproduce var-in-shorthand CSSOM
    serialization; verified in real Chrome).

## Verification

End-to-end replayed the original recording and a copy with the corruption
manually reversed (i.e. what the export would have looked like with this
fix in place at capture time). The corrupted version reproduces the
flush-to-edge symptom; the patched version renders the card with correct
~32px internal padding.

## Scope

This fix only covers the inline `<style>` path. `<link rel=stylesheet>`
inlining (also in `record.ts`) calls the same `stringifyStylesheet` and
has the same corruption, but the original CSS lives at the URL rather
than in the DOM, so the fallback strategy doesn't apply directly. Left as
a follow-up.

Tradeoff: in the rare case where a `<style>` is _both_ subject to
var-shorthand corruption _and_ has rules added later via
`document.styleSheets[i].insertRule(...)` on the same element, the
runtime-added rules will be lost (we keep the SSR text and skip the sheet
serialization). This is the minimum-impact fix; the alternative
(parse + merge textContent and sheet rules) would require shipping a CSS
parser for marginal benefit.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

(Change is in `@posthog/rrweb-snapshot`, which `posthog-js` consumes
transitively via `@posthog/rrweb-record`.)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/rrweb-snapshot-css-var-shorthand.md
+++ b/.changeset/rrweb-snapshot-css-var-shorthand.md
@@ -12,4 +12,4 @@ padding-right: ; padding-left: ;`. The previous behavior replaced the
 on replay. We now detect the empty-longhand pattern and keep the original
 textContent in that case. Affects users of any CSS-in-JS framework that
 combines `var()` with shorthands (Chakra UI v3, Panda CSS, Emotion, etc.).
-See rrweb-io/rrweb#1626 / #1567.
+Same class of bug as rrweb-io/rrweb#1667.
--- a/.changeset/rrweb-snapshot-css-var-shorthand.md
+++ b/.changeset/rrweb-snapshot-css-var-shorthand.md
@@ -1,5 +1,6 @@
 ---
 '@posthog/rrweb-snapshot': patch
+'posthog-js': patch
 ---
 
 Preserve `<style>` textContent when the browser's CSSOM serialization would
--- a/.changeset/rrweb-snapshot-css-var-shorthand.md
+++ b/.changeset/rrweb-snapshot-css-var-shorthand.md
@@ -1,5 +1,4 @@
 ---
-'@posthog/rrweb-snapshot': patch
 'posthog-js': patch
 ---
 
--- a/.changeset/rrweb-snapshot-css-var-shorthand.md
+++ b/.changeset/rrweb-snapshot-css-var-shorthand.md
@@ -1,4 +1,5 @@
 ---
+'@posthog/rrweb-snapshot': patch
 'posthog-js': patch
 ---
 
--- a/.changeset/rrweb-snapshot-css-var-shorthand.md
+++ b/.changeset/rrweb-snapshot-css-var-shorthand.md
@@ -1,0 +1,15 @@
+---
+'@posthog/rrweb-snapshot': patch
+---
+
+Preserve `<style>` textContent when the browser's CSSOM serialization would
+emit empty longhands from `var()` inside a shorthand. When a stylesheet has
+e.g. `padding: var(--p); padding-bottom: var(--pb);`, browsers store the
+shorthand's longhands with empty token lists per the CSS Custom Properties
+spec, and `CSSStyleRule.cssText` re-emits them as `padding-top: ;
+padding-right: ; padding-left: ;`. The previous behavior replaced the
+`<style>` text with that corrupted output, silently dropping layout rules
+on replay. We now detect the empty-longhand pattern and keep the original
+textContent in that case. Affects users of any CSS-in-JS framework that
+combines `var()` with shorthands (Chakra UI v3, Panda CSS, Emotion, etc.).
+See rrweb-io/rrweb#1626 / #1567.
--- a/packages/browser/playground/chakra-emotion/index.html
+++ b/packages/browser/playground/chakra-emotion/index.html
@@ -1,0 +1,12 @@
+<!doctype html>
+<html lang="en">
+    <head>
+        <meta charset="UTF-8" />
+        <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+        <title>Chakra UI v3 + emotion replay repro</title>
+    </head>
+    <body>
+        <div id="root"></div>
+        <script type="module" src="/src/main.tsx"></script>
+    </body>
+</html>
--- a/packages/browser/playground/chakra-emotion/package.json
+++ b/packages/browser/playground/chakra-emotion/package.json
@@ -1,0 +1,25 @@
+{
+    "name": "chakra-emotion",
+    "private": true,
+    "version": "0.0.0",
+    "type": "module",
+    "scripts": {
+        "dev": "vite",
+        "build": "tsc -b && vite build",
+        "preview": "vite preview"
+    },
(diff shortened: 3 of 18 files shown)


pauldambra:
* **#3542** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3542?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3542?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pauldambra:
THIS IS A BOT NOT IDENTIFYING ITSELF AS ONE DESPITE MY INSTRUCTIONS

-----

Pushed an additional commit (fd48fed6) with extra coverage on top of your fix:

## What's added

**Real-browser Playwright spec** at `packages/browser/playwright/mocked/session-recording/session-recording-stylesheet-layers.spec.ts` — runs on Chromium and Firefox via `pnpm --dir packages/browser playwright`. jsdom can't reproduce the var-in-shorthand CSSOM serialization (your existing tests use `Object.defineProperty` on `cssText` to simulate it); these tests exercise the *real* browser path so we know your fix lands the symptom, not just the simulated one.

Three test cases:
1. `@layer` block + atomic utilities + nested `@media` round-trip cleanly — regression guard for grouping at-rules.
2. **Shorthand-with-`var()` + longhand override** — the exact pattern from `app.sonia.so/login`. The captured `<style>` text must not contain `padding-(top|right|bottom|left): ;` and must contain `padding: var(--chakra-spacing-8)` AND `padding-bottom: var(--chakra-spacing-12)`. Verified to fail on `main` and pass on this branch.
3. Hybrid `<style>` (seeded text + late `sheet.insertRule()` with non-corrupting CSS) captures **both** rules — regression check for the corruption-detection branch in `serializeTextNode`, so it doesn't over-fire and lose `insertRule` additions.

**Synthetic fixture** at `packages/browser/playground/css-layers/index.html` — drives all three test cases via `document.styleSheets` API and `textContent`.

**Chakra-emotion playground** at `packages/browser/playground/chakra-emotion/` — Vite + React + `@chakra-ui/react@^3.27.0` + `@emotion/react@^11.14.0` + posthog-js (linked). Mirrors the customer's stack closely enough that I could verify on `app.sonia.so/login` directly that the live `<style>.textContent` is clean (`padding: var(--x); padding-bottom: var(--y)`) while `rule.cssText` returns the broken longhand split. Useful as a debugging artifact for future CSS-in-JS replay issues; not driven by any automated test today.

## Where I learned this

- Read `@emotion/sheet@1.4.0` source to verify emotion never produces hybrid `<style>` (textContent + insertRule on same element). Confirmed: speedy mode is empty-text + insertRule; non-speedy is one text node per rule. SSR hydration registers existing tags but doesn't write to them.
- Read `styled-components@6.4.1` source: same per-element mode separation, plus `xe` function explicitly removes SSR `<style>` tags after hydration.
- Verified the regression-window for "pre-FullSnapshot insertRule on a seeded `<style>`" by tracing `record/index.ts:651-654` — `takeFullSnapshot()` runs *before* `observe(document)` installs the `insertRule` Proxy in `initStyleSheetObserver`. Your fix's fallback inherits this — same as the previous behaviour for any path through `stringifyStylesheet`.

## Scope

Tests + playground only. No source change. Your fix is the source change.

One observation while testing: the speedy-mode path (empty `<style>` populated via `sheet.insertRule()` *before* recording starts) still routes through `stringifyStylesheet` directly via the `<style>` element handler at `snapshot.ts:670+` and would emit the same empty placeholders for shorthand-with-`var()`+override rules. Your fix in `serializeTextNode` doesn't reach that branch. This is consistent with your "scope: only the inline `<style>` path" note in the PR description, but worth flagging if a follow-up PR wants to cover it.

pauldambra (approved):
one comment from my bot on a route this misses, but that could be done mañana

PR #3545, opened 2026-05-07
ci: strip content hashes from bundled-size diff filenames

pauldambra:
## Summary

The Bundled Size CI check was reporting rollup chunks like `types-D5C8ej05.cjs` and `types-yQPl9Qhh.js` as `+N kB (new file)` on every PR. They aren't new — the source changed, so the rollup content hash changed, and `preactjs/compressed-size-action` was comparing names verbatim. The result is a phantom delete + new file pair on every chunked source change, which drowns the real signal.

`compressed-size-action` already handles this via a `strip-hash` regex input (capture-group contents replaced with `*` before name comparison) — we just weren't passing it.

This PR adds:

```yaml
strip-hash: '-(\w{8})\.'
```

to `.github/workflows/bundled-size.yaml`. It captures the 8-character rollup hash sandwiched between the chunk name and its first extension dot, normalising every variant we emit:

| input                          | normalised                    |
|--------------------------------|-------------------------------|
| `types-D5C8ej05.cjs`           | `types-********.cjs`          |
| `types-yQPl9Qhh.js`            | `types-********.js`           |
| `types-D5C8ej05.umd.cjs`       | `types-********.umd.cjs`      |
| `types-D5C8ej05.umd.min.cjs`   | `types-********.umd.min.cjs`  |

`\w` does not match `-`, so chunk names that contain dashes (e.g. `vendor-react-utils.js`) cannot be misread as a hash. The trailing `\.` anchor keeps the match tight to the hash position.

## Test plan

- [ ] Bundled Size check runs on this PR.
- [ ] PR comment reports rrweb-snapshot chunks under their normalised (`*`) names.
- [ ] No `(new file)` annotations for chunks whose source merely changed.

---

### DIFF
--- a/.github/workflows/bundled-size.yaml
+++ b/.github/workflows/bundled-size.yaml
@@ -47,3 +47,4 @@ jobs:
       - uses: preactjs/compressed-size-action@946a292cd35bd1088e0d7eb92b69d1a8d5b5d76a # pin v2.8.0
         with:
           compression: "none"
+          strip-hash: '-(\w{8})\.'


pauldambra:
* **#3545** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3545?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3545?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/bundled-size.yaml
```diff
@@ -47,3 +47,4 @@ jobs:
       - uses: preactjs/compressed-size-action@946a292cd35bd1088e0d7eb92b69d1a8d5b5d76a # pin v2.8.0
         with:
           compression: "none"
+          strip-hash: '-(\w{8})\.'
```
pauldambra, line 50: > [!NOTE]

**[convergent: paul + xp + qa-team/generalist-a + qa-team/generalist-b]** 🟢 LOW

Two soft convergent observations on this regex:

1. **Character class is broader than needed.** `\w` matches `[A-Za-z0-9_]`. Rollup's content hashes are base64-style without underscores, so `[a-zA-Z0-9]{8}` would express intent more tightly and reduce the chance of accidentally masking non-hash 8-char filename segments (e.g. a future chunk named `pkg-snapshot.cjs` — `snapshot` is 8 word chars between `-` and `.`).

2. **Hash length is hardcoded to 8.** If rollup/vite ever changes the default chunk hash length, this regex silently stops matching and the noisy `+N kB (new file)` entries will return with no error signal. Options if you want to harden: widen to `\w{8,}` to tolerate longer hashes, set `output.chunkFileNames` explicitly in the vite config so the regex has a contract to lean on, or accept the fragility and fix it when it next bites.

Both are non-blocking — symmetric stripping on base+head means false positives just produce slightly weird-looking masked names, not incorrect size diffs.

pauldambra (commented):
QA Swarm review complete — see inline comment.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE

Single-line CI ergonomics fix that enables the existing `strip-hash` input on `preactjs/compressed-size-action`. Pinned action, scoped permissions, no new outbound surface. Reviewers converge on a couple of soft observations about the regex's character class and hash-length assumption — none are merge blockers.

### Key findings

- 🟢 LOW (convergent across 4 reviewers): `\w{8}` is broader than rollup's base64-ish hash alphabet and silently couples to rollup's current 8-char default. Tightening or making the chunk-naming explicit in vite config would future-proof it. See inline comment.

### Convergence

- paul, xp, qa-team/generalist-a, and qa-team/generalist-b independently flagged the same regex robustness concern — higher confidence even at LOW severity.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Workflow trigger (`pull_request`, not `pull_request_target`) and scoped permissions are unchanged; security review is clean. Generalists flagged that the regex hardcodes rollup's 8-char default. |
| 👤 paul | "stamp" — would ship as-is, would tighten if the broader character class ever bites. |
| 📐 xp | Passes the four simplicity rules; ship it. Latent coupling with rollup defaults worth a moment's thought next time someone touches vite. |

---

.github/workflows/bundled-size.yaml
pauldambra: > [!NOTE]

Acknowledged — convergent but explicitly flagged LOW / "ship as-is" by all four reviewers. The current rollup default produces 8-char alphanumeric hashes (e.g. `D5C8ej05`, `yQPl9Qhh`) so `\w{8}` matches correctly today, and false positives only produce slightly weird-looking masked names rather than incorrect size diffs (base + head are stripped symmetrically). Closing as a YAGNI nit — will tighten if rollup ever changes its hash shape and the noise returns.

lucasheriques (approved):

PR #3547, opened 2026-05-07
feat(browser): expose uuid on CaptureOptions for cross-source idempotency

### LINKED ISSUE #3546: Expose uuid on CaptureOptions in posthog-js (web), matching posthog-node and posthog-python

The browser SDK's public `capture()` API doesn't accept a custom event `uuid`, even though every surface around it already supports one:

- The `CaptureResult` type already requires `uuid: string` ([`packages/types/src/capture.ts`](https://github.com/PostHog/posthog-js/blob/main/packages/types/src/capture.ts))
- The non-deprecated public method `calculateEventProperties(eventName, properties, timestamp?, uuid?, readOnly?)` accepts a uuid argument (`_calculate_event_properties` was deprecated in favor of it in 1.241.0)
- `capture()` itself **already passes a uuid** through to `calculateEventProperties` — it just generates one locally with `uuidv7()` and never reads from options ([`packages/browser/src/posthog-core.ts#L1259`](https://github.com/PostHog/posthog-js/blob/main/packages/browser/src/posthog-core.ts#L1259))
- **posthog-node** exposes it on `EventMessage.uuid?: string`
- **posthog-python** exposes it on `OptionalCaptureArgs.uuid` ([`args.py`](https://github.com/PostHog/posthog-python/blob/master/posthog/args.py))

`posthog-js` (browser) is the only mainstream PostHog SDK that doesn't let users provide their own uuid. This blocks the official idempotency pattern recommended by PostHog staff in [#17211](https://github.com/PostHog/posthog/issues/17211):

> "We deduplicate based on the sort key `[timestamp, distinct_id, event, uuid]` … the solution here is to make sure you send the events with the same uuid."
> — @tiina303, 2023-09-18

That guidance is impossible to follow from a browser today.

### Concrete use case

A common architecture: a Stripe (or App Store / Play Store) webhook fires from the backend at payment confirmation, and the browser-side success page also fires the same event with attribution context. Both sides know a shared business identifier (e.g. `transaction_id`). Today:

- The server can compute `uuid = uuidv5("${event}:${transaction_id}", NS)` and pass it to `posthog-node.capture({ uuid, timestamp, … })` ✅
- The browser can compute the same uuid but **has no way to pass it to `posthog.capture()`** ❌ — so the two events never share a `uuid` and PostHog's eventual dedup can't collapse them.

This is the same problem class as the App Store webhook example in [#17211](https://github.com/PostHog/posthog/issues/17211), the page-refresh purchase duplicate in the same thread, and the auto-uuid collision rate documented in [#30692](https://github.com/PostHog/posthog/issues/30692).

Two small changes — one field on `CaptureOptions`, one line in `capture()` to read it:

```ts
// packages/types/src/capture.ts
export interface CaptureOptions {
  // …existing fields
  timestamp?: Date
  uuid?: string  // NEW
}
```

```diff
// packages/browser/src/posthog-core.ts — capture(), around line 1259
  const timestamp = options?.timestamp || systemTime
- const uuid = uuidv7()
+ const uuid = options?.uuid || uuidv7()
```

`calculateEventProperties` already accepts the uuid argument, so no further plumbing is needed.

Usage:

```ts
posthog.capture('purchase', { transaction_id }, {
  uuid: deterministicUuid(event, transaction_id),
  timestamp: serverCanonicalTimestamp,
})
```

Validation behavior to consider: warn (don't throw) on malformed UUIDs, mirroring posthog-node's permissive handling.

## Describe alternatives I've considered

- **Drop client-side fires for transactional events.** Works, but loses the browser-side event-time attribution context (utm, `$current_url`, `$referrer`) that some teams need for ad-platform CAPI parity.
- **Sync only `timestamp` + `$insert_id`** between sides and rely on PostHog's auto-uuid derivation. Brittle — depends on undocumented hash internals; behavior could change between SDK versions.
- **Set `uuid` via a `before_send` hook.** Awkward — `before_send` mutates outgoing events but the auto-uuid has already been generated and assigned by then; replacing it is fragile.

## Related sub-libraries

- [ ] All of them
- [x] `posthog-js` (web)
- [ ] `posthog-js-lite`
- [ ] `posthog-node` *(already supports it)*
- [ ] `posthog-react-native`

## Additional context

- The existing test at `packages/core/src/__tests__/posthog.capture.spec.ts` already asserts `uuid: expect.any(String)` in the captured payload — extending it to assert pass-through of a user-provided uuid would be a small change.
- Is the omission deliberate? If there's a reason to keep browser events on auto-generated uuidv7s (e.g. session-replay correlation, ordering invariants in ingestion), happy to discuss alternatives. Otherwise, glad to send a PR.
- Linked issues for context: [#17211](https://github.com/PostHog/posthog/issues/17211) (idempotency request, open since 2023), [#30692](https://github.com/PostHog/posthog/issues/30692) (auto-uuid collision rate ~0.7%).

williamchong:
## Problem

  Engineers integrating `posthog-js` from the browser have no way to set a deterministic event `uuid`, which blocks the official cross-source idempotency pattern recommended by PostHog staff in [PostHog/posthog#17211](https://github.com/PostHog/posthog/issues/17211):

  > "We deduplicate based on the sort key `[timestamp, distinct_id, event, uuid]` … the solution here is to make sure you send the events with the same uuid."
  > — @tiina303, 2023-09-18

  A common architecture is a server-side webhook (e.g. Stripe `checkout.session.completed`) **and** a browser-side success page both firing for the same business transaction. Today they emit different `uuidv7()`s, so PostHog's eventual deduplication cannot collapse them — every redirect-driven flow ends up with ~2× event inflation.

  `posthog-js` (browser) is the only mainstream PostHog SDK that doesn't expose this:

  - `posthog-node` already has it on `EventMessage.uuid?: string`
  - `posthog-python` already has it on `OptionalCaptureArgs.uuid`
  - `@posthog/core` already has it on `PostHogCaptureOptions.uuid` (with the comment "Check out posthog-js for these additional options and try to keep them in
  sync")

  This PR closes the parity gap. Closes #3546.

  ## Changes

  Two surgical changes in `packages/types` and `packages/browser`, plus a test:

  1. **`packages/types/src/capture.ts`** — added `uuid?: string` to the `CaptureOptions` interface, with a JSDoc explaining the cross-source idempotency use case and linking PostHog/posthog#17211.

  2. **`packages/browser/src/posthog-core.ts:1259`** — `capture()` already passes a uuid through to `calculateEventProperties`; previously it always generated a fresh `uuidv7()` and never read from options. One-line change:

     ```diff
     - const uuid = uuidv7()
     + const uuid = options?.uuid || uuidv7()
     ```

  3. **`packages/browser/src/__tests__/posthog-core-also.test.ts`** — added a test mirroring the existing `'captures when time is overriden by caller'` test, asserting that a caller-supplied `uuid` lands on the captured event.

  4. **`.changeset/`** — minor bump for `posthog-js` (auto-bumps `@posthog/types` via the existing fixed group in `.changeset/config.json`).

  ### Usage

  ```ts
  posthog.capture('purchase', { transaction_id }, {
    uuid: deterministicUuid(eventName, transaction_id), // e.g. uuidv5(`${eventName}:${transaction_id}`, NS)
    timestamp: serverCanonicalTimestamp,
  })
  ```

  Now both the server-side `posthog-node.capture({ uuid, … })` and the browser-side `posthog.capture(name, props, { uuid, … })` can produce identical `(timestamp, distinct_id, event, uuid)` tuples and PostHog's existing dedup will collapse them.

  ### Notes for reviewers

  - I deliberately did **not** add a sentinel property like `$event_uuid_override_provided` (analogous to `$event_time_override_provided`). Adding new `$`-namespaced properties feels like a maintainer call rather than contributor scope. Happy to add one if you'd like — just let me know the preferred property name.
  - No validation on the uuid string format; matches `posthog-node`'s permissive handling. Could add a dev-mode warning if desired.
  - No deprecation needed — this is purely additive on an optional field.

  ### Libraries affected

  - [ ] All of them
  - [x] posthog-js (web)
  - [ ] posthog-node
  - [ ] posthog-react-native
  - [ ] @posthog/react
  - [ ] @posthog/ai
  - [ ] @posthog/convex
  - [ ] @posthog/next
  - [ ] @posthog/nextjs-config
  - [ ] @posthog/nuxt
  - [ ] @posthog/rollup-plugin
  - [ ] @posthog/webpack-plugin
  - [x] @posthog/types

  ## Checklist

  - [x] Tests for new code

  ### If releasing new changes


.changeset/expose-uuid-on-capture-options.md
```diff
@@ -0,0 +1,16 @@
+---
+'posthog-js': minor
+---
+
+`capture()` now accepts an optional `uuid` on `CaptureOptions`, mirroring `posthog-node` (`EventMessage.uuid`) and `posthog-python` (`OptionalCaptureArgs.uuid`). When provided, it overrides the auto-generated `uuidv7()` for that event.
+
+This unblocks the official idempotency pattern recommended in [posthog/posthog#17211](https://github.com/PostHog/posthog/issues/17211): emit a server-side webhook event and a browser-side success-page event for the same business transaction with a deterministic shared uuid (e.g. `uuidv5("${event}:${transaction_id}", NS)`) so PostHog's eventual deduplication can collapse them into one. Previously this was impossible from the browser SDK because every event was assigned a fresh `uuidv7()` inside `capture()` with no way for the caller to override.
```
marandaneto:
```suggestion
`capture()` now accepts an optional `uuid` on `CaptureOptions`
```

.changeset/expose-uuid-on-capture-options.md
```diff
@@ -0,0 +1,16 @@
...
+
+```ts
+posthog.capture('purchase', { transaction_id }, {
+  uuid: deterministicUuid(eventName, transaction_id),
+  timestamp: serverCanonicalTimestamp,
+})
+```
+
+Closes #3546.
```
marandaneto:
```suggestion
```

packages/types/src/capture.ts
```diff
@@ -97,6 +97,14 @@ export interface CaptureOptions {
      */
     timestamp?: Date
 
+    /**
+     * If set, overrides the auto-generated event uuid. Useful for cross-source idempotency
+     * (e.g. a server webhook and a browser success page both firing for the same business
+     * transaction): emit both events with the same deterministic uuid so PostHog can dedupe
+     * them. See https://github.com/PostHog/posthog/issues/17211.
```
marandaneto:
```suggestion
     * them.
```

.changeset/expose-uuid-on-capture-options.md
```diff
@@ -0,0 +1,16 @@
+---
+'posthog-js': minor
```
marandaneto, line 2: you also have to add a changeset for the `types` package

marandaneto:
@williamchong thanks
left a few comments, once addressed, we can merge and release this

marandaneto (approved):

PR #3533, opened 2026-05-05
fix: next/pages server-only components from being imported in the client

brandonhines-mialabs:
## Problem

`@posthog/next/pages` doesn't build in a Next.js Pages Router app following the documented setup. The `./pages` subpath in `package.json#exports` had no per-runtime conditions, so an `import` from `pages/_app.tsx` transitively pulled in `posthog-node` and `'server-only'`. Next.js rejects `'server-only'` in client modules before tree-shaking can drop the unused re-exports, so `next build` fails with:

> `'server-only' cannot be imported from a Client Component module.`

Pages Router users hit this on every Next version that ships the `server-only` enforcement (Next 13+).

## Changes

Split the `./pages` barrel per runtime, mirroring the pattern already used for the `.` entry:

- `src/pages.client.ts` — `browser` condition. Re-exports `PostHogProvider` and `PostHogPageView` only.
- `src/pages.edge.ts` — `edge-light` / `edge` / `worker` conditions. Re-exports `postHogMiddleware`, `PostHogPageView`, `DEFAULT_INGEST_PATH`.
- `src/pages.ts` — `react-server` / `default` conditions. Unchanged; keeps the full Pages Router surface (`getServerSidePostHog`, `getPostHog`, etc.).

Wired into `package.json#exports`:

```json
"./pages": {
    "types": "./dist/pages.d.ts",
    "edge-light": "./dist/pages.edge.js",
    "edge": "./dist/pages.edge.js",
    "worker": "./dist/pages.edge.js",
    "browser": "./dist/pages.client.js",
    "react-server": "./dist/pages.js",
    "default": "./dist/pages.js"
}
```

Existing consumer imports (`from '@posthog/next/pages'`) keep working unchanged; the resolver picks the right file per runtime.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [X] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code
- [X] Accounted for the impact of any changes across different platforms
- [X] Accounted for backwards compatibility of any changes (no breaking changes!)
- [X] Took care not to unnecessarily increase the bundle size

[...]

PR #3549, opened 2026-05-07
fix: set default next pages export to client version, matching the default export

brandonhines-mialabs:
## Problem

Users importing from `@posthog/next/pages` in environments where no specific export condition matches (i.e. the `default` fallback is used) will resolve to the server barrel (`pages.js`), which transitively imports `server-only` and `posthog-node`. This causes build failures in client-side contexts. The root `"."` export already defaults to the client barrel, but `"./pages"` was inconsistent.

## Changes

Updated the `default` export condition for `"./pages"` in `package.json` from `./dist/pages.js` (server) to `./dist/pages.client.js` (client), matching the pattern used by the root `"."` export. Updated the corresponding packaging test assertion to reflect the new default.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [X] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [X] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/posthog-next-pages-default-client.md
+++ b/.changeset/posthog-next-pages-default-client.md
@@ -1,0 +1,5 @@
+---
+'@posthog/next': patch
+---
+
+Fix `@posthog/next/pages` default export condition to resolve to the client barrel (`pages.client.js`) instead of the server barrel (`pages.js`), matching the behavior of the root `"."` export. This prevents bundlers that don't match a more specific condition from pulling in `server-only` and `posthog-node` unnecessarily.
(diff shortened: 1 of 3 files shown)


brandonhines-mialabs:
@dustinbyrne hey sorry, follow up fix for #3533

dustinbyrne (approved):
Thank you!

PR #3550, opened 2026-05-08
fix(replay): preserve recording remote config across posthog.reset()

TueHaulund:
## Problem

When an app calls `posthog.reset()` mid-session (e.g. on sign-out / sign-in), session recording can break for the new session: the recording opens with no Meta + FullSnapshot until the next periodic 5-minute checkout, leaving the player buffering against incremental events that reference an rrweb mirror it never received.

Root cause:

1. `posthog.reset()` calls `this.persistence?.clear()`, which removes **everything** from persistence, including `$session_recording_remote_config`.
2. `posthog.reset()` then calls `sessionManager.resetSessionId()`, which forces a session rotation on the next `checkAndGetSessionAndWindowId(...)` call.
3. The session rotation flows through `_updateWindowAndSessionIds` → `stop()` → `start('session_id_changed')`.
4. `start()` reads `_remoteConfig` (which is just `get_property(SESSION_RECORDING_REMOTE_CONFIG)`), gets `undefined`, and **early-returns** without calling `_startRecorder()`.
5. rrweb has been torn down by `stop()` but never restarted in the rotation path. Subsequent events flow once the SDK eventually re-fetches remote config — but they reference whatever rrweb mirror state was current when rotation happened, not a fresh FullSnapshot. The player buffers until the periodic checkout interval fires.

The remote config is server-defined recording configuration (sample rate, masking, canvas, network capture, triggers, …), not user state. There's no privacy or correctness reason for it to be wiped on `reset()`.

## Changes

- In `reset()`, snapshot `$session_recording_remote_config` before `persistence.clear()` and re-register it afterwards if it was present. Keys whose original value was `undefined` are not re-introduced.
- New `describe('reset()')` block in `posthog-core-also.test.ts` covering:
    - The recording remote config survives `reset()`, while regular user state still gets cleared.
    - `reset()` doesn't crash when no recording remote config has ever been stored.

Other persistence keys (per-session sampling decisions, trigger activations, etc.) correctly continue to be cleared, because they're tied to the session that's about to be discarded.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/preserve-recording-config-on-reset.md
+++ b/.changeset/preserve-recording-config-on-reset.md
@@ -1,0 +1,16 @@
+---
+'posthog-js': patch
+---
+
+Preserve session-recording remote config across `posthog.reset()`.
+
+`posthog.reset()` was clearing the entire persistence store, which wiped
+`$session_recording_remote_config` along with user state. On the next session
+rotation triggered by the reset, `start('session_id_changed')` would early-return
+because the remote config was missing — leaving rrweb torn down and the new
+session opening with no Meta + FullSnapshot until the next periodic 5-minute
+checkout.
+
+This affected any flow where an app calls `posthog.reset()` mid-session
+(e.g. on sign-out / sign-in) and was particularly visible on Flutter Web
+recordings that depend on a fresh FullSnapshot to anchor the CanvasKit DOM.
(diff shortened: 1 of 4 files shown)


arnohillen (approved):

marandaneto:
@PostHog/team-replay can i merge this?

PR #3558, opened 2026-05-09
fix: guard iframe buffer cleanup race

### LINKED ISSUE #3557: rrweb findAndRemoveIframeBuffer crashes when multiple iframes pagehide concurrently
## Summary

`findAndRemoveIframeBuffer` in `packages/rrweb/rrweb/src/record/observer.ts` throws `TypeError: Cannot read properties of undefined (reading 'bufferBelongsToIframe')` when multiple iframes fire `pagehide` near-simultaneously (e.g. tab close, navigation away from a doc with multiple YouTube/Twitter/PDF iframes).

## Source

https://github.com/PostHog/posthog-js/blob/main/packages/rrweb/rrweb/src/record/observer.ts#L379-L387

```ts
export function findAndRemoveIframeBuffer(iframeEl: HTMLIFrameElement) {
  for (let i = mutationBuffers.length - 1; i >= 0; i--) {
    const buf = mutationBuffers[i];
    if (buf.bufferBelongsToIframe(iframeEl)) {
      buf.reset();
      mutationBuffers.splice(i, 1);
    }
  }
}
```

## Root cause

Two `findAndRemoveIframeBuffer` invocations interleave on the shared global `mutationBuffers` array. One call `splice`s while the other is iterating, so `mutationBuffers[i]` returns `undefined` for an index that was valid at loop entry. The next line dereferences `buf.bufferBelongsToIframe` and throws.

This was introduced when `posthog-rrweb` was merged into the monorepo in https://github.com/PostHog/posthog-js/pull/3510. Code originates from the still-open upstream PR https://github.com/rrweb-io/rrweb/pull/1755.

## Impact

Reporting from one Anara production deployment over ~17 days:
- Sentry: 10,343 events, 8,532 unique users
- PostHog error tracking on the same project: 384 dedup-by-session occurrences across 3 issue groups in 30 days

Session recording itself keeps working — the throw is on iframe cleanup, not the outer page observer — but each crash leaks the `MutationBuffer` and its iframe observer, and because the loop exits on the throw, any other buffers queued for cleanup are leaked too.

## Suggested fix

```diff
 export function findAndRemoveIframeBuffer(iframeEl: HTMLIFrameElement) {
   for (let i = mutationBuffers.length - 1; i >= 0; i--) {
     const buf = mutationBuffers[i];
-    if (buf.bufferBelongsToIframe(iframeEl)) {
+    if (buf?.bufferBelongsToIframe(iframeEl)) {
       buf.reset();
       mutationBuffers.splice(i, 1);
     }
   }
 }
```

Optional chaining handles the concurrent-splice race without changing semantics; whichever invocation runs after still cleans up whatever this one skipped.

## Versions

- `posthog-js` 1.372.10 (also reproduced on 1.372.1)
- Browsers: predominantly Chromium (Edge / Chrome), some Safari
- Pages embedding YouTube via `react-lite-youtube-embed`, Twitter via `react-tweet`, and PDF preview iframes

Happy to put up the PR if you want.

marandaneto:
## Problem

Multiple same-origin iframes can fire `pagehide` close together during navigation or tab close. `findAndRemoveIframeBuffer` iterates the shared `mutationBuffers` array while cleanup can shrink it, leaving an `undefined` entry that is dereferenced and reported as `Cannot read properties of undefined (reading 'bufferBelongsToIframe')`.

Fixes #3557.

## Changes

- Guard `findAndRemoveIframeBuffer` with optional chaining before calling `bufferBelongsToIframe`.
- Add a regression test that simulates `mutationBuffers` shrinking during iframe cleanup.
- Add a patch changeset for `@posthog/rrweb`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

Additional affected package: `@posthog/rrweb`.

## Checklist

- [x] Tests for new code

### If releasing new changes

Test note: attempted `pnpm --filter @posthog/rrweb retest -- test/record/memory-leaks.test.ts`, but local `node_modules` is missing/broken (`Cannot find module .../vitest.mjs`).

### DIFF
--- a/.changeset/brave-iframes-clean.md
+++ b/.changeset/brave-iframes-clean.md
@@ -1,0 +1,5 @@
+---
+'@posthog/rrweb': patch
+---
+
+Fix iframe mutation buffer cleanup when iframe pagehide events overlap.
--- a/packages/rrweb/rrweb/src/record/observer.ts
+++ b/packages/rrweb/rrweb/src/record/observer.ts
@@ -379,7 +379,7 @@ function initViewportResizeObserver(
 export function findAndRemoveIframeBuffer(iframeEl: HTMLIFrameElement) {
   for (let i = mutationBuffers.length - 1; i >= 0; i--) {
     const buf = mutationBuffers[i];
-    if (buf.bufferBelongsToIframe(iframeEl)) {
+    if (buf?.bufferBelongsToIframe(iframeEl)) {
       buf.reset();
       mutationBuffers.splice(i, 1);
     }
--- a/packages/rrweb/rrweb/test/record/memory-leaks.test.ts
+++ b/packages/rrweb/rrweb/test/record/memory-leaks.test.ts
@@ -1,7 +1,10 @@
 import { describe, it, expect, beforeEach, vi } from 'vitest';
 import { JSDOM } from 'jsdom';
 import record from '../../src/record';
-import { mutationBuffers } from '../../src/record/observer';
+import {
+  findAndRemoveIframeBuffer,
+  mutationBuffers,
+} from '../../src/record/observer';
 import { IframeManager } from '../../src/record/iframe-manager';
 import type { eventWithTime } from '@posthog/rrweb-types';
 import { createMirror } from '@posthog/rrweb-snapshot';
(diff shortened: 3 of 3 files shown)


marandaneto:
i'm not sure this is the correct fix but its a quick fix at least @PostHog/team-replay 
you prob. need a refactor here to avoid the race condition, i will leave that up to you

dustinbyrne (approved):

turnipdabeets (approved):

PR #3568, opened 2026-05-11
fix: validate native gzip output

marandaneto:
## Problem

Native browser `CompressionStream("gzip")` can fail by resolving malformed or partial gzip output without throwing. When that happens, posthog-js sends the bad payload with gzip compression enabled and capture drops the event instead of retrying safely.

## Changes

- Validate native gzip output before returning it from `@posthog/core` by checking the gzip header, CRC32 trailer, and uncompressed size trailer.
- Throw a recognizable `NativeGzipValidationError` when native gzip output is malformed.
- Disable native async gzip for the current page lifetime after validation failures and fall back to the existing synchronous `fflate` gzip path, preserving compression.
- Add regression tests for malformed native gzip output and browser request fallback behavior.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/native-gzip-validation.md
+++ b/.changeset/native-gzip-validation.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/core': patch
+---
+
+Validate native gzip output before sending requests and fall back when CompressionStream returns malformed data.
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -594,6 +594,11 @@ describe('request', () => {
             jest.doMock('@posthog/core', () => ({
                 ...jest.requireActual('@posthog/core'),
                 gzipCompress: mockedIsolatedGzipCompress,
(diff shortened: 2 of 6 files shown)


turnipdabeets (approved):

marandaneto:
the ideal world would be:

```
    const stream = new Blob([bytes]).stream()
    const decompressed = stream.pipeThrough(new DecompressionStream("gzip"))

    // Force the stream to fully read/decompress
    await new Response(decompressed).arrayBuffer()
```

but then we'd need to pay the perf price for every request, so lets do a light check with the important parts first and see if it helps with most use cases of broken gzip payloads

PR #3569, opened 2026-05-11
fix(deadclicks): ignore clicks nested inside interactive ancestors

christiaan-ph:
## Problem

`$dead_click` events fire on pages that use `` (or other non-interactive elements) nested inside `<a>` tags. Clicking the image navigates fine, but the click target is the ``, not the anchor, so the dead-click detector's tag check misses the interactive parent and reports a dead click. The pattern shows up frequently on sites built with Framer.

## Changes

Replaces the leaf-only tag check in `_ignoreClick` with an ancestor walk via `closest()`. The walk uses the existing `autocaptureCompatibleElements` list joined into a selector, hoisted to `autocaptureCompatibleElementsSelector` for reuse.

Adds parameterized tests covering nested clicks inside each compatible element, plus a regression test for nested clicks inside non-interactive ancestors.

### Open questions for reviewers

1. **Selector scope.** I reused the full compatible-elements list (`a, button, form, input, select, textarea, label`) for the ancestor walk. The motivating bug only needs `a`/`button`. A click in the inert padding of a long `<form>` is arguably a real dead click that this change would now silently suppress. Happy to narrow to `['a', 'button']` if folks prefer fewer false negatives over consistency with autocapture.

2. **Partial parity with autocapture.** This mirrors autocapture's ancestor-walk pattern, but autocapture also pierces shadow roots and treats `cursor: pointer` parents as interactive. Shadow-root behavior is documented in a code comment; let me know if I should expand the fix to cover those cases or leave them as follow-ups.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes


marandaneto:
@christiaan-ph  is that the same as https://github.com/PostHog/posthog-js/releases/tag/posthog-js%401.374.0 released by Paul?

christiaan-ph:
@marandaneto ah yeah excellent! Glad to see it's already fixed 🙌

PR #3570, opened 2026-05-12
fix(record): stop leaking same-origin iframe state on removal

gruessi:
## Problem

Recording a page that mounts and unmounts same-origin iframes leaks each iframe's `Document`, every node serialized into the mirror, one `MutationObserver` per mount, and its `Window`. End-to-end heap test against a host that cycles 5 blob-URL iframes every 2s for 110s: **+118 MB / +390 leaked `HTMLDocument`s, ~300 KB per mount**. Long-running tabs eventually hit the browser's memory limit.

Cross-origin and `about:blank` iframes go through a different code path and weren't affected.

## Changes

Eight independent retention chains in `packages/rrweb/`. Each was masking the next — fixing any subset still showed leak in heap snapshots.

- **Listener leaks.** `onceIframeLoaded` now returns a disposer (load listener + timer); `IframeManager`'s `pagehide` handler is named and tracked. Both are removed in `removeIframeById`.
- **`recordCrossOriginIframes` cleanup gate dropped** so the same-origin path actually runs cleanup. `cleanupDetachedIframes()` sweeps orphans when an iframe is removed via its parent.
- **`iframe.src = 'about:blank'` before unmount** (typical React cleanup) swaps `contentDocument`/`contentWindow` so the mirror walks an empty doc and `nestedIframeListeners: Map<Window>` misses its entry. New `attachedDocuments` / `attachedWindows` WeakMaps capture every doc/window observed across the iframe's lifetime; detach walks all of them.
- **Global arrays accumulated forever.** `mutationBuffers[]` and `handlers[]` had no iframe-removal path. `findAndRemoveIframeBuffer` now matches by original doc; new `runAndDetachIframeCleanup` splices closures out of `handlers[]`.

<details>
<summary>Per-chain detail (file-by-file)</summary>

1. **`rrweb-snapshot/src/snapshot.ts` — `onceIframeLoaded`.** Returns a disposer that removes the listener and clears the pending timer. The `readyState === 'complete'` branch previously registered a *second* listener next to `setTimeout(listener, 0)` — that one is now named (`onSubsequentLoad`) and removed by the same disposer. `serializeNodeWithId`/`snapshot` gain an optional `onIframeListenerRegistered?(iframe, disposer)` callback so the disposer travels back to `IframeManager`.
2. **`IframeManager.attachIframe`.** Anonymous `pagehide` handler became a named function tracked in `pageHideHandlers: WeakMap<HTMLIFrameElement, Set<{win, handler}>>`. Set-of-handlers because an iframe can run through several Windows during its lifetime.
3. **`record/index.ts` `wrappedMutationEmit`.** Cleanup gate on `recordCrossOriginIframes` removed — same-origin leaks too.
4. **`IframeManager.cleanupDetachedIframes()`.** Sweeps `attachedIframes` for ids no longer in the mirror to catch nested iframe removals (mutation buffer only reports the directly-removed ancestor).
5. **`attachedDocuments: WeakMap<HTMLIFrameElement, Set<Document>>`.** Captures every contentDocument observed across an iframe's lifetime (initial blank → loaded src → cleanup blank). Detach walks the Set and calls `mirror.removeNodeFromMap(doc)` on each.
6. **`mutationBuffers[]` cleanup.** `MutationBuffer.bufferDoc()` exposes the doc; `findAndRemoveIframeBuffer` takes optional `knownDocs: Set<Document>` to match by original contentDocument. `reset()` nulls `this.doc` as defence in depth.
7. **`handlers[]` cleanup.** `runAndDetachIframeCleanup(iframeId)` runs the cleanup and splices the closure from both `iframeObserverCleanups` and `handlers`. `iframeObserverCleanups` is `Map<number, Set<cleanup>>` so navigations don't overwrite earlier cleanups.
8. **`attachedWindows: WeakMap<HTMLIFrameElement, Set<Window>>`.** Mirrors `attachedDocuments`. `removeIframeById` walks the captured Set and removes each entry from `nestedIframeListeners` and `crossOriginIframeMap`. Fixes a stale code comment that incorrectly claimed `Map<Window>` keys don't prevent GC.

</details>

## Validation

End-to-end heap snapshot, same browser, same recorder config, 61 cycles per run (≈305 mount/unmount events).

| | Unpatched | After fix |
|---|---:|---:|
| Post-GC heap delta over 110s | **+118 MB** | **−5 MB** |
| `native:HTMLDocument` Δ | +390 | 0 |
| `native:MutationObserver` Δ | +391 | 0 |
| `native:MutationObserverRegistration` Δ | +390 | −1 |
| `closure:onIframeLoad` Δ | +195 | +1 |

The negative post-GC delta is because the heap-snapshot operation forces a thorough GC that releases other accumulated state too — the real read is the constructor-level deltas. Top growth in the post-cycle snapshot is V8/JIT internals and transient `PerformanceLongTaskTiming` entries, not leaked iframe state.

## Tests

`packages/rrweb/rrweb/test/record/memory-leaks.test.ts` — **30/30 pass** (4 new regressions for this fix: same-origin cleanup, nested parent removal, captured-doc walk, multi-doc lifecycle). `check-types` clean for both `@posthog/rrweb` and `@posthog/rrweb-snapshot`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

The changeset bumps `@posthog/rrweb` and `@posthog/rrweb-snapshot` (patch). `updateInternalDependencies: "patch"` cascades the bump through `rrweb-record` → `posthog-js`.

## Checklist

- [x] Tests for new code

No public API changes. The new `onIframeListenerRegistered?` option is optional and additive; everything else is internal to `IframeManager` / `MutationBuffer` / `record/index.ts`. Bundle-size impact is small (~handful of WeakMap declarations + cleanup helpers) — worth confirming via bundle-size CI on this PR.

### If releasing new changes


pauldambra (approved):
looks safe to me  
i feel like we're accruing a very similar "disposables" pattern across the code  
but that's not the fault of this PR, or blocking here

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -180,32 +180,12 @@ function record<T = eventWithTime>(
...
-      } catch {
-        // Cross-origin: contentDocument access throws, cleanup anyway
-        cleanup();
-        iframeObserverCleanups.delete(iframeId);
-      }
-    }
-  }
+  // Forward-declared; assigned inside the try{} block where `handlers` is in scope.
+  let runAndDetachIframeCleanup: (iframeId: number) => void = () => {};
+  let cleanupDetachedIframeObservers: () => void = () => {};
```
pauldambra: > [!NOTE]

**[convergent: paul + xp + qa-team/frontend]** 🟡 MEDIUM

Forward-declaring `runAndDetachIframeCleanup` / `cleanupDetachedIframeObservers` as `() => {}` and reassigning inside the `try{}` block makes construction order load-bearing in a quiet way. Anything that ends up invoking `wrappedMutationEmit` before the try-block executes gets a silent no-op cleanup — there's no warning, the leak just comes back.

Three refactors that would tighten this:
- Extract `handlers` + `iframeObserverCleanups` into a small `IframeObserverRegistry` class with `register(id, handler)` / `detach(id)` / `cleanupDetached()` methods. The class owns its own state, no forward-declared no-ops needed.
- Or: `let runAndDetachIframeCleanup: ((id: number) => void) | undefined` and call `runAndDetachIframeCleanup?.(id)` — silent misuse becomes a no-op only when *intentionally* unset, and a future caller that fires too early visibly does nothing rather than swallowing a leak.
- Or: lift `handlers` to outer scope so the closures don't need to be reassigned at all.

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -275,27 +255,37 @@ function record<T = eventWithTime>(
...
-        if (!addedIds || !addedIds.has(id)) {
-          // Disconnect observers for this iframe to prevent memory leaks
-          const cleanup = iframeObserverCleanups.get(id);
-          if (cleanup) {
-            cleanup();
-            iframeObserverCleanups.delete(id);
+      const addedIframeElements = new Set<HTMLIFrameElement>();
+      if (m.adds.length > 0) {
+        for (const add of m.adds) {
+          const node = mirror.getNode(add.node.id);
```
pauldambra, line 270: > [!NOTE]

**[paul]** 🟠 HIGH

the reparent detection uses `mirror.getNode(add.node.id)` to map the new id back to the element. that only works because `MutationBuffer.emit` drains `mapRemoves` _before_ `pushAdd` serialises the new id. that's a deep cross-file coupling — a change to emit's ordering would silently corrupt reparent detection and we'd start tearing down observers on every iframe move.

could we add a comment over in `MutationBuffer.emit` pointing at this expectation, or a test that fails when emit's drain order changes? a silent regression here would look like "iframes occasionally stop recording after a move" — exactly the sort of thing that's painful to track down later. ship as you see fit, but i'd want a tripwire.

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -275,27 +255,37 @@ function record<T = eventWithTime>(
...
+          // Reparent: keep observers/listeners; just drop stale id mapping.
+          iframeManager.forgetIframeId(id);
+          return;
+        }
+        runAndDetachIframeCleanup(id);
+        iframeManager.removeIframeById(id);
       });
 
-      // Safety net: cleanup any iframes that have become detached or inaccessible
+      // Catch iframes removed inside a removed subtree (only the ancestor's
```
pauldambra, line 289: > [!NOTE]

The two cleanup orderings disagree. On the explicit `m.removes` path (around lines 282-284) `runAndDetachIframeCleanup(id)` runs **before** `iframeManager.removeIframeById(id)`, so the `MutationObserver` is disconnected before `findAndRemoveIframeBuffer` resets the buffer and nulls `this.doc`. But the safety-net path here runs `iframeManager.cleanupDetachedIframes()` (which transitively calls `buf.reset()` -> `this.doc = null`) **before** `cleanupDetachedIframeObservers()` disconnects the observer.

Combined with the new `reset()` nulling `this.doc`, a mutation queued in the microtask window between those two calls would land on a buffer whose `doc` is null and throw inside `pushAdd`'s `serializeNodeWithId({ doc: this.doc, ... })`. The race window is narrow (browsers deliver MutationObserver callbacks via microtasks) but real.

Suggested fix: swap the order to `cleanupDetachedIframeObservers()` first, then `iframeManager.cleanupDetachedIframes()`. That matches the explicit-remove path and closes the race.

packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -254,6 +254,12 @@ export default class MutationBuffer {
   public reset() {
     this.shadowDomManager.reset();
     this.canvasManager.reset();
+    // Defence in depth: clear doc in case a stale closure still pins the buffer.
+    this.doc = null as unknown as observerParam['doc'];
```
pauldambra: > [!NOTE]

**[convergent: paul + xp + qa-team/reliability]** 🟡 MEDIUM

`this.doc = null as unknown as observerParam['doc']` — three reviewers independently snagged on this. Either:

- `doc` really can be null after reset, in which case make the field's type honest (`Document | null`) and fix the few call sites that read it (`pushAdd`'s `serializeNodeWithId({ doc: this.doc, ... })`, `addShadowRoot(..., this.doc)`), or
- The buffer simply isn't used post-reset and the null-assignment isn't needed at all — let the GC collect the buffer naturally now that it's spliced from `mutationBuffers[]`.

The "defence in depth" comment + the double-cast suggests neither path was fully committed to. Pick a story. Bonus: `public bufferDoc(): Document | null` on the next line claims nullability but the field is typed non-null — that inconsistency is the same smell.

If you keep the null-assignment, please also harden `processMutation`/`emit` with a `if (!this.doc) return;` guard so a late MutationObserver callback during the cleanup race (see the related comment on `wrappedMutationEmit`) is harmless instead of an exception.

packages/rrweb/rrweb/src/record/iframe-manager.ts
```diff
@@ -28,13 +29,29 @@ export class IframeManager {
...
   private messageHandler: (message: MessageEvent) => void;
-  // Map window to handler for cleanup - windows are browser-owned and won't prevent GC
+  // Strong Map — keys pin Windows; every entry must be deleted on detach.
   private nestedIframeListeners: Map<Window, (message: MessageEvent) => void> =
     new Map();
+  // Originals captured per iframe so cleanup survives iframe.src swaps.
+  private attachedWindows: WeakMap<HTMLIFrameElement, Set<Window>> =
+    new WeakMap();
+  private attachedDocuments: WeakMap<HTMLIFrameElement, Set<Document>> =
+    new WeakMap();
```
pauldambra, line 39: > [!NOTE]

**[convergent: xp + paul + qa-team/performance + qa-team/compatibility]** 🟡 MEDIUM

Four-plus WeakMaps/Maps keyed by `HTMLIFrameElement` (`attachedWindows`, `attachedDocuments`, `loadListenerDisposers`, `pageHideHandlers`, plus the strong-keyed `iframeElementsById`) are a textbook **Data Clump**. They're populated together in `attachIframe` and cleared together in `removeIframeById` / `destroy`. Every retainer chain ends up with its own `let bucket = …; if (!bucket) …` boilerplate.

One `WeakMap<HTMLIFrameElement, IframeAttachmentState>` where `IframeAttachmentState = { windows: Set<Window>; documents: Set<Document>; loadDisposers: Set<() => void>; pageHide: Set<{ win: Window; handler: () => void }> }` would name the concept once. Cleanup becomes one `delete`; the "one bucket per element" pattern lives in a single helper.

Secondary observation: `iframeElementsById` is a strong `Map`, while the others are `WeakMap` — that's deliberate (you need to look up by id) but worth a comment that this map MUST be cleared on every detach pathway or it pins iframe elements forever.

Observability nudge while you're in here — five separate retainer chains is a lot to keep right forever. A debug-only counter that logs `attachedDocuments.size` / `loadDisposers.size` etc. crossing a threshold would let us see the next regression before users do.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -339,54 +339,88 @@ export function needMaskingText(
...
   }
+
   if (readyState !== 'complete') {
-    const timer = setTimeout(() => {
-      if (!fired) {
-        listener();
-        fired = true;
+    let fired = false;
+    let timer: ReturnType<typeof setTimeout> | null = null;
+    const onLoad = () => {
```

**[xp]** 🟡 MEDIUM

The "still loading" branch's `onLoad` calls `iframeEl.removeEventListener('load', onLoad)` after firing, so subsequent loads no longer re-fire the listener. Compare to the "complete" branch right below this, which carefully wires `onSubsequentLoad` precisely to keep firing on later loads.

This asymmetry is either:
- Intentional — then please leave a comment explaining why a still-loading iframe doesn't deserve subsequent-load handling, or
- An oversight — the PR description says "Accumulate every contentDocument across loads", which strongly suggests this branch should also keep the listener attached after the first load (just behind the `fired` guard) so an `iframe.src` swap *while the iframe is still loading the initial document* gets its second contentDocument tracked too.

The latter feels more consistent with the rest of the diff's intent.

packages/rrweb/rrweb/src/record/iframe-manager.ts
```diff
@@ -116,20 +187,36 @@ export class IframeManager {
...
+          wins = new Set();
+          this.attachedWindows.set(iframeEl, wins);
+        }
+        wins.add(win);
       });
     }
 
-    callSafely(() =>
-      iframeEl.contentWindow?.addEventListener('pagehide', () => {
+    callSafely(() => {
```
pauldambra, line 200: > [!NOTE]

`attachIframe` adds a pagehide listener on every call without checking for an existing entry. A reparented iframe (which calls `attachIframe` again from `MutationBuffer.emit`'s `onIframeLoad` after the move) will accumulate N pagehide listeners on the same Window. The handlers are idempotent (`mirror.removeNodeFromMap` is a no-op for an unknown id, etc.), so this isn't a correctness bug — but the `pageHideHandlers` Set grows on every reparent, and every listener fires once per event.

Dedupe before adding:
```ts
if (Array.from(bucket).some(({ win }) => win === pageHideWindow)) return;
```

packages/rrweb/rrweb/src/record/iframe-manager.ts
```diff
@@ -360,33 +447,80 @@ export class IframeManager {
...
+    // attachedIframes / mirror may both be empty for iframes removed
+    // before first load; iframeElementsById covers that case.
     const iframe =
       entry?.element ||
+      this.iframeElementsById.get(iframeId) ||
       (this.mirror.getNode(iframeId) as HTMLIFrameElement | null);
 
+    this.iframeElementsById.delete(iframeId);
+
     if (iframe) {
```
pauldambra, line 466: > [!NOTE]

**[xp]** 🟢 LOW

`removeIframeById` is now ~60 lines doing seven distinct things: resolve element, clear nested-Window listeners, clear legacy-Window path, clear crossOrigin map, drop iframe, dispose load listeners, remove pagehide, walk captured docs. Each block has a clarifying comment — those comments are extract-method opportunities (`detachNestedWindowListeners(iframe)`, `releaseCapturedDocuments(iframe)`, `disposeLoadListeners(iframe)`, etc.).

Kent Beck's ComposedMethod: keep all operations in a method at the same level of abstraction. Right now the method mixes the "detach observers" abstraction with the "release captured DOM refs" abstraction with the "forget the iframe" abstraction. The next maintainer has to read the whole thing to follow one retainer chain. Not blocking — works today — but the shape is worth pressing on before this file grows another chain.

packages/rrweb/rrweb/test/record/memory-leaks.test.ts
```diff
@@ -842,5 +841,435 @@ describe('memory leak prevention', () => {
...
+      // observer setup complete.
+      await new Promise((r) => setTimeout(r, 50));
+      const buffersBeforeMove = mutationBuffers.length;
+
+      // Move (remove + add in one atomic mutation).
+      c2.appendChild(iframe);
+      await new Promise((r) => setTimeout(r, 50));
+
+      // Move must not have spliced any per-iframe MutationBuffer entries.
+      expect(mutationBuffers.length).toBeGreaterThanOrEqual(buffersBeforeMove);
```
pauldambra, line 1227: > [!NOTE]

**[paul]** 🟢 LOW

the reparent test asserts `mutationBuffers.length >= buffersBeforeMove`. that's deliberately permissive and fine for catching the bug here — but a future regression that _adds_ buffers per move would still pass.

worth tightening to `===` if the count is stable across the reparent, or splitting into a separate assertion on the iframe-specific buffer count (e.g. count buffers whose `doc` matches the moved iframe's contentDocument). otherwise we've shipped a test that's load-bearing for one direction of regression and silent on the other.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments and the summary comment for the full verdict.

pauldambra:
> [!NOTE]
>

The leak fix itself is sound and the +118 MB / +390 leaked HTMLDocuments -> ~0 validation is a real receipt. The requested changes are about robustness and maintainability: one narrow correctness concern (safety-net cleanup ordering can race against the new `MutationBuffer.reset()`), a load-bearing reparent invariant that needs a comment or test, and three convergent design smells in the new state plumbing.

### Key findings

**🟠 High**
- **`packages/rrweb/rrweb/src/record/index.ts:267`** — reparent detection silently couples to `MutationBuffer.emit`'s drain order (`mapRemoves` before `pushAdd`). A future change there would tear down observers on every iframe move with no test failure. Add a comment in `emit` or a regression test.

**🟡 Medium**
- **`packages/rrweb/rrweb/src/record/index.ts:286`** *(qa-team/reliability)* — the safety-net path runs `iframeManager.cleanupDetachedIframes()` (which nulls `MutationBuffer.doc`) **before** `cleanupDetachedIframeObservers()` disconnects the observer. Swap the order to match the explicit-remove path and close the narrow race window.
- **`packages/rrweb/rrweb/src/record/mutation.ts:258`** *(convergent — paul + xp + qa-team)* — `this.doc = null as unknown as observerParam['doc']` plus `bufferDoc(): Document | null` on a non-null field. Pick a story: either make the field properly nullable and guard the call sites, or drop the null-assignment entirely.
- **`packages/rrweb/rrweb/src/record/index.ts:188`** *(convergent — paul + xp + qa-team/frontend)* — forward-declared `runAndDetachIframeCleanup` / `cleanupDetachedIframeObservers` no-ops reassigned inside `try{}`. Construction order is now load-bearing in a way that fails silently. Extract an `IframeObserverRegistry` class or use `let f: T | undefined; f?.(...)`.
- **`packages/rrweb/rrweb/src/record/iframe-manager.ts:39`** *(convergent — xp + paul + qa-team/performance + qa-team/compatibility)* — Data Clump: 4 WeakMaps + 1 strong Map all keyed by `HTMLIFrameElement`, set/cleared together. Collapse to one `WeakMap<HTMLIFrameElement, IframeAttachmentState>` so the concept is named once.
- **`packages/rrweb/rrweb-snapshot/src/snapshot.ts:365`** *(xp)* — the "still loading" branch removes its load listener after the first fire; the "complete" branch right below carefully keeps re-firing via `onSubsequentLoad`. Either intentional (needs a comment) or an oversight that misses contentDocument swaps during initial load.

**🟢 Low**
- **`iframe-manager.ts:200`** — `attachIframe` adds a pagehide listener every call; reparented iframes accumulate N listeners on the same Window. Dedupe.
- **`iframe-manager.ts:459`** — `removeIframeById` is now ~60 lines doing 7 distinct things. Extract `detachNestedWindowListeners`, `releaseCapturedDocuments`, `disposeLoadListeners` per ComposedMethod.
- **`memory-leaks.test.ts:1224`** — reparent test uses `>=` which catches the current regression but is silent on a future one that *adds* buffers per move. Tighten to `===` or split the assertion.

### Convergence

Three findings were flagged independently by all three reviewers (highest confidence):

| Convergent finding | Reviewers | Location |
| --- | --- | --- |
| Forward-declared no-op cleanup pattern | paul + xp + qa-team/frontend | `index.ts:186-188` |
| `this.doc = null` cast / `bufferDoc(): \| null` inconsistency | paul + xp + qa-team/reliability | `mutation.ts:254-261` |
| Data Clump across iframe state WeakMaps | xp + paul + qa-team/performance + qa-team/compatibility | `iframe-manager.ts:36-49` |

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Substantial, well-motivated fix. Main concern is the cleanup-ordering inconsistency in `wrappedMutationEmit` between the explicit-remove path and the safety-net path now that `MutationBuffer.reset()` nulls `this.doc`. Other findings are LOW/NIT. New tests cover the right regressions. No security, data-integrity, copy, or compatibility concerns. |
| 👤 paul | Solid work, +118MB -> ~0 is a real receipt. Biggest concern is the new cross-file coupling between `wrappedMutationEmit`'s reparent detection and `MutationBuffer.emit`'s drain ordering — wants a comment or test. The `this.doc = null as unknown as ...` cast is a small smell. Everything else is preference. Ship after a brief chat about the reparent invariant and the cast. |
| 📐 xp | Closes a real, validated leak with thorough tests. Passes the four rules of simple design overall, but the main smell is the Data Clump in `IframeManager` — four parallel WeakMaps keyed by the same element want to be one state object. Secondary smells: subtle behaviour change in `onceIframeLoaded`'s "still loading" branch, forward-declared mutable function bindings papering over a scoping issue, and a small encapsulation leak via `MutationBuffer.bufferDoc()`. |

---

marandaneto:
@gruessi would you address the open comments so we can merge and release this? thanks!

packages/rrweb/rrweb/src/record/index.ts
gruessi: Fixed in 84f9802f via the smaller of your three suggestions. Both helpers are now typed `((id: number) => void) | undefined` and called via `?.()`. A premature call is visibly a no-op rather than silently swallowing a leak. Holding off on the `IframeObserverRegistry` extraction in this PR though.

packages/rrweb/rrweb/src/record/index.ts
gruessi: Fixed in 84f9802f, added a comment over the `mapRemoves` drain in `MutationBuffer.emit` that points at the reparent detection in `record/index.ts` and warns future maintainers not to reorder. Held off on a regression test because asserting drain order from outside requires exposing more `MutationBuffer` internals than I'd like; happy to add one if you have a shape in mind.

packages/rrweb/rrweb/src/record/index.ts
gruessi: Fixed in 84f9802f, swapped to `cleanupDetachedIframeObservers()` then `iframeManager.cleanupDetachedIframes()`. Both paths now agree.

packages/rrweb/rrweb/src/record/mutation.ts
gruessi: Fixed in 84f9802f, went with option B. Dropped the null-assignment entirely; `bufferDoc()` returns `Document` (matches the field). The buffer's only strong refs are `mutationBuffers[]` + the cleanup closure in `handlers[]`, and both are removed before `reset()` is reached, so GC handles the doc. The nulling was creating the same race you flagged on the other thread, worse than the hedge it was meant to be.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
gruessi: Fixed in 84f9802f, was an oversight, not intentional. Still-loading branch now attaches `onSubsequentLoad` after the initial fire, matching the complete branch. An iframe.src swap while the initial document is still loading now gets its second contentDocument serialized too. Sharing the disposer wiring across both cases was the right pull.

packages/rrweb/rrweb/src/record/iframe-manager.ts
gruessi: Fixed in 84f9802f ,`attachIframe` now scans the per-iframe bucket for an existing entry with the same Window and skips registration if found.

packages/rrweb/rrweb/src/record/iframe-manager.ts
gruessi: Agreed, seven inline blocks at different abstraction levels asks to be extracted into `detachNestedWindowListeners(iframe)` / `releaseCapturedDocuments(iframe)` / etc. I'd rather land the leak fix as a clean diff handle follow up with a cleanup. Worth doing before the next chain lands in this file.

packages/rrweb/rrweb/test/record/memory-leaks.test.ts
gruessi: Tried `===` — fails because buffers actually grow 2 → 4 on reparent today: the add side re-serializes and `attachIframe` starts a fresh observer alongside the kept old one. Separate issue from this leak, out of scope. Kept `>=` and rewrote the test comment so the next reader reads it as a guarded floor, not a guarantee.

packages/rrweb/rrweb/src/record/iframe-manager.ts
gruessi: Agreed in shape, single `WeakMap<HTMLIFrameElement, IframeAttachmentState>` would be cleaner. Out of scope here though; the refactor would touch every leak-path I just landed and obscure what's new in this diff.

Adding the comment on `iframeElementsById` being strong-keyed.

gruessi:
> @gruessi would you address the open comments so we can merge and release this? thanks!

@marandaneto thanks for the ping, and @pauldambra thanks for the review. I replied to all QA Swarm comments and addressed hopefully everything in the verdict with 84f9802f. Let me know if anything is not adding up 🙏

marandaneto:
@pauldambra are you happy with their changes? should i merge if CI is green?

pauldambra:
i'll defer to @tue and @ksvat

gruessi:
Hey @TUE & @ksvat, gentle nudge on this one when you get a moment 🙏 Let me know if I can do anything to help the process!

ksvat:
Hey @gruessi i'm happy to merge this once CI is passing/green!

.changeset/iframe-memory-leak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/rrweb': patch
+'@posthog/rrweb-snapshot': patch
```
marandaneto: @PostHog/team-replay is that correct? iirc we dont release those packages alone anymore, should we just release posthog-js instead?

ksvat: Yes, you're right, We only release posthog-js now

ksvat: @gruessi can you please add 
'posthog-js': patch
to this changeset! thank you!

gruessi: @ksvat Done! The changeset now only bumps `posthog-js`. Hope that works!

PR #3581, opened 2026-05-12
fix(core): detect Oculus Browser instead of falling back to Chrome

Ashut0sh-mishra:
## Problem

Events generated on Meta Quest headsets are reported with browser `Chrome`
instead of `Oculus` because the Oculus Browser user agent contains both
`OculusBrowser` and `Chrome` (and on newer firmware also `SamsungBrowser`),
and the existing detection order returns the first matching branch.

Closes #3574

## Fix

- Add `OculusBrowser` detection in `detectBrowser` *before* the
  `SamsungBrowser` / `Chrome` branches in `packages/core/src/utils/user-agent-utils.ts`.
- Add a corresponding entry in `versionRegexes` so `detectBrowserVersion`
  returns the actual Oculus Browser version.
- Constants follow the existing pre-uglifying string-concatenation style noted
  in the file header comment.

## Tests

Added two new cases to `packages/browser/src/__tests__/utils/user-agent-utils.test.ts`:

- Quest 2 UA (older format, with `SamsungBrowser` segment).
- Quest 3-style UA (no `SamsungBrowser` segment).

Both assert `expectedBrowser: 'Oculus Browser'` and the correct
`expectedVersion`, exercising the same `detectBrowser` / `detectBrowserVersion`
suite the other browsers in this file go through.

Reference UA list: https://whatmyuseragent.com/browser/oc/oculus-browser

### DIFF
--- a/.changeset/oculus-browser-detection.md
+++ b/.changeset/oculus-browser-detection.md
@@ -1,0 +1,5 @@
+---
+"@posthog/core": patch
+---
+
+Detect Oculus Browser (Meta Quest headsets) correctly instead of falling back to Chrome
--- a/packages/browser/src/__tests__/utils/user-agent-utils.test.ts
+++ b/packages/browser/src/__tests__/utils/user-agent-utils.test.ts
@@ -205,6 +205,25 @@ describe('user-agent-utils', () => {
                 expectedVersion: 28.0,
                 expectedBrowser: 'Chrome',
             },
+            {
+                // see https://github.com/PostHog/posthog-js/issues/3574
+                // Oculus Browser is Chromium-based, so its UA includes both
+                // `OculusBrowser` and `Chrome` (and on newer headsets also `SamsungBrowser`).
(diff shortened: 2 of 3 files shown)


Ashut0sh-mishra:
Thanks @greptile-apps! Inlined OCULUS directly into OCULUS_BROWSER -- you're right, it was only used once. Pushed.

marandaneto:
@Ashut0sh-mishra thanks, pls add a changeset entry https://github.com/PostHog/posthog-js/blob/main/RELEASING.md

marandaneto (approved):

Ashut0sh-mishra:
@marandaneto  Done — added .changeset/oculus-browser-detection.md (@posthog/core: patch). Thanks!

Ashut0sh-mishra:
@marandaneto Thanks again for the helpful review feedback and for merging this. I really appreciate it.

PR #3584, opened 2026-05-12
fix(ai): preserve Error message, stack, and cause chain on $ai_error

### LINKED ISSUE #3556: @posthog/ai: $ai_error loses error fields (message, stack) and cause chain because json.stringify is used directly on error instances
## Bug description

When an error is thrown from inside a model wrapped by withTracing (e.g. a GatewayResponseError from @ai-sdk/gateway carrying a cause chain), @posthog/ai serializes it via json.stringify() before storing it on the $ai_generation event's $ai_error property. Javascript error objects store properties like message, stack, and cause as non enumerable, so when serialized, usually only name remains.

The end result on the posthog dashboard is a stripped down error blob with no message, no stack, and no usable cause chain making it nearly impossible to debug what actually went wrong.

## To reproduce

Tested against `@posthog/ai@7.11.2` and `@posthog/ai@7.18.1` same behaviour.

```ts
import { createGateway, generateText } from "ai";

const phClient = new PostHog(process.env.POSTHOG_KEY!, { host: process.env.POSTHOG_HOST });
const gateway = createGateway({ apiKey: process.env.AI_GATEWAY_API_KEY });

const tracedModel = withTracing(gateway("google/gemini-3-pro-image"), phClient, {
  posthogDistinctId: "test-repro",
  posthogTraceId: `test-repro-${Date.now()}`,
});

// Abort mid-stream to trigger a chained error from the gateway
const controller = new AbortController();
setTimeout(() => controller.abort(), 2000);

try {
  await generateText({
    model: tracedModel,
    abortSignal: controller.signal,
    messages: [
      { role: "user", content: "Generate a detailed photo of a sunset over snowy mountains." },
    ],
  });
} catch (err) {
  console.log("PostHog stores:", JSON.stringify(err));
  console.log("Actual error:", {
    name: (err as Error).name,
    message: (err as Error).message,
    cause: (err as { cause?: unknown }).cause,
  });
}

await phClient.shutdown();
```

### Output observed

```
{
"statusCode":500,
"cause":{},
"name":"GatewayResponseError",
"type":"response_error",
"response":{},
"validationError":{
"name":"AI_TypeValidationError",
"cause":{
"name":"ZodError",
"message":"[\n {\n "expected": "object",\n "code": "invalid_type",\n "path": [\n "error"\n ],\n "message": "Invalid input: expected object, received undefined"\n }\n]"
},
"value":{}
}

```
message and the cause chain's deeper fields are missing from $ai_error, even though they exist on the actual thrown error.

## Related sub-libraries

- [x] @posthog/ai

Ashut0sh-mishra:
## Problem

`JSON.stringify` on a JavaScript `Error` instance returns `{}` because
`name`, `message`, `stack`, and `cause` are non-enumerable. The
`captureAiGeneration` path stored `JSON.stringify(options.error)` directly
on ``, so events captured via `withTracing` ended up with a
stripped-down blob — no message, no stack, no usable cause chain — making
`GatewayResponseError` and friends nearly impossible to debug from the
dashboard.

Closes #3556

## Fix

- New `serializeError` helper in `packages/ai/src/utils.ts` that
  normalizes `Error` instances to plain objects (`name` / `message` /
  `stack` / `cause` plus own enumerable properties such as `statusCode`,
  `response`, `validationError`), recurses through the cause chain and
  through nested plain objects, and guards against circular cause graphs
  with a `WeakSet`.
- `captureAiGeneration` now passes the error through `serializeError`
  before `JSON.stringify`, so `` keeps the same string shape
  (backwards compatible with anything that already reads it as JSON text)
  but carries the actual error data.

## Tests

Added to `packages/ai/tests/captureAiGeneration.test.ts`:

- A three-level cause chain `Error` is captured and the test asserts
  `message`, `stack`, `statusCode`, and the recursive `cause` levels
  are all present on the parsed `` JSON.
- A circular cause chain is captured and the test asserts the inner cycle
  is replaced with `'[Circular]'` rather than overflowing the stack.

The existing `error path` parametric tests are untouched and still
assert `` is a string, so this is backwards compatible.

## Reproducer (from the issue)

The original reproducer aborts mid-stream against `@ai-sdk/gateway`;
after this change the dashboard's `` payload contains the
`GatewayResponseError` `message`, `statusCode`, and the full
`validationError` -> `ZodError` cause chain instead of a blob like
`{ statusCode: 500, cause: {}, name: 'GatewayResponseError' }`.

### DIFF
--- a/packages/ai/src/captureAiGeneration.ts
+++ b/packages/ai/src/captureAiGeneration.ts
@@ -6,7 +6,7 @@ import { v4 as uuidv4 } from 'uuid'
 import { uuidv7, ErrorTracking as CoreErrorTracking } from '@posthog/core'
 import { version } from '../package.json'
 import type { TokenUsage } from './types'
-import { AIEvent, CostOverride, getTokensSource, sanitizeValues, withPrivacyMode } from './utils'
+import { AIEvent, CostOverride, getTokensSource, sanitizeValues, serializeError, withPrivacyMode } from './utils'
 
 type AnthropicTool = AnthropicOriginal.Tool
 
@@ -115,7 +115,12 @@ export const captureAiGeneration = async (client: PostHog, options: CaptureAiGen
 
     errorData = {
       $ai_is_error: true,
-      $ai_error: sanitizeValues(JSON.stringify(options.error)),
+      // JSON.stringify on an Error instance drops `message`, `stack`, and the
+      // `cause` chain because those fields are non-enumerable, which made
+      // `$ai_error` show up as an empty-ish blob on the dashboard. Normalize
+      // Errors to plain objects first so the fields survive serialization.
+      // See https://github.com/PostHog/posthog-js/issues/3556
+      $ai_error: sanitizeValues(JSON.stringify(serializeError(options.error))),
       $exception_event_id: exceptionId,
     }
   }
--- a/packages/ai/src/utils.ts
+++ b/packages/ai/src/utils.ts
@@ -591,6 +591,53 @@ export function sanitizeValues(obj: any): any {
   return jsonSafe
 }
 
+/**
+ * Convert an unknown thrown value into a JSON-serializable representation that
+ * preserves Error fields the engine marks non-enumerable (`name`, `message`,
+ * `stack`, `cause`), so they survive `JSON.stringify` for storage on
+ * `$ai_error`. Recurses into the `cause` chain and into any own enumerable
+ * properties (e.g. SDK-attached fields like `status`, `response`,
+ * `validationError`). Tracks a `seen` set to guard against circular cause
+ * graphs. Non-object inputs are returned as-is.
+ *
+ * Fixes https://github.com/PostHog/posthog-js/issues/3556
+ */
+export function serializeError(err: unknown, seen: WeakSet<object> = new WeakSet()): unknown {
+  if (err === null || (typeof err !== 'object' && typeof err !== 'function')) {
+    return err
+  }
+  if (seen.has(err as object)) {
+    return '[Circular]'
+  }
+  seen.add(err as object)
+  if (err instanceof Error) {
+    const obj: Record<string, unknown> = {
+      name: err.name,
+      message: err.message,
+      stack: err.stack,
+    }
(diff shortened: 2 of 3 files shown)


Ashut0sh-mishra:
Thanks @greptile-apps! Good point on the diamond-reference case. Documented the limitation in the JSDoc as intentional (keeps allocation O(visited); the linear cause-chain is the primary use case). Pushed.

Ashut0sh-mishra:
Thanks @greptile-apps! Good point on the diamond-reference case. Documented the limitation in the JSDoc as intentional (keeps allocation O(visited); the linear cause-chain is the primary use case). Pushed.

marandaneto:
same as https://github.com/PostHog/posthog-js/pull/3605 i think
@PostHog/team-llm-analytics decide which one to close?

packages/ai/src/captureAiGeneration.ts
```diff
@@ -115,7 +115,12 @@ export const captureAiGeneration = async (client: PostHog, options: CaptureAiGen
 
     errorData = {
       $ai_is_error: true,
-      $ai_error: sanitizeValues(JSON.stringify(options.error)),
+      // JSON.stringify on an Error instance drops `message`, `stack`, and the
+      // `cause` chain because those fields are non-enumerable, which made
+      // `$ai_error` show up as an empty-ish blob on the dashboard. Normalize
+      // Errors to plain objects first so the fields survive serialization.
+      // See https://github.com/PostHog/posthog-js/issues/3556
+      $ai_error: sanitizeValues(JSON.stringify(serializeError(options.error))),
```
hpouillot, line 123: Just dropping a comment in case it's useful, inside posthog core we have an error parser that creates an exception list from any object which is the format we use at error tracking. This will also pull the causes from the error to reconstruct the exception chain. You could potentially reuse some of our component as well to display the stacktrace on the UI if the format is the same.

carlos-marchal-ph:
Ended up merging #3605 in response to another ticket, but the implementation was pretty similar.

@Ashut0sh-mishra, thanks for your contribution! You should be able to get the new behaviour by updating to `7.19.0` or later.

@hpouillot, I wasn't aware of that error helper. Will probably switch the implementation to it for consistency at a later point.

PR #3585, opened 2026-05-12
fix(nextjs-config): warn when withPostHogConfig is not the outermost wrapper

### LINKED ISSUE #3572: @posthog/nextjs-config withPostHogConfig silently fails if it is not the outermost config wrapper
### Bug Description

If withPostHogConfig is not the last config wrapper in next.config.js, the PostHog build plugin does not do anything. It does not error, produce logs of any kind (even with log level set to debug), or upload source maps. 

Moving withPostHogConfig to be the last wrapper makes it work as expected. 

## How to reproduce

1. Wrap your next config with withPostHogConfig per the docs
2. Wrap that with something like withNextIntl from next-intl
3. Run your build. See that no source maps get generated or uploaded. With the PostHogConfig having log level set to debug, see there are no logs output from the posthog tool. 
4. Then put withPostHogConfig as the last wrapper and see that it works as expected. 

### Debug info

```shell
```

Ashut0sh-mishra:
## Problem

If `withPostHogConfig` is not the outermost wrapper in `next.config.js`
— e.g. `withNextIntl(withPostHogConfig(config, { ... }))` — the PostHog
build plugin silently never runs. No source maps upload, no errors, no
logs even with `logLevel: 'debug'`. Moving `withPostHogConfig` to be
the last wrapper makes it work as expected.

Root cause: `withPostHogConfig` returns an *async function* that Next.js
calls during build init. If a downstream wrapper consumes that function
and returns its own plain-object config without delegating, Next.js
never invokes our function, so the webpack/compiler hooks the plugin
adds never run.

Closes #3572

## Fix

Set an `invoked` flag inside the returned config function and schedule
a `.unref()`-ed 5-second `setTimeout` that prints a clear warning if
the flag is still `false`. When Next.js (or any properly delegating
outer wrapper) does call our function, the flag flips before the timer
fires and no warning is emitted.

The warning tells the user:

- what went wrong (the function was never invoked),
- the likely cause (outer wrapper swallowed it), and
- the fix (move `withPostHogConfig` to be the outermost wrapper).

It also links the tracking issue for context.

Notes:

- `.unref()` ensures the timer never holds the Node process open by
  itself.
- The change preserves the existing return type and runtime behaviour
  in the happy path (no extra work, just a flag flip on invocation).


Ashut0sh-mishra:
Thanks @greptile-apps! Addressed both:

- Replaced the multi-step `unref()` type-guard with a single optional-chaining call -- cleaner and same null-safety.
- Extracted the 5000ms literal into `INVOCATION_TIMEOUT_MS` and added a comment about the known false-positive when `next.config.js` is imported outside the Next.js build (e.g. test contexts). I chose to leave it as a documented limitation rather than add an opt-out flag, since the warning is informational and adding API surface for an edge case felt heavier than the benefit. Happy to revisit if you'd prefer the flag.

Pushed.

Ashut0sh-mishra:
Thanks @greptile-apps! Addressed both:

- Replaced the multi-step `unref()` type-guard with a single optional-chaining call -- cleaner and same null-safety.
- Extracted the 5000ms literal into `INVOCATION_TIMEOUT_MS` and added a comment about the known false-positive when `next.config.js` is imported outside the Next.js build (e.g. test contexts). I chose to leave it as a documented limitation rather than add an opt-out flag, since the warning is informational and adding API surface for an edge case felt heavier than the benefit. Happy to revisit if you'd prefer the flag.

Pushed.

Ashut0sh-mishra:
Thanks @greptile-apps -- fair point. Hoisted INVOCATION_TIMEOUT_MS to module scope so it sits with the other module-level declarations and isn't re-created per call. Pushed.

marandaneto:
cc @PostHog/team-error-tracking i think you'd know this better

hpouillot (approved):
Thanks for you contribution ! I will create a release as a follow-up

Ashut0sh-mishra:
@hpouillot Thanks a lot for the review and merge. I appreciate the follow-up release note too.

Issue #3564, opened 2026-05-11
Native CompressionStream produces invalid/partial gzip payloads with no thrown error in 1.372.10 (post #3378 + #3464)


hcaiano:
# Native CompressionStream produces invalid/partial gzip payloads with no thrown error in 1.372.10 (post #3378 + #3464)

## Summary

posthog-js 1.372.10 is producing invalid gzip request bodies on a non-trivial fraction of real-browser requests to our self-hosted capture (`POST /e/`), despite both PR #3378 (NotReadableError fallback) and PR #3464 (avoid `Blob.stream`) being present in the bundled SDK. The browser sends `Content-Encoding: gzip` (and `?compression=gzip` per `posthog-core.js`), capture-rs decodes, and logs one of:

- `decompress_gzip_to_bytes: failed to read GZIP chunk: invalid gzip header`
- `event: request payload parsing error: failed to decode request: invalid GZIP data`
- `High GZIP compression ratio detected - potential GZIP bomb` (less frequent)

The event is dropped. There is no thrown error in the browser that triggers the SDK's no-compression retry path (line `request.ts` ~equivalent of bundle line 85636: `Ns=!0; retry uncompressed`), so the bad payload is sent as-is.

## Reproduction environment

- Self-hosted PostHog (community/master capture, posthog-clickhouse), behind Cloudflare tunnel at `t.games.gg`.
- Client: posthog-js 1.372.10, initialized with default `compression` (effectively `GZipJS`, which the SDK swaps for native `CompressionStream("gzip")` when available).
- We bumped from 1.365.4 -> 1.372.10 specifically to pick up #3378. Errors continued at the same rate.
- Verified the bundled SDK contains the fix from #3464: zero references to `Blob.stream`/`.stream()` in `dist/main.js`; the gzip path writes UTF-8 to `CompressionStream.writable` directly.

## Production data (1 hour, real users)

| metric | value |
|---|---|
| Total successful events ingested (`posthog.events`) | 13,000 – 15,500/hr |
| Capture log lines: `invalid gzip header` | 183/hr |
| Capture log lines: `invalid GZIP data` (probably same requests, different error path) | 183/hr |
| Capture log lines: GZIP-bomb-ratio | ~5/hr |
| `params_lib_version` on 100% of failing requests | `1.372.10` |
| `compression` query param on 100% of failing requests | `gzip` |
| Origin | `https://<our-site>` |
| Path | `POST /e/` |
[...]

PR #3602, opened 2026-05-12
fix: validate gzip body before browser send

marandaneto:
## Problem

Some browser requests can be marked as gzip-compressed even when the final outgoing body is not gzip data, causing capture to reject and drop events for affected users.

## Changes

- Add shared gzip helpers in `@posthog/core` to detect gzip request intent and gzip magic bytes.
- Reuse the gzip magic-byte check inside native gzip validation.
- Add a browser send-boundary guard for fetch, XHR, and sendBeacon: if a request claims gzip but the encoded body does not start with the gzip magic bytes, disable native async gzip for the page lifetime, remove the compression query param, and send JSON instead.
- Add regression coverage for fetch, XHR, sendBeacon, and native async gzip fallback paths.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/safe-gzip-send-boundary.md
+++ b/.changeset/safe-gzip-send-boundary.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+---
+
+Validate gzip request bodies at the browser send boundary and fall back to JSON if the outgoing body is not gzip data.
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -25,6 +25,7 @@ const flushPromises = async () => {
     jest.useRealTimers()
 }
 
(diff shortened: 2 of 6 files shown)


marandaneto:
Relates to https://github.com/PostHog/posthog-js/issues/3564

turnipdabeets (approved):

PR #3605, opened 2026-05-13
fix: preserve error fields in $ai_error

### LINKED ISSUE #3556: @posthog/ai: $ai_error loses error fields (message, stack) and cause chain because json.stringify is used directly on error instances

carlos-marchal-ph:
## Problem

`$ai_error` was set with `JSON.stringify(error)`, which silently drops `message`, `stack`, and `cause`. That made it impossible to debug. 

## Changes

Uses a custom serialisation function to preserve these properties and capture nested errors through `cause`. Has a maximum nested error depth and a maximum stack line count to avoid bloat.

Closes #3556. 

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code (Opus 4.7). Iterated on the serializer shape: started with a richer version (AggregateError handling, separate WeakSet cycle detection, `[Circular]`/`[Truncated]` sentinels, plain-object container recursion) and trimmed it down to the current minimal version on Carlos's direction — depth-bound only, no cycle bookkeeping, plain objects pass through. Stack truncation at 20 lines was added afterwards to bound payload size on deep cause chains. `$ai_error` deliberately remains a JSON-encoded string rather than a structured object so existing `.toContain(...)` test assertions and any downstream HogQL consumers keep working.

### DIFF
--- a/.changeset/bright-crews-beg.md
+++ b/.changeset/bright-crews-beg.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': minor
+---
+
+Improved error serialization to capture cause, message and stack
--- a/packages/ai/src/langchain/callbacks.ts
+++ b/packages/ai/src/langchain/callbacks.ts
@@ -10,6 +10,7 @@ import type { DocumentInterface } from '@langchain/core/documents'
 import { ToolCall } from '@langchain/core/messages/tool'
 import { BaseMessage } from '@langchain/core/messages'
 import { sanitizeLangChain } from '../sanitization'
+import { serializeError } from '../serializeError'
 
 interface SpanMetadata {
   /** Name of the trace/span (e.g. chain name) */
@@ -376,7 +377,7 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       eventProperties['$process_person_profile'] = false
     }
     if (outputs instanceof Error) {
-      eventProperties['$ai_error'] = outputs.toString()
+      eventProperties['$ai_error'] = JSON.stringify(serializeError(outputs))
       eventProperties['$ai_is_error'] = true
     } else if (outputs !== undefined) {
       eventProperties['$ai_output_state'] = withPrivacyMode(this.client, this.privacyMode, outputs)
@@ -436,7 +437,7 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
 
     if (output instanceof Error) {
       eventProperties['$ai_http_status'] = (output as any).status || 500
-      eventProperties['$ai_error'] = output.toString()
+      eventProperties['$ai_error'] = JSON.stringify(serializeError(output))
       eventProperties['$ai_is_error'] = true
     } else {
       // Handle token usage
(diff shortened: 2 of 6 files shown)


shashwat-runable:
Any eta on when this pr will be merged ?

packages/ai/src/langchain/callbacks.ts
```diff
@@ -376,7 +377,7 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       eventProperties['$process_person_profile'] = false
     }
     if (outputs instanceof Error) {
-      eventProperties['$ai_error'] = outputs.toString()
+      eventProperties['$ai_error'] = JSON.stringify(serializeError(outputs))
```
Radu-Raicea: These JSON.stringify can throw

Radu-Raicea (approved):

.changeset/bright-crews-beg.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/ai': minor
```
marandaneto, line 2: should have been a patch instead?

PR #3607, opened 2026-05-13
feat(js-sdk): Enable $web_vitals reporting when cookieless mode is enabled

eli-r-ph:
## Problem
Right now reporting of `$web_vitals` is disabled when `cookieless_mode` is enabled, because cookieless mode disables the session ID manager, which gates vitals logic.

I think this is a safe change because the ingestion (backend) processing because:
* Web vitals events carry the cookieless mode property 
* A top-level `$session_id` is generated for cookieless-enabled events
* Cookieless enabled events that can't be hashed or generate a proper session ID are dropped

I think this is safe on the product (access path side) because:
* Access patterns I found in the product rely on page path not session ID
* Replay inspector relies on name, rating and value, not session ID
* Top level `$session_id` is available on cookieless web vitals events

Ad-hoc HogQL queries _could_ attempt to access a _nested_ session ID (_available on cookieless-disabled events only_) and this could break in the product. I'm not sure whether this is a blocker on shipping the change

## Verification

After getting this PR through CI, I build the local distribution and replaced the `posthog/frontend/` copy of the SDK in `dist` with my `posthog-js` feature branch `dist`. Next, I updated the local PostHog app config for the test team (cookieless mode enabled, web vitals enabled.)

Finally, I ran a Python script to generate and publish cookieless web vitals events along with some control events to the test account, then verified in ClickHouse that the results were well formed and the ingestion pipeline did not reject them.

## Changes
* Enable publishing of `$web_vitals` events even when cookieless mode is enabled (session ID manager not present)
* Related test changes

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/cookieless-web-vitals.md
+++ b/.changeset/cookieless-web-vitals.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/types': patch
+---
+
+Enable $web_vitals reporting when cookieless mode is enabled
--- a/packages/browser/playwright/mocked/cookieless-web-vitals.spec.ts
+++ b/packages/browser/playwright/mocked/cookieless-web-vitals.spec.ts
@@ -1,0 +1,69 @@
+import { expect, test } from './utils/posthog-playwright-test-base'
+import { start } from './utils/setup'
+import { pollUntilEventCaptured } from './utils/event-capture-utils'
+
+const startOptions = {
+    options: {
+        cookieless_mode: 'always' as const,
+        capture_performance: {
+            web_vitals: true,
+        },
+    },
+    flagsResponseOverrides: {
+        capturePerformance: true,
+    },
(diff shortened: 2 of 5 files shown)


lricoy (approved):

dustinbyrne (approved):
send it

robbie-c (approved):

PR #3616, opened 2026-05-14
feat: use batch ingestion for browser analytics requests

marandaneto:
## Problem

Browser analytics requests still used the legacy `/e/` ingestion shape and query-param gzip signaling for regular fetch/XHR calls, while core-based SDKs use `/batch/` with standard HTTP gzip headers. We want regular browser requests to align with core ingestion without losing reliable `sendBeacon` delivery during page unload.

## Changes

- Route non-`sendBeacon` browser analytics requests through `/batch/` using the core-style `{ api_key, batch, sent_at }` payload envelope.
- Use `Content-Type: application/json` and `Content-Encoding: gzip` for gzip-compressed non-beacon analytics requests.
- Preserve legacy `/e/` + `compression=gzip-js` behavior for `sendBeacon`, since beacon cannot set custom headers.
- Clone request options before rewriting them so retry queue unload can still fall back to the legacy beacon path.
- Update unit and functional tests, including the mock ingestion server, for `/batch/` and HTTP gzip handling.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


ioannisj:
How about this? https://posthog.slack.com/archives/C0ACMS2BNTU/p1778777375342569?thread_ts=1778769774.145479&cid=C0ACMS2BNTU

turnipdabeets (approved):
LGTM, greptile has a good point about the bool

marandaneto:
@PostHog/team-ingestion 

```
"analytics": {
    "endpoint": "/i/v0/e/"
},
```

remote config still returns `/i/v0/e/` so the SDK is gonna use it anyway, /batch is just the the 1st request until remote config returns.
what should we do here, ignore `analytics.endpoint`? or still honor this since its the only SDK using /e instead of /batch
should we change the value of `analytics.endpoint`?

if the SDK should honor `analytics.endpoint` and we cannot change its value, i see no point of merging this PR

packages/browser/src/extensions/conversations/external/components/RichContent.tsx
```diff
@@ -46,36 +46,36 @@ function sanitizeUrl(url: string): string | undefined {
 
     // Block dangerous protocols
     if (
-        normalizedForCheck.startsWith('javascript:') ||
-        normalizedForCheck.startsWith('vbscript:') ||
-        normalizedForCheck.startsWith('data:') ||
-        normalizedForCheck.startsWith('file:')
+        normalizedForCheck.indexOf('javascript:') === 0 ||
```
marandaneto, line 49: startsWith isnt IE11 compat so it was always failing in the E2E test, now we have a linter, so i replaced all calls with indexOf

eli-r-ph:
👋 Hey @marandaneto FYI all the "legacy" capture ingestion endpoints (`/e/`, `/batch/`, `/i/v0/e/`, `/capture/`, and what have you) all route to the same handling pipeline in the backend atm, and all of them can process either individual or batch payloads.

marandaneto:
> 👋 Hey @marandaneto FYI all the "legacy" capture ingestion endpoints (`/e/`, `/batch/`, `/i/v0/e/`, `/capture/`, and what have you) all route to the same handling pipeline in the backend atm, and all of them can process either individual or batch payloads.

thanks, so not really needed to do this then since they are routing to the same place, no need to risk some disruption for no benefits

Issue #3619, opened 2026-05-15
Feature request: dead clicks, per-element exclusion (parity with .ph-no-rageclick)


slshults:
## Is your feature request related to a problem?

Dead click detection has no  mechanism equivalent to `.ph-no-rageclick`. The only options currently are the global `capture_dead_clicks: false` kill switch, using `before_send` to drop `$dead_click` events (which requires inspecting event properties and is not element-aware), or adding `ph-no-capture` (which throws out too much baby with the bathwater, since it also suppresses autocapture events, session replay visibility, and heatmaps for that element).

For example, when a user clicks an `<a>` tag whose `href` points to a file download: The download fires correctly, but no DOM mutation occurs, so the dead click detector flags it. The PostHog customer may want to exclude that element from dead click detection while keeping autocapture, session replay, and heatmap data for it.

## Describe the solution you'd like

Two complementary improvements (either or both would help):

1. A `.ph-no-deadclick` CSS class (or equivalent config like `deadclick.css_selector_ignorelist`) to exclude specific elements from dead click detection without affecting autocapture, session replay, or heatmaps. (This would mirror the existing `.ph-no-rageclick` pattern.)

2. Smarter heuristics in the dead click detector itself: clicks on `<a>` elements with a `download` attribute, or with an `href` matching common file extensions (`.pdf`, `.zip`, `.csv`, `.xlsx`, etc.), shouldn't be flagged as dead clicks since the expected browser behavior (file download) doesn't produce a visible DOM change.
## Describe alternatives you've considered

- `ph-no-capture`: works but is too broad; it suppresses autocapture, replay, and heatmaps for the element.
- `before_send` filtering on `$dead_click` events: functional but requires custom code to inspect event properties and isn't truly element-level.
- Disabling `capture_dead_clicks` globally: defeats the purpose of the feature for the rest of the site.

## Additional context
From: https://posthoghelp.zendesk.com/agent/tickets/58193 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/59183)

pauldambra:
both released in 1.374.0

PR #3621, opened 2026-05-16
feat(dead-clicks): ignore clicks inside a link ancestor

pauldambra:
we ignore deadclicks on `<a/>`
because they might download a file or open in a new window which we can't observe
but we don't ignore deadclicks on `<a><span /></a>` if the span is the click target

we should 

## Summary

- Walks up from the dead-click target to find an `autocaptureCompatibleElement` ancestor (`a`, `button`, `form`, `input`, `select`, `textarea`, `label`). If one is found, the click is not flagged as a dead click.
- Subsumes the previous direct-element check — clicking the element itself is just the first iteration of the walk.
- Fixes the file-download false positive reported in #3619: a click on `<span>` / `<svg>` inside an `<a href="/file.pdf">` no longer fires `$dead_click`. Same for `<svg>`-inside-`<button>` and similar icon-button shapes.

This is a built-in heuristic that complements #3620's opt-out (`.ph-no-deadclick` / `css_selector_ignorelist`). The two land together to address both halves of #3619.

## Tradeoff

A genuine dead click on a non-interactive child of an interactive ancestor (e.g. a broken `onclick` on a `<div>` inside an `<a>`) is no longer surfaced. That pattern is rare; the noise-reduction win on icon buttons and download links is much larger and matches how autocapture itself attributes clicks.

## Test plan

- [x] `pnpm exec jest src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts` — 47 passed including new cases for every `autocaptureCompatibleElement` ancestor, a deeply-nested anchor child, and a non-interactive control case that still queues a candidate.
- [x] `pnpm lint`.

---

### DIFF
--- a/packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
+++ b/packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
@@ -1,7 +1,6 @@
 import { PostHog } from '../../posthog-core'
 import LazyLoadedDeadClicksAutocapture from '../../entrypoints/dead-clicks-autocapture'
 import { assignableWindow, document } from '../../utils/globals'
-import { autocaptureCompatibleElements } from '../../autocapture-utils'
 
 // need to fake the timer before jsdom inits
 jest.useFakeTimers()
@@ -147,16 +146,102 @@ describe('LazyLoadedDeadClicksAutocapture', () => {
             expect(lazyLoadedDeadClicksAutocapture['_clicks'].length).toBe(0)
         })
 
-        it.each(autocaptureCompatibleElements)('click on %s node is never a deadclick', (element) => {
-            const el = document.createElement(element)
-            document.body.append(el)
-            triggerMouseEvent(el, 'click')
+        it('click on an anchor is never a deadclick', () => {
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', '/some/file.pdf')
+            document.body.append(anchor)
+            triggerMouseEvent(anchor, 'click')
             jest.setSystemTime(4000)
 
             lazyLoadedDeadClicksAutocapture['_checkClicks']()
 
             expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
             expect(fakeInstance.capture).not.toHaveBeenCalled()
+        })
+
+        it('click on a child of an anchor is never a deadclick', () => {
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', '/some/file.pdf')
+            const child = document.createElement('span')
+            anchor.appendChild(child)
+            document.body.append(anchor)
+
+            triggerMouseEvent(child, 'click')
+
+            expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
+        })
+
+        it('click on a deeply nested child of an anchor is never a deadclick', () => {
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', '/some/file.pdf')
+            const wrapper = document.createElement('div')
+            const icon = document.createElement('svg')
+            wrapper.appendChild(icon)
+            anchor.appendChild(wrapper)
+            document.body.append(anchor)
+
+            triggerMouseEvent(icon, 'click')
+            jest.setSystemTime(4000)
+
+            lazyLoadedDeadClicksAutocapture['_checkClicks']()
+
+            expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
+            expect(fakeInstance.capture).not.toHaveBeenCalled()
+        })
+
+        it('click on a child of an anchor inside a shadow root is never a deadclick', () => {
+            const host = document.createElement('div')
+            const shadowRoot = host.attachShadow({ mode: 'open' })
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', '/some/file.pdf')
+            const child = document.createElement('span')
+            anchor.appendChild(child)
+            shadowRoot.appendChild(anchor)
+            document.body.append(host)
+
+            triggerMouseEvent(child, 'click')
+
+            expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
+        })
+
+        // buttons, inputs, selects, textareas, labels, forms all rely on app JS handlers
+        // (or browser-native side effects we can observe via mutation/scroll/selection).
+        // If the handler ran, our observers catch the effect; if it didn't, dead-click
+        // correctly surfaces the bug. A click on a broken <button> with no handler
+        // should still flag — that's exactly the case we want to catch.
+        it.each(['button', 'input', 'select', 'textarea', 'label', 'form'])(
+            'click on a %s is still a candidate',
+            (tag) => {
+                const el = document.createElement(tag)
+                document.body.append(el)
+
+                triggerMouseEvent(el, 'click')
+
+                expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(1)
+            }
+        )
+
+        it.each(['button', 'input', 'select', 'textarea', 'label', 'form'])(
+            'click on a child of a %s is still a candidate',
+            (ancestorTag) => {
+                const ancestor = document.createElement(ancestorTag)
+                const child = document.createElement('span')
+                ancestor.appendChild(child)
+                document.body.append(ancestor)
+
+                triggerMouseEvent(child, 'click')
+
+                expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(1)
+            }
+        )
+
+        it('click on a non-interactive element with no interactive ancestor is still a candidate', () => {
+            const div = document.createElement('div')
+            document.body.append(div)
+
+            triggerMouseEvent(div, 'click')
+
+            expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(1)
         })
 
         it.each(['ph-no-deadclick', 'ph-no-capture'])('ignores clicks on elements with the %s class', (className) => {
--- a/packages/browser/src/autocapture-utils.ts
+++ b/packages/browser/src/autocapture-utils.ts
@@ -278,6 +278,21 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false 
     return { parentIsUsefulElement, targetElementList }
 }
 
+// dead-click skips a click only when the click target is — or is inside — an <a>.
+// Anchors navigate / download / open a new window via the browser, and we have no
+// observable signal for those actions (no DOM mutation, no scroll). For every other
+// element (button, input, select, textarea, label, custom JS-handled divs, etc.) we
+// rely on the existing mutation / scroll / selection / visibility observers — if the
+// app's JS handler ran, those catch the effect; if it didn't, dead-click correctly
+// surfaces the bug. A click on a broken <button> with no handler should still flag.
+export function shouldSkipDeadClick(el: Element | null): boolean {
+    if (!window || cannotCheckForAutocapture(el)) {
+        return false
+    }
+    const { targetElementList } = getElementAndParentsForElement(el, false)
+    return targetElementList.some((node) => isTag(node, 'a'))
+}
+
 /*
  * Check whether a DOM event should be "captured" or if it may contain sensitive data
  * using a variety of heuristics.
--- a/packages/browser/src/entrypoints/dead-clicks-autocapture.ts
+++ b/packages/browser/src/entrypoints/dead-clicks-autocapture.ts
@@ -1,7 +1,7 @@
 import { assignableWindow, document, LazyLoadedDeadClicksAutocaptureInterface } from '../utils/globals'
 import { PostHog } from '../posthog-core'
 import { isNull, isNumber, isUndefined } from '@posthog/core'
-import { autocaptureCompatibleElements, getEventTarget, shouldCaptureDeadClick } from '../autocapture-utils'
+import { getEventTarget, shouldCaptureDeadClick, shouldSkipDeadClick } from '../autocapture-utils'
 import { DeadClickCandidate, DeadClicksAutoCaptureConfig, Properties } from '../types'
 import { autocapturePropertiesForElement } from '../autocapture'
 import { isElementInToolbar, isElementNode, isTag } from '../utils/element-utils'
@@ -195,11 +195,7 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
             return true
         }
 
-        if (
-            isTag(click.node, 'html') ||
-            !isElementNode(click.node) ||
-            autocaptureCompatibleElements.includes(click.node.tagName.toLowerCase())
-        ) {
+        if (isTag(click.node, 'html') || !isElementNode(click.node) || shouldSkipDeadClick(click.node)) {
             return true
         }
 
--- a/.changeset/dead-clicks-interactive-ancestor.md
+++ b/.changeset/dead-clicks-interactive-ancestor.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Dead clicks: a click on an `<a>` (or any element inside an `<a>`, including across shadow DOM) is no longer flagged as a dead click — the browser navigates / downloads / opens a new window and we can't observe that. Reuses autocapture's existing DOM walker for the ancestor walk. Direct clicks on `<button>`, `<input>`, `<select>`, `<textarea>`, `<label>`, and `<form>` (previously all skipped) are now eligible for dead-click detection: if their JS handler ran, the existing mutation / scroll / selection observers see the effect; if it didn't, dead-click correctly surfaces the bug. A broken `<button>` with no handler, or an `<svg>` icon inside one, will now flag — which is exactly the dead-click case we want to catch.


pauldambra:
* **#3621** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3621?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3621?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -28,6 +28,17 @@ function checkTimeout(value: number | undefined, thresholdMs: number) {
     return isNumber(value) && value >= thresholdMs
 }
 
+function hasAutocaptureCompatibleAncestor(node: Element): boolean {
```
pauldambra: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

`hasAutocaptureCompatibleAncestor` walks via `parentElement`, which returns `null` at a `ShadowRoot` boundary. A `<span>` inside a `<button>` inside an open shadow DOM will NOT be treated as having an interactive ancestor and will still be eligible as a dead click. The sibling code in `autocapture-utils.ts` (`getElementAndParentsForElement`) explicitly hops through shadow roots via `isShadowRoot` / `.host` — consider using the same pattern here for parity, or at minimum document the gap.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -28,6 +28,17 @@ function checkTimeout(value: number | undefined, thresholdMs: number) {
...
+    let cur: Element | null = node
+    while (cur && !isTag(cur, 'body') && !isTag(cur, 'html')) {
+        if (autocaptureCompatibleElements.includes(cur.tagName.toLowerCase())) {
```
pauldambra: > [!NOTE]

`autocaptureCompatibleElements` includes `'form'`. Pages that wrap a large region (or the whole page, as classic server-rendered forms do) in a single `<form>` will now have every click inside that region suppressed as having an "interactive ancestor", silently shrinking dead-click coverage. Suggestion: either exclude `'form'` from the ancestor check (forms are containers, not click targets in the same sense as `<a>`/`<button>`) or add a test that asserts the intended behavior for a click on a non-interactive `<div>` deeply nested inside a `<form>`.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Strictly suppressive behavior change shipped as a patch with no feature flag. Customers querying historical `$dead_click` event volume will see a step-change drop after this version rolls out via the CDN-loaded extension (which always serves latest, regardless of pinned core SDK version). The changeset documents it, which is good — consider also calling out in release notes that dashboards over `$dead_click` may show a drop.

pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

silly question... the changeset says this "matches how autocapture itself attributes clicks to an interactive ancestor", but autocapture's `getElementAndParentsForElement` also treats ancestors with `cursor: pointer` as useful (see `autocapture-utils.ts` around line 252). this walk only checks tag names. is that intentional? i can see arguments either way — tag-only is cheaper and avoids `getComputedStyle` on every click — but worth a sentence in the changeset or a comment so the next person doesn't "fix" the divergence. not blocking.

pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

i'm lazily asking... how will we know in production whether this filter is over-aggressive (eating real dead clicks on `<a>` wrappers that legitimately did nothing) vs. just-right? we already capture `$dead_click` events but we don't capture the clicks we *suppressed*. i wonder if a sampled debug log or a counter we surface somewhere would help us tune this. not blocking, ship as you see fit.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -28,6 +28,17 @@ function checkTimeout(value: number | undefined, thresholdMs: number) {
     return isNumber(value) && value >= thresholdMs
 }
 
+function hasAutocaptureCompatibleAncestor(node: Element): boolean {
+    let cur: Element | null = node
+    while (cur && !isTag(cur, 'body') && !isTag(cur, 'html')) {
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

The loop guard `!isTag(cur, 'body') && !isTag(cur, 'html')` is doing nothing — neither tag is in `autocaptureCompatibleElements`, and `parentElement` returns `null` at the document root anyway. Simplifying to `while (cur)` removes a tiny piece of "why is this here?" reading load and makes the rule (walk to the top, return true if any ancestor is interactive) easier to state in one sentence.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -190,11 +201,7 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
             return true
         }
 
-        if (
-            isTag(click.node, 'html') ||
-            !isElementNode(click.node) ||
-            autocaptureCompatibleElements.includes(click.node.tagName.toLowerCase())
-        ) {
+        if (isTag(click.node, 'html') || !isElementNode(click.node) || hasAutocaptureCompatibleAncestor(click.node)) {
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

The explicit `isTag(click.node, 'html')` short-circuit is now subsumed by `hasAutocaptureCompatibleAncestor` returning `false` at the root (also covered by the `!isElementNode(click.node)` check for non-element nodes). I'd lean toward dropping it (OnceAndOnlyOnce: the "what counts as ignorable" question lives in one helper now), but ThreeStrikesAndYouRefactor — happy to leave it.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

Pre-existing smell, surfaced rather than created by this PR — `autocapture-utils.ts` already walks ancestors checking `autocaptureCompatibleElements` (lines 240-258 and 346). Three call sites of the same "is there an interactive ancestor?" question is the threshold for extracting a shared helper into `autocapture-utils.ts`. Not for this PR (YAGNI cuts both ways and the new helper is six lines), but worth noting for the next time someone touches this.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

**[qa-team/performance]** 🟢 LOW

Per-click DOM walk to `<body>` with `tagName.toLowerCase()` and `Array.includes` on every ancestor. Click frequency and tree depth are both low so impact is negligible — but if a future change moves this check to a hotter path, consider memoising the lowercase tag set as a module-scope `Set<string>` to avoid the per-call linear scan.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -28,6 +28,17 @@ function checkTimeout(value: number | undefined, thresholdMs: number) {
     return isNumber(value) && value >= thresholdMs
 }
 
+function hasAutocaptureCompatibleAncestor(node: Element): boolean {
+    let cur: Element | null = node
```
pauldambra: > [!NOTE]

`cur` reads as shorthand and the rest of this file uses fuller names (`click`, `node`, `eventTarget`). Renaming to `node` or `ancestor` is a one-token change that helps the next reader.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

The function name says "Ancestor" but the loop starts at the node itself, so it also matches when the click target IS an interactive element. That's intentional (it preserves the previous behavior of the inlined check) but the name is slightly misleading. Either rename to `hasAutocaptureCompatibleAncestorOrSelf`, or initialize `cur` to `node.parentElement` and keep the precondition that the caller already handled the self-case.

packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
```diff
@@ -158,6 +158,51 @@ describe('LazyLoadedDeadClicksAutocapture', () => {
...
+                expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
+                expect(fakeInstance.capture).not.toHaveBeenCalled()
+            }
+        )
+
+        it('click on a deeply nested child of an anchor is never a deadclick', () => {
+            const anchor = document.createElement('a')
+            anchor.setAttribute('href', '/some/file.pdf')
+            const wrapper = document.createElement('div')
+            const icon = document.createElement('svg')
```
pauldambra, line 178: > [!NOTE]

**[qa-team/generalist-b]** ⚪ NIT

`document.createElement('svg')` creates an `HTMLUnknownElement` with `tagName === 'SVG'`, not a real `SVGElement` (that needs `createElementNS`). The test still exercises the right logic because `tagName.toLowerCase()` normalizes it, but if anyone reads the test as "we handle SVG namespaced icons" they'll be misled. Optional: switch to `document.createElementNS('http://www.w3.org/2000/svg', 'svg')` for fidelity.

packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
```diff
@@ -158,6 +158,51 @@ describe('LazyLoadedDeadClicksAutocapture', () => {
             expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
             expect(fakeInstance.capture).not.toHaveBeenCalled()
         })
+
+        it.each(autocaptureCompatibleElements)(
```
pauldambra: > [!NOTE]

**[paul]** ⚪ NIT

love the parameterised `it.each` here — perfetto. the deeply-nested anchor test below is a nice complement. one tiny thought: you've got a test for "non-interactive with no interactive ancestor *is* a candidate" which is great — could you also `it.each` the negative-direction for the case where the clicked element is itself interactive? line 150's test already covers it, but a direct assertion against the new helper would be reassuring. not blocking.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments for details, summary thread incoming.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (frontend, compatibility, performance, generalist-a, generalist-b), paul-reviewer, xp-reviewer

Tight, well-tested behaviour change. Two MEDIUM observations worth thinking about before merge (shadow DOM parity, `<form>` ancestor over-suppression); everything else is LOW / NIT polish.

### Key findings

🟡 **MEDIUM** — Shadow DOM gap: `parentElement` walk silently stops at `ShadowRoot` boundaries, unlike the sibling autocapture walker which hops via `.host`.

🟡 **MEDIUM** — `<form>` ancestor will suppress dead clicks anywhere inside a form-wrapped region. Worth deciding whether forms-as-containers should be in the same list as `<a>` / `<button>`.

🟢 **LOW** — Patch ships via the always-latest CDN extension with no flag; expect a step-change drop in `$dead_click` event volume on rollout. Document in release notes.

🟢 **LOW** — Walk diverges from autocapture which also treats `cursor: pointer` ancestors as useful. Tag-only is cheaper; worth a comment so the next person doesn't "fix" it.

🟢 **LOW** — No observability for *suppressed* clicks — we'll only know if this is over- or under-aggressive from user reports. Sampled counter could help tune.

🟢 **LOW** — `while (!isTag(cur, 'body') && !isTag(cur, 'html'))` guard is redundant; `while (cur)` covers it.

🟢 **LOW** — Call-site `isTag(click.node, 'html')` is now subsumed by the walk + the `!isElementNode` check.

🟢 **LOW** — Pre-existing smell: three call sites of "is there an interactive ancestor?" — extraction opportunity for a follow-up.

### Convergence

⚪ **NIT** — Both **paul** and **xp** independently flagged `cur` as terser than the surrounding file's naming style (`click`, `node`, `eventTarget`). High-confidence rename suggestion.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Monotonically suppressive change, safe to merge. Watch shadow DOM + `<form>` over-suppression. |
| 👤 paul | nice tight change, stamp from me. only nagging thought is the deliberate divergence from autocapture's `cursor: pointer` — worth a line of context. |
| 📐 xp | Solid, well-tested fix. Only quibbles are NIT-level — a redundant loop guard and `cur` being terser than the surrounding style. |

---

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Fixed in d004a444 — renamed `cur` -> `current`.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Fixed in d004a444 — simplified to `while (current)`.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Intentional — already called out in the changeset note. Release-notes treatment is downstream of the changeset, so this should propagate.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Leaving the `isTag(click.node, 'html')` guard in place — it's defensive at the call site and removing it makes the conditional harder to read at a glance. ThreeStrikesAndYouRefactor.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Out of scope for this PR (xp itself flagged it as "not for this PR"). Noted for the next time someone touches the autocapture-utils walks.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Out of scope — finding itself notes impact is negligible. Worth revisiting only if this check moves to a hotter path.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Intentional — keeping the self-case in the same helper preserves the previous direct-element behaviour without a second branch at the call site. The slight name imprecision is the cost. Reads naturally as "is this click inside something autocapture-compatible?".

packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
pauldambra: > [!NOTE]

Test exercises the right logic via `tagName` normalisation; switching to `createElementNS` would be more faithful but adds noise without changing what's under test. Leaving as-is.

packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-clicks-autocapture.test.ts
pauldambra: > [!NOTE]

The existing `it.each(autocaptureCompatibleElements)('click on %s node is never a deadclick', …)` at line 150 already asserts the direct-interactive case parametrically. The new `it.each` for children + the deeply-nested anchor test sit alongside it. No new coverage to add.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Confirmed with the author: not worth adding suppressed-click observability. Shipping as-is.

pauldambra: > [!NOTE]

Fixed in 2a6c0e6c — replaced the local walker with `clickIsInsideInteractiveElement` in `autocapture-utils.ts`, which delegates to `getElementAndParentsForElement` and inherits its shadow-DOM hop via `isShadowRoot` / `.host`. New test asserts a `<span>` inside a `<button>` inside an open shadow root is no longer a dead-click candidate.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Fixed in 2a6c0e6c — `<form>` is now excluded from the dead-click interactive set (containers shouldn't suppress dead-click coverage). Two new tests assert that a direct click on `<form>` and a click on a non-interactive child of `<form>` both remain dead-click candidates.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]

Fixed in 2a6c0e6c — by reusing `getElementAndParentsForElement`, dead-click now inherits autocapture's `cursor: pointer` ancestor attribution. No more divergence to explain. New test covers it.

packages/browser/src/autocapture-utils.ts
```diff
@@ -260,6 +260,27 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false
...
+// dead-click skips a click when:
+//   - the target IS a direct interactive element (a/button/input/select/textarea/label),
+//     except <form> which is a container, not a click target
+//   - OR the target is inside an <a>: the browser navigates/downloads/opens-new-window
+//     and we have no observable signal for those actions (no DOM mutation, no scroll)
+// other interactive ancestors (button, input, etc.) are NOT propagated to descendants —
+// those rely on app JS handlers, so the existing mutation/scroll/selection observers
+// catch the "handler did something" case, and correctly flag the "handler did nothing"
+// case as a dead click. <svg> inside a broken <button> should still surface.
+const DEAD_CLICK_INTERACTIVE_TAGS = autocaptureCompatibleElements.filter((tag) => tag !== 'form')
```
pauldambra: > [!NOTE]

`<label>` children are explicitly asserted to remain dead-click candidates, but the rationale ("those rely on app JS handlers") doesn't quite fit `<label>`. Clicking a `<span>` or icon inside a `<label for="…">` propagates to the label and toggles the bound input via the browser — no app handler involved, no guaranteed DOM mutation/scroll/selection signal we can observe. This is a plausible false-positive source for icon-in-label patterns (custom checkbox/radio styling, switch toggles).

Options:
1. Treat `<label>` like `<a>` and propagate the skip to descendants (one-line: add `'label'` to the anchor-walk check).
2. Accept the false-positive risk explicitly in the changeset/comment.

The previous broader rule covered this case; iteration 3 deliberately gives it up. Worth a conscious decision either way.

packages/browser/src/autocapture-utils.ts
```diff
@@ -260,6 +260,27 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false
...
+export function clickIsInsideInteractiveElement(el: Element | null): boolean {
```
pauldambra: > [!NOTE]

`clickIsInsideInteractiveElement` doesn't quite describe what it now does. It returns true for direct clicks on `a/button/input/select/textarea/label`, AND for any descendant of `<a>`, but NOT for descendants of the other interactive tags. A reader who reaches for it as a generic "am I inside something interactive?" helper will get wrong answers for `<button>` or `<input>` descendants.

Consider `shouldSkipDeadClick(el)` or `clickIsOnInteractiveOrInsideAnchor(el)` to make the asymmetric semantics visible at the call site. The comment block does explain it — this is a naming nit, not a correctness issue.

packages/browser/src/autocapture-utils.ts
```diff
@@ -260,6 +260,27 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false
...
+    if (!window || cannotCheckForAutocapture(el)) {
+        return false
+    }
+    if (DEAD_CLICK_INTERACTIVE_TAGS.indexOf(el.tagName.toLowerCase()) > -1) {
+        return true
+    }
+    const { targetElementList } = getElementAndParentsForElement(el, false)
```
pauldambra, line 292: > [!NOTE]

**[convergent: paul + qa-team/frontend]** ⚪ NIT

When `el` itself is an `<a>`, both branches catch it — the direct-tag check at line 277 short-circuits, but if it didn't, `targetElementList[0]` (which includes `el`) would also match `isTag(node, 'a')`. Tiny redundancy. Could simplify to a single `targetElementList.some` check that covers both cases, or just leave it — the early-return reads fine.

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/compatibility]** ⚪ NIT

`DEAD_CLICK_INTERACTIVE_TAGS` is derived from `autocaptureCompatibleElements` minus `'form'`. If a future change adds another tag to `autocaptureCompatibleElements` for autocapture's purposes, it silently propagates into dead-click's interactive-element check.

Intent is documented in the comment, but the only safety net is that comment. Options: inline the literal list `['a','button','input','select','textarea','label']` to decouple, or keep the coupling intentional (future autocapture-compatible additions get treated similarly here by default). Either is defensible — flagging for visibility.

packages/browser/src/autocapture-utils.ts
```diff
@@ -260,6 +260,27 @@ const getElementAndParentsForElement = (el: Element, captureOnAnyElement: false
...
+//     except <form> which is a container, not a click target
+//   - OR the target is inside an <a>: the browser navigates/downloads/opens-new-window
+//     and we have no observable signal for those actions (no DOM mutation, no scroll)
+// other interactive ancestors (button, input, etc.) are NOT propagated to descendants —
+// those rely on app JS handlers, so the existing mutation/scroll/selection observers
+// catch the "handler did something" case, and correctly flag the "handler did nothing"
+// case as a dead click. <svg> inside a broken <button> should still surface.
+const DEAD_CLICK_INTERACTIVE_TAGS = autocaptureCompatibleElements.filter((tag) => tag !== 'form')
+export function clickIsInsideInteractiveElement(el: Element | null): boolean {
+    if (!window || cannotCheckForAutocapture(el)) {
```
pauldambra, line 289: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

The early return `if (!window || cannotCheckForAutocapture(el)) return false` means "if we can't determine, treat as a candidate dead-click". That's the safe default — don't lose data — but the comment block above doesn't mention the failure mode. A one-liner inside the function would help a future maintainer reason about edge cases (text nodes, `<html>`, detached elements).

pauldambra (commented):
> [!NOTE]

Second-pass review on the anchor-only refinement. xp would ship without changes; paul + qa-team have one MEDIUM design question (label semantics) and a few NIT-level observations. See inline comments + summary.

pauldambra:
> [!NOTE]
>
> Multi-perspective re-review on the anchor-only refinement (5ffa1441): qa-team, paul-reviewer, xp-reviewer

The narrowing is materially safer than iteration 1 — the user's "broken `<button>` with `<svg>` should still surface" concern is now preserved. xp would ship without changes. One MEDIUM design question worth your call: should `<label>` propagate the skip to its descendants too? (See inline.)

### Key findings

🟡 **MEDIUM** — `<label>` semantics differ from button/input/select/textarea: clicking a child propagates to the bound input via the browser, no JS handler involved. Plausible false-positive source for icon-in-label custom controls.

🟢 **LOW** — Function name `clickIsInsideInteractiveElement` is asymmetric in meaning (direct vs descendant cases differ by tag). Comment explains it; rename for clarity at the call site is optional.

⚪ **NIT** — Convergent: derived-set coupling (`DEAD_CLICK_INTERACTIVE_TAGS` from `autocaptureCompatibleElements`).
⚪ **NIT** — Convergent: redundant double-check when `el` is `<a>`.
⚪ **NIT** — Document the `cannotCheckForAutocapture` failure-mode default (returns false = treat as candidate).

### Convergence

⚪ Both **paul** and **qa-team/compatibility** flagged the derived-set coupling — future tag additions silently propagate.
⚪ Both **paul** and **qa-team/frontend** noticed the redundant double-check when `el` is `<a>`.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Iteration 3 is materially safer. One real design question on `<label>` semantics; rest is naming/coupling nits. |
| 👤 paul | "perfetto. this is the narrowing i was hoping for last round." Ship as you see fit. |
| 📐 xp | No findings. Previous extract-helper smell is gone; new rule is strictly more conservative. Ship it. |

---

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

Renamed to `shouldSkipDeadClick` in 8fe19934 to make the asymmetric semantics obvious at the call site.

marandaneto (approved):

PR #3, opened 2020-03-29
Fea/batch event posts

Tmunayyer:
**Description**
This PR adds support for batching event requests from the front end. The previous default behavior was to send a POST request on each event. With batching enabled a queue will be built of recorded events. Every 3 seconds a POST request will be sent with the queued events.

I noticed no batching was occurring from the frontend to the server. I thought it might be useful to have a system in place to group some of the default event captures together to reduce bandwidth usage and cpu usage on client and server.

**Ancillary**
1. A new configuration setting adding to turn on and off request batching at the top level.
`DEFAULT_CONFIG: { 'request_batching': true }`

2. The poll process will track empty poll counts. After 5 empty polls, it will pause itself, resuming on the next event capture.

3. If a custom callback or custom options are passed through for specific events, default non-batching behavior will persist. The reason for this is if someone using the library passes a callback and relies on the result, they don't want to wait for the next poll.

PR #3623, opened 2026-05-17
fix(node): correct edge cases in local feature flag evaluation

richardsolomou:
## Problem

Six bugs in the posthog-node local feature flag evaluator at `packages/node/src/extensions/feature-flags/feature-flags.ts`, surfaced while porting this evaluator into `@posthog/convex` (#3622).

1. **Numeric comparison falls back to lexicographic.** In `matchProperty`, the `gt`/`gte`/`lt`/`lte` branch wraps `parseFloat` in a try/catch on the assumption that it throws on bad input — it doesnt, it returns `NaN`. `NaN != null` is true, so `NaN` slipped through into `compare(...)` and every numeric comparison silently returned false. Separately, when `overrideValue` was a string (the common case — person properties arriving as `\"10\"`), the code called `compare(overrideValue, String(value), operator)`, a lexicographic comparison: `\"10\" > \"9\"` returned false.
2. **`is_not_set` always inconclusive.** Same function. Key absent threw `InconclusiveMatchError` regardless of operator; key present plus `is_not_set` threw `\"Operator is_not_set is not supported\"`. Both cases are knowable locally — the operator is precisely about presence — so the flag was being forced to return `undefined` for no good reason.
3. **`isConditionMatch` ignored `prop.negation`.** `matchPropertyGroup` (the cohort path) already inverts matches when `prop.negation` is set; the flag-level condition loop did not, so any negated property on a flag-level filter quietly passed through.
4. **Inactive flag with continuity returned undefined instead of false.** `computeFlagValueLocally` checked `ensure_experience_continuity` before `!flag.active`, so a flag with `active: false` AND `ensure_experience_continuity: true` threw `InconclusiveMatchError` instead of short-circuiting to false.
5. **`is_set` against a null/undefined value returned false.** `NULL_VALUES_ALLOWED_OPERATORS = ['is_not']` excluded `is_set`, so the null guard fired and returned false before the switch could reach the `case 'is_set'` branch. `is_set` is about key presence, not value.
6. **Cohort flag-dep silently skipped.** `matchPropertyGroup`'s `else if (prop.type === 'flag')` branch logged a warning and called `continue` without setting `errorMatchingLocally = true`. For an AND group whose other props all pass, the loop fell through to `return propertyGroupType === 'AND'` → true, silently granting cohort membership the skipped flag dependency would have denied. Symmetric failure for OR groups.

## Changes

- `matchProperty`'s numeric branch parses both sides up front and gates on `Number.isFinite` — numeric comparison when both parse as finite numbers, lexicographic `String(...)` comparison otherwise.
- `matchProperty` returns `true` for `is_not_set` when the key is absent and `false` when it's present. Any other operator with the key absent still throws `InconclusiveMatchError`. A `null`-valued property still counts as set.
- `isConditionMatch` applies `if (prop.negation) matches = !matches` before the `if (!matches) return false` check, matching the cohort path.
- `computeFlagValueLocally` checks `!flag.active` before `ensure_experience_continuity` so an inactive flag short-circuits to false regardless of continuity.
- `NULL_VALUES_ALLOWED_OPERATORS` now includes `is_set`. The switch case correctly returns true on key presence regardless of value; the bug was purely that the null guard short-circuited it.
- `matchPropertyGroup` sets `errorMatchingLocally = true` before `continue` in the flag-type branch, so the eval routes through the existing `InconclusiveMatchError` throw and bubbles up to undefined rather than a wrong definitive answer.
- Updated three existing tests in `feature-flags.spec.ts` (`'with math operators'` on `property_e`, `'with operator is_set'` on `{ key: undefined }`, `'null or undefined property/override value'` on `is_set`) that locked in the buggy behavior.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

- \`pnpm jest --no-coverage --testPathPattern feature-flags\` — 313/313 pass across the four feature-flag test files.
- Added parameterised matrix in \`feature-flags.spec.ts\` covering \`gt\`/\`gte\`/\`lt\`/\`lte\` across string×number, number×string, number×number, and a non-numeric-string lexicographic fallback case.
- Added \`is_not_set\` cases for key absent (true), key present (false), and key present with \`null\` value (false — still counts as set). Confirmed other operators with the key absent still throw \`InconclusiveMatchError\`.
- Added \`is_set\` cases for property value \`null\` (true), \`undefined\` (true), normal value (true), and absent key (throws \`InconclusiveMatchError\`).
- Added five integration tests via \`PostHog.getFeatureFlag\`: flag-level negation, flag-level \`is_not_set\`, inactive flag with \`ensure_experience_continuity: true\` resolves to false, and a cohort referencing a flag dependency resolves to undefined.
- \`pnpm lint\` clean.

The two pre-existing \`exception-autocapture.spec.ts\` worker-thread failures reproduce on \`main\` — unrelated to this change.

---


marandaneto:
@PostHog/team-feature-flags i think it makes more sense that you guys review this, since it touches the flag evaluation algo

dmarticus:
@marandaneto already in my stack – I saw the notification

packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -586,6 +592,12 @@ class FeatureFlagsPoller {
           matches = matchProperty(prop, properties, warnFunction)
         }
 
+        // `matchPropertyGroup` (cohort path) inverts on `negation`; the top-level flag condition
+        // path needs to do the same or any negated property on a flag-level filter quietly passes.
+        if (prop.negation) {
+          matches = !matches
+        }
```
dmarticus: Pushback on this one: I don't think the top-level flag-condition path is supposed to honor `negation`. Confirmed across the stack:

- `posthog-python` `feature_flags.py:387-426` (`is_condition_match`) doesn't apply `prop.negation` at the flag-condition level — only inside `match_property_group` (line 747)
- Rust evaluator: same — `negation` only in `cohort_operations.rs`, never on flag release conditions
- Backend `feature_flag.py:329` explicitly refuses to flatten negated cohort properties into flag conditions: "Local evaluation doesn't support negation"

With this change, posthog-node will diverge from posthog-python, the Rust evaluator, and the server `/flags` endpoint on any payload carrying `negation: true` at the top-level filter. The backend shouldn't generate such payloads today, but if one ever appears only posthog-node will flip. That's a hard split to debug later.

Suggest dropping this fix from the PR, or opening a coordinated change against posthog-python + rust-evaluator alongside it so the semantics land everywhere at once. The other 5 fixes are independent and can ship now.

packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -1252,6 +1263,11 @@ function matchPropertyGroup(
                 `Skipping condition with dependency on flag '${prop.key || 'unknown'}'`
             )
           }
+          // Mark the group as inconclusive so we don't silently grant cohort membership in an AND
+          // group whose missing flag dependency would have evaluated to false (or deny it in an OR
+          // group whose flag dependency would have matched). Falls through to the
+          // InconclusiveMatchError throw at the end of the loop.
+          errorMatchingLocally = true
```
dmarticus: Correct fix, but conservative. Python's `match_property_group` actually evaluates flag deps inside cohort groups (`feature_flags.py:734-743`) via `evaluate_flag_dependency`. This class already has `evaluateFlagDependency` — the stronger fix is to call it here instead of marking inconclusive. This is strictly better than the prior silent grant though, so happy to leave the harder fix as a follow-up.

packages/node/src/extensions/feature-flags/feature-flags.ts
```diff
@@ -1075,28 +1092,22 @@ function matchProperty(
...
+      // is the right guard — `NaN != null` would slip through and produce nonsense comparisons
+      // like `NaN > 5`. Likewise, when a person property arrives as the string `"10"` we want
+      // `"10" > "9"` to evaluate numerically (true), not lexicographically (false).
+      const parsedValue = typeof value === 'number' ? value : parseFloat(String(value))
+      const parsedOverride =
+        typeof overrideValue === 'number'
+          ? overrideValue
+          : overrideValue != null
+            ? parseFloat(String(overrideValue))
+            : NaN
```
dmarticus: Style nit: the nested `typeof overrideValue === 'number' ? overrideValue : overrideValue != null ? parseFloat(…) : NaN` reads better as a 4-line `if/else`. Minor — feel free to ignore.

dmarticus (approved):
Strong PR overall — 5 of the 6 fixes are clean wins. One I'd push back on, and a couple of notes.

**The pushback: fix 3 (negation at flag level).**

This is the one is probably not a real bug. I just confirmed:

- `feature_flags.py:387-426` — `is_condition_match` doesn't apply `prop.negation` at the flag-condition level (only inside `match_property_group` at line 747)
- Rust evaluator: same — negation is honored only in `cohort_operations.rs`, never at the flag's release condition
- Backend (`feature_flag.py:329`) explicitly bails on flattening negated cohort properties to flag conditions — "Local evaluation doesn't support negation"

So with this PR, posthog-node will diverge from posthog-python, the Rust evaluator, and the server `/flags` endpoint on any payload that carries `negation: true` at the top-level filter. In practice the backend shouldn't generate such payloads, but if one ever shows up, only posthog-node will flip — every other SDK and the server give the opposite answer. That's a split that's painful to debug later.

I'd either drop fix 3 from this PR, or open a coordinated change against posthog-python + the rust evaluator alongside it so the semantics land everywhere at once. The other 5 fixes are independent and can ship now.

**Other notes:**

- Python has the same bugs as 2 and 4 (`feature_flags.py:454-455` and `client.py:1480-1483`). Worth filing a sibling PR — these are 1-line fixes and would prevent users on different SDKs from getting different undefined vs false answers for the same flag config.
- Behavior changes are real but correct — `is_set` on null flipping from false→true, lexicographic `"100" < "30"` flipping to numeric false. Both called out in the changeset, which is the right level. A release note in the next minor wouldn't hurt.
- Tests, comments, and changeset are all well done — the comments explaining "WHY" (NaN slip, key-presence semantics, AND/OR silent grant) are exactly the kind of comment that belongs. `Number.isFinite` in fix 1 is the right guard.

Approving — but please consider dropping or coordinating fix #3 before merging.

packages/node/src/extensions/feature-flags/feature-flags.ts
richardsolomou: Verified across the stack — `is_condition_match` in posthog-python doesn't apply `prop.negation` at the flag-condition level, and Rust matches. Dropped the inversion and the corresponding test in dacaec7a. A coordinated cross-stack change would be a separate PR if/when needed.

packages/node/src/extensions/feature-flags/feature-flags.ts
richardsolomou: Went ahead and did the stronger fix in 157c8b9b — `matchPropertyGroup` threads a `flagDependencyEvaluator` closure and calls `evaluateFlagDependency` for `type === 'flag'` props inside cohort groups, matching `match_property_group` at posthog-python `feature_flags.py:701-709`. When `dependency_chain` is absent or the dep can't be resolved, the existing `InconclusiveMatchError` catch still fires, so the conservative-undefined fallback is preserved. Added a happy-path test for the resolves-locally case.

packages/node/src/extensions/feature-flags/feature-flags.ts
richardsolomou: Flattened to a 4-line if/else in dacaec7a.

richardsolomou:
@dmarticus Verily, I giveth thee thanks, Dylan.

PR #3632, opened 2026-05-19
test(replay): verify network wrappers preserve CSRF and auth headers

pauldambra:
see https://www.reddit.com/r/posthog/comments/1thgz33/comment/omnhoyl/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

## Summary
- Reproduces a user report that PostHog network recording strips `x-csrf-token` (and similar) from the actual outgoing request.
- Adds 68 tests across jest (fetch + XHR, single- and double-wrap) and playwright (chromium, firefox, webkit) covering every plausible code path.
- **All 68 tests pass on current `main`** — the wrappers do NOT strip headers from the outgoing request.

This PR is intentionally test-only. The goal is to share a reproducible harness with the reporter so we can pinpoint where their setup differs (older posthog-js version, additional fetch wrapper, etc.).

## What gets tested

**`packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts`**
- Adds a `does not strip headers from the actual outgoing request` block parameterised over `x-csrf-token`, `x-csrftoken`, `x-xsrf-token`, `authorization`, `x-api-key`, `cache-control`, `pragma`, exercised through `init.headers` plain object, `Headers` instance, and `Request` input.
- Adds a `double wrap (network-plugin + tracing-headers)` block that applies the tracing-headers patch on top of the network-plugin patch (and vice versa) and asserts CSRF + tracing headers all reach the underlying `fetch`.

**`packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts`** (new)
- Mocks `XMLHttpRequest` so we can record what reaches the underlying `setRequestHeader` and `send`.
- Single-wrap coverage for the same seven headers, plus double-wrap of `setRequestHeader` (mirroring how a second wrapper would patch the instance after the recorder patches it).

**`packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts`** (new)
- Three scenarios: network recording only, tracing headers only, both together.
- Two methods: `fetch` and `XHR`.
- Three CSRF header names.
- Three browsers: chromium, firefox, webkit.
- Asserts `page.on('request')` sees the CSRF header on the real outgoing request.

## Why this matters
The reporter shared two curls (with and without PostHog recording active) where `x-csrf-token` was present in one and absent in the other. None of these tests reproduce that behaviour against current `main`, so the bug is most likely:
1. An older posthog-js version that pre-dates `ce4566d9` (`fix: restore fetch wrappers to most recent working version`), OR
2. Interaction with another `window.fetch` wrapper in their app (axios interceptors, Sentry, NewRelic, Datadog RUM, etc.).

These tests will catch a regression of (1) if it ever returns.

## Test plan
- [x] `pnpm exec jest src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts` — 63 pass, 1 skip
- [x] `pnpm exec jest src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts` — 14 pass
- [x] `pnpm exec playwright test playwright/mocked/session-recording/csrf-headers-preserved.spec.ts` — 54 pass across chromium, firefox, webkit


pauldambra (commented):
QA Swarm review complete — see inline comments and summary comment below.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+        {
+            options: scenario.options,
+            flagsResponseOverrides: scenario.flagsResponseOverrides,
+            url: '/playground/cypress/index.html',
+        },
+        page,
+        context
+    )
+
+    await page.waitForFunction(() => (window as any).posthog?.__loaded === true)
```
pauldambra, line 109: > [!NOTE]

**[qa-team/reliability]** 🟠 HIGH

For the "tracing headers only" and "double wrap" scenarios the test only waits for `posthog?.__loaded === true`. Tracing-headers is loaded as a lazy external dependency (`TracingHeaders._loadScript`) so the fetch can fire before `window.fetch` / `XMLHttpRequest.prototype.open` are patched. When that happens the test passes trivially because no wrapper ever ran — false negative. Match the existing `tracing-headers.spec.ts` pattern: also `await page.waitForFunction(() => (window as any).__PosthogExtensions__?.tracingHeadersPatchFns)`, and for scenarios that need network recording wrapping use `waitForSessionRecordingToStart` from `playwright/mocked/utils/setup.ts` so you know the open/fetch patches are actually installed before triggering the request.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+                    xhr.setRequestHeader('content-type', 'application/json')
+                    xhr.setRequestHeader(header, value)
+                    xhr.onloadend = () => resolve()
+                    xhr.send(JSON.stringify({ name: 'Untitled' }))
+                }),
+            { d: DOMAIN, header: csrfHeader, value: csrfValue }
+        )
+    }
+
+    await page.waitForTimeout(500)
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul + xp]** 🟡 MEDIUM

`await page.waitForTimeout(500)` is a sleep-based assertion. Since the assertion targets headers captured from `page.on('request', ...)` and `page.evaluate(...fetch...)` already awaits the round trip, prefer `await page.waitForRequest(req => req.url().startsWith('https://example.com'))` (then read its headers directly) — eliminates the magic timeout, removes a flakiness vector on slow CI / webkit, and removes the 9s mandatory wall-time penalty across the 18 parameterised tests per browser project.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+): Promise<Record<string, string>> {
+    let capturedHeaders: Record<string, string> = {}
+
+    page.on('request', (request: Request) => {
+        if (request.url().startsWith(`https://${DOMAIN}`)) {
+            capturedHeaders = request.headers()
+        }
+    })
+
+    await context.route(`**/${DOMAIN}/**`, (route) => {
```
pauldambra, line 86: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟢 LOW

Two `context.route(...)` registrations are stacked: `**/${DOMAIN}/**` (line 80) and `https://${DOMAIN}/` (line 83). The first glob already matches `https://example.com/`, so the second is unreachable in practice — dead code that confuses future readers. Drop the second route, or document why both are needed.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+    scenario: Scenario,
+    {
+        method,
+        csrfHeader,
+        csrfValue,
+    }: { method: 'fetch' | 'xhr'; csrfHeader: string; csrfValue: string }
+): Promise<Record<string, string>> {
+    let capturedHeaders: Record<string, string> = {}
+
+    page.on('request', (request: Request) => {
```
pauldambra: > [!NOTE]

`request.headers()` in Playwright lowercases header names, so the later `headers[header]` lookup only works because every entry in `csrfHeaders` is already lowercase. If someone later adds `X-CSRF-Token` (capitalised) to the list, the assertion silently returns `undefined` and fails with a misleading "received undefined" message. Either lowercase the lookup explicitly (`headers[header.toLowerCase()]`) or use `Object.entries(headers).find(([k]) => k.toLowerCase() === header.toLowerCase())`.

**[paul]** ⚪ NIT

`capturedHeaders` is overwritten on every matching request, so this is implicitly relying on only one example.com request firing. capture into an array and assert on the last/only one to be a touch sturdier. ship as you see fit.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+                }),
+            { d: DOMAIN, header: csrfHeader, value: csrfValue }
+        )
+    }
+
+    await page.waitForTimeout(500)
+    return capturedHeaders
+}
+
+test.describe('CSRF headers survive PostHog network wrappers', () => {
```
pauldambra, line 195: > [!NOTE]

3 scenarios × 2 methods × 3 headers = 18 tests, each spinning a fresh PostHog init and a fixed wait, repeated across chromium/firefox/webkit = 54 browser sessions. By far the most expensive of the new tests and will slow CI noticeably. Either capture headers for multiple CSRF headers in a single page session (register all three with one fetch), or restrict cross-browser fan-out via `test.skip` on webkit/firefox — CSRF stripping is a JS-level invariant, not browser-dependent.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
```diff
@@ -255,4 +255,181 @@ describe('fetch wrapper', () => {
...
+        })
+    })
+
+    // The real product applies BOTH the network-plugin wrapper and the
+    // tracing-headers wrapper to window.fetch (when __add_tracing_headers
+    // is configured). They wrap independently, so whichever loads second
+    // ends up calling the other. This block reproduces both orders and
+    // asserts that a user-supplied CSRF header still reaches the
+    // underlying fetch — and that the tracing headers are also added.
+    describe('double wrap (network-plugin + tracing-headers)', () => {
```
pauldambra, line 345: > [!NOTE]

**[convergent: qa-team/generalist-b + xp + paul]** 🟠 HIGH

The "double wrap" suite inlines a copy of `tracing-headers.ts:patchFetch` rather than importing the real wrapper. The inline version has already drifted from production: no `COOKIELESS_SENTINEL_VALUE` handling, no `addTracingHeaders` early-return on invalid URLs, no `SessionIdManager`. So this test asserts "a wrapper shaped like patchFetch composes correctly", not "patchFetch composes correctly" — and can't catch the very regression it's supposed to guard.

The c2 fix: extract a pure `wrapFetchWithTracingHeaders(originalFetch, hostnames, distinctId, sessionManager)` from `tracing-headers.ts` and call it from both the patcher and this test. Smell becomes a refactor — and removes the missing seam that forced the inline copy.

(paul also flagged a related concern: "i wonder if it's worth also asserting the *other half* of the invariant — that the recording payload still has the csrf header REDACTED. right now we've proved 'the outgoing request keeps the header' but nothing pins 'the recording redacts it'. could be a one-liner extra assertion on the captured `cb` data.")

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
```diff
@@ -255,4 +255,181 @@ describe('fetch wrapper', () => {
         expect(capturedInit?.method).toBe('PUT')
         expect(capturedInit?.headers).toEqual({ 'X-Custom': 'value' })
     })
+
+    // Reproduces a user report that the wrapper strips CSRF headers from the
+    // actual outgoing request. The wrapper redacts them from the *recording*
+    // via HEADER_DENY_LIST, but that must never leak back into the live
+    // request the browser sends to the server.
+    describe('does not strip headers from the actual outgoing request', () => {
+        const headerCases = [
```
pauldambra: > [!NOTE]

**[xp]** 🟡 MEDIUM

The `headerCases` / `csrfHeaderCases` arrays are duplicated across three files (this one, `xhr-wrapper-invariants.test.ts`, and `csrf-headers-preserved.spec.ts`). OnceAndOnlyOnce: when the next reported sensitive header gets added to the production `HEADER_DENY_LIST`, the invariant tests will silently fail to cover it because nobody will remember to update three lists. Extract to a small test fixture module (e.g. `__tests__/extensions/replay/external/header-cases.ts`) exporting `csrfHeaderCases` and `sensitiveHeaderCases`, and import from all three sites.

`cache-control` and `pragma` are not on `HEADER_DENY_LIST` (config.ts line 61), so they are guaranteed to pass for trivial reasons — the wrapper never even considered redacting them. Including them in the "does not strip" suite implies they were ever at risk, which is misleading. Move them to a separate "ordinary headers also pass through" describe block so the deny-listed-but-not-leaked headers stand out as the load-bearing invariant.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
```diff
@@ -255,4 +255,181 @@ describe('fetch wrapper', () => {
...
+            ['x-csrf-token', 'r_lIDFH3NdoomvNNKK5SWHg3KFOpWvnARWDvvi_TbwY'],
+            ['x-csrftoken', 'django-style-csrf'],
+            ['x-xsrf-token', 'angular-style-xsrf'],
+            ['authorization', 'Bearer abc123'],
+            ['x-api-key', 'sk-test-1234'],
+            ['cache-control', 'no-cache'],
+            ['pragma', 'no-cache'],
+        ] as const
+
+        it.each(headerCases)('preserves %s on downstream Request when set via init.headers plain object', async (name, value) => {
```
pauldambra: > [!NOTE]

**[qa-team/generalist-a]** ⚪ NIT

The three parameterised describe blocks share identical structure varying only in how headers are supplied (plain object / Headers instance / Request input). Consider collapsing into a single `it.each` over the cartesian product `header_cases × header_supply_mechanisms` — keeps OnceAndOnlyOnce, and the diff already adds ~180 lines of repetition.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
```diff
@@ -0,0 +1,187 @@
...
+            ['x-csrf-token', 'r_lIDFH3NdoomvNNKK5SWHg3KFOpWvnARWDvvi_TbwY'],
+            ['x-csrftoken', 'django-style-csrf'],
+            ['x-xsrf-token', 'angular-style-xsrf'],
+            ['authorization', 'Bearer abc123'],
+            ['x-api-key', 'sk-test-1234'],
+            ['cache-control', 'no-cache'],
+            ['pragma', 'no-cache'],
+        ] as const
+
+        it.each(headerCases)('forwards %s to the underlying XMLHttpRequest.setRequestHeader', (name, value) => {
```
pauldambra: > [!NOTE]

**[qa-team/generalist-b]** 🟠 HIGH

This test would silently pass even if the network-plugin wrapper never installed — the mock class's own `setRequestHeader` unconditionally pushes to `setHeaderCalls`, so the assertion holds in both "wrapped" and "completely unwrapped" worlds. The only thing this test actually proves is that the wrapper doesn't short-circuit on deny-listed names (it always calls `originalSetRequestHeader`).

To make the assertion meaningful, also assert that the wrapper recorded the header into the captured network request payload (the `cb` callback passed to `plugin.observer`) — currently the test passes `() => {}` as `cb` so the recording side is entirely unobserved. Capture the cb invocations and assert e.g. that for deny-listed headers the recording shows `REDACTED` while `setHeaderCalls` shows the original value — that is the true invariant of the bug being investigated.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
```diff
@@ -0,0 +1,187 @@
...
+            const listeners = this.listeners.get(event)
+            if (listeners) {
+                const index = listeners.indexOf(listener)
+                if (index > -1) listeners.splice(index, 1)
+            }
+        }
+    }
+}
+
+function setupWrappedXhr() {
```
pauldambra, line 74: > [!NOTE]

`getRecordNetworkPlugin` uses a module-level `initialisedHandler` singleton in `network-plugin.ts:639`. Each `setupWrappedXhr()` call relies on the previous test's `cleanup()` setting it back to `null`, otherwise the second test silently gets a no-op cleanup and the mock XHR's `setRequestHeader` is the unwrapped passthrough — which then trivially "passes" the assertion. If any test ever throws between `setupWrappedXhr()` and `cleanup()`, all subsequent tests in the file are compromised. Reset the singleton in an `afterEach` (e.g. expose a test-only reset, or capture `cleanup` in a `let` and always call it in `afterEach`).

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
```diff
@@ -0,0 +1,187 @@
...
+
+            expect(setHeaderCalls).toEqual(
+                expect.arrayContaining([{ header: name, value }])
+            )
+        })
+    })
+
+    describe('does not modify what the underlying send() receives', () => {
+        it('forwards the original body to underlying send', () => {
+            const { MockXMLHttpRequest, sendCalls, cleanup } = setupWrappedXhr()
```
pauldambra: > [!NOTE]

The body-forwarding test makes one assertion `expect(sendCalls).toEqual([body])` and doesn't parameterise over `Document` / `FormData` / `Blob` / `Uint8Array` / `null` bodies the way `fetch-wrapper-invariants.test.ts` does. Either parameterise to match the fetch coverage, or rely on `network-plugin.test.ts`'s existing body coverage and remove this block — as-is it tests one trivial case that is unlikely to ever regress in isolation.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
```diff
@@ -0,0 +1,187 @@
...
+        })
+    })
+
+    // Both the network-plugin wrapper and the tracing-headers wrapper patch
+    // XMLHttpRequest.prototype.open. In the product they're both applied
+    // when __add_tracing_headers is configured alongside session recording.
+    // Each open patch ALSO patches setRequestHeader on the instance, so
+    // setRequestHeader can end up double-wrapped. The user-supplied header
+    // must still reach the underlying XHR.
+    describe('double wrap (network-plugin + tracing-headers-style setRequestHeader patch)', () => {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-b + paul]** 🟠 HIGH

The "double wrap" XHR suite emulates the two wrappers by manually replacing `xhr.setRequestHeader` on an instance — but the production tracing-headers wrapper for XHR patches `XMLHttpRequest.prototype.open` (see `entrypoints/tracing-headers.ts:patchXHR`) and does NOT touch `setRequestHeader` at all. So this test verifies a wrapping topology that doesn't exist in production.

The real composition risk is two `open` patches stacking (network-plugin's open patch wraps the prototype, tracing-headers's open patch then wraps that wrapped one). Restructure the test to patch `MockXMLHttpRequest.prototype.open` twice (once via `plugin.observer`, once via a copy of the real tracing-headers patch) and assert that headers and tracing headers both reach `setHeaderCalls`. As written, this section will not catch a regression in the real double-wrap path — and the comment at lines 122-127 will mislead the next maintainer.

**[xp]** ⚪ NIT (when reworked, add a debugging breadcrumb)

The "double wrap" block is the strongest part of the diff once it's pointed at the real topology. Add a one-line preamble comment naming what breaks if this fails ("a wrapper is replacing `setRequestHeader`/`open` without delegating to the previous one") so the 2027 reader gets a debugging breadcrumb, not just a red assertion.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+import { start } from '../utils/setup'
+import { Page, BrowserContext, Request } from '@playwright/test'
+
+// Reproduces the user report that PostHog network recording strips CSRF
+// headers from the actual outgoing request. The fix invariant: enabling
+// session recording network capture (and/or tracing-headers, both of which
+// wrap fetch/XHR) must NEVER cause user-supplied request headers to go
+// missing from what the browser actually sends to the server.
+
+const DOMAIN = 'example.com'
```
pauldambra, line 23: > [!NOTE]

**[qa-team/compatibility]** ⚪ NIT

Hostnames in `__add_tracing_headers` are matched by exact-hostname equality (see `addTracingHeaders` in `entrypoints/tracing-headers.ts:22`). Using `example.com` is fine, but worth a comment that the URL path is irrelevant — the future maintainer hitting a flake here will probably first suspect the URL.

pauldambra:
> [!NOTE]
>

Four HIGH findings — the tests have structural validity problems that need fixing before they can be trusted as regression guards for the CSRF-stripping bug they exist to catch. The intent and shape are great; the gaps are about wiring the assertions to the *real* code path.

### Key findings (HIGH)

- 🟠 **Playwright readiness wait is insufficient** — `csrf-headers-preserved.spec.ts:97` only waits for `posthog.__loaded`; tracing-headers is a *lazy* dependency, so the fetch can fire before `window.fetch` is patched. Scenarios 2 and 3 may pass trivially because no wrapper ever ran.
- 🟠 **Fetch double-wrap test inlines its own tracing-headers wrapper** — `fetch-wrapper-invariants.test.ts:338` reimplements a subset of `patchFetch` (missing `COOKIELESS_SENTINEL_VALUE`, missing hostname guard) rather than importing the real one. Can't catch regressions in the real wrapper.
- 🟠 **XHR single-wrap test passes when wrapper never installs** — `xhr-wrapper-invariants.test.ts:91`. The mock's own `setRequestHeader` unconditionally records into the sink the test asserts on. Combined with the `network-plugin.ts` `initialisedHandler` singleton, a leaked cleanup makes every subsequent test green for the wrong reason.
- 🟠 **XHR double-wrap test verifies a wrapping topology that doesn't exist in production** — `xhr-wrapper-invariants.test.ts:128`. The real `patchXHR` (in `entrypoints/tracing-headers.ts:54`) patches `prototype.open`, never `setRequestHeader`. The test as written can't catch the real composition bug.

### Convergence

Three or more reviewers independently flagged:
- **`waitForTimeout(500)` is sleep-based** — should be `waitForRequest` (qa-team/reliability, paul, xp)
- **Inlined `applyTracingHeadersWrapper` drifts from production** — extract a pure `wrapFetchWithTracingHeaders` from `tracing-headers.ts` and call it from both sites (qa-team/generalist-b, xp, paul)

Two reviewers independently flagged:
- **Redundant stacked `context.route` registrations** in the playwright spec (qa-team/reliability, paul)
- **Misleading "double wrap" framing in the XHR test** — tracing-headers doesn't actually wrap `setRequestHeader` (qa-team/generalist-b, paul)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Reasonable invariant guards but two structural validity problems: tests can pass without the wrapper ever installing, and the "double wrap" suites emulate rather than import the real wrappers. Strongest layer is the playwright spec, but it doesn't wait for the patches to be installed. |
| 👤 paul | Solid bit of test infrastructure for a genuinely important bug class. Main asks: fix the misleading XHR `setRequestHeader` framing, assert the recording-side redaction too (otherwise CSRF could quietly leak into recordings), drop the `waitForTimeout`. Approval-with-caveats. |
| 📐 xp | Good intent, good shape. Header-case lists duplicated across three files — OnceAndOnlyOnce, extract to a fixture module so the next reported header doesn't need three updates. Inlined tracing-headers wrapper is an honest workaround that points at a real missing seam in production code. |

### Suggested order of operations

1. Tighten the readiness waits in the playwright spec (wait for `tracingHeadersPatchFns` and session-recording start) and replace `waitForTimeout(500)` with `waitForRequest`.
2. Extract `wrapFetchWithTracingHeaders` from `tracing-headers.ts` and call it from both the production patcher and the fetch double-wrap test. Drop the inline copy.
3. Restructure the XHR double-wrap to patch `prototype.open` twice (not `setRequestHeader`) — the real topology — and the single-wrap to also observe the `cb` callback so the assertion depends on the wrapper running.
4. Extract `csrfHeaderCases` / `sensitiveHeaderCases` into a shared test fixture module.
5. Move `cache-control`/`pragma` to a separate "ordinary headers pass through" block so the deny-listed-but-not-leaked headers stand out.
6. Add an `afterEach` that always calls `cleanup()` to defend against the `initialisedHandler` singleton.

---

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — `captureRequestHeaders` now waits for `tracingHeadersPatchFns` when `__add_tracing_headers` is set, and `waitForSessionRecordingToStart` when network capture is enabled. The fetch/XHR is no longer fired before the wrappers are installed.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — replaced `waitForTimeout(500)` with `page.waitForRequest(req => req.url().startsWith(...) && req.method() === 'POST')` and read headers from the request object directly. Suite is ~3s faster per browser project.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — dropped the second `context.route('https://example.com/')`. The `**/example.com/**` glob already covers it.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — assertion now does a case-insensitive lookup via `Object.entries(...).find(...)`. The 'capture into array' nit isn't applied because the new `waitForRequest` flow reads from the single matched request object, so the multiple-request race no longer exists.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — extracted `csrfHeaderCases`, `sensitiveHeaderCases`, and `unaffectedHeaderCases` into a shared `header-cases.ts` fixture imported by both jest test files and the playwright spec. Also split the 'does not strip' describe into 'deny-listed' and 'ordinary' blocks so the load-bearing invariant stands out from the trivially-passing `cache-control`/`pragma` cases.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — added an `afterEach` cleanup safety net via a `setupWrappedXhrWithSafety` wrapper. Cleanup is now idempotent and registered for the safety net, so a test that throws before its own `cleanup()` no longer leaks the singleton.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 219ec73c — added a comment near the `DOMAIN` constant noting that `__add_tracing_headers` matches by exact hostname and the URL path is irrelevant.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+
+    page.on('request', (request: Request) => {
+        if (request.url().startsWith(`https://${DOMAIN}`)) {
+            capturedHeaders = request.headers()
+        }
+    })
+
+    await context.route(`**/${DOMAIN}/**`, (route) => {
+        void route.fulfill({ status: 200, contentType: 'text/plain', body: 'ok' })
+    })
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 4bcf9049 — the mocked route now returns `Access-Control-Allow-Origin/Methods/Headers` so the preflight succeeds, and the `waitForRequest` filter now requires `method() === 'POST'` so we assert on the real outgoing request rather than the OPTIONS preflight. Good catch — the suite happened to pass without this but would have been brittle to any browser tightening preflight behaviour.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 219ec73c — same fix as the QA Swarm convergent finding on this line. Now waits for `tracingHeadersPatchFns` (when tracing headers are configured) and `sessionRecording.started === true` (when network capture is enabled) before triggering the request.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+                    xhr.setRequestHeader(header, value)
+                    xhr.onloadend = () => resolve()
+                    xhr.send(JSON.stringify({ name: 'Untitled' }))
+                }),
+            { d: DOMAIN, header: csrfHeader, value: csrfValue }
+        )
+    }
+
+    await page.waitForTimeout(500)
+    return capturedHeaders
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 219ec73c — replaced `waitForTimeout(500)` with `page.waitForRequest` filtering on `method === 'POST'`.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
```diff
@@ -0,0 +1,156 @@
...
+            capturedHeaders = request.headers()
+        }
+    })
+
+    await context.route(`**/${DOMAIN}/**`, (route) => {
+        void route.fulfill({ status: 200, contentType: 'text/plain', body: 'ok' })
+    })
+    await context.route(`https://${DOMAIN}/`, (route) => {
+        void route.fulfill({ status: 200, contentType: 'text/plain', body: 'ok' })
+    })
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 219ec73c — dropped the redundant root-path `context.route`.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Deferring to author — restricting cross-browser fan-out is a coverage-policy decision. The other suggestion (capture all three CSRF headers in one page session) would reduce 18 tests to 6 but couples the per-header assertion failures together. Leaving open for human judgement.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Deferring to author — the cleanest fix is to extract a pure `wrapFetchWithTracingHeaders(originalFetch, hostnames, distinctId, sessionManager)` from `tracing-headers.ts` and call it from both the production patcher and this test. That changes the scope of this PR from test-only to test + small production refactor — needs human judgement on whether to bundle or split. Greptile's H0G finding makes the same point.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Deferring to author — collapsing into a single `it.each` over a cartesian product is a stylistic preference and now interacts with the new `header-cases` fixture extraction. Leaving for the author to decide whether the consolidation is worth the loss of describe-block grouping in the test runner output.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Deferring to author — the suggested fix (observe the `cb` callback so the assertion depends on the wrapper having run) is architecturally significant. Asserting on the recording-side payload too is a real change of test strategy. Worth doing but needs human signoff.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Deferring to author — parameterising over body types (`FormData`, `Blob`, `Uint8Array`, `null`) duplicates coverage already in `network-plugin.test.ts`. The author should decide whether to expand here or delete the block in favour of the existing coverage.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Deferring to author — making the XHR double-wrap match the real production topology (both wrappers patch `prototype.open`, not `setRequestHeader`) is a structural rework. The current test does exercise a real composition risk (instance-level `setRequestHeader` patching, which axios-style libraries sometimes do internally), but it's not the topology the comment claims. Author needs to decide: rewrite to match production `patchXHR`, or relabel.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
```diff
@@ -255,4 +255,181 @@ describe('fetch wrapper', () => {
...
+                    body: '{}',
+                })
+                cleanup()
+
+                expect(downstream!.headers.get(name)).toBe(value)
+                expect(downstream!.headers.get('x-posthog-distinct-id')).toBe('distinct-abc')
+                expect(downstream!.headers.get('x-posthog-session-id')).toBe('session-abc')
+                expect(downstream!.headers.get('x-posthog-window-id')).toBe('window-abc')
+            })
+        })
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Deferring to author — same call as the QA Swarm convergent finding on `fetch-wrapper-invariants.test.ts:352`. Cleanest fix is to extract a pure `wrapFetchWithTracingHeaders` from `tracing-headers.ts` and call it from both sites — needs human judgement on whether to bundle that small production refactor here or split it out.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in f886a3a6 — setupWrappedXhr now records cb invocations, the mock's `performance.getEntriesByName` returns a synthetic resource-timing entry so the perf-entry retry loop resolves immediately, and a `triggerDoneAndFlush` helper drives the XHR to DONE and waits for the recording cb to fire. The single-wrap test now asserts BOTH the mock-level sink (outgoing path) AND the cb payload's `requestHeaders` (recording path). The recording assertion only holds when the wrapper actually ran — so a regression that fails to install can no longer pass trivially. Triggered by the reporter confirming they use axios 0.18 XHR adapter, making this code path the one that matters.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in f886a3a6 — replaced the instance-level `setRequestHeader` patching (which doesn't match production) with prototype.open stacking that mirrors `entrypoints/tracing-headers.ts:patchXHR`. The double-wrap suite now stacks two prototype.open patches (network-plugin + tracing-headers-style), verifies both wrappers ran, and asserts the CSRF header reaches the underlying setRequestHeader through both layers — i.e. the real production topology when `__add_tracing_headers` is configured alongside session recording. Triggered by the reporter confirming axios 0.18 + XHR usage.

packages/browser/playwright/mocked/session-recording/csrf-headers-preserved.spec.ts
pauldambra: > [!NOTE]

Addressed in 094e7c07 — author chose option (b): kept cross-browser fan-out but consolidated the 3 csrf headers into a single probe request per scenario+method. 54 → 18 tests across chromium/firefox/webkit, runtime down from 10.7s to 5.8s. Each test now iterates over the captured headers map and asserts each csrf header is unchanged, so a regression on any single header still fails this test (just with a more informative message thanks to the `message` arg on `expect`).

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in 094e7c07 — author chose option (c): kept the inlined `applyTracingHeadersWrapper` but added a comment documenting that the playwright spec is the authoritative end-to-end coverage for the real composed wrapper behaviour. The jest test now only claims to prove the structural invariant that two `new Request(url, init)`-then-`originalFetch(req)`-style wrappers compose without dropping headers, irrespective of what each one adds. If someone later wants to extract `wrapFetchWithTracingHeaders` from `tracing-headers.ts` that's still the right move but is out of scope for this PR.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in 094e7c07 — same call as the QA Swarm convergent finding on this line: author chose to keep the inlined helper but make its scope explicit. Added a comment noting that the playwright spec end-to-end coverage is authoritative for the real composed wrapper, and the jest test only asserts the structural compose-without-stripping invariant. Extracting `wrapFetchWithTracingHeaders` from `tracing-headers.ts` remains a viable follow-up but is intentionally out of scope here.

packages/browser/src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in 094e7c07 — collapsed the three describe blocks into a single `it.each` over the cartesian product `sensitiveHeaderCases × [init.headers plain object, Headers instance, Request input]`. Each supply mechanism is now a `{ label, invoke }` record, the test reads `via %s: preserves %s on downstream Request`, and the source is ~50 lines shorter for the same coverage.

packages/browser/src/__tests__/extensions/replay/external/xhr-wrapper-invariants.test.ts
pauldambra: > [!NOTE]

Addressed in 094e7c07 — author chose option (a): parameterised over JSON string, plain text, empty string, URL-encoded, Blob, ArrayBuffer, Uint8Array, URLSearchParams, FormData, and null (the full XMLHttpRequestBodyInit union minus Document, which jsdom would need for and isn't relevant to the bug). Also tightened the assertion from `toEqual([body])` to `toBe(body)` so a wrapper that *copies* the body (which would break FormData boundaries and ArrayBuffer transfer) gets caught. Polyfilled `Document` in the test env because `isDocument(body)` does `instanceof Document` which throws in node.

marandaneto (approved):

marandaneto:
unclear if its all requests or just specific endpoints, if its only if session recording is enabled or any other feature, but tests are always good anyway!

PR #3634, opened 2026-05-19
feat(surveys): opt-in back button for multi-question surveys

lucasheriques:
## Summary

Adds a built-in, opt-in **Back** button to web surveys so respondents can return to a previously visited question and edit their answer. Today the surveys SDK is strictly forward-only — this fills a real UX gap, especially on multi-question and branched surveys.

- New appearance flags: `allowGoBack` (default `false`) and `backButtonText` (defaults to "Back"). Added to `@posthog/core`'s `SurveyAppearance` so the field is discoverable for future RN parity.
- **History stack**, not `currentIndex - 1`: `InProgressSurveyState` gains an optional `visitedIndices: number[]` so Back respects the path the user actually took through branching surveys (`response_based`, `specific_question`). Existing in-progress state without `visitedIndices` resumes safely with the button hidden until the next forward step.
- Prior answers are pre-filled on the question you return to via the existing `initialValue` path — no retyping.
- Back button hidden in preview mode, on the first visited question, and whenever `allowGoBack` is not set (preserves existing behavior for every current survey).

## Out of scope

- React Native (`packages/react-native/src/surveys/`) — separate parity PR, the new appearance fields live on the shared core type so it can land next.
- No public programmatic navigation API on `PostHogSurveys` yet — deferred until a concrete external ask.

## Test plan

- [x] Unit tests (`packages/browser/src/__tests__/extensions/surveys/back-button.test.tsx`, 7 cases): visibility on first question, hidden when flag unset, advance+back round-trip with pre-fill, history survives re-render from persisted state, branching respected (Q1 → Q3 → back lands on Q1, not Q2), legacy state without `visitedIndices` doesn't crash, custom `backButtonText` honored.
- [x] Playwright (`packages/browser/playwright/mocked/surveys/back-button.spec.ts`, 2 cases): full bundle smoke for the positive and negative path.
- [x] Lint, surveys typecheck, all 420 surveys unit tests pass.
- [ ] Manual smoke in playground — linear + branching survey, confirm pre-fill and CSS rendering.

---


lucasheriques:
Pushed `d6adead` addressing the review:

**Greptile #1 — DOM structure regression**: fixed. `form-buttons` div only wraps submit when the Back button is actually shown. Existing surveys keep their original DOM (submit as a direct child of `.bottom-section`).

**Greptile #2 — stale responses on branch change**: fixed. On submission (partial and final), we now prune `newResponses` to only include question IDs corresponding to the indices in `visitedIndices ++ [currentIndex]`. So if a respondent answers Q1 → Q2 → backs up to Q1 → changes their answer, now routing to Q4, the emitted payload contains only Q1 and Q4 — Q2's answer is dropped. Added a unit test covering exactly that path (`'prunes responses from abandoned branches when path changes after backing up'`). `questionsResponses` state is intentionally kept intact so re-walking the same path still pre-fills.

**Greptile #3 — missing `waitFor`**: fixed.

**Greptile #4 — parameterise tests**: respectfully pushing back. The 7 cases assert distinct behaviors (visibility, branching, persistence resume, pre-fill, custom label), each with different setup or assertions. A `test.each` table would couple them together and hurt readability. I'd rather keep one named test per behavior — easier to reason about a failure, and adding "future combinations" hasn't been a problem here.

**CI failures**: all cascaded from one root cause — adding `allowGoBack`/`backButtonText` to core's `SurveyAppearance` broke `posthog-react-native`'s `SurveyAppearanceTheme = Required<SurveyAppearance>` (the new fields became required and `defaultSurveyAppearance` didn't supply them). Fixed by adding `allowGoBack: false` and `backButtonText: 'Back'` to RN's `defaultSurveyAppearance`. This is a no-op for RN at runtime (RN doesn't render the back button yet) and pre-wires the defaults for the eventual RN parity PR.

Added a changeset (minor bump for both `posthog-js` and `@posthog/core`).

marandaneto:
> Pushed `d6adead` addressing the review:
> 
> **Greptile #1 — DOM structure regression**: fixed. `form-buttons` div only wraps submit when the Back button is actually shown. Existing surveys keep their original DOM (submit as a direct child of `.bottom-section`).
> 
> **Greptile #2 — stale responses on branch change**: fixed. On submission (partial and final), we now prune `newResponses` to only include question IDs corresponding to the indices in `visitedIndices ++ [currentIndex]`. So if a respondent answers Q1 → Q2 → backs up to Q1 → changes their answer, now routing to Q4, the emitted payload contains only Q1 and Q4 — Q2's answer is dropped. Added a unit test covering exactly that path (`'prunes responses from abandoned branches when path changes after backing up'`). `questionsResponses` state is intentionally kept intact so re-walking the same path still pre-fills.
> 
> **Greptile #3 — missing `waitFor`**: fixed.
> 
> **Greptile #4 — parameterise tests**: respectfully pushing back. The 7 cases assert distinct behaviors (visibility, branching, persistence resume, pre-fill, custom label), each with different setup or assertions. A `test.each` table would couple them together and hurt readability. I'd rather keep one named test per behavior — easier to reason about a failure, and adding "future combinations" hasn't been a problem here.
> 
> **CI failures**: all cascaded from one root cause — adding `allowGoBack`/`backButtonText` to core's `SurveyAppearance` broke `posthog-react-native`'s `SurveyAppearanceTheme = Required<SurveyAppearance>` (the new fields became required and `defaultSurveyAppearance` didn't supply them). Fixed by adding `allowGoBack: false` and `backButtonText: 'Back'` to RN's `defaultSurveyAppearance`. This is a no-op for RN at runtime (RN doesn't render the back button yet) and pre-wires the defaults for the eventual RN parity PR.
> 
> Added a changeset (minor bump for both `posthog-js` and `@posthog/core`).

make your review pr workflow to reply directly to the comment thread, this way you have to jump context and the # id is easily out of sequence/outdated

marandaneto (approved):

marandaneto:
@PostHog/team-surveys ping

lucasheriques:
@marandaneto thakns for the ping, will get this in, just wanted to run one last test before merging

marandaneto:
> @marandaneto thakns for the ping, will get this in, just wanted to run one last test before merging

moving this to draft so its out of my pending review filters until its actually good to go

PR #3639, opened 2026-05-20
feat: use async gzip for replay event compression

marandaneto:
## Problem

Session recording event compression still used synchronous `fflate` gzip on the main thread, even though the browser SDK request path can use native async `CompressionStream` gzip when available.

## Changes

- Use `@posthog/core`'s `gzipCompress` / `isGzipSupported` for session recording event compression when native async gzip is available.
- Preserve replay event ordering by queueing events while async compression is in flight.
- Stop rrweb/mutation producers synchronously during `stop()` while allowing already queued async compression to drain and flush.
- Synchronously drain pending compression on `beforeunload` so sendBeacon can include queued replay events.
- Fall back to the existing synchronous `fflate` path when native gzip is unsupported or fails.
- Export `isGzipSupported` from `@posthog/core`'s main entrypoint.
- Add unit tests for the async native gzip path, the synchronous fallback path, queued event ordering, stop-time queue draining, and beforeunload sync draining.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -206,6 +226,87 @@ function gzipField(data: unknown): string {
     return gzipToString(data)
 }
 
+async function gzipFieldAsync(data: unknown): Promise<string> {
+    if (isArray(data) && data.length === 0) {
+        if (_gzippedEmptyArray) {
+            return _gzippedEmptyArray
+        }
```
marandaneto, line 239: The async path mirrors that so we don’t create a new CompressionStream, encode, validate, read the Blob, etc. for every empty array field

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -192,11 +195,28 @@ export type compressedEventWithTime = compressedEvent & {
...
+
+async function gzipToStringAsync(data: unknown): Promise<string> {
+    const serializedData = JSON.stringify(data)
+    const compressed = await gzipCompress(serializedData, Config.DEBUG, { rethrow: true })
+    if (!compressed) {
+        return gzipStringToString(serializedData)
+    }
+
+    return strFromU8(new Uint8Array(await compressed.arrayBuffer()), true)
 }
```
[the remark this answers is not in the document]
marandaneto: Fixed in 08f971c9 — removed the unreachable null fallback now that `gzipCompress` is called with `{ rethrow: true }`.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording-compression.test.ts
```diff
@@ -0,0 +1,188 @@
...
+        expect(gzipCompress).not.toHaveBeenCalled()
+        expect(posthog.capture).toHaveBeenCalledWith(
+            '$snapshot',
+            expect.objectContaining({
+                $snapshot_data: [expect.objectContaining({ type: 2, cv: '2024-10', data: expect.any(String) })],
+            }),
+            expect.any(Object)
+        )
+    })
+})
```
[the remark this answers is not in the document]
marandaneto: Fixed in 08f971c9 — refactored the sync/async compression path coverage into parameterized tests and added stop-time queue drain coverage.

marandaneto:
Addressed Greptile feedback in 08f971c9: in-flight async compression is now drained before `stop()` teardown, the unreachable async gzip null fallback was removed, and compression path tests were refactored to use parameterized coverage.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -206,6 +222,87 @@ function gzipField(data: unknown): string {
     return gzipToString(data)
 }
 
+async function gzipFieldAsync(data: unknown): Promise<string> {
+    if (isArray(data) && data.length === 0) {
+        if (_gzippedEmptyArray) {
+            return _gzippedEmptyArray
+        }
+        _gzippedEmptyArrayPromise =
```
pauldambra, line 240: asking so i don't forget tomorrow when i'm awake enough to think

do we lose order of operation guarantees here
replay does some things with timestamps and ordering (IIRC)

way too tired to actually think though so could be being stupid

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1022,14 +1141,60 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+                this._teardown()
+                logger.info('stopped')
+            })
+
+        return true
+    }
+
     stop() {
+        if (this._stopAfterCompressionQueueDrains()) {
+            return
```
TueHaulund, line 1214: We def. need to defer the flush and clear but we should tear down rrweb mutation observers synchronously. Otherwise we keep collecting mutations and we keep producing to the compression queue.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1022,14 +1141,60 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            .catch(() => undefined)
+            .then(() => {
+                if (generation !== this._compressionQueueGeneration) {
+                    return
+                }
+
+                this._isStoppingAfterCompression = false
+                this._flushBuffer()
+                this._clearBuffer()
+                this._teardown()
```
dustinbyrne, line 1198: we stop rrweb in `teardown()`, so will it continue to write snapshots to the queue until then?

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1042,6 +1207,65 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            } finally {
+                if (generation === this._compressionQueueGeneration) {
+                    this._queuedCompressionEvents = Math.max(0, this._queuedCompressionEvents - 1)
+                }
+            }
+        }
+
+        this._compressionQueue = this._compressionQueue
+            ? this._compressionQueue.catch(() => undefined).then(processEvent)
+            : processEvent()
```
dustinbyrne, line 1339: if we hit `_onBeforeUnload` while asynchronously compressing, I think we'd end up dropping any pending events?

we can't send them via the sync `sendBeacon`

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: no, thats why we have the _compressionQueue so the order is guaranteed

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: Fixed in 074b3f24 — `stop()` now stops rrweb/mutation producers synchronously before waiting for the existing compression queue to drain and flush.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: Fixed in 074b3f24 — rrweb is stopped immediately when stop begins, so it cannot continue adding snapshots while queued compression drains.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: Fixed in 074b3f24 — `beforeunload` now synchronously drains pending compression jobs with the sync fflate path before flushing, so sendBeacon can include them.

marandaneto:
@PostHog/team-replay can you test this change before we merge it? i dont know the whole canvas code path enough, so i feel like merging this without testing properly is risky

fasyy612 (approved):
Seems to work locally 👌

PR #3646, opened 2026-05-21
fix: skip PostHogProvider init without api key

marandaneto:
## Problem

A user reported errors when rendering `PostHogProvider` without an API key/project token. The React provider warned about the missing `apiKey`/`client`, but still attempted to initialize the default PostHog instance with an undefined token. The Next.js helpers also threw when no API key was configured, which can unnecessarily break apps.

## Changes

- Return early from the React provider effect when neither `client` nor `apiKey` is provided, preserving the documented/manual global-instance behavior.
- Update Next.js API key resolution to warn and return `undefined` instead of throwing when no API key is configured.
- Make Next.js config resolution accept unknown runtime values safely: non-string API keys/hosts normalize to `undefined` and fall back to env/defaults where applicable.
- Update Next.js App Router and Pages Router providers to render children without initializing PostHog when no API key is available.
- Update Next.js middleware to skip cookie seeding when no API key is available, while still allowing proxy rewrites.
- Update Next.js server helpers to return a disabled PostHog client when no API key is available instead of throwing.
- Add regression tests covering missing/non-string API key behavior for React, Next App Router, Next Pages Router, middleware, server helpers, and shared config.
- Add a patch changeset for `posthog-js`, `@posthog/react`, and `@posthog/next`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/quiet-keys-warn.md
+++ b/.changeset/quiet-keys-warn.md
@@ -1,6 +1,7 @@
 ---
 "@posthog/react": patch
+"@posthog/next": patch
 "posthog-js": patch
 ---
 
-Avoid initializing PostHogProvider when no API key or client is provided
+Avoid throwing or initializing PostHogProvider when no API key or client is provided
--- a/.changeset/quiet-keys-warn.md
+++ b/.changeset/quiet-keys-warn.md
@@ -1,0 +1,6 @@
+---
+"@posthog/react": patch
+"posthog-js": patch
+---
+
+Avoid initializing PostHogProvider when no API key or client is provided
--- a/packages/next/src/app/PostHogProvider.tsx
+++ b/packages/next/src/app/PostHogProvider.tsx
@@ -5,7 +5,7 @@ import type { BootstrapConfig } from '../client/ClientPostHogProvider.js'
 import { cookies } from 'next/headers.js'
 import type { PostHogOptions } from 'posthog-node'
 import { getOrCreateNodeClient } from '../server/nodeClientCache.js'
-import { NEXTJS_CLIENT_DEFAULTS, resolveApiKey, resolveHostOrDefault } from '../shared/config.js'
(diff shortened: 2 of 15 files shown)


packages/next/src/app/PostHogProvider.tsx
```diff
@@ -75,6 +75,10 @@ export async function PostHogProvider({
     children,
 }: PostHogProviderProps) {
     const apiKey = resolveApiKey(apiKeyProp)
```
ioannisj, line 77: Should `resolveApiKey` return undefined for empty strings as well?

ioannisj (approved):

packages/next/src/app/PostHogProvider.tsx
marandaneto: good point

marandaneto: it was like this already btw
added some tests to confirm [`4132d46` (this PR)](https://github.com/PostHog/posthog-js/pull/3646/changes/4132d4674c477269d27c8ecc161ebdf3aa518b7c)

.changeset/quiet-keys-warn.md
```diff
@@ -0,0 +1,6 @@
+---
+"@posthog/react": patch
+"posthog-js": patch
```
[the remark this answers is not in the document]
marandaneto: it does publish the React provider under
 posthog-js/react via packages/browser package files

PR #3648, opened 2026-05-21
chore: stabilize logs performance test

marandaneto:
## Problem

The logs entrypoint unit test for logging a 2MB object with many keys is flaky on GitHub runners. A recent run measured ~111ms against a strict 100ms threshold, causing an otherwise passing unit test job to fail.

## Changes

Raised the threshold for that specific performance assertion from 100ms to 150ms and added a comment documenting runner variance. This keeps the regression check while avoiding failures from small CI timing fluctuations.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/entrypoints/logs.test.ts
+++ b/packages/browser/src/__tests__/entrypoints/logs.test.ts
@@ -419,7 +419,7 @@ describe('logs entrypoint', () => {
             console.log(`Performance test (big body): wrapped=${wrappedTime.toFixed(2)}ms`)
         })
 
-        it('should not take more than 100ms to log a 2MB object with lots of keys', () => {
+        it('should not take more than 150ms to log a 2MB object with lots of keys', () => {
             const initializeLogs = assignableWindow.__PosthogExtensions__.logs.initializeLogs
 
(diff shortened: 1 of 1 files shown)


marandaneto:
possible improvements:

> Suggestions if we want to make it genuinely faster later:
> 
>  - Main bottleneck: logs.ts fully JSON.stringifys the 2MB object, then
>    truncates to 10KB.
>  - Faster approach: implement a size-budgeted serializer that stops around
>    LOG_BODY_SIZE_LIMIT instead of serializing the whole object.
>  - Keep that as a separate change because it touches behavior around
>    circular refs, Error, unreadable getters, and body formatting.

but unblocking CI for now

dustinbyrne (approved):

PR #3649, opened 2026-05-21
fix: speed up logs serialization

marandaneto:
## Problem

Console log capture serializes log bodies with `JSON.stringify` before truncating them to 10KB. For very large objects, this means we still pay the cost of serializing the whole object even though most of the result is discarded.

## Changes

Replaced full-body serialization with a size-budgeted serializer that stops once the log body reaches `LOG_BODY_SIZE_LIMIT`. The serializer keeps the existing safety behavior for circular references, `Error` objects, unreadable properties, arrays, and objects. Added a focused regression test to ensure serialization stops reading properties once the body size limit is reached.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


marandaneto:
follow up https://github.com/PostHog/posthog-js/pull/3648

marandaneto:
didnt test it yet but you got the idea

packages/browser/src/entrypoints/logs.ts
```diff
@@ -44,59 +44,177 @@ const setupOpenTelemetry = (posthog: PostHog) => {
...
+}
+
+const stringifyStringWithLimit = (value: string, parts: string[], budget: StringifyBudget): boolean => {
+    if (value.length + 2 <= budget.remaining) {
+        return appendWithLimit(parts, JSON.stringify(value), budget)
+    }
+
+    budget.truncated = true
+    return appendWithLimit(parts, JSON.stringify(value.slice(0, budget.remaining)), budget)
+}
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 1b9ae066 — string truncation now budgets against escaped JSON output and includes a regression test for escaped characters at the limit.

packages/browser/src/entrypoints/logs.ts
```diff
@@ -44,59 +44,177 @@ const setupOpenTelemetry = (posthog: PostHog) => {
...
-        try {
-            result[key] = sanitizeForStringify((value as any)[key], seen)
-        } catch {
-            // omit properties that cannot be read
-        }
+
+    if (seen.has(value)) {
+        return stringifyStringWithLimit('[Circular]', parts, budget)
     }
+    seen.add(value)
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 1b9ae066 — moved circular-reference tracking before toJSON handling and added a regression test for toJSON returning itself.

marandaneto:
@PostHog/logs @PostHog/team-client-libraries this should be good to go, and the threshold for the perf/flaky test is back to 100ms

dustinbyrne:
@marandaneto did you benchmark this by chance?

dustinbyrne:
# Logs serializer benchmark: branch vs main

Benchmarked `fix/logs-budgeted-serializer@1b9ae0664` against `main@3d41c1d9d`.

## Method

- Mocked OpenTelemetry and `window` dependencies.
- Measured wrapped `console.log` overhead.
- Runtime: Node `v24.13.0` on macOS arm64.
- Metric: median milliseconds per log call across **30 samples**; lower is better.
- Each sample averaged many calls:
  - small object: 200,000 logs/sample
  - medium object: 60,000 logs/sample
  - large string/object cases: 300 logs/sample, except flat object at 30 logs/sample

## Results

| case | main ms/log | branch ms/log | result |
|---|---:|---:|---:|
| small object | 0.0007 | 0.0007 | ~same |
| medium object, 1KB body | 0.0017 | 0.0023 | branch ~25% slower |
| 2MB string property | 0.8716 | 0.9329 | branch ~7% slower |
| 2MB flat object, ~52k keys | 15.2714 | 7.8931 | branch ~1.93x faster |
| mixed args, first arg 2MB string | 1.1133 | 0.9522 | branch ~1.17x faster |
| circular representative object | 0.0080 | 0.0092 | branch ~13% slower |

packages/browser/src/entrypoints/logs.ts
```diff
@@ -44,58 +44,196 @@ const setupOpenTelemetry = (posthog: PostHog) => {
...
+                return false
+            }
+            isFirst = false
+
+            if (!appendWithLimit(parts, `${JSON.stringify(key)}:`, budget)) {
+                return false
+            }
+            if (!stringifyValueWithLimit(propertyValue, parts, budget, seen, false)) {
+                return false
+            }
```
dustinbyrne: A custom `toJSON` implementation could return `undefined` (or some other non-serializable value, though I don't know why you'd do this), and `isJSONSerializablePrimitive` above wouldn't catch it

The key is already written by this point, so you'd end up with an invalid output like `{ "key": }`

packages/browser/src/entrypoints/logs.ts
```diff
@@ -44,58 +44,196 @@ const setupOpenTelemetry = (posthog: PostHog) => {
...
+    }
 
-const stringifySafely = (value: any, replacer: (_: any, value: any) => any): string | undefined => {
+    if (!appendWithLimit(parts, '{', budget)) {
+        return false
+    }
+    let isFirst = true
     try {
-        return JSON.stringify(value, replacer)
+        for (const key in value) {
```
dustinbyrne, line 206: An interesting thing my agent found:

`Object('abc')` and `String('abc')` make it here to be serialized, resulting in a stringified value of `{"0":"a","1":"b","2":"c"}`

`JSON.stringify` would correctly serialize as `"abc"`.

```ts
it('should serialize String objects like JSON.stringify does', () => {
    const initializeLogs = assignableWindow.__PosthogExtensions__.logs.initializeLogs
    initializeLogs(mockPostHog)

    assignableWindow.console.log(new String('abc'))

    expect(mockEmit.mock.calls[0][0].body).toEqual(JSON.stringify(new String('abc')))
})
```

marandaneto:
> 2MB

yep with the 2mb objects which were faster and CI happier
didnt measure with smaller objects, which seems to be slower tho, a trade off :(

packages/browser/src/entrypoints/logs.ts
marandaneto: Good catch! Fixed in 66920c78 — object properties are now serialized into a temporary budget before writing the key, so properties whose toJSON returns undefined/non-serializable output are omitted instead of corrupting the body. Added a regression test too.

packages/browser/src/entrypoints/logs.ts
marandaneto: Good catch! Fixed in ed7f4467 — boxed String/Number/Boolean values now serialize via their primitive value to match JSON.stringify, and I added a regression test for the String case plus Number/Boolean.

marandaneto:
Thanks for the benchmark. The main small-log regression I spotted after the recent correctness fix was per-property allocation: each object property created a temporary parts array and budget object before appending to the real output. I pushed 1a62b725 to serialize directly into the main buffer and roll back only when a value produces no output, which keeps the non-serializable toJSON fix while reducing allocations for small/medium objects. Logs tests and ESLint pass.

marandaneto:
@PostHog/team-client-libraries review pls

packages/browser/src/entrypoints/logs.ts
```diff
@@ -44,59 +44,225 @@ const setupOpenTelemetry = (posthog: PostHog) => {
...
+        // If toJSON can't be read or throws, fall through to safe property enumeration.
     }
-    for (const key of keys) {
-        try {
-            result[key] = sanitizeForStringify((value as any)[key], seen)
-        } catch {
-            // omit properties that cannot be read
-        }
+
+    const objectTag = Object.prototype.toString.call(value)
```
dustinbyrne: This _technically_ can throw, though in practice, I'd be surprised to see it. Something like logging a cross-origin iframe/window might trigger it?

Could be worth wrapping this in an exception handler

dustinbyrne (approved):

packages/browser/src/entrypoints/logs.ts
marandaneto: good catch

marandaneto: Good catch! Fixed in 8b9e0dd5 — `Object.prototype.toString.call` (and the boxed primitive `valueOf` handling around it) is now guarded and falls back to safe property enumeration. I also checked the rest of this diff: the other Object.prototype calls are hasOwnProperty checks already inside try/catch, and the imported number/boolean checks are now guarded too.

PR #3658, opened 2026-05-22
fix(node): fire separate $feature_flag_called events per group context

### LINKED ISSUE #3651: $feature_flag_called dedupe does not include group context for group-scoped flags
## Summary

In `posthog-node`, `$feature_flag_called` events appear to be deduped by `(distinctId, flag key, response)` only. The group context passed to feature flag evaluation is attached to the captured event, but does not appear to participate in the SDK-level dedupe key.

For group-scoped feature flags / experiments, this can suppress exposure events for a second group when the same user evaluates the same flag and receives the same response in both groups.

I'm not sure if this is intended behavior or if there's a recommended usage pattern we're missing, so filing this as a question / possible bug.

## Why this matters

For a group-scoped experiment, assignment can be based on the group key, and experiment metrics may join exposure events using a group property like `$group_0`.

If the same user visits two different groups and receives the same variant in both, the second group may not get a `$feature_flag_called` event. That means group-level exposure counts and warehouse-backed experiment joins can undercount exposed groups.

## Reproduction

Using `posthog-node@5.34.3`:

```ts

const client = new PostHog('ph_test_key', {
  host: 'http://127.0.0.1:9',
  flushAt: 9999,
  flushInterval: 0,
})

const captured: any[] = []
client.capture = (event: any) => captured.push(event)

function fire({
  distinctId,
  response,
  org,
}: {
  distinctId: string
  response: string
  org: string
}) {
  // Calling the internal method here only to isolate the dedupe behavior.
  ;(client as any)._captureFlagCalledEventIfNeeded({
    distinctId,
    key: 'group-scoped-experiment',
    response,
    groups: { org },
    properties: {
      $feature_flag: 'group-scoped-experiment',
      $feature_flag_response: response,
      '$feature/group-scoped-experiment': response,
    },
  })
}

fire({ distinctId: 'user_123', response: 'control', org: 'org_A' })
fire({ distinctId: 'user_123', response: 'control', org: 'org_B' })
fire({ distinctId: 'user_123', response: 'test', org: 'org_B' })
fire({ distinctId: 'user_456', response: 'control', org: 'org_B' })

console.log(
  captured.map(({ distinctId, event, groups, properties }) => ({
    distinctId,
    event,
    groups,
    response: properties.$feature_flag_response,
  })),
)
```

## Actual behavior

The second call is skipped:

```json
[
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_A" },
    "response": "control"
  },
  {
    "distinctId": "user_123",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "test"
  },
  {
    "distinctId": "user_456",
    "event": "$feature_flag_called",
    "groups": { "org": "org_B" },
    "response": "control"
  }
]
```

`user_123 + control + org_B` is not emitted because `user_123 + group-scoped-experiment + control` was already emitted for `org_A`.

## Expected / question

For group-scoped flags, should the dedupe key include group context, at least the aggregation group used by the flag?

For example:

```ts
distinctId + flag key + response + relevant group type/key
```

Or is the intended usage for group-scoped experiments to pass the group key as `distinctId`, e.g. `evaluateFlags(orgId, { groups: { org: orgId } })`?

## Relevant SDK code

In current `posthog-js` main, the dedupe key appears to ignore groups:

```ts
const { distinctId, key, response, groups, disableGeoip, properties } = params
const featureFlagReportedKey = `${key}_${response}`

if (
  distinctId in this.distinctIdHasSentFlagCalls &&
  this.distinctIdHasSentFlagCalls[distinctId].includes(featureFlagReportedKey)
) {
  return
}
```

`groups` is then attached only when the event is captured:

```ts
this.capture({
  distinctId,
  event: '$feature_flag_called',
  properties,
  groups,
  disableGeoip,
})
```

## Environment

- Observed with `posthog-node@5.34.3`
- Current npm latest checked: `posthog-node@5.34.10`
- Current `PostHog/posthog-js` main appears to have the same dedupe shape

## Context

We hit this while instrumenting a group-scoped A/A experiment where:

- the flag is scoped to an `org` group
- the exposure event is `$feature_flag_called`
- the experiment metric joins exposure using the org group key
- the same authenticated user can switch between multiple orgs in the same app/server process

The local workaround is to use a fresh PostHog client for this specific exposure path so the SDK-level dedupe cache does not span orgs, while keeping a singleton client for other server-side usage.

gustavohstrassburger:
## Problem

In the Node SDK, `$feature_flag_called` events were not fired when the same flag+variant was evaluated under different group contexts. The deduplication key only included the flag key and variant, so only the first call would fire an event regardless of which group was used.

Closes #3651

## Changes

Include group context in the `$feature_flag_called` deduplication key in `_captureFlagCalledEventIfNeeded`, so events fire independently per group combination.

Added a test in `evaluate-flags.spec.ts` verifying that two separate `$feature_flag_called` events fire for the same flag under different group contexts.

👉 _Stay up-to-date with [PostHog coding conventions](https://posthog.com/docs/contribute/coding-conventions) for a smoother review._

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/node/src/client.ts
```diff
@@ -1757,7 +1757,11 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
    */
   protected _captureFlagCalledEventIfNeeded(params: FlagCalledEventParams): void {
     const { distinctId, key, response, groups, disableGeoip, properties } = params
-    const featureFlagReportedKey = `${key}_${response}`
+    const groupSuffix =
+      groups && Object.keys(groups).length > 0
+        ? `_${JSON.stringify(Object.fromEntries(Object.entries(groups).sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0))))}`
```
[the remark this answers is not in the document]
gustavohstrassburger: `localeCompare` is locale-sensitive and slower than the default string comparator. For a stable hash key, `.sort()` or `.sort((a, b) => (a < b ? -1 : a > b ? 1 : 0))` is both faster and locale-independent

packages/node/src/__tests__/evaluate-flags.spec.ts
```diff
@@ -175,6 +175,49 @@ describe('evaluateFlags', () => {
...
+
+      const flags2 = await posthog.evaluateFlags('user-1', { groups: { team: 'red', company: 'org-a' } })
+      flags2.isEnabled('boolean-flag')
+
+      await waitForPromises()
+      const flagCalled = captures.filter(
+        (m) => m.event === '$feature_flag_called' && m.properties.$feature_flag === 'boolean-flag'
+      )
+      expect(flagCalled).toHaveLength(1)
+    })
```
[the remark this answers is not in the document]
gustavohstrassburger: Despite that, they differ on the purpose, so I think keeping it as separate tests is better.

packages/node/src/client.ts
dmarticus: Optional simplification: you can drop the `Object.fromEntries` wrap. The array form serializes just as deterministically for a hash key and skips an allocation:

```ts
`_${JSON.stringify(Object.entries(groups).sort(([a], [b]) => (a < b ? -1 : a > b ? 1 : 0)))}`
```

Also agreed with your reply to Greptile that bare `localeCompare` is the wrong tool here — it's locale-sensitive. If you ever want to swap the verbose comparator for something more readable, `a.localeCompare(b, 'en')` pins it to a deterministic locale. Totally a nit either way.

packages/node/src/client.ts
```diff
@@ -1757,7 +1757,11 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+        : ''
+    const featureFlagReportedKey = `${key}_${response}${groupSuffix}`
```
dmarticus, line 1764: Not a blocker for this PR, just flagging for follow-up: the existing eviction logic just below only fires on the *number* of distinct IDs — the inner `string[]` per distinctId has no cap, and each membership check is `Array.includes` (O(n)).

This PR amplifies the existing risk since dedup cardinality goes from `flags × variants` to `flags × variants × group-combinations`. For long-running Node processes that evaluate flags for one logical user across many tenants/groups, the inner array can grow and lookups slow down. Worth a follow-up to bound it (LRU) or switch to a `Set` for O(1) lookup.

packages/node/src/__tests__/evaluate-flags.spec.ts
dmarticus, line 219: Optional: consider locking down that `groups: undefined` and `groups: {}` dedup together. Both currently produce no suffix, so a single test asserting they collide would pin that contract — useful since users who omit `groups` on one call and pass `{}` on another would otherwise be surprised by either outcome changing.

dmarticus (approved):
Nice fix — the sorted JSON serialization makes the dedup key stable across insertion order, and the new tests cover the important cases (different groups fire separately, same groups dedup, key-order normalized).

Left a couple of optional style nits inline plus a flag on a pre-existing cache concern this PR amplifies (not a blocker). Reminder to add a patch changeset for `posthog-node` before merging — the changeset-hygiene bot already pinged on it.

PR #3661, opened 2026-05-22
chore: bump @posthog/cli to ~0.7.13

rafaeelaudibert:
## Summary

- Bumps `@posthog/cli` from `~0.7.3` to `~0.7.13` in the workspace catalog.
- 0.7.13 drops several unused runtime dependencies (`axios`, `axios-proxy-builder`, `console.table`, `rimraf` and their transitives like `tunnel`, `easy-table`, `glob@13`, `path-scurry`), so the lockfile is also pruned.
- Adds a changeset for the four consumers: `@posthog/nextjs-config`, `@posthog/nuxt`, `@posthog/rollup-plugin`, `@posthog/webpack-plugin`.

## Test plan

- [ ] `pnpm install --frozen-lockfile` passes locally
- [ ] CI for the affected packages stays green

### DIFF
--- a/.changeset/bump-posthog-cli-0-7-13.md
+++ b/.changeset/bump-posthog-cli-0-7-13.md
@@ -1,0 +1,8 @@
+---
+'@posthog/nextjs-config': patch
+'@posthog/webpack-plugin': patch
+'@posthog/rollup-plugin': patch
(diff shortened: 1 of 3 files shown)


cat-ph (approved):
🚀

PR #3662, opened 2026-05-23
feat(convex): move credentials and refresh cron onto the component

richardsolomou:
## Problem

In v1, the project token, host, and personal API key were either constructor options on `new PostHog(...)` or threaded through every action call. The refresh cron also had to be wired up by the consumer in their own `convex/crons.ts`. Convex 1.39 added typed env vars on `defineComponent`, which lets the component own its credentials and cron.

Also closes the v1 docs review on PostHog/posthog.com#16978 where @rafaeelaudibert asked whether the cron could be component-managed.

## Changes

- Component declares `POSTHOG_TOKEN`, `POSTHOG_HOST`, `POSTHOG_PERSONAL_API_KEY`, and the optional `POSTHOG_FLAGS_POLLING_INTERVAL_SECONDS` (default 60) as typed env vars. `PostHog` client constructor takes only callbacks (`identify`, `beforeSend`) now; credentials are dropped from every action arg.
- `src/component/crons.ts` registers the refresh cron only when `POSTHOG_PERSONAL_API_KEY` is set, gated at module-load (prod needs a redeploy to toggle). Local-eval methods throw when PAK isn't configured, pointing callers at the remote `evaluate*` methods. `posthog.reloadFeatureFlags(ctx)` stays for manual on-demand refresh, matching `posthog-node`'s naming.
- `POSTHOG_API_KEY` env var renamed to `POSTHOG_TOKEN`. Convex peer dep bumped to `^1.39.0`. Migration guide and "Differences from posthog-node" section added to `packages/convex/README.md`.

### Libraries affected

- [x] @posthog/convex

## Checklist

- [x] Tests for new code

### If releasing new changes

---


examples/example-convex/convex/convex.config.ts
```diff
@@ -1,9 +1,27 @@
...
 import agent from '@convex-dev/agent/convex.config'
 
-const app = defineApp()
-app.use(posthog)
+const app = defineApp({
+    env: {
+        POSTHOG_TOKEN: v.string(),
+        POSTHOG_HOST: v.optional(v.string()),
+        POSTHOG_PERSONAL_API_KEY: v.optional(v.string()),
+        POSTHOG_FLAGS_POLLING_INTERVAL_SECONDS: v.optional(v.string()),
```
rafaeelaudibert, line 11: Should it be a number?

rafaeelaudibert: Ah, ok, this is on Convex. I've sent a message to them asking how we can change this, but this is good for now, we just need to handle it being a string ourselves

packages/convex/README.md
```diff
@@ -21,29 +21,45 @@ Found a bug? Feature request? [File it here](https://github.com/PostHog/posthog-
...
-const app = defineApp();
-app.use(posthog);
+const app = defineApp({
+  env: {
+    POSTHOG_TOKEN: v.string(),
+    POSTHOG_HOST: v.optional(v.string()),
+    POSTHOG_PERSONAL_API_KEY: v.optional(v.string()),
+    POSTHOG_FLAGS_POLLING_INTERVAL_SECONDS: v.optional(v.string()),
+  },
+});
```
rafaeelaudibert, line 49: Add a comment explaining what each of these is for, please

packages/convex/README.md
```diff
@@ -349,6 +343,59 @@ Three methods:
...
+- **Local-eval methods don't auto-fall-back to remote.** `posthog-node`'s `getFeatureFlag` quietly hits `/flags` when local eval can't reach a verdict. Ours returns `undefined` (or `null` from `getFeatureFlagResult`) and you call `evaluateFlag` / `evaluateFlagPayload` / `evaluateAllFlags` explicitly for remote. Auto-fallback would force every local-eval call into an action context (since queries can't make network calls), which would defeat the reactivity win.
+- **Local-eval methods throw when `POSTHOG_PERSONAL_API_KEY` isn't configured.** `posthog-node` returns `undefined`; the throw here points you at the remote `evaluate*` methods so you can't get stuck wondering why your rollouts don't take effect.
+
+## ⬆️ Migrating from v1
+
+v2 moves credentials from the client constructor onto the component itself, using [Convex 1.39's typed component env vars](https://docs.convex.dev/components/authoring#environment-variables). It also bundles the refresh cron inside the component. The result is less plumbing per call site and a setup that's safe to leave running on free-tier dev deployments.
+
+To upgrade:
+
+1. **Bump your Convex peer dep** to `^1.39.0`.
```
rafaeelaudibert: It won't be a peer dep in people's codebases, it's an actual dep

rafaeelaudibert (changes requested):

packages/convex/src/component/convex.config.ts
```diff
@@ -1,3 +1,23 @@
...
+/**
+ * The component declares the env vars it needs so the installing app can wire them in
+ * `convex/convex.config.ts` (typically via `app.env.*` so existing project-level env vars
+ * pass straight through). All three are read via `process.env` inside the component's
+ * actions and cron — `POSTHOG_PERSONAL_API_KEY`'s presence is also what gates the local
+ * evaluation refresh cron.
+ */
+export default defineComponent('posthog', {
+  env: {
+    POSTHOG_TOKEN: v.string(),
```
ianmacartney: note: this will make deployments start requiring this env var on deploy - which will be a breaking change (currently I think you could deploy without this env var set)

examples/example-convex/convex/convex.config.ts
richardsolomou: Ok so sounds like this isn't being done right now, but we're already handling it in `readPollingIntervalSeconds()`, so I'll resolve this for now

packages/convex/src/component/convex.config.ts
richardsolomou: yep, thanks @ianmacartney. this is a major bump, so i'll keep it required for v2 to make it fail-fast at deploy time instead of silent no-opping at runtime. added a call out for this to the migration guide.

richardsolomou:
Renamed `token` to `projectToken` and `POSTHOG_TOKEN` to `POSTHOG_PROJECT_TOKEN` after adjacent discussion with @marandaneto [here](https://github.com/PostHog/posthog-js/pull/3668#pullrequestreview-4361623550).

packages/convex/README.md
```diff
@@ -53,49 +67,19 @@ To enable local feature flag evaluation, also set a [feature flags secure API ke
 npx convex env set POSTHOG_PERSONAL_API_KEY phs_your_feature_flags_secure_api_key
 ```
 
-> Personal API keys (`phx_…`) also still work for local evaluation, but PostHog recommends the project-scoped feature flags secure API key going forward.
+> Personal API keys (`phx_…`) also still work for local evaluation, but PostHog recommends the project-scoped feature flags secure API key going forward. Setting this env var also activates the component's built-in refresh cron — without it, the cron is skipped entirely so idle dev deployments don't burn function calls.
```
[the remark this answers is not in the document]
rafaeelaudibert: This makes sense, but it's a nit

rafaeelaudibert (approved):

packages/convex/README.md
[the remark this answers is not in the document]
richardsolomou: Already landed in 636c420b

PR #3665, opened 2026-05-25
fix: don't autocapture PostHog's own network errors in react-native

### LINKED ISSUE #3664: PostHogFetchNetworkError
### Description

Most of the errors I’m seeing from the Expo app are `PostHogFetchNetworkError`. I assume these happen when the user is offline, which seems like an expected scenario.

Would it make sense to avoid reporting these as errors? Ideally, there could be a retry mechanism for failed requests due to connectivity issues, but without treating them as actual application errors.

What do you think?

Image

ioannisj:
## Problem

Closes #3664

React Native / Expo apps with error-tracking autocapture enabled were seeing their dashboards flooded with `PostHogFetchNetworkError` events. This is PostHog's *own* internal error — thrown by the core when a `fetch` fails to reach the server (device offline or request timeout). It's an expected, non-actionable connectivity condition, not an application error, but the autocapture handlers reported it as a `$exception`.

The leak path: a user (or auto-flush) calls into the SDK, the network request fails offline, and the resulting `PostHogFetchNetworkError` surfaces as an uncaught exception / unhandled rejection — which the error-tracking handlers then capture and send.

## Changes

- **`@posthog/core`**: export a new `isPostHogFetchNetworkError(err): err is PostHogFetchNetworkError` type guard (`instanceof`-based). The error class itself stays module-private; only the guard is public surface.
- **`posthog-react-native`**: the `onUncaughtException` and `onUnhandledRejection` autocapture handlers now skip errors matching the guard. The **console** autocapture path is intentionally left untouched, and `PostHogFetchHttpError` (bad API key, 5xx, etc.) is intentionally **not** filtered — those are developer-actionable and should remain visible.
- Tests: unit tests for handler wiring + an integration test that produces a genuine `PostHogFetchNetworkError` through core's real fetch path and verifies the guard recognizes it. Extracted shared `createMockPostHog()` / `createMockLogger()` into `test-utils.ts`.

Scope was deliberately limited to network errors only (not all PostHog-internal errors) to preserve visibility of misconfiguration errors like a bad API key.

### Libraries affected

- [x] posthog-react-native
- [x] @posthog/core

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -91,6 +91,18 @@ function isPostHogFetchError(err: unknown): err is PostHogFetchHttpError | PostH
   return typeof err === 'object' && (err instanceof PostHogFetchHttpError || err instanceof PostHogFetchNetworkError)
 }
 
+/**
+ * True for the network error PostHog throws when a request fails to reach the server
+ * (e.g. the device is offline or the request times out). Exposed so SDKs that autocapture
+ * errors can skip these expected connectivity failures instead of reporting them as
+ * application exceptions.
+ *
+ * @public
+ */
+export function isPostHogFetchNetworkError(err: unknown): err is PostHogFetchNetworkError {
+  return err instanceof PostHogFetchNetworkError
+}
+
 function isPostHogFetchContentTooLargeError(err: unknown): err is PostHogFetchHttpError & { status: 413 } {
   return typeof err === 'object' && err instanceof PostHogFetchHttpError && err.status === 413
 }
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -88,7 +88,7 @@ export async function logFlushError(err: any): Promise<void> {
 }
 
 function isPostHogFetchError(err: unknown): err is PostHogFetchHttpError | PostHogFetchNetworkError {
-  return typeof err === 'object' && (err instanceof PostHogFetchHttpError || err instanceof PostHogFetchNetworkError)
+  return typeof err === 'object' && (err instanceof PostHogFetchHttpError || isPostHogFetchNetworkError(err))
 }
 
 /**
--- a/packages/react-native/src/error-tracking/index.ts
+++ b/packages/react-native/src/error-tracking/index.ts
@@ -1,5 +1,11 @@
 import type { PostHog } from '../posthog-rn'
-import { JsonType, Logger, PostHogEventProperties, ErrorTracking as CoreErrorTracking } from '@posthog/core'
+import {
+  JsonType,
+  Logger,
+  PostHogEventProperties,
+  ErrorTracking as CoreErrorTracking,
+  isPostHogFetchNetworkError,
+} from '@posthog/core'
 import { trackConsole, trackUncaughtExceptions, trackUnhandledRejections } from './utils'
 import { getRemoteConfigBool, isHermes } from '../utils'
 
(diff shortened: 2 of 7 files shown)


packages/core/src/posthog-core-stateless.ts
```diff
@@ -91,6 +91,18 @@ function isPostHogFetchError(err: unknown): err is PostHogFetchHttpError | PostH
...
 
+/**
+ * True for the network error PostHog throws when a request fails to reach the server
+ * (e.g. the device is offline or the request times out). Exposed so SDKs that autocapture
+ * errors can skip these expected connectivity failures instead of reporting them as
+ * application exceptions.
+ *
+ * @public
+ */
+export function isPostHogFetchNetworkError(err: unknown): err is PostHogFetchNetworkError {
```
marandaneto, line 102: refactor the method above to use isPostHogFetchNetworkError instead of dupe `err instanceof PostHogFetchNetworkError` again

packages/react-native/src/error-tracking/index.ts
```diff
@@ -134,6 +140,11 @@ export class ErrorTracking {
         return
       }
 
+      // Offline/timeout failures are expected, not application errors.
+      if (isPostHogFetchNetworkError(error)) {
```
marandaneto, line 144: should we use `isPostHogFetchError` instead and also do not capture 4xx and 5xx fetch errors for the posthog API?

marandaneto (approved):
left one comment (suggestion) and one question, not a blocker, LGTM otherwise
the bot has comments as well

packages/react-native/src/error-tracking/index.ts
ioannisj: 5xx could make sense, but my thinking was that 4xx is probably a misconfig error that we should surface to the developers? Not strictly an application error, but not something expected as a connectivity error either. Since both are not expected to surface that often, I chose to keep them but I don't have a strong opinion on this tbh

packages/react-native/test/error-tracking-internal-errors.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    expect(mockPostHog.capture).toHaveBeenCalledTimes(1)
+  })
+
+  it('does not filter the console handler (console is left untouched)', () => {
+    ;(isPostHogFetchNetworkError as jest.Mock).mockReturnValue(true)
+    new ErrorTracking(mockPostHog, { autocapture: { console: ['error'] } }, mockLogger as any)
+    const consoleHandler = (trackConsole as jest.Mock).mock.calls[0][1]
+
+    consoleHandler(new Error('network'), false)
```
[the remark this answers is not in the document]
ioannisj: Still, the constructor options differ and mock setup is slightly different. I feel parameterizing would make the test harder to read so I'll keep as is

packages/react-native/src/error-tracking/index.ts
marandaneto: the way i see is 'is this error actionable, can i do something about it?'
for 4xx errors and 5xx errors, most likely the answer is always yes
what i am worried about is if we are entering an endless loop here

we capture an event/or error, API is 400..500, API throws, we capture the error again, and thats how we flood the server and waste cycles in the client

marandaneto: most likely @PostHog/team-error-tracking would have opinions here, and if we decide not to capture our own errors, we have to revisit this in all SDKs

ioannisj: Agreed, I'll merge this one to unblock the user though and we can revisit with a followup PR based on the discussion here

ablaszkiewicz: Seems like posthog-js is pretty bulletproof when it comes to handling errors produced by us. All stuff like fetch errors are gated behind `catch`es. And then in the whole error tracking pipeline we made sure not to throw unhandled errors (we've had these cases before). I'll look at other SDKs tho

PR #3667, opened 2026-05-25
fix(replay): stop polling preload-as-style <link> elements forever

pauldambra:
## Problem

A customer reported (with a detailed trace) that the session recorder was saturating the main thread on pages with `<link rel="preload" as="style" href="*.css">` elements — common in Next.js-chunked apps. They measured ~1,440 concurrently active polling chains on a single preload link after ~2 hours, firing ~408/s and consuming ~55% of the main thread. Backgrounding the tab + refocusing produced multi-hundred-ms freezes as Chrome flushed the queued timers.

Root cause is in `rrweb-snapshot`'s `onceStylesheetLoaded` path, which three defects compose into a runaway:

1. The link-tag predicate treats `rel="preload" as="style"` the same as `rel="stylesheet"`, but per spec preload links never instantiate a `CSSStyleSheet` — `link.sheet` stays `null` forever, so the timeout always fires.
2. `onceStylesheetLoaded` never removes its `load` listener, has no idempotency guard, and the load handler is not gated by `fired` — so every real load event multiplies the chain.
3. The timer-callback recursively re-enters `serializeNodeWithId` on the same link, scheduling another chain every cycle.

Upstream rrweb has the same bug ([rrweb-io/rrweb#1707](https://github.com/rrweb-io/rrweb/issues/1707), still open; fix attempt [#1708](https://github.com/rrweb-io/rrweb/pull/1708) closed unmerged).

## Changes

`packages/rrweb/rrweb-snapshot/src/snapshot.ts`:

- Drop `rel="preload"` from the stylesheet predicate. Preload links are still serialized normally by the surrounding pass — we just stop scheduling a never-ending wait on them.
- Replace the recursive `serializeNodeWithId` call inside the load callback with the already-computed `serializedNode` from the enclosing scope.
- Make `onceStylesheetLoaded` idempotent via a module-level `WeakSet`; share one `fired` flag between the timer path and the load path; pass `{ once: true }` so the listener self-removes after firing.

Tests:

- `packages/rrweb/rrweb-snapshot/test/snapshot.test.ts` — three new jsdom unit tests covering preload non-tracking, stylesheet load-firing-once, and synthetic-load-event no-multiplication.
- `packages/browser/playwright/mocked/session-recording/session-recording-preload-link-leak.spec.ts` + `packages/browser/playground/preload-link-leak/index.html` — real-browser repro that instruments `HTMLLinkElement.prototype.addEventListener`, lets the recorder run for two 6-second windows with a synthetic `load` dispatch in between, and asserts the load-add count stays bounded. Without the fix: 30 leaked listeners. With the fix: 0.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

---


pauldambra:
* **#3667** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3667?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -1283,50 +1292,16 @@ export function serializeNodeWithId(
...
-            maxDepth,
-          });
-
-          if (serializedLinkNode) {
-            onStylesheetLoad(
-              n as HTMLLinkElement,
-              serializedLinkNode as serializedElementNodeWithId,
-            );
-          }
+          onStylesheetLoad(
```

**[convergent: paul + qa-team/reliability]** 🟠 HIGH

Regression: late-loading external stylesheets lose their inlined CSS in the replay.

The old code re-called `serializeNodeWithId(n, …)` inside the load callback specifically so it could re-read `link.sheet` (now populated) and write `attributes._cssText = stringifyStylesheet(stylesheet)` (snapshot.ts:690-696). The new code passes the original pre-load `serializedNode` to `onStylesheetLoad`. Because `_cssText` is only set inside `serializeNode` when `link.sheet` is truthy at serialize time, the captured node has no `_cssText`, and `StylesheetManager.attachLinkElement` gates the mutation on `'_cssText' in (childSn as elementNode).attributes` (stylesheet-manager.ts:30) — so the cssText mutation is never emitted.

`trackStylesheetInLinkElement` is a TODO no-op, so there is no fallback. With `inlineStylesheet: true` being the default, every replay that has any `<link rel="stylesheet">` not loaded by the time of the FullSnapshot will silently render unstyled.

Fix: restore the `serializeNodeWithId` re-call inside the callback (the WeakSet guard prevents the chain from re-arming because `if (styleSheetLoaded) return;` returns early once the sheet is populated), or extract a smaller helper that just re-reads `_cssText` and grafts it onto the existing serializedNode. Either way, add a jsdom test that populates `link.sheet` after first serialize, fires `load`, and asserts the node delivered to `onStylesheetLoad` has a non-empty `_cssText`.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -1283,50 +1292,16 @@ export function serializeNodeWithId(
   if (
     serializedNode.type === NodeType.Element &&
     serializedNode.tagName === 'link' &&
-    typeof serializedNode.attributes.rel === 'string' &&
-    (serializedNode.attributes.rel === 'stylesheet' ||
-      (serializedNode.attributes.rel === 'preload' &&
-        typeof serializedNode.attributes.href === 'string' &&
-        extractFileExtension(serializedNode.attributes.href) === 'css'))
+    serializedNode.attributes.rel === 'stylesheet'
```
pauldambra, line 1305: > [!NOTE]

**[qa-team/compatibility]** 🟡 MEDIUM

The PR drops the `rel === 'preload' && extractFileExtension(href) === 'css'` branch entirely. That branch was originally added (upstream rrweb) to handle Next.js-style CSS preloads. Per spec, preload-as-style links never instantiate a `CSSStyleSheet` — so the wait was indeed dead — but if any consumer/test was relying on this branch to capture preloaded CSS into the replay, this is a silent behaviour change.

Non-blocking; please confirm via a search across PostHog replay tests that no playground or existing spec was depending on `_cssText` arriving for a `rel="preload" as="style"` link.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -396,12 +396,15 @@ function onceIframeLoaded(
   iframeEl.addEventListener('load', listener);
 }
 
+const stylesheetLoadTracked = new WeakSet<HTMLLinkElement>();
```

**[convergent: paul + qa-team/reliability]** 🟡 MEDIUM

Module-level WeakSet has no reset hook. In SPAs where `record()` is stopped and started again on the same document (or where the recorder is torn down and re-init'd), a `<link>` that was tracked in lifecycle N stays in the WeakSet in lifecycle N+1 — so `onceStylesheetLoaded` early-returns and the second recording never fires `onStylesheetLoad` for that link. `StylesheetManager.reset()` clears its own `trackedLinkElements`; this WeakSet has no equivalent.

Suggestion: either scope the WeakSet to the snapshot session (pass through the options bag) or expose a reset hook from the snapshot module that the recorder calls on stop.

**[xp]** 🟢 LOW

The name `stylesheetLoadTracked` describes the mechanism rather than the intent. A reader has to scroll up to learn what's being tracked and why. Something like `linksAwaitingStylesheetLoad` (or attaching the guard as a Symbol property on the link itself so the state lives with the thing it guards) would express the intent — "we never re-watch a link we've already watched" — without making the reader cross-reference the WeakSet's call sites.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +414,24 @@ function onceStylesheetLoaded(
...
+  const fire = () => {
+    if (fired) return;
     fired = true;
     listener();
-  });
+  };
+
+  const timer = setTimeout(fire, styleSheetLoadTimeout);
+
+  link.addEventListener(
```

**[xp]** 🟢 LOW

Idempotency is now expressed three times: `WeakSet` guard at the top, `{ once: true }` on the listener, and the `fired` closure flag inside `fire()`. Rule 3 (OnceAndOnlyOnce) is taking some friendly fire here.

Consider: the WeakSet alone prevents re-registration, and `{ once: true }` self-removes the listener. The `fired` boolean only matters for the timer-vs-listener race (timer fires after listener was already attached `{ once: true }`-style and ran). If you can show that race can't happen, inline the listener back to the two-line form. If the race is real, document it next to `fired` with a one-liner comment so the third guard reads as load-bearing rather than belt-and-braces.

packages/browser/playwright/mocked/session-recording/session-recording-preload-link-leak.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+                .forEach((l) => l.dispatchEvent(new Event('load')))
+        })
+
+        await page.waitForTimeout(6000)
+
+        const loadAdds = await page.evaluate(() => {
+            return (window as unknown as { __preloadLoadAdds?: number }).__preloadLoadAdds ?? 0
+        })
+
+        expect(loadAdds).toBeLessThanOrEqual(5)
```
pauldambra: > [!NOTE]

**[convergent: paul + xp + qa-team/generalist-b]** 🟡 MEDIUM

`toBeLessThanOrEqual(5)` is too loose. After the predicate change, preload-as-style links bypass `onceStylesheetLoaded` entirely — they never get a load listener from the recorder, so the correct expected value is 0. A regression that reintroduces one listener per link per timer cycle would still pass the first cycle's check at the current bound.

Fix: `expect(loadAdds).toBe(0)`. Tighter, more honest about the contract, and would catch any future regression where the preload filter is loosened.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -584,3 +584,140 @@ describe('link href capture across SPA navigations', () => {
     expect(result).not.toBe('should-not-be-used');
   });
 });
```
pauldambra, line 587: > [!NOTE]

**[convergent: paul + xp + qa-team/generalist-a]** 🟡 MEDIUM

Three near-identical tests share substantial setup (create link, stub `sheet`, append, build options blob, dispatch, assert). Two of them inline the entire `serializeNodeWithId` options object even though `serializeLink` already extracts that helper. Project conventions explicitly prefer parameterized tests.

More importantly: none of these tests assert what `onStylesheetLoad` receives. They check call count only. That blind spot is what would let the convergent HIGH finding above ship unnoticed — a test that populates `link.sheet` after first serialize, fires `load`, and asserts the node delivered to `onStylesheetLoad` has a non-empty `_cssText` would catch the regression.

Fix: parameterize with `it.each([...])` driving one body, and add at least one row that exercises a stylesheet whose sheet populates between serialize and load, asserting `_cssText` lands on the delivered serializedNode.

packages/browser/playground/preload-link-leak/index.html
```diff
@@ -0,0 +1,24 @@
+<html>
+<head>
+    <title>Preload-as-style link listener leak repro</title>
+    <link rel="preload" as="style" href="/preload-leak/chunk-a.css" />
+    <link rel="preload" as="style" href="/preload-leak/chunk-b.css" />
+    <link rel="preload" as="style" href="/preload-leak/chunk-c.css" />
+    <link rel="preload" as="style" href="/preload-leak/chunk-d.css" />
+    <link rel="preload" as="style" href="/preload-leak/chunk-e.css" />
```
pauldambra, line 11: > [!NOTE]

**[qa-team/generalist-b]** ⚪ NIT

The five referenced CSS chunks (`/preload-leak/chunk-a.css` etc.) don't exist in the playground directory. That's intentional for the leak repro (network errors are exactly what guarantees preload-as-style links never get a sheet), but it does mean the page logs five network errors. A short HTML comment documenting that the missing files are intentional would help future maintainers understand the test is deliberately pointing at non-existent URLs.

Non-blocking.

.changeset/fix-preload-link-listener-leak.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): stop polling preload-as-style `<link>` elements forever. Session recorder treated `<link rel="preload" as="style" href="*.css">` as if it were a stylesheet and waited for `link.sheet` to populate. Per spec preload links never instantiate a `CSSStyleSheet`, so the wait timed out, re-serialized the link, scheduled another wait, and leaked a `load` listener on every cycle — multiplying further on every real `load` event. Pages with Next.js-style CSS preloads accumulated thousands of active polling chains, saturating the main thread and freezing the tab on refocus
```
pauldambra, line 7: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

Changeset is clear and trace-driven — properly explains the cause not just the symptom, including the multi-defect composition. One small polish: "multiplying further on every real `load` event" reads a touch awkwardly mid-sentence; "and the count multiplied further on every real `load` event" would scan more cleanly. Non-blocking.

pauldambra (commented):
> [!NOTE]
>

One convergent HIGH correctness finding (paul + qa-team/reliability): the new code reuses the pre-load `serializedNode` instead of re-running `serializeNodeWithId` on load, which silently drops `_cssText` for late-loading external stylesheets — `StylesheetManager.attachLinkElement` gates the mutation on `'_cssText' in attributes`, so the CSS never reaches the replay.

### Key findings

- 🟠 **HIGH (convergent: paul + qa-team/reliability)** — `snapshot.ts:1301` — late-loading `<link rel="stylesheet">` loses `_cssText` in the replay because the old recursive `serializeNodeWithId` is gone.
- 🟡 **MEDIUM (convergent: paul + xp + qa-team/generalist-b)** — Playwright spec — assertion bound `toBeLessThanOrEqual(5)` is too loose; should be `toBe(0)` (preload links now bypass `onceStylesheetLoaded` entirely).
- 🟡 **MEDIUM (convergent: paul + xp + qa-team/generalist-a)** — `snapshot.test.ts` — three near-identical tests; project conventions prefer parameterized tests, and none of them assert that `_cssText` reaches the `onStylesheetLoad` payload.
- 🟡 **MEDIUM (qa-team/reliability + paul)** — module-level `stylesheetLoadTracked` WeakSet not reset across recorder lifecycles.
- 🟡 **MEDIUM (qa-team/compatibility)** — preload-as-style branch removed; confirm no replay was relying on it for CSS capture.
- 🟢 **LOW (xp)** — triple idempotency (WeakSet + `{once:true}` + `fired`) reads as belt-and-braces; consider trimming.
- ⚪ **NIT** — WeakSet name, playground comment, changeset wording.

### Convergence

- `_cssText` regression: paul + qa-team/reliability (highest confidence).
- Playwright assertion looseness: paul + xp + qa-team/generalist-b.
- Test parameterization + missing `_cssText` payload assertion: paul + xp + qa-team/generalist-a.
- WeakSet cross-lifecycle: paul + qa-team/reliability.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | REQUEST CHANGES — leak fix is correct, but the recursive-call removal regresses CSS capture for late-loading stylesheets. |
| 👤 paul | REQUEST CHANGES — solid trace-driven write-up, but wants the late-loading `_cssText` path either restored or covered by a test before stamping. |
| 📐 xp | APPROVE WITH NITS — the predicate simplification is good XP work; idempotency triple-guard and the un-parameterised tests are the cleanups he'd want. |
| 🛡 security-audit | APPROVE — strictly shrinks attack surface; no new sinks, no new user-input paths, no exploitable change. |

---

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Intentional — `WeakSet`s in this file use the `*Tracked` / `tracked*` pattern (cf. `StylesheetManager.trackedLinkElements` next door in `record/stylesheet-manager.ts`). The point of the WeakSet is also a tighter mechanism than its name implies — it's specifically there to suppress re-entry when the load callback re-calls `serializeNodeWithId` for the same link — and that's now documented by the call site (the recursive serializer follows immediately below the predicate).

Renaming to `linksAwaitingStylesheetLoad` would actually be misleading because the WeakSet also contains links whose timer has already fired, not just ones still awaiting load.

Non-blocking — keeping the existing name.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Each of the three guards handles a distinct case — they're not redundant:

- **`WeakSet`** — suppresses the re-entrant call from the load callback. With `0f73140a` restoring the recursive `serializeNodeWithId` (needed so late-loading stylesheets get `_cssText`), the inner call re-enters `onceStylesheetLoaded` for the same link. Without the WeakSet, that re-entry would schedule a fresh timer + listener and resurrect the leak.
- **`{ once: true }`** — self-cleans the load listener after it fires. If we relied only on `fired`, the listener would sit on the link forever.
- **`fired`** — handles the timer-vs-listener race. The timer fires after the listener was attached `{ once: true }`-style, the listener runs, calls `fire()` — but if it then takes longer than `stylesheetLoadTimeout` and the real `load` event arrives, the listener wouldn't re-fire (it's `once`), but the closure would still call `listener()` again from the timer path. The `fired` flag is what makes the call idempotent across the two paths.

So removing any one of the three reintroduces a real edge-case bug. Closing as intentional.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
pauldambra: > [!NOTE]

Actioned in `0f73140a` — the three tests are now driven by `it.each` over `noListenerLeakCases` and `noMultiplicationCases`, and a new test (`delivers _cssText to onStylesheetLoad when a stylesheet finishes loading after first serialize`) covers the convergent HIGH regression by populating `link.sheet` between first serialize and the dispatched `load`, asserting the delivered serializedNode contains `_cssText`.

packages/browser/playground/preload-link-leak/index.html
pauldambra: > [!NOTE]

Actioned in `0f73140a` — added an HTML comment in `playground/preload-link-leak/index.html` explaining the CSS chunks are intentionally non-existent (that's the precondition the repro depends on).

.changeset/fix-preload-link-listener-leak.md
pauldambra: > [!NOTE]

Non-blocking — leaving as-is. The phrasing is intentional ("multiplying further" continues the verb tense of the preceding clause).

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Checked — no consumer could have been benefiting from the dropped branch.

Per the HTML spec, `<link rel="preload" as="style">` instructs the browser to fetch the bytes into the HTTP preload cache; it does **not** instantiate a `CSSStyleSheet` (`link.sheet` stays `null` forever — that's what made the original code loop). So even before this PR, the preload branch could never reach the `_cssText` extraction in `serializeNode` (`if (stylesheet) { cssText = stringifyStylesheet(stylesheet); }` at L690 — `stylesheet` would always be `null`). The branch existed and looked like it was doing useful work, but it was only scheduling a wait for a sheet that physically cannot exist.

The customer's production trace showed exactly this: 168,467 leaked `addEventListener('load')` calls on a single `<link rel="preload" as="style">` over 585s, all originating from the now-removed branch, none of which ever delivered `_cssText` because `link.sheet` was permanently null.

Closing as intentional and verified.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +414,24 @@ function onceStylesheetLoaded(
...
-  link.addEventListener('load', () => {
-    clearTimeout(timer);
+  stylesheetLoadTracked.add(link);
+  let fired = false;
+  const fire = () => {
+    if (fired) return;
     fired = true;
     listener();
-  });
+  };
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

The lifecycle concern is real (deferring to a follow-up — see #PRRT_kwDODj9d5c6EmL41), but the specific suggested fix — `stylesheetLoadTracked.delete(link)` inside `fire()` — would reintroduce the original leak in the timer-fires-first case:

1. First serialize: `link.sheet` is `null`, add to WeakSet, schedule timer + listener.
2. Timer fires first (sheet still hasn't loaded): `fire()` → with the suggested change, removes link from WeakSet → calls `listener()` → calls `serializeNodeWithId(n, ...)` → re-enters the link branch → `onceStylesheetLoaded` is called again.
3. WeakSet no longer has the link → does **not** return early. `link.sheet` is still `null` so `if (styleSheetLoaded) return;` also doesn't return early. We schedule **a fresh timer + listener** → leak re-armed.

The WeakSet specifically needs to stay populated to suppress that re-entry. The broader lifecycle issue (recorder stop/restart) needs scoping via the snapshot options bag or an explicit reset hook from the recorder — bigger change than a one-line delete and out of scope for this PR.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -584,3 +584,140 @@ describe('link href capture across SPA navigations', () => {
...
+    for (let round = 0; round < 5; round++) {
+      link.dispatchEvent(new Event('load'));
+    }
+    vi.advanceTimersByTime(5000);
+
+    expect(stylesheetLoadCalls).toBeLessThanOrEqual(1);
+
+    document.head.removeChild(link);
+  });
+});
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Actioned in `0f73140a` — extracted `setupLink` + `serializeWithCapture` helpers and the three new tests are now parameterized via `it.each(noListenerLeakCases)` and `it.each(noMultiplicationCases)`. A fourth case (`delivers _cssText to onStylesheetLoad when a stylesheet finishes loading after first serialize`) was added for the convergent HIGH regression discovered in this review pass.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Addressed in `ff5f56d5`. The module-level `stylesheetLoadTracked` WeakSet now resets inside `StylesheetManager.reset()` via a new `resetStylesheetLoadTracking()` export from `@posthog/rrweb-snapshot`, so a recorder stop/restart on the same page no longer skips re-tracking links whose load was still pending in the previous lifecycle.

Added a jsdom test (`re-tracks a pending stylesheet link after resetStylesheetLoadTracking`) that exercises the exact scenario: serialize a pending link in session 1, call `resetStylesheetLoadTracking()`, serialize the same link in session 2, populate `link.sheet`, dispatch `load`, and assert session 2 receives the `_cssText` payload.

Note: the WeakSet still lives in `rrweb-snapshot` rather than being moved into `StylesheetManager`. The cleaner long-term shape is to thread the tracker through `serializeNodeWithId` options so each `snapshot()` call owns its own — out of scope for this PR.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Addressed in `ff5f56d5`. The module-level `stylesheetLoadTracked` WeakSet now resets inside `StylesheetManager.reset()` via a new `resetStylesheetLoadTracking()` export from `@posthog/rrweb-snapshot`, so a recorder stop/restart on the same page no longer skips re-tracking links whose load was still pending in the previous lifecycle.

Added a jsdom test (`re-tracks a pending stylesheet link after resetStylesheetLoadTracking`) that exercises the exact scenario: serialize a pending link in session 1, call `resetStylesheetLoadTracking()`, serialize the same link in session 2, populate `link.sheet`, dispatch `load`, and assert session 2 receives the `_cssText` payload.

Note: the WeakSet still lives in `rrweb-snapshot` rather than being moved into `StylesheetManager`. The cleaner long-term shape is to thread the tracker through `serializeNodeWithId` options so each `snapshot()` call owns its own — out of scope for this PR.

pauldambra (commented):
> [!NOTE]
>
> Incremental review of `ff5f56d5` (Option A: reset hook for `stylesheetLoadTracked`)

Fix correctly addresses the previously deferred cross-lifecycle finding. Reset is wired into `StylesheetManager.reset()`, mirrors the existing `resetMaxDepthState` precedent in the same file, and the new jsdom test genuinely exercises the fix (without the reset, the second-session capture would never fire). xp and security-audit returned no findings.

### Key findings

- 🟢 **LOW (qa-team/reliability)** — `StylesheetManager.reset()` runs from `takeFullSnapshot` on every checkout, not only on recorder stop/restart. Pending links can therefore acquire a second `{ once: true }` listener at each checkout, emitting duplicate `attachLinkElement` mutations when the sheet finally loads. Same `id`, same `_cssText` — not corrupt, just churn. Tracking-ticket worthy.
- ⚪ **NIT (paul + qa-team/compatibility)** — module-level `let stylesheetLoadTracked` is fine because the only reader (`onceStylesheetLoaded`) resolves the binding at call time, and no captured references exist. Unusual for this file; flagging for future readers.
- ⚪ **NIT (paul)** — new test assertion `expect(firstSessionCalls).toBeLessThanOrEqual(1)` is hand-wavy; the actual expected value is `0` since `load` is dispatched only after the reset.
- ⚪ **NIT (qa-team/reliability)** — new cross-lifecycle test only exercises the `load`-event delivery path. Timer fallback is covered by `noMultiplicationCases` but not specifically across reset.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | APPROVE WITH NITS — fix is right; flagging the every-checkout reset as follow-up. |
| 👤 paul | APPROVE WITH NITS — trust on the WeakSet swap; tighten the `firstSessionCalls` assertion. |
| 📐 xp | APPROVE — small, surgical, mirrors `resetMaxDepthState` precedent. No findings. |
| 🛡 security-audit | APPROVE — module-level reset hook only; no new sinks or user-input paths. No findings. |

---

packages/rrweb/rrweb/src/record/stylesheet-manager.ts
```diff
@@ -82,6 +85,7 @@ export class StylesheetManager {
   public reset() {
     this.styleMirror.reset();
     this.trackedLinkElements = new WeakSet();
+    resetStylesheetLoadTracking();
   }
 
   // TODO: take snapshot on stylesheet reload by applying event listener
```
pauldambra, line 91: > [!NOTE]

`StylesheetManager.reset()` is invoked from `takeFullSnapshot` (`record/index.ts:418`) on initial start **and on every checkout snapshot**, not just on stop/restart. After a mid-session checkout, any link element still pending becomes eligible for `onceStylesheetLoaded` to schedule a fresh `setTimeout` + `{ once: true }` listener (because the WeakSet was wiped). The listeners from the previous full snapshot are still attached to the link, so when the sheet finally loads, both listeners fire — each emitting a duplicate `attachLinkElement` mutation for the same `id` with the same `_cssText`.

Not corrupt data (the replayer applies attribute mutations idempotently), but duplicate emit cost on every checkout-pending link. Tracking-ticket follow-up — fix needs a way to abort the previous snapshot's pending listeners on reset, which is bigger than this PR's scope.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -396,12 +396,19 @@ function onceIframeLoaded(
   iframeEl.addEventListener('load', listener);
 }
 
+let stylesheetLoadTracked = new WeakSet<HTMLLinkElement>();
```

**[convergent: paul + qa-team/compatibility]** ⚪ NIT

Module-level `let stylesheetLoadTracked` + an exported `resetStylesheetLoadTracking()` that re-binds it. Works correctly because the only in-module reader (`onceStylesheetLoaded`) resolves the binding by name at call time, and no consumer imports `stylesheetLoadTracked` directly (it's not exported). Flagging only because re-bindable module state is unusual in this file and a future reader might second-guess it.

Non-blocking — the precedent here is `resetMaxDepthState` next door, which uses the same pattern.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -584,3 +585,225 @@ describe('link href capture across SPA navigations', () => {
...
+    } as unknown as CSSStyleSheet);
+
+    link.dispatchEvent(new Event('load'));
+
+    expect(firstSessionCalls).toBeLessThanOrEqual(1);
+    expect(secondSessionDelivered).not.toBeNull();
+    const attrs = ((secondSessionDelivered as unknown as elementNode) ?? {})
+      .attributes;
+    expect(attrs._cssText).toContain('.late');
+    expect(attrs._cssText).toContain('color: blue');
```
pauldambra, line 805: > [!NOTE]

**[paul]** ⚪ NIT

`expect(firstSessionCalls).toBeLessThanOrEqual(1)` is hand-wavy — the actual expected value is `0` because the link's `load` event is dispatched only after `resetStylesheetLoadTracking()` runs, which means the first session's listener and timer are still pending when the assertion runs. Tighten to `expect(firstSessionCalls).toBe(0)`.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -584,3 +585,225 @@ describe('link href capture across SPA navigations', () => {
...
+    expect(deliveredNode).not.toBeNull();
+    const attrs = ((deliveredNode as unknown as elementNode) ?? {}).attributes;
+    expect(attrs).toBeDefined();
+    expect(attrs._cssText).toContain('.a');
+    expect(attrs._cssText).toContain('color: red');
+
+    document.head.removeChild(link);
+  });
+
+  it('re-tracks a pending stylesheet link after resetStylesheetLoadTracking', () => {
```
pauldambra: > [!NOTE]

**[qa-team/reliability]** ⚪ NIT

The new cross-lifecycle test only exercises the `load`-event delivery path. It does not assert that the timer fallback (`vi.advanceTimersByTime(stylesheetLoadTimeout)` without dispatching `load`) also re-arms cleanly after `resetStylesheetLoadTracking()`. Parameterising over both paths (load vs timer) would mirror the structure of `noMultiplicationCases`.

Non-blocking — the load-event path is the canonical case.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Intentional and consistent with the precedent in this file. `resetMaxDepthState` (snapshot.ts:1009) uses the same shape — `let` module state + an exported reset that re-binds. The only reader (`onceStylesheetLoaded` within the same module) resolves the binding by name at call time, and `stylesheetLoadTracked` is not in the package's barrel exports (only `resetStylesheetLoadTracking` is), so no external consumer can hold a stale reference.

Closing as intentional.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
pauldambra: > [!NOTE]

Pinned in `cb7dee41` — but to `toBe(1)`, not `toBe(0)`. The actual value is 1: after `resetStylesheetLoadTracking()`, the first session's `{ once: true }` load listener is still attached to the link element (reset only clears the dedup tracker, not the listeners themselves). When the sheet finally loads, both session 1's and session 2's listeners fire, so `firstSessionCalls` becomes 1.

The duplicate fire is exactly the follow-up concern in the sibling thread on `stylesheet-manager.ts:91` — the test now documents it explicitly.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
pauldambra: > [!NOTE]

Fair point — the cross-lifecycle test only covers the load-event path. The timer-fallback re-arm is structurally identical (same WeakSet check at the top of `onceStylesheetLoaded`), so a separate test row would mostly assert the same code path with a different trigger. The `noMultiplicationCases` table already exercises the timer path for the non-reset case.

Keeping as-is — non-blocking, marginal coverage gain.

packages/rrweb/rrweb/src/record/stylesheet-manager.ts
pauldambra: > [!NOTE]

Intentional and consistent with the precedent in this file. `resetMaxDepthState` (snapshot.ts:1009) uses the same shape — `let` module state + an exported reset that re-binds. The only reader (`onceStylesheetLoaded` within the same module) resolves the binding by name at call time, and `stylesheetLoadTracked` is not in the package's barrel exports (only `resetStylesheetLoadTracking` is), so no external consumer can hold a stale reference.

Closing as intentional.

pauldambra: > [!NOTE]

Ignore the previous reply on this thread — wrong body landed by mistake.

Deferring this LOW finding to a follow-up. The reset on every checkout (rather than only on recorder stop) means each pending link can accumulate one extra `{ once: true }` listener per checkout. The test at `snapshot.test.ts:805` now pins this behavior with `expect(firstSessionCalls).toBe(1)` in `cb7dee41`, so the cost is documented and asserted, not hidden. Fix needs a mechanism to abort the previous snapshot's pending listeners on reset (e.g. an `AbortSignal` per snapshot lifecycle, or a Map<link, {timer,listener}> that reset can iterate over) — bigger than this PR's scope and converges back toward the rrweb upstream #1708 approach we explicitly chose not to take here.

Leaving unresolved as a deferred follow-up.

pauldambra: > [!NOTE]

Fixed in `3ed123d1`. The tracker is now `Map<HTMLLinkElement, { timer, onLoad }>`; `resetStylesheetLoadTracking()` iterates pending entries, `clearTimeout` each timer, `removeEventListener('load', onLoad)` each named handler, then clears the Map. So when `StylesheetManager.reset()` fires (start or any checkout), the previous full snapshot's pending watches are torn down before the new one re-tracks — no duplicate `attachLinkElement` mutations.

Each watch self-removes from the Map after `fire()` (wrapped in `try/finally` so a throwing listener still cleans up). Stays inside the existing browser-support matrix (no `AbortController`).

The cross-lifecycle test's `expect(firstSessionCalls).toBe(1)` is now `toBe(0)` — that's the documented behavior: the previous session's listener really is gone after reset.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +419,29 @@ function onceStylesheetLoaded(
...
+    fired = true;
+    try {
       listener();
-      fired = true;
+    } finally {
+      stylesheetLoadTracked.delete(link);
+      controller.abort();
     }
-  }, styleSheetLoadTimeout);
+  };
```
pauldambra, line 434: > [!NOTE]

**[paul]** ⚪ NIT

The `controller.abort()` inside `fire()`'s `finally` is the channel for tearing down the load listener (`{ signal }`-bound). The timer is already cleared by then — either it just fired (we're inside its callback) or `clearTimeout` is a no-op because `load` fired first. So `abort()` is doing one real job (removing the load listener) and one redundant job (re-triggering the abort handler that already-ran clearTimeout).

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +419,29 @@ function onceStylesheetLoaded(
 
   if (styleSheetLoaded) return;
 
-  const timer = setTimeout(() => {
-    if (!fired) {
+  const controller = new AbortController();
```
pauldambra, line 422: > [!NOTE]

**[paul]** ⚪ NIT

Same concern shape as the original rrweb #1708 PR (aborting previous snapshot's work when a new one starts). The reasoning for why it's safe here: `reset()` is only called from `takeFullSnapshot` (`record/index.ts:418`) and is immediately followed by `snapshot(document, …)` which re-walks the DOM and schedules a fresh wait for every still-pending link. Mirror id is preserved across checkouts, so the deferred mutation goes against the same node id. Lost mutation only if the new snapshot walk skips the link (e.g. block selector, depth cap, slimDOM removing it) — not protected by the current cross-lifecycle test.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +419,29 @@ function onceStylesheetLoaded(
```
pauldambra, line 419: > [!NOTE]

**[xp]** ⚪ NIT

A `cleanup()` closure that did `clearTimeout(timer); link.removeEventListener('load', fire)` and was called from both `fire()` and `resetStylesheetLoadTracking()` would be more direct than routing through `AbortController` + abort-event + signal-bound listener.

Trade-off: the previous shape (3ed123d1) was exactly that — `Map<HTMLLinkElement, { timer, onLoad }>` with manual `clearTimeout` + `removeEventListener` in reset. b74d3926 deliberately moved to the `AbortController` shape for declarative lifecycle. Both work; this version is slightly more indirect but the lifecycle is expressed in one object.

pauldambra (commented):
> [!NOTE]
>
> Incremental review of `b74d3926` (Map<…, {timer, onLoad}> → Map<…, AbortController>)

3ed123d1 + b74d3926 together resolve the previous LOW-severity duplicate-mutation concern. Pending watches now carry an `AbortController`, `resetStylesheetLoadTracking()` aborts each, and the `{ signal }`-bound load listener is removed declaratively. Test assertion flip from `1` to `0` documents the desired behaviour (previous session's listener is genuinely gone after reset).

### Findings (all NIT)

- ⚪ **paul** — `controller.abort()` in `fire()`'s `finally` reads as belt-and-braces; worth confirming it's intentional.
- ⚪ **paul** — module-level mutable state + `reset()` aborting in-flight waits has the #1708 concern shape; flagged for awareness.
- ⚪ **xp** — using `AbortController` for cleanup is indirect compared to a `cleanup()` closure that does `clearTimeout + removeEventListener`. Trade-off, not a bug.
- ⚪ **xp** — `try/finally` intent ("always tear down even if listener throws") could be named via extracted `cleanup` closure.
- ⚪ **paul + xp (convergent)** — test name `re-tracks a pending stylesheet link after resetStylesheetLoadTracking` is fine; a more behaviour-focused name ("does not invoke first-session listener after reset") would protect regression intent.
- ⚪ **paul** — changeset addition for `@posthog/rrweb`: positive note.

qa-team and security-audit: no findings.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | APPROVE — refactor is sound, semantics correct, AbortController scoping right. |
| 👤 paul | APPROVE WITH NITS — high trust, small change, ship it. |
| 📐 xp | APPROVE WITH NITS — refactor is a genuine lifecycle simplification; nits are about expressiveness. |
| 🛡 security-audit | APPROVE — internal helper, no new sinks, no new untrusted input. |

---

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +419,29 @@ function onceStylesheetLoaded(
 
   if (styleSheetLoaded) return;
 
-  const timer = setTimeout(() => {
-    if (!fired) {
+  const controller = new AbortController();
+  let fired = false;
+  const fire = () => {
+    if (fired) return;
```
pauldambra, line 425: > [!NOTE]

**[xp]** ⚪ NIT

The `try/finally` around `listener()` is load-bearing — guarantees `stylesheetLoadTracked.delete(link)` and `controller.abort()` happen even if the listener throws. Extracting a `cleanup` closure would name the intent.

Leaving as-is: the `finally` block is two lines and the intent reads from the structure (`try { listener() } finally { /* always clean up */ }`).

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
```diff
@@ -584,3 +585,225 @@ describe('link href capture across SPA navigations', () => {
...
+    ];
+    setSheet({
+      cssRules: fakeRules as unknown as CSSRuleList,
+      rules: fakeRules as unknown as CSSRuleList,
+      href: null,
+    } as unknown as CSSStyleSheet);
+
+    link.dispatchEvent(new Event('load'));
+
+    expect(firstSessionCalls).toBe(0);
```
pauldambra, line 800: > [!NOTE]

A more behaviour-focused test name (e.g. `does not invoke previous-session listener after resetStylesheetLoadTracking`) would protect the regression intent more clearly than the current `re-tracks a pending stylesheet link after resetStylesheetLoadTracking`. Worth a rename.

.changeset/fix-preload-link-listener-leak.md
```diff
@@ -0,0 +1,7 @@
+---
+'posthog-js': patch
+'@posthog/rrweb-snapshot': patch
+'@posthog/rrweb': patch
```
pauldambra, line 4: > [!NOTE]

**[paul]** ⚪ NIT (positive)

Good catch adding `@posthog/rrweb` — the fix touches `rrweb-snapshot` which `rrweb` depends on (and `record/stylesheet-manager.ts` now imports `resetStylesheetLoadTracking`), so the patch bump propagating is correct.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Reading right. `controller.abort()` in the `finally` is the only mechanism that removes the `{ signal }`-bound load listener. The timer is already settled (we're either inside the timer callback or `clearTimeout` is a no-op because `load` fired first). So `abort()` does one real job (auto-remove the load listener via signal) and one belt-and-braces job (re-fires the abort handler that idempotently clears the already-cleared timer).

Closing as intentional.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Same shape, different context (we worked through this earlier in the PR thread). `reset()` is only called from `takeFullSnapshot` (`record/index.ts:418`) and is followed synchronously by `snapshot(document, …)` which re-walks the DOM and schedules a fresh wait for every still-pending link. Mirror id is preserved across checkouts so the deferred mutation goes against the same node id.

The "lost mutation only if the new walk skips the link" edge case (block selector, depth cap, slimDOM) is real but isn't introduced by this PR — pre-fix it would silently emit a duplicate-but-undeliverable mutation in the same edge case.

Closing as intentional.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Fair observation. We tried that exact shape in `3ed123d1` (`Map<HTMLLinkElement, { timer, onLoad }>` + manual `clearTimeout` + `removeEventListener` in reset). `b74d3926` deliberately moved to `AbortController` to express the watch's lifetime as a single object — `controller.abort()` becomes the one verb for "this watch is done", and the signal-bound listener teardown is declarative.

Both work. Sticking with `AbortController` because the lifetime is now a first-class thing rather than two parallel handles.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Leaving as-is. The `try/finally` is two lines, the intent reads from the structure, and extracting a `cleanup` closure would add a named helper used in exactly one place — that's the wrong direction on Rule 4. The `controller.abort()` line is the load-bearing teardown; `delete()` is the bookkeeping; both are obvious in context.

packages/rrweb/rrweb-snapshot/test/snapshot.test.ts
pauldambra: > [!NOTE]

Renamed in `79af5f76` to `resetStylesheetLoadTracking tears down the previous session listener and re-tracks the link for the new session`. Leads with the behaviour the test guards.

.changeset/fix-preload-link-listener-leak.md
pauldambra: > [!NOTE]

Thanks — resolving.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -411,18 +427,25 @@ function onceStylesheetLoaded(
...
+  let timer: ReturnType<typeof setTimeout>;
+  const fire = () => {
+    if (fired) return;
+    fired = true;
+    try {
       listener();
-      fired = true;
+    } finally {
+      stylesheetLoadTracked.delete(link);
     }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

This analysis applies to the previous shape (`3ed123d1`, `Map<HTMLLinkElement, { timer, onLoad }>` with a named `onLoad` and no signal binding) — not the current `b74d3926` code.

In `b74d3926`:
- The load listener is bound with `{ signal: controller.signal, once: true }` (line 439-442).
- `controller.abort()` in the `finally` (line 431) aborts the signal.
- Aborting the signal causes the browser to auto-remove **all** listeners bound with that signal — including the load listener.

So even if `listener()` throws, the `finally` still runs `delete()` + `abort()`, and `abort()` removes the load listener. No leak.

The suggested fix `link.removeEventListener('load', onLoad)` wouldn't compile against the current code anyway (there's no named `onLoad` — we pass `fire` directly to `addEventListener`).

Closing — the bot is analyzing a stale revision.

fasyy612 (approved):
Sweet find, LGTM!

PR #3668, opened 2026-05-25
feat(ai): expose `token` in PostHogTraceExporter and deprecate `apiKey`

rafaeelaudibert:
This update introduces a new `token` parameter for the PostHogTraceExporter, aligning with our goal of standardizing terminology. The `apiKey` parameter is still supported but marked as deprecated. Adjustments have been made across examples and tests to reflect this change, ensuring backward compatibility while encouraging the use of `token` moving forward.

### DIFF
--- a/.changeset/grumpy-heads-accept.md
+++ b/.changeset/grumpy-heads-accept.md
@@ -1,0 +1,7 @@
+---
+'@posthog/ai': patch
+---
+
+Expose `token` in addition to `apiKey` for `PostHogTraceExporter`
+
+We're moving towards a place where we're calling this `token` everywhere. Let's release this as a patch version because it's a tiny change and we still support the old usage. This might be dropped in the next packages/ai major version.
--- a/.changeset/grumpy-heads-accept.md
+++ b/.changeset/grumpy-heads-accept.md
@@ -2,6 +2,6 @@
 '@posthog/ai': patch
 ---
 
-Expose `token` in addition to `apiKey` for `PostHogTraceExporter`
+Expose `projectToken` in addition to `apiKey` for `PostHogTraceExporter`
 
-We're moving towards a place where we're calling this `token` everywhere. Let's release this as a patch version because it's a tiny change and we still support the old usage. This might be dropped in the next packages/ai major version.
+We're moving towards a place where we're calling this `projectToken` everywhere. Let's release this as a patch version because it's a tiny change and we still support the old usage. This might be dropped in the next packages/ai major version.
--- a/packages/ai/src/otel/exporter.ts
+++ b/packages/ai/src/otel/exporter.ts
@@ -15,24 +15,32 @@ function normalizeHost(value?: unknown): string {
   return normalizedValue || DEFAULT_OTEL_HOST
 }
 
-export interface PostHogTraceExporterOptions {
-  /**
-   * Your PostHog project API token.
-   */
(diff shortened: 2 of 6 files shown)


packages/ai/src/otel/exporter.ts
```diff
@@ -16,10 +16,17 @@ function normalizeHost(value?: unknown): string {
 }
 
 export interface PostHogTraceExporterOptions {
+  /**
+   * Your PostHog project API token.
+   */
+  token?: string
```
marandaneto:
```suggestion
  /**
   * Your PostHog project API token.
   */
  projectToken?: string
```

we call it `projectToken` as the canonical form in some other SDKs
`token` was already before `apiKey` and some SDKs even have the `token` as deprecated and/or removed
`projectToken` has semantics about its scope, so it makes more sense imo

.changeset/grumpy-heads-accept.md
```diff
@@ -0,0 +1,8 @@
+---
+'@posthog/convex': patch
```
richardsolomou: @posthog/convex shouldn't be patched for this, there's no code changes for this package

richardsolomou (changes requested):

marandaneto:
have to fix docs/onboarding as well

rafaeelaudibert:
Docs will be fixed automatically by Inkeep, and I will be able to approve it, won't I? @marandaneto

.changeset/grumpy-heads-accept.md
rafaeelaudibert: I changed some example files and I thought it made sense to ship a new version with the changes, but I'll remove it

packages/ai/src/otel/exporter.ts
rafaeelaudibert: This is reasonable, thank you

richardsolomou: Should we be changing this to reflect this everywhere? I recently made a change to docs to change VITE_PUBLIC_POSTHOG_API_KEY to VITE_POSTHOG_TOKEN. Should that be changed to VITE_POSTHOG_PROJECT_TOKEN?

marandaneto: ideally yes but only if theres an easy migration path (deprecate the old one, add the new one, etc)
sometimes its not that easy because of positional/named params and then it requires a breaking change and ideally a major version
we dont want to add a new major just for renaming one single property
so once we have a few stacked breaking changes for a certain sdk, then we do this at once
eg https://github.com/PostHog/posthog-dotnet/pull/183

packages/ai/src/otel/exporter.ts
```diff
@@ -16,10 +16,17 @@ function normalizeHost(value?: unknown): string {
...
+   */
+  projectToken?: string
+
   /**
    * Your PostHog project API key.
+   *
+   * @deprecated Use `projectToken` instead
    */
-  apiKey: string
+  apiKey?: string
```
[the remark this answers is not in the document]
richardsolomou: @rafaeelaudibert Agree with the robot on this, would be nice to make this typesafe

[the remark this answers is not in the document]
rafaeelaudibert: Done!

richardsolomou (approved):

PR #3669, opened 2026-05-25
chore: drop IE11 from browserslist, keep ES5 bundle as polyfill canary

pauldambra:
## Summary

- Remove `IE 11` from `packages/browser/package.json#browserslist` — we don't actually support IE11.
- Keep the ES5 bundle (`array.full.es5.js`) build and the `es-check` workflow intact. Babel `preset-env` targets for the ES5 build are hard-coded in `rollup.config.mjs` (still include `'IE 11'`), so the ES5 polyfill set is unchanged. The ES5 bundle is now a canary: `es-check` failing on it tells us we've introduced something the polyfill set can't cover yet.
- Fix two latent TDZ bugs in tests (`web-vitals.test.ts`, `posthog-core-also.test.ts`). They referenced `let mockX: jest.Mock` from inside `jest.mock(...)` factories, which jest hoists above the declaration. Babel was previously transpiling `let` -> `var` (because IE 11 was in browserslist), masking the bug. Switching the two declarations to `var` — the canonical jest workaround for out-of-scope mock references — is the actual root cause of the CI failure the previous attempt hit.

## Test plan

- [ ] `pnpm lint` passes
- [ ] `pnpm build` succeeds for `packages/browser`
- [ ] `ES5 support check` workflow (es-check es5 on `array.full.es5.js`) is green — this is the IE11 bundle test we explicitly want to keep validating
- [ ] `ES6 support check` (es-check es6 on `array.full.js`) is green
- [ ] `pnpm --filter posthog-js test:unit` passes (modulo the pre-existing flaky perf assertion in `logs.test.ts`)
- [ ] `library-ci` compat playwright job behaves the same as on main (non-blocking)

---


pauldambra:
* **#3669** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3669?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3669?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

.github/workflows/testcafe.yml
```diff
@@ -58,6 +58,7 @@ jobs:
           BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
           RUN_ID: ${{ github.run_id }}
           BROWSER: ${{ matrix.browser }}
+          BROWSERSLIST: "> 0.5%, last 2 versions, Firefox ESR, not dead, IE 11"
```
pauldambra, line 61: this is what was missing when previous attempts to do this lead to these tests freezing

dustinbyrne (approved):

PR #3673, opened 2026-05-26
chore(react-native): bump posthog-react-native-session-replay peer dep to 1.6.0

turnipdabeets:
## Summary

Bumps the optional peer dependency `posthog-react-native-session-replay` from `>= 1.5.8` to `>= 1.6.0`.

`1.6.0` ([release notes](https://github.com/PostHog/posthog-react-native-session-replay/releases/tag/1.6.0)) adds an opt-in path that resolves `posthog-ios` through Swift Package Manager. Consumers who want to engage it set `"posthog.useSpm": "true"` in `ios/Podfile.properties.json` and enable `use_frameworks! :linkage => :dynamic`. Default consumer behavior is unchanged: without the property, `posthog-ios` continues to resolve through CocoaPods.

This is the React Native side of the broader [PostHog/posthog-ios#472](https://github.com/PostHog/posthog-ios/issues/472) effort to give consumers an SPM path ahead of the [CocoaPods trunk read-only date on 2026-12-02](https://blog.cocoapods.org/CocoaPods-Specs-Repo/).

## What this PR does NOT change

- **No iOS-side code in this repo.** `packages/react-native` has no `.podspec` and no native iOS sources. `posthog-ios` only enters an RN consumer's dependency tree through session-replay's podspec, so the entire CocoaPods-deprecation question reduces to this single peer-dep bump.
- **No example app SPM opt-in.** `examples/example-expo-53` continues to demonstrate the default consumer experience (static linkage, CocoaPods resolution). The SPM opt-in is already exercised by session-replay's own `build-ios-spm` CI job; duplicating it here would change what the example demonstrates without adding useful coverage.

## Files changed

- `packages/react-native/package.json` — peerDep + devDep bumped to `>= 1.6.0` / `^1.6.0`
- `examples/example-expo-53/package.json` — explicit pin bumped to `^1.6.0` (caret semver already resolved 1.6.0, but explicit-is-better-than-implicit)
- `pnpm-lock.yaml` + `examples/example-expo-53/pnpm-lock.yaml` — refreshed
- `.changeset/bump-session-replay-1.6.0.md` — new, patch bump on `posthog-react-native`

## Lockfile noise

The root `pnpm-lock.yaml` diff is larger than the change deserves (~517 lines of changes unrelated to session-replay). pnpm normalized accumulated drift in `@types/node` peer resolutions across `ts-node`/`jest`/etc. while it was rewriting the lockfile. The session-replay-specific changes themselves are only 5 lines. The unrelated normalizations would happen on any install touching the lockfile, not caused by this change.

## Test plan

- [ ] CI passes
- [ ] After merge, the new `posthog-react-native` patch release flows out via the changeset workflow
- [ ] Spot-check that example-expo-53 still builds with the bumped session-replay (the version range is satisfied by `^1.5.8`'s caret semver already, so this should be a no-op functionally)


.changeset/bump-session-replay-1.6.0.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': patch
```
marandaneto: prob a minor since it feels like a new feat?

marandaneto (approved):

PR #3677, opened 2026-05-27
feat(react-native): add native crash capture and debug symbol upload

ioannisj:
## Problem

The React Native SDK only autocaptures JS-side errors today. Native iOS and Android crashes — which happen below the JS layer — never reach PostHog. This PR lets users opt in to native crash capture through the optional native plugin, alongside session replay.

The companion package `posthog-react-native-session-replay` is superseded by **`@posthog/react-native-plugin`** (now in this monorepo, published as `2.0.1`): it keeps the same session-replay surface and adds native error tracking behind a unified `setup()` entry point. This PR wires the SDK to the new plugin while keeping the legacy package supported as a fallback.

### Runtime — native crash capture
- New `errorTracking.autocapture.nativeCrashes?: boolean` (on `AutocaptureOptions`). When `true`, the SDK forwards native crash autocapture to the iOS/Android PostHog SDKs via the optional plugin. Off by default, and additionally gated server-side by the project's Error tracking **Autocapture** setting.
- The optional plugin loader now prefers `@posthog/react-native-plugin`, falling back to `posthog-react-native-session-replay`. Renamed `src/optional/OptionalSessionReplay.ts` → `OptionalPlugin.ts` (and its exported identifiers) since it now loads either plugin.
- New-plugin path calls a unified `setup(sessionId, sdkOptions, { sessionReplay, errorTracking: { nativeAutocapture } })`; the legacy path keeps `start(...)` and warns if native crashes were requested.
- Session management (`getSessionId` / `resetSessionId` / `identify` / …) now activates when **either** replay or error tracking is initialized, with lazy replay toggling to avoid a redundant `setup()`.

### Build tooling — native symbol upload (Expo)
Native crash stack traces are unreadable without uploaded symbols, so the Expo config plugin gains an opt-in `uploadNativeSymbols` option. On `expo prebuild` it wires both platforms:
- **iOS** — a dSYM-upload Run Script phase running posthog-ios's `upload-symbols.sh`.
- **Android** — the `com.posthog.android` Gradle plugin (R8 mapping upload + matching map-id).
- `uploadNativeSymbols: { includeSource: true }` also uploads native source for crash context (**iOS only**).
- **Bare RN** is manual by design (documented dSYM phase + `com.posthog.android` classpath/apply) — a bare app's native projects are developer-owned, with no posthog-controlled generation step to inject into.

### Example
- `examples/example-expo-53`: enabled `nativeCrashes`, set `uploadNativeSymbols`, and added a "Trigger native crash" button backed by a local `native-crash` Expo module for end-to-end testing.

### Local dev wiring
- Added `@posthog/react-native-plugin` as an optional `peerDependency` + `devDependency` of `posthog-react-native`.

### Tests
- `test/native-error-tracking.spec.ts` (no flags → no init; `nativeCrashes` only → `setup()` with replay off + native on; both → both on) and `test/expoconfig.spec.ts` for the symbol-upload wiring helpers.
-  Manual testing with sample apps
CleanShot 2026-06-11 at 07 05 02@2x

CleanShot 2026-06-10 at 22 54 47@2x

### Libraries affected

- [x] posthog-react-native

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes!) — existing `posthog-react-native-session-replay` consumers are unaffected; both packages are supported, with the new plugin preferred when both are installed.

### If releasing new changes


packages/react-native/src/error-tracking/index.ts
```diff
@@ -22,6 +16,11 @@ interface AutocaptureOptions {
 
 export interface ErrorTrackingOptions {
   autocapture?: AutocaptureOptions | boolean
+  /**
+   * Enables native iOS/Android error autocapture through the optional native plugin.
+   * Disabled by default. Requires `posthog-react-native-plugin` installed.
+   */
+  autocaptureNative?: boolean
```
marandaneto: this feels starnge because we have autocapture just above
should this be part of `AutocaptureOptions` instead?

ioannisj: Makes sense, will address it when testing out the public api

packages/react-native/src/error-tracking/index.ts
```diff
@@ -22,6 +16,11 @@ interface AutocaptureOptions {
 
 export interface ErrorTrackingOptions {
   autocapture?: AutocaptureOptions | boolean
+  /**
+   * Enables native iOS/Android error autocapture through the optional native plugin.
+   * Disabled by default. Requires `posthog-react-native-plugin` installed.
+   */
```
marandaneto: this will require uploading debug symbols for ios and source maps for android, which means 
this https://github.com/PostHog/posthog-js/blob/main/packages/react-native/tooling/posthog-xcode.sh
and https://github.com/PostHog/posthog-js/blob/main/packages/react-native/tooling/posthog.gradle or https://github.com/PostHog/posthog-js/tree/main/packages/react-native/src/tooling needs to change to install the android plugin which is not yet addressed in this pr

ioannisj: handled this by updating the Expo plugin (src/tooling/expoconfig.ts with a new opt-in uploadNativeSymbols option that auto-wires xcode build phase and gradle plugin. 
Could not find a way to do the same for bare RN though. Looks like posthog-xcode.sh runs before dsyms are available? So for those users, it will have to be the same manual steps that native users need to take. Working on the docs right now

marandaneto: > Looks like posthog-xcode.sh runs before dsyms are available? So for those users, it will have to be the same manual steps that native users need to take. Working on the docs right now

its possible to automate this via expo plugins, since now the lifecycle is different
we probably need to run twice

marandaneto: and if it has to be manual, we could next step add support in the posthog wizard to do all manual steps for us, not a blocker for this PR tho, docs is a good 1st interaction

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -33,6 +38,74 @@ const withAndroidPlugin = (config: any) => {
...
+    if (config.modResults.language !== 'groovy') {
+      console.warn('Cannot configure the PostHog Android Gradle plugin because the project build.gradle is not groovy')
+      return config
+    }
+    config.modResults.contents = addPostHogAndroidGradlePluginClasspath(config.modResults.contents)
+    return config
+  })
+
+  return withAppBuildGradle(config, (config: any) => {
+    if (config.modResults.language !== 'groovy') {
```
marandaneto, line 134: we will probably need kotlin (kts) support very soon

marandaneto:
iirc there are cases that if an error is thrown in JS and crashes the app, the native SDK also captures the error, so its capturing twice and we'd need to double check for that
i dont know if the newer RN versions still do that, worth checking
https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L355

packages/react-native/src/error-tracking/index.ts
ioannisj: yeah I think for v1 I'll stick to manual steps for bare RN and just document (it's the same steps native users need to take anw)

packages/react-native/src/tooling/expoconfig.ts
ioannisj: Do you feel this is a blocker for the release or can this be a follow up?

ioannisj:
> iirc there are cases that if an error is thrown in JS and crashes the app, the native SDK also captures the error, so its capturing twice and we'd need to double check for that i dont know if the newer RN versions still do that, worth checking [getsentry/sentry-react-native@`f170ec3`/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L355](https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/android/src/main/java/io/sentry/react/RNSentryStart.java#L355)

Looks like the fix should land on native SDKs. Created these two issues for a followup 
https://github.com/PostHog/posthog-ios/issues/653
https://github.com/PostHog/posthog-android/issues/567

marandaneto:
> PostHog/posthog-android#567

this should be addressed in this SDK before release as a quick workaround https://github.com/PostHog/posthog-android/issues/567#issuecomment-4681823793 similar here https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/ios/RNSentryStart.m#L213-L219
we dont want to bill people twice for every fatal event

ioannisj:
> > PostHog/posthog-android#567
> 
> 
> 
> this should be addressed in this SDK before release as a quick workaround https://github.com/PostHog/posthog-android/issues/567#issuecomment-4681823793 similar here https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/ios/RNSentryStart.m#L213-L219
> 
> we dont want to bill people twice for every fatal event

Yeah that could work as a workaround or failsafe, but will require a 2.0.1 of the react-native-plugin.

packages/react-native/src/posthog-rn.ts
```diff
@@ -1886,18 +1898,48 @@ export class PostHog extends PostHogCore {
...
+    ) {
+      return true
+    }
+
+    // The native SDKs can't be re-initialized — a second setup() would reset the running
+    // instance. If error tracking is already running, skip setup() and start replay on the
+    // existing native instance instead (setup() is what would otherwise start it).
+    if (this._isNativePluginInitialized() && enableSessionReplay && !this._sessionReplayNativeInitialized) {
+      this._sessionReplayNativeInitialized = true
+      await OptionalReactNativePlugin.startRecording?.(true)
```
turnipdabeets, line 1936: Wouldn't this start replay on a native instance that never got the replay config? When error tracking initializes first, setup() runs with sessionReplay.enabled: false — and both native impls only apply sdkReplayConfig (screenshot mode, masking, snapshot endpoint) when enabled is true. So when replay is toggled on later via startRecording(), it does start, but with the wrong settings: wireframe mode instead of screenshot mode (broken for RN), default masking instead of the user's, default endpoint. Related: https://github.com/PostHog/posthog-js/pull/3677#discussion_r3394089404

  The JS side already sends sdkReplayConfig/decideReplayConfig even when replay is disabled, so the cleanest fix is in the plugin package: have setup() always apply the replay config, and use enabled only to decide whether to start recording.

packages/react-native/src/posthog-rn.ts
```diff
@@ -2008,27 +2050,62 @@ export class PostHog extends PostHogCore {
...
+        }
       }
       this._currentSessionId = sessionId
-      this._sessionReplayNativeInitialized = true
+      if (enableSessionReplay) {
+        this._sessionReplayNativeInitialized = true
+        this._logger.info(`Session replay started with sessionId ${sessionId}.`)
+      }
+      if (enableNativeErrorTracking) {
+        this._nativeErrorTrackingInitialized = true
```
turnipdabeets, line 2105: If a user has the old posthog-react-native-session-replay package (no setup()) and asks for nativeCrashes, we correctly warn "Native error tracking is not available" — the legacy plugin can't do crash capture. But then at the bottom of the function we still set _nativeErrorTrackingInitialized = true and log "Native error tracking started." So the SDK now believes crash capture is running when it isn't, and the logs say both "not available" and "started" for the same feature.

Fix: in the legacy branch, after the warning, set enableNativeErrorTracking = false so the flag and the success log are skipped.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -33,6 +37,77 @@ const withAndroidPlugin = (config: any) => {
...
+  config = withProjectBuildGradle(config, (config: any) => {
+    if (config.modResults.language !== 'groovy') {
+      console.warn('Cannot configure the PostHog Android Gradle plugin because the project build.gradle is not groovy')
+      return config
+    }
+    config.modResults.contents = addPostHogAndroidGradlePluginClasspath(config.modResults.contents)
+    return config
+  })
+
+  return withAppBuildGradle(config, (config: any) => {
```
turnipdabeets, line 133: These two Gradle edits need to be coupled: the classpath line (downloads the plugin) and the `apply plugin` line (uses it). Right now, if the classpath insertion is skipped — kts build.gradle, or no `buildscript { dependencies {` block — we warn "skipping… Native symbols will not be uploaded" but **still inject `apply plugin: "com.posthog.android"` into the app module**. Gradle then can't resolve the plugin and the whole Android build fails with `Plugin with id 'com.posthog.android' not found`

Suggestion: have `addPostHogAndroidGradlePluginClasspath` report whether it actually inserted the line, and only run `applyPostHogAndroidGradlePlugin` when it did. Then the failure mode matches the warning: symbols skipped, build still green.

Related: the `/dependencies\s*\{/` search runs from `buildscript {` to end of file, so on a buildscript with no dependencies block the classpath line can land in a *later* block (e.g. `allprojects`) — also a build break. Bounding the search to the buildscript block body fixes that too.

packages/react-native/test/native-error-tracking.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+    const { PostHog } = await import('../src/posthog-rn')
+    const posthog = new PostHog('test-token', {
+      persistence: 'memory',
+      flushInterval: 0,
+      errorTracking: { autocapture: { nativeCrashes: true } },
+    })
+
+    await posthog.ready()
+
+    await waitForExpect(100, () => {
```
turnipdabeets, line 64: The happy paths are well covered. Two gaps worth closing because a regression could ship silently:

1. No test mocks a legacy plugin without `setup` — the entire fallback branch in `initializeNativePlugin` (the `start()`/`isEnabled()` flow existing session-replay-package users depend on, plus the new "nativeCrashes with old plugin → warn" behavior) runs in zero tests after this refactor.
2. The linkedFlag-gated branch (`enableSessionReplay: true`, cached flag resolves off, `nativeCrashes: true` → error-tracking-only `setup()`) is untested — deleting that if-block in `startSessionReplay` passes the whole suite.

packages/react-native/src/posthog-rn.ts
ioannisj: Right, this is indeed a bug, config should be applied unconditionally in `react-native-plugin` and only gate `sessionReplay = true`. Working on a PR for that plugin to address this workaround https://github.com/PostHog/posthog-js/pull/3677#issuecomment-4682084736, I'll add this there

packages/react-native/src/posthog-rn.ts
ioannisj: Fixed

packages/react-native/src/tooling/expoconfig.ts
ioannisj: Done

packages/react-native/src/tooling/expoconfig.ts
marandaneto: not a blocker

packages/react-native/src/posthog-rn.ts
ioannisj: Addressed in https://github.com/PostHog/posthog-js/pull/3824, not we gate only `config.sessionReplay`
Tested this out on Android sample project

ioannisj:
> > [PostHog/posthog-android#567](https://github.com/PostHog/posthog-android/issues/567)
> 
> this should be addressed in this SDK before release as a quick workaround [PostHog/posthog-android#567 (comment)](https://github.com/PostHog/posthog-android/issues/567#issuecomment-4681823793) similar here [getsentry/sentry-react-native@`f170ec3`/packages/core/ios/RNSentryStart.m#L213-L219](https://github.com/getsentry/sentry-react-native/blob/f170ec355cd622d3e935478a23660ad4cb97e896/packages/core/ios/RNSentryStart.m#L213-L219) we dont want to bill people twice for every fatal event

Addressed in https://github.com/PostHog/posthog-js/pull/3824
Unfortunately, workaround works for Android (old + new arch) and iOS (old arch only). Tested and verified btw on sample expo app. 

For the new architecture, we get only a SIGABRT from plcarshreporter, so even native side we'll probably need a little bit more work to catch this case

I'll wait for the plugin to publish, then update the min version on this PR. 

With this + fix CI below, we should be good to go

packages/react-native/src/tooling/expoconfig.ts
ioannisj: https://github.com/PostHog/posthog-js/issues/3826

packages/react-native/src/optional/OptionalPlugin.ts
```diff
@@ -0,0 +1,30 @@
...
+  setup?: (sessionId: string, sdkOptions: { [key: string]: any }, pluginConfig: { [key: string]: any }) => Promise<void>
+  startRecording?: (resumeCurrent: boolean) => Promise<void>
+  stopRecording?: () => Promise<void>
+}
+
+export let OptionalReactNativePlugin: PostHogReactNativePluginExtended | undefined = undefined
+
+if (Platform.OS !== 'macos' && Platform.OS !== 'web') {
+  try {
+    OptionalReactNativePlugin = require('@posthog/react-native-plugin')
```
marandaneto, line 22: error tracking should work for `macos` right

examples/example-expo-53/package.json
```diff
@@ -14,6 +14,7 @@
     "dependencies": {
         "@expo/metro-runtime": "~5.0.4",
         "@expo/vector-icons": "^14.1.0",
+        "@posthog/react-native-plugin": "*",
```
marandaneto, line 17: i think we can kill all calls from `posthog-react-native-session-replay` here now in the sample

marandaneto (approved):
left 2 comments otherwise LGTM

packages/react-native/src/optional/OptionalPlugin.ts
ioannisj: plugin is iOS only today (because of legacy plugin and session replay). So not a one line change. I'll do as a follow-up though since now with error tracking, we should change that

ioannisj: https://github.com/PostHog/posthog-js/issues/3827

marandaneto: good point!

ioannisj:
Did a last manual test since things have changed since I last opened. Signal is good, merging

PR #3678, opened 2026-05-27
fix(replay): fix absolute position bug in ph-no-capture

ksvat:
## Problem

https://posthoghelp.zendesk.com/agent/tickets/59042 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/60912)  
`ph-no-capture` elements come back during replay with `position: absolute` + inline `left/top` regardless of their original positioning. In-flow elements (e.g. flex children, paragraph spans) get pulled out of flow and sibling layout collapses.

## Changes

- snapshot.ts: when blocking, also capture `rr_position`, `rr_transform` (only if not `none`), and `rr_display` (only if `inline`).
- rebuild.ts: skip the forced `position: absolute` when `rr_position === 'static'` and no contributing transform; promote inline placeholders to `inline-block` so width/height hold the slot. Old recordings (no `rr_position`) keep the legacy absolute path.

Tested:

- `pnpm --filter @posthog/rrweb-snapshot test` — added 3 snapshot cases (static block, transformed, inline) and 5 rebuild cases (in-flow, transformed-in-flow, non-static, inline promotion, inline already-positioned).
- `pnpm --filter @posthog/rrweb test` — full integration suite passes; one block-element fixture snapshot updated to reflect the intended behavior change (blocked divs rebuild as `width/height` only, no `position: absolute`).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/rrweb/rrweb-snapshot/src/rebuild.ts
+++ b/packages/rrweb/rrweb-snapshot/src/rebuild.ts
@@ -291,6 +291,28 @@ function buildNode(
         }
       }
 
+      // Old recordings (no rr_position) fall through to the legacy absolute path.
+      // relative/sticky also occupy a normal-flow slot — original offsets aren't
+      // captured but staying in-flow beats collapsing siblings.
+      const inFlow =
+        specialAttributes.rr_position === 'static' ||
+        specialAttributes.rr_position === 'relative' ||
+        specialAttributes.rr_position === 'sticky';
+      const wasInFlow =
+        inFlow &&
+        (specialAttributes.rr_transform === undefined ||
+          specialAttributes.rr_transform === 'none');
+
+      // Restore captured display; plain `inline` ignores width/height so promote
+      // to `inline-block`. `inline-flex`/`inline-grid`/`inline-block` apply as-is.
+      if (wasInFlow && typeof specialAttributes.rr_display === 'string') {
+        const display = specialAttributes.rr_display;
+        (node as HTMLElement).style.setProperty(
+          'display',
+          display === 'inline' ? 'inline-block' : display,
+        );
+      }
+
       for (const name in specialAttributes) {
         const value = specialAttributes[name];
         // handle internal attributes
@@ -327,11 +349,15 @@ function buildNode(
         } else if (name === 'rr_height') {
           (node as HTMLElement).style.setProperty('height', value.toString());
         } else if (name === 'rr_left') {
-          (node as HTMLElement).style.setProperty('left', value.toString());
-          (node as HTMLElement).style.setProperty('position', 'absolute');
+          if (!wasInFlow) {
+            (node as HTMLElement).style.setProperty('left', value.toString());
+            (node as HTMLElement).style.setProperty('position', 'absolute');
+          }
         } else if (name === 'rr_top') {
-          (node as HTMLElement).style.setProperty('top', value.toString());
-          (node as HTMLElement).style.setProperty('position', 'absolute');
+          if (!wasInFlow) {
+            (node as HTMLElement).style.setProperty('top', value.toString());
+            (node as HTMLElement).style.setProperty('position', 'absolute');
+          }
         } else if (
           name === 'rr_mediaCurrentTime' &&
           typeof value === 'number'
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -873,6 +873,7 @@ function serializeElementNode(
   // block element
   if (needBlock) {
     const { width, height, left, top } = n.getBoundingClientRect();
+    const computed = doc.defaultView?.getComputedStyle(n);
     attributes = {
       class: attributes.class,
       rr_width: `${width}px`,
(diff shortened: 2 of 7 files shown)


ksvat:
* **#3678** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3678?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3678?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

fasyy612 (approved):
what a changeset name :feet: LGTM!

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -873,13 +873,26 @@ function serializeElementNode(
...
+    // instead of forcing `position: absolute` and collapsing sibling layout.
+    if (computed) {
+      // JSDOM-style envs return '' for unset computed values; normalize to the CSS default.
+      attributes.rr_position = computed.position || 'static';
+      if (computed.transform && computed.transform !== 'none') {
+        attributes.rr_transform = computed.transform;
+      }
+      if (computed.display === 'inline') {
+        attributes.rr_display = 'inline';
+      }
```
TueHaulund: **Capture all inline-level displays, not just `inline`** (pairs with the rebuild suggestion).

A `static` `inline-block`/`inline-flex`/`inline-grid` placeholder takes the in-flow path on rebuild, but its original inline `style` is stripped — so the rebuilt bare tag falls back to its default display (`inline` for a `<span>`), which ignores `rr_width`/`rr_height` and collapses the slot. Same bug class this PR fixes. Capture the real inline-level display so rebuild can restore it.

```suggestion
      if (computed.display && computed.display.startsWith('inline')) {
        attributes.rr_display = computed.display;
      }
```

packages/rrweb/rrweb-snapshot/src/rebuild.ts
```diff
@@ -291,6 +291,17 @@ function buildNode(
...
+      // Old recordings (no rr_position) fall through to the legacy absolute path.
+      const wasInFlow =
+        specialAttributes.rr_position === 'static' &&
+        (specialAttributes.rr_transform === undefined ||
+          specialAttributes.rr_transform === 'none');
+
+      // Inline boxes ignore width/height — promote so the recorded slot is preserved.
+      if (wasInFlow && specialAttributes.rr_display === 'inline') {
+        (node as HTMLElement).style.setProperty('display', 'inline-block');
+      }
```
TueHaulund: **Restore the captured inline-level display** (pairs with the snapshot suggestion).

Promote plain `inline` (which ignores width/height) to `inline-block`; `inline-block`/`inline-flex`/`inline-grid` already honor width/height, so apply them as-is.

```suggestion
      // Inline boxes ignore width/height — restore the captured display so the
      // recorded slot is preserved. Plain `inline` is promoted to `inline-block`;
      // `inline-flex`/`inline-grid` already honor width/height.
      if (wasInFlow && typeof specialAttributes.rr_display === 'string') {
        const display = specialAttributes.rr_display;
        (node as HTMLElement).style.setProperty(
          'display',
          display === 'inline' ? 'inline-block' : display,
        );
      }
```

packages/rrweb/rrweb-snapshot/src/rebuild.ts
```diff
@@ -291,6 +291,17 @@ function buildNode(
         }
       }
 
+      // Old recordings (no rr_position) fall through to the legacy absolute path.
+      const wasInFlow =
+        specialAttributes.rr_position === 'static' &&
+        (specialAttributes.rr_transform === undefined ||
+          specialAttributes.rr_transform === 'none');
```
TueHaulund, line 304: **Treat `relative`/`sticky` as in-flow too.** These positions also occupy a normal-flow slot (then get visually offset), so forcing `position: absolute` pulls them out of flow and collapses the slot — the same symptom this PR fixes for `static`. The snapshot already records the real `rr_position`, so this is rebuild-only.

Tradeoff worth noting: the original `left/top/right/bottom` offset isn't captured (style was stripped), so a `relative`/`sticky` placeholder stays in its in-flow slot without the visual nudge — strictly better than collapsing siblings. This also flips the existing `relative → absolute` rebuild test expectation, so that case needs updating.

```suggestion
      const inFlow =
        specialAttributes.rr_position === 'static' ||
        specialAttributes.rr_position === 'relative' ||
        specialAttributes.rr_position === 'sticky';
      const wasInFlow =
        inFlow &&
        (specialAttributes.rr_transform === undefined ||
          specialAttributes.rr_transform === 'none');
```

TueHaulund (commented):
Two optional follow-ups as suggested edits — both extend the in-flow handling this PR adds, neither is a blocker.

1. **Inline-level displays** (snapshot + rebuild): capture/restore `inline-block`/`inline-flex`/`inline-grid`, not just plain `inline`. Today a static `inline-block` placeholder takes the in-flow path but rebuilds with the tag's default display and collapses — the bug this PR fixes. The two suggestions go together.
2. **`relative`/`sticky` positions** (rebuild): also in-flow, so they shouldn't be forced absolute either. Note this flips one existing rebuild test expectation.

Nice fix overall — backward-compat and test coverage look solid.

PR #3686, opened 2026-05-28
fix(persistence): only persist session activity timestamp every 5s

pauldambra:
Problem

there is an open P2 bug in chromium that local storage writes cause a memory leak https://issues.chromium.org/issues/394823902

we write to local storage a lotthe PH memory leak in https://github.com/PostHog/posthog/issues/32479 is fiendish to replicate in a test

Changes

let's write to local storage less since it can't hurt...

first, let's not track activitity timestamp in millsecond increments writing each to local storage. we can afford to drift a little

\-----

## Problem

`SessionIdManager` writes the session-activity-timestamp into PostHog persistence on every event capture (4+ times per second during active use). Each write serializes the full props blob, writes through localStorage (or cookie + localStorage), and broadcasts a cross-tab `storage` event in every same-origin tab — even when no listener reacts. Chrome allocates the payload buffer in mojo IPC regardless of subscribers, so this is real work.

## Changes

Throttle activity-timestamp-only writes to a 5s granularity. Id-changes and session-start-changes still persist immediately. The in-memory `_sessionActivityTimestamp` is updated at full resolution so within-tab reads still see the freshest tick.

Throttle logic:

- Compare against the last _persisted_ value, not the last in-memory tick, so cumulative sub-granularity drift still crosses the gate at the right point.
- `Math.abs` defends against backwards clock jumps (NTP correction, devtools time travel, another tab writing an older value).
- On `destroy` and `beforeunload`, flush the latest in-memory value through to persistence so a tab closing inside the granularity window does not leave the persisted timestamp up to 5s stale for other tabs / future loads.

### Cross-tab semantics

This is a deliberate, narrow change to cross-tab behaviour: other tabs reading `SESSION_ID` directly from `persistence.props[SESSION_ID]` now see an activity timestamp that can lag the freshest tick by up to 5s. With the minimum configurable `session_idle_timeout_seconds = 60`, that's ~8% of the minimum idle threshold — well under the safety margin, and a regression test pins the invariant. Downstream consumers that read `SESSION_ID` directly (session replay, surveys, feature flags) are not sensitive to sub-5s timestamp freshness.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code


pauldambra:
* **#3690** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3690?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3688** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3687** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3686** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3686?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3686?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pauldambra (commented):
> [!NOTE]

Multi-perspective review complete. See inline comments + summary thread below.

packages/browser/src/sessionid.ts
```diff
@@ -164,19 +177,42 @@ export class SessionIdManager {
...
+        // happen on every event capture (4+ times per second during active
+        // use); persist them only when the timestamp has moved enough that
+        // a stale read could affect idle detection. The in-memory value
+        // remains at full resolution above.
+        if (!idChanged && !startChanged) {
+            const lastPersisted = this._lastPersistedActivityTimestamp
+            if (
+                lastPersisted !== null &&
+                sessionActivityTimestamp !== null &&
+                Math.abs(sessionActivityTimestamp - lastPersisted) <
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + paul + xp]** ⚪ NIT

`Math.abs(...)` on a delta that should be monotonically increasing — defensive against clock skew (NTP correction, devtools time travel, malicious storage writes). All three reviewers landed on the same observation, with two different takes:

- Treat it as deliberate defense and add a one-word comment (`// clock can go backwards`) so a future reader doesn't simplify to a plain subtraction.
- Or treat it as accidental defense — if backwards jumps should always persist (so the stored value tracks the freshest in-memory state) drop the `Math.abs`.

Not blocking either way. Pick a stance and pin it with a comment or a test.

packages/browser/src/sessionid.ts
```diff
@@ -31,6 +39,11 @@ export class SessionIdManager {
     private _sessionStartTimestamp: number | null
 
     private _sessionActivityTimestamp: number | null
+    // Tracks the activity timestamp we most recently persisted. Used to
+    // decide whether a new activity-only change is worth writing to
+    // storage. Reset to null on `resetSessionId` so the next real value
+    // always lands in storage.
+    private _lastPersistedActivityTimestamp: number | null = null
```
pauldambra, line 40: > [!NOTE]

The doc comment says `_lastPersistedActivityTimestamp` is "Reset to null on `resetSessionId`" but there's no explicit reset — it falls out of `resetSessionId` calling `_setSessionId(null, null, null)`, which makes `activityChanged` true, short-circuits the throttle gate (because `sessionActivityTimestamp` is null), persists, and assigns `null` here.

That works *today* but couples three pieces of control flow to maintain the invariant. A future "optimization" to skip the null persist would silently break it. Either:
- add an explicit one-line reset in `resetSessionId`, or
- update the comment to describe how the reset actually happens ("kept in sync via `_setSessionId`").

packages/browser/src/sessionid.ts
```diff
@@ -164,19 +177,42 @@ export class SessionIdManager {
...
-            sessionStartTimestamp !== this._sessionStartTimestamp
-        ) {
-            this._sessionStartTimestamp = sessionStartTimestamp
-            this._sessionActivityTimestamp = sessionActivityTimestamp
-            this._sessionId = sessionId
-
-            this._persistence.register({
-                [SESSION_ID]: [sessionActivityTimestamp, sessionId, sessionStartTimestamp],
-            })
+        const idChanged = sessionId !== this._sessionId
```
pauldambra, line 184: > [!NOTE]

**[convergent: qa-team/frontend + xp]** 🟢 LOW

`_setSessionId` has grown from ~10 lines doing one thing to ~30 lines doing three at mixed abstraction levels (detect changes, update memory, decide whether to persist).

XP's ComposedMethod nudge: extract `_shouldPersistActivityOnlyChange(newTimestamp): boolean` so the body reads as a table of contents rather than rationale-in-comments. The `!idChanged && !startChanged` predicate is also repeated as the gate for the throttling block — a local `const onlyActivityChanged` would name the concept and remove the duplication.

Not load-bearing, but consistent with PostHog's general "prefer extract method over explanatory comment" style.

packages/browser/src/sessionid.ts
```diff
@@ -19,6 +19,14 @@ export const MAX_SESSION_IDLE_TIMEOUT_SECONDS = 10 * 60 * 60 // 10 hours
 const MIN_SESSION_IDLE_TIMEOUT_SECONDS = 60 // 1 minute
 const SESSION_LENGTH_LIMIT_MILLISECONDS = 24 * 3600 * 1000 // 24 hours
 
+// Granularity for persisting the session activity timestamp to the
+// persistence store. The in-memory value is updated at full resolution on
+// every call; only writes to storage are throttled.
+// Activity-timestamp-only changes within this window are not persisted.
+// The minimum configurable idle timeout is 60s, so 5s of staleness in the
+// persisted value cannot cause incorrect idle detection.
+const ACTIVITY_TIMESTAMP_PERSIST_GRANULARITY_MS = 5_000
```
pauldambra: > [!NOTE]

Two observations about the hardcoded 5_000ms constant:

1. **paul**: should this be exposed via config? slightly allergic to "we'll tune this with a deploy" for a browser SDK pinned per customer. 5s is a fine default — but a customer with a very chatty integration who wants 1s, or a very quiet one who wants 30s, is stuck until the next upgrade. Hardcoded internal constant today so the threat ratchet is low — flagging for discussion.

2. **xp**: the 5s constant's safety relies on `MIN_SESSION_IDLE_TIMEOUT_SECONDS = 60` being meaningfully larger. The comment captures this, but if someone later lowers the min idle timeout the invariant breaks silently. A unit test that fails when `MIN_SESSION_IDLE_TIMEOUT_SECONDS * 1000 < 6 * GRANULARITY` would catch the regression cheaply.

pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

Observability — we ship to a lot of 3rd party sites, and if there's some weird call pattern where `_setSessionId` never moves more than 5s we'd silently stop persisting forever. Nothing in the diff lets us tell from grafana / a customer's own posthog whether this is doing what we think.

Could we capture a debug-level counter (writes skipped vs writes performed) behind the existing `debug` config, just so we can sanity-check in the wild? Not blocking — the change looks right — but every time we've added a "drop on condition X" gate without observability we've regretted it later.

packages/browser/src/sessionid.ts
```diff
@@ -164,19 +177,42 @@ export class SessionIdManager {
...
+                lastPersisted !== null &&
+                sessionActivityTimestamp !== null &&
+                Math.abs(sessionActivityTimestamp - lastPersisted) <
+                    ACTIVITY_TIMESTAMP_PERSIST_GRANULARITY_MS
+            ) {
+                return
+            }
         }
+
+        this._lastPersistedActivityTimestamp = sessionActivityTimestamp
```
pauldambra, line 203: > [!NOTE]

Cross-tab activity-timestamp staleness — `_getSessionId()` reads from persistence (the cross-tab source of truth). With activity-only writes throttled to once per 5s, a second tab observing the storage event sees an activity timestamp that lags reality by up to 5s.

Math: minimum configurable `session_idle_timeout_seconds` is 60s and the throttle is 5s → worst-case skew is ~8% of the minimum idle threshold. Well under the safety margin, so this **cannot** cause incorrect idle detection. Still, it's an intentional change to cross-tab semantics worth being explicit about for downstream consumers (session replay, surveys, feature flags) that read SESSION_ID directly via `persistence.props[SESSION_ID]`.

Suggestion: note the up-to-5s cross-tab lag in the commit body / PR description, and verify session-replay's use of SESSION_ID isn't sensitive to sub-5s timestamp freshness.

pauldambra, line 203: > [!NOTE]

No flush-on-unload of the latest activity timestamp. The `destroy()` / `beforeunload` paths don't force a final persist of `_sessionActivityTimestamp`. If a tab is closed within the 5s window after the last event capture, the persisted value lags reality by up to 5s.

Not a correctness bug — another tab/page reopened immediately still sees the session correctly because 5s ≪ 60s minimum idle timeout. Just a slightly more pessimistic "last activity" view than before. Optional: on `destroy()` or in `_beforeUnloadListener`, if `_sessionActivityTimestamp !== _lastPersistedActivityTimestamp`, write the latest value through. Only worth doing to preserve the pre-change semantics exactly.

packages/browser/src/__tests__/sessionid.test.ts
```diff
@@ -307,6 +307,116 @@ describe('Session ID manager', () => {
         })
     })
 
+    describe('activity timestamp persistence granularity', () => {
```
pauldambra, line 317: > [!NOTE]

**[qa-team/frontend]** ⚪ NIT

Test coverage is good but missing two cases worth adding:

1. **Activity timestamp going backwards within 5s** (clock skew / NTP adjustment) — `Math.abs` covers this in code but no test pins the behavior.
2. **Post-id-change baseline check**: after a sessionId change persists the new activity timestamp, verify subsequent within-5s activity-only ticks correctly compare against that new baseline (not the prior session's last-persisted value).

Suggestion: extend the existing `it.each` patterns with backward-delta and post-id-change-baseline cases.

packages/browser/src/__tests__/sessionid.test.ts
```diff
@@ -307,6 +307,116 @@ describe('Session ID manager', () => {
...
+            const sessionIdManager = sessionIdMgr(persistence)
+            sessionIdManager['_setSessionId']('id', 1_000_000, 1_000_000)
+            sessionIdManager['_setSessionId']('id', 1_000_001, 1_000_000)
+            // Under the 5_000 ms granularity, persistence is skipped, but the
+            // in-memory value still reflects the latest tick — important for
+            // any in-process readers and for the next granularity comparison.
+            expect(sessionIdManager['_sessionActivityTimestamp']).toBe(1_000_001)
+        })
+
+        it('compares against last persisted, not last in-memory', () => {
```
pauldambra, line 392: > [!NOTE]

**[convergent: paul + xp]** ⚪ NIT — positive

The "compares against last persisted, not last in-memory" test is the single highest-value test in this PR — it pins down the exact failure mode the design prevents (the trap of comparing to the previous in-memory tick).

paul's note: "this is a great test". xp's note: "exactly what good TDD looks like — it specifies the design rather than just exercising the code". Worth keeping this prominent in the describe block so a future reader sees the design intent first.

pauldambra:
> [!NOTE]
>

All four reviewers landed in the same place: the change is well-scoped, the logic is correct, the tests pin the right boundaries. The 5s throttle vs 60s minimum idle timeout gives a 12x safety margin so this cannot cause incorrect idle detection. No critical or high findings. Several low/nit observations worth considering but none block merge.

### Key findings

🟢 LOW
- **Implicit `resetSessionId` semantics** (convergent: paul + xp) — comment claims a null reset that's actually achieved indirectly via `_setSessionId(null,null,null)`. Tighten the comment or add explicit reset.
- **`_setSessionId` is doing three things at mixed abstraction** (convergent: qa-team + xp) — candidate for ComposedMethod extraction (`_shouldPersistActivityChange`).
- **Cross-tab staleness contract** (qa-team) — downstream consumers reading `SESSION_ID` from persistence now see up to 5s lag. Math is safe (5s ≪ 60s min idle) but worth noting in the commit body.
- **No flush-on-unload** (qa-team) — tab closed within 5s of last event leaves the persisted activity timestamp up to 5s stale. Not a correctness bug; optional polish.
- **No observability** (paul) — every "drop if X" gate without a counter is hard to verify in the wild. Suggest a debug-level skip counter.

⚪ NIT
- **`Math.abs`** (convergent x3: qa-team + paul + xp) — clock-skew defense or accidental defense? Pin a stance.
- **5s as a hardcoded constant** (convergent: paul + xp) — paul wants it config-exposed; xp wants a test asserting the safety relationship with `MIN_SESSION_IDLE_TIMEOUT_SECONDS`.
- **Missing test cases** (qa-team) — backwards-delta + post-id-change baseline.
- **Positive: parameterized tests + last-persisted test** (paul + xp) — both reviewers explicitly praised these. The "compares against last persisted, not last in-memory" test is the most valuable in the suite.

### Convergence

Findings flagged independently by 2+ reviewers (high confidence):
- `Math.abs` on the timestamp delta — 3 reviewers (qa-team + paul + xp)
- Implicit `resetSessionId` reset semantics — 2 reviewers (paul + xp)
- `_setSessionId` structure / extract method — 2 reviewers (qa-team + xp)
- 5s constant tradeoffs (config vs safety coupling) — 2 reviewers (paul + xp)
- "Compares against last persisted" test is the best — 2 reviewers, positive (paul + xp)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Focused, well-scoped optimisation. 12x safety margin vs minimum idle timeout makes incorrect idle detection impossible. Findings are minor (comment-vs-extract-method, optional unload flush, test-coverage gaps, cross-tab contract note). Approve with optional nits. |
| 👤 paul | high trust, ship as you see fit. logic is clear, tests cover boundaries plus the subtle compare-against-last-persisted case. one real ask: observability — a debug-level skip counter would let us answer "is this doing what we think" without a redeploy. rest is preference. |
| 📐 xp | Focused perf fix in the spirit of MakeItWorkMakeItRightMakeItFast. Naming intention-revealing, comments explain *why* not *what*, parameterized tests pin boundary behaviour. Quibble: `_setSessionId` could use ComposedMethod extraction; reset semantics rely on implicit fallout. Otherwise clean, simple, ship it. |
| 🛡 security-audit | 0 findings. The throttle window (5s) is an order of magnitude below the minimum configurable idle timeout (60s). SessionId/start changes always force a write. `Math.abs` covers backwards time. JS is single-threaded → no race on `_lastPersistedActivityTimestamp`. An attacker who controls localStorage already controls the session tuple — no new sink, no control removed. |

---

packages/browser/src/sessionid.ts
pauldambra: Decided not to add this in this PR. A `debug`-gated counter is only useful when someone is already investigating, which is the same threat model as just reading the code. And if the throttle is misbehaving in the pathological way, idle detection also breaks, which we can already see in dogfood data (session length / idle-reset rates in our own PostHog project). Will revisit if we see anything weird in the wild.

packages/browser/src/sessionid.ts
pauldambra: Done in 86ebe2ac — `resetSessionId` now explicitly sets `_lastPersistedActivityTimestamp = null` before calling `_setSessionId(null, null, null)`, and the doc comment on the field is updated to match.

packages/browser/src/sessionid.ts
pauldambra: Done in 86ebe2ac — extracted `_isActivityChangeBelowGranularity(newActivityTimestamp)` and introduced `const isActivityOnlyChange` to remove the duplicated `!idChanged && !startChanged` predicate. Clock-skew rationale now lives in the helper alongside the `Math.abs`.

packages/browser/src/sessionid.ts
pauldambra: Done — the PR body now has an explicit "Cross-tab semantics" section calling out the up-to-5s skew and noting that downstream SESSION_ID readers (session replay, surveys, feature flags) are not sensitive at sub-5s freshness.

pauldambra: Done in 86ebe2ac — added `_flushPendingActivityTimestamp` which is called from both `destroy()` and the `beforeunload` listener. Two new tests pin both branches (pending data flushed; no-op when in-memory matches persisted).

packages/browser/src/__tests__/sessionid.test.ts
pauldambra: Done in 86ebe2ac — added an `it.each` covering backward deltas (-1ms, -4_999ms, -5_000ms boundary, -6_000ms clock-skew jump) and a "compares against the new persisted baseline after an id change" test.

packages/browser/src/__tests__/sessionid.test.ts
pauldambra: Thanks. Keeping it as-is in its current position in the describe block.

packages/browser/src/sessionid.ts
pauldambra: xp half done — invariant test `persist granularity stays well under the minimum idle timeout` added in 86ebe2ac, asserts `MIN_SESSION_IDLE_TIMEOUT_SECONDS * 1000 >= 6 * ACTIVITY_TIMESTAMP_PERSIST_GRANULARITY_MS`.

paul half — not exposing via config for now. 5s is so far below the granularity anyone is realistically tuning for that the "stuck until next upgrade" risk is hypothetical. Unlike `persistence_save_debounce_ms` (correctness-neutral throttle), this one has a safety floor (must stay well under the min idle timeout), so exposing it would also need range validation. Revisit if a real customer asks.

packages/browser/src/sessionid.ts
```diff
@@ -157,26 +172,73 @@ export class SessionIdManager {
...
+        }
+
+        // SESSION_ID changes and session-start changes are semantically
+        // important — always persist them. Activity-timestamp-only changes
+        // happen on every event capture (4+ times per second during active
+        // use); persist them only when the timestamp has moved enough that
+        // a stale read could affect idle detection. The in-memory value
+        // remains at full resolution above.
+        if (isActivityOnlyChange && this._isActivityChangeBelowGranularity(sessionActivityTimestamp)) {
+            return
```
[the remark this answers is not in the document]
pauldambra: Fixed in 29e12d4b. Added `_freshestActivityTimestamp()` which returns `max(persisted, in_memory)`, used in both `checkAndGetSessionAndWindowId` and the `_resetIdleTimer` callback. Regression test: "does not rotate session early when in-memory activity is fresher than persisted".

packages/browser/src/sessionid.ts
```diff
@@ -157,26 +172,73 @@ export class SessionIdManager {
...
-            this._sessionId = sessionId
-
-            this._persistence.register({
-                [SESSION_ID]: [sessionActivityTimestamp, sessionId, sessionStartTimestamp],
-            })
+            return
         }
+        this._lastPersistedActivityTimestamp = this._sessionActivityTimestamp
+        this._persistence.register({
+            [SESSION_ID]: [this._sessionActivityTimestamp, this._sessionId ?? null, this._sessionStartTimestamp],
```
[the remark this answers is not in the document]
pauldambra: Fixed in 29e12d4b. `_flushPendingActivityTimestamp` now calls `_persistence.load()` to refresh from storage, then bails out if `persistedSessionId !== this._sessionId || persistedStart !== this._sessionStartTimestamp`. Two tests added: "flush does not clobber a cross-tab session rotation" and "flush proceeds when persisted session still matches the tab".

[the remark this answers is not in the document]
pauldambra: Declining — `_sessionId` is declared as `string | null | undefined` (not `string | null`) because the constructor initializes it to `undefined` before any `_setSessionId` call. The `?? null` normalizes `undefined` to `null` for the persisted tuple. The other two fields are `number | null` (no `undefined`), which is why they do not need the fallback.

lricoy (approved):
Let's test it, LGTM.   
  
How will we track whether it IS working or failing, though?

marandaneto:
> [issues.chromium.org/issues/394823902](https://issues.chromium.org/issues/394823902)

uff...

> How will we track whether it IS working or failing, though?

yeah not sure either, down to try tho!

marandaneto (approved):

PR #3690, opened 2026-05-28
feat(persistence): refresh from storage before declaring an idle session

pauldambra:
a session now rotates only when every tab has gone idle past the timeout, instead of whenever any single tab decides it's idle. On the active event path, when a tab's own idle check says "rotate", it re-reads SESSION_ID from shared localStorage first; if a sibling kept the session alive it doesn't rotate, and if a sibling already rotated it adopts that id rather than minting a third. The idle timer re-arms instead of firing once, and (when persistence_save_debounce_ms > 0) adoption emits onSessionId.

## Problem

Part of the cross-tab session work that came out of the browser memory investigation in PostHog/posthog#32479 — the same multi-tab behaviour that buffered storage events in hidden renderers also fragments sessions across tabs.

Each tab in PostHog runs its own `SessionIdManager` with its own idle timer. When the timer fires (or the next `checkAndGetSessionAndWindowId` call runs the idle check), it reads from `_persistence.props[SESSION_ID]` — but PostHogPersistence does not listen for `storage` events, so that cache only ever reflects what THIS tab has loaded or written. Sibling tabs' writes are invisible.

The consequence: a user with three tabs open, active in one and idle in the other two, gets their session **rotated** by the idle tabs after `session_idle_timeout_seconds` — splitting one continuous session into multiple session IDs in the analytics data.

## Changes

Add `_isSessionIdleAfterCrossTabRefresh(timestamp)` which calls `_persistence.load()` before consulting `_freshestActivityTimestamp`. Wire it into both:

- the rotation path in `checkAndGetSessionAndWindowId` (only on the rotation branch, to avoid the cost on every event capture)
- the idle timer callback in `_resetIdleTimer`

When the timer fires and the refresh shows the session is still alive (because a sibling kept it active), re-arm the timer so the next check happens after another timeout window.

## Tests

- jest unit tests pin both directions (no rotate when sibling is active; rotate when storage agrees)
- Playwright tests run two real tabs in the same context and check the behaviour against real `localStorage` and real `PostHogPersistence.load()` — they pass with the fix, and stashing the fix produces a fail on the cross-tab test

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code


pauldambra:
* **#3690** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3690?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3690?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3688** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3688?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3687** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3687?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3686** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3686?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

packages/browser/src/sessionid.ts
```diff
@@ -336,8 +343,13 @@ export class SessionIdManager {
...
         const noSessionId = !sessionId
-        const activityTimeout =
+        let activityTimeout =
             !noSessionId && !readOnly && this._sessionHasBeenIdleTooLong(timestamp, lastActivityTimestamp)
+        if (activityTimeout) {
+            // Re-check against the freshest cross-tab view: a sibling tab
+            // may have kept the session alive.
+            activityTimeout = this._isSessionIdleAfterCrossTabRefresh(timestamp)
+        }
         if (noSessionId || activityTimeout || sessionPastMaximumLength) {
```
pauldambra, line 409: > [!NOTE]

**[convergent: xp + security-audit]** 🟠 HIGH

After `_isSessionIdleAfterCrossTabRefresh` triggers a `_persistence.load()`, the surrounding code keeps using the `sessionId` and `startTimestamp` sampled at line 337 — *before* the load. If a sibling tab has rotated to a new session, the write at line 376 (`_setSessionId(sessionId, newActivityTimestamp, sessionStartTimestamp)`) clobbers the sibling's new session with the stale in-memory id. This is the **exact** failure mode `_flushPendingActivityTimestamp` was hardened against — but the same defence is missing here. Re-sample `_getSessionId()` after the cross-tab refresh and either bail or use the refreshed values. The asymmetry between the flush guard and the rotation hot path is begging for the bug.

packages/browser/src/sessionid.ts
```diff
@@ -389,14 +401,14 @@ export class SessionIdManager {
     private _resetIdleTimer() {
```
pauldambra, line 468: > [!NOTE]

**[convergent: xp + paul + qa-team/reliability]** 🟠 HIGH

The new self-perpetuating `setTimeout` recursion has three converging concerns from three reviewers:

1. **destroy() race (xp)**: the `else` branch unconditionally calls `_resetIdleTimer()`. If `destroy()` ran between the timer firing and the callback executing, the freshly assigned `_enforceIdleTimeout` outlives the manager and will call `resetSessionId` on a destroyed instance. Guard the re-arm on `this._enforceIdleTimeout !== undefined` (`destroy` clears it to undefined) or introduce a destroyed flag.

2. **Observability (paul)**: an idle tab whose sibling stays active will `setTimeout` itself indefinitely, calling `_persistence.load()` every ~66s for the lifetime of the tab. No counter, no breadcrumb. Very Knight Capital — a runaway re-arm loop is invisible. Worth at least a debug counter so we can see in prod whether re-arm cycles ever balloon.

3. **External rotation (qa-team)**: `resetSessionId()` does not clear `_enforceIdleTimeout`. If something rotates the session externally (`posthog.reset()`, manual capture-pixel reset) between two timer fires, the next fire looks at the freshly rotated session and may `forcedIdleReset` it. Pre-existing latent bug surfaced more often by the new re-arm.

All three are about the same wiring. Fix together: clear `_enforceIdleTimeout` in `resetSessionId`, guard re-arm on non-undefined, add a debug counter.

packages/browser/src/sessionid.ts
```diff
@@ -239,6 +239,13 @@ export class SessionIdManager {
         return Math.max(persisted, inMemory)
     }
 
+    // PostHogPersistence does not listen for `storage` events, so this tab's
+    // cached props lag sibling tabs' writes. Re-load before deciding idle.
+    private _isSessionIdleAfterCrossTabRefresh(timestamp: number): boolean {
+        this._persistence.load()
```
pauldambra: > [!NOTE]

**[qa-team/data-integrity]** 🟠 HIGH

`_persistence.load()` does `this.props = extend({}, entry)` — it replaces the in-memory props with whatever storage currently holds. Previously this side-effect only fired from `destroy()` / `beforeunload` / `_flushPendingActivityTimestamp` (all terminal paths), where losing un-persisted in-memory state is acceptable.

Moving this onto the live event-capture hot path means that any `register()` write held in memory due to `persistence_save_debounce_ms > 0` (the new dated default at `2026-05-30` sets it to 250ms) will be silently wiped the next time an idle-tab event triggers `activityTimeout`. Default debounce is 0 for most deployments today, but anyone on the new dated default — including PostHog itself once it rolls out — has a latent data-loss path on the activity hot path.

Either flush pending debounced writes (`this._persistence.flush()`) before `load()`, or reload into a scratch object and reconcile rather than overwriting `props` wholesale.

packages/browser/src/__tests__/sessionid.test.ts
```diff
@@ -551,6 +551,38 @@ describe('Session ID manager', () => {
...
+            const queryTime = 1_000_000 + sessionIdManager.sessionTimeoutMs + 5_000
+            // Sibling tab wrote 1 second ago, so the session is alive.
+            persistence.props[SESSION_ID] = [queryTime - 1_000, 'sessionA', 1_000_000]
+
+            const result = sessionIdManager.checkAndGetSessionAndWindowId(false, queryTime)
+            expect(result.sessionId).toBe('sessionA')
+            expect(result.changeReason?.activityTimeout).toBeFalsy()
+        })
+
+        it('checkAndGetSessionAndWindowId rotates when cross-tab refresh confirms idle', () => {
```
pauldambra, line 662: > [!NOTE]

**[convergent: xp + qa-team/reliability + paul]** 🟡 MEDIUM

Three reviewers independently flagged that the new unit tests are weaker than they look:

- **xp**: this test (`rotates when cross-tab refresh confirms idle`) passes whether `_persistence.load()` is called or not — the mock load is a no-op and `persistence.props` isn't mutated, so the cross-tab refresh just re-confirms the original check. A future refactor that drops the `load()` call would still pass.
- **qa-team**: the tests mock `persistence.load` as a bare `jest.fn()` and simulate the cross-tab refresh by directly mutating `persistence.props`. The behaviour drift the PR is meant to prevent is only really covered by the Playwright tests.
- **paul**: there's no unit test that exercises the timer callback re-arm loop. Could `jest.useFakeTimers()` advance once with a sibling write (assert no rotation + timer re-armed), then again with no sibling activity (assert rotation finally fires)?

Concrete fix: drop a `load: jest.fn().mockImplementation(() => { persistence.props[SESSION_ID] = [latestStorageActivity, sessionId, start] })` so the mock models the real semantic, then add a fake-timer-based re-arm test. As written, only the first new unit test pins the fix; the second is decoration and the timer re-arm is uncovered.

packages/browser/src/sessionid.ts
```diff
@@ -336,8 +343,13 @@ export class SessionIdManager {
 
         let valuesChanged = false
         const noSessionId = !sessionId
-        const activityTimeout =
+        let activityTimeout =
             !noSessionId && !readOnly && this._sessionHasBeenIdleTooLong(timestamp, lastActivityTimestamp)
+        if (activityTimeout) {
```
pauldambra, line 391: > [!NOTE]

This PR fixes the case where a sibling tab has a recent **persisted** activity timestamp. It does not address the case where the sibling tab has a recent activity timestamp only *in memory* due to the 5s `ACTIVITY_TIMESTAMP_PERSIST_GRANULARITY_MS` throttle.

Scenario: user clicks tab A at t=0 (persisted), clicks again at t=4s (throttled — in-memory only), then goes silent. Tab B fires its idle check at t=65s, reads persisted=0, gap=65s > 60s, calls `load()` which still shows persisted=0, rotates. The user's t=4s click is invisible across tabs. This is narrow but the same shape of silent corruption the PR is targeting.

Consider documenting the residual gap, or having the active tab unconditionally persist on `visibilitychange: hidden` so handoff to a sibling is consistent.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+
+async function startSecondTab(context: BrowserContext): Promise<Page> {
+    const tab = await context.newPage()
+    await tab.clock.install()
+    await start(startOptions, tab, context)
+    return tab
+}
+
+test.describe('cross-tab session lifecycle', () => {
+    test.beforeEach(async ({ page, context }) => {
```
pauldambra, line 102: > [!NOTE]

**[qa-team/frontend]** 🟡 MEDIUM

Only `activeTab` gets `tab.clock.install()` via `startSecondTab`; `idleTab` (the base `page` fixture) runs on the real wall clock. The asymmetry isn't documented and isn't required by the test's logic (the check is driven by explicit `timestamp` arguments to `checkAndGetSessionAndWindowId`).

Two risks: (1) on slow CI the real-clock `setTimeout(..., timeoutMs * 1.1)` on `idleTab` may actually fire during the test run and rotate the session under the assertion, producing flakes; (2) future maintainers reading this test won't know whether the asymmetry is load-bearing.

Either install the clock on both tabs (and frame the test with `setFixedTime`), or remove it from `activeTab` and explain why neither needs it. Looking at the base test fixture — `page.clock.install()` may already be applied via the fixture; if so, the redundant install on `activeTab` should go.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        expect(await getSessionId(idleTab)).toBe(sharedSessionId)
+    })
+
+    test('session rotates when all tabs have been idle past the timeout', async ({ page, context }) => {
+        // Sanity check: when no tab has been active, the session must still
+        // rotate after the timeout. This guards against the cross-tab
+        // refresh fix accidentally keeping dead sessions alive.
+        const tabA = page
+        const tabB = await startSecondTab(context)
+
```
pauldambra: > [!NOTE]

`session rotates when all tabs have been idle past the timeout` passes equally with and without the fix (in both cases the idle-past-timeout tab rotates). It's a valuable regression guard against the fix accidentally pinning dead sessions alive, but it does not discriminate the new behaviour.

Worth a one-line comment in the test so a future reader doesn't assume both tests cover the fix — only the first one does.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+        await start(startOptions, page, context)
+    })
+
+    test('idle sibling tab does not rotate the session while another tab keeps it alive', async ({ page, context }) => {
+        // Pins the cross-tab refresh contract in `checkAndGetSessionAndWindowId`:
+        //   PostHogPersistence does not subscribe to `storage` events, so an
+        //   idle tab's cached `_persistence.props[SESSION_ID]` lags writes from
+        //   sibling tabs. Without the fix, when the idle tab's next idle check
+        //   runs, it reads its own stale activity timestamp and rotates the
+        //   session — even though another tab has been writing fresh activity.
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

both tests share roughly the same setup-and-assert shape (start a sibling, drive activity, trigger a check, assert rotation behaviour). could be parameterized over `(sibling-writes-activity, expected-rotation)` to make it obvious at a glance that these two cases are the same logic with different inputs. not blocking, you know how i feel about parameterized tests.

packages/browser/src/sessionid.ts
```diff
@@ -239,6 +239,13 @@ export class SessionIdManager {
         return Math.max(persisted, inMemory)
     }
 
+    // PostHogPersistence does not listen for `storage` events, so this tab's
+    // cached props lag sibling tabs' writes. Re-load before deciding idle.
+    private _isSessionIdleAfterCrossTabRefresh(timestamp: number): boolean {
```
pauldambra, line 281: > [!NOTE]

**[paul]** ⚪ NIT

nit picking — `_isSessionIdleAfterCrossTabRefresh` reads as a pure predicate but its whole reason for existing is the `_persistence.load()` side effect. someone refactoring later might inline it not realising that. could the name lean harder on that, or could we keep a comment-on-the-name? i don't have a great rename to offer, just flagging. ship as you see fit.

packages/browser/src/sessionid.ts
```diff
@@ -336,8 +343,13 @@ export class SessionIdManager {
 
         let valuesChanged = false
         const noSessionId = !sessionId
-        const activityTimeout =
+        let activityTimeout =
```
pauldambra, line 389: > [!NOTE]

**[paul]** ⚪ NIT

i'm slightly allergic to the `let activityTimeout = ... ; if (activityTimeout) { activityTimeout = ... }` shape — reads like it might be a bug for half a second. could be `const activityTimeout = !noSessionId && !readOnly && this._sessionHasBeenIdleTooLong(...) && this._isSessionIdleAfterCrossTabRefresh(timestamp)` which keeps the short-circuit. feel free to disagree, this is taste.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+    // is the same call `posthog.capture` makes internally — but unlike
+    // `posthog.capture` we don't follow it with a real-clock `Date.now()`
+    // write that would clobber the simulated timestamp.
+    await page.evaluate((ts: number) => {
+        const ph = (window as WindowWithPostHog).posthog as any
+        ph.sessionManager.checkAndGetSessionAndWindowId(false, ts)
+    }, timestampMs)
+}
+
+async function triggerNextEventCheck(page: Page, timestampMs: number): Promise<string | null> {
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

`(window as WindowWithPostHog).posthog as any` reaches into a private surface (`sessionManager`, `checkAndGetSessionAndWindowId`). A rename on either silently breaks this test instead of failing typecheck.

A small typed helper on the playwright base (`getSessionManagerHandle(page)`) would express the coupling once and let the rest of the file read declaratively.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,134 @@
...
+//
+// Why we drive the rotation check explicitly with a future timestamp
+// instead of waiting for the real idle timer:
+//   The real `setTimeout(..., timeout * 1.1)` callback would take >60s to
+//   fire (the configured-min idle timeout) in a deterministic Playwright
+//   clock setup, and the Playwright clock's interaction with paused-vs-
+//   ticking behaviour is fiddly across `setFixedTime` / `pauseAt`.
+//   Instead we synthesize the same call the timer would make — passing a
+//   `timestamp` value to `checkAndGetSessionAndWindowId` that puts the
+//   tab well past the idle threshold relative to its own in-memory and
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

The header comment explains *why* the timer is bypassed but not *what is therefore not covered*. The re-arm path and the throttle-plus-cross-tab-refresh interactions both live behind the real timer and are not exercised here or in the unit tests.

Worth one explicit timer-fires-and-re-arms test (with `jest.useFakeTimers()` in the unit suite — see the convergent test-weakness finding) rather than asserting in prose that "unit tests cover the timer callback flow." They cover one fire, not the re-arm.

pauldambra:
> [!NOTE]
>
>
> The reviewer was briefed that **session-rotation changes are load-bearing for downstream analytics** — once a session is incorrectly rotated, the data is unrecoverable. Findings calibrated accordingly: if a reviewer felt uneasy about something subtle, they were asked to surface it rather than drop it.

Three convergent HIGH findings — two correctness bugs that match the failure mode this PR is targeting (silent session corruption), and a latent data-loss path when combined with the new `persistence_save_debounce_ms` dated default landing in the same stack. The fix's shape is right, but the implementation has unfinished symmetry with the existing `_flushPendingActivityTimestamp` guard and the re-arm wiring has a `destroy()` race.

### Key findings

**🟠 HIGH — Convergent (xp + security-audit):** `sessionid.ts:353` writes back the pre-load `sessionId` after `_persistence.load()` — can clobber a sibling tab's rotation. Same shape as the bug the `_flushPendingActivityTimestamp` recheck guards against; the same guard is missing here.

**🟠 HIGH — Convergent (xp + paul + qa-team):** `sessionid.ts:401` re-arm logic — destroy() race (timer outlives the manager), no observability (silent infinite re-arm), and the pre-existing fact that `resetSessionId()` does not clear `_enforceIdleTimeout` becomes more dangerous.

**🟠 HIGH — qa-team/data-integrity:** `sessionid.ts:245` `_persistence.load()` overwrites in-memory `props` wholesale. Now fires on the live event-capture hot path. When combined with the `persistence_save_debounce_ms: 250` dated default landing in #3688, debounced writes get silently wiped.

**🟡 MEDIUM — Convergent (xp + qa-team + paul):** New unit tests are weaker than they look — `load: jest.fn()` is a no-op, and props are mutated by the test itself. The tests would still pass if the fix were removed. Only the first new test pins behaviour; the second is decoration. No fake-timer-based re-arm test.

**🟡 MEDIUM — qa-team/data-integrity:** Throttle edge case persists — sibling tab's in-memory activity within the 5s throttle window is invisible to idle-tab refresh.

**🟡 MEDIUM — qa-team/frontend:** Asymmetric `clock.install()` in Playwright test — flake risk on slow CI.

### Convergence

| Concern | Reviewers | File:line |
|---|---|---|
| Pre-load `sessionId` clobbers sibling rotation | xp + security-audit | `sessionid.ts:353` |
| Re-arm wiring (race + observability + external-rotation timer leak) | xp + paul + qa-team | `sessionid.ts:401` |
| Unit tests don't actually exercise `load()` semantic / re-arm path | xp + qa-team + paul | `sessionid.test.ts:576` |

### Pre-existing issues this PR surfaces more strongly

- **`Math.abs` in `_sessionHasBeenIdleTooLong`** — clock-skew defence currently turns "future activity" into "long-ago activity" and falsely declares idle. Now reached more often via the new code path.
- **`_persistence.load()` cost on idle hot path** — extra `localStorage.getItem` + `JSON.parse` of the full persistence blob when activityTimeout fires. Gated correctly behind the rare idle condition.
- **Two-tab race on rotation** — when this tab rotates, sibling reading storage between rotation and next event capture can independently generate a new session. Pre-existing, not introduced here.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Right shape, two leaky abstractions: `load()` clobbers in-memory pending writes when debounce is enabled, and unit tests model the simulated semantic via direct `props` mutation rather than the actual `load()` call. |
| 👤 paul | Right fix for a scary bug, but the new indefinite re-arm loop has no unit coverage and no observability — very Knight Capital-y to silently `setTimeout` forever. |
| 📐 xp | Fix expresses the right idea OnceAndOnlyOnce, but the symmetry with `_flushPendingActivityTimestamp` is unfinished (clobber risk) and the re-arm races with destroy(). |
| 🛡 security-audit | No exploitable security findings — the trust boundary for cross-tab `localStorage` is unchanged. Notes the same correctness concern as xp (post-refresh sessionId clobber) but doesn't file as security. |

---

packages/browser/src/sessionid.ts
pauldambra: Fixed in 183c8d05 — after `_isSessionIdleAfterCrossTabRefresh` runs, `_getSessionId()` is re-sampled so the post-refresh write uses the freshest values. Mirrors the same defence `_flushPendingActivityTimestamp` already had on the unload path. New regression test: `does not clobber a sibling tab session id when cross-tab refresh keeps the session alive`.

packages/browser/src/sessionid.ts
pauldambra: Partially fixed. The three sub-concerns:

- **destroy() race**: fixed in 183c8d05 by a `_destroyed` flag. `destroy()` sets it; both the entry of `_resetIdleTimer` and the timer callback short-circuit on it. Regression test: `idle timer does not re-arm after destroy()`.
- **`resetSessionId()` does not clear `_enforceIdleTimeout`**: fixed earlier on #3686 in 639ae7de.
- **Observability**: not adding a counter — would need a debug surface we do not have today. Paul flagged this as "consider, not blocking". Filing as a follow-up if we see actual runaway re-arm in production telemetry.

packages/browser/src/sessionid.ts
pauldambra: Fixed in 183c8d05 — `_isSessionIdleAfterCrossTabRefresh` now calls `this._persistence.flush()` before `this._persistence.load()`, mirroring the same fix already applied to `_flushPendingActivityTimestamp` earlier in the stack. Pending debounced writes are pushed to storage before `load()` replaces in-memory `props`, so they survive. Regression test: `flushes pending debounced writes before _persistence.load() clobbers them`.

packages/browser/src/__tests__/sessionid.test.ts
pauldambra: Strengthened in 183c8d05 — the new `cross-tab refresh hardening` block includes a regression test `flushes pending debounced writes before _persistence.load() clobbers them` that pins the actual `load()`-and-`flush()` invocation order via `invocationCallOrder`. It would now fail if `load()` were skipped or moved before `flush()`. The other two new tests (sibling-id-not-clobbered, no-rearm-after-destroy) also use a `load: jest.fn().mockImplementation(...)` that mutates `props` to simulate the real load-from-storage semantic — addresses the original convergent concern that the tests were not actually exercising the side effect.

packages/browser/src/sessionid.ts
pauldambra: Acknowledged. The 5s throttle blind spot (sibling in-memory-only activity within the throttle window is invisible) is documented in the `ACTIVITY_TIMESTAMP_PERSIST_GRANULARITY_MS` constant comment in `sessionid.ts` (added on #3686). Fixing it properly needs explicit cross-tab signaling, which is out of scope for this stack. Resolving — residual gap is acknowledged in the code.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: Both tabs DO get the Playwright clock installed: the first tab via `playwright/fixtures/page.ts` which auto-installs on the fixture-provided `page`, and the second tab via the explicit `tab.clock.install()` in `startSecondTab`. They are symmetric — the asymmetric appearance was an artefact of one being implicit. Rewritten in 5aaf8bcd to document this explicitly in the helper comment so a future reader does not wonder.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: Addressed in 5aaf8bcd by adding an inline note to the spec header: the all-idle case is a regression guard, not a fix-pin, and the file-level header now explicitly enumerates what the suite does and does not cover.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: Done in 5aaf8bcd — both tests parameterized over `(siblingKeepsAlive)` with one shared body, since they only differ in that one branch.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: Fixed in 5aaf8bcd — dropped the `as any` casts. `WindowWithPostHog` already exposes a typed `posthog: PostHog`, and `PostHog.sessionManager` / `SessionIdManager.checkAndGetSessionAndWindowId` are public on the typed surface. The Playwright tests now call those via the typed accessor, so a rename fails typecheck rather than silently breaking.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: Rewritten in 5aaf8bcd — the header now enumerates what the file does cover (cross-tab localStorage sharing, real `load()`, both sides of the contract) AND what it does not (real timer callback, throttle interactions). Points the reader at the unit tests for the bits not exercised here.

packages/browser/src/sessionid.ts
pauldambra: Keeping the name. `_isSessionIdleAfterCrossTabRefresh` already encodes the side effect — "after cross-tab refresh" reads as "we did a refresh, then checked." Inlining without realising would also have to inline a `_persistence.load()` call, which is hard to do accidentally. Open to a renaming PR if you spot one you like — not blocking.

packages/browser/src/sessionid.ts
pauldambra: Keeping the shape — the `if` block now does TWO things (re-check + re-sample `sessionId`/`startTimestamp` after the load to fix the convergent HIGH clobber). A single `const activityTimeout = X && Y && Z()` cannot express that conditional state mutation without becoming uglier (IIFE or extracted helper). The mutation is now load-bearing, not a style choice.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skill not present in this repo, skipped)

No correctness blockers and security is clean. Findings are polish + one comment-accuracy issue and one dedup. The cross-tab test coverage (parameterized legacy/hardened, ACCEPT/REJECT/PRESERVE/KNOWN-LIMITATION) drew praise from both reviewers.

### Key findings

**🟡 MEDIUM**
- **Gating comment overclaims** (`sessionid.ts`) — the cross-tab refresh on the active idle path *and* the self-re-arming idle timer ship to **all** users; only the per-key `refreshKey` mechanism and the `onSessionId` emit are gated on `persistence_save_debounce_ms > 0`. The "new behaviour and bug surface roll out together" comment reads as untrue. Fix the comment (the design is sound — bug fix ships to everyone via the right mechanism per path) or widen the gate.
- **No observability on the path this PR exists for** (`sessionid.ts`) — we log on rotation but nothing when the cross-tab refresh *prevents* a spurious rotation or *adopts* a sibling session. That rare, timing-dependent event is exactly what wants a breadcrumb. Add `logger.info` on both branches.
- **Verbatim duplication** (`sessionid.ts`) — the `if (hardening) refreshKey(SESSION_ID) else flush()+load()` block appears identically in `_isSessionIdleAfterCrossTabRefresh` and `_flushPendingActivityTimestamp`. Extract `_refreshSessionIdFromStorage()` (no flag param — avoids the wrong-abstraction trap). The `_useCrossTabRefreshHardening()` gate appearing in 3 callers is **not** a smell — leave it.

**🟢 LOW / ⚪ NIT**
- Adopted-rotation fires handlers with an all-false `changeReason`; consider naming the condition (`adoptedSiblingSession`) and/or a dedicated reason flag so a future reader of a recorder restart knows why.
- Self-re-arming idle timer is a real (correct) behaviour change — worth a one-line "re-arms until all tabs die" note. `_destroyed` guard is the right fix, well tested.
- Test comments reference "Finding 2" / "Three convergent HIGH findings from QA Swarm" — chat/PR artifacts that belong in the commit message, not source (house rule).

### Convergence
- **Timer re-arm is unbounded by design** — flagged independently by paul + xp. Both conclude it's correct (session genuinely alive); just wants acknowledgement.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 👤 paul | Solid, excellent tests; reconcile the gating comment + add the two log lines, then ship. |
| 📐 xp | Clean; one real once-and-only-once fix (extract the refresh block); the 3-call gate is fine, leave it. |
| 🛡 security-audit | 0 findings — client-side session-id state, no trust boundary; `refreshKey` prop is the hardcoded `SESSION_ID` constant so no prototype-pollution path. |

---

pauldambra:
> [!NOTE]

All qa-swarm findings addressed in `28cee398`:

1. **Dedup** — extracted `_refreshSessionIdFromStorage()`; both `_isSessionIdleAfterCrossTabRefresh` and `_flushPendingActivityTimestamp` now call it (OnceAndOnlyOnce).
2. **Chat-artifact comments** — removed the "Finding 2" / "QA Swarm" references from the test files.
3. **Gating comment** — corrected to state accurately that the cross-tab refresh + idle-timer re-arm run for all callers, and only the per-key mechanism and handler-emit are gated on `persistence_save_debounce_ms > 0`.
4. **Observability** — added `logger.info` on the "cross-tab refresh kept the session alive" and "adopted cross-tab session id" branches.
5. **Naming** — named the `adoptedSiblingSession` condition.
6. **Timer re-arm** — documented that the re-arm is unbounded by design (until all tabs idle / destroy).

Lint failure (unused var) also fixed. 491 affected unit tests green; security-audit found nothing.

pauldambra:
> [!NOTE]

Added a replay-safety regression test in `4342b24a`: `lazy-sessionrecording.test.ts` → *"recorder follows an adopted sibling-tab session id (does not record under the stale id)"*.

It drives the real `SessionIdManager` + recorder: an idle tab whose sibling kept the session alive under a different id adopts that id, and the test asserts the recorder switches `_sessionId` to the adopted id and restarts exactly once (no churn) — so snapshots are stamped with the live session, not the stale one. Closes the gap where cross-tab convergence was only covered at the `SessionIdManager` level, never through the recorder.

pauldambra:
> [!NOTE]
>
> Final pass (paul-reviewer, xp-reviewer, security-audit) on `c071886c`

## Verdict: ✅ APPROVE

Prior-round feedback verified genuinely addressed (shared `_refreshSessionIdFromStorage()`, accurate gating comment, observability logs firing, recorder-convergence regression test passing). No correctness or security findings.

### Findings (all NIT/LOW, all actioned in `c071886c`)
- **NIT** — redundant `valuesChanged ? {...} : undefined` inside the `if (valuesChanged)` handler dispatch → extracted `changeReason` once and reused (also de-duplicates the object literal across dispatch + return).
- **NIT** — the "Same guard pattern as `_flushPendingActivityTimestamp`" comment over-claimed (both re-read, but one bails, one adopts) → reworded.
- **LOW** — the idle-timer re-arm doesn't adopt a sibling's id (only the event path does) → added a comment confirming the asymmetry is deliberate.

### Greptile P2 (now `outdated`)
"`changeReason` silent on cross-tab adoption" — same point as the prior round. The all-false `changeReason` on adoption is **intentional**: it correctly suppresses a spurious `$pageview` and the `$session_ending` replay link while still firing handlers so consumers follow the id. Adoption is signalled via `logger.info('adopted cross-tab session id')` instead. No change.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 👤 paul | clean, ship it — every prior item addressed, not papered over |
| 📐 xp | dedup landed cleanly, no new duplication; one comment reworded |
| 🛡 security-audit | 0 findings — session id is a client-minted correlation id, the new logs leak nothing, no prototype-pollution path |

Also added the missing changeset (`patch`).

---

lricoy (approved):

packages/browser/src/sessionid.ts
```diff
@@ -376,9 +408,20 @@ export class SessionIdManager {
...
+            }
+            if (this._useCrossTabRefreshHardening() && sessionId !== preRefreshSessionId) {
+                // Cross-tab refresh adopted a sibling tab's session id (we
+                // keep our own window id). Fire handlers so downstream
+                // consumers (session recorder, session-scoped props,
+                // $pageview) react to the new id. Gated to debounce>0 to tie
+                // the handler-emit behaviour to the same dated default as the
+                // rest of the hardening.
+                valuesChanged = true
+            }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]
> 🤖 posthog-code agent comment — driven by @pauldambra

addressed in 4616cd2f: `changeReason` now carries an additive `crossTabAdoption: true` when a tab adopts a sibling rotation, handler emission is no longer gated on debounce (an adopted id that handlers never hear about leaves consumers on the old session), and `PageViewManager` clears its pageview state on adoption so `$prev_pageview_*` stats cannot leak across the boundary.

pauldambra:
> [!NOTE]
> 🤖 posthog-code agent comment — driven by @pauldambra
>
> Multi-perspective review: qa-team (compatibility/reliability), paul-reviewer, xp-reviewer. security-audit skipped (PostHog MCP unavailable this run).

Solid fix for a real data-quality bug. The cross-tab refresh design is correct, the test suite is excellent (especially the `KNOWN LIMITATION` pinning pattern), and the extraction of `_refreshSessionIdFromStorage` is clean. One medium observability gap; the rest are low/nit.

### Key findings

- 🟡 MEDIUM — Timer re-arm path has no observability: when a sibling tab keeps the session alive, `_resetIdleTimer` re-arms silently. The only log (`cross-tab refresh kept the session alive`) is on the event path, not the timer re-arm path. Hard to validate the fix is working in the wild.
- 🟢 LOW — `crossTabAdoption?: boolean` is optional while the other three changeReason flags are required. Correct for SDK backwards compat, but the asymmetry will confuse a future reader without a comment explaining why.
- 🟢 LOW — Gradual CDN rollout note: old-SDK tabs won't use the cross-tab refresh, so mixed-version users will still see some fragmentation proportionally to old-SDK tab share. Worth noting in the changeset or PR description.
- ⚪ NIT — The two-path refresh trade-off is explained in multiple comment blocks. The extraction is good; the callers' inline comments can be trimmed.
- ⚪ NIT — `cross-tab-session.spec.ts` blind-spot comment says "known blind spot" without stating the consequence (spurious rotation if sibling activity isn't persisted within 5s).

### Convergence

None — all findings were independent.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 👤 paul | Approves with a medium observability ask and a low type-comment nit; loves the KNOWN LIMITATION test pattern |
| 📐 xp | Approves; notes OnceAndOnlyOnce is good on logic but over-duplicated on comments; praises the parameterized KNOWN LIMITATION pinning |
| 🔍 qa-team/compatibility | Approves; notes mixed-version rollout behavior and optional typing are both correct and acceptable |

---

pauldambra (commented):
QA Swarm inline findings — see below

packages/browser/src/sessionid.ts
```diff
@@ -395,36 +451,46 @@ export class SessionIdManager {
...
-                    windowId,
-                    valuesChanged ? { noSessionId, activityTimeout, sessionPastMaximumLength } : undefined
-                )
-            )
+            this._sessionIdChangedHandlers.forEach((handler) => handler(sessionId, windowId, changeReason))
         }
 
         return {
             sessionId,
             windowId,
```
pauldambra, line 461: > [!NOTE]

**[paul]** 🟡 MEDIUM

i wonder if we have an observability gap on the timer re-arm path. when a sibling tab keeps the session alive, this timer callback just re-arms silently — no log, no counter, nothing to graph. the event path has `logger.info('cross-tab refresh kept the session alive')` but this timer path says nothing.

the whole point of this feature is 'cross-tab kept alive' — without some signal here we can't validate it's working in the wild. not blocking, just: should we be able to observe what we can't easily reproduce?

packages/types/src/session-recording.ts
```diff
@@ -100,7 +100,13 @@ export type CapturedNetworkRequest = Writable<Omit<PerformanceEntry, 'toJSON'>>
...
     sessionId: string,
     windowId: string | null | undefined,
-    changeReason?: { noSessionId: boolean; activityTimeout: boolean; sessionPastMaximumLength: boolean }
+    changeReason?: {
+        noSessionId: boolean
+        activityTimeout: boolean
+        sessionPastMaximumLength: boolean
+        /** This tab picked up a session rotation another tab had already written to storage, instead of minting its own. */
+        crossTabAdoption?: boolean
+    }
```
pauldambra, line 109: > [!NOTE]

**[paul + qa-team/compatibility]** 🟢 LOW

`crossTabAdoption?: boolean` is optional while the other three changeReason flags are required booleans. this is correct for SDK backwards compat — old callers who have `onSessionId` handlers predate this field and would get `undefined` if we made it required. but the asymmetry will confuse a future reader.

the existing JSDoc on the field (`/** This tab picked up... */`) is good. adding one sentence like `// Optional rather than required: external callers on SDK versions before this field was added will receive undefined here.` on the type would prevent someone from 'fixing' it to required and breaking old callers.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
```diff
@@ -0,0 +1,136 @@
...
+const startOptions = {
+    waitForFlags: true,
+    options: {
+        session_idle_timeout_seconds: TEST_TIMEOUT_SECONDS,
+    },
+    flagsResponseOverrides: {
+        sessionRecording: undefined,
+        capturePerformance: true,
+        autocapture_opt_out: true,
+    },
```
pauldambra, line 47: > [!NOTE]

**[xp]** ⚪ NIT

the comment says "known blind spot" but doesn't state the consequence. Code wants to express every idea — the consequence is: if tab B has activity that happened within the last 5s and hasn't been flushed to localStorage yet, tab A won't see it and may still rotate despite tab B being active. Adding that one sentence makes the comment actually inform future readers rather than just warn them.

packages/browser/src/sessionid.ts
pauldambra: > [!NOTE]

intentional suggestion — low priority. the re-arm path was previously a no-op (just 'rotate or don't'); adding observability there is a follow-up concern. deferring to author.

packages/types/src/session-recording.ts
pauldambra: > [!NOTE]

intentional — optional is correct for SDK backwards compat (existing onSessionId handlers predating this field). the existing JSDoc explains the intent. resolving as nit.

packages/browser/playwright/mocked/cross-tab-session.spec.ts
pauldambra: > [!NOTE]

nit — the comment is in a block that explicitly describes known limitations; the consequence is implied by the surrounding context. resolving as non-blocking.

packages/browser/src/sessionid.ts
```diff
@@ -395,36 +451,46 @@ export class SessionIdManager {
...
-            if (this._sessionHasBeenIdleTooLong(new Date().getTime(), lastActivityTimestamp)) {
+            if (this._destroyed) {
+                return
+            }
+            if (this._isSessionIdleAfterCrossTabRefresh(new Date().getTime())) {
                 const idleSessionId = this._sessionId
                 this.resetSessionId()
                 this._eventEmitter.emit('forcedIdleReset', { idleSessionId })
+            } else {
+                // A sibling tab kept the session alive — re-arm so we
```
ksvat, line 485: one day the SDK will be more comment than code

ksvat (approved):

packages/browser/src/sessionid.ts
pauldambra: :see_no_evil:

PR #3692, opened 2026-05-28
fix(replay): fix idle session rotation race condition

ksvat:
## Problem

A customer reported a session replay stuck on "Buffering" for ~5 minutes. Investigating the recording showed the session ID had rotated due to an inactivity gap right before recording resumed, and the first DOM full snapshot landed under the OLD session ID. The new session received only incremental mutations until the next periodic snapshot fired, so rrweb had nothing to paint onto and the player sat spinning.

Root cause: a hole in the `forcedIdleReset` path in `lazy-loaded-session-recorder.ts`.

Two paths normally re-snapshot on session rotation:
- **Path A** — `_onSessionIdCallback` restarts rrweb when `_isIdle === true`
- **Path B** — `_updateWindowAndSessionIds` (driven by `onRRwebEmit`) restarts rrweb on session change for non-idle rotations

When the session manager's idle enforcement timer fires `forcedIdleReset`, the listener sets `_isIdle = 'unknown'` and calls `stop()` BEFORE the new session id is observed. When `_onSessionIdCallback` subsequently fires for the rotated id:
- `_isIdle` is `'unknown'`, not `true` → Path A's guard misses
- rrweb is already torn down → Path B can never fire (no events to emit)

The recorder ends up running under the new session id with no rrweb instance and no full snapshot, until something else restarts recording.

## Changes

- `_onSessionIdCallback`: extend the restart guard to also fire when rrweb isn't running (`!this.isStarted`). `stop()` is safe to call when already stopped, and the `isStarted` getter already exists at line 867.
- Add a regression test that simulates the `forcedIdleReset` path and asserts `rrweb.record` is called a second time after the rotation.

Diff is tiny — one boolean condition + one test. No new abstractions.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

- [ ] Ran \`pnpm changeset\` to generate a changeset file

---

### DIFF
--- a/.changeset/fix-replay-idle-rotation-snapshot.md
+++ b/.changeset/fix-replay-idle-rotation-snapshot.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): take a fresh full snapshot after session ID rotates via `forcedIdleReset`. Previously, when the session manager's idle enforcement timer rotated the session id, the recorder tore down rrweb and set `_isIdle = 'unknown'` before the new session id was observed. Neither restart path then fired (the `_onSessionIdCallback` guard only restarted when `_isIdle === true`, and `_updateWindowAndSessionIds` could not run with rrweb stopped), so the new session received only incremental mutations until a later snapshot — leaving the player stuck on "Buffering". The restart guard now also fires when rrweb isn't running.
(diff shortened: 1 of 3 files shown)


ksvat:
* **#3692** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3692?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3692?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

TueHaulund (approved):
I think this is it. The _final_ bandaid :-D

PR #3694, opened 2026-05-28
fix(react-native): preserve property types for flag evaluation

gustavohstrassburger:
## Summary

- After PR #2564, `posthog-react-native` (4.43.x) started attaching cached person/group properties to every `/flags` request — but the RN call sites coerced every value via `String(value)`. Booleans became `"true"`, arrays became `"wl"`, nested objects became `"[object Object]"`. Flag conditions using boolean equality or array `contains` then failed on device while the PostHog UI still evaluated correctly against the properly-typed stored profile, producing a device-vs-UI split for the same user / flag / moment after upgrading 4.10.5 → 4.43.x.
- The fix drops the `String()` coercion at the three internal call sites (`identify()`, `group()`, `_setDefaultPersonPropertiesForFlags()`) and widens the accumulator types to `Record<string, JsonType>`. Core's `setPersonPropertiesForFlags` already accepted `JsonType`; this PR also widens core's `setGroupPropertiesForFlags` to match.
- Web SDK already preserved types end-to-end, which is exactly why UI eval and device eval diverged.

## Workaround for customers blocked on a release

```ts
PostHog.init({ ..., setDefaultPersonProperties: false })
```

Reverts `/flags` payload behavior to pre-4.10.5 (empty `person_properties`); server resolves flags from the correctly-typed stored profile.

## Test plan

- [x] New unit tests in `packages/react-native/test/posthog.spec.ts` assert booleans, arrays, numbers, and nested objects survive through `identify($set / $set_once)` and `group()`, with explicit regression guards (`.not.toBe('true')`, `.not.toBe('wl')`).
- [x] Verified the new tests **fail** against the pre-fix source (failure output shows the exact bug shape: `hasApprovedSubscription: "true"`, `user_categories: "wl"`, `tags: "enterprise,beta"`, `metadata: "[object Object]"`).
- [x] Verified the new tests **pass** with the fix.
- [x] Updated two pre-existing tests that asserted the buggy coercion (`employees: '50'` → `employees: 50`).
- [x] `pnpm -F posthog-react-native build` clean, `pnpm -F posthog-react-native lint` clean.
- [x] Full test suites green: RN 335/335, core 587/587, node 583/583.
- [ ] Manual repro on a sample RN app pinned to this build against a staging project with a flag that uses boolean equality + array `contains` — confirm `/flags` payload `person_properties` contains `true` (boolean) and `["wl"]` (array), and the SDK now returns the same variant the PostHog UI returns.

### DIFF
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -701,7 +701,7 @@ export class PostHog extends PostHogCore {
    * @param reloadFeatureFlags Whether to reload feature flags after setting the properties. Defaults to true.
    */
   private _setDefaultPersonPropertiesForFlags(reloadFeatureFlags = true): void {
-    const defaultProps: Record<string, string> = {}
+    const defaultProps: Record<string, JsonType> = {}
     const relevantKeys = [
       '$app_version',
       '$app_build',
@@ -714,16 +714,16 @@ export class PostHog extends PostHogCore {
     relevantKeys.forEach((key) => {
       const value = this._appProperties[key]
       if (value !== null && value !== undefined) {
-        defaultProps[key] = String(value)
+        defaultProps[key] = value
       }
     })
 
     const commonProps = this.getCommonEventProperties()
     if (commonProps.$lib) {
-      defaultProps.$lib = String(commonProps.$lib)
+      defaultProps.$lib = commonProps.$lib as JsonType
     }
     if (commonProps.$lib_version) {
-      defaultProps.$lib_version = String(commonProps.$lib_version)
+      defaultProps.$lib_version = commonProps.$lib_version as JsonType
     }
 
     if (Object.keys(defaultProps).length > 0) {
@@ -1041,11 +1041,11 @@ export class PostHog extends PostHogCore {
 
     // Automatically cache group properties for feature flag evaluation
     if (properties && Object.keys(properties).length > 0) {
-      const propsToCache: Record<string, string> = {}
+      const propsToCache: Record<string, JsonType> = {}
       Object.keys(properties).forEach((key) => {
         const value = properties[key]
         if (value !== null && value !== undefined) {
-          propsToCache[key] = String(value)
+          propsToCache[key] = value as JsonType
         }
       })
       if (Object.keys(propsToCache).length > 0) {
@@ -1188,7 +1188,7 @@ export class PostHog extends PostHogCore {
    * @param properties The group properties to set for flag evaluation
    * @param reloadFeatureFlags Whether to reload feature flags after setting the properties. Defaults to true.
    */
-  setGroupPropertiesForFlags(properties: Record<string, Record<string, string>>, reloadFeatureFlags = true): void {
+  setGroupPropertiesForFlags(properties: Record<string, Record<string, JsonType>>, reloadFeatureFlags = true): void {
     super.setGroupPropertiesForFlags(properties)
 
     if (reloadFeatureFlags) {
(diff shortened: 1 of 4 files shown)


packages/react-native/src/posthog-rn.ts
```diff
@@ -714,16 +714,16 @@ export class PostHog extends PostHogCore {
...
       if (value !== null && value !== undefined) {
-        defaultProps[key] = String(value)
+        defaultProps[key] = value
       }
     })
 
     const commonProps = this.getCommonEventProperties()
     if (commonProps.$lib) {
-      defaultProps.$lib = String(commonProps.$lib)
+      defaultProps.$lib = commonProps.$lib as JsonType
```
dustinbyrne: nit: given `getCommonEventProperties` returns type `PostHogEventProperties`, `commonProps.$lib` should already be typed as `JsonType`

packages/react-native/src/posthog-rn.ts
```diff
@@ -714,16 +714,16 @@ export class PostHog extends PostHogCore {
...
     }
     if (commonProps.$lib_version) {
-      defaultProps.$lib_version = String(commonProps.$lib_version)
+      defaultProps.$lib_version = commonProps.$lib_version as JsonType
```
dustinbyrne: same goes here

packages/react-native/src/posthog-rn.ts
```diff
@@ -1041,11 +1041,11 @@ export class PostHog extends PostHogCore {
 
     // Automatically cache group properties for feature flag evaluation
     if (properties && Object.keys(properties).length > 0) {
-      const propsToCache: Record<string, string> = {}
+      const propsToCache: Record<string, JsonType> = {}
       Object.keys(properties).forEach((key) => {
         const value = properties[key]
         if (value !== null && value !== undefined) {
-          propsToCache[key] = String(value)
+          propsToCache[key] = value as JsonType
```
dustinbyrne: same goes here

packages/react-native/src/posthog-rn.ts
```diff
@@ -1513,20 +1513,20 @@ export class PostHog extends PostHogCore {
...
-    const propsToCache: Record<string, string> = {}
+    const propsToCache: Record<string, JsonType> = {}
     if (userProps) {
       Object.entries(userProps).forEach(([key, value]) => {
         if (value !== null && value !== undefined) {
-          propsToCache[key] = String(value)
+          propsToCache[key] = value as JsonType
         }
       })
     }
```
dustinbyrne: `value` requires a cast because the type of `userProps` is ambiguous at the time we enumerate it
```suggestion
    if (userProps && typeof userProps === 'object' && !Array.isArray(userProps)) {
      Object.entries(userProps).forEach(([key, value]) => {
        if (value !== null && value !== undefined) {
          propsToCache[key] = value
        }
      })
    }
```

packages/react-native/src/posthog-rn.ts
```diff
@@ -1513,20 +1513,20 @@ export class PostHog extends PostHogCore {
...
-    const propsOnceToCache: Record<string, string> = {}
+    const propsOnceToCache: Record<string, JsonType> = {}
     if (userPropsOnce && typeof userPropsOnce === 'object') {
       Object.entries(userPropsOnce as Record<string, unknown>).forEach(([key, value]) => {
         if (value !== null && value !== undefined) {
-          propsOnceToCache[key] = String(value)
+          propsOnceToCache[key] = value as JsonType
         }
       })
     }
```
dustinbyrne: same here sort of
```suggestion
    if (userPropsOnce && typeof userPropsOnce === 'object' && !Array.isArray(userPropsOnce)) {
      Object.entries(userPropsOnce).forEach(([key, value]) => {
        if (value !== null && value !== undefined) {
          propsOnceToCache[key] = value
        }
      })
    }
```

dustinbyrne (approved):
looks good, would prefer if we could drop all the explicit typecasts before merging as they may mask a future type issue

dustinbyrne:
oh and please add a changeset before merging or this will not release

PR #3693, opened 2026-05-28
fix(core): stop leaking DOM globals into public type surface

dustinbyrne:
## Problem

The `Bump versions and commit to main` step of the release workflow was failing because `posthog-react-native#generate-references` errored with:

> ERROR: Internal Error: Unable to follow symbol for "ErrorEvent"

After fixing that, the same failure surfaced for `PromiseRejectionEvent`. (Failed run: https://github.com/PostHog/posthog-js/actions/runs/26583417207)

Root cause: PR #3681 (`fix: unify capture exception`) added `createErrorPropertiesBuilder` to `posthog-rn.ts`, which references `CoreErrorTracking.ErrorEventCoercer` / `PromiseRejectionEventCoercer`. Those classes are typed `ErrorTrackingCoercer<ErrorEvent>` / `ErrorTrackingCoercer<PromiseRejectionEvent | EventWithDetailReason>` in `@posthog/core`. The emitted `.d.ts` files therefore reference DOM-only globals (`ErrorEvent`, `PromiseRejectionEvent`, `Event`). When api-extractor analyses the react-native package — whose tsconfig `lib` is `["ESNext"]` and excludes DOM — it cannot resolve those symbols and aborts.

The browser package (`posthog-js`) doesn't hit this because its tsconfig includes DOM. The release-workflow bump step still fails because it runs `generate-references` across all three publishable packages.

## Changes

Replace the DOM-typed generic parameters on `ErrorEventCoercer` and `PromiseRejectionEventCoercer` with structural subsets that capture only the fields actually read at runtime (`message`, `error`, `reason`, `detail.reason`). Runtime behaviour, `match()` semantics, and the public class shapes are unchanged; the emitted `.d.ts` files no longer reference DOM globals, so api-extractor can analyse them in a non-DOM tsconfig.

Verified locally:
- `pnpm exec turbo --filter=posthog-js --filter=posthog-node --filter=posthog-react-native generate-references` now passes for all three packages.
- `@posthog/core` unit tests: 587 passed.
- `posthog-react-native` unit tests: 333 passed.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
[...]

PR #3697, opened 2026-05-28
ci: run generate-references on PRs

dustinbyrne:
## Problem

The release workflow's `Bump versions and commit to main` step runs `pnpm bump && pnpm generate-references`. If `generate-references` (which runs `api-extractor` over `posthog-js`, `posthog-node`, and `posthog-react-native`) fails, the release is blocked — but the failure isn't visible until release time.

PR #3681 introduced a type-extraction breakage (`@posthog/core` leaked DOM globals into its public type surface; api-extractor couldn't follow them when analysing `posthog-react-native`'s non-DOM tsconfig). The release ran, the bump step failed, and #3693 had to chase it down after the fact. Running the same command on the PR would have caught it pre-merge.

## Changes

Add a `generate-references` job to `library-ci.yml`. Same machine class as the other PR jobs; uses the shared setup action; runs the existing root-level `pnpm generate-references` script (which fans out to `posthog-js`, `posthog-node`, `posthog-react-native` via turbo). The job's only purpose is to fail when api-extractor errors — no commits, no artifacts, no special tokens.

Local runtime: ~9s end-to-end.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

(CI-only change — no library version bump.)

## Checklist

- [x] Tests for new code (the job itself is the check)

### If releasing new changes

- [ ] Ran \`pnpm changeset\` to generate a changeset file (n/a — CI-only)

---

### DIFF
--- a/.github/workflows/library-ci.yml
+++ b/.github/workflows/library-ci.yml
@@ -295,3 +295,21 @@ jobs:
           WRITE_MANGLED_PROPERTIES: 1
 
       - run: git diff --exit-code # fail if e.g. the mangled properties list has changed, see rollup.config.js
+
+  generate-references:
+    # Runs the same api-extractor pipeline the release workflow uses to bump
+    # versions, so type-extraction breakages (e.g. unresolved DOM globals in
+    # `@posthog/core` types when extracted from a non-DOM tsconfig) surface
+    # on the PR instead of blocking the release.
+    name: Generate references
+    runs-on: ubuntu-22.04
+    permissions:
+      contents: read
+    steps:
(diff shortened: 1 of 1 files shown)


dustinbyrne:
failing CI as intended - we should mark this a required check imo

marandaneto (approved):

PR #3698, opened 2026-05-28
fix(replay): forward (url, init) instead of new Request on Safari

HaynesPostHog:
## Summary

- Safari refuses stream uploads. Both fetch wrappers (session-replay network plugin and tracing-headers) reconstructed `new Request(url, init)` from `(url, init)` and forwarded the `Request` to native fetch. Per the Fetch spec, constructing a Request with a string body stores it as a `ReadableStream`; forwarding that Request makes Safari throw `NotSupportedError: ReadableStream uploading is not supported`.
- Fix: forward the caller's `(url, init)` for non-Request inputs so the string body stays a string. Keep forwarding `req` when the caller passed a `Request` directly (constructing the local `req` consumed the caller's body, so `req` is the only usable copy, and the caller already chose a Request body).
- Affects any Safari user with `__add_tracing_headers` or `session_recording.recordBody` enabled; Chrome/Firefox unaffected. Both options are off by default.

## Test plan

- [x] Browser unit tests pass (99 affected, including un-skipped issue #1326 contract)
- [x] Webkit Playwright pass: `tracing-headers.spec.ts` (5, including new POST-with-string-body Safari regression)
- [x] Webkit Playwright pass: `csrf-headers-preserved.spec.ts` (9, including double-wrap)
- [x] Webkit Playwright pass: `session-recording-network-recorder.spec.ts` (4)
- [ ] Manual repro in Safari: PostHog init with `__add_tracing_headers` and `session_recording: { recordBody: true }`, then `await fetch('/anything', { method: 'POST', body: JSON.stringify({a:1}) })` — should succeed (throws `NotSupportedError` before the fix)
- [ ] Confirm Chrome behaviour unchanged
- [ ] Confirm session-replay body capture still records the request body
- [ ] Confirm tracing headers still added on outgoing requests for listed domains

---

### DIFF
--- a/packages/browser/playwright/mocked/tracing-headers.spec.ts
+++ b/packages/browser/playwright/mocked/tracing-headers.spec.ts
@@ -88,4 +88,49 @@ test.describe('tracing headers', () => {
             expect(headers['x-posthog-window-id']).toBeUndefined()
         })
     }
+
+    // Safari refuses stream uploads. The fetch wrapper used to construct `new Request(url, init)`
+    // and forward that Request — which upgraded a string body to a ReadableStream and triggered
+    // `NotSupportedError: ReadableStream uploading is not supported` on Safari/webkit.
+    test('POST with string body succeeds on webkit and preserves the body', async ({ page, context, browserName }) => {
+        test.skip(browserName !== 'webkit', 'Safari-only regression: stream-upload rejection only happens on webkit')
+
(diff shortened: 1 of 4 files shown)


marandaneto:
@dustinbyrne is working on this here https://github.com/PostHog/posthog-js/pull/3706

PR #3701, opened 2026-05-28
perf(react-native): coalesce storage writes with a debounce window

turnipdabeets:
## Problem

React Native apps that fire many events quickly (e.g. a screen that captures several events on navigation) can stutter on low-end Android. A customer is experiencing performance implications on low end androids (#3493).

It comes from how the RN SDK stores data. Native iOS/Android write one file per event. The RN SDK instead keeps everything in a single in-memory object and, on **every** change, serializes that whole object and writes the entire blob to disk (AsyncStorage). So a burst of N captures re-serializes a growing blob N times (O(n²) bytes) and does N disk writes — all on the JS thread. Reads aren't affected; they come from the in-memory copy. Only the disk write is costly.

## Changes

Batch the disk writes instead of writing on every change:

- A change updates the in-memory copy immediately (reads unchanged), but the disk write is scheduled **100ms** later. Other changes in that window join the same write, so a burst becomes one write instead of one per event.
- Durability is kept: paths that must be on disk *now* force the write synchronously — events flush, app background, shutdown, and **login/logout** (so a crash right after logout can't leave the previous user's identity on disk).

Two write triggers, whichever comes first wins:
- The SDK already flushes events every **20** (`flushAt`) or every 10s, and a flush forces a write — so in a fast burst you get ~1 write per 20 events.
- The **100ms** timer covers the gaps between flushes so nothing waits too long.

Measured on a Pixel (100 captures in a tight loop): **203 writes / 8.7 MB → 1 write / 85 KB**.

Why 100ms: a small simulation swept window sizes against realistic traffic. Beyond ~100ms there's no further reduction (the write count is already floored by `flushAt`), while a larger window only widens how much could be lost if the app is killed before the timer fires. 100ms reaches that floor with the smallest loss window.

Tests are TDD-style — each behavior has a test confirmed to fail without the change. 345 RN + 587 core pass. Also adds a "Storage" demo screen to `example-expo-53`.

### Libraries affected
- [x] posthog-react-native

## Checklist
- [x] Tests for new code

### If releasing new changes


marandaneto:
`captureException` should be sync or awaited and bypass the debounce window otherwise we'd miss exceptions when the app is about to crash
also i'd test this if now we are gonna miss logs, events, etc every time if we are about to crash

turnipdabeets:
> `captureException` should be sync or awaited and bypass the debounce window otherwise we'd miss exceptions when the app is about to crash also i'd test this if now we are gonna miss logs, events, etc every time if we are about to crash

@marandaneto Added a sync drain on fatal exceptions in `captureException`

marandaneto:
PR is very chatty with very long comments, can we simplify?

packages/react-native/src/storage.ts
```diff
@@ -16,6 +32,12 @@ export class PostHogRNStorage {
   preloadPromise: Promise<void> | undefined
   private _storageKey: string
   private _pendingPromises: Set<Promise<void>> = new Set()
```
dustinbyrne, line 34: I don't know that the concept of multiple pending promises makes sense anymore, considering we want to limit the maximum number of times we're writing to once per 100ms.

Allowing multiple in-flight writes seems like it could introduce ordering issues - though, the actual risk of this seems relatively low?

dustinbyrne (approved):
Makes sense to me. I think the decision to debounce persistence writes over 100ms windows is a good call. Definitely an improvement over batching once per tick of the event loop 👍

packages/react-native/src/storage.ts
turnipdabeets: Agreed the actual risk is small since memoryCache is the authoritative latest state. And we can keep it for ergonomics.

PR #3705, opened 2026-05-29
feat(flags): support early_exit in posthog-node local evaluation

gustavohstrassburger:
## Problem

PostHog/posthog#57321 added an **early exit** option to feature flags: when enabled, condition evaluation stops and returns `false` as soon as a user matches a group's targeting but is excluded by its rollout percentage, instead of falling through to later groups. That PR implemented the behavior server-side (Rust engine), but `posthog-node`'s local evaluation re-implements the matching loop in TypeScript and did not honor the new `filters.early_exit` field — so locally-evaluated flags could diverge from server-side results.

## Changes

- Add `early_exit?: boolean` to the feature-flag `filters` type.
- Honor it in local evaluation: when enabled and a condition group's property filters match but rollout excludes the user (the engine's `OutOfRolloutBound` case), return `false` immediately rather than evaluating later groups. Property mismatches still fall through, mirroring the Rust semantics exactly.
- Tests for early-exit on, default off (regression), and the property-mismatch case.

The `/flags/definitions` endpoint already returns `filters.early_exit` (the serializer dumps the full `filters` dict verbatim), so no SDK fetch/parse changes were needed.

### Libraries affected

- [x] posthog-node

## Checklist

- [x] Tests for new code

### If releasing new changes

Added unit tests in `feature-flags.spec.ts` covering early-exit enabled (returns `false` without evaluating later groups), default/unset (falls through to a later matching group), and a property-mismatch case (does not early exit). Full `posthog-node` suite passes (586 tests); lint and build are clean.

---

### DIFF
--- a/packages/node/src/types.ts
+++ b/packages/node/src/types.ts
@@ -282,7 +282,7 @@ export type PostHogFeatureFlag = {
     payloads?: Record<string, string>
     // Flag-level toggle: when true, condition evaluation stops and returns false as soon as a
     // group's property filters match but the rollout percentage excludes the user, rather than
-    // continuing to evaluate later groups. Mirrors the server-side (Rust) early-exit behavior.
+    // continuing to evaluate later groups.
     early_exit?: boolean
   }
   deleted: boolean
--- a/packages/node/src/types.ts
+++ b/packages/node/src/types.ts
@@ -280,6 +280,10 @@ export type PostHogFeatureFlag = {
       }[]
     }
     payloads?: Record<string, string>
+    // Flag-level toggle: when true, condition evaluation stops and returns false as soon as a
+    // group's property filters match but the rollout percentage excludes the user, rather than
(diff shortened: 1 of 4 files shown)


packages/node/src/types.ts
```diff
@@ -280,6 +280,10 @@ export type PostHogFeatureFlag = {
       }[]
     }
     payloads?: Record<string, string>
+    // Flag-level toggle: when true, condition evaluation stops and returns false as soon as a
+    // group's property filters match but the rollout percentage excludes the user, rather than
+    // continuing to evaluate later groups. Mirrors the server-side (Rust) early-exit behavior.
```
marandaneto:
```suggestion
    // continuing to evaluate later groups.
```

marandaneto: no need otherwise all SDKs are adding this at some point
usually those comments get stale over time

marandaneto (approved):

Issue #2866, opened 2026-01-09
ERR_BLOCKED_BY_CLIENT with AdBlocker


dbauszus-glx:
Posthog should be able to check whether it is possible to pass a client AdBlocker.

We check feature flags as outlined in the documentation which works.

```js
const response = await fetch("https://eu.i.posthog.com/flags?v=2", {
    method: "POST",
    headers: {
        "Content-Type": "application/json",
        // Browser automatically adds Origin, Referer, Sec-Fetch-* headers
    },
    body: JSON.stringify({
        api_key: "phc_hw1YFnOqlGTj6J36D18zWkMWCM18YQwhEHcSpQlEr1o",
        distinct_id: "user-id"
    })
});
```

We then await the posthog.init method which works.

Posthog then tries fetch assests. This call is blocked by the AdBlocker and fails.

<img width="725" height="247" alt="Image" src="https://github.com/user-attachments/assets/445cfe93-c90a-4476-9011-d48b5312d785" />

Posthog will then continue to throw an Error and retry every request with the same result.

The call to the decide endpoint should check whether assets can be loaded.

The init method should abort if assests can not be loaded.

dbauszus-glx:
A note in regards to the domains.

We are able to connect to the `https://eu.posthog.com` domain through the AdBlocker. However assets are loaded from `https://eu-assets.i.posthog.com` domain which is blocked.

As a workaround we check flags from `https://eu.i.posthog.com/decide` which will fail, and then abort the init.

dbauszus-glx:
I think we should bail and warn allowing the client to take appropriate steps, ie disable adblocker.

ctsstc:
I setup the reverse proxy and it seems some things are getting through but not everything. I'm assuming due to the name of the resource being requested. I didn't try before adding the proxy unfortunately, so maybe nothing changed. But it would be nice if we could obfuscate the file names. I don't assume they're doing any pattern matting on the file contents given the file is completely blocked.

This is via: ublock origin lite

<img width="745" height="249" alt="Image" src="https://github.com/user-attachments/assets/9b93904a-7810-4fe2-ae9a-738b80bd738c" />

Fansy-lab:
Same issue as @ctsstc. Some js files just don't load even with reverse proxy probbaly because of the naming

juanpablob:
[...]

PR #3706, opened 2026-05-29
fix(browser): avoid Safari ReadableStream upload with fetch wrappers

dustinbyrne:
## Problem

Safari can throw `NotSupportedError: ReadableStream uploading is not supported` when PostHog fetch wrappers pass an internally-created `Request` object to downstream fetch wrappers/interceptors. Those downstream wrappers can forward `request.body`, which is a `ReadableStream`, even when the app originally called `fetch(url, { method: 'POST', body: 'some string' })`.

## Changes

- Avoid passing internally-created `Request` objects downstream for original `fetch(url, init)` calls in tracing headers and replay network capture wrappers.
- Keep using a cloned `Request` for original `fetch(Request)` calls, since creating the clone can consume the original request body.
- Add Playwright coverage for a Safari/WebKit regression with an inner fetch wrapper that forwards `Request.body`.
- Update fetch wrapper invariant tests for the new downstream behavior.

Validated locally:

- `pnpm --dir packages/browser typecheck`
- `pnpm --dir packages/browser exec eslint src/entrypoints/tracing-headers.ts src/extensions/replay/external/network-plugin.ts src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts playwright/mocked/fetch-post-body.spec.ts`
- `cd packages/browser && pnpm exec jest src/__tests__/tracing-headers.test.ts src/__tests__/extensions/replay/external/fetch-wrapper-invariants.test.ts --runInBand`
- `cd packages/browser && pnpm exec playwright test playwright/mocked/fetch-post-body.spec.ts --project webkit --project chromium --project firefox --reporter=line`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/entrypoints/tracing-headers.ts
```diff
@@ -49,9 +50,18 @@ const patchFetch = (hostnames: string[], distinctId: string, sessionManager?: Se
...
+            if (url instanceof Request) {
+                addTracingHeaders(hostnames, distinctId, sessionManager, req.url, req.headers)
+                return originalFetch(req)
+            }
+
+            const headers = new Headers(init?.headers)
+            addTracingHeaders(hostnames, distinctId, sessionManager, req.url, headers)
 
-            return originalFetch(req)
+            return originalFetch(url, { ...(init || {}), headers })
```
dustinbyrne: Does anyone foresee any issues with spreading `init` like this?

ikrishanand:
@marandaneto  @HaynesPostHog @dustinbyrne When i can expect this fix to go live

packages/browser/src/entrypoints/tracing-headers.ts
marandaneto: no idea tbh, what my agent says is:

> I think it’s fine for normal fetch(url, { ...plain RequestInit }) usage, but yes, there’s a narrow edge-case problem.
> 
>  RequestInit is a WebIDL dictionary: native fetch reads known fields by property lookup. Object spread only copies own
>  enumerable properties. So this can drop values if someone passes a non-plain init object, e.g.:
> 
>  - class instance with get method() { ... }
>  - object with inherited/non-enumerable body, signal, credentials, etc.
>  - a Request object used as the second argument
> 
>  In those cases, headers is preserved because the code explicitly reads init?.headers, but other fields like method, body,
>  signal, duplex, etc. could be lost.
> 
>  For the common case, this is very unlikely to break anything, and the approach is reasonable for fixing the Safari
>  ReadableStream issue. But since this wraps global fetch, I’d mention the edge case in review and maybe add/adjust a test if we
>  want to be strict about preserving native fetch semantics.

ikrishanand:
@marandaneto @HaynesPostHog @dustinbyrne Our experience has been very disappointing so far, https://github.com/PostHog/posthog-js/issues/2866 is this issue even being tracked? /s/?ip=0& this exact path is blocked for me, I don't see a use for proxy domain

fasyy612 (approved):

packages/browser/src/entrypoints/tracing-headers.ts
marandaneto: ok i burnt a few more tokens comparing some other approaches with ours (eg otel and others)
looks like all of them have trade-offs, edge cases
the result is https://github.com/PostHog/posthog-js/pull/3711
this pr is on top of yours, so feel free to incorporate or drop it

dustinbyrne: thanks!! I've incorporated it

marandaneto (approved):

dustinbyrne:
Hi @ikrishanand, I'll merge this once it's CI passing, and it should be released shortly after

PR #1038, opened 2024-02-22
feat: improve user agent detection

pauldambra:
A user reports that they had significantly more matches on "desktop" as a type than "mobile" as a type after moving from another provider. See https://posthoghelp.zendesk.com/agent/tickets/10775 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/13369)

We default to "desktop" for user agents we can't identify so this points at their user base having a high proportion of useragents we don't match

They're on an older version of posthog-js that doesn't report raw user agent so we can't confirm that guess

Until they update we can target general improvements

This PR copies opensource test cases from other user agent matcher and uses them to improve our matching.

In this PR we 

* don't try to attempt to start matching smart tvs
* improve matching for OS and device
* add matching for watchOS
* adds <2k rather than the 6k of ua-parser-js
* pre-uglify the code to help roll-up minify it more by making many constants for repeated strings and using old fashioned concatenation to concat them together

## some notes as I worked on it

* started with 88 of 427 tests failing
    * these are heavily skewed towards wearables and smart tvs which we've never tried to detect
* ignoring smart tv and wearables 
    * only 42 of 427 tests failing
* so we detect 90% the same as ua-parser-js to start with 👍 
* we can skip embedded devices like echo dot
* distinguishing between android tablet and mobile has quite wide variation
* using `$blah $foo` doesn't necessarily help minification as it converts to `"".concat(blah, foo)` and those 7 characters of `.concat` might outweigh gain from constantising. just doing `blah + ' ' + foo` does work though

PR #3708, opened 2026-05-29
feat(core): detect Brave, Vivaldi, Yandex, Whale, DuckDuckGo, Pale Moon, Waterfox

pauldambra:
## Problem

`detectBrowser` in `@posthog/core` hadn't been meaningfully extended since 2024-02 (#1038). As a result, anyone on Brave, Vivaldi, Yandex Browser, Naver Whale, DuckDuckGo, Pale Moon or Waterfox was getting bucketed as `Chrome` or `Firefox` — making `$browser`-based breakdowns blind to a real (and growing) slice of traffic.

## Changes

**UA-string detection** added for the Chromium and Firefox forks that stamp themselves into the UA:

| Browser | UA marker | Inserted before |
| --- | --- | --- |
| Vivaldi | `Vivaldi/` | Chrome check |
| Yandex Browser | `YaBrowser/` | Chrome check |
| Naver Whale | `Whale/` | Chrome check |
| DuckDuckGo | `DuckDuckGo/` (Android/desktop), `Ddg/` (iOS) | Chrome / Safari check |
| Brave (iOS only) | `Brave/` | Safari check |
| Pale Moon | `PaleMoon/` | Firefox check |
| Waterfox | `Waterfox/` | Firefox check |

**Hint-based detection** for desktop / Android Brave, which is Chromium-based and carries no UA marker:

- `detectBrowser` / `detectBrowserVersion` accept an optional 3rd argument, `BrowserDetectionHints`, with a single `brave` flag (from `navigator.brave`).
- The browser SDK collects it automatically via a `getBrowserDetectionHints()` helper and passes it through `getEventProperties()`. No caller changes required; the original two-argument signature still works for non-DOM callers.
- Desktop / Android Brave has no UA version marker, so `$browser_version` is `null` for those (honest rather than a fake Chrome version). Brave on iOS does carry `Brave/X`, so its version is parsed from the UA.

### On Arc (deliberately dropped)

Arc was in the original scope but has been **removed**. It presents as plain Chromium in the UA *and* in `navigator.userAgentData.brands`, and — confirmed by installing Arc — it does **not** inject its `--arc-palette-*` CSS custom properties by default (those only appear when a user applies a theme/Boost to a specific site). There is no reliable synchronous in-page signal for Arc, so detecting it would have meant a per-event `getComputedStyle` cost plus a page-controllable false-positive vector (any site that set that CSS var) for almost no real true positives. Better to not claim support than to mis-bucket.

### Tests

- A real-world UA per added browser feeding the existing `test.each(browserTestcases)` matrix, so both `detectBrowser` and `detectBrowserVersion` are checked.
- Waterfox's classic-numeric vs `Waterfox/G6.0.13` versioning (the latter is recognised as Waterfox; version is `null`).
- A `detectBrowser with out-of-band hints` block: no hints, empty hints, desktop/Android Brave via the `brave` hint, hint precedence over UA, and Brave-on-iOS via the UA marker (no hint needed).
- A `detectBrowserVersion with out-of-band hints` block: desktop/Android Brave → `null`, Brave-on-iOS version from the UA marker, UA-only fallback when hints absent.
- End-to-end `getEventProperties` wiring for Brave (hint → `$browser: Brave`, `$browser_version: null`).


packages/core/src/utils/user-agent-utils.ts
```diff
@@ -50,10 +50,85 @@ const GENERIC_MOBILE = GENERIC + ' ' + MOBILE.toLowerCase()
...
+  // forks (Arc, etc.) that surface themselves only through Client Hints.
+  userAgentDataBrands?: UserAgentBrand[]
+  // Pass `!!navigator.brave` here. The Brave object exists on all Brave
+  // builds and is the recommended sync detection signal.
+  brave?: boolean
+}
+
+// Brand names we recognise in `navigator.userAgentData.brands`. Comparison is
+// case-insensitive against the brand field.
+const CLIENT_HINTS_BRAND_MAP: Record<string, string> = {
```
pauldambra: > [!NOTE]

**[qa-team/data-integrity]** 🟠 HIGH

Arc detection rests entirely on an `"Arc"` entry showing up in `navigator.userAgentData.brands` (via this `CLIENT_HINTS_BRAND_MAP`). The worry: in production Arc deliberately mimics Chrome and (per community detection write-ups) does **not** advertise `"Arc"` in client-hint brands — it's usually detected via signals like a `--arc-palette-title` CSS var or `window.arc`. The end-to-end Arc test passes only because it injects a synthetic `{brand:'Arc'}` that may not occur in the wild.

Could you confirm against a **captured real Arc UA + client-hints dump** that `brands` actually contains `"Arc"`? If it doesn't, Arc users stay bucketed as Chrome despite this change. (Brave genuinely does report `"Brave"` in brands, so that half is fine.)

Also minor design note (convergent: paul + xp): this map reads like the general "add a browser via client hints" mechanism, but in practice it only does Arc (Brave short-circuits on the `brave` flag) while everything else goes through the UA-string ladder. Worth a one-liner so the next person knows which path to extend.

packages/browser/src/utils/event-utils.ts
```diff
@@ -261,6 +268,25 @@ export function getTimezoneOffset(): number | undefined {
...
+    const nav = typeof navigator !== 'undefined' ? (navigator as Record<string, any>) : undefined
+    if (!nav) {
+        return {}
+    }
+    const hints: BrowserDetectionHints = {}
+    const brands = nav.userAgentData?.brands
+    if (brands) {
+        hints.userAgentDataBrands = brands
+    }
+    if (nav.brave) {
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/frontend]** 🟡 MEDIUM

`if (nav.brave)` treats the mere presence of the `navigator.brave` object as "this is Brave". But `navigator.brave` is an object exposing an async `isBrave()` promise — presence is the conventional *sync* heuristic, and the authoritative check is `await navigator.brave.isBrave()`. If that object ever exists in a Brave-derived/embedded webview that isn't user-facing Brave, those users silently become `$browser: Brave` and nothing errors.

Reasonable trade-off for a per-event hot path, but worth (a) a short comment noting the async confirm was deliberately skipped, and (b) a thought on how we'd notice in prod if the brave-hint path starts misfiring.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -171,8 +288,20 @@ const versionRegexes: Record<string, RegExp[]> = {
...
+  vendor: string | undefined,
+  hints?: BrowserDetectionHints
+): number | null {
+  const browser = detectBrowser(userAgent, vendor, hints)
+
+  // Browsers detected purely through Client Hints have no UA marker we can
+  // regex against — read the version directly from the brands entry. Brave
+  // hides itself from `brands` by default so it usually returns null here.
+  if (browser === ARC || browser === BRAVE) {
+    return brandVersionFromHints(hints, browser)
```
pauldambra: > [!NOTE]

For Brave, `$browser` is set from the `hints.brave` presence flag, but `$browser_version` comes from `brandVersionFromHints(hints, 'Brave')`, which only reads `userAgentData.brands`. So when brands lacks a `"Brave"` entry (older builds, or `userAgentData` unavailable e.g. non-secure context), you get `$browser: Brave` with `$browser_version: null`. There's a test for the null case, so the behaviour is intentional — just flagging the asymmetry. Also note Brave's client-hint version tends to reflect the underlying **Chromium** major, not Brave's own version, so the number (when present) may surprise downstream consumers; a doc note would help.

pauldambra (commented):
QA Swarm review complete. See inline comments and the summary comment below.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -113,6 +205,17 @@ export const detectBrowser = function (user_agent: string, vendor: string | unde
     return SAMSUNG_INTERNET
   } else if (includes(user_agent, EDGE) || includes(user_agent, 'Edg/')) {
     return MICROSOFT_EDGE
+  }
+  // Chromium forks that DO stamp themselves into the UA. These must be
+  // checked before Chrome because their UA also contains `Chrome/`.
+  else if (includes(user_agent, VIVALDI + '/')) {
+    return VIVALDI
+  } else if (includes(user_agent, 'YaBrowser/')) {
```
pauldambra, line 164: > [!NOTE]

The new UA markers are half `CONST + '/'` (`VIVALDI`, `WHALE`, `WATERFOX`) and half raw literals (`'YaBrowser/'` here, `'PaleMoon/'`, `'Ddg/'`). The file's own header comment explains the concat-from-constants pattern is a deliberate minification optimization — so the half-and-half is the worst of both worlds: a maintainer renaming `YANDEX` would see detection silently *not* follow, with no type error. Pick one approach and apply it consistently (or, if some are deliberate exceptions, say so).

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -50,10 +50,85 @@ const GENERIC_MOBILE = GENERIC + ' ' + MOBILE.toLowerCase()
...
+  }
+  if (hints.brave) {
+    return BRAVE
+  }
+  const brands = hints.userAgentDataBrands
+  if (brands && brands.length) {
+    for (let i = 0; i < brands.length; i++) {
+      const brand = brands[i]?.brand
+      if (brand) {
+        const mapped = CLIENT_HINTS_BRAND_MAP[brand.toLowerCase()]
```
pauldambra: > [!NOTE]

**[security-audit/type-confusion]** 🟢 LOW

`CLIENT_HINTS_BRAND_MAP` is a plain object literal, so `CLIENT_HINTS_BRAND_MAP[brand.toLowerCase()]` for an attacker-supplied brand like `"__proto__"` or `"constructor"` resolves to an inherited `Object.prototype`/`Object` member (both truthy), passing the `if (mapped)` guard. `brandFromHints` then returns a non-string, which `detectBrowser` returns as `$browser` (violating its `string` contract).

Impact is low — it's a read (no prototype write), reaches only a JSON-serialized analytics property on the attacker's own events, and enables no injection or DoS. ReDoS was checked and ruled out (the new `Name/(\d+(\.\d+)?)` regexes have no catastrophic backtracking). Cheap robustness fix: require `typeof mapped === 'string'` before returning, or build the map with `Object.create(null)`.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -113,6 +205,17 @@ export const detectBrowser = function (user_agent: string, vendor: string | unde
...
+  }
+  // Chromium forks that DO stamp themselves into the UA. These must be
+  // checked before Chrome because their UA also contains `Chrome/`.
+  else if (includes(user_agent, VIVALDI + '/')) {
+    return VIVALDI
+  } else if (includes(user_agent, 'YaBrowser/')) {
+    return YANDEX
+  } else if (includes(user_agent, WHALE + '/')) {
+    return WHALE
+  } else if (includes(user_agent, DUCKDUCKGO + '/') || includes(user_agent, 'Ddg/')) {
```
pauldambra, line 168: > [!NOTE]

**[qa-team/data-integrity]** 🟢 LOW

The `includes(ua, 'DuckDuckGo/')` half of this OR may be dead code — real DDG app UAs use the `Ddg/` token. If `DuckDuckGo/` is backed by a captured real UA, great; otherwise it's worth dropping to keep this size-sensitive file lean. Separately (a gap, not a regression): DDG on iOS/Android often ships **no** DDG marker at all, so a chunk of mobile DDG traffic will still bucket as Safari/Chrome.

pauldambra:
> [!NOTE]
>

The headline feature works: I verified the new fork branches sit **before** the Chrome check (Vivaldi/Yandex/Whale/DuckDuckGo) and before the Firefox/Gecko checks (Pale Moon/Waterfox), the version-regex `matches.length - 2` math holds for the new alternation patterns, the new `@posthog/core` exports are wired through the package index, and the optional third `BrowserDetectionHints` arg is backwards compatible. No security blockers (pure read-only UA/navigator inspection). The open items are correctness/consistency discussion points, not merge blockers.

### Key findings (see inline comments)
- 🟠 **HIGH** — Arc detection depends on an `"Arc"` client-hints brand that Arc may not actually emit in production; please confirm against a captured real Arc client-hints dump. (`user-agent-utils.ts:90`)
- 🟡 **MEDIUM** *(convergent: paul + qa-team)* — `if (nav.brave)` uses object-presence as the Brave signal rather than the async `isBrave()`; document the trade-off. (`event-utils.ts:284`)
- 🟡 **MEDIUM** — Brave can yield `$browser: Brave` with a `null`/Chromium-major version. (`user-agent-utils.ts:302`)
- 🟡 **MEDIUM** *(convergent: paul + xp)* — `CONST + '/'` vs raw-literal (`'YaBrowser/'`, `'PaleMoon/'`, `'Ddg/'`) inconsistency; a rename could silently break detection. (`user-agent-utils.ts:213`)
- 🟢 **LOW** — `CLIENT_HINTS_BRAND_MAP[brand.toLowerCase()]` lets `__proto__`/`constructor` slip the truthiness guard (low-impact type confusion). (`user-agent-utils.ts:107`)
- 🟢 **LOW** — `DuckDuckGo/` literal may be dead code; mobile DDG often has no marker. (`user-agent-utils.ts:217`)

### Nits (non-blocking, not posted inline)
- Waterfox "G" releases give `$browser: Waterfox` + `null` version (tested + intended — flagging the silent partial result).
- `getEventProperties` scans the UA twice per event (`detectBrowser` + `detectBrowserVersion` re-deriving it) — pre-existing, not a regression; could halve the scan while you're here.
- New `}\n  else if` brace style differs from the surrounding `} else if` chain.

### Dropped on verification (so you don't chase them)
Two findings raised by one reviewer were checked against source and found **invalid**: (1) a claim that the fork branches were dead-coded after the Chrome check — they're correctly ordered before it; (2) a claim that `isNaN`/`null` violate a repo lint rule — `isNaN` is used idiomatically in 7+ places and no such rule exists. An "untyped helper params" note was also dropped (the helpers are fully typed).

### Convergence
The `navigator.brave` heuristic concern and the const-vs-literal inconsistency were each raised independently by two reviewers — higher confidence.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Ordering/regex correctness all verified; main risk is real-world Arc detection and Brave version asymmetry. |
| 👤 paul | Approving in spirit; wants the brave-signal reasoning written down and the const/literal inconsistency tidied. |
| 📐 xp | Client-hints approach is the right design; keep the two hint loops separate (not premature-dedup); make the constant convention consistent. (Two HIGH claims did not hold up on inspection.) |
| 🛡 security-audit | One low-impact type-confusion read via the brand map; ReDoS/DoS ruled out. No real vulnerabilities. |

---

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

Addressed in `ba3d2dd1` — the client-hints-brand approach was replaced with a real Arc signal (the `--arc-palette-title` CSS custom property Arc injects on `document.documentElement`). Resolving.

packages/browser/src/utils/event-utils.ts
pauldambra: > [!NOTE]

Addressed in `ba3d2dd1` — still presence-based (`navigator.brave`), but this is now deliberately documented as Brave's recommended sync signal, with an explicit iOS UA fallback (`Brave/`) since WebKit doesn't ship `navigator.brave`. Resolving the heuristic concern as a documented, intentional trade-off.

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

Addressed in `ba3d2dd1` — desktop/Android Brave now returns a null version (documented), and iOS Brave gets a real version from the new `Brave/` UA marker via `versionRegexes`. The Chromium-major confusion is gone (brands are no longer read). Resolving.

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

Addressed in `ba3d2dd1` — `CLIENT_HINTS_BRAND_MAP` and the attacker-controllable object-literal lookup were removed entirely; hints are now plain boolean flags (`arc`/`brave`). The `__proto__`/`constructor` type-confusion path no longer exists. Resolving.

packages/browser/src/utils/event-utils.ts
```diff
@@ -261,6 +268,40 @@ export function getTimezoneOffset(): number | undefined {
...
+// Arc deliberately presents as plain Chromium in the UA *and* in the
+// low-entropy `navigator.userAgentData.brands` list, so neither helps here.
+const ARC_CSS_PROPERTY = '--arc-palette-title'
+
+function detectArc(): boolean {
+    if (typeof document === 'undefined' || !document?.documentElement) {
+        return false
+    }
+    try {
+        const value = window?.getComputedStyle?.(document.documentElement).getPropertyValue(ARC_CSS_PROPERTY)
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + paul + xp]** 🟠 HIGH

`detectArc()` calls `window.getComputedStyle(document.documentElement).getPropertyValue('--arc-palette-title')`, and `getBrowserDetectionHints()` is invoked inside `getEventProperties` — which runs on **every captured event** (pageview, autocapture click, custom event). `getComputedStyle` can force a synchronous style/layout recalc on the main thread, and the result here is **invariant for the page lifetime** (neither Arc's injected CSS var nor `navigator.brave` change after load). On high-volume autocapture/scroll-heavy pages this is repeated wasted work with no SDK-side observability — the classic "PostHog made my site janky" profile.

Three reviewers raised this independently. Recommend computing the hints **once** (module-level lazy memo, or at init) and reusing them, rather than recomputing per event. "Make it fast" once it works — the value is genuinely cacheable.

pauldambra (commented):
QA Swarm second-pass review (against the fix commit ba3d2dd1). See inline comments and summary.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -88,10 +133,27 @@ const safariCheck = (ua: string, vendor?: string) => (vendor && includes(vendor,
  * This function detects which browser is running this script.
  * The order of the checks are important since many user agents
  * include keywords used in later checks.
+ *
+ * `hints` is an optional bag of extra signals (Client Hints `brands`,
```
pauldambra: > [!NOTE]

**[convergent: qa-team + paul + xp]** 🟡 MEDIUM

Stale JSDoc. This `detectBrowser` doc comment still says `hints` is "an optional bag of extra signals (Client Hints `brands`, `navigator.brave`)", but `BrowserDetectionHints` no longer carries `brands` — the interface is now `{ arc?, brave? }`, and the (accurate) interface-level JSDoc right above it explicitly states Arc does **not** use brands. A comment that contradicts the code it documents sends the next maintainer hunting for brand-parsing that no longer exists. Retitle it to reference the real signals (`--arc-palette-title` CSS property + `navigator.brave`).

packages/browser/src/utils/event-utils.ts
```diff
@@ -261,6 +268,40 @@ export function getTimezoneOffset(): number | undefined {
     }
 }
 
+// Arc injects this CSS custom property on `document.documentElement` very
+// early on every page it loads. It's the most reliable sync signal we have —
+// Arc deliberately presents as plain Chromium in the UA *and* in the
+// low-entropy `navigator.userAgentData.brands` list, so neither helps here.
+const ARC_CSS_PROPERTY = '--arc-palette-title'
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/data-integrity]** 🟡 MEDIUM

False-positive surface. `--arc-palette-title` is an ordinary CSS custom property — any site can set `--arc-palette-title: <non-empty>` on `:root`/`html` (deliberately, or by copying an Arc-theming snippet), and every visitor on any browser would then be reported as `$browser: 'Arc'`. Same fragile-external-signal shape as the `navigator.brave` presence check, just relocated — and because hints win over UA detection, it silently overrides the real browser.

Reasonable trade-off (Arc is otherwise invisible), but worth: (a) a one-line note that `$browser === 'Arc'` now means "Arc **or** a page that set this var", (b) confirming nothing downstream buckets/bills on `$browser` accuracy, and (c) a parameterized test pinning the boundary (page sets the var on a Chrome UA → assert intended behaviour).

packages/browser/src/utils/event-utils.ts
```diff
@@ -261,6 +268,40 @@ export function getTimezoneOffset(): number | undefined {
...
+    } catch {
+        return false
+    }
+}
+
+// Gathers signals that aren't in the UA string but help identify browsers that
+// hide themselves there (Arc desktop, Brave on desktop/Android). Brave on iOS
+// is picked up via the `Brave/` UA marker by `detectBrowser` itself, so no
+// hint is needed for that case.
+export function getBrowserDetectionHints(): BrowserDetectionHints {
```
pauldambra, line 275: > [!NOTE]

**[xp]** 🟢 LOW

Once-and-only-once nit: this file already imports SSR-safe `document`/`window` from `./globals`, and `globals` also exports a safe `navigator`. `detectArc` adds its own `typeof document === 'undefined'` guard (redundant with the `!document?.documentElement` that follows, since `document` here is the already-guarded import), and `getBrowserDetectionHints` re-derives `nav` via `typeof navigator !== 'undefined'`. Prefer the existing `globals` exports for consistency with the rest of the file.

pauldambra:
> [!NOTE]
>
> Second-pass multi-perspective review against fix commit `ba3d2dd1`

The `ba3d2dd1` fix is a real improvement: the fake client-hints-brand approach (and its prototype-pollution surface) is gone, Arc now uses a genuine signal (`--arc-palette-title`), and iOS Brave is handled via its `Brave/` UA marker. Detection ordering re-verified correct (forks before Chrome; iOS Brave before Safari; Pale Moon/Waterfox before Firefox). Security: **0 findings** (ReDoS ruled out — the new regexes are linear; no dangerous sink). Remaining items are quality/perf, not blockers.

### Key findings (see inline)
- 🟠 **HIGH** *(convergent: qa-team/performance + paul + xp)* — `detectArc()`'s `getComputedStyle` runs on the **per-event hot path** but is page-lifetime-invariant; memoize the hints once. (`event-utils.ts:282`)
- 🟡 **MEDIUM** *(convergent: qa-team + paul + xp)* — `detectBrowser` JSDoc still references the removed "Client Hints `brands`". (`user-agent-utils.ts:137`)
- 🟡 **MEDIUM** *(convergent: paul + qa-team/data-integrity)* — `--arc-palette-title` is page-settable, so `$browser: Arc` can false-positive; document + add a test. (`event-utils.ts:275`)
- 🟢 **LOW** — reuse the file's existing SSR-safe `globals` (`document`/`window`/`navigator`) instead of re-rolling `typeof` guards. (`event-utils.ts:293`)

### Resolved since first pass
The first-pass HIGH (Arc detection via a signal Arc doesn't emit) and the LOW security finding (object-literal map type-confusion) are both fixed by this commit. Four first-pass threads were resolved with notes.

### Now considered intentional (dropped)
The const-vs-literal mix (`VIVALDI + '/'` vs `'YaBrowser/'`) — both paul and xp concluded it matches the file's existing pre-uglify convention (consts where the marker == result name and can be shared; literals where they differ, e.g. `YaBrowser`→`Yandex`). Leave as-is.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Fix resolves prior HIGH; one perf item (per-event getComputedStyle) worth fixing; stale JSDoc + spoofable-signal notes. |
| 👤 paul | Approve-with-caveats; wants the hot-path getComputedStyle hoisted and the page-settable Arc signal acknowledged before merge. |
| 📐 xp | Clean simplification (`browserFromHints`); must-fix the lying JSDoc; memoize the per-event style read; const/literal mix is intentional, leave it. |
| 🛡 security-audit | 0 findings — regexes linear, output only JSON-serialized, prior type-confusion vector removed. |

---

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

On the second review pass, both reviewers concluded this matches the file's existing pre-uglify convention (consts where the UA marker equals the result name and can be shared; raw literals where they differ, e.g. `YaBrowser`->`Yandex`, `PaleMoon`->`Pale Moon`). Treating as intentional and resolving.

packages/browser/src/utils/event-utils.ts
pauldambra: > [!NOTE]

Fixed in `72093b29` — `getBrowserDetectionHints` is now memoized in a module-level cache, so `detectArc`'s `getComputedStyle` runs once per page instead of on every captured event. Both signals are page-lifetime-invariant. (Tests reset the cache via a new `resetBrowserDetectionHintsCache` export.)

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

Fixed in `72093b29` — the `detectBrowser` JSDoc no longer references the removed Client Hints `brands`; it now describes the actual out-of-band signals (Arc CSS custom property + `navigator.brave`).

packages/browser/src/utils/event-utils.ts
pauldambra: > [!NOTE]

Fixed in `72093b29` — added a comment documenting that the CSS custom property is page-controllable, so `$browser === 'Arc'` means "Arc, or a page that set this variable" (best-effort, not authoritative). The spoof boundary is already pinned by the existing e2e test (`--arc-palette-title` present on a Chrome UA -> Arc).

packages/browser/src/utils/event-utils.ts
pauldambra: > [!NOTE]

Fixed in `72093b29` — `navigator` is now imported from `./globals`, `detectArc` uses the imported SSR-safe `document`/`window` (dropped the redundant `typeof document` guard), and `getBrowserDetectionHints` no longer re-rolls a `typeof navigator` check.

packages/core/src/utils/user-agent-utils.ts
pauldambra: > [!NOTE]

Not dead code — the `DuckDuckGo Browser on Android` test case exercises this branch (UA: `...Mobile DuckDuckGo/5 Safari/537.36`). The `DuckDuckGo/` token is the real Android DDG marker; `Ddg/` is the iOS one. Both are needed. Resolving.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -50,10 +50,85 @@ const GENERIC_MOBILE = GENERIC + ' ' + MOBILE.toLowerCase()
...
+  userAgentDataBrands?: UserAgentBrand[]
+  // Pass `!!navigator.brave` here. The Brave object exists on all Brave
+  // builds and is the recommended sync detection signal.
+  brave?: boolean
+}
+
+// Brand names we recognise in `navigator.userAgentData.brands`. Comparison is
+// case-insensitive against the brand field.
+const CLIENT_HINTS_BRAND_MAP: Record<string, string> = {
+  arc: ARC,
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Already addressed in `ba3d2dd1` — Arc detection was switched from UA-CH `brands` to the `--arc-palette-title` CSS custom property Arc injects on `document.documentElement`, which is a signal Arc actually emits. Resolving this stale comment.

packages/core/src/utils/user-agent-utils.ts
```diff
@@ -88,10 +163,27 @@ const safariCheck = (ua: string, vendor?: string) => (vendor && includes(vendor,
...
+  hints?: BrowserDetectionHints
+): string {
   vendor = vendor || '' // vendor is undefined for at least IE9
 
+  // Client Hints / API-based signals win over UA sniffing because the
+  // browsers we care about here (Arc, Brave) are deliberately invisible in
+  // the UA string and would otherwise be misdetected as Chrome.
+  const fromHints = brandFromHints(hints)
+  if (fromHints) {
+    return fromHints
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Already addressed in `ba3d2dd1` — iOS Brave is now detected via the `Brave/` UA marker (checked before the Safari branch, since iOS Brave's UA otherwise looks like Mobile Safari). Resolving this stale comment.

pauldambra:
> [!NOTE]
>
> Third / final-state multi-perspective review against `5c8e323f`

## Verdict: ✅ APPROVE (one fix applied)

Re-reviewed the whole feature after the fix/research/simplicity rounds. Detection ordering re-verified correct, the `matches.length - 2` index holds for the alternation regexes, Arc/desktop-Brave correctly return `null` versions via the `versionRegexes` lookup (no special-case needed), and `navigator` optional-chaining is sound. **Security: 0 findings** (ReDoS ruled out — the new regexes are linear; output only flows to JSON-serialized analytics properties; the prior type-confusion vector is gone).

### Fixed this round
- 🟡 **MEDIUM** *(convergent: qa-team + paul)* — the changeset still named the old `--arc-palette-title` variable while the code uses `--arc-palette-background`. Since the changeset ships to the public npm changelog, this would misdirect anyone debugging an unexpected `$browser: Arc`. **Fixed in `10217845`.**

### Non-blocking notes (no change made)
- 🟢 **LOW** — memo gate (`hints.arc || readyState === 'complete'`): a page that never reaches `complete` (or stalls at `interactive`) means a confirmed non-Arc user keeps paying the per-event `getComputedStyle`. Bounded — worst case is the pre-memo cost — and such pages are rare. Defensible tradeoff; could short-circuit `detectArc` when `brave` is already set if it ever matters.
- 🟢 **Observability** *(paul, standing item)* — the detection has documented silent failure modes (early-load Arc users bucketed as Chrome until next page; any page setting the CSS var looks like Arc) with no in-product counter for re-bucketing hit/false-positive rate. Reasonable to watch the live `$browser` distribution shift post-release instead — flagging so that's a **conscious** decision, not an omission.
- ⚪ **NIT** — `!!value && value.trim().length > 0` in `detectArc` could be `!!value?.trim()`; the new detection cases could collapse into a single parameterized `it.each`; one Arc fact is stated in both files' comments (acceptable — they're across a package boundary). All cosmetic.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Correct & well-tested; only the stale changeset was worth acting on (now fixed). |
| 👤 paul | Approve; fix the changeset (done), and make "watch `$browser` distribution post-release" a stated decision. |
| 📐 xp | Converged well; both prior simplifications confirmed; remaining items are NITs. Ship it. |
| 🛡 security-audit | 0 findings. |

---

packages/browser/src/utils/event-utils.ts
```diff
@@ -261,6 +266,72 @@ export function getTimezoneOffset(): number | undefined {
...
+    const hints: BrowserDetectionHints = {}
+    if (nav?.brave) {
+        hints.brave = true
+    }
+    if (detectArc()) {
+        hints.arc = true
+    }
+    if (hints.arc || document?.readyState === 'complete') {
+        cachedBrowserDetectionHints = hints
+    }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Good catch — fixed in `d6b801d5`. Added `hints.brave` to the memo gate (`if (hints.arc || hints.brave || document?.readyState === 'complete')`), since `navigator.brave` is stable from page start and can be frozen immediately. This stops Brave users from re-running `detectArc`'s `getComputedStyle` on every event during load. Added a regression test asserting `getComputedStyle` is not re-invoked on the second call once Brave is detected pre-load. Resolving.

pauldambra:
> [!NOTE]
>
> Final-state multi-perspective review against `844269e9` (post Arc-removal)

## Verdict: ✅ APPROVE — nothing to fix

Re-reviewed the simplified, Arc-free feature. The Arc teardown is **clean**: no leftover `getComputedStyle` / `readyState` / memoization / `--arc-palette` / test-reset machinery, no orphaned imports/consts/types/comments (verified by repo-wide grep across all four reviewers). Detection ordering re-confirmed correct (forks before Chrome; iOS `Brave/` before Safari; Pale Moon/Waterfox before Firefox); `detectBrowserVersion` correctly returns `null` for desktop Brave; the `matches.length - 2` index holds for the alternation regexes. **Security: 0 findings** (ReDoS ruled out — all version regexes are linear; output only flows to JSON-serialized analytics properties).

### Non-blocking notes (no change made)
- ⚪ **NIT** — the changeset *filename* is still `detect-arc-brave-and-friends.md`; its **content** is correct (no Arc). Changeset filenames are ephemeral (consumed at release, never in the changelog), so not worth churning CI to rename.
- 🟢 **LOW** — `detectBrowserVersion` returning `null` for desktop Brave depends implicitly on the `[BRAVE]` `versionRegexes` entry only matching the iOS `Brave/` marker. Works today and is pinned by a test; flagged only so a future maintainer doesn't add a Chrome-style regex there and silently start reporting Chrome versions for desktop Brave.
- ⚪ **NIT** — `getBrowserDetectionHints` reads `navigator` via a `typeof` guard while siblings use the bare global; this is **deliberate** (the late-bound read is needed so test mocks that replace `navigator` are seen — the globals snapshot would miss them).

### Dropped on verification
A reported MEDIUM "stray `@posthog/mcp` hunk in `release.yml`" was checked against the actual PR diff (`origin/main...HEAD`) and is **not present** — the PR contains exactly 5 files (2 source, 2 test, 1 changeset). That was a local-environment diff artifact, not part of this PR.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Arc removal clean, nothing dangling; only note was the (false) release.yml hunk + cosmetic filename. |
| 👤 paul | Approving; right call dropping Arc; `getBrowserDetectionHints` is the right size now. Keep watching `$browser` distribution post-release. |
| 📐 xp | Complete & clean teardown; trio is at the right altitude; only a NIT on navigator-access style (deliberate). Ship it. |
| 🛡 security-audit | 0 findings. |

---

marandaneto (approved):

PR #3713, opened 2026-06-01
fix: honor disableGeoip constructor option in event capture

### LINKED ISSUE #3818: [posthog-node] `disableGeoip` on the constructor PostHogOptions is typed but ignored at runtime
### SDK
`posthog-node@5.35.5`

### Bug

`PostHogOptions.disableGeoip` is part of the constructor option type — TypeScript accepts it without an error — but the runtime client never reads it during event capture. The flag is consulted **per event** (`EventMessage.disableGeoip`, `IdentifyMessage.disableGeoip`) and the constructor-level setting has no effect, so `$geoip_*` properties are still derived from the request IP for projects that expected a global opt-out.

This is a silent privacy regression for any consumer that intends \"do not let PostHog enrich events from the request IP\" as a global posture. We hit it in a privacy-sensitive product where we mean to send PostHog **no** geo data ever; the typecheck passed, no runtime log was emitted, and we only caught the leak when we inspected the captured properties on the PostHog side.

### Minimal repro

```ts

// Constructor option is typed and accepted, but ignored at runtime.
const client = new PostHog(API_KEY, {
  disableGeoip: true, // ⬅ no effect
})

// Geo enrichment happens here regardless of the constructor option above.
await client.captureImmediate({
  distinctId: 'user-1',
  event: 'test',
  // disableGeoip: true,  // ← this is the only honored placement
})
```

The captured event arrives with `$geoip_*` properties derived from the ingest server's view of the client IP, despite the constructor opt-out.

### Expected

Either:
1. **Honor it at the constructor level** — wrap every \`captureImmediate\` / \`identifyImmediate\` / \`capture\` call so the constructor flag propagates to the per-event payload when the call site doesn't override it. This is what consumers reading the type definitions expect.
2. **Remove it from \`PostHogOptions\`** so the type system surfaces the misuse at compile time (TS2353).

(1) is the less-surprising fix and lines up with the option being privacy-relevant.

### Current workaround

Set \`disableGeoip: true\` on every per-event payload at the wrapper layer. See [our wrapper](https://github.com/maniator/co-parenting-coordination/blob/main/packages/server/src/analytics/posthog.service.ts) for the pattern we landed on.

### Why this matters

Silent-runtime-ignore of a documented privacy option is the worst class of typed API bug for a sensitive consumer: the type system says you opted out, the SDK shipped the data anyway. Either path above (honor or remove the type) fixes it; the status quo is the only outcome that's hard to defend.

gavin913-lss:
## Summary

Fixes PostHog/posthog-js#3818

The `PostHogOptions.disableGeoip` constructor option was typed and accepted by TypeScript but never read at runtime. The flag was only consulted per-event (`EventMessage.disableGeoip`), silently ignoring the global opt-out set in the constructor.

This is a **privacy regression** for consumers who set `disableGeoip: true` globally expecting all events to opt out of geo-IP enrichment. The type system accepted the option, no runtime error was emitted, and geo data continued to leak.

## Changes

Modified `prepareEventMessage()` in `packages/node/src/client.ts` to propagate the constructor-level `disableGeoip` option to the event payload using nullish coalescing (`??`):

```typescript
// Before
disableGeoip: eventMessage.disableGeoip,
// After  
disableGeoip: eventMessage.disableGeoip ?? this.options.disableGeoip,
```

This ensures:
1. Per-event `disableGeoip` still takes precedence when explicitly set
2. Constructor-level `disableGeoip` is used as fallback when the event doesn't specify it
3. No breaking changes - existing per-event usage continues to work


gavin913-lss:
Thanks for the review! I've addressed both issues:

1. **identify/alias methods**: Applied the same  fallback to Version: ImageMagick 7.1.2-21 Q16-HDRI x86_64 51e1d5d97:20260421 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/license/
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib freetype heic jng jpeg lcms ltdl lzma png tiff webp xml zlib zstd
Compiler: clang (16.0.0)
Usage: identify [options ...] file [ [options ...] file ... ]

Image Settings:
  -alpha option        on, activate, off, deactivate, set, opaque, copy
                       transparent, extract, background, or shape
  -antialias           remove pixel-aliasing
  -authenticate password
                       decipher image with this password
  -clip                clip along the first path from the 8BIM profile
  -clip-mask filename  associate a clip mask with the image
  -clip-path id        clip along a named path from the 8BIM profile
  -colorspace type     alternate image colorspace
  -crop geometry       cut out a rectangular region of the image
  -define format:option
                       define one or more image format options
  -density geometry    horizontal and vertical density of the image
  -depth value         image depth
  -endian type         endianness (MSB or LSB) of the image
  -extract geometry    extract area from image
  -features distance   analyze image features (e.g. contrast, correlation)
  -format "string"     output formatted image characteristics
  -fuzz distance       colors within this distance are considered equal
  -gamma value         of gamma correction
  -interlace type      type of image interlacing scheme
  -interpolate method  pixel color interpolation method
  -limit type value    pixel cache resource limit
  -matte               store matte channel if the image has one
  -moments             report image moments
  -monitor             monitor progress
  -ping                efficiently determine image attributes
  -precision value     maximum number of significant digits to print
  -quiet               suppress all warning messages
  -regard-warnings     pay attention to warning messages
  -respect-parentheses settings remain in effect until parenthesis boundary
  -sampling-factor geometry
                       horizontal and vertical sampling factor
  -seed value          seed a new sequence of pseudo-random numbers
  -set attribute value set an image attribute
  -size geometry       width and height of image
  -strip               strip image of all profiles and comments
  -unique              display the number of unique colors in the image
  -units type          the units of image resolution
  -verbose             print detailed information about the image
  -virtual-pixel method
                       virtual pixel access method

Image Operators:
  -auto-orient         automagically orient (rotate) image
  -channel mask        set the image channel mask
  -grayscale method    convert image to grayscale
  -negate              replace every pixel with its complementary color

Miscellaneous Options:
  -debug events        display copious debugging information
  -help                print program options
  -list type           print a list of supported option arguments
  -log format          format of debugging information
  -version             print version information

By default, the image format of 'file' is determined by its magic
number.  To specify a particular image format, precede the filename
with an image format name and a colon (i.e. ps:image) or specify the
image type as the filename suffix (i.e. image.ps).  Specify 'file' as
'-' for standard input or output., , , and  methods, matching the pattern in .

2. **Regression tests**: Added two tests:
   -  with constructor  and no per-event override → asserts 
   - Version: ImageMagick 7.1.2-21 Q16-HDRI x86_64 51e1d5d97:20260421 https://imagemagick.org
Copyright: (C) 1999 ImageMagick Studio LLC
License: https://imagemagick.org/license/
Features: Cipher DPC HDRI Modules 
Delegates (built-in): bzlib freetype heic jng jpeg lcms ltdl lzma png tiff webp xml zlib zstd
Compiler: clang (16.0.0)
Usage: identify [options ...] file [ [options ...] file ... ]

Image Settings:
  -alpha option        on, activate, off, deactivate, set, opaque, copy
                       transparent, extract, background, or shape
  -antialias           remove pixel-aliasing
  -authenticate password
                       decipher image with this password
  -clip                clip along the first path from the 8BIM profile
  -clip-mask filename  associate a clip mask with the image
  -clip-path id        clip along a named path from the 8BIM profile
  -colorspace type     alternate image colorspace
  -crop geometry       cut out a rectangular region of the image
  -define format:option
                       define one or more image format options
  -density geometry    horizontal and vertical density of the image
  -depth value         image depth
  -endian type         endianness (MSB or LSB) of the image
  -extract geometry    extract area from image
  -features distance   analyze image features (e.g. contrast, correlation)
  -format "string"     output formatted image characteristics
  -fuzz distance       colors within this distance are considered equal
  -gamma value         of gamma correction
  -interlace type      type of image interlacing scheme
  -interpolate method  pixel color interpolation method
  -limit type value    pixel cache resource limit
  -matte               store matte channel if the image has one
  -moments             report image moments
  -monitor             monitor progress
  -ping                efficiently determine image attributes
  -precision value     maximum number of significant digits to print
  -quiet               suppress all warning messages
  -regard-warnings     pay attention to warning messages
  -respect-parentheses settings remain in effect until parenthesis boundary
  -sampling-factor geometry
                       horizontal and vertical sampling factor
  -seed value          seed a new sequence of pseudo-random numbers
  -set attribute value set an image attribute
  -size geometry       width and height of image
  -strip               strip image of all profiles and comments
  -unique              display the number of unique colors in the image
  -units type          the units of image resolution
  -verbose             print detailed information about the image
  -virtual-pixel method
                       virtual pixel access method

Image Operators:
  -auto-orient         automagically orient (rotate) image
  -channel mask        set the image channel mask
  -grayscale method    convert image to grayscale
  -negate              replace every pixel with its complementary color

Miscellaneous Options:
  -debug events        display copious debugging information
  -help                print program options
  -list type           print a list of supported option arguments
  -log format          format of debugging information
  -version             print version information

By default, the image format of 'file' is determined by its magic
number.  To specify a particular image format, precede the filename
with an image format name and a colon (i.e. ps:image) or specify the
image type as the filename suffix (i.e. image.ps).  Specify 'file' as
'-' for standard input or output. with constructor  → asserts

gavin913-lss:
Thanks for the review! I've addressed both issues:

1. **identify/alias methods**: Applied the same `?? this.options.disableGeoip` fallback to `identify`, `identifyImmediate`, `alias`, and `aliasImmediate` methods, matching the pattern already used in `prepareEventMessage` for `capture`.

2. **Regression tests**: Added two tests:
   - `capture` with constructor `disableGeoip: true` and no per-event override asserts `$geoip_disable: true` appears in event properties
   - `identify` with constructor `disableGeoip: true` asserts `$geoip_disable: true` appears in event properties

marandaneto:
https://github.com/PostHog/posthog-js/issues/3818#issuecomment-4707641052

PR #3719, opened 2026-06-02
feat(browser): add __preview_cookie_wins_on_conflict for multi-subdomain identify

lricoy:
## Summary

Adds an opt-in `__preview_cookie_wins_on_conflict` config that fixes a latent multi-subdomain identify / session-disconnect bug in `'localStorage+cookie'` persistence.

**Bug:** in the default `'localStorage+cookie'` mode, identity keys (`distinct_id`, `$device_id`, `$session_id`, `$user_state`, `$initial_person_info`, etc.) are stored both in the cross-subdomain cookie and in the per-subdomain localStorage. On `_parse` the two are merged with `extend(cookieProperties, localStorageData)` (`storage.ts`) and `extend()` lets later args win, so **localStorage wins on conflict**. For sites running posthog on more than one subdomain, a stale per-subdomain localStorage can clobber a fresh cookie written by an `identify()` on another subdomain — surfacing as identified users looking anonymous after a subdomain transition, plus spurious new `$session_id` mints when the merged stale activity timestamp trips the idle-timeout check.

**Fix:** opt-in flag that flips the merge order. When `true`, cookie values win for the keys it holds; null/empty cookie values are filtered before the merge so a malformed legacy cookie cannot wipe out valid localStorage data. The existing self-heal write at the end of `_parse` means stale localStorage gets overwritten with the merged value on the first load with the flag enabled. localStorage-only keys (flag caches, surveys, super properties, registered URL params) are unaffected.

**Why not just flip the default:** localStorage-wins has shipped for years; some users may rely on it. Opt-in unblocks the multi-subdomain use case immediately with zero blast radius. The changeset telegraphs this may become the default in a future major.

**Why not `persistence: 'cookie'`:** that mode dumps the entire props object (including `$surveys`, `$feature_flag_details`, payloads, super properties) into a single 4KB-capped cookie, which is too small for many real configs.

## Scope

- `packages/types/src/posthog-config.ts` — new `__preview_cookie_wins_on_conflict: boolean` with full JSDoc, default `false`, init-time only, no effect for other persistence modes
- `packages/browser/src/posthog-core.ts` — default `false`
- `packages/browser/src/storage.ts` — second param `preferCookieOnConflict` on `createLocalPlusCookieStore`; conditional merge order; defensive filter for null/empty cookie values
- `packages/browser/src/posthog-persistence.ts` — plumbs the config through
- `packages/types/src/__tests__/__snapshots__/config-snapshot.spec.ts.snap` — auto-regenerated

## Tests

- **+10** in `posthog-persistence.test.ts` under a new `merge precedence` describe block:
  - flag-off lock-in (locks in current localStorage-wins default — regression coverage that didn't exist)
  - flag-on conflict (cookie wins for keys in both stores)
  - cross-subdomain identify scenario (full stale-anonymous localStorage + fresh-identified cookie → all five keys come from cookie)
  - self-heal write-back assertion
  - localStorage-only keys preserved
  - empty cookie, empty localStorage round-trips
  - defensive filter: null and empty-string cookie values do NOT clobber valid localStorage values
  - no effect on non-conflicting keys
- **+7** in `utils.test.ts` pinning `extend()` semantics so a future refactor cannot silently change them (later wins for defined values, skips `undefined`, applies `null`/`''`/`0`/`false`, mutates first arg)

## Test plan

- [x] `pnpm test:unit` (browser) — 4073/4073 pass, 102/102 suites
- [x] `pnpm test:unit` (types) — 2/2 pass, snapshot clean
- [x] `pnpm lint` — clean (both packages)
- [x] `pnpm typecheck` — clean (production; pre-existing `test:typecheck` failure on missing dompurify/dotenv type defs is unrelated to this change)
- [ ] Playwright cross-host scenario — follow-up; the existing fixtures don't simulate eTLD+1 cookie scope and the unit coverage above exercises the same code path directly.

## Notes for reviewers

- Flag is read once at SDK init when `_buildStorage` runs; toggling via `set_config` after init has no effect. Documented in JSDoc.
- No effect for `'cookie'`, `'localStorage'`, `'memory'`, `'sessionStorage'` persistence modes.
- `createLocalPlusCookieStore` is not part of the public export surface — the signature change is internal.


dustinbyrne (approved):

dustinbyrne (approved):

packages/browser/src/posthog-core.ts
```diff
@@ -219,6 +217,7 @@ export const defaultConfig = (defaults?: ConfigDefaults): PostHogConfig => ({
     defaults: defaults ?? 'unset',
     __preview_deferred_init_extensions: false, // Opt-in only for now
     __preview_external_dependency_versioned_paths: false,
+    __preview_cookie_wins_on_conflict: false, // Opt-in: fixes cross-subdomain stale-localStorage bug
```
dustinbyrne, line 220: fwiw, I think it's likely safe to enable this by default if the persistence mode is `localStorage+cookie` and `cross_subdomain_cookie` is true

It's _possible_ some users rely on the broken behavior, but I can't think of a legitimate use case that's not improved by this

dustinbyrne (approved):
third time's the charm

turnipdabeets (approved):

PR #3721, opened 2026-06-02
feat(persistence): split feature-flag config into separate storage

pauldambra:
## Problem

posthog-js persists everything in a single localStorage blob keyed `ph_<token>_posthog`. The feature-flag config cluster (`$feature_flag_details`, `$enabled_feature_flags`, `$active_feature_flags`, `$feature_flag_payloads`, `$feature_flag_request_id`, `$feature_flag_evaluated_at`) is by far the largest entry, yet it changes rarely. Because it rides on the same blob as high-frequency bookkeeping (session / activity / super-property churn), every one of those frequent writes re-serialises and re-writes the whole blob — and each `localStorage.setItem` fires a cross-tab `storage` event that broadcasts the entire payload to every same-origin tab.

There is also an open P2 Chromium bug where localStorage writes leak memory (https://issues.chromium.org/issues/394823902); writing less, and less per write, should amplify it less.

## Changes

Adds an opt-in `split_storage` config that moves the atomic flag-config cluster into a `<name>__flags` localStorage entry, leaving the frequently-written bookkeeping in the main blob. A flag response writes only its own (rare) entry, and the frequent main-blob writes no longer re-serialise or re-broadcast the large flag payload.

- **Storage-layer only.** Every flag read still resolves against the single in-memory `props`. On `load()` the group entry is merged back into `props`, so all readers are unchanged.
- **One-shot migration.** `props[key] = group[key] ?? main[key]` — the old main-blob location is read once and carried forward, then the first `save()` strips the keys from main. Upgrades never miss a cached flag.
- **Atomicity preserved.** The whole flag cluster stays in one entry written in one `_set`, so the cached values and `$feature_flag_evaluated_at` (the TTL gate) cannot tear.
- **Write-failure safe.** storage `_set` now returns whether the durable write landed; the no-op fingerprint is cached only after a confirmed write, so a swallowed quota failure is retried, not stranded.
- **Freshness-checked.** On load a stale group entry never beats a fresher main blob (compared via `$feature_flag_evaluated_at`).
- **Per-entry fingerprints + per-slot dirty tracking**, so a main-blob change neither rewrites nor re-serialises `__flags`.
- **Backends.** Only `localStorage` / `localStorage+cookie`. `memory` / `sessionStorage` / `cookie` keep the single blob. `reset()` / opt-out wipe the entry.
- **Gated** behind the `2026-05-30` dated default. Gate off is behaviourally identical to the current single-blob path, with one improvement: a failed main-blob write (e.g. quota) is now retried on the next save instead of being cached as already-saved.

Survey config is split out in the stacked follow-up #3727.

## Release info — sub-libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code
- [x] Ran `pnpm changeset`

---


pauldambra:
* **#3727** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3727?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3721** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3721?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3721?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pauldambra (commented):
> [!NOTE]

QA Swarm multi-perspective review — inline findings below; full verdict, convergence, and the low-severity / nit list are in the top-level summary comment.

packages/browser/src/posthog-persistence.ts
```diff
@@ -343,8 +448,16 @@ export class PostHogPersistence {
         // remove both domain and subdomain cookies
         this._storage._remove(this._name, false)
         this._storage._remove(this._name, true)
-        // Storage entry is gone — any future save() must write through.
-        this._lastSavedSerialized = undefined
+        // Always wipe the group entries too — even when the split is currently
+        // off — so a default flip-flop or version downgrade cannot strand an
+        // orphaned flag/survey entry that would leak across users on
+        // reset()/opt-out. They are localStorage-only, so one removal each.
+        for (const group of PERSISTENCE_STORAGE_GROUPS) {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + qa-team/compatibility]** 🔴 CRITICAL

`remove()` unconditionally deletes the shared-named `<name>__flags` / `<name>__surveys` localStorage entries (the comment right here even says "even when the split is currently off"). The trap: the SDK runs **two** `PostHogPersistence` instances that share the same `_name` — the main one (localStorage / localStorage+cookie, split-eligible) and `sessionPersistence` (forced `persistence: 'sessionStorage'`, never split-eligible). `_groupEntryName` is ``${this._name}__${group}`` and `parseName` keys `_name` off the token only, so **both instances resolve the same `__flags`/`__surveys` keys**.

`posthog.set_config(...)` reconstructs `sessionPersistence` from scratch (posthog-core.ts:3051 — pre-existing code, hence not in this diff). The fresh constructor runs `update_config` -> `set_secure(secure_cookie)` with `_secure === undefined`; since the default `secure_cookie` is a real boolean, `secure !== undefined` is true -> `remove()` runs -> it deletes the **main instance's** `__flags`/`__surveys` from localStorage. `set_config` fires from `_onRemoteConfig` (every page load after the remote-config round-trip, posthog-core.ts:940) plus start/stop recording and exception toggles. The main instance does not re-save afterward (its `update_config` short-circuits when nothing changed), and its per-slot fingerprint cache still believes the groups are written, so it won't rewrite them until the flag/survey **content** changes.

Net effect once `__split_storage` is on (which the `2026-05-30` dated default does for everyone on the default localStorage+cookie backend): the on-disk flag & survey cache is wiped shortly after almost every page load, so cached flags/surveys do **not** survive a reload and a sibling tab reads an empty cache — flag flicker / default-variant exposure on the mission-critical replay+analytics path. Not permanent data loss (in-memory props survive the current pageview; flags refetch), but it silently defeats the feature's core durability guarantee for the default config. The unit tests miss it because they exercise `PostHogPersistence` in isolation, never the dual-instance `set_config` path.

The fix needs to decouple two things that currently both flow through this one unconditional loop: (a) the **primary** instance legitimately cleaning orphaned group entries on reset/opt-out/backend-downgrade, vs (b) a **non-split sibling** never touching the primary's group keys. A naive `if (this._splitStorageEligible)` guard fixes the sibling collision but reintroduces orphan-leak for a non-eligible primary (cookie/memory). Better: give `sessionPersistence` a distinct name namespace, or pass it a flag that suppresses group reads/writes/removes entirely, so only the localStorage owner manages the group keys. Please add an integration-level regression test that drives `posthog.set_config(...)` on a `localStorage+cookie` + `__split_storage:true` instance on an https origin and asserts `__flags` survives on disk.

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +358,84 @@ export class PostHogPersistence {
...
 
-        this._storage._set(
-            this._name,
-            this.props,
-            this._expire_days,
-            this._cross_subdomain,
-            this._secure,
-            this._config.debug
-        )
+        storage._set(name, props, this._expire_days, this._cross_subdomain, this._secure, this._config.debug)
```
pauldambra: > [!NOTE]

**[convergent: qa-team/generalist-b (reproduced) + qa-team/reliability]** 🟡 MEDIUM

`_writeEntry` records `this._lastSavedSerialized[slot] = fingerprint` (line 433) **before** calling `storage._set(...)` (line 438), and `localStore._set` swallows quota/serialization errors internally. So if the `__flags` write fails (QuotaExceededError on a full localStorage, private mode, a SecurityError) while the main-blob write succeeds, the cache believes flags are persisted; every subsequent identical `save()` short-circuits at the fingerprint check, so disk never catches up — on reload `_loadGroupEntries` finds no `__flags` and the flags are gone until their content next changes (getFeatureFlag undefined, surveys / replay-linked-flag mis-evaluate). One reviewer reproduced this with a unit test that makes `_set` throw for `__flags` only.

Honest calibration: the fingerprint-before-write ordering is **pre-existing** (the single-blob code did the same), and splitting actually shrinks each entry, so it isn't strictly more likely — it just changes the failure shape so `__flags` can strand independently. The trigger (a write failure scoped to the flags entry) is uncommon, hence MEDIUM not HIGH, but the consequence is nasty and the fix is cheap: only record the fingerprint after a confirmed-successful write — have `_set` return a success boolean and gate the cache update on it (or set `_lastSavedSerialized[slot]` only if `_set` didn't throw).

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +358,84 @@ export class PostHogPersistence {
...
+    // localStorage / sessionStorage the last three are ignored by the
+    // storage backend so including them just costs a redundant write
+    // when cookie options change on a non-cookie store — rare and cheap.
+    //
+    // JSON.stringify can throw on BigInt / circular refs. We let the
+    // underlying storage layer keep its existing try/catch behaviour
+    // (log and drop) by falling through on serialization errors.
+    private _writeEntry(storage: PersistentStore, name: string, props: Properties, slot: string): void {
         try {
             const fingerprint =
```
pauldambra: > [!NOTE]

The stated goal is that a high-frequency main-blob write "no longer re-serializes" the flag+survey payloads — but `_writeEntry` always runs `JSON.stringify(props)` to build the fingerprint **before** the unchanged-check returns. With the gate on, every save still does `JSON.stringify(main)` + `JSON.stringify(flags)` + `JSON.stringify(surveys)`, so the large rarely-changing payloads are fully re-serialized on every SESSION_ID / super-prop write even though their `setItem` (and the cross-tab `storage` broadcast) is correctly skipped. The broadcast/IPC win — the dominant cost — is real and delivered; but the **serialization** cost the goal calls out is not actually saved (it's marginally higher than the pre-split single stringify, given 3 stringify calls + the `_partitionProps` O(n) pass). Non-blocking. To close the gap: track a per-group dirty flag (set in `_setProp`/`_deleteProp` keyed by the prop's `storageGroup`) or the last-written group object identity, and short-circuit `_writeEntry` before `JSON.stringify` for an unchanged group.

packages/browser/src/posthog-persistence.ts
```diff
@@ -222,6 +251,30 @@ export class PostHogPersistence {
...
+    // directly, never the cookie or the localPlusCookie re-write-on-parse path).
+    private _loadGroupEntries(): void {
+        for (const group of PERSISTENCE_STORAGE_GROUPS) {
+            // `localStore._parse` returns `{}` (not null) for a missing key, so
+            // gate on real content: an empty/absent entry is not "persisted" and
+            // must not be tracked, or `_writeNowSplit` would re-create it as `{}`.
+            const groupEntry = localStore._parse(this._groupEntryName(group))
+            if (groupEntry && !isEmptyObject(groupEntry)) {
+                this._persistedGroups.add(group)
+                extend(this.props, groupEntry)
```
pauldambra: > [!NOTE]

`_loadGroupEntries` does `extend(this.props, groupEntry)`, so a group entry unconditionally wins over the main blob with no freshness comparison. Hazard on a hard SDK downgrade-then-upgrade, or a mixed gate-on/gate-off fleet: a gate-ON tab writes `__flags=v2` and strips main; then a page loads under an OLDER SDK (or a tab where `__split_storage` resolved false) that doesn't know about `__flags`, fetches flags v3, and (gate off) writes the full cluster — including v3 — back into the MAIN blob, leaving the orphaned `__flags=v2` untouched (an old SDK never cleans it up). On the next gate-ON load, `props=main` (v3) is overwritten by `extend(props, __flags)` (v2) -> **stale flags served**. It's transient (self-heals on the next flags response) and partially masked when `feature_flag_cache_ttl_ms` is set, but with the default unset TTL there's no guard. Fix: prefer the newer source when both carry `$feature_flag_evaluated_at` (keep the larger timestamp instead of letting the group blindly win), or stamp a schema/version marker on the group entry that the loader validates; at minimum document that the split must roll out monotonically and gate the dated-default opt-in on that.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (reliability, data-integrity, compatibility, performance, adversarial generalist), paul-reviewer, xp-reviewer, security-audit

## Verdict: 🚫 BLOCKED

One **verified, convergent CRITICAL**: the new unconditional group-wipe in `remove()` is reachable from the `sessionStorage` sibling that shares the main instance's storage `_name`, so on the default config (localStorage+cookie, opted in by the `2026-05-30` dated default) the on-disk flag/survey cache is destroyed on nearly every page load — silently defeating the feature's core "cached flags survive reload" guarantee. Not a security or permanent-data issue, but it should block until fixed and covered by an integration test. CI is green precisely because the regression lives in the dual-instance `set_config` path that the isolation unit tests never exercise.

- 🔴 **`remove()` sibling collision wipes `__flags`/`__surveys`** — flagged independently by reliability (CRITICAL) and compatibility (HIGH); the unconditional removal was also independently called out by the adversarial generalist (byte-for-byte break) and xp (sessionPersistence reach-through). Verified end-to-end against the code.
- 🟡 **fingerprint set before a swallowed write** — generalist-b (reproduced silent flag loss) + reliability (pre-existing).

### Inline findings
- 🔴 CRITICAL — `posthog-persistence.ts:455` — sibling `remove()` wipes the split entries
- 🟡 MEDIUM — `posthog-persistence.ts:438` — fingerprint recorded before a swallowed `_set` -> silent flag loss on a scoped quota failure
- 🟡 MEDIUM — `posthog-persistence.ts:428` — perf goal partially unmet (always stringifies before the skip-check)
- 🟡 MEDIUM — `posthog-persistence.ts:275` — group-wins merge can serve stale flags on downgrade / mixed-fleet

### Low / nits (not posted inline, to keep signal high)
- 🟢 **paul — observability** (his one genuine ask): it's behind a config + dated default (good), but there's no signal to tell if the split is working in the wild or to watch adoption climb. A cheap one-time `$split_storage_migrated` diagnostic would let us funnel it on posthog-on-posthog before it's the default for everyone. The failure mode (silently lose a cached flag) is exactly the invisible kind.
- 🟢 **xp — `slot: string`** (his one real ask): Primitive Obsession; tighten to `PersistenceStorageGroup | typeof MAIN_STORAGE_SLOT` so the fingerprint-map invariant is compiler-enforced.
- 🟢 **data-integrity** — the changeset / doc comment says survey **seen-state** moves into `__surveys`; only `$surveys` does (seen-state lives in separate `seenSurvey_*` / `inProgressSurvey_*` keys, cleared via `surveys.reset()`). Tighten the wording so a future maintainer isn't misled.
- 🟢 **generalist-b** — `<name>__flags` can collide with another instance whose `persistence_name` ends in `__flags`; consider reserving the suffixes. Also a corrupted **array-shaped** `__flags` entry passes the `!isEmptyObject` guard and pollutes `props` — a cheap `isObject(groupEntry)` check before `extend` closes it.
- ⚪ **paul / xp polish** — `_buildStorage` returns the store **and** mutates `_splitStorageEligible` as a side effect (consider returning `{ store, splitEligible }`); `_partitionProps` adds a per-key policy lookup to the write path (memoise or comment the cost); the `_lastSavedSerialized` doc comment still points at `_writeNow` (the skip moved to `_writeEntry`); `__split_storage` reads opaquely for SDK users if it ever becomes a recommended setting.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 🔄 reliability | CRITICAL — sibling `remove()` destroys the on-disk cache on the default config; durability guarantee broken. |
| 📊 data-integrity | Key-routing is exact and safe (identity keys stay in main); one stale-flag merge-precedence gap on downgrade/mixed-fleet. |
| 🔗 compatibility | Dated-default opt-in, backend gating, ES5, terser list all correct; same sibling-collision bug (HIGH). |
| ⚡ performance | Broadcast/write win delivered; serialization win not (always stringifies before the skip-check). |
| 🕵️ generalist-b | Reproduced silent flag loss on a swallowed `__flags` quota write; plus collision / corruption robustness gaps. |
| 👤 paul | "really nicely put-together pr… high trust, stamp" — one ask: observability before it's the default. |
| 📐 xp | Hard change, well executed, tests express intent; one ask: tighten the `slot` type. |
| 🛡 security-audit | No exploitable issues; neutral-to-positive (tighter opt-out wipe, flag cluster kept out of the cookie). |

---

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `3ef62441`, test-first:

- **Repro (red):** added an integration test driving the real `set_config` -> `sessionPersistence` reconstruction path; before the fix it asserted the owner's `__flags`/`__surveys` came back `null` after `set_config({})` — i.e. the sibling wiped them.
- **Fix:** `PostHogPersistence` now takes an `ownsSplitStorage` flag (default `true`); `remove()`'s group-entry wipe is guarded by it, and the two `sessionPersistence` sibling sites in `posthog-core` pass `false`. The localStorage **owner** still cleans orphaned group entries on reset / opt-out / backend-downgrade (the orphan-cleanup intent is preserved), while the sessionStorage sibling — which shares the owner's storage name — no longer touches the shared keys. This is the "primary cleans orphans vs. non-owner never touches the shared keys" decouple, without the eligibility-guard's orphan-leak tradeoff.
- Test now green; persistence (355), core (30) and feature-flags (167) suites all pass; `terser-mangled-names.json` regenerated for the new private field.

Resolving — happy to reopen if you'd prefer a different decoupling (e.g. namespacing the sibling's entries instead of the ownership flag).

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `b573accb` (test-first).

- **Fix:** storage `_set` now returns whether the durable write landed (`storage.ts`, all backends; `PersistentStore._set` typed `=> boolean`). `_writeEntry` records the no-op fingerprint **only after** a confirmed-successful `_set`, so a swallowed quota/security failure scoped to `__flags`/`__surveys` leaves the slot un-fingerprinted and the next `save()` retries instead of stranding.
- **Repro (red->green):** new test "retries the __flags write on the next save after a failed (swallowed) write" stubs `localStore._set` to return `false` for the `__flags` name only; pre-fix the slot was cached as saved and never retried.
- persistence suite green (359).

Resolving — reopen if you'd prefer a different shape.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `b573accb` (test-first).

- **Fix:** per-slot dirty tracking (`_dirtySlots`). `_setProp`/`_deleteProp` mark a prop's storage group dirty via `_markGroupDirty`; `_writeEntry` short-circuits a group slot **before** `JSON.stringify` when it is not dirty and already written, so a high-frequency main-blob save no longer re-serializes the large unchanged flag/survey payloads. The main slot still always serializes (small; carries cookie options in its fingerprint).
- **Repro (red->green):** new test "skips JSON.stringify of __flags / __surveys when only a main key changed" spies on `JSON.stringify` and asserts no group payload is serialized after a main-only change.
- persistence suite green (359).

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `b573accb` (test-first).

- **Fix:** `_loadGroupEntries` no longer lets a group entry blindly win. `_groupEntryIsStale` compares `$feature_flag_evaluated_at` on the group entry vs the main blob; when both are numbers and main is strictly newer (the gate-off / older-SDK-wrote-fresher-flags-back-to-main case you described), the main cluster is kept and the orphaned group entry is healed by the next `save()` repartition. `surveys` (no timestamp) keeps prior behaviour.
- **Repro (red->green):** two new tests — "keeps the newer main-blob cluster when the group entry is older" and "still prefers the group entry when it is newer than the main blob".
- persistence suite green (359).

packages/browser/src/posthog-persistence.ts
```diff
@@ -222,6 +263,44 @@ export class PostHogPersistence {
...
+                // both carry a flag-eval timestamp and the main blob's is newer,
+                // keep the main blob and let the next save heal the group entry.
+                if (!this._groupEntryIsStale(groupEntry)) {
+                    extend(this.props, groupEntry)
+                }
+            }
+        }
+    }
+
+    private _groupEntryIsStale(groupEntry: Properties): boolean {
```
pauldambra: > [!NOTE]

`_groupEntryIsStale` keys entirely off `$feature_flag_evaluated_at`, so it only protects the `flags` group. The `surveys` group has no timestamp, so `isNumber(groupEvaluatedAt)` is always false and a stale `__surveys` entry always wins over a fresher survey payload written into the main blob by a gate-off / older-SDK tab. The downgrade-staleness scenario this commit closes for flags stays open for surveys. Either document that surveys are intentionally last-writer-wins, or carry a comparable freshness signal for the survey group. (Self-healing on the next `/surveys` response, hence MEDIUM not HIGH.)

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +384,110 @@ export class PostHogPersistence {
...
+    }
+
+    // Partition `props` by storage group and write each entry independently:
+    // the main blob without the grouped keys (stripping them completes the
+    // migration), plus one entry per group holding only its keys. Per-entry
+    // fingerprints mean a main-blob change does not rewrite the rarely-changing
+    // flag/survey entries, and vice-versa — which is the whole bandwidth win.
+    // Group entries go to `localStore` directly so they never hit the 4 KB
+    // cookie or the localPlusCookie re-write-on-parse path.
+    private _writeNowSplit(): void {
```
pauldambra, line 466: > [!NOTE]

**[paul]** 🟡 MEDIUM

torn-state + no observability. in `_writeNowSplit` the stripped main blob is written first, then each group; if a group `_set` returns false (quota) main is now persisted *without* flags and the flags entry never landed, so the cluster survives only in memory until the next successful save. the dirty-slot retry covers the common path, but a repeated group-write failure is completely invisible. could we `logger.warn` (debug-gated) when a group `_set` returns false? log before you limit — every dropped cache entry is pain. not blocking.

packages/browser/src/posthog-persistence.ts
```diff
@@ -571,9 +748,20 @@ export class PostHogPersistence {
...
     }
 
     private _deleteProp(prop: string): void {
         delete this.props[prop]
+        this._markGroupDirty(prop)
+    }
+
+    // Mark the prop's storage group dirty so its entry is re-serialized on the
+    // next write. Props with no group live in the main blob, which always writes.
+    private _markGroupDirty(prop: string): void {
```
pauldambra, line 867: > [!NOTE]

**[xp]** 🟡 MEDIUM

The group fast-path skip rests on an implicit three-part invariant: every `props` mutation marks its slot dirty, the skip checks `!dirty && lastSaved defined`, and `clear()` resets both caches together. It holds today because all writes route through `_setProp`/`_deleteProp` — but a future `this.props[x] = y` written outside those methods would silently break the skip and strand a stale group entry. A one-line comment at the `props` field declaration ("mutate only via `_setProp`/`_deleteProp` so `_dirtySlots` tracking stays correct") would protect the next maintainer cheaply. Not a bug today.

packages/browser/src/storage.ts
```diff
@@ -307,9 +309,11 @@ export const createLocalPlusCookieStore = (customCookieProperties: readonly stri
                 if (Object.keys(cookiePersistedProperties).length) {
                     cookieStore._set(name, cookiePersistedProperties, days, cross_subdomain, is_secure, debug)
                 }
+                return stored
```
pauldambra: > [!NOTE]

**[paul]** 🟢 LOW

i'm slightly allergic to this: `createLocalPlusCookieStore._set` returns `stored` (the localStorage result), but a cookie write throwing *afterwards* drops into the catch and returns `false` — even though localStorage already succeeded. that reports a durable write as failed, so the slot stays dirty and we re-serialize + re-write the already-persisted localStorage entry on every later save whenever cookies are flaky. no data loss, but it defeats the no-op-rejection win for `localStorage+cookie` users. consider returning the localStorage success and treating the cookie write as best-effort.

packages/browser/src/posthog-persistence.ts
```diff
@@ -71,11 +75,33 @@ export class PostHogPersistence {
...
+    private readonly _ownsSplitStorage: boolean
+    // Group slots already materialized on disk this session (loaded at startup
+    // or written since). Lets `_writeNowSplit` skip creating an entry for a
+    // still-empty group while still writing through to clear one that held data.
+    private _persistedGroups = new Set<PersistenceStorageGroup>()
+    // Group slots whose props changed since their last successful write. Lets a
+    // high-frequency main-blob save skip re-serializing the large, unchanged
+    // flag/survey payloads. Only group slots are tracked; the main slot always
+    // serializes (it is small and carries cookie options in its fingerprint).
+    private _dirtySlots = new Set<string>()
```
pauldambra: > [!NOTE]

`_dirtySlots` is `Set<string>` but only ever holds `PersistenceStorageGroup` values — the main slot is explicitly excluded. Typing it `Set<PersistenceStorageGroup>` would let the compiler stop anyone ever marking `'main'` dirty (mild Primitive Obsession; type-system-over-convention). Cheap and protects the invariant. Fine to defer.

packages/browser/src/__tests__/posthog-persistence.test.ts
```diff
@@ -876,6 +888,504 @@ describe('persistence', () => {
...
+
+            const serializedAGroupPayload = stringifySpy.mock.calls.some(
+                ([arg]) => arg && typeof arg === 'object' && (ENABLED_FEATURE_FLAGS in arg || SURVEYS in arg)
+            )
+            expect(serializedAGroupPayload).toBe(false)
+            jest.restoreAllMocks()
+        })
+    })
+
+    describe('a stale group entry does not overwrite fresher flags in the main blob', () => {
```
pauldambra: > [!NOTE]

`_groupEntryIsStale` uses strict `>`, so an *equal* timestamp lets the group entry win — almost certainly the right tie-break (the group is the migrated-forward canonical home), but it's load-bearing and untested. Add a third case asserting `mainTs === groupTs -> group wins`, which pins the boundary against a future well-meaning `>=` "fix". And these three really want to be one parameterized `it.each` over `(mainTs, groupTs, expectedWinner)`.

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +384,110 @@ export class PostHogPersistence {
...
+    // The fingerprint covers all four meaningful inputs to `storage._set`:
+    // serialized props, expire_days, cross_subdomain, secure. For
+    // localStorage / sessionStorage the last three are ignored by the
+    // storage backend so including them just costs a redundant write
+    // when cookie options change on a non-cookie store — rare and cheap.
+    //
+    // JSON.stringify can throw on BigInt / circular refs. We let the
+    // underlying storage layer keep its existing try/catch behaviour
+    // (log and drop) by falling through on serialization errors.
+    private _writeEntry(storage: PersistentStore, name: string, props: Properties, slot: string): void {
```
pauldambra: > [!NOTE]

**[qa-team/performance]** 🟢 LOW

When `JSON.stringify` throws (BigInt / circular ref) the catch sets `fingerprint = undefined`; on a successful `_set` the dirty flag clears but `_lastSavedSerialized[slot]` stays `undefined`. The fast-path keys "written before" off `!isUndefined(_lastSavedSerialized[slot])`, so a slot whose payload always throws will re-attempt `JSON.stringify` and re-write on *every* save — permanently defeating the no-op optimization for that entry. Narrow (needs a non-serializable value in a flag/survey payload) but it inverts the intended perf win for the affected slot.

packages/browser/src/types.ts
```diff
@@ -453,14 +453,17 @@ export interface PersistentStore {
     _error: (error: any) => void
     _parse: (name: string) => any
     _get: (name: string) => any
+    // Returns whether the durable write succeeded. Backends that swallow
```
pauldambra: > [!NOTE]

**[qa-team/data-integrity]** ⚪ NIT

The `_set` return is best-effort "did not throw" rather than "durably landed" — `localStore._set` returns `true` straight after `setItem` with no read-back, and Safari private mode has historically returned success on a write that did not persist. The comment "Returns whether the durable write succeeded" slightly overstates the guarantee; consider softening to "...whether the write was accepted without throwing".

pauldambra (commented):
QA Swarm review complete. See inline comments and the summary comment below.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists + generalists), paul-reviewer, xp-reviewer, security-audit. Target: the hardening commit `b573accb` (full PR diff in scope).

The hardening commit does exactly what it claims and all three fixes are sound and directly tested: the `_set` boolean + fingerprint-after-confirmed-write closes the write-failure stranding, per-slot dirty tracking delivers the serialization win, and `_groupEntryIsStale` stops a stale flag group entry clobbering a fresher main blob. Every reviewer said nothing blocks. The remaining items are edge-cases and cheap hardening, not merge blockers.

### Key findings
- 🟡 **MEDIUM** — staleness/downgrade protection is **flag-only**; the `surveys` group has no timestamp so a stale `__surveys` entry always wins (self-heals on next `/surveys`). _(qa-team/data-integrity)_
- 🟡 **MEDIUM** — **no observability** when a group `_set` fails: main is written stripped, the group write can silently fail (quota), cluster survives only in memory until next save. Add a debug-gated `logger.warn`. _(paul)_
- 🟡 **MEDIUM** — the fast-path skip rests on an **implicit invariant** (all `props` mutations route through `_setProp`/`_deleteProp`); a guarding comment at the `props` field would protect the next maintainer. _(xp)_
- 🟢 **LOW** — `localStorage+cookie` `_set` returns `false` when only the *cookie* write throws, masking a durable localStorage success and defeating no-op rejection. _(paul)_
- 🟢 **LOW** — equal-timestamp tie-break in `_groupEntryIsStale` is untested; parameterize the three staleness cases. _(paul + xp, convergent)_
- 🟢 **LOW** — `_dirtySlots` should be `Set<PersistenceStorageGroup>`, not `Set<string>`. _(paul + xp, convergent)_
- 🟢 **LOW** — a non-serializable flag/survey payload permanently defeats the no-op optimization for that slot (`JSON.stringify` throw leaves the fingerprint `undefined`). _(qa-team/performance)_
- ⚪ **NIT** — `_set`'s "durable write succeeded" comment overstates a best-effort "did not throw". _(qa-team/data-integrity)_

### Convergence
- **Equal-timestamp tie-break untested** — flagged independently by paul and xp.
- **`_dirtySlots` should be typed `PersistenceStorageGroup`** — paul (type-system-over-convention) and xp (primitive obsession) independently.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | LOW-to-MEDIUM. Fixes are correct and the highest-blast-radius interaction (sessionStorage sibling sharing the storage name) is double-guarded and tested; residuals are the surveys-staleness gap and a narrow stringify-throw perf edge. |
| 👤 paul | approve, ship as you see fit. wants a tie-break test (ideally parameterized) and debug-gated logging when a group write fails so a repeated quota failure isn't silent. |
| 📐 xp | A disciplined MakeItRight pass with no scope creep; the one cheap mitigation worth taking is a comment guarding the implicit dirty-tracking invariant. |
| 🛡 security-audit | No findings. Cross-user leakage on a shared browser (`reset()`/opt-out/runtime-split-off) is explicitly and correctly handled; no new sensitive data persisted; localStorage is same-origin. |

---

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Resolved in `8aa20392` — by documenting + testing the intentional behaviour. A robust fix is genuinely impossible without an end-to-end survey freshness signal: a gate-off/older-SDK write-back to the main blob is **indistinguishable** from a stale main-blob leftover during the one-shot migration (both are "main carries $surveys + a `__surveys` entry exists", but the desired winner is opposite). The flag group only works because `$feature_flag_evaluated_at` is supplied by the flags response, not the storage layer; surveys have no equivalent. I prototyped a `mainCarriesGroupKeys`-based generalization and it broke the migration test (`prefers the group entry over a stale value in the old main blob`), confirming the ambiguity. So the survey group entry stays last-writer-wins (self-healing on the next `/surveys`), now documented in `_loadGroupEntries` and pinned by a test.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `8aa20392`. `_writeEntry` now emits a debug-gated `logger.warn("failed to persist storage entry ...; will retry on next save")` when `storage._set` returns false, so a repeated quota failure on a group entry is no longer silent. Log before you limit. (Kept it debug-gated to avoid noise on the common path, as you suggested.)

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Partially addressed in `8aa20392`, and acknowledged. `_dirtySlots` is now `Set<PersistenceStorageGroup>`, so the compiler prevents the main slot ever being marked dirty — that closes the most likely break. The broader invariant (every `props` mutation routing through `_setProp`/`_deleteProp`) I've deliberately **not** documented with a prose comment: the repo owner keeps a strict no-narrative-comments house style, so a "mutate only via …" comment would cut against it. The type tightening is the durable guard here; flagging the trade-off rather than silently dropping it.

packages/browser/src/storage.ts
pauldambra: > [!NOTE]

Fixed in `8aa20392`. `createLocalPlusCookieStore._set` now captures the localStorage result up front and only the cookie mirror runs inside the try, so a cookie throw can no longer flip an already-landed localStorage write to `false`. Added a storage test asserting the write reports success even when `cookieStore._set` throws.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Fixed in `8aa20392`. `_dirtySlots` is now `Set<PersistenceStorageGroup>`; `_writeEntry`'s `slot` param is typed `PersistenceStorageGroup | typeof MAIN_STORAGE_SLOT` and the group-only `.delete(slot)` calls are guarded by the `!== MAIN_STORAGE_SLOT` narrow. The main slot can no longer be marked dirty by construction.

packages/browser/src/__tests__/posthog-persistence.test.ts
pauldambra: > [!NOTE]

Fixed in `8aa20392`. The three flag-staleness cases are now one parameterized `it.each` over `(mainTs, groupTs, expectedBeta)`, **including the equal-timestamp tie-break** asserting the group (canonical migrated-forward home) wins. Added survey staleness coverage alongside.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Looked at closely and resolving **as intended behaviour** rather than with a code change. The re-attempt on an un-fingerprinted slot is the *same* retry path that finding #1 (write-failure stranding) deliberately relies on — a slot stays dirty/uncached until a write confirms. For the BigInt/circular sub-case specifically: `localStore._set` itself does `JSON.stringify(value)` on the same object, so it also throws and returns `false` — the write genuinely never lands, so re-attempting on the next save is correct, not wasteful churn (no `setItem`/broadcast happens). Caching a fingerprint to suppress the retry would re-introduce exactly the stranding #1 fixes. Real flag/survey payloads are server JSON and always serializable, so this path isn't hit in practice. Happy to reopen if you'd prefer a last-object-identity short-circuit, but it'd trade correctness for a micro-optimization on an unreachable path.

packages/browser/src/types.ts
pauldambra: > [!NOTE]

Fixed in `8aa20392`. Softened the `_set` doc comment to "Returns whether the write was accepted without throwing … best-effort, not a durable-persistence guarantee — e.g. Safari private mode has historically reported success on a write that did not persist."

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]

Update: on a closer look my earlier "unfixable without an end-to-end signal, so documenting" was the wrong call — the signal was cheap to add. Fixed properly in `b693538a`:

- `getSurveys()` now stamps `$surveys_loaded_at` (in the `surveys` storage group) on every `/surveys` load — the direct analogue of `$feature_flag_evaluated_at`.
- `_groupEntryIsStale` is generalized to compare each group's own freshness timestamp (`GROUP_FRESHNESS_KEY`), so a stale `__surveys` entry no longer wins over a fresher main-blob write-back. With no timestamp on either side the group still wins (migrated-forward home), so the migration path is unchanged.
- Parameterized survey staleness tests (older/newer/equal) added, plus a no-timestamp fallback test.

Because the stamp ships with the split feature, every version in the split cohort carries it; the only residual is downgrading to a pre-split-feature SDK, the same monotonic-rollout caveat already documented for flags.

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +398,125 @@ export class PostHogPersistence {
...
+        for (const group of PERSISTENCE_STORAGE_GROUPS) {
+            const groupProps = groups[group]
+            // Don't materialize an entry just to hold `{}`: skip a group that is
+            // empty and has never been persisted. Once a group has held content
+            // we keep writing it (even when empty) so a later clear actually lands.
+            if (isEmptyObject(groupProps) && !this._persistedGroups.has(group)) {
+                continue
+            }
+            this._persistedGroups.add(group)
+            this._writeEntry(localStore, this._groupEntryName(group), groupProps, group)
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Good catch — fixed in `bb2d9d3a`. `_persistedGroups.add` is now done inside `_writeEntry` **only after a confirmed-successful `storage._set`** (alongside the existing fingerprint/dirty-slot bookkeeping), and the premature add in `_writeNowSplit` is removed. So a failed (e.g. quota) group write no longer marks the group as materialized on disk — the slot stays dirty/un-fingerprinted and retries, matching the documented contract. Added a regression assertion to the swallowed-write test: `_persistedGroups.has('flags')` is `false` after a failed write and `true` only after the retry lands. Persistence suite green.

packages/browser/src/persistence-key-policy.ts
```diff
@@ -79,8 +79,20 @@ import type { Properties, Property } from './types'
...
+/**
+ * Keys sharing a `storageGroup` are persisted together in their own storage
+ * entry (`<name>__<group>`) instead of the main persistence blob, when the
+ * `__split_storage` config is enabled. Keys without a group
+ * stay in the main blob. Members of one group are written in a single entry so
+ * an atomic `register` of the whole group lands as one write (no torn state).
+ */
+export type PersistenceStorageGroup = 'flags'
+
+export const PERSISTENCE_STORAGE_GROUPS: readonly PersistenceStorageGroup[] = ['flags']
```
pauldambra, line 91: an array just so we can follow up with surveys as the next split

packages/browser/src/posthog-persistence.ts
```diff
@@ -222,6 +282,49 @@ export class PostHogPersistence {
...
+    // the main blob. Group entries are localStorage-only (read via `localStore`
+    // directly, never the cookie or the localPlusCookie re-write-on-parse path).
+    private _loadGroupEntries(): void {
+        for (const group of PERSISTENCE_STORAGE_GROUPS) {
+            // `localStore._parse` returns `{}` (not null) for a missing key, so
+            // gate on real content: an empty/absent entry is not "persisted" and
+            // must not be tracked, or `_writeNowSplit` would re-create it as `{}`.
+            const groupEntry = localStore._parse(this._groupEntryName(group))
+            if (groupEntry && !isEmptyObject(groupEntry)) {
+                this._slotWriteState(group).persisted = true
```
haacked: `blocking`: A returning visitor still rewrites the whole flag blob and rebroadcasts it to every open tab once on every page load, which is the exact cost this split is meant to remove.

The reason: when the flags entry is loaded from disk, the code records that it exists but not a snapshot of its contents. So the first frequent main-blob save (which posthog does at startup, before fresh flags come back from the network) can't tell the flags entry is unchanged, and writes it again. It only settles down on the second save.

The `does not re-serialize __flags` test doesn't catch this because it writes the flags first, which records the snapshot, so it never starts from the load-from-disk state a real page load has.

Fix: record the snapshot when loading the entry, the same value a normal write produces, so the first save recognizes it as unchanged and skips it. A test that loads an existing entry into a fresh instance, changes only a main key, and asserts `localStorage` isn't written for `__flags` would lock it in.

packages/browser/src/posthog-persistence.ts
```diff
@@ -88,12 +133,14 @@ export class PostHogPersistence {
...
      */
-    constructor(config: PostHogConfig, isDisabled?: boolean) {
+    constructor(config: PostHogConfig, isDisabled?: boolean, ownsSplitStorage: boolean = true) {
         this._config = config
+        this._ownsSplitStorage = ownsSplitStorage
         this.props = {}
         this._campaign_params_saved = false
         this._name = parseName(config)
         this._storage = this._buildStorage(config)
+        this._splitStorage = this._splitStorageEligible && !!config['split_storage']
```
haacked: `suggestion`: This eligibility rule lives here and again in `update_config` (`const wantSplit = this._splitStorageEligible && !!config['split_storage']`). With two copies, a later change to one and not the other makes the constructor and the runtime toggle disagree about whether the split is on.

```suggestion
        this._splitStorage = this._resolveSplitStorage(config)
```

Pull it into a `private _resolveSplitStorage(config: PostHogConfig): boolean` helper and call it from both spots.

packages/browser/src/__tests__/posthog-persistence.test.ts
```diff
@@ -876,6 +887,459 @@ describe('persistence', () => {
...
+        const newerTs = 2000
+
+        // flags carry $feature_flag_evaluated_at, so freshness is compared by
+        // timestamp; an equal timestamp keeps the group entry (the canonical
+        // migrated-forward home).
+        it.each([
+            { name: 'group older than main -> main wins', mainTs: newerTs, groupTs: olderTs, expectedBeta: 'main' },
+            { name: 'group newer than main -> group wins', mainTs: olderTs, groupTs: newerTs, expectedBeta: 'group' },
+            { name: 'equal timestamps -> group wins', mainTs: newerTs, groupTs: newerTs, expectedBeta: 'group' },
+        ])('flags: $name', ({ mainTs, groupTs, expectedBeta }) => {
```
haacked, line 1646: `suggestion`: Every case here uses a real timestamp on both sides, so the "one side has no timestamp" path is never tested, and that's the path that protects against silently flipping which flags win.

When a group entry has no `$feature_flag_evaluated_at` (an older SDK, or a write from before the timestamp existed), `_groupEntryIsStale` is supposed to keep the group entry as the canonical one. The number checks (`isNumber(...)`) are what make that happen. If someone later simplifies the comparison to `mainTs > groupTs` and drops them, a missing timestamp would start losing to an undefined one and the cached flags would quietly change, with no test to flag it.

Add a case where one side omits `$feature_flag_evaluated_at` and assert the group still wins.

packages/browser/src/posthog-persistence.ts
```diff
@@ -295,42 +408,121 @@ export class PostHogPersistence {
...
+        this._writeEntry(this._storage, this._name, main, MAIN_STORAGE_SLOT)
+        for (const group of PERSISTENCE_STORAGE_GROUPS) {
+            const groupProps = groups[group]
+            // Don't materialize an entry just to hold `{}`: skip a group that is
+            // empty and has never been persisted. Once a group has held content
+            // we keep writing it (even when empty) so a later clear actually lands.
+            if (isEmptyObject(groupProps) && !this._slotState[group]?.persisted) {
+                continue
+            }
+            // `_writeEntry` records the group in `_persistedGroups` only after a
```
haacked: `nit`: This comment names `_persistedGroups`, a field that no longer exists (it's now `state.persisted` on `_slotState`), so the next reader will go looking for something that's gone. Worth updating the name.

haacked (commented):
Nice change. One blocking finding inline (a perf gap that undercuts the headline win on returning visitors), plus a couple of non-blocking notes.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Fixed in `635adb83`. `_loadGroupEntries` now seeds the no-op fingerprint from the snapshot it just read (`state.fingerprint = this._entryFingerprint(groupEntry, group)`), so the first startup save recognises the unchanged `__flags` and skips it.

Digging in, the seed alone wasn't enough for the real init path: posthog-core merges concrete cookie defaults *before* constructing persistence, so `set_cross_subdomain`/`set_secure` each fired `remove()+save()` on their undefined->value transition — churning `__flags` 3 writes + 2 removes per load. So two more changes complete the decoupling: (1) group fingerprints exclude cookie options (group entries are localStorage-only, scope-independent), and (2) those cookie-scope `remove()`s now keep the group entries. Net: an unchanged returning visitor writes `__flags` **zero** times.

New test `a returning visitor does not rewrite the unchanged __flags it loaded` is parameterised over both unset and concrete (production) cookie options — it asserts construction performs no `__flags` write and the following main-only `register` doesn't either.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Done in `635adb83`. Extracted `private _resolveSplitStorage(config: PostHogConfig): boolean` and call it from both the constructor and `update_config`, so the two can't drift.

packages/browser/src/__tests__/posthog-persistence.test.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Added in `635adb83`. New parameterised cases `flags: the group entry wins when {main omits the stamp | group omits the stamp | neither side carries a stamp}` — each asserts the group entry stays canonical, locking in the `isNumber` guards so a future simplification to `mainTs > groupTs` would fail the suite.

packages/browser/src/posthog-persistence.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Fixed in `635adb83` — comment now reads "marks the slot `persisted` (on `_slotState`)" instead of naming the removed `_persistedGroups` field.

pauldambra:
> [!NOTE]
>
> Multi-perspective review of the human-review fix commits: qa-team (data-integrity, reliability, compatibility + generalists), paul-reviewer, xp-reviewer, security-audit. Reviewed `635adb83`; the one blocking finding is resolved in `908fb4c5`.

## Verdict: ⚠️ REQUEST CHANGES → ✅ resolved in `908fb4c5`

The review found one **HIGH** regression introduced by the load-time fingerprint seed; it's now fixed and covered by tests in this push. Everything else is LOW/NIT.

### Convergent HIGH — fixed (`908fb4c5`)
**Seeded load fingerprint + `_writeEntry` dirty fast-path silently dropped a partial-migration leftover.** (flagged independently by qa-team/data-integrity and paul.)

Seeding `state.fingerprint` at load while leaving `state.dirty` falsy let the group fast-path (`!state.dirty && !isUndefined(state.fingerprint) → return`) short-circuit **before** the fingerprint is compared. So when `__flags` exists on disk **and** the main blob still carries a flag-group key absent from `__flags` (partial migration, a key newly grouped in a later SDK, or a mixed-fleet/stale tab that wrote a flag key back to main), the first save stripped that key from main but skipped writing `__flags` — losing it on the next reload, and never healing a stale orphan.

**Fix:** seed the fingerprint only when the main blob carries no key for that group (`_mainCarriesGroupKey`). The steady-state returning-visitor optimisation (the headline win) is preserved; the migration/heal write now lands. Added two tests: a main-leftover-folds-into-`__flags` case and a stale-`__flags`-heals-on-disk case (both fail before the fix, pass after).

### Non-blocking
- 🟢 **LOW** (paul): a repeatedly-failing group write strands the flag cache and the only signal is a `logger.warn` gated behind `config.debug` — worth an in-the-field, graphable signal before widening the gate. Pre-existing; logged as a follow-up, not addressed here.
- ⚪ **NIT** (paul + xp, convergent): `remove(true)` was a mystery boolean at the call site. **Done in `908fb4c5`** — `remove()` now takes `{ keepGroupEntries }`, so the cookie-option setters read `remove({ keepGroupEntries: true })`.
- ⚪ **NIT** (paul/xp): the main-slot fingerprint is a stringly-typed `|`-delimited concat. Pre-existing, fingerprint-not-security-boundary; left as-is.

### Cross-user leak check (the original worry): clear
security-audit returned **0 findings**. `reset()`/logout and the `update_config` migrate path go through `clear() → remove()` (default, full wipe incl. `__flags`/`__surveys`); opt-out goes through `set_disabled(true) → remove()` (full wipe). The `keepGroupEntries` path is only reachable from the cookie-option setters, which don't clear identity — no cross-user retention introduced.

### Reviewer summaries
| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | One HIGH data-integrity regression (seed vs dirty fast-path); mechanics otherwise sound, sibling guard correct. |
| 👤 paul | Same HIGH hole; good parameterised tests; nits on call-site boolean + debug-only failure logging. |
| 📐 xp | `keepGroupEntries` is a defensible scope flag, not a wrong-abstraction split; recommended an options object (done). |
| 🛡 security-audit | 0 findings — no cross-user/leak path; reset/opt-out still wipe everything. |

---

packages/browser/src/__tests__/posthog-persistence.test.ts
```diff
@@ -876,6 +887,586 @@ describe('persistence', () => {
...
+            }
+        })
+    })
+
+    describe('the session-scoped sibling must not wipe the localStorage owner', () => {
+        // posthog-core runs a second, sessionStorage-backed PostHogPersistence that
+        // shares the main instance's storage name. set_config reconstructs it, and a
+        // fresh instance's update_config -> set_secure calls remove() — which must not
+        // delete the localStorage owner's __flags entry.
+        it('set_config leaves the owner __flags entry intact', () => {
```
haacked, line 1554: `suggestion`: One coverage gap worth noting: the only path where `_ownsSplitStorage` is actually doing the work, a sibling whose persistence gets disabled and calls a bare `remove()` with no `keepGroupEntries` to save it, isn't tested. The cases here all preserve `__flags` via `keepGroupEntries` instead, so that guard could regress without anything catching it. Your call whether it's worth a case.

packages/browser/src/__tests__/posthog-persistence.test.ts
```diff
@@ -876,6 +887,586 @@ describe('persistence', () => {
...
+
+            lib.update_config(makeConfig(), off)
+
+            expect(parse(FLAGS)[ENABLED_FEATURE_FLAGS]).toEqual(FLAG_CLUSTER[ENABLED_FEATURE_FLAGS])
+            const main = parse(MAIN)
+            Object.keys(FLAG_CLUSTER).forEach((k) => expect(main[k]).toBeUndefined())
+            expect(main['distinct_id']).toBe('d')
+        })
+
+        it('turning the gate off folds grouped keys back into main and drops the group entries', () => {
```
haacked, line 1420: `suggestion`: The runtime-toggle tests all flip the gate on a fixed localStorage backend, so there's no coverage for `update_config` moving to a backend that can't host the split while the gate stays on (localStorage to memory, say). That's where the flags would silently strand if eligibility resolution regressed. Up to you whether to add one.

haacked (approved):
Nice work. A couple of test gaps I noticed, but nothing blocking.

packages/browser/src/__tests__/posthog-persistence.test.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Good catch — agreed the existing `set_config` test is saved by `keepGroupEntries` on the cookie-setter path, so it would still pass with the guard deleted. Added a case that makes `_ownsSplitStorage` load-bearing (`5c4e3b33`): a non-owning sibling (`ownsSplitStorage=false`, mirroring the sessionStorage instance posthog-core spins up) is disabled, which fires a **bare** `remove()` with no `keepGroupEntries`; the test asserts the owner's `__flags` survives.

Verified it's actually load-bearing: with the `!keepGroupEntries && this._ownsSplitStorage` guard temporarily reduced to `!keepGroupEntries`, the test goes red (the sibling wipes the owner's entry); restored, it's green.

(The stack also got rebased onto latest main in the same push — the trunk-drift in `posthog-core.ts` / `posthog-config.ts` / the config snapshot / mangled names is that resync, not part of this test.)

packages/browser/src/__tests__/posthog-persistence.test.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — not written by a human

Added (`b1852c88`). New case `drops the split when the backend becomes ineligible while the gate stays on`: gate on + localStorage, register flags, then `update_config` to `persistence: 'memory'` with `split_storage` still true. Asserts `_splitStorage` re-resolves to `false`, `__flags` is dropped, and the grouped keys fold back into the single in-memory blob (not stranded). The `_splitStorage === false` assertion is what would catch an eligibility-resolution regression directly.

pauldambra:
### Merge activity

* **Jun 8, 11:48 AM UTC**: @pauldambra merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/3721).

PR #3722, opened 2026-06-02
feat: add browser SDK dist channel property

marandaneto:
## Problem

Browser SDK events identify the SDK and version, but they don't tell which distribution channel loaded the SDK.

## Changes

- Add `$sdk_dist_channel` to browser SDK event properties when an entrypoint sets a distribution channel.
- Set `$sdk_dist_channel` to `npm` for module/npm package entrypoints.
- Set `$sdk_dist_channel` to `cdn` for snippet/array-loader entrypoints.
- Add unit coverage for unset, empty, npm, and cdn distribution channel behavior.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/browser/src/posthog-core.ts
```diff
@@ -4107,6 +4108,7 @@ export function init_from_snippet(): void {
 }
 
 export function init_as_module(): PostHog {
+    Config.SDK_INSTALL_SOURCE = 'npm'
```
turnipdabeets: nit: `npm` sounds a bit confusing because then I think what about yarn, pnpm etc... so maybe this should be called `module` or `bundler` to be more generic

turnipdabeets (approved):

packages/browser/src/utils/event-utils.ts
```diff
@@ -325,4 +325,10 @@ export function getEventProperties(
             $time: Date.now() / 1000, // epoch time in seconds
         }
     )
+
+    if (Config.SDK_INSTALL_SOURCE) {
+        properties['$sdk_install_source'] = Config.SDK_INSTALL_SOURCE
```
ioannisj: Should we extract $sdk_install_source to a constant? Similar to SDK_DEBUG_EXTENSIONS_INIT_METHOD for example

packages/browser/src/posthog-core.ts
```diff
@@ -4048,6 +4048,7 @@ const add_dom_loaded_handler = function () {
 }
 
 export function init_from_snippet(): void {
+    Config.SDK_INSTALL_SOURCE = 'script_loader'
```
ioannisj: Not sure if script_loader describes the "source" here. I was thinking of "cdn" but that also could not be the case if people self host array.js? Same goes for "npm" below. Good timing to make a conscious decision here but I don't have a better alternative in mind

ioannisj (approved):

packages/browser/src/posthog-core.ts
marandaneto: npm is the package manager repository as in https://www.npmjs.com/
theres no such thing for yarn, etc

turnipdabeets: ah ok so maybe `package_manager` then?

packages/browser/src/posthog-core.ts
marandaneto: its not about where its hosted but rather how its installed
option 1 https://posthog.com/docs/libraries/js#option-1-add-the-javascript-snippet-to-your-html-span-classbg-accent-text-gray-font-semibold-align-middle-text-sm-p-1-roundedrecommendedspan
option 2 npm|yarn|pnpm|etc install posthog-js which pulls from npmjs.com

marandaneto: how should we call this property then? Which values would differ between option 1 and 2?

packages/browser/src/posthog-core.ts
marandaneto: npm means node package manager
i prefer to be more specific to its name because we could use the same property for other sdks eg
for python that would be 'mypi'

marandaneto: another option is `$sdk_distribution_channel`, is it clearer?

turnipdabeets: ok maybe a small comment then? Just so we know which one and don't expect the others

marandaneto: or `$sdk_install_method`

packages/browser/src/posthog-core.ts
marandaneto: see this thread https://github.com/PostHog/posthog-js/pull/3722#discussion_r3342173228

packages/browser/src/posthog-core.ts
turnipdabeets: I think $sdk_distribution_channel is better

packages/browser/src/utils/event-utils.ts
marandaneto: Good catch! Fixed in 45e70214 — extracted the event property key to SDK_DIST_CHANNEL and renamed the config field/value to report $sdk_dist_channel as npm/cdn.

packages/browser/src/__tests__/sdk-install-source.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        expect(getEventProperties()).not.toHaveProperty('$sdk_install_source')
+    })
+
+    it('reports npm for module usage', () => {
+        Config.SDK_INSTALL_SOURCE = 'script_loader'
+
+        init_as_module()
+
+        expect(getEventProperties()['$sdk_install_source']).toBe('npm')
+    })
```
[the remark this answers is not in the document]
marandaneto: Fixed in 01f26641 — renamed the npm test to clarify it intentionally verifies overriding a prior dist channel.

packages/browser/src/__tests__/sdk-install-source.test.ts
```diff
@@ -0,0 +1,39 @@
...
+        Config.SDK_INSTALL_SOURCE = undefined
+
+        expect(getEventProperties()).not.toHaveProperty('$sdk_install_source')
+    })
+
+    it('does not report empty install source', () => {
+        ;(Config as any).SDK_INSTALL_SOURCE = ''
+
+        expect(getEventProperties()).not.toHaveProperty('$sdk_install_source')
+    })
```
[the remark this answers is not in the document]
marandaneto: Fixed in 01f26641 — merged the absent-property cases into a parameterized test for undefined and empty dist channel values.

dustinbyrne (approved):
very useful

turnipdabeets (approved):

PR #3737, opened 2026-06-03
fix(record): port rrweb 2.0.1 WebKit MutationObserver fix

arnohillen:
## Problem

I reviewed upstream rrweb's releases (`rrweb-io/rrweb`) against our vendored fork (`packages/rrweb/*`, `@posthog/rrweb`). The fork's upstream baseline is **`rrweb@2.0.0-alpha.18`**; upstream has since cut **`2.0.0`** and **`2.0.1`**.

The headline of `2.0.0` is a packaging/dist restructure (split plugin packages, new `@rrweb/browser-client`, new file extensions) that **we already do our own incompatible way** (our `@posthog/rrweb-*` workspace split + property mangling + custom vite build), so adopting it would fight our build. Stripping out docs/CI/`browser-client` noise, the **only functional code change** upstream landed after our baseline is the `2.0.1` WebKit fix.

That fix matters: on **WebKit/Safari**, when a 3rd-party library has monkey-patched the page, rrweb grabs a pristine `MutationObserver` constructor from a throwaway iframe to avoid the tainted one. WebKit tears down a **detached** iframe's `ScriptExecutionContext`, so the observer built from it silently stops delivering mutations once we remove the iframe ([webkit.org/b/179224](https://bugs.webkit.org/show_bug.cgi?id=179224)) — i.e. **no DOM mutations get recorded** on affected Safari pages.

## Changes

Ported [rrweb-io/rrweb#1854](https://github.com/rrweb-io/rrweb/pull/1854), reconciled against our diverged fork:

- `@posthog/rrweb-utils` (`getUntaintedPrototype`): on Safari, keep the untainted-prototype iframe **attached** (hidden via `display:none`, tagged `rr-block` + `__rrwebUntaintedMutationObserver` so it isn't serialized into snapshots) instead of removing it. `mutationObserverCtor()` now returns `[ctor, cleanup]`.
- `@posthog/rrweb` (`initMutationObserver` / `initObservers`): thread the `iframeCleanup` through and invoke it on observer teardown so we don't leak the iframe.
- Non-Safari behavior is unchanged (iframe still removed immediately; cleanup is a no-op).

Adapted to our fork's shape: our `initMutationObserver` returns an object (`{ observer, buffer }`) rather than upstream's tuple, so `shadow-dom-manager` needs no change (it just ignores the extra `iframeCleanup` field). Deliberately **not** porting the upstream WebKit Docker/CI test harness (`Dockerfile.webkit-test`, `vitest.config.webkit.ts`) — that's a separate CI infrastructure decision.

Typecheck + eslint clean on both packages.

### Libraries affected

- [x] posthog-js (web)

(Changeset bumps `posthog-js`, `@posthog/rrweb`, `@posthog/rrweb-utils` as patch.)

## Checklist

- [ ] Tests for new code — source-only port; the upstream regression test needs a WebKit CI runner we don't have wired up. Open to adding a Chromium-safe variant if wanted.
- [x] Accounted for the impact of any changes across different platforms — Safari-gated; non-Safari path byte-for-byte equivalent.
- [x] Accounted for backwards compatibility — `mutationObserverCtor` signature changed but it has a single internal caller; no external `dom.mutationObserver(` consumers.
- [x] Took care not to unnecessarily increase the bundle size — a few lines, Safari-gated.

### If releasing new changes

## 🤖 Agent context

Scope was deliberately narrowed from "sync all rrweb updates": a literal full merge of `alpha.18 → 2.0.1` (65 commits) would drag in upstream's 2.0.0 packaging restructure which conflicts with our custom build/mangling. The genuine functional delta since our baseline is just the `2.0.1` WebKit MutationObserver fix, which this PR ports.

### DIFF
--- a/packages/rrweb/rrweb/test/record/untainted-mutation-observer.test.ts
+++ b/packages/rrweb/rrweb/test/record/untainted-mutation-observer.test.ts
@@ -1,0 +1,60 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { JSDOM } from 'jsdom';
+
+const safariUA =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/18.3 Safari/605.1.15';
+const chromeUA =
+  'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/136.0.0.0 Safari/537.36';
+
+class HijackedMutationObserver {
+  constructor(_callback: MutationCallback) {}
+  observe() {}
+  disconnect() {}
+  takeRecords() {
+    return [];
+  }
+}
+
+async function loadUtils(userAgent: string) {
+  vi.resetModules();
+  const dom = new JSDOM('<!DOCTYPE html><html><body></body></html>', {
+    url: 'http://localhost',
+  });
+  const { window } = dom;
+  vi.stubGlobal('window', window);
+  vi.stubGlobal('document', window.document);
+  vi.stubGlobal('navigator', { userAgent });
+  vi.stubGlobal(
+    'MutationObserver',
+    HijackedMutationObserver as unknown as typeof MutationObserver,
+  );
+
+  return import('@posthog/rrweb-utils');
+}
+
+describe('untainted MutationObserver iframe', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals();
+    vi.resetModules();
+  });
+
+  it('keeps the iframe attached on Safari', async () => {
+    const utils = await loadUtils(safariUA);
+    utils.mutationObserverCtor();
+
+    const iframe = document.querySelector('iframe.rr-block');
+    expect(iframe).not.toBeNull();
+    expect(iframe?.getAttribute('__rrwebUntaintedMutationObserver')).toBe('');
+
+    utils.cleanupUntaintedIframe();
+    expect(document.querySelector('iframe')).toBeNull();
+  });
+
+  it('removes the iframe immediately on non-Safari browsers', async () => {
+    const utils = await loadUtils(chromeUA);
+    utils.mutationObserverCtor();
+
+    expect(document.querySelector('iframe')).toBeNull();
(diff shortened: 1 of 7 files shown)


packages/rrweb/utils/src/index.ts
```diff
@@ -107,27 +110,46 @@ export function getUntaintedPrototype<T extends keyof BasePrototypeCache>(
...
-      document.body.removeChild(iframeEl);
-    }
   }
 }
 
+export function cleanupUntaintedIframe(
+  key: keyof BasePrototypeCache = 'MutationObserver',
+): void {
+  untaintedBaseIframeCleanup[key]?.();
+  delete untaintedBaseIframeCleanup[key];
```
TueHaulund, line 150: Must-fix: this removes the iframe but never evicts `untaintedBasePrototype[key]`. Since `getUntaintedPrototype` early-returns the cached prototype, the next `record()` on Safari hands back a constructor bound to the now-removed iframe — so mutations record on the first session only and every later start/stop (SPA reinit, `reset()`, stop/restart) records nothing, the exact bug this PR fixes, one cycle later.

```suggestion
  untaintedBaseIframeCleanup[key]?.();
  delete untaintedBaseIframeCleanup[key];
  delete untaintedBasePrototype[key];
```

packages/rrweb/rrweb/test/record/untainted-mutation-observer.test.ts
```diff
@@ -0,0 +1,60 @@
...
+  it('keeps the iframe attached on Safari', async () => {
+    const utils = await loadUtils(safariUA);
+    utils.mutationObserverCtor();
+
+    const iframe = document.querySelector('iframe.rr-block');
+    expect(iframe).not.toBeNull();
+    expect(iframe?.getAttribute('__rrwebUntaintedMutationObserver')).toBe('');
+
+    utils.cleanupUntaintedIframe();
+    expect(document.querySelector('iframe')).toBeNull();
```
TueHaulund, line 50: This stops at one ctor → cleanup, so nothing here catches the bug above. Add a second `mutationObserverCtor()` after cleanup and assert a fresh `iframe.rr-block` is re-created — that fails today (the cached prototype is never evicted) and passes once it is.

PR #3739, opened 2026-06-04
chore(ai): move provider SDKs to optional peer dependencies + deprecation cleanup

### LINKED ISSUE #3610: Consider moving Google GenAI, LangChain, and LangSmith packages to "peerDependencies" to reduce supply-chain blast radius
### Summary

Hello! I'm currently trying to enforce stricter minimum ages for my company's monorepo, in light of the Mini Shai-Hulud / Shai-Hulud style attacks on TanStack, similar to the ones that affected PostHog late last year (https://github.com/PostHog/posthog-js/issues/2633).

I noticed `@posthog/ai` currently installs several provider/framework SDKs even when an application does not use those integrations.
 - `@google/genai`
 - `langchain`
 - `@langchain/core`
 - transitive dependencies like `langsmith`

If possible, I'd prefer these packages were instead optional peer dependencies! I'm using PostHog AI with the Vercel AI SDK, meaning these packages are installed and are additional vectors I wouldn't have otherwise.

### Current dependency chain

From @posthog/ai@7.16.13:

```text
   @posthog/ai
   ├── @google/genai
   ├── @langchain/core
   │   └── langsmith
   └── langchain
       └── langsmith
```

The package currently declares dependencies like:

```json
   {
     "@google/genai": "^1.43.0",
     "@langchain/core": "^1.1.29",
     "langchain": "^1.2.28"
   }
```

 ### Requested change

 Please consider moving these packages to optional peer dependencies, for example:

 ```json
   {
     "peerDependencies": {
       "@google/genai": "^1.43.0",
       "@langchain/core": "^1.1.29",
       "langchain": "^1.2.28"
     },
     "peerDependenciesMeta": {
       "@google/genai": {
         "optional": true
       },
       "@langchain/core": {
         "optional": true
       },
       "langchain": {
         "optional": true
       }
     }
   }
 ```

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

 - Mini Shai-Hulud / TanStack compromise writeup: https://www.aikido.dev/blog/mini-shai-hulud-is-back-tanstack-compromised
 - PostHog related incident issue: https://github.com/PostHog/posthog-js/issues/2633

turnipdabeets:
## Problem

Closes https://github.com/PostHog/posthog-js/issues/3610

`@posthog/ai` installs **every** provider SDK even if you use only one integration. A user on the Vercel AI SDK still gets `@google/genai`, `langchain`, and `@langchain/core` (which drags in `langsmith` + `langgraph`) into their `node_modules` — extra supply-chain surface for zero benefit (#3610, in light of the recent npm attacks).

Since this requires a major bump, it also bundles in the staged `@posthog/ai` deprecation cleanup so it all ships in one major release.

## Changes

**Optional peer dependencies (#3610)**

- Moved `openai`, `@anthropic-ai/sdk`, `@google/genai`, `@langchain/core` to optional `peerDependencies`, matching the pattern already used for `@ai-sdk/provider` / `@openai/agents`. They install only when you use that integration.
- Dropped `langchain` — never imported in `src/`; it only pulled in `langgraph` + `langsmith`.
- Integration clients now import from their **subpaths** (`@posthog/ai/openai`, `/anthropic`, `/gemini`, `/langchain`) instead of the root barrel, so importing the package no longer eagerly loads every SDK. The root still exports the SDK-agnostic surface (`withTracing`, `captureAiGeneration`, `Prompts`).
- Build fix: the rollup `external` config now externalizes dependency subpaths, so `@langchain/core` is required from the consumer's copy instead of being inlined into our bundle.
- Kept SDK types out of the root `.d.ts` (`captureAiGeneration`'s `tools` is `unknown[]`), so importing from the root needs no provider SDK installed.

**Deprecation cleanup (bundled into this major)**

- `Prompts.get()` now always returns a `PromptResult` (`{ source, prompt, name, version }`) — the plain-string return and `withMetadata` option are gone. Read the template from `result.prompt`.
- `PostHogTraceExporter` / `PostHogSpanProcessor` no longer accept the deprecated `apiKey` option — use `projectToken`.
- `projectToken` is now **required** on both (it was only optional to support the old `apiKey`). A blank token still disables the integration as a defensive runtime no-op.

**Breaking** (hence the `major` changeset): root imports move to subpaths and you install the SDK yourself, e.g. `import { OpenAI } from '@posthog/ai'` → `import { OpenAI } from '@posthog/ai/openai'` + `npm install openai`. Full migration diff is in the changeset.

Verified: build + lint + 416 unit tests pass; the root and `/vercel` bundles require only `@posthog/core` + `uuid`; root bundle dropped ~123 kB (−70%); `langsmith` is gone from the dependency tree.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


marandaneto:
Are there any more major breaking changes that we want to do so we can release them together?

turnipdabeets:
> Are there any more major breaking changes that we want to do so we can release them together?

Good call — I went through the @posthog/ai deprecations + API surface and folded in everything that's genuinely breaking so we only ship one major, that has been committed now. Also went through issues and all seems additive and not breaking. But I could have missed something.

.changeset/ai-optional-peer-deps.md
```diff
@@ -0,0 +1,22 @@
+---
+"@posthog/ai": major
```
marandaneto, line 2: having 2 major changesets is strange, lets merge them

marandaneto:
worth double-checking with @PostHog/team-ai-observability before merging, fine from my side

marandaneto:
i'd also update pr title/description since its doing more than just the peer deps now

packages/ai/src/otel/exporter.ts
```diff
@@ -33,14 +33,7 @@ function normalizeHost(value?: unknown): string {
...
  */
-export type PostHogTraceExporterOptions =
-  | { projectToken?: string; apiKey?: never; host?: string }
-  | {
-      /** @deprecated Use `projectToken` instead */
-      apiKey?: string
-      projectToken?: never
-      host?: string
-    }
+export type PostHogTraceExporterOptions = { projectToken?: string; host?: string }
```
marandaneto: `projectToken` can be non nullable now right

.changeset/ai-optional-peer-deps.md
turnipdabeets: Merged into a single `@posthog/ai` major changeset in d8ab5b05 — the optional-peer-deps and deprecation-removal sections now live in one file.

packages/ai/src/otel/exporter.ts
turnipdabeets: Done in deb77d4f — `projectToken` is now required on both `PostHogTraceExporter` and `PostHogSpanProcessor`. It was only optional to support the old `apiKey` alternative (exporter) / for the disable-when-missing path; with `apiKey` gone, required is correct. Kept the runtime no-op guard so a blank token (e.g. `process.env.PH_KEY ?? ''`) still disables defensively instead of firing empty-auth requests.

carlos-marchal-ph (commented):
Looks great, thanks for taking the time to do this. Especially appreciate the cleanup of deprecated stuff 👑 

Would you mind holding off on merging this until we merge [this PR I just opened too](https://github.com/PostHog/posthog-js/pull/3762)? It isn't necessarily a breaking change but I'd still feel way better if we pushed it as part of a major.

carlos-marchal-ph (approved):
Just merged the PR I mentioned above. Also did a round of testing locally and it works as expected. Green to go on our end!

PR #3749, opened 2026-06-05
feat(autocapture): ignore rageclicks on text inputs and steppers

pauldambra:
## Problem

Rageclick detection is purely positional (same spot, fast, N times) and can't tell intentional repeated clicking from frustration. Analysis of autocapture data showed a large share of `$rageclick` events land on interactions where rapid repeated clicks are the *expected* behaviour, not rage:

- **Text-editing surfaces** — double/triple-clicking inside an `input`/`textarea`/`contenteditable` to select a word or line. These had a markedly higher rageclick rate than genuine clickable surfaces like buttons and links.
- **Stepper / nav controls** — `+`/`-` quantity steppers and `next`/`prev` style buttons that are designed to be clicked repeatedly.

This over-captures rageclicks, dilutes the signal for real frustration, and adds noise to heatmaps.

## Changes

All new suppression is **gated behind the `2026-05-30` config defaults** — existing users and `rageclick: true` are unchanged.

- **Text selection** (`rageclick.ignore_text_selection`, new opt-in flag, on from `2026-05-30`): skip rageclick capture when the click target is a `textarea`, a text-like `input` (`text`/`search`/`email`/`password`/`url`/`tel`/`number`/no-type), or a `contenteditable` element. Still fires on `checkbox`/`button`/`submit` inputs, buttons, and links.
- **Steppers + nav**: `+`, `-`, `−`, `–` added to the default `content_ignorelist` from `2026-05-30` (extending the existing `next`/`prev`/`>`/`<` keywords).
- **Exact symbol matching**: `content_ignorelist` now matches symbol-only keywords (`+`, `-`, `>`, `<`) against the element's text *exactly* instead of as a substring. This is what makes `+`/`-` safe to add, and it also fixes the pre-existing over-broad behaviour where labels like `sign-up`, `5 > 3`, or `C++` were treated as repeatedly-clicked controls. Word keywords still match as substrings.
- **Heatmaps consistency**: the heatmaps `rageclick` marker now routes through `shouldCaptureRageclick`, so it respects the same suppression (and the pre-existing selector/content ignore lists) as the `$rageclick` event — previously it bypassed all element filtering.

### Backwards compatibility

- New behaviour is opt-in via the `2026-05-30` defaults date; `rageclick: true` and earlier defaults are untouched.
- One non-gated change: making symbol keywords exact-match slightly tightens `>`/`<` handling for existing `content_ignorelist: true` users (`2025-11-30`+). It only ever *reduces* false suppression (more correct), never the reverse. Happy to scope exact-matching to the new stepper keywords only if reviewers prefer `2025-11-30` to stay byte-for-byte identical.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

---


pauldambra:
* **#3750** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3750?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3749** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3749?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3749?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/autocapture-utils.ts
```diff
@@ -152,13 +152,19 @@ export function getParentElement(curEl: Element): Element | false {
...
 const MAX_CONTENT_IGNORELIST_ENTRIES = 10
 
 interface ElementWithText {
     safeText: string
     ariaLabel: string
 }
 
+// symbol keywords (e.g. +, -, >) match exactly so we don't suppress "sign-up", "5 > 3", "C++", etc.
+const matchesContentKeyword = (text: string, keyword: string): boolean =>
+    /[a-z0-9]/i.test(keyword) ? text.includes(keyword) : text === keyword
```
pauldambra, line 166: > [!NOTE]

**[convergent: paul + qa-team/compatibility]** 🟡 MEDIUM

The substring->exact change in `matchesContentKeyword` applies to **all** `content_ignorelist` paths, including `content_ignorelist: true` — the existing `2025-11-30` default, which already contains `>` and `<`. So users already on `2025-11-30` get a behaviour change on upgrade: a label whose text *contains* `>`/`<` (e.g. `home > settings`, `Learn more >`, `< Back`) was suppressed before and now isn't. It's gated by neither `2026-05-30` nor opt-in. It's arguably more correct, but the changeset frames the whole PR as a clean `2026-05-30` opt-in. Please add a sentence to the changeset noting the retroactive `>`/`<` effect, and add a legacy-config test (`content_ignorelist: true`): bare `>` still suppressed, `Learn more >` no longer suppressed.

packages/browser/src/autocapture-utils.ts
```diff
@@ -152,13 +152,19 @@ export function getParentElement(curEl: Element): Element | false {
...
+export const DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS = [...DEFAULT_CONTENT_IGNORELIST, '+', '-', '−', '–']
 const MAX_CONTENT_IGNORELIST_ENTRIES = 10
 
 interface ElementWithText {
     safeText: string
     ariaLabel: string
 }
 
+// symbol keywords (e.g. +, -, >) match exactly so we don't suppress "sign-up", "5 > 3", "C++", etc.
+const matchesContentKeyword = (text: string, keyword: string): boolean =>
```
pauldambra, line 165: > [!NOTE]

**[convergent: xp + paul + qa-team/correctness]** 🟡 MEDIUM

`matchesContentKeyword` picks exact-vs-substring by sniffing whether the keyword contains an alphanumeric char (`/[a-z0-9]/i`). The matching *strategy* is coupled to the *shape of the data* rather than named — a reader has to run the regex in their head. The accompanying comment is a hint an extract/rename is hiding. Consider explicit word/symbol lists or a `{keyword, exact}` shape so the rule is stated once. Also sanity-check the new `content_ignorelist` doc wording against the regex: a mixed keyword like `->` or `<3` is treated as a substring keyword, which a user reading "symbol-only keywords match exactly" might not expect.

packages/browser/src/autocapture-utils.ts
```diff
@@ -208,26 +216,58 @@ export function shouldCaptureDeadClick(el: Element | null, _config: PostHogConfi
 // autocapture check will already filter for ph-no-capture,
 // but we include it here to protect against future changes accidentally removing that check
 const DEFAULT_RAGE_CLICK_IGNORE_LIST = ['.ph-no-rageclick', '.ph-no-capture']
+
+const TEXT_SELECTION_INPUT_TYPES = ['', 'text', 'search', 'email', 'password', 'url', 'tel', 'number']
+
+function isContentEditableTarget(el: Element): boolean {
```
pauldambra, line 222: > [!NOTE]

**[convergent: qa-team/compatibility + xp + paul]** 🟡 MEDIUM

jsdom does not implement `HTMLElement.isContentEditable`, so the unit tests only exercise the `getAttribute('contenteditable')` fallback — the `el.isContentEditable` property branch is untested. That branch is exactly what covers the most common real-browser case: the rageclick target is usually the inner clicked node *inside* a `contenteditable` container, which has no `contenteditable` attribute of its own and relies entirely on the inherited `isContentEditable` property. Worth a Playwright/integration check (or at least an explicit note that this path is browser-verified), since jsdom gives false confidence here. Minor related gap: `contenteditable="plaintext-only"` is missed on the attribute path (the property branch covers it in a real browser).

pauldambra (commented):
> [!NOTE]

QA Swarm review — inline findings below. Verdict: 💬 APPROVE WITH NITS (no blockers).

packages/browser/src/heatmaps.ts
```diff
@@ -236,7 +236,10 @@ export class Heatmaps implements Extension {
 
         const properties = this._getProperties(e, type)
 
-        if (this.rageclicks?.isRageClick(e.clientX, e.clientY, new Date().getTime())) {
+        if (
```
pauldambra, line 239: > [!NOTE]

**[convergent: xp + qa-team/reliability]** 🟡 MEDIUM

Good DRY win routing the heatmap `rageclick` marker through `shouldCaptureRageclick` — it removes the drift where the heatmap marker ignored all element suppression. Two notes: (1) no test covers this new heatmaps suppression (the existing heatmaps test targets `document.body`, which is never suppressed) — add a case asserting a suppressed target downgrades the marker to `click`. (2) `isRageClick(...)` mutates the click buffer as a side effect and is correctly evaluated first (matching autocapture.ts order); this is fine **because** heatmaps and autocapture each own a separate `RageClick` instance, so the buffers don't desync. Worth confirming that stays true.

packages/browser/src/autocapture-utils.ts
```diff
@@ -152,13 +152,19 @@ export function getParentElement(curEl: Element): Element | false {
 }
 
 const DEFAULT_CONTENT_IGNORELIST = ['next', 'previous', 'prev', '>', '<']
+// +/- steppers are built to be clicked repeatedly; enabled from the 2026-05-30 config defaults
+export const DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS = [...DEFAULT_CONTENT_IGNORELIST, '+', '-', '−', '–']
```
pauldambra, line 156: > [!NOTE]

**[convergent: xp + qa-team/correctness]** ⚪ NIT

`DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS` is 9 entries vs `MAX_CONTENT_IGNORELIST_ENTRIES = 10` — one spare slot. If more glyph variants are added later (em-dash `—`, fullwidth `＋`) the default silently approaches the cap, and crossing it makes `shouldIgnoreByContent` return `false`, disabling **all** content filtering rather than just dropping the extra keyword. Thin headroom for a list that ships to everyone on `2026-05-30`.

packages/browser/src/posthog-core.ts
```diff
@@ -190,7 +191,12 @@ const defaultsThatVaryByConfig = (
     | 'internal_or_test_user_hostname'
     | 'persistence_save_debounce_ms'
 > => ({
-    rageclick: defaults && defaults >= '2025-11-30' ? { content_ignorelist: true } : true,
+    rageclick:
```
pauldambra, line 194: > [!NOTE]

**[paul]** 🟢 LOW

The nested ternary for `rageclick` across three date thresholds now takes two reads. Not blocking, but a fourth threshold makes it four levels deep — a small `rageclickDefaultsFor(defaults)` helper alongside the others would read more like a table. Ship as you see fit.

packages/browser/src/autocapture-utils.ts
```diff
@@ -208,26 +216,58 @@ export function shouldCaptureDeadClick(el: Element | null, _config: PostHogConfi
 // autocapture check will already filter for ph-no-capture,
 // but we include it here to protect against future changes accidentally removing that check
 const DEFAULT_RAGE_CLICK_IGNORE_LIST = ['.ph-no-rageclick', '.ph-no-capture']
+
```
pauldambra, line 219: > [!NOTE]

**[paul]** 🟢 LOW

Observability: a suppressed rageclick just silently `return false`s. The only production signal that this did the right thing (vs over-suppressing) is rageclick volume on the `2026-05-30` cohort, which conflates text-selection + steppers + the `>`/`<` change. Not asking for SDK-side counters (expensive) — just worth deciding up front how we'll confirm we didn't over-suppress, so we don't find out from a "where did my rageclicks go" ticket.

packages/browser/src/__tests__/autocapture.test.ts
```diff
@@ -566,6 +566,105 @@ describe('Autocapture system', () => {
...
+                        expect(rageClickThreeTimes(buttonWithText(text))).not.toContain('$rageclick')
+                    }
+                )
+
+                it.each(['C++', '5 > 3', 'sign-up', 'Save'])(
+                    'rapid clicks on a "%s" button still capture $rageclick (symbol keywords match exactly)',
+                    (text) => {
+                        expect(rageClickThreeTimes(buttonWithText(text))).toContain('$rageclick')
+                    }
+                )
```
pauldambra, line 653: > [!NOTE]

**[convergent: xp + qa-team/tests]** 🟢 LOW

The parameterized tests are exactly the table-driven shape this repo asks for — nice. Two small things: (1) this final `it.each` asserts two unrelated behaviours (textarea + `+` stepper) in one body; on failure you won't immediately know which suppression regressed — split or add messages. (2) No case asserts the one backwards-compat-affecting change under `content_ignorelist: true` (bare `>` still suppressed, `Next >` no longer) — adding it would lock in the intended legacy behaviour.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists), paul-reviewer, xp-reviewer, security-audit

No CRITICAL/HIGH findings and no blockers — the new text-selection + stepper suppression is correctly gated behind the `2026-05-30` defaults and the gating is proven by tests. The flags worth the author's attention are compatibility/coverage gaps, not bugs.

### Key findings

- 🟡 **MEDIUM** — the `>`/`<` substring→exact change is **unconditional**, so it silently changes rageclick volume for users already on the `2025-11-30` `content_ignorelist: true` default (not just `2026-05-30` adopters). Likely an improvement; needs a changeset note + a legacy-config test.
- 🟡 **MEDIUM** — `isContentEditable` is unimplemented in jsdom, so the common real-browser case (clicking a child inside a `contenteditable` container) is untested. Consider a Playwright check.
- 🟡 **MEDIUM** — `matchesContentKeyword` infers exact-vs-substring by regex-sniffing the keyword; strategy is implicit. Consider explicit word/symbol lists.
- 🟡 **MEDIUM** — heatmaps suppression path has no test (existing test targets `document.body`, never suppressed).
- 🟢 **LOW** — nested-ternary depth in `posthog-core.ts`; over-suppression observability; split the two-behaviour test.
- ⚪ **NIT** — `DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS` is 9/10 entries; thin headroom before the cap disables all content filtering.

### Convergence

The `>`/`<` retroactive compat, the implicit `matchesContentKeyword` strategy, the `isContentEditable` test gap, the missing heatmaps test, and the entry-count headroom were each flagged independently by 2+ reviewers — highest confidence.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Well-scoped and well-tested; APPROVE WITH NITS — two compat/coverage gaps (`>`/`<` retroactivity, `isContentEditable` jsdom blind spot). |
| 👤 paul | Solid, tests are the right parameterized shape; main flag is the retroactive `>`/`<` change reading as a clean opt-in when it isn't. |
| 📐 xp | Good XP hygiene and a real OnceAndOnlyOnce win on heatmaps; the one wrinkle is `matchesContentKeyword` coupling strategy to data shape. |
| 🛡 security-audit | No findings — client-side only; the new regex is not ReDoS-prone and suppression doesn't broaden captured data. |

---

packages/browser/src/__tests__/autocapture.test.ts
```diff
@@ -566,6 +566,105 @@ describe('Autocapture system', () => {
...
+            const rageClickThreeTimes = (el: Element): string[] => {
+                document.body.appendChild(el)
+                const fakeEvent = makeMouseEvent({ target: el, clientX: 5, clientY: 5 })
+                Object.setPrototypeOf(fakeEvent, MouseEvent.prototype)
+                autocapture['_captureEvent'](fakeEvent)
+                autocapture['_captureEvent'](fakeEvent)
+                autocapture['_captureEvent'](fakeEvent)
+                const captured = beforeSendMock.mock.calls.map((args) => args[0].event)
+                document.body.removeChild(el)
+                return captured
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `152626de` — added `autocapture[\'rageclicks\'].clicks = []` at the start of `rageClickThreeTimes` to reset the rage-click buffer between calls, and `beforeSendMock.mockClear()` after capturing to isolate mock state. Both buffers now reset between invocations, so multi-element tests in the same `it.each` body are independent.

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

Fixed in `152626de`:
- Updated `.changeset/rageclick-ignore-intentional-clicks.md` with a note that `content_ignorelist: true` users (2025-11-30 config) will see the same `>`/`<` substring->exact change on upgrade, and what to do if they relied on the old behaviour.
- Added a `describe` block in `autocapture.test.ts` covering `content_ignorelist: true` (legacy): bare `>` and `<` are still suppressed; `Learn more >`, `< Back`, `Go >`, `home > settings` are not suppressed (symbol keywords now match exactly).

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

Deferred — converting `matchesContentKeyword` to explicit word/symbol lists (or a `{keyword, exact}` shape) is an architectural decision that changes the public mental model for `content_ignorelist` entries. The existing comment is meant to explain the implicit contract; whether to make it explicit via a richer data shape is a design call for the PR author. Left for human review.

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

Deferred — adding a Playwright/integration test for the `isContentEditable` property path (inherited editability on a child node) requires real-browser test infrastructure that is out of scope for this PR. The unit test correctly exercises the `getAttribute('contenteditable')` fallback path; the property-path gap and the `plaintext-only` note are surfaced for the PR author to decide whether a follow-up issue is warranted.

packages/browser/src/heatmaps.ts
pauldambra: > [!NOTE]

Fixed in `152626de` — added `should downgrade rageclick to click for suppressed targets` to `heatmaps.test.ts`: three rapid clicks on a `+` stepper with `content_ignorelist: DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS` produces three `click` entries and no `rageclick` entry in the heatmap buffer.

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

NIT — the one-slot headroom is intentional thin guard for the current set; the cap-exceeded path disabling all content filtering is worth documenting if more glyph variants are added. Out of scope for this PR.

packages/browser/src/posthog-core.ts
pauldambra: > [!NOTE]

NIT — the nested ternary is intentional for now; a helper could be extracted if a fourth threshold lands. Leaving for the PR author.

packages/browser/src/autocapture-utils.ts
pauldambra: > [!NOTE]

NIT / heads-up — no SDK-side counter is being added (the observation is correct: the only signal is rageclick volume on the 2026-05-30 cohort). Surfacing for the PR author to decide up-front how over-suppression would be detected. No code change needed here.

packages/browser/src/__tests__/autocapture.test.ts
pauldambra: > [!NOTE]

The legacy-config test (bare `>` suppressed, `Learn more >` not suppressed under `content_ignorelist: true`) was added in `152626de` as part of the changeset/test fix. The assertion split for textarea vs. stepper in the `it.each` is a minor readability nit — leaving for the PR author.

packages/browser/src/heatmaps.ts
```diff
@@ -236,7 +236,10 @@ export class Heatmaps implements Extension {
...
+            this.rageclicks?.isRageClick(e.clientX, e.clientY, new Date().getTime()) &&
+            shouldCaptureRageclick(getEventTarget(e), this.instance.config.rageclick)
```
dustinbyrne, line 241: `isRageClick` mutates state to track clicks, regardless of whether or not `shouldCaptureRageclick` discards the click afterwards. So two clicks in a text input followed by one misclick just outside could trigger a rage click.

Seems relatively minor / edge-case-y, but should these conditions be checked in reversed?

packages/browser/src/posthog-core.ts
```diff
@@ -190,7 +191,12 @@ const defaultsThatVaryByConfig = (
...
+        defaults && defaults >= '2026-05-30'
+            ? { content_ignorelist: DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS, ignore_text_selection: true }
```
dustinbyrne, line 196: Should we merge the default config with the user supplied config? If they don't supply all of these default properties, they'll get dropped. E.g., consider:
```ts
rageclick: { threshold_px: 30 }
```

dustinbyrne (approved):
Left a couple of comments, but nothing blocking. Looks good overall 👍

packages/browser/src/heatmaps.ts
pauldambra: > [!NOTE]
> 🤖 Automated reply drafted by an agent (PostHog Code) on @pauldambra's behalf — not written by a human

Good catch, and you'\''re right it'\''s a real (if minor) edge case. Keeping the current order deliberately:

`isRageClick` is the cheap O(1) gate; `shouldCaptureRageclick` walks the ancestor chain and calls `getSafeText` per node. Checking suppression first would run that DOM work on **every** click, not just the rare ones that complete a rage sequence. The same `isRageClick`-first order is in `autocapture.ts`, and the two paths hold independent `RageClick` buffers, so they stay consistent.

The "suppressed clicks still prime the buffer" behaviour is also pre-existing — it already applied to `.ph-no-rageclick` and `content_ignorelist` targets; this PR just widens the suppressed set. If it bites in practice, the cheap fix is a tag/type-only pre-gate for text-selection targets (no DOM walk) before `isRageClick` — happy to do that as a follow-up rather than reorder the whole check.

packages/browser/src/posthog-core.ts
pauldambra: > [!NOTE]
> 🤖 Automated reply drafted by an agent (PostHog Code) on @pauldambra's behalf — not written by a human

Agreed — fixed in `09fd72f1`. A partial `rageclick` object is now deep-merged with the date-gated defaults, so `rageclick: { threshold_px: 50 }` keeps `content_ignorelist` / `ignore_text_selection`. Explicit sub-options still override (`{ content_ignorelist: false }`) and a boolean still replaces the whole thing. Added tests + a changeset note.

PR #3751, opened 2026-06-05
feat(ai): populate $ai_base_url on Vercel and OpenAI Agents paths

richardsolomou:
## Problem

Ingestion dedups gateway-routed `$ai_generation` events by checking whether `$ai_base_url` is a PostHog gateway host. The Vercel and OpenAI Agents paths never set it, so gateway calls instrumented through them couldn't be detected.

## Changes

- Vercel: recover the base URL from the provider's internal `config` (the `config.url({ path })` closure or `config.baseURL`), replacing the hardcoded `''`. Verified against the real published `@ai-sdk` providers.
- OpenAI Agents: emit it from `model_config.base_url` when the SDK exposes it. Best-effort, since the SDK omits it for Responses calls and chat calls with model settings.
- OTel already maps `server.address` to `$ai_base_url` server-side, so no SDK change there.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

---

### DIFF
--- a/packages/ai/src/vercel/middleware.ts
+++ b/packages/ai/src/vercel/middleware.ts
@@ -24,7 +24,7 @@ import {
 } from '../utils'
 import { captureAiGeneration } from '../captureAiGeneration'
 import { redactBase64DataUrl } from '../sanitization'
-import { isString } from '../typeGuards'
+import { isObject, isString } from '../typeGuards'
 
 // Union types for dual version support
 type LanguageModel = LanguageModelV2 | LanguageModelV3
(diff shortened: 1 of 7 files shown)


packages/ai/tests/vercel.test.ts
```diff
@@ -1140,4 +1140,76 @@ describe('Vercel AI SDK - Dual Version Support', () => {
...
+
+    it('recovers the base URL from a config.baseURL string (@ai-sdk/anthropic style)', async () => {
+      const props = await runGenerate({ baseURL: 'https://gateway.posthog.com/anthropic' })
+      expect(props['$ai_base_url']).toBe('https://gateway.posthog.com/anthropic')
+    })
+
+    it('falls back to an empty string when config is absent or unrecognized', async () => {
+      expect((await runGenerate(undefined))['$ai_base_url']).toBe('')
+      expect((await runGenerate({ somethingElse: true }))['$ai_base_url']).toBe('')
+    })
```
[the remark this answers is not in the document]
richardsolomou: done — split into two tests so each asserts its own capture call

brandonleung (approved):

packages/ai/src/vercel/middleware.ts
```diff
@@ -283,6 +283,35 @@ const extractProvider = (model: LanguageModel): string => {
...
+        path: '',
+        modelId: model.modelId,
+      })
+      return typeof url === 'string' ? url : ''
+    }
+  } catch {
+    // Unknown config shape or url() threw.
+  }
+  return ''
+}
```
carlos-marchal-ph, line 311: Since we're coupling this to SDK internals that we can't control, I'd much rather we use an actual type guard than a type cast here. Something might change in the future and bite us in the ass.

carlos-marchal-ph (approved):

packages/ai/src/vercel/middleware.ts
richardsolomou: done — swapped the casts for the existing `isObject`/`isString` guards plus `typeof` narrowing, so a future shape change degrades to '' instead of sailing through. only the off-spec `config` read stays a (minimal, honest) cast since the `LanguageModel` type doesnt declare it.

PR #3756, opened 2026-06-07
fix: drop events whose token property is removed by before_send (#3438)

### LINKED ISSUE #3438: `before_send` silently strips `properties.token` → ingest 401 "event submitted without an api_key"
## Summary

When a user-supplied `before_send` hook removes the `properties.token` key from an event (e.g. via a generic PII scrubber that matches `/token/i`), every subsequent event is rejected by ingest with HTTP 401 `event submitted without an api_key`.

There is no documented list of reserved property keys that `before_send` must preserve, and the SDK performs no post-`before_send` safety check to re-assert `properties.token = config.token`. This is a silent footgun.

## SDK version

- Reproduced on `posthog-js@1.368.2` (last-known-good prior to 1.369.x).
- Source path still identical on `main` (`packages/browser/src/posthog-core.ts`).

## Reproduction

1. Initialize posthog-js normally, with any valid token and api_host.
2. Add a `before_send` hook that deletes keys matching a generic token-redaction regex:

```ts
posthog.init('phc_...', {
  api_host: '...',
  before_send: (event) => {
    if (!event) return null
    if (event.properties) {
      for (const key of Object.keys(event.properties)) {
        if (/token/i.test(key)) {
          delete event.properties[key]  // strips the project api_key
        }
      }
    }
    return event
  },
})
```

3. Fire any event. Observe in DevTools Network:
   - Request to `/i/v0/e/?ip=0&_=...&ver=1.368.2&compression=gzip-js`
   - No `api_key` query parameter
   - Decoded gzip body is a bare JSON array with no `token` anywhere in event properties
   - Response: HTTP **401** `event submitted without an api_key`

## A/B confirmation (manual curl against the same endpoint)

Same URL, headers, compression. Only difference is the `token` field inside event properties.

| Body | Response |
|---|---|
| `[{"uuid":"...","event":"test","properties":{"token":"phc_...","distinct_id":"..."}}]` | **200** `{"status":"Ok"}` |
| `[{"uuid":"...","event":"test","properties":{"distinct_id":"..."}}]` | **401** `event submitted without an api_key` |

The api_key is carried *only* via `properties.token` inside the event — the `/i/v0/e/` path takes no `?api_key=` query parameter.

## Code path

`packages/browser/src/posthog-core.ts` — `calculateEventProperties`:

```ts
```

`before_send` runs after this (`_runBeforeSend` is called from the capture pipeline). If `before_send` returns an event object whose `properties` no longer contains `token`, the SDK sends the tokenless payload to ingest as-is.

## Why this is an easy foot-gun

1. The `token` property uses a human-sensible name (rather than e.g. `$token` or `$api_key`), so it's indistinguishable from an app-level token field a PII scrubber would legitimately target (e.g. `auth_token`, `reset_token`, `magic_link_token`).
2. There's no public enumeration of reserved keys that `before_send` must leave alone.
3. The SDK emits no warning when `before_send` returns an event without the required `token`.
4. The ingest error ("event submitted without an api_key") is opaque from the SDK side — it surfaces as a 401 the SDK just logs without pointing at the likely cause.

### 1. SDK safeguard — re-assert `token` after `before_send`

In `_runBeforeSend` (or at the enqueue boundary), re-assign `properties.token` to `this.config.token` if `before_send` dropped it. This is the fully backstopped behavior:

```ts
const result = beforeSendFn(event)
if (result && result.properties) {
  result.properties.token ??= this.config.token
  result.properties.distinct_id ??= /* existing distinct_id resolution */
}
return result
```

### 2. SDK warning when `before_send` drops `token`

If `result.properties.token` is missing after the hook runs, `logger.warn(...)` once with a pointer at the likely cause. Helps users who use `before_send` as a transform layer and want the explicit signal rather than the silent fix above.

### 3. Docs — publish the reserved-keys list

Add a section to the `before_send` docs (`posthog.com/docs/libraries/js/features`) enumerating the property keys the SDK relies on and warning that `before_send` must not drop them:

- `token`
- `distinct_id`
- `$session_id`
- `$window_id`
- `$lib`, `$lib_version`
- `$insert_id`
- `$device_id`
- `$time`

Even without the SDK safeguard, this alone would prevent the class of bug.

## Workaround

Downstream fix that landed in our codebase — maintain an explicit allowlist of reserved keys and skip them before the pattern scan:

```ts
const PROTECTED_KEYS = new Set(['token', 'distinct_id'])

before_send: (event) => {
  if (!event?.properties) return event
  for (const key of Object.keys(event.properties)) {
    if (PROTECTED_KEYS.has(key)) continue
    if (PII_KEY_PATTERNS.some((p) => p.test(key))) delete event.properties[key]
  }
  return event
}
```

Happy to send a PR for any of the three fixes above if there's maintainer interest.

---

Context: came up during production PostHog integration on a multi-tenant web app behind a `/relay-ph/*` reverse proxy. Took a multi-round investigation to pin down because the cookie proved init ran correctly, events appeared to flow on the network tab, and the 401 body was the only actionable signal.

archievi:
## Problem

Fixes #3438.

The project `api_key` is sent to ingest inside each event as `properties.token` (set in `calculateEventProperties`). User `before_send` hooks run after that. If a hook removes the `token` key — e.g. a generic PII scrubber matching `/token/i`, since the key is named like an app-level `auth_token`/`reset_token` field — the SDK sends a tokenless payload and every event is rejected by ingest with `HTTP 401 event submitted without an api_key`. There is no warning, so it's a silent footgun.

## Changes

Following review feedback from @marandaneto, this **drops the event and logs a warning** when a `before_send` hook removes a required property, rather than restoring it (the removal may be intentional).

Mirroring the existing `knownUnsafeEditableEvent` / `isKnownUnsafeEditableEvent` pattern:

- Adds `knownUnsafeEditableEventProperty` (currently just `['token']`) and an `isKnownUnsafeEditableEventProperty` predicate in `@posthog/core`.
- In `_runBeforeSend`, snapshots which of these required properties were present *before* the hook chain runs (so an in-place mutation can't hide a removal), then after the chain, if such a property is now missing from a still-present `properties` object, logs a warning and drops the event (returns `null`).

Scope note: this only triggers when `properties` still exists but a required key was removed (the reported scrubber case). The pre-existing "no properties after beforeSend" path is left unchanged.

Updates the regression test to assert the event is dropped, and the changeset.

### Sub-libraries affected

- [x] posthog-js (web)

### DIFF
--- a/.changeset/before-send-token.md
+++ b/.changeset/before-send-token.md
@@ -1,5 +1,6 @@
 ---
-"posthog-js": patch
+'@posthog/core': patch
+'posthog-js': patch
 ---
 
-Re-assert the `token` property if a `before_send` hook removes it, so events are no longer silently rejected by ingest with a 401.
+Drop the event and log a warning when a `before_send` hook removes the `token` property, instead of silently sending an event that ingest rejects with a 401.
--- a/.changeset/before-send-token.md
+++ b/.changeset/before-send-token.md
@@ -1,0 +1,5 @@
+---
+"posthog-js": patch
+---
+
+Re-assert the `token` property if a `before_send` hook removes it, so events are no longer silently rejected by ingest with a 401.
--- a/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
@@ -174,4 +174,31 @@ describe('posthog core - before send', () => {
             `Event '${randomUnsafeEditableEvent}' was rejected in beforeSend function. This can cause unexpected behavior.`
         )
     })
+
+    it('re-asserts the token property if a beforeSend function strips it', () => {
+        // Regression test for #3438: the project api_key is sent on
+        // properties.token, so a generic token/PII scrubber that drops it would
+        // otherwise make every event 401 with "submitted without an api_key".
+        const posthog = posthogWith({
+            before_send: (cr) => {
+                if (cr.properties) {
+                    for (const key of Object.keys(cr.properties)) {
+                        if (/token/i.test(key)) {
+                            delete cr.properties[key]
+                        }
+                    }
+                }
+                return cr
+            },
+        })
+        ;(posthog._send_request as jest.Mock).mockClear()
+
+        const capturedData = posthog.capture(eventName, {}, {})
+
+        expect(capturedData).toHaveProperty(['properties', 'token'], 'testtoken')
+        expect(posthog._send_request).toHaveBeenCalled()
+        expect(mockLogger.warn).toHaveBeenCalledWith(
+            `Event '${eventName}' is missing its 'token' property after beforeSend (it may have been removed by a token or PII scrubber); re-adding it so the event can be ingested.`
+        )
+    })
 })
--- a/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
@@ -175,10 +175,11 @@ describe('posthog core - before send', () => {
         )
     })
 
-    it('re-asserts the token property if a beforeSend function strips it', () => {
+    it('drops the event if a beforeSend function strips a required property like token', () => {
         // Regression test for #3438: the project api_key is sent on
         // properties.token, so a generic token/PII scrubber that drops it would
         // otherwise make every event 401 with "submitted without an api_key".
+        // We drop the event and warn rather than send something ingest will reject.
         const posthog = posthogWith({
             before_send: (cr) => {
                 if (cr.properties) {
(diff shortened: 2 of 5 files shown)


marandaneto:
@PostHog/team-client-libraries i think we should drop the event in this case, and log to the console, but we should not restore the token (what if it is intentional? prob not tho)
we have a `isKnownUnsafeEditableEvent` list for events, but not for properties, maybe something similar for properties would make sense

marandaneto:
thanks @archievi 
lemme check with the team

marandaneto:
@archievi lets do something similar to what i mentioned above
add a isKnownUnsafeEditableEventProperty with `token` as the item of this list for now
then we log out but do not restore the token

archievi:
Appreciate the feedback @marandaneto  - updated to match your suggestion:

- Added `knownUnsafeEditableEventProperty` (`['token']` for now) and an `isKnownUnsafeEditableEventProperty` predicate in `@posthog/core`, mirroring the existing `knownUnsafeEditableEvent` / `isKnownUnsafeEditableEvent` pair.
- `_runBeforeSend` no longer restores the token. Instead it snapshots the required properties present before the hook chain runs, and if a hook removes one, it logs a warning and drops the event (returns `null`).

One implementation detail worth flagging: the snapshot is taken before the hooks run because a hook can mutate `data.properties` in place (the reported scrubber does `delete cr.properties[key]`), so comparing against the original object afterwards wouldn't detect the removal.

I scoped the drop to the case where `properties` still exists but a required key was removed, so the existing "no properties after beforeSend" warn-but-send path is unchanged. Let me know if you'd rather that path drop too.

Regression test updated to assert the event is dropped, and changeset reworded.

marandaneto:
thanks, i pushed a few changes and signed commits

marandaneto (approved):

PR #3759, opened 2026-06-08
feat(react-native-plugin): port posthog-react-native-plugin into the monorepo

ioannisj:
## Problem

`posthog-react-native-plugin` is the successor to `posthog-react-native-session-replay`, adding native error tracking (iOS/Android crash autocapture) alongside session replay. It was developed in its own repo and now needs to live in the `posthog-js` monorepo so it ships through the same release pipeline as the other PostHog JS packages.

## Changes

Ports the package from [`PostHog/posthog-react-native-plugin#1`](https://github.com/PostHog/posthog-react-native-plugin/pull/1) into `packages/react-native-plugin/` and wires it into the monorepo:

- Imports the package and drops standalone-repo scaffolding (own `.github/`, `.changeset/`, workspace/lock files, etc.)
- Aligns `package.json`, scripts, `tsconfig`, eslint/prettier, and dev-deps (`catalog:`) with monorepo conventions
- Adds the package to the release matrix + an initial-release changeset (`0.0.0` → `1.0.0`)
- Moves the example app to `examples/example-rn-native-plugin/` (linked to the in-repo plugin)
- Adds a native CI workflow that builds the example's Android + iOS (CocoaPods and SPM) code, gated to the plugin

The public API that consumes this package lands separately in [#3677](https://github.com/PostHog/posthog-js/pull/3677).

I left some review comments in this [PR](https://github.com/PostHog/posthog-react-native-plugin/pull/1) since it should be easier to review since it's probably a smaller diff. I'll address any feedback directly here though

### Libraries affected

- [x] posthog-react-native-plugin (new package)

## Checklist

- [x] Accounted for backwards compatibility of any changes (no breaking changes\!)


packages/react-native-plugin/src/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  enabled?: boolean
+  sdkReplayConfig?: PostHogReactNativePluginMap
+  decideReplayConfig?: PostHogReactNativePluginMap
+}
+
+export interface PostHogReactNativePluginErrorTrackingConfig {
+  nativeAutocapture?: boolean
+}
+
+export interface PostHogReactNativePluginConfig {
```
ioannisj, line 32: New nested config shape to carry sessionReplay and errorTracking keys

packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+        let host = sdkOptions["host"] as? String ?? PostHogConfig.defaultHost
+        let debug = sdkOptions["debug"] as? Bool ?? false
+
+        PostHogSessionManager.shared.setSessionId(sessionId)
+
+        let config = PostHogConfig(projectToken: projectToken, host: host)
+        config.captureApplicationLifecycleEvents = false
+        config.captureScreenViews = false
+        config.debug = debug
+        config.errorTrackingConfig.autoCapture = nativeErrorTrackingAutocapture
```
ioannisj, line 79: This is the new line that matters most. This is marked as a new file because of the rename

packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+            sdkOptions: sdkOptions,
+            sessionReplayEnabled: true,
+            sdkReplayConfig: sdkReplayConfig,
+            decideReplayConfig: decideReplayConfig,
+            nativeErrorTrackingAutocapture: false,
+            resolve: resolve
+        )
+    }
+
+    private func setupNativeSdk(
```
ioannisj, line 50: shared setup helper for both setup() and start()

packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
```diff
@@ -0,0 +1,217 @@
...
+        config.captureApplicationLifecycleEvents = false
+        config.captureScreenViews = false
+        config.debug = debug
+        config.errorTrackingConfig.autoCapture = nativeErrorTrackingAutocapture
+
+        if #available(iOS 15.0, *) {
+            config.surveys = false
+        }
+
+        if sessionReplayEnabled {
```
ioannisj, line 85: This is now conditional based on config. Previously this was not gated cause call to start() implied replay = enabled

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -0,0 +1,301 @@
...
+import com.posthog.PostHogConfig
+import com.posthog.android.PostHogAndroid
+import com.posthog.android.PostHogAndroidConfig
+import com.posthog.internal.PostHogPreferences
+import com.posthog.internal.PostHogPreferences.Companion.ANONYMOUS_ID
+import com.posthog.internal.PostHogPreferences.Companion.DISTINCT_ID
+import com.posthog.internal.PostHogSessionManager
+import java.util.UUID
+
+class PosthogReactNativePluginModule(
```
ioannisj, line 20: mirror-image changes of iOS PosthogReactNativePlugin.swift

packages/react-native-plugin/src/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  sessionReplay?: PostHogReactNativePluginSessionReplayConfig
+  errorTracking?: PostHogReactNativePluginErrorTrackingConfig
+}
+
+export function setup(
```
ioannisj, line 37: New primary entrypoint that replaces the old flat 4-arg start()

packages/react-native-plugin/src/index.tsx
```diff
@@ -0,0 +1,119 @@
...
+  sessionId: string,
+  sdkOptions: PostHogReactNativePluginMap,
+  pluginConfig: PostHogReactNativePluginConfig = {}
+): Promise<void> {
+  return PosthogReactNativePlugin.setup(sessionId, sdkOptions, pluginConfig)
+}
+
+export function start(
```
ioannisj, line 45: start() was kept as legacy/back-compat but now that I'm writing this maybe there's no reason to do so and I'm being overly defensive here. An older posthog-react-native version will never land here (since we explicitly require the old package name). Could be removed and simplified I think

ioannisj (commented):
Dropping some comments for the error tracking related changes that are on-top of old `posthog-react-native-session-replay` plugin

You can focus on src/index.tsx, ios/PosthogReactNativePlugin.swift and android/PosthogReactNativePluginModule.kt. 

The feature change is one line on each native side: errorTrackingConfig.autoCapture = nativeErrorTrackingAutocapture

Also, I was probably being too defensive at keeping the old entrypoint start() in this package as well. Was thinking of back compat and matching the public APIs, but there's prob no reason to

The other changes is this PR is: 
- renames from posthog-react-native-session-replay to posthog-react-native-plugin and moving things around 
- Changes in repo scaffolding so that the plugin is now hosted in this monorepo
- Adding the new package in the release flow of this repo

.changeset/posthog-react-native-plugin-initial-release.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native-plugin': major
```
marandaneto, line 2: current version if 0.0.0 so this means 1.0.0
do you want to start this with 1.0.0 or eg 0.1.0?
the current plugin is 1.6.0
whats about this one be 2.0.0 which has a package rename?

.github/workflows/react-native-plugin-native-ci.yml
```diff
@@ -0,0 +1,211 @@
...
+
+on:
+  pull_request:
+    paths:
+      - 'packages/react-native-plugin/**'
+      - 'examples/example-rn-native-plugin/**'
+      - '.github/workflows/react-native-plugin-native-ci.yml'
+  merge_group:
+    types:
+      - checks_requested
```
marandaneto: what does that mean?

packages/react-native-plugin/src/index.tsx
marandaneto: > An older posthog-react-native version will never land here (since we explicitly require the old package name)

all packages are optionals right, so its possible i guess?
so either we also bump a major in the rn SDK or the rn SDK has to work with the older and newer plugin

marandaneto (approved):
left comments otherwise LGTM, approving to unblock

.changeset/posthog-react-native-plugin-initial-release.md
ioannisj: yeah agreed, 2.0.0 sounds more sensible

packages/react-native-plugin/src/index.tsx
ioannisj: My thinking is that old RN sdk imports `posthog-react-native-session-replay` explicitly. So even if optional `posthog-react-native-plugin` is there, it will not be picked up and used. Only newer versions of rn SDK will know to import and use `posthog-react-native-plugin`

However, based on the comment below, if we are treating this as v2.0.0, I will keep just to keep the contract compatible

.github/workflows/react-native-plugin-native-ci.yml
ioannisj: No other workflow uses this, will remove

PR #3760, opened 2026-06-08
fix(conversations): re-attach support widget after SPA body replacement

benben:
## Problem

The Conversations (Support) widget mounts `#ph-conversations-widget-container` into `document.body` **once** and is never recreated. SPA frameworks that replace the entire `document.body` on navigation — notably **Turbo Drive / Rails Turbo** — detach the container without any teardown call, so the bubble silently disappears until a full page reload.

Surveys already self-heal on a 1s interval; Conversations did not.

## Fix

Adds a lightweight 1s watcher (`_startReattachWatcher`) that re-attaches the **same** container element when it detects the node has been detached (`!isConnected`). Re-attaching the existing element (rather than re-rendering) preserves widget state and avoids refetching messages. The watcher is stopped in `hide()` and `destroy()` so intentional removals stay removed.

## Tests

Two new unit tests in `conversations-manager.test.tsx`:
- re-attaches the widget when its container is detached from the DOM
- does **not** re-attach after `hide()`

All 51 conversations-manager tests pass; ESLint + Prettier clean.


veryayskiy (approved):
🚢

PR #3764, opened 2026-06-08
fix(nextjs-config): strip dangling sourceMappingURL comments from Turbopack chunks

### LINKED ISSUE #3058: Cannot deploy sourcemaps using `deleteAfterUpload` with Next.js and TurboPack
## Bug description

When configuring Next.js with PostHog with `sourcemaps.enabled = true`, `deleteAfterUpload: true` and Next's `productionBrowserSourceMaps = false`, `withPostHogConfig` from `@posthog/nextjs-config` forces the value `productionBrowserSourceMaps = true`:

https://github.com/PostHog/posthog-js/blob/8d9bb16ea664fbd83529715edf1e330d45832e2d/packages/nextjs-config/src/config.ts#L30-L32

This results in:

✅ SourceMaps uploaded to PostHog when building
✅ SourceMaps are deleted from production deployment
🚫 SourceMaps are still referenced in production chunks, resulting in `404` errors

> [!NOTE]
> _I see a [forum post](https://posthog.com/questions/next-js-sourcemaps) asking about this.. I tried posting replies, but had issues with the forum so posting here._ 

## How to reproduce

1. Configure NextJS with:

```ts
const baseConfig: NextConfig = {
  // Not sure if this is a relevant detail, but I output "standalone" in my deployments
  output: "standalone",
  productionBrowserSourceMaps: false,
  // ...
};

const configWithPosthog = withPostHogConfig(baseConfig, {
  // ...
  sourcemaps: {
    // ...
    enabled: true,
    deleteAfterUpload: true,
  },
});

export default configWithPosthog;
```

2. Deploy
3. Look at compiled output
    - Still contains `//# sourceMappingURL=` references

## Notes

- If there is a good reason this should not be the case, there should be a comment in the code explaining why TurboPack is forced to use this

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

emmayusufu:
## Problem

Fixes #3058. With `@posthog/nextjs-config` on Next.js + Turbopack and `sourcemaps.enabled` + `deleteAfterUpload: true`:

1. `withPostHogConfig` forces `productionBrowserSourceMaps = true` so Turbopack emits browser source maps.
2. The CLI uploads those maps and then deletes them (`--delete-after`).
3. The emitted chunks still carry a `//# sourceMappingURL=<file>.js.map` comment pointing at the now-deleted map, which 404s in production.

The webpack path avoids this with `SourceMapDevToolPlugin({ append: false })` and the rollup path with `sourcemap: 'hidden'`, but Turbopack's `productionBrowserSourceMaps` has no hidden mode, it always writes the comment.

## Fix

After `processSourceMaps` runs (in the Turbopack `runAfterProductionCompile` hook), when `deleteAfterUpload` is set, strip the dangling comment from the browser chunks under `<distDir>/static`. A comment is only removed when the `.map` it references no longer exists, so a still-present map is never touched. Inline (`data:`) maps and server chunks are left alone.

## Verification

- `pnpm --filter @posthog/nextjs-config test:unit` — 8 unit tests (strips dangling, keeps live, inline `data:` untouched, server chunks untouched, CRLF, `.mjs`/`.cjs`, no-op on missing dir).
- Tested against a real `next build --turbopack` (Next 15.5, `productionBrowserSourceMaps: true`): with maps present the helper changed nothing; after deleting the maps it stripped all 13 dangling comments with zero residual references, and the chunks were otherwise byte-for-byte identical.

One note: I exercised the delete step by removing the `.map` files directly (a real `posthog-cli` upload needs project credentials), which matches what `--delete-after` does. Happy to move the strip into `@posthog/cli` instead if you'd prefer it live next to the deletion.

### DIFF
--- a/.changeset/nextjs-turbopack-strip-sourcemap-comments.md
+++ b/.changeset/nextjs-turbopack-strip-sourcemap-comments.md
@@ -1,0 +1,5 @@
+---
+'@posthog/nextjs-config': patch
+---
+
+Strip dangling `//# sourceMappingURL=` comments from Turbopack browser chunks when `deleteAfterUpload` is set, so deleted source maps are no longer referenced (and 404) in production.
--- a/packages/nextjs-config/rslib.config.mjs
+++ b/packages/nextjs-config/rslib.config.mjs
@@ -6,4 +6,10 @@ export default defineConfig({
   shims: { esm: { __dirname: true } },
   syntax: 'es6',
   lib: [{ format: 'esm' }, { format: 'cjs' }],
+  source: {
+    entry: {
+      index: ['src/**/*', '!src/**/*.spec.ts'],
+    },
+    tsconfigPath: './tsconfig.build.json',
+  },
 })
--- a/packages/nextjs-config/src/config.ts
+++ b/packages/nextjs-config/src/config.ts
@@ -1,6 +1,7 @@
 import type { NextConfig } from 'next'
 import { PosthogWebpackPlugin, PluginConfig, resolveConfig, ResolvedPluginConfig } from '@posthog/webpack-plugin'
 import { hasCompilerHook, isTurbopackEnabled, processSourceMaps } from './utils'
+import { stripDanglingSourceMapComments } from './strip-sourcemap-comments'
 
 type NextFuncConfig = (phase: string, { defaultConfig }: { defaultConfig: NextConfig }) => NextConfig
 type NextAsyncConfig = (phase: string, { defaultConfig }: { defaultConfig: NextConfig }) => Promise<NextConfig>
@@ -134,6 +135,9 @@ function withCompilerConfig(
       await userCompilerHook?.(config)
       console.debug('Processing source maps from compilation hook...')
       await processSourceMaps(posthogConfig, config.distDir)
+      if (posthogConfig.sourcemaps.deleteAfterUpload) {
+        await stripDanglingSourceMapComments(config.distDir)
+      }
     }
     return newConfig
   }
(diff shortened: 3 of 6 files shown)


emmayusufu:
Thanks for the review. Addressed both:

1. Remote URLs: `http(s)` and protocol-relative `//` sourceMappingURL comments are now guarded the same way as inline `data:` maps, so a still-valid remote map is never stripped. Added a test for it.
2. Parameterized the `.mjs`/`.cjs` extension test with `it.each`. I left the dangling/kept/inline/CRLF cases as separate tests since they assert different behaviours rather than the same logic over different inputs, collapsing those felt like it would cost more clarity than it saves, but happy to fold them in if you'd prefer.

Unit suite is at 10 tests, build and lint pass.

marandaneto:
cc @PostHog/team-error-tracking

ablaszkiewicz:
Hey @emmayusufu. Thanks for the contribution. I've tested your change locally on our `example-nextjs` project. Forced it to use turbo using `pnpm build:turbo`. Confirmed it solves the described issue and confirmed that everything is working as expected with the new package

ablaszkiewicz (approved):

ablaszkiewicz:
Had to rebase to re-sign commits as our CI won't let any unsigned commits be merged

emmayusufu:
Hey, this merged a while back but I never got a merch code for it. Is there something I need to fill in, or did it just slip through? Can send shipping details wherever they need to go.

PR #3767, opened 2026-06-08
fix(replay): jump scrolls instantly on seek with scroll-behavior smooth

arnohillen:
## Problem

Silk-style bottom sheets reveal content by scrolling an internal track with `scroll-behavior: smooth`. Merged #3736 re-applies scroll offsets after fast-forward flush when they clamp to 0 before layout settles, but sync scroll application still used `behavior: 'auto'`, which **inherits** the page's CSS `scroll-behavior: smooth`.

On seek/scrub (and skip-inactivity jumps), the replayer animates the reveal scroll from 0 instead of jumping to the recorded offset. The backdrop scrim renders while sheet content stays off-screen — the empty black rectangle customers report on Owner/vaquerorestaurant.com-style `?item=` menu flows.

## Changes

- Use `behavior: 'instant'` instead of `'auto'` for sync (`isSync`) scrolls in `applyScroll` — all three scroll targets (iframe window, nested document, element)
- Add regression test fixture mimicking a Silk sheet with `scroll-behavior: smooth` and assert seek lands at the recorded offset immediately

## Scope

- **Replayer-only** fix; complements #3736
- Does **not** include #3742 capture-side `scrollend` changes (those help future recordings where scroll clamps to 0 at record time; existing exports with full scroll animation sequences are unaffected)

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/rrweb

## Checklist

- [x] Tests for new code
- [x] Ran `pnpm changeset`

## Test plan

- [x] `vitest run test/replayer.test.ts -t "lands a scroll instantly"`
- [x] `vitest run test/replayer.test.ts -t "re-applies a scroll"` (#3736 regression)


arnohillen:
## Scope note (from export analysis)

This fix addresses **seek/skip-inactivity** only: sync scrolls with `behavior: 'auto'` inherit `scroll-behavior: smooth` and animate from 0 instead of jumping.

It does **not** fix black sheets when playing linearly from 0 if the recording never captured the reveal scroll. Analyzed export `019e89dc-2580-76f8-bdbd-2886899d51ae`:

- **15.6s** (first item open): only `y=0` + backdrop `y=300`, no track reveal scroll → broken on linear playback regardless of this PR
- **423s** (taco-al-pastor): same — no track scroll captured
- **183s / 451s**: full reveal captured (peak ~900+) → this PR helps on seek; linear playback should also work if capture is good

That capture gap is what #3742 (`scrollend` resting offset) targets. Recommend merging both, with clear separation:
- **#3767** = replayer seek behavior
- **#3742** = recorder capture for clamped-to-0 reveals

TueHaulund (approved):

PR #3779, opened 2026-06-09
chore(react-native-plugin): scope under @posthog and revert premature release

ioannisj:
## Problem

A premature release (`31b8693c`) bumped `posthog-react-native-plugin` to `2.0.0` and consumed its changeset, but **published nothing** to npm — the package had never been set up for publishing. Separately, the package is moving to the `@posthog` npm scope and renamed to drop the now-redundant prefix.

## Changes

1. **Revert the bad release** (`31b8693c`) — restores the changeset, drops `package.json` back to `1.6.0`, and removes the premature `2.0.0` CHANGELOG entry.
2. **Scope + rename to `@posthog/react-native-plugin`** — npm name, changeset, `release.yml` publish matrix, native-CI `is-affected` filter, example dependency + imports, README, and the podspec git-source tag. Because it's scoped under `@posthog`, the redundant `posthog-` prefix is dropped: **`@posthog/react-native-plugin`** (not `@posthog/posthog-react-native-plugin`).

The iOS pod name (`posthog-react-native-plugin`) and the native module are intentionally left unscoped — CocoaPods names can't contain `@`/`/`, and React Native autolinking resolves scoped npm packages fine.

On merge, the restored changeset re-triggers the release as `@posthog/react-native-plugin@2.0.0`.

Follows up on #3759. Reverts `31b8693c`.

### Libraries affected

## Checklist

- [x] Changeset present (`@posthog/react-native-plugin`, major)
- [x] Accounted for cross-platform impact (iOS pod / Android module names unchanged)


packages/react-native-plugin/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "posthog-react-native-plugin",
-  "version": "2.0.0",
+  "name": "@posthog/posthog-react-native-plugin",
```
dustinbyrne: Since we're already renaming, should we consider `@posthog/react-native-plugin` for less redundancy + following the other packages in this repo?

It also matches the name of the package in the file system

dustinbyrne (approved):

packages/react-native-plugin/package.json
ioannisj: Renamed to `@posthog/react-native-plugin`

PR #3780, opened 2026-06-09
fix: resample replay decisions when sample rate changes

### LINKED ISSUE #3778: Trigger matching for session recordings works unexpectedly
## Bug description

In [how-to-control-which-sessions-you-record](https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record) it explicitly states that [combining-controls](https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#combining-controls) with [with-all-matching](https://posthog.com/docs/session-replay/how-to-control-which-sessions-you-record#with-all-matching) should take into account all `Recording conditions` with `Sampling Rate`.
But, in my case where we have configured `Trigger matching = all`, `URL matches`, `Feature flag` and `Sampling Rate = 0` I still see sessions being recorded

## How to reproduce
- Trigger matching -> Start when `all` triggers below match
- Some `URL matches`
- `Feature flag`
- `Sampling Rate` set to `0`
- SDK version `1.369.1`, web, react
- Zero `postHog.startSessionRecording()` calls from code, everything configured only in PostHog dashboard

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

After some investigation and playing around with PostHog AI, I find out SQL query for session recording.
I've run:
```
SELECT
    properties.$has_recording AS has_recording,
    properties.$recording_status AS recording_status,
    properties.$session_recording_start_reason AS start_reason,
    properties.$sdk_debug_recording_script_not_loaded AS script_blocked,
    properties.$sdk_debug_replay_url_trigger_status AS url_trigger,
    properties.$sdk_debug_replay_event_trigger_status AS event_trigger,
    properties.$sdk_debug_replay_linked_flag_trigger_status AS flag_trigger,
    properties.$replay_sample_rate AS sample_rate,
    properties.$lib_version AS sdk_version
FROM events
WHERE $session_id = '<your-session-id>'
ORDER BY timestamp DESC
LIMIT 1
```
Here is some results it gave me:
```
// No.1:
[
    {
        "has_recording": null,
        "recording_status": "lazy_loading",
        "start_reason": null,
        "script_blocked": null,
        "url_trigger": null,
        "event_trigger": null,
        "flag_trigger": null,
        "sample_rate": null,
        "sdk_version": "1.369.1"
    }
]

---

// No.2:
[
    {
        "has_recording": null,
        "recording_status": "sampled",
        "start_reason": "linked_flag_matched",
        "script_blocked": null,
        "url_trigger": "trigger_activated",
        "event_trigger": "trigger_disabled",
        "flag_trigger": "trigger_activated",
        "sample_rate": null,
        "sdk_version": "1.369.1"
    }
]

---

// No.3:
[
    {
        "has_recording": null,
        "recording_status": "sampled",
        "start_reason": "linked_flag_matched",
        "script_blocked": null,
        "url_trigger": "trigger_activated",
        "event_trigger": "trigger_disabled",
        "flag_trigger": "trigger_activated",
        "sample_rate": null,
        "sdk_version": "1.369.1"
    }
]
```

Those behaviour (especially No.2 and No.3) suggest to me that recording still happens no matter what sampling rate you have if you have `"start_reason": "linked_flag_matched"`

Also not that If I intentionally want to have `Sample Rate = 0` to record sessions if I can simply disable recordings then. Issue is observed even if you select `Sample Rate = 1..5%`, where basically I do expect to have really small amount of recordings but they are still recorded like `Sample Rate = 100%`

dustinbyrne:
## Problem

Session replay sampling decisions are persisted for the current PostHog session. When a session was sampled in at one replay sample rate and the project config later changed to a lower rate, the SDK could reuse the stale sampled-in decision after a reload/re-instantiation within the same session.

This could cause replay capture to continue even when the configured sample rate should now sample that session out.

Closes #3778

## Changes

- Persist the replay sample rate alongside V1 session replay sampling decisions.
- Recompute replay sampling when the stored sample rate differs from the current configured rate.
- Treat legacy sampled-in decisions without a stored rate as needing a fresh decision, so lowering the sample rate to `0` takes effect after reload.
- Store V2 trigger-group sampling decisions as `{ sessionId, sampleRate, sampled }` and recompute them when the group sample rate changes.
- Preserve explicit sampling overrides when the configured sample rate is `0`.
- Expose the new replay sample-rate persistence key through the persistence key policy.
- Add regression coverage for V1, V2 trigger groups, and explicit overrides.

Validated with an MCTS-style review. The reviewed scope is stale sampled-in decisions after replay sample-rate changes; live reconfiguration of an already-started recorder remains existing/out-of-scope behavior.

Tests run:

```bash
git diff --check
pnpm --dir packages/browser exec jest src/__tests__/extensions/replay/lazy-sessionrecording.test.ts --runInBand
pnpm --dir packages/browser exec jest src/__tests__/persistence-key-policy.test.ts --runInBand
```

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] posthog-react-native-plugin
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


dustinbyrne:
Addressed Greptile feedback in dc936c820: added parameterized V1 sampling persistence coverage including the legacy sampled-in-without-stored-rate path, stored-rate re-sampling, unchanged-rate reuse, and explicit override at sampleRate 0.

dustinbyrne:
Fixed the Playwright regression in cb5b11c01: explicit sampling overrides from configs without a sample rate now persist as the existing null-rate override sentinel instead of being treated as legacy missing-rate state. Local validation now includes the full replay unit test file plus the failing Playwright sampling specs on Chromium and Firefox.

marandaneto:
makes sense, @dustinbyrne worth checking if we do the same with the mobile SDKs

marandaneto (approved):

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1494,6 +1495,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         this._instance.persistence?.register({
             // short-circuits the `makeSamplingDecision` function in the session recording module
             [SESSION_RECORDING_IS_SAMPLED]: this.sessionId,
+            [SESSION_RECORDING_SAMPLE_RATE]: this._sampleRate,
```
TueHaulund: This stores `this._sampleRate`, which in the common path (override applied after remote config has loaded) is a number, not the `null` sentinel — so an override at rate 0 stores rate `0`, and any later rate change triggers `sampleRateChanged` → re-decision → the "definitely record this session" override gets reverted mid-session. Before this PR, overrides survived all rate changes; now only the narrow override-before-config path (which writes the sentinel) does. Suggest writing the `null` sentinel unconditionally here.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1815,6 +1817,7 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
 
         return {
             $recording_status: this.status,
+            [SESSION_RECORDING_SAMPLE_RATE]: this._sampleRate,
```
TueHaulund: `$replay_sample_rate` is now emitted from two conflicting sources: this one (current config value) and the new `exposure: 'event'` policy entry (persisted decision-time value). Due to merge order this one always shadows the policy entry while the recorder is loaded, making it dead code. Also, since `sdkDebugProperties` attaches to every event, all replay-enabled projects — including ones not using sampling — now get `$replay_sample_rate: null` on every event. Suggest keeping exactly one source.

packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -238,9 +241,19 @@ export class V1RecordingStrategy implements RecordingStrategy {
...
+        // 3. Sample rate changed since the stored decision was made
+        //
+        // Older SDK versions only stored the session ID for sampled-in sessions, not the rate. Treat
+        // those sampled-in decisions as needing a fresh decision so lowering sampleRate to 0 takes
+        // effect after a page reload within the same PostHog session.
+        // An explicit override from a config without sampling stores null as the rate, so only an
+        // undefined rate is considered legacy missing-rate state.
         const sessionChanged = typeof storedValue === 'string' && storedValue !== sessionId
-        const makeDecision = sessionChanged || !isBoolean(storedIsSampled)
+        const sampleRateChanged = isNumber(storedSampleRate) && storedSampleRate !== currentSampleRate
```
TueHaulund, line 253: The legacy migration is asymmetric: legacy sampled-in decisions without a stored rate are re-decided, but a legacy bare `false` (sampled-out) is still reused even when the rate was raised — same in the V2 legacy branch. Making the legacy branch symmetric is a one-liner.

FYI: I have an in-flight fix for the related bug where a sampled-out decision is inherited across sessions (it's behind a recent support escalation) — I'll stack it on top of this PR. Your `{sessionId, sampleRate, sampled}` object format subsumes the session-tagging I was doing for V2, so I'll converge on that.

TueHaulund (commented):
Found two real issues plus a legacy-migration asymmetry — details inline. The V2 object format is a nice improvement, and mixed-version tabs self-heal since `sampleOnProperty` is deterministic, so no compat concerns there.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
dustinbyrne: Addressed in 5f87eaa07300687a0bb24b4477a5dd3789de782e

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
dustinbyrne: Addressed in 5f87eaa07300687a0bb24b4477a5dd3789de782e

packages/browser/src/extensions/replay/external/recording-strategies.ts
dustinbyrne: Addressed in 5f87eaa07300687a0bb24b4477a5dd3789de782e

PR #3781, opened 2026-06-09
feat(mcp): add server-agnostic PostHog MCP Analytics capture API

lucasheriques:
## Summary

Adds **`createMcpAnalytics(posthog, options?)`** — a server-agnostic capture API for MCP servers that have no `@modelcontextprotocol/sdk` `Server`/`McpServer` instance to wrap (e.g. custom hono/HTTP dispatchers, edge/serverless MCP servers).

`instrument()` works by monkeypatching a server's request handlers. A growing class of MCP servers — Cloudflare Workers, hono, Vercel edge, multiplexed dispatchers — don't expose such an object, so they currently can't use the SDK at all and fall back to hand-rolled `posthog.capture()`. PostHog's own dotcom MCP server is one of these after its hono refactor.

`createMcpAnalytics` gives them a first-class entry point: the host resolves identity/context per request and calls the capture methods directly. Each call runs through the **exact same pipeline** as `instrument()` (sanitize → truncate → `$exception` fan-out → `beforeSend`) and emits the same canonical `$mcp_*` events.

```ts
const analytics = createMcpAnalytics(posthog, { beforeSend })

await analytics.captureToolCall({
  toolName: 'execute-sql',
  parameters, response, durationMs, isError, error,
  distinctId,            // → distinct_id (+ person processing on)
  sessionId,             // → $session_id (omitted when absent)
  groups,                // → $groups
  properties,            // extra props, spread verbatim
})
await analytics.captureInitialize({ clientName, clientVersion, distinctId })
await analytics.captureToolsList({ toolNames, distinctId })
await analytics.capture({ event, distinctId, properties })   // custom
```

## Design notes

- **Shared pipeline, not a forked one.** Both `instrument()` and `createMcpAnalytics()` go through `McpEventSink` / `processMcpEvent`, so event shapes can't drift. `instrument`'s server-WeakMap/session/identity machinery is left untouched — the genuinely shared core is the pipeline, and rebuilding the published `instrument` path on top of the new API was unjustified regression risk for no behavior gain.
- **One backward-compatible behavior change:** `$session_id` is now omitted from events when no session is supplied (was always set). The auto-capture path always resolves a session id, so it's unaffected; this only matters for the new stateless path, where a fabricated `$session_id` would wrongly bucket events into a non-existent Session Replay session.

## Testing

- New `manual.test.ts` (14 cases): canonical props, identity/`$set`/`$groups`, `$exception` fan-out + suppression, anonymous/`$process_person_profile`, `$session_id` omission, `beforeSend` mutate/drop.
- Full package suite green: **302 tests**, lint clean, build emits `dist/extensions/manual.js` and the new exports.

## Release

Changeset included (`minor`).

---


pauldambra:
~~i am a dinosaur question  
why is this in our MCP and not in an SDK  
if someone else has a hand-rolled server they would have to pull this in _and_ our SDK or....?~~

ugh i saw packages/mcp late at night and made a silly

packages/mcp/src/extensions/manual.ts
```diff
@@ -0,0 +1,135 @@
...
+ *   toolName: "execute-sql",
+ *   durationMs: 42,
+ *   isError: false,
+ *   distinctId: "user-123",
+ *   groups: { organization: "org-1", project: "proj-1" },
+ *   properties: { $mcp_client_name: "claude-code" },
+ * })
+ * ```
+ */
+export function createMcpAnalytics(posthog: PostHog, options: CreateMcpAnalyticsOptions = {}): McpAnalyticsManual {
```
marandaneto: maybe a better approach is having:

`posthog = PostHog(options): PostHogBackendClient(options)` (a class that extends PostHog from node)
that you can then:
`posthog.captureToolCall(...)`

so you can add specific methods to PostHog and still inherit all methods from Node

if you look at all the other SDKs, its either this way or:

`posthog = PostHog.init|start(options)`
that you can `posthog.capture(...)`

this one kinda does something totally different, requires a prebuilt `PostHog` instance, returns yet another type `McpAnalyticsManual`

Node for example:
`PostHog extends PostHogBackendClient`
and `class PostHogBackendClient extends PostHogCoreStateless`

RN:
`class PostHog extends PostHogCore`
`class PostHogCore extends PostHogCoreStateless`

so the idea is the same

packages/mcp/README.md
```diff
@@ -61,6 +61,39 @@ Events for sessions with no resolved identity are sent with `$process_person_pro
 
 The full event + property catalog (including `$mcp_resources_*` / `$mcp_prompts_*`) lives in [`docs/ARCHITECTURE.md`](./docs/ARCHITECTURE.md).
 
+## No `Server`/`McpServer` to wrap? Use `createMcpAnalytics`
```
marandaneto: the readme is nice but i'd prefer all the documentation living in posthog.com/docs
this will be easily outdated, under-documented on posthog.com/docs
this is already a problem in all SDKs and i am slowly changing that, so lets not add one more
can you port everything to posthog.com/docs?
this readme should just point out to the docs

packages/mcp/src/extensions/manual.ts
pauldambra: yep, across all of these changes +1000 to keeping the cognitive surface low for our users

lucasheriques: done, went with `PostHogMCP extends PostHog` in fa51462c. it's a drop-in posthog-node client (capture/identify/flush/shutdown/flags all inherited) with captureToolCall + captureInitialize on top, so no third type to learn, and the client's own beforeSend/enableExceptionAutocapture apply. bonus: dotcom just makes getPostHogClient() return a PostHogMCP, so existing capture calls keep working, mcp code gets the new methods, no second client. good call.

packages/mcp/README.md
lucasheriques: done, readme just points to posthog.com/docs/mcp-analytics now and the prose all lives there (the custom-servers page covers PostHogMCP). agree on not spinning up another place to drift.

packages/mcp/src/extensions/manual.ts
```diff
@@ -0,0 +1,140 @@
...
+      event.listedToolNames = data.toolNames
+      event.parameters = data.parameters
+      event.duration = data.durationMs
+      return emit(event)
+    },
+
+    async capture(data: ManualCustomCaptureData): Promise<void> {
+      if (!data || typeof data.event !== 'string' || data.event.length === 0) {
+        throw new Error('`capture` requires an `event` name, e.g. analytics.capture({ event: "feedback_submitted" })')
+      }
```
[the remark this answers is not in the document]
lucasheriques: moot after the rework. the custom capture() override is gone, custom events use the inherited posthog-node capture(), so there's nothing left to reject.

packages/mcp/src/__tests__/manual.test.ts
```diff
@@ -0,0 +1,218 @@
...
+      expect(onlyCapture(PostHogMCPAnalyticsEvent.ToolCall).properties.redacted).toBe(true)
+    })
+
+    it('can drop a payload by returning null', async () => {
+      const analytics = createMcpAnalytics(fakePostHog(), { beforeSend: () => null })
+
+      await analytics.captureToolCall({ toolName: 'execute-sql', distinctId: 'user-123', isError: false })
+
+      expect(capture.findCapturesByEvent(PostHogMCPAnalyticsEvent.ToolCall)).toHaveLength(0)
+    })
```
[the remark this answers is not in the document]
lucasheriques: rewrote the test file, so the beforeSend pair this referenced is gone. keeping the error cases as separate it() blocks since they differ by assertion, not just input, so it.each would just push the difference into per-row callbacks. matches the rest of the tests here.

packages/mcp/README.md
```diff
@@ -14,226 +12,24 @@ npm install @posthog/mcp posthog-node
...
-})
-```
+Full documentation lives at **[posthog.com/docs/mcp-analytics](https://posthog.com/docs/mcp-analytics)** — installation, custom servers, capturing intent, identifying users, custom events, privacy/redaction, the event + property reference, and sample queries. The design narrative lives in [`docs/ARCHITECTURE.md`](./docs/ARCHITECTURE.md).
 
-## Docs
+## Building in public
 
-- [`docs/ARCHITECTURE.md`](./docs/ARCHITECTURE.md): wire-up, request lifecycle, event pipeline, full event + property catalog, intent resolution, file map, HogQL query recipes.
-- [PostHog docs](https://posthog.com/docs): the rest.
+The SDK source lives in the [`posthog-js` monorepo](https://github.com/PostHog/posthog-js/tree/main/packages/mcp). Issues, PRs, and feedback are welcome. We started from a duplicated copy of the MIT-licensed [MCPcat TypeScript SDK](https://github.com/MCPCat/mcpcat-typescript-sdk) — the event schema, identity model, and feature surface have since diverged.
```
marandaneto, line 35: if thats the case, you need to apend their license here https://github.com/PostHog/posthog-js/blob/main/LICENSE
and all files that are derived from their SDK should have a license header eg https://github.com/PostHog/posthog-js/blob/f601c496338ed0be8853f94160ee3edca542ac7d/packages/node/src/extensions/error-tracking/autocapture.ts#L1-L3
matching exactly the company name and license from mcpcat repo

lucasheriques: done in d644c638. added their mit license + copyright to the repo LICENSE (matching the sentry/expo blocks already there), and put the derived-from header on the source files that came from the original mcpcat import. left it off posthog-mcp.ts since that's new code, and off the tests since i couldn't confirm they were copied from mcpcat vs written here. should i stamp the tests too?

marandaneto (approved):
much better imo and still under 1.0.0 so we can still improve things later, approving to unblock

PR #3791, opened 2026-06-10
chore: deprecate preview request configs

marandaneto:
## Problem

The browser SDK has two request-related config options that were still marked as preview: `__preview_disable_beacon` and `__preview_disable_xhr_credentials`. `disable_beacon` should now have a stable config name, while the XHR credentials option should be retired because the SDK should no longer set `XMLHttpRequest.withCredentials`.

## Changes

- Added stable `disable_beacon` config option.
- Deprecated `__preview_disable_beacon` in favor of `disable_beacon`.
- Deprecated `__preview_disable_xhr_credentials` as a no-op.
- Removed setting `req.withCredentials = true` from XHR requests.
- Removed the proposed `disable_xhr_credentials` config option.
- Updated type snapshots and tests for config rename/XHR credentials behavior.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/__tests__/posthog-core-also.test.ts
+++ b/packages/browser/src/__tests__/posthog-core-also.test.ts
@@ -62,25 +62,19 @@ describe('posthog core', () => {
     })
 
     describe('configRenames()', () => {
-        it('maps deprecated preview beacon option to the stable config name', () => {
-            expect(
-                configRenames({
-                    __preview_disable_beacon: true,
-                })
-            ).toMatchObject({
-                disable_beacon: true,
-            })
-        })
-
-        it('prioritizes stable beacon option name over deprecated preview name', () => {
-            expect(
-                configRenames({
-                    disable_beacon: false,
-                    __preview_disable_beacon: true,
-                })
-            ).toMatchObject({
-                disable_beacon: false,
-            })
+        it.each([
+            [
+                'maps deprecated preview beacon option to the stable config name',
+                { __preview_disable_beacon: true },
+                { disable_beacon: true },
+            ],
+            [
+                'prioritizes stable beacon option name over deprecated preview name',
+                { disable_beacon: false, __preview_disable_beacon: true },
+                { disable_beacon: false },
+            ],
+        ] as [string, Partial<PostHogConfig>, Partial<PostHogConfig>][])('%s', (_description, input, expected) => {
+            expect(configRenames(input)).toMatchObject(expected)
         })
     })
 
--- a/packages/browser/src/__tests__/posthog-core-also.test.ts
+++ b/packages/browser/src/__tests__/posthog-core-also.test.ts
@@ -7,7 +7,7 @@ import { isUndefined } from '@posthog/core'
 import { ENABLE_PERSON_PROCESSING, SESSION_RECORDING_REMOTE_CONFIG, USER_STATE } from '../constants'
 import { createPosthogInstance, defaultPostHog } from './helpers/posthog-instance'
 import { PostHogConfig, RemoteConfig } from '../types'
-import { PostHog } from '../posthog-core'
+import { configRenames, PostHog } from '../posthog-core'
 import { PostHogPersistence } from '../posthog-persistence'
 import { SessionIdManager } from '../sessionid'
 import { RequestQueue } from '../request-queue'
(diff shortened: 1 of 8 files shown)


marandaneto:
added in posthog-js@1.265.1

packages/browser/src/posthog-core.ts
```diff
@@ -1055,8 +1057,13 @@ export class PostHog implements PostHogInterface {
             ...options.headers,
         }
         options.compression = options.compression === 'best-available' ? this.compression : options.compression
-        options.disableXHRCredentials = this.config.__preview_disable_xhr_credentials
-        if (this.config.__preview_disable_beacon) {
+        options.disableXHRCredentials = isUndefined(this.config.disable_xhr_credentials)
+            ? this.config.__preview_disable_xhr_credentials
+            : this.config.disable_xhr_credentials
+        const disableBeacon = isUndefined(this.config.disable_beacon)
```
marandaneto, line 1059: you could also disable that by passing `CaptureOptions.transport`  to capture but its not a global config

packages/types/src/posthog-config.ts
```diff
@@ -1728,13 +1728,23 @@ export interface PostHogConfig {
...
-     * Enabling this option may hurt the reliability of sending $pageleave events
+     * Enabling this option may hurt the reliability of sending $pageleave events.
      */
-    __preview_disable_beacon?: boolean
+    disable_beacon?: boolean
 
     /**
      * Disables sending credentials when using XHR requests.
      */
+    disable_xhr_credentials?: boolean
```
robbie-c: Hmm thinking about this, we should remove `disable_xhr_credentials` altogether and make this always-on behaviour.

(we don't want to do this with `disable_beacon`, that should remain false by default)

robbie-c: Just adding a link to the original PR so there's a bidirectional connection https://github.com/PostHog/posthog-js/pull/2318

marandaneto: ok so i should reprecate `__preview_disable_xhr_credentials`, make it a no op (so setting true/false does not do anything) and we never set `req.withCredentials = true`, correct?

marandaneto: [`7dbefb9` (this PR)](https://github.com/PostHog/posthog-js/pull/3791/changes/7dbefb90bf57603fd796e395497edb792eca0b8e)

packages/browser/src/__tests__/posthog-core-also.test.ts
```diff
@@ -61,6 +61,34 @@ describe('posthog core', () => {
...
+                    __preview_disable_beacon: true,
+                    disable_xhr_credentials: false,
+                    __preview_disable_xhr_credentials: true,
+                })
+            ).toMatchObject({
+                disable_beacon: false,
+                disable_xhr_credentials: false,
+            })
+        })
+    })
```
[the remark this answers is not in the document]
marandaneto: Addressed in 5d45045a — converted the config rename tests to an `it.each` table.

robbie-c (approved):

packages/browser/src/request.ts
```diff
@@ -207,11 +207,6 @@ const xhr = (options: RequestWithOptions) => {
     if (options.timeout) {
         req.timeout = options.timeout
     }
-    if (!options.disableXHRCredentials) {
-        // send the ph_optout cookie
-        // withCredentials cannot be modified until after calling .open on Android and Mobile Safari
-        req.withCredentials = true
-    }
```
ioannisj, line 214: I think this is worth mentioning in the changeset since it's not just preview flag change?

ioannisj (approved):
LG, left a minor comment on options.disableXHRCredentials since it seems it's not just affecting preview-flag users?

packages/browser/src/request.ts
marandaneto: it is, the changeset mentions __preview_disable_xhr_credentials

marandaneto:
> LG, left a minor comment on options.disableXHRCredentials since it seems it's not just affecting preview-flag users?

disableXHRCredentials is internal, this is set from __preview_disable_xhr_credentials

PR #3792, opened 2026-06-10
chore: make release version bump pull main before release and fail if main is durty

marandaneto:
## Problem

The release workflow can fail if `main` advances while the approved version-bump job is generating package versions, changelogs, references, and lockfile updates. This happened in https://github.com/PostHog/posthog-js/actions/runs/27272295505/attempts/1, where `planetscale/ghcommit-action` refused to commit because `main` no longer pointed at the workflow's starting SHA.

## Changes

Before committing the generated version bump, update the checkout to the latest `main` and fail if dirty before release

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -94,6 +94,9 @@ jobs:
                   fetch-depth: 0
                   token: ${{ steps.releaser.outputs.token }}
 
+            - name: Update to latest main before generating release artifacts
+              run: git pull --ff-only origin main
+
             - name: Setup environment
               uses: ./.github/actions/setup
               with:
@@ -122,8 +125,13 @@ jobs:
                   echo "version=$VERSION" >> "$GITHUB_OUTPUT"
                   echo "Resolved posthog-js version: $VERSION"
 
-            - name: Update to latest main before committing
-              run: git pull --ff-only --autostash origin main
+            - name: Ensure main did not advance during release artifact generation
+              run: |
+                  git fetch origin main
+                  if [ "$(git rev-parse HEAD)" != "$(git rev-parse FETCH_HEAD)" ]; then
+                    echo "::error::main advanced while generating release artifacts. Re-run the workflow so versions and changelogs include the latest changesets."
+                    exit 1
+                  fi
 
             - name: Commit version bump and lockfile
               id: commit-version-bump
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -122,6 +122,9 @@ jobs:
                   echo "version=$VERSION" >> "$GITHUB_OUTPUT"
                   echo "Resolved posthog-js version: $VERSION"
 
+            - name: Update to latest main before committing
+              run: git pull --ff-only --autostash origin main
```
dustinbyrne: we've already version bumped by this point, so isn't it possible we pull in an additional, releasable commit? it may get released during this run, but it'd have missed the changelog, and will still trigger another (empty, other than version bump & changelog) release afterwards

marandaneto: Good catch — pulling after `pnpm bump` was unsafe. Fixed in 1dcda75: the workflow now fast-forwards before generating release artifacts, then fails before committing if `main` advanced during generation so the run can be retried with fresh changelogs/version bumps.

.github/workflows/release.yml
```diff
@@ -122,6 +125,14 @@ jobs:
                   echo "version=$VERSION" >> "$GITHUB_OUTPUT"
                   echo "Resolved posthog-js version: $VERSION"
 
+            - name: Ensure main did not advance during release artifact generation
+              run: |
+                  git fetch origin main
+                  if [ "$(git rev-parse HEAD)" != "$(git rev-parse FETCH_HEAD)" ]; then
+                    echo "::error::main advanced while generating release artifacts. Re-run the workflow so versions and changelogs include the latest changesets."
+                    exit 1
+                  fi
```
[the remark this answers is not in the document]
marandaneto: updated pr desc

dustinbyrne (approved):

PR #3800, opened 2026-06-10
fix(replay): keep linkedFlagSeen sticky once the flag has matched

ksvat:
## Problem

While investigating a customer's missing session replays, I traced a class of recordings that:

1. Started recording correctly — linked flag matched, `recording_status: active`, snapshots flowing into the local buffer
2. Then flipped to `recording_status: buffering` mid-session
3. Lost the entire buffered recording when the user navigated or closed the tab

The trigger is any operation that causes `posthog.featureFlags` to reload while the recorder is running — `identify()`, `group()`, `setPersonPropertiesForFlags()`, or an explicit `reloadFeatureFlags()`. During the in-flight window, the `onFeatureFlags` callback can fire with the linked flag absent or set to `false` (e.g. before group properties have been re-sent to the server), and `LinkedFlagMatching` flipped `linkedFlagSeen` back to `false` based on that transient value. The recorder state machine moved to `BUFFERING`, and if the user left the page before the next reload restored the flag, `_onBeforeUnload` silently called `_clearBuffer()` on the way out, dropping everything rrweb had captured for the session.

## Changes

`LinkedFlagMatching.onConfig` no longer un-sets `linkedFlagSeen` once it's been set. The boolean is now monotonic per matcher instance: once the configured flag has been observed to match for this session, the matcher stays activated even if a later callback reports the flag as missing or false. Recording continues uninterrupted across the reload window.

This applies to both:
- V1 strategy — the recorder-level `_linkedFlagMatching` instance
- V2 strategy — the per-group `LinkedFlagMatching` inside each `TriggerGroupMatching`

since both code paths share the same class.

Updated three existing test cases in the `linked flags` describe block that asserted the previous "flip back to buffering on flag drop" behavior, since that behavior was the bug. Added a new regression test (`keeps recording when a flag reload transiently drops the flag`) that explicitly walks through the customer-reported sequence: flag matches → reload returns empty variants → explicit `false` response — and asserts the recorder stays `active` the whole way.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/sticky-linked-flag-seen.md
+++ b/.changeset/sticky-linked-flag-seen.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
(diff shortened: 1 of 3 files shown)


ksvat:
* **#3800** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3800?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3800?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #3801, opened 2026-06-10
fix(replay): prevent silent recorder teardown on session-id rotation

ksvat:
## Problem

When the session id rotates during active rrweb capture, `_updateWindowAndSessionIds` calls `stop()` and then synchronously `start('session_id_changed')`. If `stop()` takes the `_stopAfterCompressionQueueDrains` path — which it does whenever the compression queue is non-empty, the steady-state condition for any actively-recording session on a modern browser — its async cleanup later resolves and calls `_teardown()` against the *freshly-started* recorder. The teardown:

- removes all event listeners (`beforeunload`, `offline`, `online`, `visibilitychange`)
- stops rrweb (`_stopRrweb?.()` and `_stopRrweb = undefined`)
- calls `_strategy.stop()`, which empties V2's `_triggerGroupMatchers` array

After that, the recorder still has its `_strategy` reference, so the `status` getter keeps returning `active`/`sampled` for any captured event. But rrweb is no longer producing snapshots, no unload handler is registered, and the V2 trigger groups are gone. From the server's perspective, events keep arriving with `$recording_status: active` while no `$snapshot` data ever lands.

### How often does the bug actually fire?

The compression queue is populated by `shouldUseNativeAsyncSessionRecordingGzip`, which is `true` whenever:

1. The browser exposes `CompressionStream` (Chrome 80+ / Edge 80+ / Firefox 113+ / Safari 16.4+). In 2026 this is essentially every browser actually used in production — Safari 16.4 shipped March 2023, the others years earlier.
2. The event is a `FullSnapshot` or an `IncrementalSnapshot` with source `Mutation`/`StyleSheetRule`.

The first event after rrweb attaches is always a `FullSnapshot`, and every DOM mutation produces an Incremental/Mutation. So **any session where the user is actually doing things in the app has a populated compression queue at the moment of rotation**, and any session-id rotation during that activity exposes this race.

### The path that triggers it

1. User has a long-running tab. rrweb is steady, compression queue is non-empty.
2. Session id rotates — idle timeout, session-past-max-length, or after `posthog.reset()`.
3. `_updateWindowAndSessionIds` runs `stop()` → `_stopAfterCompressionQueueDrains` takes the async branch, captures the current `_compressionQueueGeneration`, queues teardown for after the drain.
4. `start('session_id_changed')` runs synchronously: constructs a new V2 strategy, calls `_startRecorder()` (rrweb attaches, new `_stopRrweb` set), attaches new event listeners.
5. Drain promise resolves. The pending cleanup re-checks the generation, finds it unchanged, and runs `_teardown()` on the now-live new recorder.

This matches a clean signal in real customer data: long sessions (minutes to hours) entered from an already-open tab where status reports `active` throughout but no replay data arrives. Short `/login`-started sessions for the same user on the same day record fine, because they start fresh without an in-flight `stop()` pending.

## Changes

`start()` now invalidates the compression-queue state at the top, via `_invalidateCompressionQueue()` — the same block `_teardown()` already runs: generation bump plus reset of `_pendingCompressionEvents`, `_queuedCompressionEvents`, `_compressionQueue`, and `_isStoppingAfterCompression`. Any in-flight async cleanup from a prior `stop()` then bails out at its existing generation check (`if (generation !== this._compressionQueueGeneration) return`) instead of tearing down the new recorder.

The mechanism is the same one the code already uses elsewhere: every other place that wants to invalidate stale async work — `_teardown()` itself, `discard()` — already does this. `start()` was the only lifecycle point that could leave behind stale cleanup without invalidating it. This change completes the pattern.

The full reset (rather than only a generation bump, as in the first revision of this PR) matters: the bailed-out drain never resets `_isStoppingAfterCompression`, and stale-generation events never decrement `_queuedCompressionEvents`, so a bare bump would leave every later `stop()` of the new recorder hitting the in-progress guard in `_stopAfterCompressionQueueDrains` and silently no-oping — the new recorder could never be stopped (second rotation, `posthog.reset()`, etc.). Caught by Greptile's review.

Trade-off: any compressed events for the prior session that hadn't yet been flushed are dropped. These are end-of-life trailing data for a session that already rotated out; the alternative is losing the entire new session's replay.

## Tests

Added a unit test that:

1. Establishes a recording session and manually populates `_pendingCompressionEvents` and `_compressionQueue` to simulate the in-flight async drain.
2. Rotates the session id, triggering `_updateWindowAndSessionIds` → `stop()` + `start('session_id_changed')`.
3. Resolves the drain promise to fire the queued cleanup.
4. Asserts the new recorder is still alive (`isStarted === true`, status is one of `active`/`sampled`/`buffering`).

The test fails on `main` (the assertion at step 4 produces `isStarted === false` because `_teardown()` ran against the new recorder) and passes with this fix applied.

A second unit test covers the review finding: after the same rotation + drain resolution, a subsequent `stop()` must actually tear the recorder down (`isStarted === false`). It fails with only the generation bump (stop() silently no-ops on the stale in-progress state) and passes with the full invalidation.

## Related

#3690 (Paul) reduces how often session-id rotations spuriously fire in multi-tab scenarios by refreshing from storage before declaring idle. That PR and this one are orthogonal and complementary:

- Paul's PR makes rotations happen less often.
- This PR makes the recorder survive correctly whenever a rotation does happen.

Both should land. Neither makes the other unnecessary.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes


ksvat:
* **#3801** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3801?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3801?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -880,6 +880,14 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
             return
         }
 
+        // Invalidate any in-flight async cleanup queued by a prior stop(). On a session-id
+        // rotation, _updateWindowAndSessionIds calls stop() then start() synchronously; if
+        // stop() took the _stopAfterCompressionQueueDrains path (non-empty compression queue),
+        // its async cleanup would later call _teardown() and silently tear down THIS new
+        // recorder once the drain promise resolves. Bumping the generation here causes that
+        // pending cleanup to bail at its generation check.
+        this._compressionQueueGeneration += 1
```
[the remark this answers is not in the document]
pauldambra: Good catch — confirmed real, and slightly worse than described: besides `_isStoppingAfterCompression` staying `true`, the stale-generation event never decrements `_queuedCompressionEvents` (the generation check in `_finishQueuedCompressionEvent` skips it), so the synchronous stop path stays blocked too and every later `stop()` is a silent no-op.

Fixed in 19bbfd83 by going one step further than the suggested one-liner: `_teardown()`'s existing compression-state reset is extracted into `_invalidateCompressionQueue()` and `start()` now calls that instead of a bare generation bump. That resets the flag, the queued-event count, the pending events, and the queue reference together — so a generation bump always travels with the full state reset and the mismatch-bail branch needs no special-casing.

Added the missing test: after the rotation survives, a subsequent `stop()` must actually tear the recorder down. It fails with only the generation bump and passes with the full invalidation.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments and the summary comment for the overall verdict.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -880,6 +880,16 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
 
+        // Invalidate any in-flight async cleanup queued by a prior stop(). On a session-id
+        // rotation, _updateWindowAndSessionIds calls stop() then start() synchronously; if
+        // stop() took the _stopAfterCompressionQueueDrains path (non-empty compression queue),
+        // its async cleanup would later call _teardown() and silently tear down THIS new
+        // recorder once the drain promise resolves. Bumping the generation here causes that
+        // pending cleanup to bail at its generation check, and resetting the rest of the
+        // compression-stop state means a future stop() of this new recorder is not mistaken
+        // for the still-in-progress old one (which would make it a silent no-op).
+        this._invalidateCompressionQueue()
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟡 MEDIUM

When `start()` bumps the generation, the old `stop()`'s drain promise later bails at the generation check *before* it reaches `_flushBuffer()`/`_clearBuffer()` — so the tail of the previous session (the queued compressed events, plus whatever was already sat in `_buffer` that the synchronous `stop()` deliberately deferred) just evaporates. The changeset acknowledges the trade-off and it is clearly the right call (losing the old tail beats losing the whole new session), but it is silent data loss at the session boundary with no log or metric: replay completeness degrades invisibly and there is no way to tell in grafana how often this fires or how much it sheds.

in paul's words: "every dropped message is a probably-unplayable bit of recording... could we increment-and-log here — even just a `logger.info('dropping stale compression queue on restart', { queued: this._queuedCompressionEvents })` inside `_invalidateCompressionQueue` when it's non-empty — so we can watch it rather than guess? not blocking, but i'd really like the meter."

pauldambra: > [!NOTE]

`_invalidateCompressionQueue()` is now unconditional at the top of `start()`, but `start()` has a third caller (`session-recording.ts` `_startCapturing`) that does NOT pair with a preceding `stop()`. The two rotation callers always `stop()` first so they are safe — but if `start()` is ever re-entered on an already-active recorder with a non-empty compression queue (queued snapshots for the CURRENT, still-live session), this call drops `_pendingCompressionEvents` and zeroes `_queuedCompressionEvents`, silently discarding in-flight snapshots for a session that is continuing.

The invalidation's safety depends on an invariant (start() is always preceded by stop(), OR the queue is genuinely stale) that is not enforced or documented at the call sites — a latent trap for a future caller invoking `start()` to "refresh" a live recorder. Suggestion: guard the invalidation so it only fires when there is stale stop-in-progress state to clear (e.g. only when `_isStoppingAfterCompression` is true), or document the invariant where it can't be missed.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1110,8 +1120,12 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
         this._stopRecordingProducers()
 
-        // Invalidate any in-flight async compression work so it does not capture events
-        // after stop()/discard() has cleared the buffer or after a future restart.
+        this._invalidateCompressionQueue()
+    }
+
+    // Invalidate any in-flight async compression work so it does not capture events
+    // after stop()/discard() has cleared the buffer or after a future restart.
+    private _invalidateCompressionQueue() {
```
pauldambra, line 1133: > [!NOTE]

**[xp]** 🟢 LOW

The name `_invalidateCompressionQueue` undersells what it does — it resets five fields of compression-*stop* state (pending events, queued count, queue ref, the stopping flag, plus the generation), not just "the queue". Your own second commit message says it better: "reset all compression-stop state, not just the generation." Consider `_resetCompressionStopState()` — at the `start()` call site that name would carry more of the meaning, and the 8-line comment there could shrink to a one-liner like "a prior stop() may have queued async cleanup keyed on the old generation". The full stop()/start() sequencing narrative arguably belongs at the `_stopAfterCompressionQueueDrains` generation-check (the thing being defended), not at the call that defends it.

pauldambra, line 1133: > [!NOTE]

**[paul]** 🟢 LOW

nice extract — having `start()` and `_teardown()` reset through the one method means the two sites can't drift apart, which is exactly the bug class we'd want to avoid here. one tiny thing while you're in here: `_stopAfterCompressionQueueDrains` snapshots `const generation` and the drain's `.then` compares against the live field, which is the whole mechanism. worth a half-second sanity check that nothing _else_ resets `_isStoppingAfterCompression` to false out from under an in-flight drain now that two callers (start + teardown) can touch this state — i think it's fine because the generation guard fences it, but i'm lazily asking for your read rather than re-deriving it myself.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -1359,6 +1359,88 @@ describe('Lazy SessionRecording', () => {
...
                 }
             })
+
+            // Verifies the suspected teardown-race between stop()'s async compression-queue
+            // drain and the synchronous start('session_id_changed') that follows. If the
+            // pending cleanup proceeds (its generation check passes because nothing bumped
+            // it during start()), it calls _teardown() on the recorder — silently undoing
+            // the restart. The recorder then appears to be in an ACTIVE strategy state but
+            // rrweb is dead, no listeners are attached, and snapshots never reach the server.
+            it('keeps the new recorder alive when stop() had an in-flight compression queue', async () => {
```
pauldambra, line 1369: > [!NOTE]

**[paul]** 🟢 LOW

these two tests are the same ~15 lines of setup (emit, fake a non-empty queue, rotate, resolve the drain) with a different tail — one asserts the recorder survived, the other asserts a subsequent stop() still works. you know i can't see two near-identical test bodies without saying the word "parameterised" 🙂 either a shared `arrangeRotationWithInflightQueue()` helper returning `{ lazyRecorder, resolveDrain }`, or a single it.each with the differing assertion as the param. not blocking, ship as you see fit — but the second test is the one that actually guards the "reset everything not just the generation" decision, so i'd hate for the duplication to make someone delete it later thinking it's redundant.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -1359,6 +1359,88 @@ describe('Lazy SessionRecording', () => {
...
+                const recordMock = assignableWindow.__PosthogExtensions__.rrweb.record as Mock
+                const lazyRecorder = sessionRecording['_lazyLoadedSessionRecording']
+
+                // Establish a recording session and pretend the compression queue is non-empty.
+                emitActiveEvent(startingTimestamp + 100)
+                expect(recordMock).toHaveBeenCalledTimes(1)
+                expect(lazyRecorder['isStarted']).toEqual(true)
+
+                lazyRecorder['_queuedCompressionEvents'] = 1
```
pauldambra, line 1378: > [!NOTE]

**[xp]** 🟢 LOW

Both tests hand-assemble the race by poking privates (`_queuedCompressionEvents = 1`, a manually-resolved `_compressionQueue`). Pragmatic for a hard-to-provoke timing bug, and I wouldn't block on it — but it couples the tests to internal field names, so a future refactor of the queue mechanics breaks the tests for reasons unrelated to the behaviour. If there's any seam to make compression genuinely async in-test (a slow/awaitable compressor), driving the real path would be more durable. Otherwise fine.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -1359,6 +1359,88 @@ describe('Lazy SessionRecording', () => {
...
+                // stop(). With the bug, the cleanup calls _teardown() on the new recorder.
+                resolveDrain()
+                await Promise.resolve()
+                await Promise.resolve()
+
+                // The new recorder must still be alive. If the teardown race fired, isStarted
+                // would be false (rrweb stopped) and the V2 strategy would have its matchers
+                // cleared so status returns 'disabled'.
+                expect(lazyRecorder['isStarted']).toEqual(true)
+                expect(['active', 'sampled', 'buffering']).toContain(sessionRecording.status)
```
pauldambra, line 1408: > [!NOTE]

**[paul]** ⚪ NIT

silly question — `expect(['active', 'sampled', 'buffering']).toContain(status)` is a slightly loose assertion. is it loose because the strategy/sampling decision is non-deterministic across the v1/v2 branch, or just defensiveness? if we actually know which status this session lands in given the test's config, pinning it exactly would catch a regression where the recorder survives but lands in the wrong state. if it genuinely varies, ignore me — the set-membership check is the honest thing to write.

pauldambra:
> [!NOTE]
>

The fix is correct and well-tested — the generation-bump-plus-full-reset is the right mechanism, the orphaned async closures all re-check the generation and no-op cleanly, and the second test pins the exact regression a naive generation-only fix would have introduced. The two MEDIUM findings are an observability gap and a latent invariant, not correctness bugs.

### Key findings

**🟡 MEDIUM**
- **Silent drop of the prior session's unflushed tail with no metric/log** (`lazy-loaded-session-recorder.ts:891`) — the bailed drain skips `_flushBuffer()`; the trade-off is right but currently unobservable. Suggest counting/logging when the invalidation discards a non-empty queue. *(convergent: qa-team/reliability + paul)*
- **Unconditional invalidation relies on an undocumented invariant** (`lazy-loaded-session-recorder.ts:891`) — `start()` has a caller (`_startCapturing`) that doesn't `stop()` first; a future re-entrant `start()` on a live recorder would drop its own in-flight snapshots. Guard on `_isStoppingAfterCompression` or document the invariant. *(qa-team/generalist-b)*

**🟢 LOW / ⚪ NIT**
- Rename `_invalidateCompressionQueue` -> `_resetCompressionStopState` and shrink the 8-line call-site comment *(xp)*
- The two new tests share ~15 lines of setup — extract a helper or parameterize *(paul)*
- Tests poke private fields to assemble the race — pragmatic but coupled to internals *(xp)*
- Loose `toContain(status)` assertion — pin it if the status is deterministic *(paul)*

### Convergence

The silent old-session snapshot drop was flagged independently by qa-team (4 of its internal personas) and paul-reviewer — highest-confidence finding in this review. security-audit independently verified the same code path and confirmed it is a reliability trade-off, **not** a data-attribution or consent issue.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | MEDIUM — mechanism verified correct (no leak/double-count from orphaned closures); flags the unobservable drop and the re-entrant-start invariant; compatibility/security personas: no findings. |
| 👤 paul | approve with comments, high trust — "lovely bit of debugging"; wants a meter on the knowingly-dropped tail before assuming it only ever sheds a little. |
| 📐 xp | Ship it — Four Rules satisfied (tests pin both failure modes, extract honors OnceAndOnlyOnce); rename + comment-trim suggested as optional polish. |
| 🛡 security-audit | No exploitable findings — traced cross-session attribution (buffer self-stamps its own session id, mismatch flushes under the OLD id), opt-out stop integrity, double-rrweb, and listener duplication end to end; all sound. |

---

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
pauldambra: > [!NOTE]

Declined by maintainer — the SDK runs in customer sites and we should be sparing with observability. Logging inside the SDK for this edge-case is out of scope here; the generation-guard already prevents the data loss, and any future need for metrics can be addressed as a separate opt-in mechanism.

pauldambra: > [!NOTE]

The invariant (start() is always preceded by stop(), or the queue is genuinely stale) holds at all current call sites — both rotation callers run stop()+start() synchronously. The single-reviewer concern about a future caller invoking start() to "refresh" a live recorder is noted. Treating this as out-of-scope for this fix; the bug being addressed here is specifically the teardown race on the stop()+start() rotation path. A follow-up can add a guard if a re-entrant start() use-case materialises.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
pauldambra: > [!NOTE]

LOW/style rename — out of scope for this fix. The current name is consistent with the existing codebase naming convention (queue-focused). A rename can be done as a follow-up if desired.

pauldambra: > [!NOTE]

Yes, the generation guard fences it correctly. _isStoppingAfterCompression is only reset to false inside _invalidateCompressionQueue(), which is called from exactly two places: start() and _teardown(). Both sites also bump _compressionQueueGeneration in the same call. So when an in-flight drain's .then() fires after a start() or _teardown() has already reset the flag, it also sees a mismatched generation and bails at the generation check before doing anything. There is no path where the flag is cleared out from under a drain that still has the right generation.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
pauldambra: > [!NOTE]

LOW — agree the duplication is real and a shared arrange helper would be cleaner. Leaving as-is for this PR to keep the fix minimal; can be addressed in a follow-up refactor.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
pauldambra: > [!NOTE]

LOW — poking private fields is pragmatic for a hard-to-provoke timing race; the alternative (a genuinely async compressor seam) is a larger change than the bug fix warrants. The coupling is accepted as a trade-off here.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
pauldambra: > [!NOTE]

NIT — the set-membership check is intentional: the v1/v2 strategy branch means the exact status is config-dependent (v2 TriggerGroupStrategy vs v1 RecordingStrategy produce different initial states depending on trigger config). Pinning the exact value would make the test fragile across strategy variants. The set membership check is the honest assertion here.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
pauldambra: > [!NOTE]

Correcting the earlier resolution: this finding was right, and CI proved it. The opting-out and linked-flags Playwright suites failed on all three browsers — opt-in flows call `start()` re-entrantly via `_startCapturing` with no preceding `stop()`, and the unconditional invalidation dropped the live session's queued FullSnapshot (async-compressed) while the meta event (sync path) survived, producing `[meta, meta]` instead of `[meta, full_snapshot]`.

Fixed in `a8d72861` with the guard this comment suggested: the invalidation now only fires when `_isStoppingAfterCompression` is true — i.e. exactly when a prior `stop()` left the in-flight async cleanup this fence exists for. Added a unit test (`re-entrant start() preserves the live compression queue`) that fails without the guard, and verified both failing Playwright specs pass with it.

TueHaulund (approved):

PR #3832, opened 2026-06-14
fix: Guard remaining survey localStorage accesses against SecurityError (#3588)

archievi:
## Changes

Fixes #3588: surveys throw an uncaught `SecurityError` when `localStorage` is unavailable (most commonly in cross-origin iframes), which floods users' error monitoring (PostHog, Sentry, etc.).

Most survey storage access is already wrapped in `try/catch` (e.g. `surveys-extension-utils.tsx`), but two spots were still unprotected and are the remaining source of the reported errors:

- `PostHogSurveys.reset()` in `packages/browser/src/posthog-surveys.ts` — `localStorage.removeItem` / `.length` / `.key(i)`.
- The `lastSeenSurveyDate` write in the `PHSurveySent` handler in `packages/browser/src/extensions/surveys.tsx`.

Both are now wrapped in `try/catch`, matching the existing best-effort idiom used elsewhere in the survey code (`// localStorage not available`). Survey reset and the last-seen-date write are non-critical, so swallowing the error is correct here.

## Tests

Added a regression test in `posthog-surveys.test.ts` that mocks `Storage.prototype.removeItem` to throw and asserts `reset()` no longer throws. It fails on `main` and passes with this change.

### Verification (run locally)
- `jest src/__tests__/posthog-surveys.test.ts` — 41 pass (regression-proven against revert).
- `jest src/__tests__/extensions/surveys.test.ts` — 68 pass (no regressions).
- `tsc --noEmit` clean, `eslint` clean, `prettier --check` clean.

## Checklist
- [x] Added a changeset (`patch`).
- [x] Added a test that reproduces the bug and verifies the fix.
- [x] Followed the existing storage-guarding pattern in the survey code.

👋 Hi, I use PostHog and hit this while running surveys in an embedded context. Happy to extend the same guard to any other unprotected survey storage access if you'd like.

### DIFF
--- a/.changeset/surveys-localstorage-iframe-guard.md
+++ b/.changeset/surveys-localstorage-iframe-guard.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Surveys: guard the remaining unprotected `localStorage` accesses (`reset()` and the `lastSeenSurveyDate` write) so a `SecurityError` in cross-origin iframes is swallowed instead of bubbling up to user monitoring.
--- a/packages/browser/src/__tests__/posthog-surveys.test.ts
+++ b/packages/browser/src/__tests__/posthog-surveys.test.ts
@@ -151,6 +151,18 @@ describe('posthog-surveys', () => {
             localStorage.clear()
         })
 
+        describe('reset', () => {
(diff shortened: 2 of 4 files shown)


marandaneto:
thanks @archievi

marandaneto (approved):

PR #3834, opened 2026-06-15
ci: check public API references

marandaneto:
## Problem

Public API reference generation and checking were combined in a single CI command, which made it harder to tell whether failures came from the API extraction/generation step or from outdated committed reference snapshots.

## Changes

- Split the library CI public API job into separate generate and check steps.
- Added a `check:public-api` script that only verifies generated latest reference snapshots are clean.
- Added `scripts/check-public-api.js` to fail when tracked latest public API reference snapshots are out of date.
- Handle `git status` failures with a concise error instead of an unhandled Node.js stack trace.
- Limit the CI check to `*-references-latest.json` files so frozen versioned reference files are not regenerated/required on PRs.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


marandaneto:
similar to flutter/dotnet https://github.com/PostHog/posthog-flutter/pull/437

scripts/check-public-api.js
```diff
@@ -0,0 +1,13 @@
+const { execFileSync } = require('child_process')
+
+const referencePaths = ['packages/browser/references', 'packages/node/references', 'packages/react-native/references']
+
+const status = execFileSync('git', ['status', '--porcelain', '--', ...referencePaths], { encoding: 'utf8' }).trim()
```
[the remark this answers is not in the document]
marandaneto: Fixed in 7b01d897 — the git status call is now wrapped and exits with a concise error if it fails.

scripts/check-public-api.js
```diff
@@ -0,0 +1,19 @@
...
+let status
+try {
+    status = execFileSync('git', ['status', '--porcelain', '--', ...referencePaths], { encoding: 'utf8' }).trim()
+} catch (error) {
+    console.error('Failed to check public API references:', error.message)
+    process.exit(1)
+}
+
+if (status) {
+    console.error('Public API references are out of date. Run `pnpm generate-references` and commit the updated files.')
```
ioannisj, line 18: Out of all implementations, I'm not really sure about this one. checking `packages/<package>/references` I see that we keep a frozen file per version. Won't pnpm generate-references force the PR author to regenerate an already released version?

marandaneto: Good point. Fixed in 8f40bf03 — the check now only looks at the three tracked `*-references-latest.json` files, so CI won't require changes to frozen versioned reference files generated by `pnpm generate-references`.

turnipdabeets (approved):

ioannisj (approved):

PR #3848, opened 2026-06-15
feat(flags): add disableRemoteFeatureFlags option and runtime updateFlags

### LINKED ISSUE #3786: Feature flag parity for posthog-react-native: add a full-disable option and a runtime updateFlags (flags, payloads)

Yes, we have a React Native app `posthog-react-native` that wraps our web app in a WebView. We locally evaluate feature flags on our own backend `posthog-node` with onlyEvaluateLocally) for our own stable distinct ID and pass the results down to native, so we'd like the native SDK to use exactly those values and never fetch or evaluate flags on its own. This is the same approach our web app already uses, where we set advanced_disable_feature_flags and then call posthog.updateFlags(flags, payloads).

We can already do the init-time half of this on mobile:

- preloadFeatureFlags: false stops the startup /flags load, which is roughly the equivalent of the web's advanced_disable_feature_flags_on_first_load.
- The bootstrap option accepts both flags and payloads at construction time, matching the web's bootstrap config.

What we can't do is the runtime half. Our flags arrive over the WebView bridge after the user logs in, not when the SDK is constructed, and there are two missing pieces:

1. There's no way to stop the SDK from fetching flags on identify. preloadFeatureFlags: false only covers the initial load. As soon as we call identify (or alias, group, or setPersonProperties), the SDK calls reloadFeatureFlags() and hits /flags again. There's no equivalent of the web's advanced_disable_feature_flags, and the only full off switch is disabled: true, which also turns off event capture.

2. There's no way to hand the SDK locally evaluated flags with payloads at runtime. The only runtime option is overrideFeatureFlag(flags), which doesn't accept payloads, is stored as a sticky override rather than a normal flag set, and reads back with override semantics. There's nothing like the web's updateFlags(flags, payloads, { merge }) or receivedFeatureFlags(...).

These two depend on each other. Without the first, the SDK refetches and overwrites whatever we set. Without the second, we can't supply the values (with payloads) in the first place.

Bring `posthog-react-native` and `@posthog/core` in line with `posthog-js` for the runtime flag model:

1. A full feature-flag disable option, like the web's advanced_disable_feature_flags, that:
   - turns reloadFeatureFlags() into a no-op so the reloads triggered by identify, alias, group, and setPersonProperties don't fire, and
   - sends disable_flags: true on any /flags request that still goes out, matching what `posthog-js` does.

2. A public runtime updateFlags method matching the web signature:

   ```ts
   updateFlags(
     flags: Record<string, boolean | string>,
     payloads?: Record<string, JsonType>,
     options?: { merge?: boolean }
   ): void

turnipdabeets:
## Problem

https://posthoghelp.zendesk.com/agent/tickets/60063 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61806)

Apps that evaluate feature flags outside the SDK — e.g. on their own backend with `posthog-node` local evaluation, then pushed into a React Native app over a WebView bridge after login — can't tell `posthog-react-native` to stop fetching flags and use the supplied values at runtime. The web SDK already supports this (`advanced_disable_feature_flags` + `updateFlags`); core/RN didn't. Closes #3786.

The init-time half already existed (`preloadFeatureFlags: false`, `bootstrap`). This adds the missing runtime half.

Docs added: https://github.com/PostHog/posthog.com/pull/17664

## Changes

Two additions to `@posthog/core` (inherited by `posthog-react-native` and `posthog-js-lite`):

- **`disableRemoteFeatureFlags` option** (parity with web's `advanced_disable_feature_flags`): turns `reloadFeatureFlags()` and the reloads triggered by `identify()`/`group()`/`setPersonPropertiesForFlags()`/`reset()` into no-ops, sends `disable_flags: true` on any config/survey request that still goes out, and never lets a server response overwrite locally supplied flags.
- **Public `updateFlags(flags, payloads?, { merge? })`** (parity with web's `updateFlags`): supply locally evaluated flags + payloads at runtime. Values persist, `getFeatureFlag()`/`getFeatureFlagPayload()` read them back, and the `featureflags` event fires (so linked-flag session replay re-evaluates). No network request.

Notes for reviewers:
- All behavior is gated behind the option, which defaults to `false` — existing clients are unchanged.
- `posthog-node` is unaffected (it extends `PostHogCoreStateless`, not `PostHogCore`; the only stateless touch is `disable_flags` riding through `getFlags`'s existing `extraPayload`).
- `reset()` clears pushed flags (same as web) — callers re-push after an identity change. Until `updateFlags()` runs, reads return not-loaded defaults, so `bootstrap` is the way to supply startup values.
- **Manual testing** (beyond unit tests):
  - reproduced the issue's exact setup end-to-end — `posthog-node` (`onlyEvaluateLocally`) evaluated flags + payloads for a stable distinct id, handed them to a React Native app (iOS simulator) running `disableRemoteFeatureFlags`, and pushed them via `updateFlags`. Confirmed `identify()`/`group()` trigger zero `/flags` calls, reads match the node-evaluated values + payloads, and the captured event lands in PostHog (capture still works under the option).
  - exercised on web (browser), iOS (simulator), and Android (device) — no `/flags` traffic under the option, `updateFlags` round-trips flags + payloads, cold-start persistence, and the logout/login re-push cycle.

### Libraries affected

Minor bump for `@posthog/core`, `posthog-react-native`, and `posthog-js-lite` (additive option + method; the latter two gain the new public API by inheritance).

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/core/src/posthog-core.ts
```diff
@@ -1142,6 +1169,65 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+        flagDetails[key] = {
+          key,
+          enabled: getEnabledFromValue(value),
+          variant: getVariantFromValue(value),
+          reason: undefined,
+          // id 0 marks a locally supplied flag
+          metadata:
+            payload !== undefined
+              ? { id: 0, version: undefined, description: undefined, payload: JSON.stringify(payload) }
+              : undefined,
```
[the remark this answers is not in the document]
turnipdabeets: addressed as undefined

packages/node/references/posthog-node-references-5.37.1.json
```diff
@@ -2153,7 +2153,7 @@
...
+      "example": "{\n    host?: string;\n    flushAt?: number;\n    flushInterval?: number;\n    maxBatchSize?: number;\n    maxQueueSize?: number;\n    disabled?: boolean;\n    defaultOptIn?: boolean;\n    sendFeatureFlagEvent?: boolean;\n    preloadFeatureFlags?: boolean;\n    advancedDisableFeatureFlags?: boolean;\n    disableRemoteConfig?: boolean;\n    disableSurveys?: boolean;\n    bootstrap?: {\n        distinctId?: string;\n        isIdentifiedId?: boolean;\n        featureFlags?: Record<string, FeatureFlagValue>;\n        featureFlagPayloads?: Record<string, JsonType>;\n    };\n    fetchRetryCount?: number;\n    fetchRetryDelay?: number;\n    requestTimeout?: number;\n    featureFlagsRequestTimeoutMs?: number;\n    remoteConfigRequestTimeoutMs?: number;\n    sessionExpirationTimeSeconds?: number;\n    disableCompression?: boolean;\n    disableGeoip?: boolean;\n    historicalMigration?: boolean;\n    evaluationContexts?: readonly string[];\n    evaluationEnvironments?: readonly string[];\n    personProfiles?: 'always' | 'identified_only' | 'never';\n    before_send?: BeforeSendFn | BeforeSendFn[];\n    addTracingHeaders?: string[];\n}"
```
dustinbyrne: changes to `node/references` and `react-native/references` should be dropped

these files are written by CI / release, and are versioned to render reference documentation on posthog.com ([example](https://posthog.com/docs/references/posthog-node-5.37.1))

packages/core/src/posthog-core.ts
```diff
@@ -1093,6 +1112,10 @@ export abstract class PostHogCore extends PostHogCoreStateless {
 
   // Used when we want to trigger the reload but we don't care about the result
   reloadFeatureFlags(options?: { cb?: (err?: Error, flags?: PostHogFlagsResponse['featureFlags']) => void }): void {
+    if (this.advancedDisableFeatureFlags && !this.disabled) {
+      this._initPromise.then(() => options?.cb?.(undefined, this.getFeatureFlags()))
+      return
+    }
```
dustinbyrne: this short circuit will prevent `setKnownFeatureFlagDetails` from being called on reset

it's important because it emits the `'featureFlags'` event which has a decent amount of listeners via the `onFeatureFlags` registration method

dustinbyrne: this is worth a test or two

packages/core/src/types.ts
```diff
@@ -53,6 +53,24 @@ export type PostHogCoreOptions = {
...
+   * related methods keep working against those values. Until `updateFlags()` runs, reads
+   * return their not-loaded defaults, so use `bootstrap` for any flags needed at startup.
+   * Equivalent to the web SDK's `advanced_disable_feature_flags`.
+   *
+   * Note: surveys gated on feature flags will not evaluate unless the survey targeting
+   * flags are also provided via `updateFlags()`. This option cannot be toggled at runtime.
+   *
+   * @default false
+   */
+  advancedDisableFeatureFlags?: boolean
```
dustinbyrne: I know this is derived from `posthog-js`, but this is a good chance for us to give it a decent name.

what do you think? `disableRemoteFeatureFlags`, `enableRemoteFeatureFlags`, `remoteFeatureFlags` - open to anything here. what ever you decide on we can port to posthog-js v2

.changeset/rn-flags-disable-and-update-flags.md
```diff
@@ -0,0 +1,7 @@
+---
+'@posthog/core': minor
+'posthog-react-native': minor
+'posthog-js-lite': minor
+---
+
+feat(flags): add the `advancedDisableFeatureFlags` option and a public `updateFlags(flags, payloads?, { merge? })` method, matching the web SDK's `advanced_disable_feature_flags` and `updateFlags`. With the option set, `reloadFeatureFlags()` and the reloads triggered by `identify()`, `group()`, `setPersonPropertiesForFlags()` and `reset()` become no-ops, and any flags request that still goes out for remote config or surveys carries `disable_flags: true` so the server skips flag evaluation. `updateFlags` supplies locally evaluated flag values (with payloads) at runtime: values persist, `getFeatureFlag()`/`getFeatureFlagPayload()` read them back, and `onFeatureFlags` listeners fire — so React Native session replay gated on a linked flag re-evaluates when flags are pushed in.
```
dustinbyrne: nit - this goes right into the CHANGELOG / release notes as markdown under a header like `## Minor Changes`, so the conventional commit tag `feat(flags)` isn't needed

packages/core/src/posthog-core.ts
turnipdabeets: updated

packages/core/src/types.ts
turnipdabeets: Went with `disableRemoteFeatureFlags`. Seems like it was named for handling more originally.

.changeset/rn-flags-disable-and-update-flags.md
turnipdabeets: good call, updated

packages/node/references/posthog-node-references-5.37.1.json
turnipdabeets: removed

turnipdabeets: Hm tried dropping all, but ran into a snag where the `check:public-api` job (looks like due to https://github.com/PostHog/posthog-js/pull/3834 and https://github.com/PostHog/posthog-js/pull/3849) regenerates references on every PR and fails if the committed -latest.json files are stale. So if I drop the node/RN -latest changes too, "Generate references" goes red.

packages/core/src/posthog-core.ts
```diff
@@ -1142,6 +1176,67 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
+          key,
+          enabled: getEnabledFromValue(value),
+          variant: getVariantFromValue(value),
+          reason: undefined,
+          // Locally supplied flags have no server-side id/version
+          metadata: {
+            id: undefined,
+            version: undefined,
+            description: undefined,
+            payload: payload !== undefined ? JSON.stringify(payload) : undefined,
```
ioannisj: I think this could potentially throw and crash the host app? We may want to try-catch serializing payload here and log if there's an error instead?

packages/core/src/types.ts
```diff
@@ -53,6 +53,24 @@ export type PostHogCoreOptions = {
...
+   * related methods keep working against those values. Until `updateFlags()` runs, reads
+   * return their not-loaded defaults, so use `bootstrap` for any flags needed at startup.
+   * Equivalent to the web SDK's `advanced_disable_feature_flags`.
+   *
+   * Note: surveys gated on feature flags will not evaluate unless the survey targeting
+   * flags are also provided via `updateFlags()`. This option cannot be toggled at runtime.
+   *
+   * @default false
+   */
+  disableRemoteFeatureFlags?: boolean
```
ioannisj, line 74: nit: I think this will be picked up by posthog-node so we probably need an Omit there or just update the changeset saying it's a no-op?

ioannisj (commented):
LG for me, waiting for Dustin to validate changes on his review

packages/core/src/types.ts
turnipdabeets: Good point about Omit. I noticed there's about 8 of these in node that should be omitted and was thinking of doing that, but it'll technically be breaking changes (can do this in a separate PR too). Do you prefer omitting just this one or think it's ok to omit them all? Or think it's fine to have a no-op like the others? 

For now I added a comment that it's no-op since it matches pattern of the others.

packages/core/src/posthog-core.ts
turnipdabeets: good catch, updated

packages/node/references/posthog-node-references-5.37.1.json
dustinbyrne: ah okay, this looks better now though, at least it's changing `*-latest` vs `*-5.37.1`

packages/core/src/posthog-core.ts
```diff
@@ -178,7 +180,14 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         }
       }
 
-      this.reloadFeatureFlags()
+      if (this.disableRemoteFeatureFlags && !this.disabled) {
+        // No reload runs to emit the change, so emit the now-cleared flags directly
+        // (drives onFeatureFlags listeners, e.g. session replay re-arm). Callers re-push
+        // the new identity's flags via updateFlags().
+        this.setKnownFeatureFlagDetails({ flags: {} })
```
dustinbyrne: we should validate `FeatureFlagDetails` is not in `allPropertiesToKeep` before clearing it here (and test this case)

turnipdabeets: good catch, addressed

dustinbyrne (approved):
🚀

PR #3854, opened 2026-06-15
feat(flags): add configurable flags_request_path

### LINKED ISSUE #3840: uBlock origin started blocking /flags

Hi there is experimental feature in uBlock origin where they block /flags endpoint. Do you have suggestions how to bypass this?

https://github.com/uBlockOrigin/uAssets/commit/4a771f2422058f19915c988358ed21105d148708

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

haacked:
## Problem

uBlock Origin added a filter that blocks the `/flags` request path on any domain. It matches the `/flags/` substring anywhere in the URL, not just on `*.posthog.com`. Because the match is on the path, the usual reverse-proxy workaround doesn't help: a proxy changes the host, but the SDK still requests `…/flags/?v=2`, so the request is still blocked. The flags host is already configurable via `flags_api_host`, but the `/flags/` path is hardcoded.

Fixes #3840

## Changes

Adds a `flags_request_path` config option (default `/flags/`) used when building the flags request URL, so the path can be moved off `/flags`.

```js
posthog.init('<token>', {
    api_host: 'https://your-proxy.example.com',
    flags_request_path: '/api/eaH8aiyu/', // your proxy routes this to PostHog's /flags
})
```

New optional `flags_request_path` on `PostHogConfig` (`@posthog/types`), defaulting to `/flags/` in `defaultConfig`. Both flag request call sites in `posthog-featureflags.ts` (the main flag load and `getRemoteConfigPayload`) use the configured path and fall back to `/flags/`. It is backwards compatible: unset or default behaves exactly as before, and the request is byte-identical to `…/flags/?v=2…`.

This is the SDK half of the fix. It immediately helps anyone running their own reverse proxy who can add a rewrite rule mapping the custom path to PostHog's `/flags`. It does not by itself fix PostHog Cloud or the managed reverse proxy, which only rewrite the host and not the path. That needs a companion server-side change to accept flag requests on an alternate path, and this PR is the prerequisite for it.

One constraint worth noting: the new path must not contain the literal `flags` substring, since the filter matches it anywhere. So `/api/<random>/` works but `/api/flags/` would not.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/.changeset/flags-configurable-request-path.md
+++ b/.changeset/flags-configurable-request-path.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+Add a `flags_request_path` config option to route feature flag requests via a custom path instead of the default `/flags`. Some ad blockers block the `/flags` path on any domain, which a reverse proxy alone doesn't fix because it only changes the host. Point `flags_request_path` at a path your reverse proxy maps to PostHog's flags endpoint to avoid the block. Defaults to `/flags/`, so existing behavior is unchanged.
--- a/packages/browser/src/__tests__/featureflags.test.ts
+++ b/packages/browser/src/__tests__/featureflags.test.ts
@@ -1049,6 +1049,24 @@ describe('featureflags', () => {
             expect(instance._send_request.mock.calls[0][0].data.disable_flags).toBe(undefined)
         })
 
+        it('requests the default /flags path when flags_request_path is not set', () => {
+            featureFlags.reloadFeatureFlags()
+            jest.runOnlyPendingTimers()
+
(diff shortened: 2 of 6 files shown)


haacked:
Ignore this. Claude got carried away. :cry:

PR #3855, opened 2026-06-15
fix(flags): drop ip query param from flags requests

### LINKED ISSUE #3840: uBlock origin started blocking /flags

haacked:
## Problem

uBlock Origin added a filter, `/flags/*^ip=`, that blocks PostHog flag evaluation on any domain. It matches two things anywhere in the request URL: the `/flags/` path and an `ip=` query param. The SDK appends `ip=0` to every request in `_send_request`, so flag requests look like `/flags/?v=2&ip=0`, which the filter matches. A reverse proxy doesn't help, because the filter keys on the path and param rather than the host.

Fixes #3840

## Changes

The `ip` query param is vestigial on the flags endpoint. The server derives the client IP from the connection and honors the `$geoip_disable` property in the body, so this query param has no effect on flag evaluation. Removing it from flag requests breaks the filter match with no behavior change.

- Add a `skipIPParam` option to the internal request options.
- `_send_request` skips appending `ip` when it is set.
- Both flag request paths set it: the main flag load and `getRemoteConfigPayload`.
- Event and session recording requests are unchanged and still send `ip`.

One note on durability: this defeats the filter as written today, which requires the `ip=` token. If uBlock later drops that condition and matches `/flags/` alone, a path-level change would be needed too. This is the smallest fix that unblocks flag evaluation now, including on PostHog Cloud, with no proxy or server-side change required.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/references/posthog-js-references-1.386.8.json
+++ b/packages/browser/references/posthog-js-references-1.386.8.json
@@ -3526,6 +3526,10 @@
           "name": "noRetries"
         },
         {
+          "type": "boolean",
+          "name": "skipIPParam"
+        },
+        {
           "type": "number",
           "name": "timeout"
         },
--- a/packages/browser/references/posthog-js-references-1.386.8.json
+++ b/packages/browser/references/posthog-js-references-1.386.8.json
@@ -3526,10 +3526,6 @@
           "name": "noRetries"
         },
         {
-          "type": "boolean",
-          "name": "skipIPParam"
-        },
-        {
           "type": "number",
           "name": "timeout"
         },
--- a/.changeset/flags-drop-ip-param.md
+++ b/.changeset/flags-drop-ip-param.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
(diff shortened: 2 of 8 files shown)


packages/browser/references/posthog-js-references-1.386.8.json
```diff
@@ -3525,6 +3525,10 @@
           "type": "boolean",
           "name": "noRetries"
         },
+        {
+          "type": "boolean",
+          "name": "skipIPParam"
+        },
```
marandaneto: this should not be there since this version was already released

marandaneto: https://github.com/PostHog/posthog-js/pull/3858

marandaneto (approved):

PR #3861, opened 2026-06-16
feat: add exception steps to react native error tracking

ioannisj:
## Problem

The React Native SDK has no way to record breadcrumb-style context leading up to an
exception. The browser, iOS, and Android SDKs already support **exception steps** —
context records that accumulate over time and attach to every captured `$exception`
as `$exception_steps`, giving the error tracking UI a timeline of what happened right
before the error. This brings RN to parity.

```js
posthog.addExceptionStep('User tapped Checkout', { screen: 'cart' })
```

## Changes

- **`addExceptionStep(message, properties?)`** public API on the RN `PostHog` class.
- **Managed (JS) layer** — reuses `@posthog/core`'s `ExceptionStepsBuffer` (FIFO,
  UTF-8 byte-budget eviction, oldest evicted first, oversized step rejected).
  `$timestamp` is captured at call time; reserved keys (`$message`/`$timestamp`) are
  stripped; recording never throws.
- **Config** — `errorTracking.exceptionSteps` with `enabled` (default `true`) and
  `maxBytes` (default `32768`).
- **Capture wiring** — buffered steps attach to a `$exception` only when the caller
  didn't already provide `$exception_steps`. The buffer is a rolling window scoped to
  the SDK instance: it persists across captures and identity changes, and clears on
  `_shutdown` (a clean launch starts empty).
- **Hybrid / native forwarding** — when native crash capture is enabled, each step is
  forwarded through `@posthog/react-native-plugin` to the embedded native SDK (iOS
  Swift + Android Kotlin bridge), so native crashes carry the same timeline. The
  native byte budget is kept in sync via the plugin's `setup` config.

### Testing  
- unit tests for the managed-layer buffer/API (ordering, validation,
reserved-key stripping, disabled no-op, byte-budget eviction, oversized rejection,
clear, config mapping) and capture-path tests (attach order, rolling window across
captures, no-overwrite, disabled). Full RN suite: 438 passing.
- Manual testing on sample projects

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes


packages/react-native/src/posthog-rn.ts
```diff
@@ -1676,6 +1681,39 @@ export class PostHog extends PostHogCore {
...
+   *
+   * @param {string} message A non-empty description of the step
+   * @param {Object} [properties] Optional additional context to attach to the step
+   * @returns {void}
+   */
+  addExceptionStep(message: string, properties?: PostHogEventProperties): void {
+    this._errorTracking.addExceptionStep(message, properties)
+    // Mirror the step to the embedded native SDK so native crashes carry the same steps.
+    this._forwardExceptionStepToNative(message, properties)
+  }
```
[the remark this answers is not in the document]
ioannisj: Done, even though native already guards this

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -99,6 +102,14 @@ class PosthogReactNativePluginModule(
               flushAt = theFlushAt
               errorTrackingConfig.autoCapture = nativeErrorTrackingAutocapture
 
+              // Keep the native exception-steps buffer aligned with the JS layer (one logical buffer).
+              if (hasKey(exceptionStepsConfig, "enabled")) {
+                errorTrackingConfig.exceptionSteps.enabled = getBoolean(exceptionStepsConfig, "enabled", true)
+              }
+              if (hasKey(exceptionStepsConfig, "maxBytes")) {
+                errorTrackingConfig.exceptionSteps.maxBytes = getInt(exceptionStepsConfig, "maxBytes", errorTrackingConfig.exceptionSteps.maxBytes)
+              }
```
marandaneto: if the 3rd param of gerBoolean and getInt are the default values, we dont need a `hasKey` check unless it throws

packages/react-native/src/error-tracking/index.ts
```diff
@@ -19,8 +28,25 @@ interface AutocaptureOptions {
...
+ */
+export interface ExceptionStepsOptions {
+  /**
+   * Whether exception steps are recorded and attached. Defaults to `true`.
+   */
+  enabled?: boolean
+  /**
+   * Total UTF-8 byte budget for the in-memory buffer. Oldest steps are evicted first when exceeded.
+   * Defaults to `32768` (~32KB).
+   */
```
marandaneto, line 46: use jsdoc style for defaults eg @default false

packages/react-native/src/posthog-rn.ts
```diff
@@ -1733,6 +1738,51 @@ export class PostHog extends PostHogCore {
...
+   * @example
+   * ```js
+   * posthog.addExceptionStep('User tapped Checkout', { screen: 'cart' })
+   * ```
+   *
+   * @param {string} message A non-empty description of the step
+   * @param {Object} [properties] Optional additional context to attach to the step
+   * @returns {void}
+   */
+  addExceptionStep(message: string, properties?: PostHogEventProperties): void {
```
marandaneto: do we use the PostHogEventProperties type for non event type properties? at least the obj name feels strange as addExceptionStep param

packages/react-native/src/posthog-rn.ts
```diff
@@ -1733,6 +1738,51 @@ export class PostHog extends PostHogCore {
...
+
+  private _forwardExceptionStepToNative(message: string, properties?: PostHogEventProperties): void {
+    if (!this._nativeErrorTrackingInitialized || !OptionalReactNativePlugin?.addExceptionStep) {
+      return
+    }
+    // Fire-and-forget: the native layer validates and buffers independently and must never block.
+    void Promise.resolve(OptionalReactNativePlugin.addExceptionStep(message, properties)).catch((e) => {
+      this._logger.warn(`Failed to forward exception step to native: ${e}`)
+    })
+  }
```
marandaneto: could these be moved within the _errorTracking class to keep this one cleaner?

marandaneto:
side note for the plugin, we should refactor/rename the decideReplayConfig flag name for the plugin since we dont even have a decide call anymore, rename to remoteConfig?

marandaneto (approved):
left a few suggestions but LGTM anyway

packages/react-native/src/error-tracking/index.ts
ioannisj: Done. Just a note, this mangles a little bit the generated references since `@default` is stripped from the comment

```
"description": "Total UTF-8 byte budget (~32KB) for the in-memory buffer. Oldest steps are evicted first when exceeded.   32768"
```

but I kept it to be consistent with the rest of the codebase

packages/react-native/src/posthog-rn.ts
ioannisj: Good call here, even though practically the same type the object did not make sense. I copied the signature that web sdk uses

packages/react-native/src/posthog-rn.ts
ioannisj: Done

PR #3865, opened 2026-06-16
feat(logs): migrate web programmatic logs to the shared @posthog/core pipeline

turnipdabeets:
## Problem

The browser had two separate logging implementations: the programmatic API (`posthog.captureLog` / `posthog.logger.*`), hand-rolled in the browser, and console auto-capture (`captureConsoleLogs`) on OpenTelemetry. The programmatic path re-implemented batching/retry that already exists in `@posthog/core` (the pipeline React Native uses).

## Changes

Routes the **programmatic** web logs API (`posthog.captureLog` / `posthog.logger.*`) through the shared `@posthog/core` logs pipeline — **no public API change, same behavior**. Web now gets the same resilience RN already has:

- adaptive 413 batch-splitting,
- per-record retry with exponential-backoff re-arm,
- an `online`-event reconnect flush,
- a FIFO eviction cap (`maxQueueSize`) decoupled from the flush trigger (`maxBufferSize`, default 100 unchanged). On web it's sized to the per-interval rate cap, so a burst the cap admits is held and delivered rather than evicted. This isn't free — it holds up to the rate cap's worth of records in memory during a burst — but **web memory is cheap and there's no cellular cost, so we favor delivery completeness**. (RN evicts sooner, tuned for mobile memory/cellular budgets — an intentional per-platform difference.) To bound web memory, lower `maxLogsPerInterval`.

Console auto-capture is untouched (still OTel). This is the first step toward consolidating both web logging paths into core and removing the OpenTelemetry dependencies.

> **`logs.beforeSend` (the user-facing filter) is split into a stacked follow-up: #3869.** This PR is the plumbing only.

## Compatibility notes (for reviewers)

Existing manual-logs behavior is preserved — `service.name` default (`unknown_service`), the per-interval rate cap, delivered volume, opt-out gating, and the `sendBeacon`-on-unload drain. Two benign shifts from the async core pipeline:

- **`maxBufferSize` keeps its default (100) and meaning** (the flush trigger). Records leave the queue only after the server acks, so during a synchronous burst the queue can grow past `maxBufferSize`; the eviction backstop (`maxQueueSize`, sized to the rate cap on web) drops the oldest only above that. Net: bursts the rate cap admits are still delivered in full, matching the pre-PR behavior — just held in memory during the burst rather than flushed synchronously as they arrive.
- **Unload:** because records leave the queue only after an ack, an unload mid-flight re-sends that batch — favoring a possible duplicate over the previous risk of losing an in-flight batch the navigation cancelled.

RN/node are unaffected — core changes are additive (`maxQueueSize` defaults to `maxBufferSize` when unset, so RN stays byte-identical). `BeforeSendLogFn` is relocated to `@posthog/types` (core re-exports it — non-breaking).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] @posthog/types

(`@posthog/core` is also bumped — additive, backwards-compatible.)

## Checklist

- [x] Tests for new code

### If releasing new changes

---

> [!NOTE]
> Migrates the browser's programmatic logs API (`posthog.captureLog` / `posthog.logger.*`) to use the shared `@posthog/core` `PostHogLogs` pipeline that React Native already uses, gaining adaptive 413 batch-splitting, exponential-backoff retry, online-event reconnect flush, and a FIFO eviction cap decoupled from the flush trigger. Also adds `fireCallbackOnDrop` to `_send_request` so dropped requests surface as retries instead of silent stalls, and wires `posthog.reset()` to clear buffered logs.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit ae1ff413622f82d2fce8dbf1a145c957e4c27b4f.</sup>


turnipdabeets:
Thanks @greptile-apps — both addressed in d8e972d6.

<details>
<summary>✅ Silent drop when <code>beforeSend</code> empties the body — fixed</summary>

The empty-body drop in `packages/core/src/logs/index.ts` now emits the same `info` diagnostic as the null-return path (`"Log was rejected in beforeSend function"`), so both `beforeSend` drop paths are equally observable. Added a regression test (`logs the same info line when a fn empties the body`).
</details>

<details>
<summary>✅ In-flight flush vs core rebuild on <code>set_config</code> — accepted trade-off, now documented</summary>

Yes, the concurrent `_persistQueueAdvance` race is an accepted trade-off. As you noted, the FIFO advance makes the worst case a **duplicate** head batch, never a loss — the same dup-over-loss stance used on the unload path. The trigger is narrow: a `set_config({ logs })` call with a *new* object reference that lands while a flush is in flight. I've updated the `_getCore()` comment so it covers the in-flight async path explicitly, not just the armed-timer path.
</details>

<sub>Also declared `posthog-react-native: patch` in the changeset for the regenerated references — clears the changeset-hygiene note.</sub>

.changeset/web-logs-before-send.md
```diff
@@ -0,0 +1,25 @@
...
+            }
+            return log
+        },
+    },
+})
+```
+
+`beforeSend` accepts a single function or an array of functions (applied left to right); returning `null` from any of them drops the record. It runs for logs sent via both `posthog.captureLog()` and `posthog.logger.*`.
+
+This release also makes log delivery more resilient: oversized batches are split automatically, failed sends retry with exponential backoff, and delivery resumes when the browser comes back online.
```
dustinbyrne: can this be two PRs? 😬 

the `beforeSend` enhancement seems like it'd be a pretty straightforward approval, but it's getting lost in here

turnipdabeets: I could separate it, but it's here because it's kind of getting it for free by doing this refactor.

turnipdabeets: moved it here: https://github.com/PostHog/posthog-js/pull/3869 but it depends on this PR to merge first.

packages/core/src/logs/index.ts
```diff
@@ -186,6 +219,8 @@ export class PostHogLogs {
 
   private async _flushInner(): Promise<void> {
     this._clearFlushTimer()
+    // Count only evictions from here on; the read below already reflects prior ones.
+    this._evictedSinceAdvance = 0
```
dustinbyrne: if we send multiple batches in the loop below, a log message captured in between iterations (due to the `await`) can evict from the queue and this eviction accounting would be thrown off

we should move this into the loop

packages/browser/src/posthog-logs.ts
```diff
@@ -49,14 +95,39 @@ export class PostHogLogs implements Extension {
     }
 
     reset(): void {
-        this._logBuffer = []
-        if (this._flushTimeout) {
-            clearTimeout(this._flushTimeout)
-            this._flushTimeout = undefined
+        this._queue = []
+        this._core?.reset()
```
dustinbyrne, line 99: should `posthog.reset()` call `logs.reset()`?

dustinbyrne (approved):
the `_evictedSinceAdvance` accounting issue below is worth addressing before merge

packages/core/src/logs/index.ts
turnipdabeets: Good catch — fixed in ae1ff413. Moved the `_evictedSinceAdvance` reset into the loop (per batch), so an eviction during the persist-await between iterations is no longer counted against the next batch's advance (which could under-drop and re-send a record). Added a multi-batch regression test that reproduces it.

packages/browser/src/posthog-logs.ts
turnipdabeets: Yes — added in ae1ff413. `posthog.reset()` now calls `this.logs?.reset()` alongside the other extension resets, so buffered log records are dropped on reset/logout. Added a test asserting the wiring.

PR #3868, opened 2026-06-16
fix(replay): scope flushed-size tracker to the session

pauldambra:
## Problem

The session recorder writes `$sdk_debug_replay_flushed_size` ("total bytes flushed to PostHog") on every event, intended as a per-session figure. It isn't one.

It was stored as a single bare number in `posthog.persistence` (localStorage/cookie) — device-global and shared across tabs, persisted across page loads — and was only ever cleared by `reset()` inside `_onSessionIdCallback`, i.e. on an *in-page* session rotation. But the recorder constructor pre-sets the current session id, so the immediate `onSessionId` emission on a fresh load hits the early-return guard and skips `reset()`. Net effect: the counter leaks across page loads and tabs and keeps growing, so short sessions on a returning visitor report hundreds of MB "flushed".

Querying live data made this concrete: a single device showed four distinct session ids all reading ~939–940 MB with no reset, and 2–9 second recordings reporting ~950 MB flushed against a few MB actually stored. This blocks using the value as a per-session byte budget / recording cutoff.

## Changes

Make the tracker genuinely session-scoped instead of relying on the fragile `reset()`-on-rotation path:

- Store `{ sessionId, size }` keyed to the current session. A flush for a different session overwrites the stored entry rather than adding to it.
- `currentTrackedSize(sessionId)` returns the stored size only when the persisted session id matches, otherwise `0`. A fresh load adopting an ongoing session reads the correct running total; a new session starts at `0`; a legacy bare-number value (from older SDKs) reads as `0`.
- `reset()` is removed — session-scoping makes it redundant, which also trims a little bundle.
- Three call sites in `lazy-loaded-session-recorder.ts` updated to pass the session id (`snapshotBuffer.sessionId` on flush, `oldSessionId` for the `$session_ending` event, `this.sessionId` for the debug props).

The emitted value stays in the same units (uncompressed serialized bytes, matching `$snapshot_bytes` and ClickHouse `size`).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Paul directed this: started from a data investigation comparing the SDK's self-reported flushed bytes against server-side stored size, which surfaced the cross-session leak, then asked to verify the cause in the SDK source and fix the tracking before considering a recording cutoff.

Decisions along the way:
- Confirmed in source that the leak is the persistence store (device-global, cross-tab) plus the constructor/early-return path skipping `reset()` on fresh loads — not a missing reset call.
- Chose keying the stored value to the session id over patching the reset path, since correctness then no longer depends on observing a session transition. This also let us delete `reset()` (smaller bundle).
- Stored an object `{ sessionId, size }` rather than two persistence keys to keep the change minimal.
- Known limitation left in place: multiple tabs of the same session do a read-modify-write on one localStorage key, so an increment can occasionally be lost — acceptable for a coarse byte budget.

Verification (run locally, not manual browser testing): `flushed-size-tracker.test.ts` 12 pass (added session-scoping + legacy-value cases), full `lazy-sessionrecording.test.ts` suite 229 pass, eslint clean on the changed files.

---


packages/browser/src/extensions/replay/external/flushed-size-tracker.ts
```diff
@@ -16,17 +20,15 @@ export class FlushedSizeTracker {
...
 
-    trackSize(size: number) {
-        const currentFlushed = Number(this._getProperty(SESSION_RECORDING_FLUSHED_SIZE)) || 0
-        const newValue = currentFlushed + size
-        this._setProperty(SESSION_RECORDING_FLUSHED_SIZE, newValue)
-    }
-
-    reset() {
-        return this._setProperty(SESSION_RECORDING_FLUSHED_SIZE, 0)
+    trackSize(sessionId: string, size: number) {
```
pauldambra, line 27: > [!NOTE]

**[convergent: qa-team/reliability + paul]** 🟠 HIGH

`trackSize` does a read-modify-write (`currentTrackedSize(sessionId) + size`) against a single shared localStorage/cookie key, and `currentTrackedSize` reads the in-memory `persistence.props` (loaded once per tab), not a fresh storage read. Two tabs recording the **same** session each hold an independent total and clobber each other on save — lost increments, net undercount. The old device-global counter had the same race, so this is **not a regression** and is fine for a passive debug property. But the stated goal is to drive a per-session byte budget / recording cutoff on top of this value — at that point an undercount makes the cutoff fire late or never (fail-open). Not blocking for this PR; flagging so the race is closed (re-read storage before incrementing, storage-event merge, or per-tab counters summed server-side) **before** anything gates on it. A `// best-effort, not safe to gate on yet` comment now would set expectations.

packages/browser/src/extensions/replay/external/flushed-size-tracker.ts
```diff
@@ -1,10 +1,14 @@
...
 const SESSION_RECORDING_FLUSHED_SIZE = '$sess_rec_flush_size'
 
+interface FlushedSize {
+    sessionId: string
+    size: number
+}
+
 export class FlushedSizeTracker {
-    private readonly _getProperty: (property_name: string) => Property | undefined
+    private readonly _getProperty: (property_name: string) => any
```
pauldambra: > [!NOTE]

**[convergent: paul + xp + qa-team/compatibility]** 🟡 MEDIUM

`_getProperty` was widened from `Property | undefined` to `any`, and then re-narrowed by assertion (`as FlushedSize | undefined`) in `currentTrackedSize`. That gives up compiler help right where the value is now a union of legacy-number and new-object shapes. Prefer `unknown` (forcing a deliberate narrow) or an explicit `FlushedSize | number | undefined`, and do the narrowing in one place. `any` tends to leak.

packages/browser/src/extensions/replay/external/flushed-size-tracker.ts
```diff
@@ -16,17 +20,15 @@ export class FlushedSizeTracker {
...
+        this._setProperty(SESSION_RECORDING_FLUSHED_SIZE, {
+            sessionId,
+            size: this.currentTrackedSize(sessionId) + size,
+        })
     }
 
-    get currentTrackedSize(): number {
-        return Number(this._getProperty(SESSION_RECORDING_FLUSHED_SIZE)) || 0
+    currentTrackedSize(sessionId: string): number {
+        const stored = this._getProperty(SESSION_RECORDING_FLUSHED_SIZE) as FlushedSize | undefined
```
pauldambra: > [!NOTE]

**[convergent: xp + paul]** 🟢 LOW

The legacy bare-number case is handled by coincidence: an old plain number has no `.sessionId`, so `stored?.sessionId === sessionId` is false and we fall through to `0`. That is the correct, safe behaviour (the old device-global number was garbage), and the test pins it — but the production code never *says* that's the intent. Consider an explicit shape guard (`typeof stored === 'object' && stored?.sessionId === sessionId`) so a future reader debugging migration sees the named branch rather than an accident of `===`.

packages/browser/src/extensions/replay/external/flushed-size-tracker.ts
```diff
@@ -1,10 +1,14 @@
 import { PostHog } from '../../../posthog-core'
-import { Property } from '../../../types'
 
 const SESSION_RECORDING_FLUSHED_SIZE = '$sess_rec_flush_size'
```
pauldambra: > [!NOTE]

**[qa-team/data-integrity]** 🟡 MEDIUM (verified)

`$sess_rec_flush_size` has no entry in `persistence-key-policy.ts`. In `PostHogPersistence.properties()` (posthog-persistence.ts:286) keys with no policy hit the `!policy` branch and are attached to **every captured event** as a super-property (`p[k] = v`, commented "Unknown keys are treated as user-defined super properties and remain event-visible"). So this internal counter already rides on every event for every SDK user — and this PR changes that emitted value from a number to a nested `{ sessionId, size }` object, i.e. a schema change to a property present on all events. Recommend promoting the key to a shared constant and adding a policy entry that keeps it out of the event stream (an `exposure: 'derived'` entry with no `transformToEventProperties` is skipped from event props). Not sensitive (session id is already on events), but unintended noise + a scalar→object change on the wire.

pauldambra (commented):
> [!NOTE]

Multi-perspective review (qa-team specialists, paul-reviewer, xp-reviewer). security-audit skipped (MCP unavailable; no security surface in this diff). See inline comments.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists), paul-reviewer, xp-reviewer. security-audit skipped (MCP unavailable; no auth/SQL/SSRF/IDOR surface in this diff).

The core fix is correct and a genuine improvement: keying the flushed-size total to the session id removes the cross-page-load / cross-tab over-counting and deletes the fragile `reset()`-on-rotation path that fresh page loads silently skipped. All three call sites pass the right session id (the `_flushBuffer` switch to `snapshotBuffer.sessionId` is actually *more* correct than before), legacy bare-number values degrade safely to `0`, and the new tests pin the never-tracked / no-leak / new-session-from-zero / legacy cases.

### Key findings

- 🟠 **HIGH (non-blocking, pre-existing)** — multi-tab read-modify-write race on the single shared `$sess_rec_flush_size` key undercounts when two tabs record the same session. Harmless for a debug prop; a fail-open gap **once this value gates dropping recordings**. Close the race before the cutoff lands on top of it.
- 🟡 **MEDIUM (verified)** — `$sess_rec_flush_size` has no `persistence-key-policy` entry, so it ships on every captured event as a super-property; this PR changes that emitted value from a number to a nested object. Add a policy entry to keep it out of the event stream.
- 🟡 **MEDIUM** — `_getProperty: any` gives up type safety exactly where a shape-sensitive cast + legacy branch were added; prefer `unknown`/explicit union.
- 🟢 **LOW** — legacy bare-number handling works by coincidence of `===`; make the shape check explicit in prod code.
- 🟢 **LOW** — a final old-session split flushed *after* `$session_ending` emits can under-report `flushed_size` in that event (reporting inaccuracy, not a correctness bug).
- ⚪ **NIT** — in `flushed-size-tracker.test.ts`, the comment "storage only ever holds the current session's total" would read better as the test name.

- The multi-tab race was flagged independently by **qa-team/reliability** and **paul**.
- The `any` type downgrade was flagged independently by **paul**, **xp**, and **qa-team/compatibility**.
- The implicit legacy-value handling was flagged by **xp** and **paul**.

### Note (not a code finding)

The fix commit carries `Generated-By` / `Task-Id` trailers (mandated by the authoring tooling). qa-team flagged the internal `Task-Id` UUID as public-repo metadata — surfacing for the author's awareness; it is tooling-required, not an accidental leak.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Correct fix; approve with nits — close the multi-tab race and add a persistence-key policy before this gates recording drops. |
| 👤 paul | High trust, would stamp; real flag is the *next* PR — don't build a recording cutoff on a racy shared key, and confirm the debug prop still graphs the fix in prod. |
| 📐 xp | Satisfying simplification (data carries its own scope, redundant `reset()` gone); tighten the typing and make the legacy intent explicit. |
| 🛡 security-audit | Skipped — MCP unavailable; diff has no security surface. |

---

packages/browser/src/extensions/replay/external/flushed-size-tracker.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PR Shepherd** — relaying the author's decision, not written by a human

Accepted and resolving. For this PR the flushed-size counter only needs to be *relatively* correct, and the read-modify-write undercount is harmless for a passive debug property — the pre-existing device-global counter had the same race, so this is not a regression.

Noting the close for when this value gates a recording cutoff: at that point the undercount becomes fail-open, so we'll re-read the key before incrementing (`_refreshKeyFromStorage`), merge on the `storage` event, or use per-(session, window) counters summed server-side — rather than gating naively on a value that can lose increments.

arnohillen (approved):

PR #3872, opened 2026-06-17
refactor(flags): browser reuses core parsePayload

turnipdabeets:
## Problem

`posthog-js`'s `getFeatureFlagPayload` reimplements payload JSON parsing inline (a `try/catch` around `JSON.parse`), duplicating `@posthog/core`'s `parsePayload`. Duplicated logic like this is how the two flag implementations drift apart.

## Changes

- Export `parsePayload` from `@posthog/core`.
- Use it in `posthog-js`'s `getFeatureFlagPayload` instead of the inline parse block.

Pure refactor, **no behavior change** (core's `parsePayload` returns non-strings as-is, parses strings, and falls back on parse failure — equivalent to the inline logic). Full feature-flag suite still passing (171).

I vetted other dedup candidates and left them out on purpose: `getFeatureFlagValue` doesn't fit (the browser's `getFeatureFlagResult` returns `FeatureFlagResult`, not core's `FeatureFlagDetail`), and `normalizeFlagsResponse` / `id: 0` vs `undefined` / payload storage format have genuinely diverged — those are behavior reconciliations, not free swaps.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

- [ ] Ran `pnpm changeset` to generate a changeset file (n/a — internal refactor, no user-facing change)

### DIFF
--- a/packages/browser/src/posthog-featureflags.ts
+++ b/packages/browser/src/posthog-featureflags.ts
@@ -37,7 +37,7 @@ import {
     PERSISTENCE_OVERRIDE_FEATURE_FLAG_PAYLOADS,
 } from './constants'
 
-import { isUndefined, isArray, isNull } from '@posthog/core'
+import { isUndefined, isArray, isNull, parsePayload } from '@posthog/core'
 import { createLogger } from './utils/logger'
 import { getTimezone } from './utils/event-utils'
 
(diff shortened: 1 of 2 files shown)


turnipdabeets:
Folded into #3871 (same theme + both touch the core `featureFlagUtils` export line, so they belong together).

PR #3874, opened 2026-06-17
fix(react-native): support skip on conflict in expo plugin

### LINKED ISSUE #3867: Expo config-plugin: allow passing extra upload args (e.g. --skip-on-conflict) to the posthog-cli
Follow-up to #48179. That issue asked for two things; only one shipped:

1. ✅ `--skip-on-conflict` flag on the upload CLI — shipped in #59001 (posthog-cli **v0.7.12**, 2026-05-21).
2. ❌ A way for the **Expo / React Native config plugin** to pass that flag through to the CLI — **never built.** #48179 was closed when item 1 merged, so this half fell off the radar.

The CLI supports `--skip-on-conflict`, but there is currently **no way to enable it from the plugin**, because the plugin hardcodes the `hermes upload` arguments and exposes no passthrough.

## Why there's no workaround short of patching the package

All paths in `packages/react-native/`:

- **`PostHogPluginProps`** (`src/tooling/expoconfig.ts`) only accepts `disableSandboxing` and `uploadNativeSymbols` — no way to add upload args.
- **Android** (`tooling/posthog.gradle`): the `hermes upload` call only appends auto-detected `--release-name` / `--release-version` / `--build`.
- **iOS** (`tooling/posthog-xcode.sh`): the `hermes upload` line only appends the same auto-detected `$CLI_RELEASE_ARGS`.
- No environment-variable escape hatch exists (`grep` for `skip-on-conflict` / `extraUploadArgs` / `EXTRA*` → nothing).

So today a customer who needs `--skip-on-conflict` has to `patch-package` the installed plugin. The plugin side is pure passthrough wiring — the CLI already does the work.

## Proposed fix

Add a generic passthrough prop, plus a typed convenience alias for the common case:

```ts
type PostHogPluginProps = {
  disableSandboxing?: boolean
  uploadNativeSymbols?: boolean | { includeSource?: boolean }
  // NEW:
  extraUploadArgs?: string[]   // appended verbatim to `hermes upload` on both platforms
  skipOnConflict?: boolean     // convenience: adds `--skip-on-conflict` to extraUploadArgs
}
```

Wire the resolved args into both generated upload commands:
- `tooling/posthog.gradle` — append to the `hermes upload` `args` list (alongside the existing `extraArgs`).
- `tooling/posthog-xcode.sh` — append to the `hermes upload` line (alongside `$CLI_RELEASE_ARGS`).

Also **bump the CLI version floor to `0.7.12`** (currently `0.7.8` in `posthog-xcode.sh`, and unchecked in `posthog.gradle`) since `--skip-on-conflict` requires it; otherwise passing the flag fails with a confusing build error.

## Acceptance criteria

- [ ] `skipOnConflict: true` in `app.json`/`app.config` causes `--skip-on-conflict` to appear in the `hermes upload` invocation on **both** iOS and Android.
- [ ] `extraUploadArgs: ["--foo", "bar"]` is appended verbatim on both platforms.
- [ ] Min CLI version checks updated to `0.7.12` (or guarded so the flag isn't sent to an unsupported CLI).
- [ ] Docs / prop table updated.

_Refs: #48179, #59001._

marandaneto:
## Problem

The Expo config plugin can upload React Native Hermes sourcemaps, but it had no option to enable the CLI's `--skip-on-conflict` behavior for upload commands. Customers who hit duplicate sourcemap uploads had to patch the package locally.

## Changes

- Add a `skipOnConflict` Expo plugin option for `posthog-react-native`.
- Wire `skipOnConflict: true` into iOS sourcemap uploads by passing an internal flag to `posthog-xcode.sh`, which appends `--skip-on-conflict` only to `posthog-cli hermes upload`.
- Wire `skipOnConflict: true` into Android sourcemap uploads via `project.ext.posthogReactNativeSkipOnConflict` in `posthog.gradle`.
- Add focused tests for the new option.

Closes #3867.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with pi after investigating the linked GitHub issue. The implementation intentionally keeps scope to the typed `skipOnConflict` option only, rather than adding a generic upload-argument passthrough. Follow-up review feedback removed package README docs and CLI min-version checks from this PR, and kept JSON parsing scoped to the Xcode script branches that need parsed content. Verified with focused React Native tests and linting.

### DIFF
--- a/packages/react-native/README.md
+++ b/packages/react-native/README.md
@@ -8,6 +8,20 @@ SDK usage examples and code snippets live in the official documentation so they 
 
 - [React Native library docs](https://posthog.com/docs/libraries/react-native)
 
+## Expo config plugin options
+
+```json
+{
+  "expo": {
+    "plugins": [["posthog-react-native", { "skipOnConflict": true }]]
+  }
+}
+```
+
+| Option | Type | Description |
+| --- | --- | --- |
+| `skipOnConflict` | `boolean` | Appends `--skip-on-conflict` to `posthog-cli hermes upload` on iOS and Android. Requires `posthog-cli >= 0.7.12`. |
+
 ## Questions??
 
 ### [Check out our community page.](https://posthog.com/posts)
--- a/packages/react-native/README.md
+++ b/packages/react-native/README.md
@@ -8,20 +8,6 @@ SDK usage examples and code snippets live in the official documentation so they 
 
 - [React Native library docs](https://posthog.com/docs/libraries/react-native)
 
-## Expo config plugin options
-
-```json
-{
-  "expo": {
-    "plugins": [["posthog-react-native", { "skipOnConflict": true }]]
-  }
-}
-```
-
-| Option | Type | Description |
-| --- | --- | --- |
-| `skipOnConflict` | `boolean` | Appends `--skip-on-conflict` to `posthog-cli hermes upload` on iOS and Android. Requires `posthog-cli >= 0.7.12`. |
-
 ## Questions??
 
 ### [Check out our community page.](https://posthog.com/posts)
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -11,7 +11,16 @@ const POSTHOG_ANDROID_GRADLE_PLUGIN_VERSION = '1.2.0'
 const resolvePostHogReactNativePackageJsonPath =
   "[\"node\", \"--print\", \"require('path').join(require('path').dirname(require.resolve('posthog-react-native')), '..', 'tooling', 'posthog.gradle')\"].execute().text.trim()"
 
-const withAndroidPlugin = (config: any) => {
+const POSTHOG_ANDROID_SKIP_ON_CONFLICT_PROPERTY = 'posthogReactNativeSkipOnConflict'
+
+export function buildAndroidSkipOnConflictGradleLine(skipOnConflict: boolean): string | null {
+  if (!skipOnConflict) {
+    return null
+  }
+  return `project.ext.${POSTHOG_ANDROID_SKIP_ON_CONFLICT_PROPERTY} = true`
+}
+
+const withAndroidPlugin = (config: any, skipOnConflict = false) => {
   return withAppBuildGradle(config, (config: any) => {
     if (config.modResults.language !== 'groovy') {
       console.warn('Cannot configure PostHog in the app gradle because the build.gradle is not groovy')
@@ -19,8 +28,19 @@ const withAndroidPlugin = (config: any) => {
 
     const buildGradle = config.modResults.contents
     const applyFrom = `apply from: new File(${resolvePostHogReactNativePackageJsonPath})`
+    const skipOnConflictLine = buildAndroidSkipOnConflictGradleLine(skipOnConflict)
+    const applyBlock = skipOnConflictLine ? `${skipOnConflictLine}\n${applyFrom}` : applyFrom
+    const skipOnConflictPattern = new RegExp(
+      `^project\\.ext\\.${POSTHOG_ANDROID_SKIP_ON_CONFLICT_PROPERTY}\\s*=\\s*(true|false)\\n?`,
+      'm'
+    )
 
     if (buildGradle.includes(applyFrom)) {
+      let contents = buildGradle.replace(skipOnConflictPattern, '')
+      if (skipOnConflictLine) {
+        contents = contents.replace(applyFrom, `${skipOnConflictLine}\n${applyFrom}`)
+      }
+      config.modResults.contents = contents
       return config
     }
 
@@ -28,7 +48,7 @@ const withAndroidPlugin = (config: any) => {
     const pattern = /^android\s*\{/m
 
     if (buildGradle.match(pattern)) {
-      config.modResults.contents = buildGradle.replace(pattern, `${applyFrom}\n\nandroid {`)
+      config.modResults.contents = buildGradle.replace(pattern, `${applyBlock}\n\nandroid {`)
     } else {
       console.warn('PostHog: Could not find "android {" block in build.gradle')
     }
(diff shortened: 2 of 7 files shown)


packages/react-native/README.md
```diff
@@ -8,6 +8,20 @@ SDK usage examples and code snippets live in the official documentation so they
 
 - [React Native library docs](https://posthog.com/docs/libraries/react-native)
 
+## Expo config plugin options
```
marandaneto: skip this, we will document in the docs, rollback this file

packages/react-native/tooling/posthog.gradle
```diff
@@ -109,6 +115,7 @@ plugins.withId('com.android.application') {
 
                         doFirst {
                             def cliPackage = PostHogCli.resolveCliPackagePath(rootDirFile, reactRoot)
+                            PostHogCli.assertMinCliVersion(cliPackage, osCompatibility, reactRoot)
```
marandaneto: lets remove this, we will mention min version in the docs

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -85,7 +102,7 @@ if [ -z "$PH_CLI_PATH" ] || [ ! -x "$PH_CLI_PATH" ]; then
   exit 1
 fi
 
-MIN_POSTHOG_CLI_VERSION="0.7.8"
+MIN_POSTHOG_CLI_VERSION="0.7.12"
```
marandaneto: lets remove the min version check to simplify this

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -277,6 +310,11 @@ type PostHogPluginProps = {
    * Requires `posthog-cli` to be available and authenticated during release builds.
    */
   uploadNativeSymbols?: boolean | { includeSource?: boolean }
+
+  /**
+   * Whether to append `--skip-on-conflict` to `posthog-cli hermes upload` on iOS and Android.
```
marandaneto, line 315: mention the default value here if not set

packages/react-native/README.md
marandaneto: Rolled back the README changes in 15f3b9f4.

packages/react-native/tooling/posthog.gradle
marandaneto: Removed the Android CLI min-version assertion and helper in 15f3b9f4.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Removed the iOS CLI min-version check from this PR in 15f3b9f4.

packages/react-native/src/tooling/expoconfig.ts
marandaneto: Added the default value to the prop docs in 15f3b9f4.

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -85,7 +102,7 @@ if [ -z "$PH_CLI_PATH" ] || [ ! -x "$PH_CLI_PATH" ]; then
   exit 1
 fi
```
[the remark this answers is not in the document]
marandaneto: Addressed in 15f3b9f4 by removing the iOS CLI min-version change from this PR.

packages/react-native/tooling/posthog.gradle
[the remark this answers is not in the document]
marandaneto: Addressed in 15f3b9f4 by removing the unconditional Android CLI min-version assertion.

packages/react-native/test/expoconfig.spec.ts
```diff
@@ -248,6 +276,16 @@ describe('resolveNativeSymbolUpload', () => {
   })
 })
 
+describe('buildAndroidSkipOnConflictGradleLine', () => {
+  it('returns no Gradle line when skipOnConflict is disabled', () => {
+    expect(buildAndroidSkipOnConflictGradleLine(false)).toBeNull()
+  })
+
+  it('serializes skipOnConflict for posthog.gradle', () => {
+    expect(buildAndroidSkipOnConflictGradleLine(true)).toBe('project.ext.posthogReactNativeSkipOnConflict = true')
```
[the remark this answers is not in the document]
marandaneto: Updated in 15f3b9f4 to use an it.each table for this helper.

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -85,16 +102,6 @@ if [ -z "$PH_CLI_PATH" ] || [ ! -x "$PH_CLI_PATH" ]; then
...
 
-MIN_POSTHOG_CLI_VERSION="0.7.8"
```
marandaneto, line 88: our docs mention min version

marandaneto:
> - Expo plugin generates skipOnConflict wiring for both platforms.
>  - Android Release build succeeds.
>  - Android duplicate Hermes sourcemap upload is skipped.
>  - iOS Release build succeeds with code signing disabled.
>  - iOS Hermes upload receives --skip-on-conflict.
>  - iOS duplicate Hermes sourcemap upload is skipped.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -146,21 +166,23 @@ const withAndroidNativeSymbolsPlugin = (config: any) => {
 
 type BuildPhase = { shellScript: string }
 
-export function modifyExistingXcodeBuildScript(script: BuildPhase): void {
+export function modifyExistingXcodeBuildScript(script: BuildPhase, skipOnConflict = false): void {
   if (!script.shellScript.match(/(packager|scripts)\/react-native-xcode\.sh\b/)) {
     return
   }
 
+  const code = JSON.parse(script.shellScript)
```
ioannisj: This will now parse for `script.shellScript.includes('posthog-react-native')` branch which didn't before. Can't tell if this will break things but since branches are mutually exclusive, maybe it's better each branch does JSON.parse when needed

packages/react-native/tooling/posthog-xcode.sh
ioannisj: No strong opinion but I would not mind keeping this. Even though one more doc to maintain

ioannisj (approved):
Looks good but did not test it out

packages/react-native/tooling/posthog-xcode.sh
marandaneto: the CLI will fail with a missing flag, so it'd be easy to figure out its an outdated version, also this flag has existed for quite a while
i hate dupe, if this becomes an issue, we can roll it back

packages/react-native/src/tooling/expoconfig.ts
marandaneto: Good catch — fixed in 45b11b18 by only parsing inside the branches that need the parsed script, and added a regression test to ensure the existing posthog-react-native early-return path does not parse.

PR #3879, opened 2026-06-17
feat: deprecate disableRemoteConfig and promote experimental config options to GA

ioannisj:
## Problem

Several React Native / core config options were marked "Experimental support" but have long since stabilized. `disableRemoteConfig` in particular is now redundant: remote config is always loaded, and the native mobile SDKs already deprecated their equivalent toggle and made it a no-op. Keeping these as "experimental" misrepresents their maturity.

## Changes

- **Deprecate `disableRemoteConfig`** (`@posthog/core` types) and make it a **no-op** in `posthog-react-native` — remote config is always loaded. Removed the now-unreachable "remote config disabled" survey-loading paths (`_flagsAsyncWithSurveys`, `_loadSurveysFromAPI`).
- **Promote `disableSurveys`** (`@posthog/core`) and **`maskAllSandboxedViews`** (`posthog-react-native`) to GA — dropped the "Experimental support" note only; behavior unchanged.
- Left `screenshotModeBackgroundCapture` experimental (still experimental in the native SDK).
- Updated two tests that used `disableRemoteConfig: true` purely as setup convenience.

> [!NOTE]
> Setting `disableRemoteConfig: true` is now a no-op — remote config always loads. The option remains accepted (no API break) and will be removed in a future major version.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

---

> [!NOTE]
> Deprecates `disableRemoteConfig` (now a no-op), removes the fallback survey-loading paths that ran when remote config was disabled, and promotes `disableSurveys` and `maskAllSandboxedViews` from experimental to GA.
> 
> <sup>Written by [Mendral](https://mendral.com) for commit 789ba96b8790a4c9126c9f6f168d0e825517348b.</sup>


packages/react-native/src/types.ts
```diff
@@ -138,7 +138,6 @@ export type PostHogSessionReplayConfig = {
    * Enable masking of all sandboxed system views
    * These may include UIImagePickerController, PHPickerViewController and CNContactPickerViewController
    * iOS only
-   * Experimental support
```
marandaneto, line 141: worth checking the plugin and ios sdk, maybe they also have experimental tags

marandaneto:
update docs if they are mentioned as experimental as well

marandaneto (approved):

packages/react-native/src/types.ts
ioannisj: I think we are good, iOS sdk already checked and I check plugin just now. No changes needed

ioannisj:
Looks like this is breaking tests from https://github.com/PostHog/posthog-js/pull/3848, so may be breaking that feature as well. Will pull back to draft and have a look tomorrow

ioannisj:
@turnipdabeets since you worked on https://github.com/PostHog/posthog-js/pull/3848 recently, these are the failing tests

- `PostHog RN disableRemoteFeatureFlags and updateFlags › with remote config disabled (the flags endpoint also carries config/surveys) › caches surveys from the config request but does not store its flags`
- `PostHog RN disableRemoteFeatureFlags and updateFlags › with remote config disabled (the flags endpoint also carries config/surveys) › makes a single config request with disable_flags and never refetches on identify/reset`
- `PostHog RN session replay re-arm after flags reload › arms recording from updateFlags when the SDK never fetches flags (disableRemoteFeatureFlags)`

Since we are deprecating `disableRemoteConfig` option, this means that the purely offline scenarios (no net calls on startup) are gone. Would it be okay to just rewrite these tests to keep CI honest, or should we remove the deprecation you think and just make the option GA? Trying to understand the impact of this change to `disableRemoteFeatureFlags` option

ioannisj:
@turnipdabeets ping on this? https://github.com/PostHog/posthog-js/pull/3879#issuecomment-4738028529

turnipdabeets:
> @turnipdabeets since you worked on #3848 recently, these are the failing tests

Those tests were just pinning the wire path for the disableRemoteConfig + disableRemoteFeatureFlags combo — the actual capability is unaffected. Go ahead and rewrite them.

PR #3882, opened 2026-06-17
feat: add browser-common package

dustinbyrne:
## Problem

Looking forward to a major version release of posthog-js, we must acknowledge that many users will remain on v1 for some time. As a result, we need a path forward to support both the current and next releases for the time being. Much of the code is shared, such as the persistence layer and extensions (session replay, surveys, conversations, etc.). Many extensions are already asynchronous. 

Duplicating these and maintaining fixes across two code bases is not tenable. As a result, we move extensions into their own package, and provide interfaces that allow incompatible major versions to adapt to support these extensions.

> [!NOTE]
> This change is purely additive, and doesn't affect any released SDKs. I'm opting to merge this now, as is, to cut down on the size of future PRs.

## Changes

- Adds a new source-only `@posthog/browser-extensions` workspace package.
- Defines the shared `Extension` / `Client` contract, lifecycle/disposable types, persistence/logging interfaces, extension tokens, and remote/session/event types.
- Adds a concrete `Publisher` helper for extension-owned event streams, with Node test coverage for publish/dispose/re-entrant listener behavior.
- Adds package README and an `author-extension` skill documenting extension authoring and v1 porting guidance.
- Configures modern TypeScript type-checking only (`build` runs `tsc --noEmit`); consumers are responsible for bundling/transpiling source.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

No changeset: this adds a private/source-only workspace package that is not being released yet.


packages/browser-common/src/logger.ts
```diff
@@ -0,0 +1,15 @@
+/**
+ * Minimal logging surface the client exposes to extensions. Warnings and errors
+ * are always emitted; `info` and `debug` are expected to be gated behind the
+ * client's debug setting (the extension does not decide that).
+ */
+export interface Logger {
```
marandaneto: not sure if this will cause a circular dep but if not we can move this to core and reuse from core, maybe core already has it btw

packages/browser-common/src/persistence.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * Async key-value store for small extension state. Backed by whatever the
+ * client provides — a synchronous store resolves immediately, an asynchronous
+ * one (e.g. IndexedDB) does real I/O — so reads and writes are always awaited.
+ *
+ * The store is namespaced to the client instance, so keys are local to the
+ * extension and never collide with core SDK state. Values must be
+ * JSON-serializable; setting `null`/`undefined` removes the key.
+ */
+export interface KeyValueStore {
```
marandaneto, line 10: i think core has this already, similar to https://github.com/PostHog/posthog-js/pull/3882/changes#r3457723600

marandaneto:
i like the idea, this is not going to be published, right?

marandaneto (approved):

packages/browser-common/package.json
```diff
@@ -0,0 +1,39 @@
...
+    },
+    "dependencies": {
+        "@posthog/core": "workspace:^"
+    },
+    "devDependencies": {
+        "@posthog-tooling/tsconfig-base": "workspace:*",
+        "@types/jest": "catalog:",
+        "jest": "catalog:",
+        "ts-jest": "catalog:",
+        "typescript": "^6.0.3"
```
turnipdabeets, line 37: I think other workspace package uses `"typescript": "catalog:"` (catalog pins 5.8.2). Hardcoding `^6.0.3` here pulls a second major TS version into the lockfile (plus a duplicate ts-node tree). Intentional signal for the v2 direction, or should this be a catalog entry? Why `^6.0.3` specifically?

packages/browser-common/.agents/skills/develop-extension/SKILL.md
```diff
@@ -0,0 +1,208 @@
...
+
+- `name` — unique; used for de-duplication and diagnostics.
+- `setup(client)` — may be `async` so you can read async state (kv, remote config) before the extension is ready; the
+  host awaits it.
+- `dispose()` — may be `async` (final flush, etc.); the host awaits it on teardown.
+- Extensions should define their own static configuration derived from the SDK's initialization configuration, and
+  accept it via the **constructor**. This keeps the configuration contract explicit and avoids the need for handling
+  generic configuration types.
+- Once `setup()` runs, the extension will not be called again until `dispose()`. Use `setup()` to initialize listeners
+  to react to changes changes.
```
turnipdabeets: typo: "changes changes".

packages/browser-common/package.json
dustinbyrne: `6.0.3` is the latest stable release. I'll probably take on upgrading the rest of the repo to v6 as a follow up.

packages/browser-common/src/logger.ts
dustinbyrne: good call, I switched to the core logger interface

packages/browser-common/src/persistence.ts
dustinbyrne: I didn't find a clear analog like `Logger` had. Closest matches were `PostHogCoreStateless.getPersistedProperty/setPersistedProperty` and `CookieStore`, but neither really fit

dustinbyrne:
> i like the idea, this is not going to be published, right?

that's right, it's private and won't even require it's own build or bundling step

PR #3884, opened 2026-06-17
feat(replay): rotate recording session at a configurable size budget

pauldambra:
## Problem

Per-session replay size has an extreme tail: web p99.9 is ~338 MiB, ~0.06% of web sessions exceed 500 MiB, and the largest run into the tens of GiB. A small number of unbounded sessions hold a large share of replay storage, strain the player, and push ingestion limits. We want to bound any single recording without losing data or interrupting recording.

## Changes

Adds size-based **session rotation**. When a recording session's flushed bytes reach a configurable budget, the SDK rotates to a new, linked session rather than stopping — so no single recording exceeds the budget, but the recording continues and nothing is dropped.

- **New option `session_recording.maxSessionSizeMb`** — defaults to `300` under the `2026-05-30` defaults bundle (off otherwise), clamped to the **1–500 MiB** range. It carries a comment noting it is normally only changed after interaction with the PostHog support team.
- **Rotation is linked, not a reset** — a new `sessionMaximumSize` change reason drives `SessionIdManager.rotateSessionForReplaySize()`, which mints a fresh session and fires the existing `$session_ending` / `$session_starting` linking events (so the player can stitch the timeline back together). Deliberately *not* `resetSessionId()`, which reads downstream as a `posthog.reset()` and would suppress linking.
- **A `$session_size_rotation` custom event** marks the point in the recording where the budget was reached (`{ sizeBytes, thresholdBytes }`).
- Built on the session-scoped flushed-size tracker that shipped in #3868. The budget check runs after the buffer flush + clear, so the rotation events land in the fresh buffer rather than being discarded.

Units are MiB (matching the server-side `size` column and the SDK's `$sdk_debug_replay_flushed_size`).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Paul directed this, sized the cutoff against cross-team replay data (US + EU, via Metabase), and chose the parameters: a **300 MiB** default with a **1–500 MiB** range, rotation (not hard-stop, so customers keep their data), a session-ended event, and an in-recording custom event marking the rotation point. The threshold and behaviour were chosen because rotation neither saves storage nor meaningfully changes revenue at this scale — its purpose is bounding per-session size, so the value is a guardrail tunable by support rather than a pricing lever.

Decisions made along the way:
- Rotate via a dedicated `sessionMaximumSize` change reason rather than reusing `resetSessionId()`, because the latter suppresses the `$session_ending` linking the player needs.
- Run the budget check *after* `_clearBuffer()` so the rotation custom events aren't discarded with the flushed buffer.
- Default gated behind the `2026-05-30` defaults bundle (off for existing integrations until they opt in), matching how `split_storage` / `persistence_save_debounce_ms` shipped.
- Remote-config override deferred to a follow-up (needs a backend change); this PR is local-config + bundle default only.

Verification (run locally, not manual browser testing): 353 tests pass across the recorder, sessionid, config, and flushed-size-tracker suites — including new coverage for the 300 default, the clamp, the rotation method minting a linked (non-reset) session, and the recorder emitting `$session_size_rotation` + rotating at the budget. Lint clean; `@posthog/types` typechecks.

Note: the underlying session-scoped flushed-size tracker merged separately in #3868; this PR is the rotation feature on top of it.

---

### DIFF
--- a/.changeset/replay-size-rotation.md
+++ b/.changeset/replay-size-rotation.md
@@ -1,0 +1,9 @@
+---
+'posthog-js': minor
+---
+
+feat(replay): rotate the recording session when it reaches a size budget
+
+Adds a `session_recording.maxSessionSizeMb` option that rotates a recording to a new, linked session once the session's flushed size reaches the budget. It defaults to `300` when using the `2026-05-30` defaults (off otherwise) and is clamped to the 1–500 MiB range; it is normally only changed after interaction with the PostHog support team.
+
+The previous session is linked to the new one via the existing `$session_ending` / `$session_starting` events, and a `$session_size_rotation` custom event marks the point at which the rotation happened. This bounds any single recording so it stays playable and within ingestion limits, without dropping data or interrupting recording.
--- a/packages/browser/src/__tests__/config.test.ts
+++ b/packages/browser/src/__tests__/config.test.ts
@@ -34,6 +34,22 @@ describe('config', () => {
                 content_ignorelist: DEFAULT_CONTENT_IGNORELIST_WITH_STEPPERS,
                 ignore_text_selection: true,
             })
+            expect(posthog.config.session_recording).toStrictEqual({
+                strictMinimumDuration: true,
+                maxSessionSizeMb: 300,
+            })
+        })
+
+        it('does not set maxSessionSizeMb before the 2026-05-30 defaults', () => {
+            const posthog = new PostHog()
+            posthog._init('test-token', { defaults: '2025-11-30' })
+            expect(posthog.config.session_recording.maxSessionSizeMb).toBeUndefined()
+        })
+
+        it('lets an explicit maxSessionSizeMb override the date-gated default', () => {
+            const posthog = new PostHog()
+            posthog._init('test-token', { defaults: '2026-05-30', session_recording: { maxSessionSizeMb: 100 } })
+            expect(posthog.config.session_recording.maxSessionSizeMb).toEqual(100)
         })
 
         it('merges a partial rageclick object with the date-gated defaults', () => {
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -4096,6 +4096,72 @@ describe('Lazy SessionRecording', () => {
             expect(tryAddCustomEvent).toHaveBeenCalledTimes(3)
         })
 
+        it('emits session linking events on session maximum size', () => {
+            const tryAddCustomEvent = sessionRecording['_lazyLoadedSessionRecording']['_tryAddCustomEvent'] as any
+            tryAddCustomEvent.mockClear()
+
+            const newSessionId = 'new-session-id-size'
+            const newWindowId = 'new-window-id-size'
+
+            sessionManager['_sessionIdChangedHandlers'].forEach((handler) => {
+                handler(newSessionId, newWindowId, {
+                    noSessionId: false,
+                    activityTimeout: false,
+                    sessionPastMaximumLength: false,
+                    sessionMaximumSize: true,
+                })
+            })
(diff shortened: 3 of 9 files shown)


pauldambra:
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists), paul-reviewer, xp-reviewer. security-audit skipped (no security surface in this diff).

Sound, well-scoped change — reuses the existing `$session_ending`/`$session_starting` linking and the change-reason vocabulary rather than inventing new machinery, and minting a linked session (instead of `resetSessionId()`, which would read as a `posthog.reset()` and suppress linking) is the correct approach. No correctness blocker for this PR.

### Actioned in `55f530cf`
- 🟢 `maxSessionSizeMb` now guarded with `Number.isFinite` (an `isNumber(NaN)` slipped past, working only by the downstream falsy check).
- 🟡 named the `1`/`500` MiB bounds + `BYTES_PER_MIB` (were bare magic numbers).
- 🟡 added `overshootBytes` to `$session_size_rotation` so the late-firing (from the multi-tab undercount) is observable in prod.
- 🟢 documented that rotation fires session-change handlers **synchronously** (must not flush, or it re-enters), and that `maxSessionSizeMb: 0` clamps to `1` rather than disabling.

### Deferred for team-replay's judgement
- 🟠 **OnceAndOnlyOnce (xp):** `rotateSessionForReplaySize()` re-spells the rotation core (`generate ids → set → reset idle → fire handlers`) that also lives in `checkAndGetSessionAndWindowId`. A shared `_rotateSession(changeReason)` would be cleaner — **deferred** because the existing path interleaves idle / cross-tab / `readOnly` / activity-timestamp logic, so a clean extraction touches the critical session-read hot path and deserves a maintainer's eye.
- 🟠 **Multi-tab double-rotation (qa-team):** two tabs of one session share the localStorage flushed-size counter, so both can cross the budget and each rotate — two `$session_starting` events, racing writes, and a window where the non-rotating tab emits under a stale id. This is the same shared-counter limitation already accepted as "relatively correct"; a proper fix needs cross-tab coordination (bigger scope). **Deferred** as a known limitation for a guardrail.
- 🟡 **`$session_size_rotation` attribution (qa-team):** the marker is added old-tagged into a buffer that's then re-tagged to the new session, so it may ingest under the new session. Minor (debug marker); worth confirming against the server-side capture path.
- 🟢 **Remote-config kill switch (qa-team):** no server-side override yet, so misbehaviour can only be mitigated by an SDK release. Already noted as a follow-up in the PR description.

- The **synchronous-handler / re-entrancy** concern was raised independently by **qa-team** and **paul** → addressed with a documenting comment (and it mirrors the existing rotation path, so it's not new risk).
- The **magic-number / bounds-naming** point was raised by **xp** and **paul** → actioned.

(Note: `paul` asked for a test that rotation does *not* suppress linking — already covered by `emits session linking events on session maximum size`.)

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Design sound; single-tab re-entrancy/loop concerns don't hold. Multi-tab double-rotation + marker attribution worth attention; rest is hardening. |
| 👤 paul | High trust; worries are coupling + observability, not correctness — wanted the overshoot graphable (done) and the load-bearing ordering explained. |
| 📐 xp | Follows the grain; the one real pull is the OnceAndOnlyOnce duplication with the existing rotation core — extract a shared minter. |

---

pauldambra:
i think we can get away without this if recent other changes are effective enough

PR #3885, opened 2026-06-18
feat(replay): capture canvas frames at reduced resolution

pauldambra:
## Problem

Canvas recordings (`record_canvas: true`) capture the canvas at full display resolution, 4fps, as WebP, with no ceiling. For animation/video canvases this is the dominant cost in a session's serialized bytes — we've seen canvas-heavy orgs produce enormous sessions where canvas frames are the overwhelming majority of the payload. There was no lever to trade a little fidelity for a lot of bytes, and crucially playback fidelity must not regress.

There's also a reliability angle: a single very large canvas frame can push a recording past our **infra ingest cap of 64 MB compressed**, at which point the recording can't be ingested at all. Reducing capture resolution shrinks the worst-case frame as well as the average.

## Changes

Adds `session_recording.canvasCapture.resolutionScale` — a `(0, 1]` fraction of the canvas display size to capture replay frames at. The captured bitmap is downscaled (pixel-area savings are quadratic) while the canvas's **true display size is still recorded**, so replay stretches the smaller frame back to the correct dimensions and aspect ratio — only sharpness drops, never layout.

```ts
posthog.init('...', {
  session_recording: {
    canvasCapture: { resolutionScale: 0.6 }, // (0,1]; 1 = full res = today's behaviour
  },
})
```

- Defaults to `1` (full resolution, unchanged from today). The latest `defaults` bundle (`2026-05-30`) opts new installs into `0.6`; older bundles are untouched.
- The captured display size is tracked on the **main thread** (not echoed from the encode worker), so downscaling can never mislabel a canvas's dimensions.
- At full resolution the capture path is byte-for-byte unchanged — the `'medium'` resampling hint is only applied when actually downscaling.

This started as a broader adaptive spike (varying fps + quality + resolution against per-session byte thresholds) and was deliberately simplified down to **resolution only, static**. Resolution is the highest-leverage lever (quadratic in scale) at the lowest perceptual cost, so dropping the fps/quality levers and the threshold engine removed a lot of surface area for almost none of the saving. Measured ~44% smaller encoded frames at `0.5` on a synthetic fixture, with playback dimensions provably preserved.

### Validation

A playwright test records a canvas whose backing store (2000×1500) differs from its CSS display size (1000×750) — the "is it 1000 or 2000?" case — drives the real `canvasCapture` init config across two recordings (full res vs `0.5`), and inspects the actual rrweb `$snapshot_data`: the recorded `drawImage` destination stays **1000×750 at both scales** while the encoded frame shrinks meaningfully.

### Explicitly out of scope (known follow-ups)

Two other places we capture pixels as data URIs are **not** touched here, by choice:

1. **The initial canvas full-snapshot** (`rrweb-snapshot` `canvas.toDataURL()`). At a ~5-minute full-snapshot interval vs 4fps incremental frames, the ratio is roughly **1:1200**, so shrinking it saves a fraction of a percent of canvas bytes. It would also require a matching replayer change (the rebuild path draws the frame at its intrinsic size rather than stretching to the canvas dimensions). **Caveat:** this initial frame captures the backing store at full resolution and is *not* size-capped, so for a very large canvas it's the frame most likely to breach the 64 MB compressed ingest cap. Worth revisiting specifically as an ingest-reliability fix, not a byte-efficiency one.
2. **Oversized inline `` images.** Today any inlined data-URI image over 1 MB is dropped entirely (replaced with a placeholder SVG) rather than shrunk. A downscale-before-cap there would *rescue* those images. It's niche (and PostHog doesn't enable `inlineImages` at all today), so it's a separate potential change.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

(plus `@posthog/rrweb` patch for the canvas observer resolution option)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with PostHog Code, directed by Paul. Began as an investigation into why canvas-heavy recordings are so large, root-caused to full-resolution 4fps capture with no ceiling, prototyped as a broad adaptive config, then deliberately cut back to a single static `resolutionScale` once it was clear resolution carries almost all of the saving at the least fidelity risk.

Key decisions:
- Default `1` everywhere except the `2026-05-30` defaults bundle (which opts into `0.6`), so existing installs are byte-for-byte unchanged.
- Static scale set once at record start — no mid-session rrweb reconfigure surface to maintain.
- Display size tracked main-thread, validated end-to-end by a playwright test that drives the real config path (an earlier worker-echo approach mislabeled dimensions and was caught by exactly this test).
- The two other data-URI capture sites (initial canvas snapshot, oversized inline images) are documented above as deliberate non-goals, with the 64 MB ingest-cap caveat called out.

Opt-in/low-risk but requires human review — do not self-merge.

---


packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -277,19 +316,28 @@ export class CanvasManager {
...
+            // display size, so playback dimensions/aspect are unchanged, just softer.
+            const scale = this._captureConfig!.scale;
+            const captureWidth = Math.max(1, Math.round(displayWidth * scale));
+            const captureHeight = Math.max(1, Math.round(displayHeight * scale));
             const bitmap = await createImageBitmap(canvas, {
-              resizeWidth: width,
-              resizeHeight: height,
+              resizeWidth: captureWidth,
+              resizeHeight: captureHeight,
+              resizeQuality: 'medium',
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + xp]** 🟡 MEDIUM

`resizeQuality: 'medium'` is now passed to `createImageBitmap` on **every** canvas frame, including the default `scale: 1` path. Previously the call used the browser default (`'low'`). This is the one change in the PR that alters behaviour for canvas recordings that have **not** opted into `canvasCapture` — any customer with `captureCanvas.recordCanvas` enabled gets a different (higher-cost) per-frame resampling filter even though `varyResolution` defaults to 1, which escapes the "opt-in, defaults to prod" framing. Consider only setting `resizeQuality: 'medium'` when `scale < 1`, leaving the full-res path byte-for-byte identical to today. (xp also flags it as a hardcoded fidelity knob that appears with no rationale — a one-word why would help the next reader.)

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -88,6 +88,25 @@ const DEFAULT_CANVAS_QUALITY = 0.4
...
+const CANVAS_VARY_THRESHOLD_MAX_MB = 1024
+const DEFAULT_CANVAS_VARY_THRESHOLDS_MB = [50, 150, 300]
+const MIN_CANVAS_VARY_FPS = 1
+const CANVAS_QUALITY_STEP = 0.05
+const CANVAS_VARIED_QUALITY_FLOOR = 0.25
+// capture-resolution scale per crossed threshold (index === level) used when varyResolution is
+// `true`. aspect ratio is preserved and replay upscales back to display size, so playback
+// dimensions are unchanged, just softer. never 1 (opting into stepping means never full res),
+// floored at 0.5 (half each dimension ~= a quarter of the pixels) to keep it legible.
+const CANVAS_RESOLUTION_SCALE_BY_LEVEL = [0.75, 0.6, 0.5, 0.5]
```
pauldambra: > [!NOTE]

**[convergent: xp + paul]** 🟡 MEDIUM

`CANVAS_RESOLUTION_SCALE_BY_LEVEL = [0.75, 0.6, 0.5, 0.5]` has 4 entries indexed by `level` (0..`CANVAS_VARY_THRESHOLD_COUNT`). The coupling between the threshold count (3) and the ladder length (4) is silent and load-bearing: bump `CANVAS_VARY_THRESHOLD_COUNT` to 4 and `level` can reach 4, `ladder[4]` is `undefined`, and `reconfigureCanvas({ scale: undefined })` survives only because rrweb's `scale > 0` guard happens to reject it — luck, not design. Clamp with `CANVAS_RESOLUTION_SCALE_BY_LEVEL[Math.min(level, len - 1)]`, or derive the ladder length from the threshold count so they can't drift. The duplicated trailing `0.5` is the seam where this hides — paul read it as a copy-paste typo; a one-line comment or the clamp would say the quiet part out loud.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -554,6 +573,97 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            targetFps === this._lastAppliedCanvasFps &&
+            targetQuality === this._lastAppliedCanvasQuality &&
+            targetScale === this._lastAppliedCanvasScale
+        ) {
+            return
+        }
+        this._lastAppliedCanvasFps = targetFps
+        this._lastAppliedCanvasQuality = targetQuality
+        this._lastAppliedCanvasScale = targetScale
+        getRRWebRecord()?.reconfigureCanvas?.({ fps: targetFps, quality: targetQuality, scale: targetScale })
```
pauldambra: > [!NOTE]

**[paul]** 🟡 MEDIUM

we silently degrade recorded canvas fidelity here with no breadcrumb in the recording. from prod we can't tell how often varying kicks in, at what level, or what it's costing in sharpness. `_tryAddCustomEvent` is right there (already used for `$session_options`, idle, pause) — emit one when the level changes (e.g. `$canvas_capture_varied` `{level, fps, quality, scale}`) so we can graph adoption/cost and correlate with any playback complaints. measure-before-you-limit — i'd really want this in before we vary anyone's real recording, even though the levers ship off by default. not a correctness blocker.

packages/browser/playwright/mocked/session-recording/canvas-resolution.spec.ts
```diff
@@ -0,0 +1,125 @@
...
+        })
+
+        // phase 1: default (full resolution, scale 1)
+        await page.resetCapturedEvents()
+        await recordCanvasActivity(page, 3)
+        const fullRes = await latestCanvasFrame(page)
+
+        // phase 2: downscale live to half resolution
+        await page.evaluate(() => {
+            ;(window as any).__PosthogExtensions__.rrweb.record.reconfigureCanvas({ scale: 0.5 })
```
pauldambra: > [!NOTE]

**[convergent: xp + paul]** 🟡 MEDIUM

this drives `record.reconfigureCanvas({ scale: 0.5 })` directly, so the rrweb downscale plumbing — the "does downscaling mislabel dimensions?" question — is covered end-to-end (good, that's the right test). but the adaptive config chain (`canvasCapture` + `thresholdsMb` -> `_maybeVaryCanvasCapture` -> level -> `reconfigureCanvas`) is never joined e2e: jest mocks one side and the spec hand-rolls the other. consider one playwright case that sets `canvasCapture: { varyResolution: true, thresholdsMb: [1, 2, 3] }`, drives enough flushed bytes to cross a threshold, and asserts `base64Len` shrinks — proving the config actually wires through before this stops being experimental.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -159,6 +186,10 @@ export class CanvasManager {
       true,
     );
     const snapshotInProgressMap: Map<number, boolean> = new Map();
+    // the display (replay) size per canvas id, kept on the main thread so a downscaled capture
+    // is still recorded at — and stretched back to — its real display size. never relies on the
+    // worker echoing it across the thread boundary.
+    const displayDimsMap: Map<number, { width: number; height: number }> = new Map();
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + paul]** 🟢 LOW

`displayDimsMap` grows one entry per canvas id and is never pruned when a canvas is removed from the DOM. it mirrors the pre-existing `snapshotInProgressMap` right beside it (same lifetime), so it is **not** a regression — but this PR now adds a second per-id map that leaks on canvas teardown. low severity (entries are tiny `{width, height}` and the canvas-id space is bounded in practice); flagging so it isn't read as deliberately-managed when it just leaks like its neighbour.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -554,6 +573,97 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            return CANVAS_RESOLUTION_SCALE_BY_LEVEL[level]
+        }
+        return 1
+    }
+
+    // adaptively step canvas capture fidelity down as the session accumulates bytes, gated by
+    // session_recording.canvasCapture.{varyFps,varyQuality,varyResolution}. keyed on the
+    // per-session flushed byte counter, so it resets to the configured fidelity when a new
+    // session starts.
+    private _maybeVaryCanvasCapture(): void {
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

`_maybeVaryCanvasCapture` braids guard + level computation + three-target resolution + change-guard + apply. the three levers rightly stay separate (resist deduplicating them — they compute differently, and a flag-driven merge would be the wrong abstraction). but the change-guard is accidental complexity: three parallel `_lastAppliedCanvasFps/Quality/Scale` fields + a three-way `===`. extract a `CanvasCaptureTarget { fps, quality, scale }` value object with a single `_lastAppliedCanvasTarget` and a shallow-equals — collapses three fields to one, names the concept, and a pure `(canvas, config, sessionBytes) -> CanvasCaptureTarget` helper becomes the unit-test seam, killing the `recorder['_maybeVaryCanvasCapture']()` private-poking in the jest suite.

packages/types/src/posthog-config.ts
```diff
@@ -563,6 +563,39 @@ export interface SessionRecordingOptions {
...
+     *   successive degradation step kicks in. Must be exactly three values, strictly increasing,
+     *   each between 1 and 1024 MiB — anything else is ignored and the defaults ([50, 150, 300])
+     *   are used. Lower these in tests to exercise the behaviour without recording huge sessions.
+     *
+     * Only has any effect when canvas recording is enabled.
+     */
+    canvasCapture?: {
+        varyFps?: boolean
+        varyQuality?: boolean
+        varyResolution?: boolean | number
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

`varyResolution: boolean | number` is two ideas under one name (fixed-scale vs stepped ladder), needing several doc sub-bullets to explain one field — the Primitive-Obsession / flag-parameter tell. the `false` case is also dead (indistinguishable from `undefined`). if the fixed-scale form only exists to ease testing, that's speculative generality — `varyResolution: boolean` + the ladder would be enough until product asks for a fixed scale. judgment call, not a blocker on an experimental flag.

packages/types/src/posthog-config.ts
```diff
@@ -563,6 +563,39 @@ export interface SessionRecordingOptions {
...
+     * so heavy canvas recordings (e.g. animation/video) cost less without dropping data. Keyed
+     * on the per-session flushed-bytes counter; resets to full fidelity on a new session.
+     *
+     * All levers are off by default and intended for careful, opt-in rollout. Playback fidelity
+     * matters, so each step is gentle: at each crossed threshold fps drops by one (e.g. 4 -> 3 -> 2),
+     * webp quality drops slightly, and/or the captured resolution is scaled down.
+     *
+     * - `varyFps`: step the canvas snapshot fps down by one per crossed threshold.
+     * - `varyQuality`: step the webp encode quality down slightly per crossed threshold.
+     * - `varyResolution`: scale the captured frame resolution down (aspect ratio preserved; replay
```
pauldambra: > [!NOTE]

**[convergent: qa-team/data-integrity + paul]** 🟢 LOW

subtle, intended, but the sharpest corner of the API: `varyResolution: true` returns scale `0.75` at `level` 0 (zero bytes), so a brand-new session under `varyResolution: true` starts already downscaled and never sits at full resolution. the fps/quality levers do reset to base on a new session; resolution under `true` intentionally does not. worth a one-liner in this docstring spelling out "`true` => below full res from the first frame" so an integrator doesn't read "resets to full fidelity on a new session" as "resets to scale 1".

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -38,6 +38,33 @@ export class CanvasManager {
   private rafIdFlush: number | null = null;
   private refCount = 0;
   private torndown = false;
+  // live fps/quality/scale for the FPS-snapshot canvas observer, so callers can
+  // adjust capture fidelity mid-session without restarting recording.
+  // scale (0,1] downsamples the captured frame; it is upscaled back to its display
+  // size on replay, so playback dimensions/aspect are unchanged, just softer.
+  // undefined until the FPS observer is initialised (sampling is a number).
```
pauldambra: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

the inline comment `// undefined until the FPS observer is initialised (sampling is a number).` reads as a sentence fragment — `(sampling is a number)` looks like a dangling condition. suggest `// undefined until the FPS observer is initialised, which only happens when sampling is a number`. minor; the comments are otherwise clear and explain the why.

pauldambra (commented):
> [!NOTE]
>
> Multi-perspective review: qa-team (specialists + generalists), paul-reviewer, xp-reviewer, security-audit. Inline comments below; summary in a separate top-level comment.

pauldambra:
> [!NOTE]
>

Clever, well-documented, genuinely opt-in and off-by-default — no correctness blockers, and the central risk (downscaling mislabeling canvas dimensions) is handled correctly and proven by the new playwright test. The findings are about observability, a latent fragility, test coverage of the config path, and design polish — none block merge for an experimental flag.

### Key findings

**🟡 Medium**
- **`resizeQuality: 'medium'` touches the non-opt-in path** (`canvas-manager.ts:329`) — applied on every frame including `scale: 1`, so all `recordCanvas` customers get a changed per-frame resampling filter even though `varyResolution` defaults to 1. Gate behind `scale < 1` to keep the full-res path identical to today. _(qa-team + xp)_
- **Silent fidelity degradation has no breadcrumb** (`lazy-loaded-session-recorder.ts:664`) — `_tryAddCustomEvent` exists; emit `$canvas_capture_varied {level,fps,quality,scale}` on level change so prod cost/adoption is graphable. Measure-before-you-limit. _(paul)_
- **Threshold-count vs ladder-length coupling is silent** (`lazy-loaded-session-recorder.ts:106`) — `level` can index past `CANVAS_RESOLUTION_SCALE_BY_LEVEL` if the threshold count grows; survives only by rrweb's `scale > 0` guard. Clamp or derive the length. _(xp + paul)_
- **Playwright test exercises raw `reconfigureCanvas`, not the config path** (`canvas-resolution.spec.ts:103`) — the adaptive chain (`canvasCapture` -> threshold -> level -> reconfigure) is never joined e2e. Add one case with tiny `thresholdsMb` proving config wires through. _(xp + paul)_

**🟢 Low / ⚪ Nit**
- `displayDimsMap` leaks on canvas teardown like its neighbour `snapshotInProgressMap` (not a regression). _(qa-team + paul)_
- `_maybeVaryCanvasCapture` change-guard wants a `CanvasCaptureTarget` value object (collapses 3 `_lastApplied*` fields, gives a clean unit-test seam). _(xp)_
- `varyResolution: boolean | number` is two ideas under one name; `false` is dead. _(xp)_
- Docstring should spell out that `varyResolution: true` starts below full res from the first frame (never scale 1 on a fresh session). _(qa-team + paul)_
- Confusing `(sampling is a number)` comment fragment. _(qa-team/copy)_

### Convergence

High-confidence findings flagged independently by 2+ reviewers: the **playwright-config-path gap** (xp + paul), the **ladder/threshold coupling** (xp + paul), the **`resizeQuality` non-opt-in change** (qa-team + xp), the **`displayDimsMap` leak** (qa-team + paul), and the **`varyResolution: true` doc nuance** (qa-team + paul).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Low blast radius, off-by-default; one finding (`resizeQuality: 'medium'`) leaks into the non-opt-in path and warrants gating; downscale correctness and CDN back-compat are sound. |
| 👤 paul | Stamping — lovely display-size/downscale trick, properly parameterized tests; wants an observability custom-event before varying any real recording. |
| 📐 xp | Solid make-it-right work; act on the threshold/ladder coupling and extract a `CanvasCaptureTarget`; add an end-to-end config-path test before de-experimentalizing. |
| 🛡 security-audit | 0 findings — host-developer config in their own users' browsers; thresholds/scale validated and clamped before use, no tenant crossover, no prototype-pollution path. |

---

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -554,6 +558,18 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+    // (0,1] fraction of the canvas display size to capture frames at, from
+    // session_recording.canvasCapture.resolutionScale. clamped to a sane range; defaults to 1
+    // (full resolution, matching today's production) so capture only drops below full resolution
+    // when explicitly configured. replay upscales the frame back to its display size.
+    private get _canvasResolutionScale(): number {
+        const configured = this._instance.config.session_recording.canvasCapture?.resolutionScale
+        if (!isNumber(configured) || !Number.isFinite(configured)) {
+            return 1
+        }
+        return Math.min(1, Math.max(MIN_CANVAS_SCALE, configured))
```
pauldambra: > [!NOTE]

This hand-rolled clamp (`Math.min(1, Math.max(MIN_CANVAS_SCALE, configured))` + the `isNumber`/`Number.isFinite` guard) reinvents `clampToRange`, which this same file already uses two getters down for fps and quality — and which logs a warn when it clamps. The bespoke version is silent, so a fat-fingered `resolutionScale: 60` (meaning 0.6) is quietly coerced to 1 with no breadcrumb. xp verified all five jest cases stay green under `clampToRange(configured, MIN_CANVAS_SCALE, 1, createLogger('canvas recording resolution scale'), 1)` (it handles `Infinity` via `value > max -> max`, so the `Number.isFinite` special-case goes away too). OnceAndOnlyOnce — fold the three canvas knobs into one clamp idiom.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -105,10 +109,18 @@ export class CanvasManager {
       blockSelector,
       recordCanvas,
       dataURLOptions,
+      resolutionScale,
     } = options;
     this.mutationCb = options.mutationCb;
     this.mirror = options.mirror;
 
+    const scale =
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + xp]** 🟡 MEDIUM

Scale is validated twice with **divergent policies**: the SDK getter clamps to `[0.1, 1]` (MIN_CANVAS_SCALE floor), but this rrweb constructor accepts anything in `(0, 1]` (else 1) with no floor. A consumer using `@posthog/rrweb` directly with `resolutionScale: 0.001` sails straight past the degenerate-resolution protection the 0.1 floor exists to provide. The duplication itself is defensible (this is a separately-published package boundary — validity normalization, not SDK policy), but (a) say so in a comment so it doesn't read as accidental duplication, and (b) reconcile the range — enforce a small floor here too, or document that out-of-range at the rrweb layer means full res.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -168,16 +186,22 @@ export class CanvasManager {
...
+      const dw = display ? display.width : width;
+      const dh = display ? display.height : height;
       this.mutationCb({
         id,
         type: CanvasContext['2D'],
         commands: [
           {
             property: 'clearRect', // wipe canvas
-            args: [0, 0, width, height],
+            args: [0, 0, dw, dh],
```
pauldambra, line 205: > [!NOTE]

The changeset/JSDoc claim the full-resolution path is "byte-identical / unchanged from before" — that's overstated. Pre-feature the emitted command used the 3-arg `drawImage [image, 0, 0]`; this now always emits the 5-arg form `[image, 0, 0, dw, dh]` and `clearRect` uses display dims, for **all** canvas-recording users including those who never set `resolutionScale`. On replay it's pixel-equivalent at scale 1 (`dw === displayWidth === width`, and the replayer applies both arg forms to native `drawImage`), so playback is correct — but the serialized snapshot bytes differ for every existing canvas recording. Recommend softening the wording to "pixel-equivalent on replay" rather than "byte-identical."

packages/browser/src/posthog-core.ts
```diff
@@ -201,7 +201,12 @@ const defaultsThatVaryByConfig = (
               ? { content_ignorelist: true }
               : true,
     capture_pageview: defaults && defaults >= '2025-05-24' ? 'history_change' : true,
-    session_recording: defaults && defaults >= '2025-11-30' ? { strictMinimumDuration: true } : {},
+    session_recording:
+        defaults && defaults >= '2026-05-30'
+            ? { strictMinimumDuration: true, canvasCapture: { resolutionScale: 0.6 } }
```
pauldambra, line 206: > [!NOTE]

**[convergent: qa-team/data-integrity + paul]** 🟡 MEDIUM

The `2026-05-30` defaults bundle opts new installs into `resolutionScale: 0.6` — a real, visible fidelity reduction for canvas replays (text/line sharpness on charts and drawing apps). The rollout shape is right (defaults-bundle-gated, existing installs untouched, per-customer reversible, and it lands in `$session_options` so the cohort is measurable). Two asks before shipping: (1) confirm this is intended product behaviour — new installs no longer capture canvas at full resolution — and document it downstream; (2) is `0.6` a validated sweet spot or a first guess? Worth a sentence in the changeset either way.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -159,6 +173,10 @@ export class CanvasManager {
       true,
     );
     const snapshotInProgressMap: Map<number, boolean> = new Map();
+    // the display (replay) size per canvas id, kept on the main thread so a downscaled capture
+    // is still recorded at — and stretched back to — its real display size. never relies on the
+    // worker echoing it across the thread boundary.
+    const displayDimsMap: Map<number, { width: number; height: number }> = new Map();
```
pauldambra: > [!NOTE]

**[convergent: qa-team/performance + xp]** 🟢 LOW

`displayDimsMap` grows one entry per distinct canvas mirror id for the life of the FPS observer and is never pruned on canvas removal. It's bounded by canvas count within a session and GC'd when the CanvasManager is recreated on record restart, and it mirrors the pre-existing `snapshotInProgressMap` (same shape, also never cleared) — so not a regression. Flagging only because this adds a *second* per-id map; if you ever add a `resetObservers` cleanup for one, clear both together so they don't drift.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

Dead-ish local: `const scale` is computed in the constructor but only consumed in the `sampling === 'number'` (FPS-snapshot) branch — for the `sampling === 'all'` mutation-observer path it's computed and discarded, which reads as if scale applies to both observers. Consider computing it inside `initCanvasFPSObserver` from the raw `resolutionScale` so the FPS-only scope is structural rather than incidental, and that method owns the whole concern end-to-end.

packages/types/src/posthog-config.ts
```diff
@@ -563,6 +563,20 @@ export interface SessionRecordingOptions {
...
+     * Tune how canvas frames are captured for replay. Only has any effect when canvas recording
+     * is enabled.
+     *
+     * - `resolutionScale`: capture canvas frames at a fraction of their display resolution. A
+     *   number in `(0, 1]`; `1` is full resolution (the default, matching today's behaviour) and,
+     *   e.g., `0.6` captures at 60%. Aspect ratio is preserved and replay upscales the frame back
+     *   to the original display size, so playback dimensions are unchanged, just softer. Resolution
+     *   is the highest-leverage lever for canvas byte size, since bytes scale with pixel area.
+     */
+    canvasCapture?: {
```
pauldambra, line 577: > [!NOTE]

**[qa-team/frontend]** 🟢 LOW

`canvasCapture` is typed inline as `{ resolutionScale?: number }`, whereas the adjacent `captureCanvas` uses a named `SessionRecordingCanvasOptions`. A named interface (e.g. `SessionRecordingCanvasCaptureOptions`) would read more consistently, surface better in generated types/IntelliSense, and leave room for the future knobs the JSDoc hints at. Minor style alignment.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -226,10 +252,7 @@ export class CanvasManager {
     };
 
     const takeCanvasSnapshots = (timestamp: DOMHighResTimeStamp) => {
-      if (
-        lastSnapshotTime &&
-        timestamp - lastSnapshotTime < timeBetweenSnapshots
-      ) {
+      if (lastSnapshotTime && timestamp - lastSnapshotTime < 1000 / fps) {
```
pauldambra: > [!NOTE]

**[convergent: qa-team + paul]** ⚪ NIT

The refactor inlined `1000 / fps` into the rAF callback and dropped the named `timeBetweenSnapshots` const — recomputed every tick (negligible) but it lost a good name for no reason this feature needed, and it conflates incidental churn with the feature diff. Re-introducing the named binding would aid the next reader. (Pre-existing: `fps === 0` -> `Infinity` -> never snapshots, i.e. "off"; the inline form makes the divide-by-fps slightly less obvious.)

packages/browser/playwright/mocked/session-recording/canvas-resolution.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+    }) => {
+        // two sequential recordings on the same page: each start() navigates fresh and re-inits
+        // posthog with the given canvasCapture config (the later route registration wins).
+        // full resolution (resolutionScale unset -> 1)
+        const fullRes = await recordCanvasFrame(page, context, undefined)
+        // half resolution via the real init config
+        const halfRes = await recordCanvasFrame(page, context, 0.5)
+
+        // eslint-disable-next-line no-console
+        console.log('\n=== CANVAS RESOLUTION VALIDATION ===\n' + JSON.stringify({ fullRes, halfRes }, null, 2))
```
pauldambra: > [!NOTE]

**[convergent: qa-team + xp]** ⚪ NIT

Drop the leftover `console.log('=== CANVAS RESOLUTION VALIDATION ===' ...)` debug line (with its `eslint-disable no-console`) before merge — it's dev scaffolding, not a permanent assertion, and it's noise in CI output. The test itself targets exactly the right invariant.

packages/browser/playwright/mocked/session-recording/canvas-resolution.spec.ts
```diff
@@ -0,0 +1,132 @@
+import { expect, test, WindowWithPostHog } from '../utils/posthog-playwright-test-base'
+import { start, waitForSessionRecordingToStart } from '../utils/setup'
+import { BrowserContext, Page } from '@playwright/test'
+
+// Validates that capturing the canvas at a reduced resolution (session_recording.canvasCapture.
+// resolutionScale) does NOT mislabel dimensions: the recorded drawImage stretches the (smaller)
+// encoded frame back to the canvas's display size, so playback dimensions/aspect are unchanged -
+// only the encoded bytes shrink. Uses a canvas whose backing store (2000x1500) differs from its
+// CSS display size (1000x750), i.e. the "is it 1000 or 2000?" case, and drives the real init
+// config path (not a private method) across two independent contexts.
```
pauldambra: > [!NOTE]

**[paul]** ⚪ NIT

Stale header comment: it says the test runs "across two independent contexts," but the test actually does two sequential `start()`s on the same page+context (the in-test comment says so). Delete the "two contexts" claim so a future reader doesn't go hunting for a second context that isn't there. Also worth a one-liner that the canvas redraws random content each frame, so the `< fullRes * 0.7` assertion is a byte-size *trend*, not pixel identity.

packages/types/src/posthog-config.ts
```diff
@@ -563,6 +563,20 @@ export interface SessionRecordingOptions {
      */
     captureCanvas?: SessionRecordingCanvasOptions
 
+    /**
+     * Tune how canvas frames are captured for replay. Only has any effect when canvas recording
+     * is enabled.
+     *
+     * - `resolutionScale`: capture canvas frames at a fraction of their display resolution. A
+     *   number in `(0, 1]`; `1` is full resolution (the default, matching today's behaviour) and,
```
pauldambra: > [!NOTE]

**[qa-team/copy]** ⚪ NIT

Two small doc points: (1) "matching today's behaviour" is time-relative and will age poorly in committed API docs once `0.6` is the default for new installs — consider "matching full-resolution capture." (2) The type is `number`, which permits `0` and negatives, while the prose says `(0, 1]`; the runtime clamps, so this is documentation-only, but a reader trusting the type could pass `0`. (Also: the jest `it.each` has no case for `0` or a negative — both clamp to the `0.1` floor — cheap to add a row.)

pauldambra (commented):
> [!NOTE]
>
> Multi-perspective review (resolution-only rewrite): qa-team, paul-reviewer, xp-reviewer, security-audit. Inline comments below; summary in a separate top-level comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review of the resolution-only rewrite: qa-team (specialists + generalists), paul-reviewer, xp-reviewer, security-audit

No correctness blockers. The core data-integrity invariant — downscaling never mislabels canvas dimensions — is sound (display size tracked main-thread before the worker postMessage, validated end-to-end by the playwright test), the full-res path is pixel-equivalent to before, back-compat across the lazy-CDN boundary holds, and security-audit found nothing. The simplification (deleting the adaptive engine for one static knob) is the right call. Findings are polish + two "confirm intent" asks.

### Key findings

**🟡 Medium**
- **`clampToRange` dedup** (`lazy-loaded-session-recorder.ts:570`) — the hand-rolled silent clamp reinvents `clampToRange`, already used two getters down for fps/quality (and it logs a warn). Folding it in unifies the three knobs and restores the breadcrumb. _(paul + xp; xp verified all 5 jest cases stay green)_
- **Divergent clamp policies** (`canvas-manager.ts:117`) — SDK floors at 0.1, the rrweb constructor accepts `(0,1]` with no floor, so a direct rrweb caller passing `0.001` slips past the degenerate-resolution protection. Reconcile the range + document the boundary intent. _(qa-team + xp)_
- **"byte-identical" wording overstated** (`canvas-manager.ts:201`) — full-res now emits the 5-arg `drawImage` for all canvas users; pixel-equivalent on replay but not byte-identical. Soften the changeset/JSDoc wording. _(qa-team)_
- **2026-05-30 default `0.6`** (`posthog-core.ts:206`) — visibly softer canvas replays for new installs. Confirm it's intended + document; is `0.6` measured or a first guess? (observable via `$session_options`). _(qa-team + paul)_

**🟢 Low / ⚪ Nit**
- `displayDimsMap` unbounded per-session (consistent with `snapshotInProgressMap`). _(qa-team + xp)_
- `scale` computed in the constructor but only used on the FPS path — push it into `initCanvasFPSObserver`. _(xp)_
- `canvasCapture` typed inline vs a named interface like its neighbour `captureCanvas`. _(qa-team)_
- Inlined `1000 / fps` dropped the named `timeBetweenSnapshots` const. _(qa-team + paul)_
- Leftover debug `console.log` in the spec; stale "two independent contexts" header comment; time-relative "matching today's behaviour" JSDoc; no jest case for `0`/negative. _(qa-team + paul + xp)_

### Convergence

Independently flagged by 2+ reviewers: **clampToRange dedup** (paul + xp), **divergent clamp policy** (qa-team + xp), **2026-05-30 0.6 default** (qa-team + paul), **`displayDimsMap` growth** (qa-team + xp), **dropped `timeBetweenSnapshots` const** (qa-team + paul), **leftover debug log** (qa-team + xp).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Low-risk, opt-in, defaults to prod behaviour; main-thread display-size handling is sound and the replayer takes the 5-arg form; soften the "byte-identical" claim and confirm the 0.6 default. |
| 👤 paul | Likes it — right call dropping the adaptive engine for one visible knob; wants the silent clamp folded into `clampToRange` and to know whether `0.6` is measured or a guess. |
| 📐 xp | Good YAGNI simplification with no dead remnants; two OnceAndOnlyOnce asks (clampToRange, reconcile the double-clamp) and tests hit the right seams. |
| 🛡 security-audit | 0 findings — host-developer config in their own users' browsers, clamped at two layers, no DoS / prototype-pollution / cross-tenant path. |

---

pauldambra:
> [!NOTE]
> 🤖 Automated comment by **PR Shepherd (review-triage)** — not written by a human

Triaged the QA Swarm findings.

**Actioned**
- 🟡 **`clampToRange` dedup** (`lazy-loaded-session-recorder.ts`, pushed `eedfce0c`): adopted. _Correction to an earlier draft of this comment:_ I initially deferred this thinking `clampToRange(NaN)` would return `NaN` — that's wrong. PostHog's `isNumber` excludes `NaN` (`x === x` is false for NaN), so `clampToRange` returns the fallback for NaN/non-numbers and clamps `Infinity` to max. It's a true equivalence to the hand-rolled guard, and it logs a warn when it clamps. Added regression tests in `number-utils.spec.ts` pinning the NaN/Infinity behaviour.
- 🟡 "byte-identical at full res" wording → softened to "pixel-equivalent on replay" (changeset + JSDoc).
- ⚪ Dropped the leftover playwright debug `console.log`; fixed the stale "two independent contexts" spec header (two sequential recordings on one page) and noted the assertion is a byte-size *trend*, not pixel identity.
- ⚪ JSDoc: replaced time-relative "matching today's behaviour" with "full-resolution capture"; documented clamping.
- ⚪ Added `0`, negative, and `NaN` rows to the parameterized `canvasResolutionScale` test.

**Deferred — needs author/product judgement**
- 🟡 **Divergent clamp policy** (`canvas-manager.ts`): the rrweb constructor accepts `(0,1]` with no `0.1` floor, so a *direct* `@posthog/rrweb` consumer could pass `0.001`. Through the PostHog SDK this can't happen (the getter floors at 0.1 first). Whether to add a floor in the vendored package vs document the boundary is an rrweb-package decision — left for the author.
- 🟡 **2026-05-30 default `0.6`**: intentional; `0.6` is a reasoned starting point rather than an empirically-tuned value (resolution is quadratic in bytes at low perceptual cost). It lands in `$session_options`, so the cohort is measurable — validation can follow rollout. Confirming intent + downstream docs is an author/product call.
- 🟢 `displayDimsMap` unbounded-per-session: left as-is — matches the pre-existing `snapshotInProgressMap` shape and is GC'd on record restart.
- 🟢/⚪ rrweb structural nits (`scale` local scope, re-introduce `timeBetweenSnapshots` const): skipped to avoid churn in the vendored package for marginal benefit.

.changeset/adaptive-canvas-capture.md
```diff
@@ -0,0 +1,12 @@
+---
+'posthog-js': minor
+'@posthog/rrweb': patch
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `4fd359a5` — added `'@posthog/types': patch` to the changeset so the new `canvasCapture` declarations publish alongside the SDK change.

packages/browser/src/posthog-core.ts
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Confirmed and fixed in `4fd359a5`. `session_recording` is now deep-merged like `rageclick`, so a partial user-supplied `session_recording` keeps the date-gated `canvasCapture.resolutionScale` (and `strictMinimumDuration`) defaults instead of replacing them. Added config tests covering both the keep-default and user-override cases.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
pauldambra: > [!NOTE]

Fixed in `199701e0` — the rrweb `CanvasManager` now clamps `resolutionScale` to `[0.1, 1]` (non-finite/unset -> full res), matching the SDK floor, so a direct `@posthog/rrweb` caller can't request a degenerate resolution. The two layers now agree.

pauldambra: > [!NOTE]

Fixed in `c90303af` — the `resolutionScale` clamp now lives inside `initCanvasFPSObserver` (the only path that uses it); the constructor passes the raw option through and the mutation-observer path no longer computes a scale it never uses.

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
pauldambra: > [!NOTE]

Fixed in `c90303af` — restored the named `timeBetweenSnapshots` constant in place of the inlined `1000 / fps`.

pauldambra:
> [!NOTE]
>
> Third multi-perspective pass (focused on the new logic since the last review: the stateless worker-carry refactor + the session_recording deep-merge). qa-team, paul-reviewer, xp-reviewer, security-audit.

## Verdict: ✅ APPROVE (nits only)

No CRITICAL/HIGH; security-audit found **0**. The headline change — dropping the per-id `displayDimsMap` and carrying the display size through the encode worker as **required** message fields — was independently validated as correct by all three reviewers: every `base64`-bearing reply carries `displayWidth`/`displayHeight`, and all `{id}`-only early-returns are no-draw paths, so no frame is ever drawn without its dimensions. The `session_recording` deep-merge mirrors the existing `rageclick` pattern and is covered by new config tests.

### Findings (all LOW/NIT/CONSIDER)

**Actioned (`45597fd8`)**
- The `session_recording` merge comment over-promised: it's a *shallow one-level* merge (a user-supplied nested `canvasCapture` still replaces the default wholesale). Comment corrected to say so. _(convergent: qa-team + paul + xp)_
- Named the magic `0.7` byte-ratio ceiling in the playwright spec. _(xp)_

**Acknowledged — deliberate, no change**
- The `0.1` floor exists as two independent constants (`MIN_CANVAS_SCALE` in the SDK, `MIN_CANVAS_RESOLUTION_SCALE` in rrweb). This is intentional cross-package boundary defense — rrweb is published separately and can't trust its caller — and both sites cross-reference each other in comments. Can't be deduped across the package boundary. _(convergent: qa-team + paul + xp; all three agreed it's justified)_
- Extracting the two near-identical `rageclick` / `session_recording` merge blocks into a loop: xp flagged as CONSIDER-not-must-fix; two is the threshold and a typed loop over `PostHogConfig` keys adds cast friction, so left as-is.
- `dw`/`dh` aliasing and a `clampResolutionScale` helper in rrweb: cosmetic, not worth another vendored-package rebuild cycle.

**Confirmed (paul's observability question)**
- `canvasResolutionScale` is set on `sessionRecordingOptions`, which is emitted in the `$session_options` custom event — so the `0.6` cohort is inspectable per-session, confirming the safe-rollout shape.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | New worker-carry plumbing is complete and correct; merge mirrors rageclick; findings all LOW/NIT. |
| 👤 paul | "The version of the change I was hoping for" — stateless carry reads clean, parameterised tests earn their keep, the replay test proves the stretch; approving. |
| 📐 xp | Strong simplification (removes state + its whole cleanup lifecycle); fix the over-promising comment (done), watch the twin floor constants. |
| 🛡 security-audit | 0 findings — dims derive from `canvas.clientWidth` not attacker input; merge keys are the host developer's own config; clamps reject NaN/Infinity. |

---

packages/browser/playwright/mocked/session-recording/canvas-resolution.spec.ts
```diff
@@ -0,0 +1,131 @@
...
+        await page.locator('[data-cy-input]').type('x')
+        await page.waitForTimeout(1000)
+    }
+    await page.evaluate(() => (window as WindowWithPostHog).posthog?.capture('flush'))
+    await page.waitForTimeout(800)
+
+    return latestCanvasFrame((await page.capturedEvents()) || [])
+}
+
+test.describe('canvas capture resolution', () => {
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Good catch — fixed in `8e1b915e`. The canvas FPS observer requires `OffscreenCanvas` (absent on webkit), so the spec now `test.skip`s on webkit rather than asserting on a frame that can't be captured there.

packages/types/src/posthog-config.ts
```diff
@@ -563,6 +563,21 @@ export interface SessionRecordingOptions {
...
+     * - `resolutionScale`: capture canvas frames at a fraction of their display resolution. A
+     *   number in `(0, 1]`; `1` is full-resolution capture (the default) and, e.g., `0.6` captures
+     *   at 60%. Out-of-range or non-finite values are clamped into `(0, 1]`. Aspect ratio is
+     *   preserved and replay upscales the frame back to the original display size, so playback
+     *   dimensions are unchanged, just softer. Resolution is the highest-leverage lever for canvas
+     *   byte size, since bytes scale with pixel area.
+     */
+    canvasCapture?: {
+        resolutionScale?: number
+    }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `8e1b915e` — regenerated the `@posthog/types` config snapshot; it now includes `canvasCapture`.

ksvat (approved):

PR #3886, opened 2026-06-18
fix: remove ignored batch metadata fields

marandaneto:
## Problem

Event batch payloads included top-level `type`, `library`, and `library_version` fields for every event. The capture backend hydrates `/batch/` events into `RawEvent` and uses `properties.$lib` / `properties.$lib_version` for SDK metadata, so those high-level fields are ignored and add unnecessary payload size. The top-level `type` field is deprecated and a no-op; top-level `library` and `library_version` are deprecated in favor of the official `$lib` and `$lib_version` event properties.

## Changes

- Stop adding deprecated top-level `type`, `library`, and `library_version` to core event messages.
- Normalize messages after `before_send` and during queue flush so legacy queued messages do not send those ignored fields.
- Preserve backwards compatibility for legacy queued messages by copying top-level `library` / `library_version` into `properties.$lib` / `properties.$lib_version` when the official `$` properties are missing.
- Preserve existing behavior for malformed/non-event queue entries by not attempting to normalize missing messages during flush.
- Document in the changeset and code that `type` is a deprecated no-op, and that `library` / `library_version` should use the official `$lib` / `$lib_version` properties instead.
- Add patch changesets for `@posthog/core` and downstream packages that use the core batch code path directly or via `posthog-node`: `posthog-node`, `posthog-js-lite`, `posthog-react-native`, `@posthog/convex`, `@posthog/next`, and `@posthog/nuxt`.
- Update core and node tests to assert the new batch payload shape.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

This PR was implemented with pi. I verified the backend batch path only models SDK metadata via event properties (`$lib` / `$lib_version`) and then made the SDK stop sending the ignored high-level fields while keeping a fallback for legacy queued events. The changeset was expanded to include downstream packages that use the core event batch code path and to call out the deprecated/no-op fields.

Validation run:

```sh
pnpm --filter @posthog/types build
pnpm --filter @posthog/core build
pnpm --filter posthog-node build
pnpm --filter @posthog/core test:unit
pnpm --filter posthog-node test:unit
pnpm --filter @posthog/core lint
pnpm --filter posthog-node lint
pnpm --filter @posthog/core test:unit -- posthog.capture.spec.ts posthog.enqueue.spec.ts posthog.listeners.spec.ts
pnpm --filter posthog-react-native prebuild
pnpm --filter posthog-react-native test:unit -- posthog.spec.ts
```


packages/core/src/posthog-core-stateless.ts
```diff
@@ -1056,22 +1058,38 @@ export abstract class PostHogCoreStateless {
...
+  protected prepareMessage(_type: string, _message: any, options?: PostHogCaptureOptions): PostHogEventProperties {
+    const message = this.normalizeMessage({
       ..._message,
-      type: type,
-      library: this.getLibraryId(),
-      library_version: this.getLibraryVersion(),
       timestamp: options?.timestamp ? options?.timestamp : currentISOTime(),
       uuid: options?.uuid ? options.uuid : uuidv7(),
-    }
+    })
```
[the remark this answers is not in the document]
marandaneto: Fixed in 0c02f9e8 — removed the unused prepareMessage type parameter and the redundant normalizeMessage call while keeping normalization after before_send and during queue flush.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1056,12 +1058,32 @@ export abstract class PostHogCoreStateless {
...
   }
 
-  protected prepareMessage(type: string, _message: any, options?: PostHogCaptureOptions): PostHogEventProperties {
+  private normalizeMessage(message: PostHogEventProperties): PostHogEventProperties {
+    // Top-level `type` is deprecated and a no-op for batch ingestion.
+    // Top-level `library` and `library_version` are deprecated; `$lib` and
+    // `$lib_version` in properties are the canonical SDK metadata fields.
+    // Keep these as fallbacks for legacy persisted queue entries.
+    const { type: _type, library, library_version, ...sanitizedMessage } = message
+    let properties = sanitizedMessage.properties as PostHogEventProperties | undefined
```
dustinbyrne: there are a couple of strange things going on with the typing here:
- at the call site, `message` is typed as `any`
- casting `properties` as `PostHogEventProperties | undefined` is unsafe

typing `message` as `PostHogEventProperties` means that properties could also be `null`, `string`, `number`, etc.

something like this would allow us to drop the unsafe cast on `properties` (though I'm sure the message has a stronger type than just `PostHogEventProperties`?):
```ts
type PostHogEventMessage = PostHogEventProperties & { properties: PostHogEventProperties }
```

the real root of the issue is that in `PostHogQueueItem`, `message` is typed as `any`, otherwise the type checker would catch this

dustinbyrne: fwiw, this could be a follow up item, because without types this still looks correct.

we should make a goal to drop usage of `any` (there are a few cases where it's necessary, but this is not one of them)

dustinbyrne (approved):

packages/core/src/posthog-core-stateless.ts
marandaneto: Addressed in a940f2e2 — `PostHogQueueItem.message` is no longer `any`, and `normalizeMessage` / `$geoip_disable` now narrow `properties` with an object guard instead of casting. I kept `message` optional so malformed legacy/persisted queue entries are still tolerated during flush.

marandaneto: yeah totally agree, `any` is in lots of places so def a bigger change

PR #3900, opened 2026-06-19
fix(browser): remove unused ip query parameter

marandaneto:
## Problem

The browser SDK still appends an `ip` query parameter to SDK requests even though capture no longer uses it. This creates unnecessary URL surface area for ad blockers without changing ingestion behavior.

## Changes

- Stop adding the `ip` query parameter in the shared browser request path.
- Remove the now-unused `skipIPParam` request option and feature flag special-casing.
- Update browser tests and request logging matchers that expected `ip=0` in SDK request URLs.
- Add a changeset for `posthog-js`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with pi. Before changing the SDK, we checked capture backend usage: `/e/` and `/batch` ignore the `ip` query parameter, and `beacon=1` only changes the response status to 204 for sendBeacon requests. Based on that, this PR removes the unused browser-side `ip` query parameter logic while preserving the deprecated `ip` config field for compatibility.

Validation attempted: `pnpm --filter=posthog-js test:unit -- --runInBand src/__tests__/posthog-core-also.test.ts src/__tests__/featureflags.test.ts src/__tests__/extensions/exception-autocapture/exception-observer.test.ts`, but it could not run because this worktree has no installed `node_modules` (`jest: command not found`). CI reported a Prettier lint issue and an out-of-date public API reference; this branch now includes the Prettier formatting fix and the updated `posthog-js-references-latest.json` change from `pnpm generate-references`.

### DIFF
--- a/packages/browser/references/posthog-js-references-latest.json
+++ b/packages/browser/references/posthog-js-references-latest.json
@@ -3565,10 +3565,6 @@
           "name": "noRetries"
         },
         {
-          "type": "boolean",
-          "name": "skipIPParam"
-        },
-        {
           "type": "number",
           "name": "timeout"
         },
--- a/.changeset/quiet-pumas-listen.md
+++ b/.changeset/quiet-pumas-listen.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Stop adding the unused `ip` query parameter to browser SDK requests.
--- a/packages/browser/playwright/mocked/session-recording/session-recording-network-recorder.spec.ts
+++ b/packages/browser/playwright/mocked/session-recording/session-recording-network-recorder.spec.ts
@@ -149,10 +149,7 @@ test.beforeEach(async ({ context }) => {
                 expect(hasEntry(/http:\/\/localhost:\d+\/playground\/cypress\//, 'navigation')).toBe(true)
                 expect(hasEntry(/https:\/\/localhost:\d+\/static\/array.js/, 'script')).toBe(true)
                 expect(
-                    hasEntry(
-                        /https:\/\/localhost:\d+\/array\/test%20token\/config\?_=\d+&ver=1\.\d\d\d\.\d+/,
-                        'fetch'
-                    )
+                    hasEntry(/https:\/\/localhost:\d+\/array\/test%20token\/config\?_=\d+&ver=1\.\d\d\d\.\d+/, 'fetch')
                 ).toBe(true)
                 expect(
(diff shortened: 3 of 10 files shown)


marandaneto:
### /e/ and /batch use the same handler

 Both routes point to v0_endpoint::event:

 - rust/capture/src/router.rs
     - /batch, /batch/
     - /e, /e/
     - /i/v0/e, /i/v0/e/

 ### ip query parameter

 ip is not used.

 The backend query struct is:

 ```rust
   pub struct EventQuery {
       pub compression: Option<Compression>,
       pub data: Option<String>,
       #[serde(alias = "ver")]
       pub lib_version: Option<String>,
       #[serde(alias = "_")]
       pub sent_at: Option<i64>,
       pub beacon: bool,
   }
 ```

 Source: rust/capture/src/payload/types.rs

 There is no ip field, so ?ip=0, ?ip=1, or any ?ip=... is ignored for /e/
  and /batch.

 The actual IP stored on ingested events comes from the request/client IP
 extractor:

 ```rust
   InsecureClientIp(ip)
   ...
   client_ip: ip.to_string()
 ```

 Then later:

 ```rust
   ip: resolved_ip
 ```

 Source:
 - rust/capture/src/payload/analytics.rs
 - rust/capture/src/events/analytics.rs

 So dropping ip=0 from browser /e/ requests should not change backend
 behavior.

marandaneto:
@PostHog/team-ingestion pls double check that this is not used at all in any other system before we cause an incident :)

packages/browser/references/posthog-js-references-latest.json
```diff
@@ -3564,10 +3564,6 @@
           "type": "boolean",
           "name": "noRetries"
         },
-        {
-          "type": "boolean",
-          "name": "skipIPParam"
-        },
```
ioannisj, line 3570: Is this breaking public api or is this internal type?

marandaneto: its internal but exported so leaks to public api if people reference directly (still internal tho)
excluding all those internal exports from the reference would be tricky

ioannisj (approved):

PR #3901, opened 2026-06-19
fix(browser): remove beacon query parameter

marandaneto:
## Problem

The browser SDK adds `beacon=1` to `navigator.sendBeacon` ingestion request URLs. Capture only uses this query parameter to return `204 No Content` instead of the normal `200 OK` JSON response; event processing is unchanged. Removing it further reduces URL surface area that can be matched by blocker rules.

This PR is stacked on #3900 (`fix/remove-ip-query-param`) to avoid conflicts with the related `ip` query parameter cleanup.

## Changes

- Stop adding the `beacon=1` query parameter in the sendBeacon transport.
- Update sendBeacon request tests to expect the same URL without `beacon=1`.
- Add a changeset for `posthog-js`.

Trade-offs / behavior notes:

- Without `beacon=1`, capture returns `200 OK` with a small JSON response instead of `204 No Content`.
- The SDK uses `navigator.sendBeacon`, which does not expose the response status or body to JavaScript, and this code path does not invoke SDK callbacks. So SDK behavior should not change.
- The browser may receive and discard a small response body instead of no body.
- DevTools, proxies, WAFs, or customer monitoring that key off `beacon=1` or expect `204` for these requests may observe the URL/status change, but this is not part of the public SDK contract.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with pi. We investigated the backend and SDK behavior first: capture uses `beacon=1` only to choose `204 No Content`, while browser `sendBeacon` does not expose the response to the SDK. Based on that, this PR removes the unused query parameter and documents the observable response-status trade-off for reviewers.

Validation attempted: `pnpm --filter=posthog-js test:unit -- --runInBand src/__tests__/request.test.ts`, but it could not run because this worktree has no installed `node_modules` (`jest: command not found`). After CI reported a Prettier lint error in the stacked base change, the Playwright expectation was reformatted to match Prettier.


marandaneto:
If the SDK sends navigator.sendBeacon() without beacon=1:

 ### Browser / SDK behavior

 - Browser still sends the request normally.
 - navigator.sendBeacon() does not expose the HTTP response to JS.
 - The SDK’s _sendBeacon path does not register a callback and ignores
   the sendBeacon() boolean return value.
 - So the SDK will not see whether the API returned 200, 204, or anything
   else.

 In practice:

 ```ts
   navigator.sendBeacon(url, body)
 ```

 is fire-and-forget. The browser may receive the 200 OK response body,
 but it is discarded/not exposed to the page.

 ### API behavior

 Without beacon=1, capture returns normal:

 - 200 OK
 - JSON response body

 With beacon=1, it returns:

 - 204 No Content

 Event ingestion happens the same either way.

 ### Is it breaking?

 For normal SDK users: no, it should not be a breaking change.

 Potential minor differences:

 - DevTools/network logs may show 200 instead of 204.
 - The browser may download a tiny JSON response body instead of no body.
 - If a customer/proxy/WAF/custom monitoring relies specifically on
   beacon=1 to classify these requests, that unofficial dependency could
   change.

 But from the SDK API and ingestion perspective, removing beacon=1 should
 be safe.

marandaneto:
@PostHog/team-ingestion same question as https://github.com/PostHog/posthog-js/pull/3900#issuecomment-4749636596

ioannisj (approved):

PR #3909, opened 2026-06-19
fix(rrweb): avoid CSP violations for style mutation diffs

### LINKED ISSUE #3906: CSP Violation on style mutations within rrweb
## Bug description

The session recorder's attribute-mutation handler trips a `style-src-attr` Content Security Policy violation (blocked-uri `inline`) on every `style`-attribute mutation on the page.

To compute a style diff, it applies the previous style string with `setAttribute('style', oldValue)` on a throwaway element:

`packages/rrweb/rrweb/src/record/mutation.ts` (~line 690 on `master`):

```js
if (attributeName === 'style') {
  if (!this.unattachedDoc) {
    try {
      // avoid upsetting original document from a Content Security point of view
      this.unattachedDoc = document.implementation.createHTMLDocument();
    } catch (e) {
      this.unattachedDoc = this.doc;
    }
  }
  const old = this.unattachedDoc.createElement('span');
  if (m.oldValue) {
    old.setAttribute('style', m.oldValue);   // ← triggers style-src-attr
  }
  for (const pname of Array.from(target.style)) { ... }
  ...
}
```

The code already tries to avoid this (it builds the element in a `document.implementation.createHTMLDocument()`, per the *"avoid upsetting original document from a Content Security point of view"* comment), but `setAttribute('style', …)` is still evaluated against `style-src-attr` and reported, and the `catch` fallback writes to the real document. On an SPA with strict CSP this produces a constant stream of reports.

## How to reproduce

1. Serve a page with a CSP containing `style-src 'self' 'nonce-…'` and **no** `style-src-attr 'unsafe-inline'` (plus a `report-uri`/`report-to`).
2. Initialize posthog-js with session recording enabled.
3. Mutate any element's `style` attribute.
4. Observe `style-src-attr` / blocked-uri `inline` CSP violation reports.

Environment: `posthog-js` 1.388.1 (also present on `master`); reproduced in Chrome (across Chrome 148, not browser-specific).

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

marandaneto:
## Problem

Fixes #3906.

Strict CSP configurations with `style-src-attr` disabled report violations during session recording whenever rrweb diffs a `style` attribute mutation. The recorder parsed the previous style value by calling `setAttribute('style', oldValue)` on a throwaway element, which still triggers CSP evaluation.

## Changes

- Parse the previous style value with `old.style.cssText = m.oldValue` instead of `setAttribute('style', ...)` so CSSOM can diff styles without triggering `style-src-attr`.
- Added a Puppeteer-backed rrweb integration test that records style mutations under `style-src-attr 'none'` and asserts no CSP violation is emitted.
- Added a patch changeset for `posthog-js`, which bundles the rrweb recorder into the browser SDK during release.

Validation:

- `pnpm --filter @posthog/rrweb build`
- `PUPPETEER_HEADLESS=true pnpm --filter @posthog/rrweb exec vitest run test/integration.test.ts -t "does not trigger style-src-attr"`
- Standalone Puppeteer reproduction against the rebuilt rrweb bundle: `cspViolationCount: 0`
- `pnpm eslint packages/rrweb/rrweb/src/record/mutation.ts packages/rrweb/rrweb/test/integration.test.ts`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the pi coding agent in a dedicated worktree. The fix keeps the existing style-diffing behavior but switches the parsing mechanism from a style attribute write to CSSOM `cssText`, which the focused Puppeteer reproduction confirmed avoids `style-src-attr` CSP reports.

### DIFF
--- a/.changeset/quiet-spiders-dance.md
+++ b/.changeset/quiet-spiders-dance.md
@@ -1,0 +1,6 @@
+---
+'@posthog/rrweb': patch
+'posthog-js': patch
+---
+
+Avoid `style-src-attr` CSP violations when diffing rrweb style mutations.
--- a/.changeset/quiet-spiders-dance.md
+++ b/.changeset/quiet-spiders-dance.md
@@ -1,5 +1,5 @@
 ---
-'@posthog/rrweb': patch
+'posthog-js': patch
 ---
 
 Avoid `style-src-attr` CSP violations when diffing rrweb style mutations.
--- a/.changeset/quiet-spiders-dance.md
+++ b/.changeset/quiet-spiders-dance.md
@@ -1,6 +1,5 @@
 ---
 '@posthog/rrweb': patch
-'posthog-js': patch
 ---
 
 Avoid `style-src-attr` CSP violations when diffing rrweb style mutations.
--- a/packages/rrweb/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/rrweb/src/record/mutation.ts
@@ -687,7 +687,7 @@ export default class MutationBuffer {
             }
             const old = this.unattachedDoc.createElement('span');
             if (m.oldValue) {
-              old.setAttribute('style', m.oldValue);
+              old.style.cssText = m.oldValue;
             }
             for (const pname of Array.from(target.style)) {
               const newValue = target.style.getPropertyValue(pname);
--- a/packages/rrweb/rrweb/test/integration.test.ts
+++ b/packages/rrweb/rrweb/test/integration.test.ts
@@ -325,6 +325,47 @@ describe('record integration tests', function (this: ISuite) {
     await assertSnapshot(snapshots);
   });
 
+  it('does not trigger style-src-attr CSP violations while diffing style mutations', async () => {
+    const page: puppeteer.Page = await browser.newPage();
+    const client = await page.target().createCDPSession();
+    const cspViolations: string[] = [];
+    const collectCSPViolation = (text: string) => {
+      if (text.includes('style-src-attr')) {
+        cspViolations.push(text);
+      }
+    };
+
+    await client.send('Log.enable');
+    client.on('Log.entryAdded', ({ entry }) => collectCSPViolation(entry.text));
+    page.on('console', (message) => collectCSPViolation(message.text()));
+
+    await page.goto('about:blank');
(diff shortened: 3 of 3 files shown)


.changeset/quiet-spiders-dance.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/rrweb': patch
```
marandaneto: not sure we need this anymore?

marandaneto:
@PostHog/team-replay can you double check this?

.changeset/quiet-spiders-dance.md
marandaneto: Adjusted in bf4a1787 — the changeset now only declares `@posthog/rrweb`, matching the package with source changes and clearing the changeset hygiene warning.

marandaneto: i think it was the oposite, check the release pipeline in the ci, isnt rrweb packaged with posthog-js automatically?

marandaneto: You were right — `.github/workflows/release.yml` omits `@posthog/rrweb*` from the publish matrix because those packages are bundled into `posthog-js` at build time. Updated in 980cdb44 so the changeset bumps `posthog-js` only.

ioannisj (approved):

TueHaulund (approved):

PR #3911, opened 2026-06-20
feat(replay): redact binary bodies, datadog, and credential headers in network capture

pauldambra:
## Problem

Investigating unusually large session recordings, the network-capture plugin was found to be the dominant size driver in some of them — and one example was recording a third-party auth token. Three independent issues, all when a customer has `recordBody`/`recordHeaders` enabled:

- **Binary/asset bodies are recorded.** With `recordBody: true`, every content-type is captured, including image/video/font responses. In one recording, ~80 MB (42% of its captured bodies) was `image/webp` response bodies — duplicating what the replay already shows, for no debugging value.
- **Third-party telemetry is captured.** Datadog browser-RUM intake traffic was recorded (~20 MB in the same recording) despite having no replay value — the deny-list mechanism already exists for similar vendors (Sentry, Clarity, GA, New Relic) but Datadog wasn't on it.
- **Credential headers leak.** Header redaction only ran on **request** headers and matched an exact deny list, so `set-cookie` (a response header) and credential-shaped custom names (observed: `x-gist-encoded-user-token`) were captured verbatim.

## Changes

- `shouldRecordBody`: never record bodies whose content-type is binary/asset (`image/`, `video/`, `audio/`, `font/`, `application/octet-stream`, `pdf`, `zip`, `wasm`), regardless of `recordBody`. Returning `false` here also means the body is never read — no extra processing.
- Add Datadog intake hosts (`datadoghq.com`, `datadoghq.eu`, `ddog-gov.com`) to the default `payloadHostDenyList`.
- Header redaction now applies to **both request and response** headers, and matches credential-shaped names by substring (auth/token/secret/session/api-key/credential/password/cookie/csrf/xsrf) in addition to the exact deny list.

All three are default-on and reduce both recording size and accidental sensitive-data capture. A **follow-up** will make the payload size cap enforce on the decompressed/stored body length (today it trusts `content-length`, which is the compressed size for gzipped responses — so a 0.5 MB-on-the-wire response that decompresses to 3+ MB slips the 1 MB cap).

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with PostHog Code (Claude) at Paul's direction, from analysis of three exported customer recordings: their size was driven by canvas frames, DOM-mutation churn, and (here) network body capture respectively — i.e. "large recording" wasn't one problem. These three redactions are the network-capture slice. The decompressed-size cap was deliberately split into a follow-up because, for cross-origin responses, `content-encoding` isn't readable and resource-timing sizes are zeroed, so doing it without reading the body needs a separate bounded-reader approach.

Tests: added a parameterized binary-content-type case matrix and a request+response header-redaction case (incl. a credential-shaped custom name). 73 tests pass in the replay config/network/denylist suites.

---

### DIFF
--- a/.changeset/replay-network-redaction.md
+++ b/.changeset/replay-network-redaction.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Session replay network capture: never record binary/asset response bodies (images, video, audio, fonts, etc.), skip Datadog browser-RUM intake hosts, and redact credential-bearing headers on both request and response by name heuristic (e.g. custom `*-token` headers) - reducing recording size and avoiding accidental credential capture.
--- a/packages/browser/src/__tests__/extensions/replay/config.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/config.test.ts
@@ -67,6 +67,34 @@ describe('config', () => {
                 })
             })
 
+            it('redacts denied request and response headers, including credential-shaped custom names', () => {
+                const networkOptions = buildNetworkRequestOptions(defaultConfig(), {})
+                const cleaned = networkOptions.maskRequestFn!({
+                    name: 'something',
+                    requestHeaders: {
+                        'x-gist-encoded-user-token': 'abc',
+                        'content-type': 'application/json',
+                    },
+                    responseHeaders: {
+                        'set-cookie': 'session=secret',
+                        'x-session-id': 'xyz',
+                        'content-type': 'application/json',
+                    },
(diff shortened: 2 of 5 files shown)


pauldambra:
Superseded by a 3-PR Graphite stack splitting these changes for independent review:

- #3912 — skip binary/asset network bodies
- #3913 — deny-list Datadog intake hosts
- #3914 — redact credential headers (request + response, heuristic)

Closing this combined PR in favour of the stack.

PR #3912, opened 2026-06-20
feat(replay): skip binary/asset bodies in network capture

pauldambra:
## Problem

With `recordBody` enabled, the network recorder captures _every_ content-type, including image/video/font response bodies. In analysed recordings this was the single largest size driver (e.g. ~80 MB / 42% of one recording was `image/webp` bodies). The replay network inspector renders bodies only as a text/JSON `CodeSnippet` — it has no image viewer — so a binary body shows as unreadable [mojibake](https://en.wikipedia.org/wiki/Mojibake). Capturing them is pure cost with zero human value.

## Changes

`shouldRecordBody` now returns `false` for binary/asset content-types (`image/`, `video/`, `audio/`, `font/`, `application/octet-stream`, `pdf`, `zip`, `wasm`) regardless of the `recordBody` setting. Returning false here also means the body is never read.

**Stack 1/3** of the replay network-capture reductions (binary bodies → Datadog hosts → credential headers).

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility (no breaking changes)
- [x] Took care not to unnecessarily increase the bundle size

### If releasing new changes

- [x] Ran `pnpm changeset`

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted)

Authored with PostHog Code (Claude) at Paul's direction, from analysis of exported recordings whose size was dominated by network body capture. Confirmed via the replay inspector code that binary bodies are never rendered as images.

---

_Created with_ [_PostHog Code_](https://posthog.com/code?ref=pr)

PR #3913, opened 2026-06-20
feat(replay): skip third-party analytics/telemetry hosts in network capture

pauldambra:
## Problem

Session replay was capturing request/response **bodies** sent to third-party analytics, RUM, and session-replay vendors — pure telemetry with no replay-debugging value, and often large batched POSTs. One analysed recording spent ~20 MB on Datadog RUM intake alone. The payload-host deny-list mechanism already existed (Sentry, Clarity, GA, New Relic, LogRocket) but missed the common offenders.

## Changes

Expand the default `payloadHostDenyList` (matched by hostname suffix, so it covers regional intake subdomains):

- **Datadog** — `datadoghq.com`, `datadoghq.eu`, `ddog-gov.com`
- **Segment** — `segment.io`
- **RudderStack** — `rudderstack.com`
- **Amplitude** — `amplitude.com`
- **Mixpanel** — `mixpanel.com`
- **Hotjar** — `hotjar.com`
- **FullStory** — `fullstory.com`

Selection criteria: vendor-owned domains (so suffix-matching can't catch a customer's own API), that POST sizable telemetry batches (GET pixels have no body, so ad pixels are intentionally excluded), and are widely deployed.

Also corrects two existing entries while here:
- **Google Analytics**: `analytics.google.com` → `google-analytics.com`. The old entry was the GA *dashboard* host; GA4/gtag beacons go to `*.google-analytics.com`, so GA bodies weren't actually being skipped.
- **New Relic**: `bam.nr-data.net` → `nr-data.net`, to also cover `bam-cell.nr-data.net`.

The deny-list only suppresses the body — the request still appears in the inspector timeline.

**Stack 2/3** of the replay network-capture reductions.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset`

## 🤖 Agent context

Authored with PostHog Code (Claude) at Paul's direction. The vendor set was curated against an explicit "keep the list short" constraint — heavy-deployment vendors that batch-POST telemetry — and deliberately excludes GET-only ad pixels. The GA dashboard-vs-beacon-host correction surfaced while reviewing the existing entries.

---

### DIFF
--- a/packages/browser/src/__tests__/extensions/replay/config.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/config.test.ts
@@ -214,8 +214,19 @@ describe('config', () => {
                     '.lr-ingest.io',
                     '.ingest.sentry.io',
                     '.clarity.ms',
+                    'google-analytics.com',
                     'analytics.google.com',
-                    'bam.nr-data.net',
+                    'nr-data.net',
+                    'datadoghq.com',
+                    'datadoghq.eu',
+                    'ddog-gov.com',
+                    'segment.io',
+                    'rudderstack.com',
+                    'amplitude.com',
+                    'mixpanel.com',
+                    'hotjar.com',
+                    'hotjar.io',
+                    'fullstory.com',
                 ])
             })
 
@@ -229,8 +240,19 @@ describe('config', () => {
                     '.lr-ingest.io',
                     '.ingest.sentry.io',
                     '.clarity.ms',
+                    'google-analytics.com',
                     'analytics.google.com',
-                    'bam.nr-data.net',
+                    'nr-data.net',
+                    'datadoghq.com',
+                    'datadoghq.eu',
+                    'ddog-gov.com',
+                    'segment.io',
+                    'rudderstack.com',
+                    'amplitude.com',
+                    'mixpanel.com',
+                    'hotjar.com',
+                    'hotjar.io',
+                    'fullstory.com',
                 ])
             })
         })
--- a/.changeset/replay-denylist-datadog.md
+++ b/.changeset/replay-denylist-datadog.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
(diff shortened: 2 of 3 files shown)


pauldambra:
* **#3914** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3914?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#3913** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3913?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3913?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#3912** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/3912?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

.changeset/replay-skip-binary-bodies.md
marandaneto: non related to this pr? wrong base branch? or i'd split in 2 prs at least

pauldambra: oh mad, the robot replaced the content with base pr 🤯

packages/browser/src/__tests__/extensions/replay/config.test.ts
```diff
@@ -214,8 +214,19 @@ describe('config', () => {
                     '.lr-ingest.io',
                     '.ingest.sentry.io',
                     '.clarity.ms',
+                    'google-analytics.com',
```
marandaneto, line 217: nit: some people prefer to dupe everything in tests
i prefer to extract `payloadHostDenyList` into an exported array and reuse it in tests

marandaneto (approved):

packages/browser/src/__tests__/extensions/replay/config.test.ts
pauldambra: here the duplication (in theory) stops someone editing production list without considering it  
  
(i guess in practice their agent would simpy change both places but 🙈)  
  
(i don't have a strong opinion here)

marandaneto: yep, humans would do the same tbh
there are lots of cases where that makes sense to be duplicated, not sure if this is one of them, anyway approved and just a nit so GG

PR #3915, opened 2026-06-20
fix(replay): cap oversized data URIs in SVG <image> elements

pauldambra:
## Problem

Session replay already caps oversized base64 `data:` image URIs (via `maxBase64ImageLength`, default 1MB), replacing them with a small striped placeholder so recordings don't balloon. But that cap only applied to `` elements. Inline SVG `<image>` elements with `data:` URIs — common on graphics-heavy marketing pages and animation libraries that render into SVG — slipped through entirely and were recorded in full, both in the full snapshot and in mutations. On pages that re-render these SVGs frequently, that inflates recording storage with large, duplicated image payloads.

## Changes

Extend the existing size cap from `` to SVG `<image>` elements in `transformAttribute`:

- `href` (modern SVG2 attribute) and `xlink:href` (legacy SVG1.1 attribute) `data:` URIs on SVG `<image>` are now size-capped to the striped placeholder when they exceed `maxBase64ImageLength`.
- The cap logic is extracted into a single `capDataURLSize` helper, now shared by the ``, SVG `<image>`, and `xlink:href` paths.
- SVG `<image>` is an `SVGImageElement` and cannot be redrawn through a canvas, so unlike `` we never attempt webp re-encoding for it — only the size cap, which is a pure synchronous string-length check. This keeps the serialization path synchronous (no new async work) and captures nothing extra.

Because `transformAttribute` is the shared serializer, this also covers SVG `<image>` data URIs added via mutations, not just the initial full snapshot.

### Tests

- jsdom unit tests in `snapshot.test.ts` exercising `transformAttribute` directly (parameterized over `href` / `xlink:href`): oversized URIs are capped, small ones preserved, and nothing is touched when no cap is configured.
- Puppeteer integration tests + an extended `base64-images.html` example page covering large SVG `<image>` data URIs for both `href` and `xlink:href`, plus a small-image preservation case.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

This started as an investigation into why some recordings carry large inline images despite the existing data-URI cap. The cap turned out to be gated on `tagName === 'img'`, leaving SVG `<image>` data URIs (on both `href` and `xlink:href`) completely unhandled — and the `xlink:href` branch had no data-URL handling at all.

Worked test-first: extended the existing `base64-images.html` example page and added integration tests asserting the placeholder for oversized SVG images; confirmed they failed (proving the gap) before implementing. Chose size-cap-only for SVG `<image>` rather than canvas re-encoding, because `SVGImageElement` has no canvas decode path and the guiding constraint was to keep the serialization path synchronous and avoid capturing anything extra. Verified the unit suite, integration base64 suite, type-check, and lint all pass locally; the one affected round-trip snapshot was regenerated (large canvas URIs collapse to a deterministic placeholder).

Tools: PostHog Code (Claude Opus).

---

### DIFF
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -211,8 +211,8 @@ export function transformAttribute(
     // href starts with a # is an id pointer for svg
     const transformedValue = absoluteToDoc(doc, value);
 
-    if (transformedValue.startsWith('data:') && element) {
-      if (tagName === 'img') {
+    if (transformedValue.startsWith('data:')) {
+      if (tagName === 'img' && element) {
         let processedDataURL = transformedValue;
 
         if (dataURLOptions?.type || dataURLOptions?.quality !== undefined) {
@@ -236,7 +236,7 @@ export function transformAttribute(
   } else if (name === 'xlink:href' && value[0] !== '#') {
     // xlink:href starts with # is an id pointer
     const transformedValue = absoluteToDoc(doc, value);
-    if (transformedValue.startsWith('data:')) {
+    if (tagName === 'image' && transformedValue.startsWith('data:')) {
       return capDataURLSize(transformedValue, dataURLOptions);
     }
     return transformedValue;
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -174,6 +174,16 @@ function getHref(doc: Document, customHref?: string) {
   return a.href;
 }
 
+function capDataURLSize(
+  dataURL: string,
+  dataURLOptions?: DataURLOptions,
+): string {
+  if (dataURLOptions?.maxBase64ImageLength) {
+    return checkDataURLSize(dataURL, dataURLOptions.maxBase64ImageLength);
+  }
+  return dataURL;
+}
+
 export function transformAttribute(
   doc: Document,
   tagName: Lowercase<string>,
@@ -201,34 +211,35 @@ export function transformAttribute(
     // href starts with a # is an id pointer for svg
     const transformedValue = absoluteToDoc(doc, value);
 
-    if (tagName === 'img' && transformedValue.startsWith('data:') && element) {
-      const img = element as HTMLImageElement;
+    if (transformedValue.startsWith('data:') && element) {
+      if (tagName === 'img') {
+        let processedDataURL = transformedValue;
 
-      let processedDataURL = transformedValue;
+        if (dataURLOptions?.type || dataURLOptions?.quality !== undefined) {
+          processedDataURL = recompressBase64Image(
+            element as HTMLImageElement,
+            transformedValue,
+            dataURLOptions.type,
+            dataURLOptions.quality,
+          );
+        }
 
-      if (dataURLOptions?.type || dataURLOptions?.quality !== undefined) {
-        processedDataURL = recompressBase64Image(
-          img,
-          transformedValue,
-          dataURLOptions.type,
-          dataURLOptions.quality,
-        );
+        return capDataURLSize(processedDataURL, dataURLOptions);
       }
 
-      if (dataURLOptions?.maxBase64ImageLength) {
-        processedDataURL = checkDataURLSize(
-          processedDataURL,
-          dataURLOptions.maxBase64ImageLength,
-        );
+      if (tagName === 'image') {
+        return capDataURLSize(transformedValue, dataURLOptions);
       }
-
-      return processedDataURL;
     }
 
     return transformedValue;
   } else if (name === 'xlink:href' && value[0] !== '#') {
     // xlink:href starts with # is an id pointer
-    return absoluteToDoc(doc, value);
+    const transformedValue = absoluteToDoc(doc, value);
+    if (transformedValue.startsWith('data:')) {
+      return capDataURLSize(transformedValue, dataURLOptions);
+    }
+    return transformedValue;
   } else if (
     name === 'background' &&
     (tagName === 'table' || tagName === 'td' || tagName === 'th')
--- a/packages/rrweb/rrweb-snapshot/test/integration.test.ts
+++ b/packages/rrweb/rrweb-snapshot/test/integration.test.ts
@@ -612,6 +612,92 @@ describe('base64 image compression tests', function (this: ISuite) {
 
     await page.close();
   });
+
+  it.each([
+    { id: 'svg-large-href-image', attr: 'href' },
+    { id: 'svg-large-xlink-image', attr: 'xlink:href' },
+  ])(
+    'should replace oversized SVG <image> $attr data URIs with striped placeholder',
+    async ({ id, attr }) => {
+      const page: puppeteer.Page = await browser.newPage();
+      await page.goto(`${serverURL}/html/base64-images.html`, {
+        waitUntil: 'load',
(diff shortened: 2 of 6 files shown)


packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -201,34 +211,35 @@ export function transformAttribute(
...
-
-      return processedDataURL;
     }
 
     return transformedValue;
   } else if (name === 'xlink:href' && value[0] !== '#') {
     // xlink:href starts with # is an id pointer
-    return absoluteToDoc(doc, value);
+    const transformedValue = absoluteToDoc(doc, value);
+    if (transformedValue.startsWith('data:')) {
```

**[convergent: paul + xp + qa-team]** 🟡 MEDIUM

three reviewers independently flagged the asymmetry between the two `data:` paths. the `href`/`src` branch (line 214) caps only when `tagName` is `img`/`image` **and** `element` is truthy; this `xlink:href` branch caps **any** tag with a `data:` value and has no `element` guard.

since `capDataURLSize` never touches `element`, the `&& element` guard on the href path is only load-bearing for the `` recompress call — so an SVG `<image>` passed without `element` would skip the cap on the `href` path while `xlink:href` would still cap it. not reachable today (every real caller passes `element`), and the broader xlink capping is arguably more correct, but the two paths currently tell different stories.

worth pinning the intent: either gate the xlink branch on `tagName === 'image'` too, or hoist the `image` cap out of the `&& element` guard so both attributes share one rule.

packages/rrweb/rrweb-snapshot/test/integration.test.ts
```diff
@@ -612,6 +612,92 @@ describe('base64 image compression tests', function (this: ISuite) {
...
+      waitUntil: 'load',
+    });
+    await page.waitForSelector('#svg-small-image', { timeout: 2000 });
+
+    const result = await page.evaluate(`${code}
+      const snapshot = rrwebSnapshot.snapshot(document, {
+        dataURLOptions: { type: "image/webp", quality: 0.4, maxBase64ImageLength: 1048576 }
+      });
+
+      function findByTag(node, tagName, acc = []) {
```
pauldambra, line 674: > [!NOTE]

**[xp]** 🟢 LOW

the `findByTag` recursive walker is duplicated verbatim across the two new `page.evaluate` blocks (also at line 633), embedded inside template-literal strings where the type checker can't catch drift. second occurrence — ThreeStrikesAndYouRefactor says allow it, but if a third copy appears, hoist it to a shared `const findByTagSrc` string reused in both evaluate calls.

pauldambra (commented):
QA Swarm multi-perspective review — see inline comments. Summary posted separately.

pauldambra:
> [!NOTE]
>

Tight, well-scoped change. It reuses the existing `maxBase64ImageLength` cap (same default) rather than inventing a new threshold, routes everything through a clean `capDataURLSize` extraction, and correctly resists recompressing SVG `<image>` (no canvas decode path for `SVGImageElement`). Test coverage is strong — parameterized jsdom units over capped/uncapped/no-config for both `href` and `xlink:href`, plus puppeteer integration on real canvas-generated data URIs.

- 🟡 **`href` vs `xlink:href` asymmetry** — flagged independently by **paul + xp + qa-team**. The `href` path caps only `img`/`image` and is gated on `element`; the `xlink:href` path caps any tag with a `data:` value and has no `element` guard. Not reachable today (every real caller passes `element`), but the two paths tell different stories. See inline comment for the fix options. This is the one worth a decision.

### Other findings (non-blocking)

- 🟢 **LOW** — `findByTag` walker duplicated verbatim across the two new `page.evaluate` blocks; hoist if a third copy appears.
- ⚪ **NIT** — `capDataURLSize` reads against `maxBase64ImageLength` which is a string-length, not byte size; `capDataURLLength` would mirror the option name. Leave if "size" is established vocabulary.
- ⚪ **NIT** — the oversized-SVG integration test passes `type`/`quality` options that are inert for `SVGImageElement`; the unit test already asserts the cap purely via `maxBase64ImageLength`.
- ℹ️ **rollout note (paul)** — every capped image becomes an unrecoverable striped placeholder, and SVG `<image>` is a slightly different population than ``; worth a heads-up to anyone watching replay-quality dashboards.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Approve with nits — sync string-length cap, no hot-path cost regression, both call sites covered; only the `element`-guard asymmetry and slightly-broader xlink scope to note. |
| 👤 paul | Solid and genuinely well-tested; reusing the existing cap is the safe-rollout move. One real question: the href/xlink asymmetry — pin the intent. |
| 📐 xp | Tidy, well-scoped; `capDataURLSize` is the right extraction and the ``/`<image>` split honestly avoids the wrong abstraction. Tests exemplary. |
| 🛡 security-audit | No findings — pure synchronous string-length cap returning a constant placeholder; no regex (no ReDoS), no input-derived sink (no XSS), strictly reduces recorded data. |

---

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -201,34 +211,35 @@ export function transformAttribute(
...
+      return capDataURLSize(transformedValue, dataURLOptions);
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `45feaf68` — gated the `xlink:href` cap on `tagName === 'image'` so legacy SVG links and `<use>` data refs are no longer rewritten to the image placeholder. Added a parameterized regression test asserting `<use>`/`<a>` `xlink:href` (and `<use href>`) data URIs are left untouched.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -201,34 +211,35 @@ export function transformAttribute(
...
-        processedDataURL = checkDataURLSize(
-          processedDataURL,
-          dataURLOptions.maxBase64ImageLength,
-        );
+      if (tagName === 'image') {
+        return capDataURLSize(transformedValue, dataURLOptions);
       }
-
-      return processedDataURL;
     }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `45feaf68` — moved the `&& element` guard into the `img` branch (it was only ever load-bearing for `recompressBase64Image`). The SVG `<image>` cap now applies even when `transformAttribute` is called without an element, matching the `xlink:href` path.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
pauldambra: > [!NOTE]

Fixed in `45feaf68`. Pinned the intent: both attribute paths now tell one story — cap `data:` URIs on SVG `<image>` only, and the ``-specific `element` guard lives inside the `img` branch (it's only needed for recompression). External reviewers (greptile, codex) converged on the same fix; the `xlink:href` branch is now gated on `tagName === 'image'` too.

packages/rrweb/rrweb-snapshot/test/integration.test.ts
pauldambra: > [!NOTE]

Acknowledged as a NIT — leaving the duplicated `findByTag` walker as-is per ThreeStrikesAndYouRefactor (two copies). Will hoist to a shared string constant if a third `page.evaluate` walker appears.

TueHaulund (approved):

PR #3917, opened 2026-06-21
fix(logs): emit posthogDistinctId from captureConsoleLogs so Logs auto-link to person

### LINKED ISSUE #3916: Logs: `captureConsoleLogs` stamps person id as `distinct_id`, but "Link to person" default key is `posthogDistinctId` — console logs do not link out of the box
### Summary

The two browser logging paths emit the person identifier under **different attribute keys**, and only one of them matches the PostHog Logs "Link to person" default. As a result, logs captured via `captureConsoleLogs` do **not** link to person profiles unless you manually override the link key — even though the docs/default imply it should work automatically.

### Two SDK paths, two keys

- **`captureConsoleLogs`** (the externally-loaded logs extension) hardcodes the key as **`distinct_id`** — `packages/browser/src/entrypoints/logs.ts` (~line 375):

```js
logger.emit({
  severityText: SEVERITY_MAP[level],
  body,
  attributes: {
    'log.source': `console.${level}`,
    distinct_id: posthog.get_distinct_id(),   // ← hardcoded `distinct_id`
    'location.href': assignableWindow.location.href,
    ...logAttributes,
  },
})
```

It is a literal object key — not configurable.

- **`posthog.logger.*` / `captureLog`** (the explicit logger API, in the main bundle) emits **`posthogDistinctId`**:

```js
e.distinctId && (o.posthogDistinctId = e.distinctId)
```

### Backend default

In `PostHog/posthog`, the Logs "Link to person" default is `products/logs/backend/models.py` → `DEFAULT_LOGS_DISTINCT_ID_ATTRIBUTE_KEY = "posthogDistinctId"`, with a test (`products/logs/backend/test/test_models.py`) asserting it and commenting *"Matches the JS SDK / docs convention so logs from posthog-js are linked to [persons]."* — which holds only for the `posthog.logger.*` path, **not** for `captureConsoleLogs`.

### Repro

1. `posthog.init(KEY, { logs: { captureConsoleLogs: true } })`.
2. Leave the project's Logs "Link to person" key at its default (`posthogDistinctId`).
3. Trigger a `console.warn(...)` and open the person's profile → **Logs** tab.

### Expected

Console-captured logs link to the person under the default config (as the docs / `test_models.py` comment imply).

### Actual

They don't. The records are stored with `distinct_id`, but the default link key matches `posthogDistinctId`, so nothing resolves. The only fix is to override `logs_distinct_id_attribute_key` to `distinct_id` via `logs_config` — at which point the *explicit* `posthog.logger.*` logs (which use `posthogDistinctId`) stop matching instead. There is **no single key value that links both paths**.

### Suggested fix (in rough order of preference)

1. **Make the SDK read the configured key from remote config.** The logs feature already consumes remote config (`onRemoteConfig` → `t.logs.captureConsoleLogs`). Including the project's `logs_distinct_id_attribute_key` in that payload and emitting the person id under it would make SDK output and the backend "Link to person" key **consistent by construction** — no drift possible, zero per-project setup. This is the most robust fix and reuses existing plumbing.
2. **Expose the key as an SDK option**, e.g. `logs: { distinctIdAttributeKey: 'posthogDistinctId' }`, so apps can align it with their configured link key (and with the `posthog.logger.*` path) explicitly.
3. **Align `captureConsoleLogs` to emit `posthogDistinctId`** — match `captureLog` and the backend default so console logs link with zero config. Simplest, but a breaking change for anyone who already worked around this by setting the link key to `distinct_id`.
4. **At minimum, document the mismatch** in the Logs "Link to person" docs and the `captureConsoleLogs` option, since today the two browser logging paths silently use different keys and only one matches the default.

### Environment

- `posthog-js` 1.386.8 (source verified against `main`)
- Logs product, OTLP ingestion, "Link to person" enabled

tsushanth:
Fixes #3916.

## Problem

`captureConsoleLogs` (the externally-loaded logs extension) stamps the person identifier under `distinct_id`, while every other logs path in the SDK uses `posthogDistinctId`:

- The OTLP record builder in [`@posthog/core/src/logs/logs-utils.ts:122`](https://github.com/PostHog/posthog-js/blob/main/packages/core/src/logs/logs-utils.ts#L122) — `autoAttributes.posthogDistinctId = sdkContext.distinctId`
- The browser `posthog.logger.*` / `captureLog` path
- React Native, Node, Web Worker — same key

The PostHog backend Logs "Link to person" default is also `posthogDistinctId` — [`products/logs/backend/models.py`](https://github.com/PostHog/posthog/blob/master/products/logs/backend/models.py) → `DEFAULT_LOGS_DISTINCT_ID_ATTRIBUTE_KEY = "posthogDistinctId"`, with a test in `products/logs/backend/test/test_models.py` commenting *"Matches the JS SDK / docs convention so logs from posthog-js are linked to [persons]."* — which holds only for the `posthog.logger.*` path, **not** for `captureConsoleLogs`.

Net effect today: logs captured via `captureConsoleLogs` never auto-link to a profile out of the box. The user has to manually change the project's link key, or move every site to `posthog.logger.*`.

## Fix

Switch the attribute key emitted at [`packages/browser/src/entrypoints/logs.ts:375`](https://github.com/PostHog/posthog-js/blob/main/packages/browser/src/entrypoints/logs.ts#L375) from `distinct_id` to `posthogDistinctId` so the console-logs path matches every other SDK path and the backend default.

```diff
-                        distinct_id: posthog.get_distinct_id(),
+                        posthogDistinctId: posthog.get_distinct_id(),
```

Updates the golden test's exact-record assertions accordingly. The 111 unit tests under `entrypoints/logs.{test,golden.test}.ts` + `posthog-logs.test.ts` pass.

## Risk

- **Backward compat:** Anyone whose Link-to-person key was already overridden to `distinct_id` to work around this bug will need to revert it back to the default `posthogDistinctId` (or keep the override). Anyone using the default (the documented happy path) gets linking working for the first time.
- **No public API surface change** — the attribute is emitted by an extension and only read server-side.

## Notes for maintainers

I left the rest of the divergence between `captureConsoleLogs` (which still uses `location.href`) and the OTLP builder (`url.full`) untouched — it's a separate concern from the linking bug the reporter filed, and the OTLP spec name (`url.full`) for the canonical browser path would be its own change with its own backend coordination. Happy to follow up if you'd like that aligned too.

### DIFF
--- a/packages/browser/src/entrypoints/logs.ts
+++ b/packages/browser/src/entrypoints/logs.ts
@@ -372,7 +372,13 @@ const initializeLogs = (posthog: PostHog) => {
                     body: body,
                     attributes: {
                         'log.source': `console.${level}`,
-                        distinct_id: posthog.get_distinct_id(),
+                        // `posthogDistinctId` matches the OTLP attribute key emitted by the
+                        // canonical `posthog.logger.*` / `captureLog` path (see
+                        // `@posthog/core/src/logs/logs-utils.ts`) and the backend default
+                        // `DEFAULT_LOGS_DISTINCT_ID_ATTRIBUTE_KEY` used by Logs → "Link
+                        // to person". Without this, console logs never auto-link to a
+                        // profile even though the docs/default imply they should.
+                        posthogDistinctId: posthog.get_distinct_id(),
                         'location.href': assignableWindow.location.href,
                         ...logAttributes,
                         ...(isObject(args[0]) ? flattenObject(args[0]) : {}),
--- a/.changeset/captureconsolelogs-posthog-distinct-id.md
+++ b/.changeset/captureconsolelogs-posthog-distinct-id.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fix `captureConsoleLogs` so console logs auto-link to person profiles in PostHog Logs. The browser console-logs entrypoint was emitting the person identifier under `distinct_id`, while every other logs path in the SDK (`posthog.logger.*` / `captureLog`, React Native, Node, the OTLP record builder in `@posthog/core`) emits `posthogDistinctId` — which is also the backend default for "Link to person". The mismatched key meant the console-logs path never linked out of the box. The browser path now emits `posthogDistinctId` to match the rest of the SDK and the backend.
--- a/packages/browser/src/__tests__/entrypoints/logs.golden.test.ts
+++ b/packages/browser/src/__tests__/entrypoints/logs.golden.test.ts
@@ -118,7 +118,7 @@ describe('logs entrypoint — golden (current console-capture wire output)', () 
             body: '"hello"',
             attributes: {
                 'log.source': 'console.log',
-                distinct_id: 'user-123',
+                posthogDistinctId: 'user-123',
                 'location.href': 'https://example.com/test',
                 sessionStartTimestamp: String(SESSION_START),
                 lastActivityTimestamp: String(LAST_ACTIVITY),
@@ -151,7 +151,7 @@ describe('logs entrypoint — golden (current console-capture wire output)', () 
             body: '{"user":{"id":5},"msg":"hi"}',
             attributes: {
                 'log.source': 'console.warn',
-                distinct_id: 'user-123',
+                posthogDistinctId: 'user-123',
                 'location.href': 'https://example.com/test',
                 sessionStartTimestamp: String(SESSION_START),
                 lastActivityTimestamp: String(LAST_ACTIVITY),


packages/browser/src/entrypoints/logs.ts
```diff
@@ -372,7 +372,13 @@ const initializeLogs = (posthog: PostHog) => {
...
                     attributes: {
                         'log.source': `console.${level}`,
-                        distinct_id: posthog.get_distinct_id(),
+                        // `posthogDistinctId` matches the OTLP attribute key emitted by the
+                        // canonical `posthog.logger.*` / `captureLog` path (see
+                        // `@posthog/core/src/logs/logs-utils.ts`) and the backend default
+                        // `DEFAULT_LOGS_DISTINCT_ID_ATTRIBUTE_KEY` used by Logs → "Link
+                        // to person". Without this, console logs never auto-link to a
+                        // profile even though the docs/default imply they should.
+                        posthogDistinctId: posthog.get_distinct_id(),
```
marandaneto, line 381: @turnipdabeets @PostHog/logs should we send both or just `posthogDistinctId`?

marandaneto:
@tsushanth thanks
pls add a changeset entry https://github.com/PostHog/posthog-js/blob/main/RELEASING.md

tsushanth:
Thanks @marandaneto — added [the changeset](https://github.com/PostHog/posthog-js/pull/3917/commits/8a88e85) for `posthog-js: patch`. Same content as the PR body summary, sized to fit a release-notes bullet.

turnipdabeets:
@tsushanth closing this PR since I already have a PR open that has the fix for this plus a few other things: https://github.com/PostHog/posthog-js/pull/3895

tsushanth:
Thanks @turnipdabeets — saw #3895 landed, glad the OTel removal is in. Closing the loop on this PR was the right call.

PR #3919, opened 2026-06-22
feat(replay): opt-in streaming reader for network body size limit

pauldambra:
## Problem

Session replay enforces a network body payload limit (`payloadSizeLimitBytes`, capped at 1MB), but today it does so **after** reading the entire body: `_tryReadBody` buffers the whole request/response body into a string via `clone().text()`, and only then does `config.ts` check the size and replace an over-limit body with a placeholder. For a very large body that means fully materialising it in memory before discarding it — and for **request** bodies the read is awaited *before* the real `fetch`, so it adds work on the page's critical path.

Network capture is the one area where a recording bug becomes a customer-facing outage, so we want a more careful body-limit mechanism, rolled out as conservatively as possible.

## Changes

Add an opt-in **streaming reader** (`_tryReadBodyStreaming`) that reads the body chunk by chunk off a `clone().body` reader and stops as soon as the running total exceeds the limit, so an oversized body is never fully buffered.

A **content-length pre-check** runs first: if the body's `content-length` header already declares it over the limit, we emit the too-large placeholder without cloning or reading the body at all. This trusts `content-length` the same way `config.ts` `enforcePayloadSizeLimit` already does. The streaming reader is the fallback for bodies with no/under-stated content-length (chunked, or a header that lies).

The reader keeps every observer-only invariant of the existing buffered path:

- reads **only a clone** of the body — never the stream the page itself consumes
- **always resolves, never rejects** (success, reader error, sync throw, and a 500ms timeout all resolve) so it can never reject the page's `fetch`
- **feature-detects** `ReadableStream.getReader` / `TextDecoder` and falls back to the buffered `clone().text()` read when they're absent
- returns a placeholder string (not a partial body) when over the limit

**Rollout — defaults-gated, no remote toggle.** The behaviour is off by default and turned on through the date-versioned `defaults` mechanism at `2026-05-30` (already adopted by the PostHog app, so we dogfood it). Existing customers are unaffected until they bump their `defaults` date; it can also be set directly via `session_recording.streamNetworkBody`. There is deliberately no server/remote switch.

The buffered path is unchanged and remains the default, so this is additive.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Paul flagged that network-capture body handling is the highest-risk area (anything that alters the body or breaks the request can break customers' sites) and asked for the streaming body-limit reader to ship behind the `defaults` flag for the slowest possible opt-in rollout, with no remote change — then asked to check `content-length` first so we never read a body that's already definitely over the limit.

Before writing anything, traced the capture path to confirm the safety boundary (the real `res` is returned before any capture work; capture runs in the `finally` `.then().catch()`; bodies are read from a clone; `isHostOnDenyList` URL parsing is throw-safe). The streaming reader and the content-length pre-check are built to preserve those same invariants, and are wired into the existing `2026-05-30` `session_recording` defaults tier so they dogfood on the PostHog app while every existing customer keeps the buffered path until they opt in. Tests cover the reader directly (under/at/over limit, no-stream fallback, clone failure, reader-error-never-rejects), the content-length pre-check (parameterized over header values, plus "skips the read when over" / "still reads when under" / "ignored when the flag is off"), the config→network-options flow, and the defaults-date gating. The `getEventUuid` typecheck/test noise seen locally was stale workspace `dist` for `@posthog/core` / `@posthog/types`, not part of this change.

Tools: PostHog Code (Claude Opus).

---

### DIFF
--- a/.changeset/replay-stream-network-body-limit.md
+++ b/.changeset/replay-stream-network-body-limit.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Session replay network capture: add an opt-in streaming reader for request/response bodies that stops at the payload size limit instead of buffering the whole body and then discarding it — bounding memory and pre-request latency when a body is very large. It reads only a clone of the body, so it never consumes the stream the page itself reads, and always resolves (never rejects) into the page's `fetch`. Off by default; enabled for `defaults: '2026-05-30'` and settable directly via `session_recording.streamNetworkBody`.
--- a/.changeset/replay-stream-network-body-limit.md
+++ b/.changeset/replay-stream-network-body-limit.md
@@ -1,5 +1,6 @@
 ---
 'posthog-js': patch
+'@posthog/types': patch
 ---
 
 Session replay network capture: add an opt-in streaming reader for request/response bodies that stops at the payload size limit instead of buffering the whole body and then discarding it — bounding memory and pre-request latency when a body is very large. It reads only a clone of the body, so it never consumes the stream the page itself reads, and always resolves (never rejects) into the page's `fetch`. Off by default; enabled for `defaults: '2026-05-30'` and settable directly via `session_recording.streamNetworkBody`.
--- a/.changeset/replay-stream-network-body-limit.md
+++ b/.changeset/replay-stream-network-body-limit.md
@@ -3,4 +3,4 @@
 '@posthog/types': patch
 ---
 
-Session replay network capture: add an opt-in streaming reader for request/response bodies that stops at the payload size limit instead of buffering the whole body and then discarding it — bounding memory and pre-request latency when a body is very large. It reads only a clone of the body, so it never consumes the stream the page itself reads, and always resolves (never rejects) into the page's `fetch`. Off by default; enabled for `defaults: '2026-05-30'` and settable directly via `session_recording.streamNetworkBody`.
+Session replay network capture: add an opt-in streaming reader for request/response bodies that stops at the payload size limit instead of buffering the whole body and then discarding it — bounding memory and pre-request latency when a body is very large. It reads only a clone of the body, so it never consumes the stream the page itself reads, and always resolves (never rejects) into the page's `fetch`. Off by default; enabled for `defaults: '2026-06-25'` and settable directly via `session_recording.streamNetworkBody`.
--- a/packages/browser/src/extensions/replay/external/network-plugin.ts
+++ b/packages/browser/src/extensions/replay/external/network-plugin.ts
@@ -651,10 +651,14 @@ export function _tryReadBodyStreaming(r: Request | Response, limitBytes: number)
         const body = clone.body
         // no readable stream (or no TextDecoder) available — fall back to the buffered read of the clone
         if (!isReadableStreamBody(body) || typeof TextDecoder === 'undefined') {
-            clone.text().then(
-                (txt) => done(txt),
-                (reason) => done(bodyReadFailedMessage(reason))
-            )
+            try {
+                clone.text().then(
+                    (txt) => done(txt),
+                    (reason) => done(bodyReadFailedMessage(reason))
+                )
+            } catch {
+                done(BODY_READ_FAILED_MESSAGE)
+            }
             return
         }
 
--- a/packages/browser/src/extensions/replay/external/network-plugin.ts
+++ b/packages/browser/src/extensions/replay/external/network-plugin.ts
@@ -31,7 +31,7 @@ import { createLogger } from '../../../utils/logger'
 import { formDataToQuery } from '../../../utils/request-utils'
 import { patch } from '../rrweb-plugins/patch'
 import { isHostOnDenyList } from '../../../extensions/replay/external/denylist'
-import { defaultNetworkOptions, MAX_PAYLOAD_SIZE_BYTES } from './config'
+import { defaultNetworkOptions, effectivePayloadLimitBytes } from './config'
 
 const logger = createLogger('[Recorder]')
 
@@ -549,10 +549,6 @@ function bodyTooLargeMessage(limitBytes: number): string {
     return `[SessionReplay] Body too large to record (> ${limitBytes} bytes)`
 }
 
-function effectivePayloadLimitBytes(options: NetworkRecordOptions): number {
-    return Math.min(MAX_PAYLOAD_SIZE_BYTES, options.payloadSizeLimitBytes ?? MAX_PAYLOAD_SIZE_BYTES)
-}
-
 // when the body declares a content-length over the limit we can skip reading it entirely.
 // this trusts content-length the same way config.ts enforcePayloadSizeLimit does.
 // known accepted risk: for a compressed response content-length is the compressed size while the
@@ -697,7 +693,7 @@ export function _tryReadBodyStreaming(r: Request | Response, limitBytes: number)
 
                     pump()
                 },
-                (reason) => done('[SessionReplay] Failed to read body: ' + reason)
+                (reason) => done(bodyReadFailedMessage(reason))
             )
         }
 
--- a/packages/browser/src/extensions/replay/external/network-plugin.ts
+++ b/packages/browser/src/extensions/replay/external/network-plugin.ts
@@ -536,6 +536,17 @@ function isReadableStreamBody(body: unknown): body is ReadableStream<Uint8Array>
     return isObject(body) && isFunction(body.getReader) && isFunction(body.tee)
 }
 
+// matches the hard ceiling enforced in config.ts limitPayloadSize
+const MAX_PAYLOAD_SIZE_BYTES = 1000000
+
+function effectivePayloadLimitBytes(options: NetworkRecordOptions): number {
+    return Math.min(MAX_PAYLOAD_SIZE_BYTES, options.payloadSizeLimitBytes ?? MAX_PAYLOAD_SIZE_BYTES)
+}
+
+function _readBody(r: Request | Response, options: NetworkRecordOptions): Promise<string> {
+    return options.streamNetworkBody ? _tryReadBodyStreaming(r, effectivePayloadLimitBytes(options)) : _tryReadBody(r)
+}
+
 function _tryReadBody(r: Request | Response): Promise<string> {
     // there are now already multiple places where we're using Promise...
     // eslint-disable-next-line compat/compat
@@ -556,6 +567,95 @@ function _tryReadBody(r: Request | Response): Promise<string> {
     })
 }
 
+// Reads a clone of the body chunk by chunk, stopping as soon as the running total exceeds the
+// limit, so a very large body is never fully buffered. Like _tryReadBody it only ever reads a
+// clone (never the stream the page consumes) and is guaranteed to resolve, never reject.
+export function _tryReadBodyStreaming(r: Request | Response, limitBytes: number): Promise<string> {
+    // eslint-disable-next-line compat/compat
+    return new Promise((resolve) => {
+        let settled = false
+        const timeout = setTimeout(() => done('[SessionReplay] Timeout while trying to read body'), 500)
+
+        function done(value: string): void {
+            if (settled) {
+                return
+            }
+            settled = true
+            clearTimeout(timeout)
+            resolve(value)
+        }
+
+        let clone: Request | Response
+        try {
+            clone = r.clone()
+        } catch {
+            done('[SessionReplay] Failed to read body')
+            return
+        }
+
+        const body = clone.body
+        // no readable stream (or no TextDecoder) available — fall back to the buffered read of the clone
+        if (!body || !isFunction(body.getReader) || typeof TextDecoder === 'undefined') {
+            clone.text().then(
+                (txt) => done(txt),
+                (reason) => done('[SessionReplay] Failed to read body: ' + reason)
+            )
+            return
+        }
+
+        let reader: ReadableStreamDefaultReader<Uint8Array>
+        try {
+            reader = body.getReader()
+        } catch {
+            done('[SessionReplay] Failed to read body')
+            return
+        }
+
+        function cancel(): void {
+            try {
+                void reader.cancel()
+            } catch {
+                // the reader may already be released; nothing to clean up
+            }
+        }
+
+        const chunks: Uint8Array[] = []
+        let received = 0
+
+        function pump(): void {
+            reader.read().then(
+                ({ done: streamDone, value }) => {
+                    if (streamDone) {
+                        const merged = new Uint8Array(received)
+                        let offset = 0
+                        for (const chunk of chunks) {
+                            merged.set(chunk, offset)
+                            offset += chunk.byteLength
+                        }
+                        done(new TextDecoder().decode(merged))
+                        return
+                    }
+
+                    if (value) {
+                        if (received + value.byteLength > limitBytes) {
+                            cancel()
+                            done(`[SessionReplay] Body too large to record (> ${limitBytes} bytes)`)
+                            return
+                        }
+                        received += value.byteLength
+                        chunks.push(value)
+                    }
+
+                    pump()
+                },
+                (reason) => done('[SessionReplay] Failed to read body: ' + reason)
+            )
+        }
+
+        pump()
+    })
+}
+
 async function _tryReadRequestBody({
     r,
     options,
@@ -570,7 +670,7 @@ async function _tryReadRequestBody({
         return Promise.resolve(hostname + ' is in deny list')
     }
 
-    return _tryReadBody(r)
+    return _readBody(r, options)
 }
 
 async function _tryReadResponseBody({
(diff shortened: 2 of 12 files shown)


packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -536,6 +536,17 @@ function isReadableStreamBody(body: unknown): body is ReadableStream<Uint8Array>
     return isObject(body) && isFunction(body.getReader) && isFunction(body.tee)
 }
 
+// matches the hard ceiling enforced in config.ts limitPayloadSize
+const MAX_PAYLOAD_SIZE_BYTES = 1000000
```
marandaneto:
```suggestion
const MAX_PAYLOAD_SIZE_BYTES = 1000000 // 1 MB
```

pauldambra: not even my robot has reviewed it yet :)

packages/browser/src/types.ts
```diff
@@ -503,6 +507,12 @@ export type NetworkRecordOptions = {
      * NB this will be at most 1MB even if set larger
      */
     payloadSizeLimitBytes: number
+    /**
+     * when true, read bodies through a streaming reader that stops at payloadSizeLimitBytes
+     * instead of buffering the whole body and then enforcing the limit. Reads only a clone of
+     * the body, so it never consumes the stream the page itself reads.
```
marandaneto, line 509: set @default value similar to packages/types/src/posthog-config.ts

packages/browser/src/extensions/replay/external/network-plugin.ts
marandaneto: for you only the fastest gun in the east

marandaneto:
idea: check `Content-Length` headers so you know if its within the limits even before reading it

marandaneto (approved):

pauldambra:
> idea: check `Content-Length` headers so you know if its within the limits even before reading it

​Ah, so what i discovered while poking around in large recordings is that the content length is the compressed size, we store the uncompressed body, and we can't always detect that  
  
The intention is to check content size/length first as a short circuit for sure

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +567,95 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+                        }
+                        received += value.byteLength
+                        chunks.push(value)
+                    }
+
+                    pump()
+                },
+                (reason) => done('[SessionReplay] Failed to read body: ' + reason)
+            )
+        }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Good catch — fixed in `019fd5785`. `done()` now always cancels the reader, and `pump()` bails (and cancels) once `settled` is true, so a slow/hung stream stops being pulled and stops buffering the clone the moment we resolve (timeout, cap, success, or error). Added a fake-timers test asserting the hung-stream timeout cancels the reader.

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +567,95 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
 }
 
+// Reads a clone of the body chunk by chunk, stopping as soon as the running total exceeds the
+// limit, so a very large body is never fully buffered. Like _tryReadBody it only ever reads a
+// clone (never the stream the page consumes) and is guaranteed to resolve, never reject.
+export function _tryReadBodyStreaming(r: Request | Response, limitBytes: number): Promise<string> {
+    // eslint-disable-next-line compat/compat
+    return new Promise((resolve) => {
+        let settled = false
+        const timeout = setTimeout(() => done('[SessionReplay] Timeout while trying to read body'), 500)
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `019fd5785` — same fix as the other timeout thread: `done()` cancels the reader and `pump()` stops once settled, so the clone is no longer read/buffered after the 500ms timeout.

.changeset/replay-stream-network-body-limit.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Done in `019fd5785` — the changeset now bumps `@posthog/types` as well, since `SessionRecordingOptions.streamNetworkBody` ships from that package.

packages/browser/src/types.ts
pauldambra: > [!NOTE]

Added `@default false` in `019fd5785`, matching the field in `packages/types/src/posthog-config.ts`.

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +567,95 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+
+        const body = clone.body
+        // no readable stream (or no TextDecoder) available — fall back to the buffered read of the clone
+        if (!body || !isFunction(body.getReader) || typeof TextDecoder === 'undefined') {
+            clone.text().then(
+                (txt) => done(txt),
+                (reason) => done('[SessionReplay] Failed to read body: ' + reason)
+            )
+            return
+        }
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Intentional graceful degradation. The fallback only runs where there's no `ReadableStream.getReader`/`TextDecoder`, where `text()` must buffer the whole body anyway — so enforcing `limitBytes` there buys no memory win (the read has already happened). Two things still bound it: the `content-length` pre-check in `_readBody` runs before the streaming attempt and short-circuits header-declared-large bodies even in this env, and `limitPayloadSize` in config.ts caps the recorded payload downstream, so no oversized data is recorded. The only residual is a body with no/under-stated content-length in a no-streams environment, which degrades to the existing buffered behaviour.

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Folded the 1MB into the existing comment in `c47a4772e` (`// 1MB — matches the hard ceiling enforced in config.ts limitPayloadSize`) rather than adding a second trailing comment.

marandaneto:
> > idea: check `Content-Length` headers so you know if its within the limits even before reading it
> 
> ​Ah, so what i discovered while poking around in large recordings is that the content length is the compressed size, we store the uncompressed body, and we can't always detect that
> 
> The intention is to check content size/length first as a short circuit for sure

yes, thats the downside, i recall doing this + combined estimation depending on the compressing type which isnt 100% accurate but it definitely helped a bit, but only if this is still a bottleneck and we should improve further, all fine for now

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -536,6 +536,44 @@ function isReadableStreamBody(body: unknown): body is ReadableStream<Uint8Array>
...
+            return false
+        }
+        const contentLength = parseInt(headerValue, 10)
+        return Number.isFinite(contentLength) && contentLength > limitBytes
+    } catch {
+        return false
+    }
+}
+
+export function _readBody(r: Request | Response, options: NetworkRecordOptions): Promise<string> {
```
pauldambra, line 570: > [!NOTE]

**[paul]** 🟠 HIGH

this is the bit that makes my spider-sense tingle. we're turning on a brand new streaming reader in the highest-risk area we have (network capture) via the 2026-05-30 default, and there's no way to tell what it's doing in the wild. there are four distinct "we lost a body" exits — the content-length skip, the mid-stream cap, the `text()` fallback (no ReadableStream/TextDecoder), and the 500ms timeout — and every one is invisible right now. canonical reference is Knight Capital: measure first. before this rides the default to everyone, could we ship a counter/verbose-log on each of those four exits so we can see their shape in production first? even `logger` behind verbose + simple counters we can eyeball would do — not a metrics pipeline. (This is the judgement call for the author — flagged, not auto-fixable.)

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +594,104 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+            if (settled) {
+                return
+            }
+            settled = true
+            clearTimeout(timeout)
+            cancel()
+            resolve(value)
+        }
+
+        const timeout = setTimeout(() => done('[SessionReplay] Timeout while trying to read body'), 500)
```
pauldambra: > [!NOTE]

Duplication between `_tryReadBody` and `_tryReadBodyStreaming`: the `500`ms timeout and the placeholder strings (`Timeout while trying to read body`, `Failed to read body`) are spelled verbatim in both, and `MAX_PAYLOAD_SIZE_BYTES` re-states the `Math.min(1000000, ...)` that already lives in config.ts `limitPayloadSize`. The strings are load-bearing (tests assert on them) so a wording change has to be made twice. Also the two "too large" placeholders don't match: config.ts emits `[SessionRecording] Request body too large to record (N bytes)` while this path emits `[SessionReplay] Body too large to record (> limit bytes)` — different prefix, limit-not-size, no request/response label. Hoist the timeout to a named const, the messages to shared helpers (alongside `bodyTooLargeMessage`), and ideally share the limit so it's said OnceAndOnlyOnce.

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -536,6 +536,44 @@ function isReadableStreamBody(body: unknown): body is ReadableStream<Uint8Array>
...
+    }
+}
+
+export function _readBody(r: Request | Response, options: NetworkRecordOptions): Promise<string> {
+    if (!options.streamNetworkBody) {
+        return _tryReadBody(r)
+    }
+    const limitBytes = effectivePayloadLimitBytes(options)
+    // skip the read for bodies the headers already tell us are over the limit
+    if (_contentLengthExceedsLimit(r, limitBytes)) {
```
pauldambra, line 576: > [!NOTE]

The content-length pre-check compares the header against the limit, but for a compressed response (`Content-Encoding: gzip/br`) `content-length` is the *compressed* byte count while the streaming reader on `clone.body` yields *decoded* bytes. A response whose compressed length is over the limit but decodes to under it would be dropped with the too-large placeholder instead of recorded. It never breaks the page, and it matches the pre-existing content-length trust in config.ts, so it's acceptable — but worth a one-line note that the two byte counts aren't directly comparable so a future maintainer doesn't "optimise" by trusting the header harder.

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +594,104 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+        try {
+            clone = r.clone()
+        } catch {
+            done('[SessionReplay] Failed to read body')
+            return
+        }
+
+        const body = clone.body
+        // no readable stream (or no TextDecoder) available — fall back to the buffered read of the clone
+        if (!body || !isFunction(body.getReader) || typeof TextDecoder === 'undefined') {
```
pauldambra: > [!NOTE]

**[convergent: paul + xp + qa-team]** 🟢 LOW

`isReadableStreamBody` already exists right above (line 535) as a type guard for "is this a readable stream with getReader". The streaming path re-spells it inline as `!body || !isFunction(body.getReader)`. Reuse the existing guard so it's said once and the two checks can't silently drift — or, if the looser check is deliberate (you don't need `tee` here), a one-liner saying why.

packages/browser/src/extensions/replay/external/network-plugin.ts
```diff
@@ -556,6 +594,104 @@ function _tryReadBody(r: Request | Response): Promise<string> {
...
+            reader!.read().then(
+                ({ done: streamDone, value }) => {
+                    // already resolved (e.g. the timeout fired) — stop reading and release the stream
+                    if (settled) {
+                        cancel()
+                        return
+                    }
+
+                    if (streamDone) {
+                        const merged = new Uint8Array(received)
```
pauldambra: > [!NOTE]

**[xp]** 🟢 LOW

ComposedMethod: the `pump` closure does three things at one level — detect the settled race, merge-and-decode on completion, and enforce the cap per chunk. The merge block reads as its own idea; extracting `concatChunks(chunks, received): Uint8Array` would let `pump` scan like a table of contents. Cheapest readability win in a ~90-line function.

pauldambra (commented):
QA Swarm multi-perspective review — see inline comments. Summary posted separately.

pauldambra:
> [!NOTE]
>

The implementation is careful and well-tested — observer-only invariants are all respected (clone-only reads, always-resolve-never-reject, cancel-on-settle, the 500ms timeout, feature-detection with a `text()` fallback) and memory is tightly bounded (the chunk that would cross the cap is detected before buffering). Security audit found nothing. The "request changes" is one judgement call plus convergent cleanup, not a correctness bug.

### Key findings

- 🟠 **HIGH (paul) — ships blind.** This is the riskiest surface we have, and the four "we lost a body" exits (content-length skip, mid-stream cap, `text()` fallback, timeout) have no counters or logs. Measure-first: add lightweight instrumentation so we can see their shape in production before this rides the 2026-05-30 default to everyone. *(Author decision.)*
- 🟡 **MEDIUM — duplication & placeholder drift.** The 500ms timeout + placeholder strings are spelled twice; `MAX_PAYLOAD_SIZE_BYTES` restates config.ts's limit; the two "too large" placeholders use different prefixes/formats. Hoist to shared consts/helpers.
- 🟡 **MEDIUM — content-length is compressed bytes** but the reader counts decoded bytes; worth a clarifying note (harmless, matches config.ts trust).

- **Duplication** of timeout/placeholder strings — flagged by **paul + xp**.
- **Reuse `isReadableStreamBody`** instead of the inline `getReader` check — flagged by **paul + xp + qa-team**.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Approve with nits — invariants hold, memory bounded; one MEDIUM on compressed-vs-decoded content-length. |
| 👤 paul | High trust on the craft, but won't ship blind on the riskiest surface — wants measure-first instrumentation before the default rollout. |
| 📐 xp | Sound design (the reader split is right, not a flag-driven wrong abstraction); polish — hoist the duplicated timeout/strings, reuse the stream guard, extract the chunk-merge. |
| 🛡 security-audit | No findings — clone-only, always-resolves, bounded memory, content-length only ever skips (never captures more). |

---

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Won't add instrumentation — agreed with the author this isn't new risk. The buffered default path already produces the same "lost body" outcomes and the same placeholder strings (`Timeout…`, `Failed to read body`, and config.ts's `too large`); three of the four exits listed already exist today. The placeholders are themselves the in-recording signal (queryable in ingested recordings), and there's no clean per-exit counter channel from the SDK. The one genuinely new divergence (content-length skip on a compressed/over-stated header) is now documented as an accepted risk. Resolving.

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Fixed in `2c6fd48b6`. `MAX_PAYLOAD_SIZE_BYTES` now lives once in config.ts and is imported (also used by config's own limiter); `BODY_READ_TIMEOUT_MS` and the read placeholder strings are hoisted into shared consts/helpers used by both readers. Left config.ts's `[SessionRecording] … (N bytes)` message as-is — it's a long-standing recorded string (grep/query risk to change) and each file keeps its own established prefix; the new path stays internally consistent on `[SessionReplay]`.

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Documented in `2c6fd48b6` as a known, accepted risk on `_contentLengthExceedsLimit`: for a compressed response, content-length is the compressed size while the reader counts decoded bytes, so an over-compressed-but-under-decoded body drops to the placeholder. It never breaks the page, so we accept it.

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Done in `2c6fd48b6` — the streaming feature check now reuses the existing `isReadableStreamBody` guard (`!isReadableStreamBody(body) || typeof TextDecoder === 'undefined'`). It also requires `tee`, which every real ReadableStream has, so it's equivalent in practice and says the intent once.

packages/browser/src/extensions/replay/external/network-plugin.ts
pauldambra: > [!NOTE]

Done in `2c6fd48b6` — extracted `concatChunks(chunks, totalBytes)` so `pump` reads at one level of abstraction.

TueHaulund (approved):

PR #3921, opened 2026-06-22
feat: strip URL hashes from captured URLs

marandaneto:
## Problem

Automatically captured browser URLs can include fragment identifiers (`#...`). URL hashes may contain sensitive state or routing data and are not generally useful for analytics attribution.

## Changes

- Adds `disable_capture_url_hashes`, enabled by default, to strip URL fragments from automatically captured URL fields.
- Adds a shared `stripUrlHash()` utility in `@posthog/core` and reuses it across browser-derived URL capture paths.
- Applies the option to browser SDK event/session/person URLs, web vitals, heatmaps, logs, conversations, replay URL fields, autocaptured href attributes, posthog-js-lite context, and Next Pages Router pageview URLs.
- Leaves places that already did not capture hashes unchanged, such as Next App Router pageviews and the replay `$url_changed` fallback.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

An agent implemented the requested SDK option in a dedicated worktree, auditing URL capture paths and applying hash stripping only where hashes were previously captured. During implementation we avoided adding hashes to paths that already omitted them, such as Next App Router pageviews.

Validation run:

```bash
pnpm --filter posthog-js exec jest src/__tests__/utils/event-utils.test.ts src/__tests__/autocapture.test.ts
pnpm --filter @posthog/next exec jest tests/PagesPostHogPageView.test.tsx tests/PostHogPageView.test.tsx
pnpm --filter posthog-js typecheck
pnpm --filter posthog-js build
pnpm --filter posthog-js-lite build
pnpm --filter @posthog/next build
```


packages/browser/src/utils/event-utils.ts
```diff
@@ -281,7 +293,8 @@ export function getBrowserDetectionHints(): BrowserDetectionHints {
 export function getEventProperties(
     maskPersonalDataProperties?: boolean,
     customPersonalDataProperties?: string[],
-    detectGoogleSearchApp?: boolean
+    detectGoogleSearchApp?: boolean,
+    disable_capture_url_hashes: boolean = true
 ): Properties {
```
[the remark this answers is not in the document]
marandaneto: Fixed in a02cb9ce0 — renamed the helper parameter to camelCase to match the surrounding signature.

marandaneto:
Addressed the Greptile feedback in a02cb9ce0: renamed the  parameter to camelCase and collapsed the duplicated hash tests into parameterized cases.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -653,6 +655,19 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+        return this._instance.config.disable_capture_url_hashes ? stripUrlHash(url) : url
+    }
+
+    private _maskReplayUrl(url: string, forceStripHash: boolean = false): string | undefined {
+        const href = forceStripHash ? stripUrlHash(url) : this._stripUrlHash(url)
+        const paramsToMask = this._instance.config.mask_personal_data_properties
+            ? [...PERSONAL_DATA_CAMPAIGN_PARAMS, ...(this._instance.config.custom_personal_data_properties || [])]
+            : []
+
+        return this._maskUrl(maskQueryParams(href, paramsToMask, MASKED))
```
marandaneto, line 668: this is technically a separate issue since we were not running the recording metadata event href prop through `maskQueryParams` and this could have leaked sensitive data
now it does, but its a breaking change
i dont know if this would break recordings though
 cc @PostHog/team-replay

marandaneto:
see discussion https://posthog.slack.com/archives/C04MZFDA5KK/p1782031268596399

some places were already stripping hashes
now all places strip hashes by default
additional replay fix here https://github.com/PostHog/posthog-js/pull/3921#discussion_r3451854237
people can turn it back on
people can use the before send calback to strip manoelly
its a minor with a breaking change due to security reasons, not a new defaults

marandaneto:
TLDR:

We can use heuristics, but they wouldn’t be reliably safe.

 A URL hash has no browser-level meaning that tells us whether it’s an SPA route or sensitive
 state. Examples:

 - SPA route: https://app.com/#/settings
 - Anchor: https://site.com/docs#install
 - OAuth/token: https://app.com/#access_token=...
 - App state/PII: https://app.com/#email=user@example.com
 - Router state: https://app.com/#/users/123?tab=billing

 A heuristic like “keep hashes starting with #/” would preserve many SPA routes, but SPA routes
 can still include IDs, emails, tokens, query params, or other sensitive data. Also, not all SPAs
 use #/; some use plain #route, custom formats, or normal history routing.

 So any automatic heuristic risks leaking data while only being partially correct. Safer behavior
 is:

 - strip hashes by default
 - let users opt back in with disable_capture_url_hashes: false
 - if they need selective capture, use before_send with app-specific knowledge to redact or keep
   exactly what’s safe.

.changeset/quiet-ravens-strip.md
```diff
@@ -0,0 +1,11 @@
...
+'posthog-js': minor
+'@posthog/core': minor
+'@posthog/types': minor
+'posthog-js-lite': minor
+'@posthog/next': minor
+---
+
+Add `disable_capture_url_hashes`, enabled by default, to strip URL fragments from automatically captured URLs. This can be a behavior change for projects that previously relied on hashes being captured in URL fields such as `$current_url`, `$initial_current_url`, `$session_entry_url`, replay URLs, heatmaps, web vitals, logs, conversations, or Next.js Pages Router pageviews.
+
+To keep capturing hashes, set `disable_capture_url_hashes: false`. If you only want to capture some hashes, leave hash capture enabled and use `before_send` to remove or redact sensitive hash values before events are sent.
```
ioannisj: I think we'll need a more prominent WARN here that SPAs will now see routes collapsed to bare original after this version?

packages/core/src/utils/index.ts
```diff
@@ -38,6 +38,14 @@ export function removeTrailingSlash(url: string): string {
   return url?.replace(/\/+$/, '')
 }
 
+export function stripUrlHash<T extends string | undefined>(url: T): T extends string ? string : undefined {
```
ioannisj, line 41: Should we cover this util function with tests?

packages/browser/src/utils/event-utils.ts
```diff
@@ -209,20 +209,29 @@ export function getReferrerInfo(): Record<string, any> {
     }
 }
 
-export function getPersonInfo(maskPersonalDataProperties?: boolean, customPersonalDataProperties?: string[]) {
+export function getPersonInfo(
```
ioannisj, line 212: Will this rewrite existing $initial_current_url on update?

packages/core/src/utils/index.ts
marandaneto: Added targeted `stripUrlHash` coverage in 3f2a6c1e7, including hash routes, query strings before hashes, multiple hashes, empty strings, and undefined.

packages/browser/src/utils/event-utils.ts
marandaneto: Not in persistence or on already-set person properties, but the emitted $set_once value from existing local persistence will be hash-stripped by default

feliperalmeida (approved):
LGTM on the security side.

lricoy (approved):
Small nit: add to AGENTS.md that all URLs (href/location/etc...) capturing must use `stripUrlHash` now, so it helps us not shoot ourselves in the foot when adding/changing features.

marandaneto:
ok changes
it is disabled by default to keep back compatibility
it is enabled by default if defaults=2026-06-25

PR #3924, opened 2026-06-22
Add 'enabled' property to ModuleOptions interface and quietly exit when no key found

JamieCurnow:
## Problem

1. The Nuxt build succeeds but runtime throws a server error if no posthog public key is found. This resulted in a production build failure for us.

2. There should be an easy way to turn the module off to support applications with different build environments. Eg. enabled: false in dev/staging envs.

## Changes

1. Check for a public key to exist in the nuxt config and boot out with a console error if none exists.

2. Added an `enabled` flag to the nuxt config options for posthog. Defaults true so is backwards compatible, but if it's set to false the module quietly boots out early.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [ ] Tests for new code

### If releasing new changes

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — or — Fully autonomous

### DIFF
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -44,6 +44,7 @@ interface SourcemapsConfig {
 }
 
 export interface ModuleOptions {
+  enabled: boolean
   host: string
   publicKey: string
   debug?: boolean
@@ -71,6 +72,7 @@ export default defineNuxtModule<ModuleOptions>({
     },
   },
   defaults: () => ({
+    enabled: true,
     host: 'https://us.i.posthog.com',
     debug: false,
     clientConfig: {},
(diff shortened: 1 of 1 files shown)


JamieCurnow:
Sorry gang, just found and read this:
https://github.com/PostHog/posthog-js/pull/3057

I feel like it's still relevant to be able to disable posthog in certain envs, but workaround is sufficient. Closing for now.

PR #3925, opened 2026-06-23
feat(next): enable exception capture by default

dustinbyrne:
## Problem

Next.js apps using `@posthog/next` should get useful exception telemetry with minimal setup, covering both browser-side errors and server request errors from Next.js instrumentation.

Server-side request error capture also needs to avoid sending potentially sensitive URL query strings or fragments as event properties.

## Changes

- Enable browser-side exception capture by default in the Next.js providers while preserving explicit `clientOptions.capture_exceptions` overrides.
- Add `onRequestError` / `createOnRequestError` helpers for Next.js instrumentation request-error capture.
- Support Node and Edge runtimes with runtime-appropriate client caching and capture behavior.
- Respect PostHog opt-out cookie state and attach cookie-derived identity/session properties where available.
- Capture path-only `$pathname` values for server request errors, stripping query strings and hashes.
- Add unit, edge-runtime, packaging, and barrel-export coverage.
- Add an App Router example route/page and README notes for server-side error capture.

Validation run:

```bash
pnpm --filter @posthog/next test:unit -- onRequestError
pnpm --filter @posthog/next lint
pnpm --filter @posthog/next build
```

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

This PR was implemented with Pi agent assistance under human direction. The implementation was reviewed for PostHog SDK guideline alignment, with an explicit privacy decision to store only pathnames for server-side exception events rather than full URLs containing query strings or fragments.


packages/next/src/server/captureRequestError.ts
```diff
@@ -0,0 +1,172 @@
...
+        // eslint-disable-next-line compat/compat
+        const parsed = new URL(url, 'http://localhost')
+        return parsed.pathname
+    } catch {
+        return stripSearchAndHash(url)
+    }
+}
+
+function stripSearchAndHash(value: string): string {
+    return value.split(/[?#]/, 1)[0]
```
marandaneto: theres a new config for capturing hashes or not, lets put behind that config

marandaneto (approved):

PR #3942, opened 2026-06-24
fix(browser): route log session attributes through shared SDK context

hpouillot:
## Problem

The browser logs extension could throw during initialization when the session manager returned a null `lastActivityTimestamp`, causing `Cannot read properties of null (reading 'toString')`.

Separately, the console-logs entrypoint hand-rolled session attributes (`window.id`, `sessionStartTimestamp`, `lastActivityTimestamp`) onto each record — duplicating context that the logs pipeline already assembles for `sessionId` / `distinctId` / `url.full` / `feature_flags`.

## Changes

Builds on @hpouillot's original null-timestamp fix (#3942) and routes the session attributes through the shared logs SDK context instead of the console entrypoint:

- Move `window.id` / `sessionStartTimestamp` / `lastActivityTimestamp` into `LogSdkContext` and emit them from `buildOtlpLogRecord` (`@posthog/core`), alongside the existing auto-attributes. Same OTLP keys and stringified values as before.
- Populate them in `_getSdkContext` (browser), guarding the timestamps with `isNullish` — this is where the original null-timestamp crash fix now lives.
- Drop the hand-rolled session block from the console entrypoint; it now carries only `host`. No per-log `getLogAttributes()` closure is needed because `_getSdkContext` is already called per `captureLog`, so the attributes stay live per log line.

**Behavioural note:** these three attributes now appear on **all** log records — user-facing `posthog.log.*` as well as console logs — consistent with how `sessionId` already behaves. The wire format (attribute keys + stringified values) is unchanged.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-logs.ts
+++ b/packages/browser/src/posthog-logs.ts
@@ -355,8 +355,16 @@ export class PostHogLogs implements Extension {
         context.distinctId = this._instance.get_distinct_id()
 
         if (this._instance.sessionManager) {
-            const { sessionId } = this._instance.sessionManager.checkAndGetSessionAndWindowId(true)
+            const { sessionId, windowId, sessionStartTimestamp, lastActivityTimestamp } =
+                this._instance.sessionManager.checkAndGetSessionAndWindowId(true)
             context.sessionId = sessionId
+            context.windowId = windowId
+            if (!isNullish(sessionStartTimestamp)) {
+                context.sessionStartTimestamp = sessionStartTimestamp
+            }
+            if (!isNullish(lastActivityTimestamp)) {
+                context.lastActivityTimestamp = lastActivityTimestamp
+            }
         }
 
         if (assignableWindow?.location?.href) {
--- a/packages/core/src/logs/logs-utils.ts
+++ b/packages/core/src/logs/logs-utils.ts
@@ -124,6 +124,15 @@ export function buildOtlpLogRecord(options: CaptureLogOptions, sdkContext: LogSd
   if (sdkContext.sessionId) {
     autoAttributes.sessionId = sdkContext.sessionId
   }
+  if (sdkContext.windowId) {
+    autoAttributes['window.id'] = sdkContext.windowId
+  }
+  if (!isUndefined(sdkContext.sessionStartTimestamp)) {
+    autoAttributes.sessionStartTimestamp = String(sdkContext.sessionStartTimestamp)
+  }
+  if (!isUndefined(sdkContext.lastActivityTimestamp)) {
+    autoAttributes.lastActivityTimestamp = String(sdkContext.lastActivityTimestamp)
+  }
   if (sdkContext.currentUrl) {
     autoAttributes['url.full'] = sdkContext.currentUrl
   }
--- a/packages/core/src/logs/logs-utils.ts
+++ b/packages/core/src/logs/logs-utils.ts
@@ -10,7 +10,7 @@ import type {
   OtlpSeverityText,
 } from '@posthog/types'
 import type { LogSdkContext, ResolvedPostHogLogsConfig } from './types'
-import { isArray, isBoolean, isNull, isUndefined } from '../utils'
+import { isArray, isBoolean, isNull, isNullish, isUndefined } from '../utils'
 
 // ============================================================================
 // Severity mapping
@@ -127,10 +127,10 @@ export function buildOtlpLogRecord(options: CaptureLogOptions, sdkContext: LogSd
   if (sdkContext.windowId) {
     autoAttributes['window.id'] = sdkContext.windowId
   }
-  if (!isUndefined(sdkContext.sessionStartTimestamp)) {
+  if (!isNullish(sdkContext.sessionStartTimestamp)) {
     autoAttributes.sessionStartTimestamp = String(sdkContext.sessionStartTimestamp)
   }
-  if (!isUndefined(sdkContext.lastActivityTimestamp)) {
+  if (!isNullish(sdkContext.lastActivityTimestamp)) {
     autoAttributes.lastActivityTimestamp = String(sdkContext.lastActivityTimestamp)
   }
   if (sdkContext.currentUrl) {
--- a/.changeset/fresh-logs-timestamps.md
+++ b/.changeset/fresh-logs-timestamps.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fix browser console log capture when session activity timestamps are missing and refresh session attributes for each log.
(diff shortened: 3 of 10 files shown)


frankh (approved):

ablaszkiewicz (approved):

packages/core/src/logs/logs-utils.ts
```diff
@@ -124,6 +124,15 @@ export function buildOtlpLogRecord(options: CaptureLogOptions, sdkContext: LogSd
   if (sdkContext.sessionId) {
     autoAttributes.sessionId = sdkContext.sessionId
   }
+  if (sdkContext.windowId) {
+    autoAttributes['window.id'] = sdkContext.windowId
+  }
+  if (!isUndefined(sdkContext.sessionStartTimestamp)) {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/data-integrity + paul + xp]** 🟡 MEDIUM

The guard story for the two timestamps is told in two dialects that disagree. Here in core you use `!isUndefined(...)`, so a literal `null` would slip through to `String(null)` -> the string `"null"`. Upstream, `_getSdkContext` filters with `!isNullish(...)` (drops null *and* undefined). Today that's safe only because the browser is the sole producer and never forwards a null — but `LogSdkContext` types these as `number | undefined`, and the PR's stated purpose is to make core the crash-safe home for the guard. Switching these two checks to `!isNullish(...)` makes the record-builder defensive on its own terms and says the rule OnceAndOnlyOnce; the epoch-0 test still passes since `isNullish(0)` is false. Not blocking.

pauldambra (commented):

packages/browser/src/posthog-logs.ts
```diff
@@ -355,8 +355,16 @@ export class PostHogLogs implements Extension {
         context.distinctId = this._instance.get_distinct_id()
 
         if (this._instance.sessionManager) {
-            const { sessionId } = this._instance.sessionManager.checkAndGetSessionAndWindowId(true)
+            const { sessionId, windowId, sessionStartTimestamp, lastActivityTimestamp } =
+                this._instance.sessionManager.checkAndGetSessionAndWindowId(true)
             context.sessionId = sessionId
+            context.windowId = windowId
```
pauldambra, line 361: > [!NOTE]

`context.windowId = windowId` is assigned unconditionally while the two sibling timestamps right below get an `isNullish` guard. It works out because `buildOtlpLogRecord` does a truthy `if (sdkContext.windowId)` so a null is dropped downstream — but the three sibling fields reading differently makes a future reader pause. Consider guarding all three the same way so the intent ("only set when present") is expressed uniformly. Not blocking.

pauldambra:
> [!NOTE]
>

A clean, well-tested refactor: the three hand-rolled session attributes are removed from the console-logs entrypoint and emitted once from the shared logs SDK context, alongside `sessionId`/`distinctId`/`url.full`/`feature_flags`. The original null-timestamp crash fix is preserved (now in `_getSdkContext` via `isNullish`). No blocking issues.

### Key findings

- 🟡 **MEDIUM** — `logs-utils.ts`: guard asymmetry across the package boundary. Core uses `!isUndefined` while `_getSdkContext` uses `!isNullish`; align core on `isNullish` so the record-builder is defensive on its own terms (epoch-0 test still passes).
- 🟢 **LOW** — `posthog-logs.ts`: `context.windowId` is assigned unguarded while the sibling timestamps are guarded; smooth the asymmetry.
- ⚪ **NIT / awareness** — these three attributes now appear on **all** log records (console + user-facing `posthog.log.*`), not just console logs, and are computed per-capture rather than frozen at init. This is the more-correct behaviour, but worth a line in the PR description and a check that the logs ingestion side imposes no required-absent expectation.

### Convergence

- The **guard asymmetry** (`logs-utils.ts`) was flagged independently by qa-team, paul, and xp — highest-confidence finding.
- **`windowId` unguarded** (`posthog-logs.ts`) flagged independently by paul and xp.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Clean, well-scoped refactor; correct null handling; thorough parameterized tests. Approve. |
| 👤 paul | High trust, stamp it; two sentences wanted (guard seam + widening), neither blocks. |
| 📐 xp | Tidy consolidation, says the idea OnceAndOnlyOnce; one substantive concern — align the guard predicates. |
| 🛡 security-audit | No findings. Attributes sit in the same trust class as the already-emitted `sessionId`; same first-party endpoint. |

---

packages/core/src/logs/logs-utils.ts
pauldambra: > [!NOTE]

Fixed in `b65d672a8` — swapped `!isUndefined(...)` to `!isNullish(...)` for both timestamp guards in `packages/core/src/logs/logs-utils.ts` and added `isNullish` to the import. All 43 unit tests pass.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Intentional asymmetry — `windowId` is a `string | undefined` (never `null`) from the session manager, so the downstream truthy `if (sdkContext.windowId)` in `buildOtlpLogRecord` is sufficient. The two timestamp siblings need `isNullish` because their types permit `null` and the guard is now fixed in the MEDIUM finding above. Leaving this as-is to avoid widening the diff scope; a follow-up can unify the pattern.

packages/browser/src/posthog-logs.ts
```diff
@@ -355,8 +355,16 @@
         context.distinctId = this._instance.get_distinct_id()
 
         if (this._instance.sessionManager) {
-            const { sessionId } = this._instance.sessionManager.checkAndGetSessionAndWindowId(true)
+            const { sessionId, windowId, sessionStartTimestamp, lastActivityTimestamp } =
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Out of scope for this PR — the `Math.random()` usage is inside `checkAndGetSessionAndWindowId` for generating session and window IDs, which is a pre-existing pattern not introduced by these changes. Session IDs in analytics SDKs do not require cryptographic randomness. This PR only adds null-safety guards around the values returned from that call.

turnipdabeets (approved):
LGTM, but I think you'll need a changeset for it to be released.

PR #3945, opened 2026-06-24
fix(surveys): guard handlePageUnload against version-skewed surveys instance

### LINKED ISSUE #3251: Error "TypeError: t2.handlePageUnload is not a function" after page refresh
## Bug description

After refreshing the page, the error "TypeError: t2.handlePageUnload is not a function" appears in the console

## How to reproduce

1. Open the browser
2. Enable the "preserve log" function
3. Refresh the page
4. The console error is "TypeError: t2.handlePageUnload is not a function"

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

After updating the "posthog-js" version, the error "TypeError e.handlePageUnload is not a function" started falling in the Sentry of the work project. It occurs when the user refreshes the page, so it can only be noticed in the console when the "preserve log" mode is enabled 

The path to the error is "node_modules/posthog-js/dist/module.js"

The problem appeared in version `1.310.0`

Presumably, it could appear because of [this change](https://github.com/PostHog/posthog-js/pull/2786).

My config looks like this (the "api_host" and "defaults" fields are hidden) :

```
export const postHogConfig: Partial<PostHogConfig> = {
    api_host: '',
    defaults: '',
    session_recording: {
        recordBody: true,
        recordCrossOriginIframes: true,
        recordHeaders: true,
    },
};
```

ioannisj:
## Problem

Closes #3251

On page unload PostHog calls `this.surveys.handlePageUnload()`. After a deploy, a browser can still be holding a **cached, older surveys extensions bundle** from the CDN whose `PostHogSurveys` instance predates `handlePageUnload`. The method is absent on that stale instance, so the unload handler throws:

```
TypeError: t2.handlePageUnload is not a function
```

A previous patch (#3165, v1.358.0) added optional chaining — `this.surveys?.handlePageUnload()` — and the issue was closed, but it was reopened because the crash still reproduces. Optional chaining on the *receiver* only guards against `surveys` being `undefined`/`null`; when `surveys` is a real object whose `handlePageUnload` is `undefined`, calling it still throws.

## Changes

Optional-call the method, not just the receiver:

```ts
// before
this.surveys?.handlePageUnload()

// after
this.surveys?.handlePageUnload?.()
```

- `?.()` short-circuits when `handlePageUnload` is absent (reads as `undefined`), which is exactly the version-skew failure mode. Matches the existing `?.()` idiom in this file (e.g. `extension.initialize?.()`).
- Scope is deliberately narrow: only `handlePageUnload` is guarded, since it's the only auto-invoked surveys method new enough to be missing from cached bundles. The sibling auto-calls (`reset()`, `loadIfEnabled()`) are older methods present in any cached bundle, so they don't reproduce this crash.
- Added unit tests in `posthog-core-also.test.ts`: method-present is still invoked once; a stale `surveys: {}` stub missing the method no longer throws and the rest of the unload path (`$pageleave` capture) still runs.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code, directed by @ioannisj.

### DIFF
--- a/.changeset/proud-ferns-wander.md
+++ b/.changeset/proud-ferns-wander.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(surveys): guard handlePageUnload against version-skewed surveys instance missing the method
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1051,7 +1051,12 @@ export class PostHog implements PostHogInterface {
     }
 
     _handle_unload(): void {
-        this.surveys?.handlePageUnload()
+        // Guard on the method being callable, not just present: after a deploy a cached older
+        // lazy-loaded surveys chunk can yield an instance whose prototype lacks handlePageUnload,
+        // and optional chaining alone would still throw "handlePageUnload is not a function".
+        if (isFunction(this.surveys?.handlePageUnload)) {
+            this.surveys.handlePageUnload()
+        }
 
         if (!this.config.request_batching) {
             if (this._shouldCapturePageleave()) {
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1051,12 +1051,10 @@ export class PostHog implements PostHogInterface {
     }
 
     _handle_unload(): void {
-        // Guard on the method being callable, not just present: after a deploy a cached older
+        // Optional-call the method, not just the receiver: after a deploy a cached older
         // lazy-loaded surveys chunk can yield an instance whose prototype lacks handlePageUnload,
-        // and optional chaining alone would still throw "handlePageUnload is not a function".
-        if (isFunction(this.surveys?.handlePageUnload)) {
-            this.surveys.handlePageUnload()
-        }
+        // and `this.surveys?.handlePageUnload()` would still throw "handlePageUnload is not a function".
+        this.surveys?.handlePageUnload?.()
 
         if (!this.config.request_batching) {
             if (this._shouldCapturePageleave()) {
--- a/packages/browser/src/__tests__/posthog-core-also.test.ts
+++ b/packages/browser/src/__tests__/posthog-core-also.test.ts
@@ -805,6 +805,40 @@ describe('posthog core', () => {
             expect(flushLogs).toHaveBeenCalledWith('sendBeacon')
         })
 
+        it('calls surveys.handlePageUnload when present', () => {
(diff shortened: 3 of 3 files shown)


.changeset/proud-ferns-wander.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(surveys): guard handlePageUnload against version-skewed surveys instance missing the method
```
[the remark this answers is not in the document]
ioannisj: Good catch — this was a stale PR description from a copy/paste error (an unrelated feature-flags body). The description has been corrected to describe the `surveys.handlePageUnload` version-skew guard, which matches the diff, title, and changeset. No code change needed.

packages/browser/src/posthog-core.ts
```diff
@@ -1068,7 +1051,12 @@ export class PostHog implements PostHogInterface {
     }
 
     _handle_unload(): void {
-        this.surveys?.handlePageUnload()
+        // Guard on the method being callable, not just present: after a deploy a cached older
+        // lazy-loaded surveys chunk can yield an instance whose prototype lacks handlePageUnload,
+        // and optional chaining alone would still throw "handlePageUnload is not a function".
+        if (isFunction(this.surveys?.handlePageUnload)) {
+            this.surveys.handlePageUnload()
+        }
```
marandaneto: i think `this.surveys?.handlePageUnload?.()` would do

marandaneto (approved):

packages/browser/src/posthog-core.ts
ioannisj: Ah, so optional chain on the function as well. Let me test cause it's much cleaner

PR #3955, opened 2026-06-24
fix: Dedupe feature flag called events by response

marandaneto:
## Problem

Feature flag called events could be emitted again for the same flag response after flags were reloaded or updated. This adds noise to `$feature_flag_called` analytics while still needing to report when a flag's returned value actually changes.

## Changes

- Track feature flag called dedupe state by flag key and returned response in `@posthog/core`.
- Stop clearing core feature flag called dedupe state on flag reload/update; clear it on shutdown instead.
- Clear Node.js feature flag called dedupe state on shutdown.
- Update core v1/v4 feature flag tests and add a Node.js shutdown regression test.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A Pi implementation agent prepared the PR from the existing dedicated worktree at the maintainer's request. The change keeps `$feature_flag_called` deduplication tied to each returned flag value and includes targeted regression tests for core and Node.js behavior.

### DIFF
--- a/.changeset/silver-pandas-shake.md
+++ b/.changeset/silver-pandas-shake.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-node': patch
+---
+
+Dedupe feature flag called events per returned flag value.
--- a/packages/core/src/__tests__/posthog.featureflags.spec.ts
+++ b/packages/core/src/__tests__/posthog.featureflags.spec.ts
@@ -807,7 +807,7 @@ describe('PostHog Feature Flags v4', () => {
         }
       )
 
-      it('should capture $feature_flag_called again if new flags', async () => {
+      it('should not capture $feature_flag_called again if reloaded flags keep the same value', async () => {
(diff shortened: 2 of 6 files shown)


marandaneto:
Addressed Greptile feedback in 55761ecd9: split the feature flag called dedupe coverage into two clearly named tests—one for firing when a new value is seen and one for not firing when cycling back to a previously seen value.

dustinbyrne (approved):

PR #3959, opened 2026-06-25
fix(core): isFeatureEnabled returns false instead of undefined

### LINKED ISSUE #3930: React Native: align isFeatureEnabled missing-flag fallback with sdk-specs
## Summary
Align React Native `isFeatureEnabled()` missing-flag behavior with the canonical SDK spec default/fallback behavior.

## Spec reference
- `PostHog/sdk-specs`: `openspec/specs/is-feature-enabled/spec.md`

## Current React Native behavior
RN inherits core behavior from `packages/core/src/posthog-core.ts`:

```ts
isFeatureEnabled(key: string): boolean | undefined {
  const response = this.getFeatureFlag(key)
  if (response === undefined) {
    return undefined
  }
  return !!response
}
```

This means `isFeatureEnabled()` can return `undefined` when flags are empty/not loaded or when `getFeatureFlag()` returns `undefined`.

## Expected behavior
The canonical `is-feature-enabled` spec expects missing cached flags to resolve to `false` for the boolean API.

For RN SDK-spec parity, `isFeatureEnabled('missing')` should return `false` once the SDK has no value for that flag, or otherwise expose a default-value/fallback mechanism consistent with the spec and other SDKs.

## Notes
This affects the boolean convenience API only. `getFeatureFlag()` / `getFeatureFlagResult()` can continue representing unknown/unavailable flags distinctly where appropriate.

tsushanth:
## Problem

`isFeatureEnabled` in `@posthog/core` declares a return type of `boolean | undefined` and returns `undefined` when the underlying `getFeatureFlag` call finds no cached flag value. This third state is unnecessary: the function's semantics are \"is this flag enabled?\", and the answer to that question when a flag is absent or not yet loaded is `false`, not an unknown.

Closes #3930

## Changes

`isFeatureEnabled` in `packages/core/src/posthog-core.ts` now returns `false` instead of `undefined` when `getFeatureFlag` returns `undefined`. The return type is narrowed from `boolean | undefined` to `boolean`. The interface in `packages/types/src/posthog.ts` is updated to match.

Tests in `posthog.featureflags.spec.ts` and `posthog.featureflags.v1.spec.ts` that previously asserted `undefined` for the not-loaded and error-state cases are updated to assert `false`.

Note: `posthog-featureflags.ts` in the browser package has its own `isFeatureEnabled` that retains `undefined` for the \"flags still loading\" state — that is an intentional, separate concern and is not changed here.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Took care not to unnecessarily increase the bundle size## 🤖 Agent context

Used Claude Code to help identify the issue, draft the fix, and update the affected test files. Reviewed the changes and verified the browser package's `isFeatureEnabled` was intentionally left unchanged (it has a distinct "not-yet-loaded" semantic that warrants the `undefined` return). The rebase and changeset were also done with Claude's assistance.

### DIFF
--- a/.changeset/fix-is-feature-enabled-return-false.md
+++ b/.changeset/fix-is-feature-enabled-return-false.md
@@ -1,0 +1,6 @@
+---
+"@posthog/core": patch
+"@posthog/types": patch
+---
+
+fix(core): isFeatureEnabled returns false instead of undefined when flag is not found
--- a/packages/react-native/src/posthog-rn.ts
+++ b/packages/react-native/src/posthog-rn.ts
@@ -959,9 +959,9 @@ export class PostHog extends PostHogCore {
    *
    * @param key The feature flag key
    * @param options Optional per-call settings
-   * @returns True if enabled, false if disabled, undefined if not loaded
+   * @returns True if enabled, false if disabled
    */
-  isFeatureEnabled(key: string, options?: FeatureFlagResultOptions): boolean | undefined {
+  isFeatureEnabled(key: string, options?: FeatureFlagResultOptions): boolean {
     return super.isFeatureEnabled(key, options)
   }
 
--- a/packages/types/src/posthog.ts
+++ b/packages/types/src/posthog.ts
@@ -333,7 +333,7 @@ export interface PostHog {
      * @param options.fresh - If true, only return values loaded from the server, not cached localStorage values (default: false)
      * @returns Whether the feature flag is enabled
      */
-    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean | undefined
+    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean
 
     /**
      * Reload feature flags from the server.
--- a/packages/core/src/__tests__/posthog.featureflags.spec.ts
+++ b/packages/core/src/__tests__/posthog.featureflags.spec.ts
@@ -144,9 +144,9 @@ describe('PostHog Feature Flags v4', () => {
       expect(posthog.getFeatureFlagPayload('my-flag')).toEqual(undefined)
     })
 
-    it('isFeatureEnabled should return undefined if not loaded', () => {
-      expect(posthog.isFeatureEnabled('my-flag')).toEqual(undefined)
-      expect(posthog.isFeatureEnabled('feature-1')).toEqual(undefined)
+    it('isFeatureEnabled should return false if not loaded', () => {
+      expect(posthog.isFeatureEnabled('my-flag')).toEqual(false)
+      expect(posthog.isFeatureEnabled('feature-1')).toEqual(false)
     })
 
     it('should load persisted feature flags', () => {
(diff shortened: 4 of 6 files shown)


turnipdabeets:
@tsushanth let's get CI working and the project building first. Please follow our [AI_POLICY](https://github.com/PostHog/posthog-js/blob/main/AI_POLICY.md)

tsushanth:
Rebased onto current main and added a changeset file to fix the hygiene check. The merge conflict in `posthog-core.ts` is resolved — kept the `options?: FeatureFlagResultOptions` parameter that was added upstream while still narrowing the return type to `boolean`. All four modified files (`posthog-core.ts`, `posthog.ts`, and both spec files) are clean on the new base.

tsushanth:
Fixed the TS2416 error: updated `isFeatureEnabled` override in `posthog-rn.ts` to return `boolean` (matching the base class) instead of `boolean | undefined`.

packages/types/src/posthog.ts
```diff
@@ -333,7 +333,7 @@ export interface PostHog {
      * @param options.fresh - If true, only return values loaded from the server, not cached localStorage values (default: false)
      * @returns Whether the feature flag is enabled
      */
-    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean | undefined
+    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean
```
turnipdabeets, line 336: This interface is implemented by the browser `PostHog` class, which intentionally keeps `boolean | undefined` (#2276 made missing/disabled flags return `undefined` on web), so narrowing it here breaks the browser typecheck (TS2416 in `packages/browser/src/posthog-core.ts`). Core and RN can return the narrower `boolean` without changing the shared interface — could we revert this line?

```suggestion
    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean | undefined
```

turnipdabeets, line 336: [question for @dustinbyrne] curious for your take here since #2276 deliberately gave web the tri-state (`undefined` = missing/disabled, `false` = not rolled out). The sdk-spec says missing → `false`, so web and the spec currently disagree — is js v2 the right place to decide whether web goes two-state, so we don't pre-empt that decision in this PR? Or do you think 3 states is actually the right approach and this issue is wrong?

.changeset/fix-is-feature-enabled-return-false.md
```diff
@@ -0,0 +1,6 @@
+---
+"@posthog/core": patch
+"@posthog/types": patch
```
turnipdabeets, line 3: The diff also touches `packages/react-native/src/posthog-rn.ts`, so the changeset needs a `posthog-react-native` entry. If `@posthog/types` change is reverted, that entry should come out. Should be based on Dustin’s response above. 

```suggestion
"@posthog/core": patch
"posthog-react-native": patch
```

packages/react-native/src/posthog-rn.ts
```diff
@@ -959,9 +959,9 @@ export class PostHog extends PostHogCore {
    *
    * @param key The feature flag key
    * @param options Optional per-call settings
-   * @returns True if enabled, false if disabled, undefined if not loaded
+   * @returns True if enabled, false if disabled
```
turnipdabeets, line 962: `false` is what callers get before flags have loaded — worth saying so here, and that `getFeatureFlag` still returns `undefined` for anyone who needs to tell the states apart.

```suggestion
   * @returns True if enabled, false if disabled or not yet loaded
```

turnipdabeets (commented):
Thanks for the updates @tsushanth

turnipdabeets:
@tsushanth thanks for your contribution, I spoke with team and we're going with a slightly different approach. The issue it self was a bit unclear. Closing for https://github.com/PostHog/posthog-js/pull/4222

PR #3960, opened 2026-06-25
perf(browser): speed up truncated log capture

marandaneto:
## Problem

Console log capture performance tests can be flaky on CI for large objects with many keys. The motivating CI failure was [the logs performance test exceeding the 100ms threshold](https://github.com/PostHog/posthog-js/actions/runs/28094176074/job/83179337718?pr=3939). Profiling showed the hot path walked large objects twice: once to build the bounded log body and again to flatten object attributes.

## Changes

- Collect flattened attributes during the bounded body serialization pass, avoiding a second walk over the same large object.
- Preserve bounded structured attributes even when the serialized body is truncated.
- Added regression coverage for preserving useful attributes on truncated large logs.
- Added a patch changeset for `posthog-js`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

An agent profiled the flaky logs performance path and found time split between bounded body serialization and flattening the same large object into attributes. After review feedback, the fix now keeps structured attributes for truncated logs while avoiding the duplicate root-object walk by collecting attributes during serialization. Verification included the targeted logs entrypoint Jest suite, ESLint for the changed files, browser typecheck, and `git diff --check`.

### DIFF
--- a/packages/browser/src/__tests__/entrypoints/logs.test.ts
+++ b/packages/browser/src/__tests__/entrypoints/logs.test.ts
@@ -152,6 +152,28 @@ describe('logs entrypoint', () => {
             )
         })
 
+        it('should preserve bounded attributes when the log body is truncated', () => {
+            const initializeLogs = assignableWindow.__PosthogExtensions__.logs.initializeLogs
+            initializeLogs(mockPostHog)
+
+            assignableWindow.console.log({
+                code: 'E_TOO_LARGE',
+                userId: 'user-123',
+                payload: 'x'.repeat(10001),
+            })
+
+            expect(mockEmit).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    body: expect.stringContaining('...'),
+                    attributes: expect.objectContaining({
+                        body_truncated: 'true',
+                        code: 'E_TOO_LARGE',
+                        userId: 'user-123',
+                    }),
+                })
+            )
+        })
+
         it('should not read object properties after the body size limit is reached', () => {
             const initializeLogs = assignableWindow.__PosthogExtensions__.logs.initializeLogs
             initializeLogs(mockPostHog)
--- a/packages/browser/src/__tests__/entrypoints/logs.test.ts
+++ b/packages/browser/src/__tests__/entrypoints/logs.test.ts
@@ -451,6 +451,10 @@ describe('logs entrypoint', () => {
             expect(mockEmit.mock.calls[0][0].attributes).toEqual(
                 expect.objectContaining({
                     'log.source': 'console.error',
+                    code: 'E_BOOM',
+                    name: 'CustomError',
+                    message: 'boom',
+                    stack: 'CustomError: boom\n    at test',
                 })
             )
         })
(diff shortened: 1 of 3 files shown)


packages/browser/src/entrypoints/logs.ts
```diff
@@ -325,7 +322,9 @@ const initializeLogs = (posthog: PostHog) => {
                             attributes: {
                                 'log.source': `console.${level}`,
                                 ...logAttributes,
-                                ...(isObject(args[0]) ? flattenObject(args[0]) : {}),
+                                // If the body already hit the size limit, avoid walking the same large object
+                                // again just to collect duplicate partial attributes.
+                                ...(!truncated && isObject(args[0]) ? flattenObject(args[0]) : {}),
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in ac5af9fa9 — truncated logs now preserve bounded structured attributes while collecting them during the body serialization pass, so we avoid the duplicate root-object walk without dropping useful fields.

packages/browser/src/entrypoints/logs.ts
```diff
@@ -68,12 +76,65 @@ const isNumberOrBoolean = (value: any): boolean => {
...
+
+    collector.result[key] = value
+}
+
+const collectFlattenedAttributes = (value: any, key: string, collector: AttributeCollector): void => {
+    if (collector.truncated) {
+        return
+    }
+
+    if (typeof value === 'object' && !isNull(value) && !isArray(value)) {
```
marandaneto: should we use `isObject` here

marandaneto: Yep, agreed. Updated in 7c857876f to use the existing isObject helper here.

packages/browser/src/__tests__/entrypoints/logs.test.ts
ioannisj, line 459: Agent flagged the possibility that we are dropping attributes for Error objects (do we want to?) I think this should also check for attributes as well cause if we are unintentionally dropping attributes, the test will silently pass

marandaneto: Good catch — this was a regression in the PR. Fixed in d440d8ba6 by passing the attribute collector through Error serialization and extending the test to assert Error attributes.

marandaneto: yeah good one

ioannisj (approved):
LG but some checks are red?

marandaneto:
> LG but some checks are red?

question or affirmation? :)
just network instability on the github side

PR #3974, opened 2026-06-25
fix(node): apply in/not_in operator on flag-level cohort conditions in local evaluation

### LINKED ISSUE #3972: Local evaluation ignores the in/not_in operator on a flag-level cohort condition (not in cohort evaluated as in cohort)
## Bug description

When a feature flag's release condition is **"Cohort X — `not in`"** and the SDK evaluates it **locally** (`personalApiKey` set), `posthog-node` ignores the cohort property's membership operator. It computes only *"is this person in the cohort?"* and uses that boolean directly as the match result. So `not in cohort` produces the **same** result as `in cohort` — the negation is silently dropped.

Consequence: a flag intended to *exclude* a cohort instead *includes* it. An "exempt our team / internal users" flag enforces on exactly the people it was meant to exempt, while `$feature_flag_called` reports `locally_evaluated: true` and no error is raised.

Confirmed on `posthog-node@5.21.2` and still present on `main` (`5.38.5`, code unchanged). `operator: "in"` works correctly; only `not_in` is inverted.

Root cause (links pinned to `d6fc0a50fa2c2650869e3444a030533753ff0b4f`):

**`isConditionMatch`** — for a `cohort` property it calls `matchCohort` then `if (!matches) return 'no_match'`. The condition property's `operator` (`in`/`not_in`) is never read:
https://github.com/PostHog/posthog-js/blob/d6fc0a50fa2c2650869e3444a030533753ff0b4f/packages/node/src/extensions/feature-flags/feature-flags.ts#L606-L617

```ts
if (propertyType === 'cohort') {
  matches = await matchCohort(prop, properties, this.cohorts, this.debugMode, (depProp) =>
    this.evaluateFlagDependency(depProp, properties, evaluationContext)
  )
} else if (propertyType === 'flag') {
  ...
} else {
  matches = matchProperty(prop, properties, warnFunction)
}

if (!matches) {
  return 'no_match'
}
```

**`matchCohort`** — uses only `property.value` (the cohort id) and returns raw membership; never inspects `property.operator` / `property.negation`:
https://github.com/PostHog/posthog-js/blob/d6fc0a50fa2c2650869e3444a030533753ff0b4f/packages/node/src/extensions/feature-flags/feature-flags.ts#L1206-L1217

```ts
async function matchCohort(property, propertyValues, cohortProperties, debugMode = false, flagDependencyEvaluator?) {
  const cohortId = String(property.value)
  checkCohortExists(cohortId, cohortProperties)
  const propertyGroup = cohortProperties[cohortId]
  return matchPropertyGroup(propertyGroup, propertyValues, cohortProperties, debugMode, flagDependencyEvaluator)
}
```

The negation machinery already exists for properties *inside* a cohort group (`matchPropertyGroup`), it's just never applied to the flag-condition-level cohort operator:
https://github.com/PostHog/posthog-js/blob/d6fc0a50fa2c2650869e3444a030533753ff0b4f/packages/node/src/extensions/feature-flags/feature-flags.ts#L1299-L1314

The flag-condition property from the `/local_evaluation` payload carries the operator, but it's not honored:

```json
{ "key": "id", "type": "cohort", "operator": "not_in", "value": 12345 }
```

## How to reproduce

1. Create a person-property cohort, e.g. **Team** = `email` matches regex `.*@example\.com$`.
2. Create a boolean flag with one release condition: **Cohort `Team` — `not in`**, rolled out to **100%** (intent: ON for everyone *except* the Team).
3. Evaluate locally for a user who **is** in the cohort and observe `true` instead of `false`:

```ts

const ph = new PostHog(PROJECT_API_KEY, {
  personalApiKey: PERSONAL_API_KEY, // enables local evaluation
})
await ph.reloadFeatureFlags()

const enabled = await ph.isFeatureEnabled('my_flag', 'user-123', {
  personProperties: { email: 'someone@example.com' }, // matches the cohort
  onlyEvaluateLocally: true, // prove it's the local path, not /decide
})

console.log(enabled) // got: true   expected: false
```

| | Expected | Actual |
|---|---|---|
| User **in** "not in" cohort | `false` (excluded → no match) | **`true`** (treated as match) |
| User **not in** "not in" cohort | `true` | `true` |

Evaluating the same flag/person via `/decide` (server) returns the correct `false`. Because a single person-property cohort *is* locally evaluable, the SDK never falls back to `/decide`, so callers get the inverted local result.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

**Impact:** any server-side flag that uses a cohort to *exclude* a group is inverted under local evaluation (e.g. "rate-limit/paywall everyone except internal users", "roll out to everyone except cohort Y"). It fails silently — `locally_evaluated: true`, no error.

**Suggested fix:** in `isConditionMatch`, apply the cohort property's membership operator to the `matchCohort` result before the `!matches` check (mirror the `negation` handling already in `matchPropertyGroup`) — invert when `prop.operator === 'not_in'` (or `prop.negation === true`).

**Workaround:** replace the cohort condition with a direct person-property condition on the flag itself (e.g. `email` `is_not` `[...]`, or `email` `regex ...`); `matchProperty` honors these operators correctly under local evaluation.

mjfaga:
Fixes https://github.com/PostHog/posthog-js/issues/3972

## Problem

Anyone using `posthog-node` server-side **local evaluation** for a feature flag whose release condition *excludes* a cohort (`Cohort X not_in`) gets the wrong result. Local evaluation ignored the cohort membership operator, so `not in cohort` was evaluated identically to `in cohort` and the negation was silently dropped.

The effect is the opposite of what the flag author intended: a flag meant to be **off** for a cohort (e.g. "enabled for everyone *except* internal/team users") is instead **on** for exactly that cohort. It fails silently and the `$feature_flag_called` event reports `locally_evaluated: true` so it's easy to ship and hard to notice. I hit this with a real "exclude this cohort" flag that was enforcing on the very users it was supposed to exempt.

`operator: "in"` is unaffected (membership → match is already correct); only `not_in` is inverted.

## Changes

In `FeatureFlagsPoller.isConditionMatch`, the `cohort` branch called `matchCohort` (which returns raw membership) and used that boolean directly as the condition match. It  previously never read the property's `in`/`not_in` operator. The fix applies the operator to the membership result:

```ts
const inCohort = await matchCohort(prop, properties, this.cohorts, this.debugMode, ...)
matches = prop.operator === 'not_in' ? !inCohort : inCohort
```

Applied at the condition level (not inside `matchCohort`) on purpose: `matchPropertyGroup` already handles `negation` for cohorts referenced *inside* another cohort's property group, so inverting inside `matchCohort` would double-negate nested cohorts. This change only affects the flag-condition-level cohort operator, which previously had no operator handling at all.

Adds a regression test, `computes 'not in' cohort conditions locally`, that:
- **fails on the previous code** (an in-cohort user gets `true` instead of `false`),
- passes with the fix,
- covers both `not_in` and `in` to confirm `in` is unchanged, and
- asserts the `/decide` (flags) API is never consulted (i.e. it's the local path).

**Behavioral note for reviewers:** flags using a `not_in` cohort condition under local evaluation will now return the *corrected* (opposite) value. That's the intended fix, but it does change evaluated output for those flags — worth calling out for anyone who may have unknowingly come to depend on the inverted behavior. Confirmed against current `main` (`posthog-node@5.38.5`); full `feature-flags.spec.ts` suite passes (227/227).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

## If releasing new changes
- [x] Ran pnpm changeset to generate a changeset file

### DIFF
--- a/.changeset/warm-windows-doubt.md
+++ b/.changeset/warm-windows-doubt.md
@@ -1,0 +1,7 @@
+---
+'posthog-node': patch
+---
+
+Fix local evaluation ignoring the `in`/`not_in` operator on cohort-based flag conditions. "Not in
+cohort" was evaluated as "in cohort", inverting cohort-exclusion flags. Now applies the operator to
+the cohort membership result.
--- a/packages/node/src/__tests__/feature-flags.spec.ts
+++ b/packages/node/src/__tests__/feature-flags.spec.ts
@@ -1751,6 +1751,89 @@ describe('local evaluation', () => {
     expect(mockedFetch).not.toHaveBeenCalledWith(...anyFlagsCall)
   })
 
+  it("computes 'not in' cohort conditions locally (negated cohort membership)", async () => {
+    const flags = {
+      flags: [
+        {
+          id: 1,
+          name: 'Exclude Team',
+          key: 'exclude-team',
+          active: true,
+          rollout_percentage: 100,
+          filters: {
+            groups: [
+              {
+                properties: [{ key: 'id', value: 98, type: 'cohort', operator: 'not_in' }],
+                rollout_percentage: 100,
+              },
+            ],
+          },
+        },
+        {
+          id: 2,
+          name: 'Include Team',
+          key: 'include-team',
+          active: true,
+          rollout_percentage: 100,
+          filters: {
+            groups: [
+              {
+                properties: [{ key: 'id', value: 98, type: 'cohort', operator: 'in' }],
+                rollout_percentage: 100,
+              },
+            ],
+          },
+        },
+      ],
+      cohorts: {
+        '98': {
+          type: 'OR',
+          values: [{ key: 'email', operator: 'regex', value: '.*@example\\.com$', type: 'person' }],
+        },
+      },
+    }
+    mockedFetch.mockImplementation(
+      apiImplementation({
+        localFlags: flags,
+        decideFlags: {},
+      })
+    )
+
+    posthog = new PostHog('TEST_API_KEY', {
+      host: 'http://example.com',
+      personalApiKey: 'TEST_PERSONAL_API_KEY',
+      ...posthogImmediateResolveOptions,
+    })
+
+    // In the cohort + `not_in` => excluded => OFF. (Pre-fix this incorrectly returned `true`.)
+    expect(
+      await posthog.getFeatureFlag('exclude-team', 'some-distinct-id', {
+        personProperties: { email: 'team@example.com' },
+      })
+    ).toEqual(false)
+    // Not in the cohort + `not_in` => ON.
+    expect(
+      await posthog.getFeatureFlag('exclude-team', 'some-distinct-id', {
+        personProperties: { email: 'outsider@example.org' },
+      })
+    ).toEqual(true)
+
+    // `in` keeps working: in the cohort => ON, not in the cohort => OFF.
+    expect(
+      await posthog.getFeatureFlag('include-team', 'some-distinct-id', {
+        personProperties: { email: 'team@example.com' },
+      })
+    ).toEqual(true)
+    expect(
+      await posthog.getFeatureFlag('include-team', 'some-distinct-id', {
+        personProperties: { email: 'outsider@example.org' },
+      })
+    ).toEqual(false)
+
+    // All evaluated locally; the /flags (decide) API must not be consulted.
+    expect(mockedFetch).not.toHaveBeenCalledWith(...anyFlagsCall)
+  })
+
   it('gets feature flag with variant overrides', async () => {
     const flags = {
       flags: [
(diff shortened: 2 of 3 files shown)


ioannisj (approved):
LG, just the nit about parameterized tests

PR #3976, opened 2026-06-25
fix(mcp): capture tools/list on servers that register handlers after instrument()

gesh:
## Why

The injected `context` parameter (and the `$mcp_tools_list` event) was missing when instrumenting an [MCP-Nest](https://github.com/rekog-labs/MCP-Nest) server.

## Why it misses
Our SDK's `instrument()` wraps the `tools/list` handler that exists at call time. MCP-Nest builds a bare server, hands it to `instrument()` via `serverMutator`, and only *then* registers `tools/list` directly at the low level — so there's no handler to wrap yet, and we bail. Unlike `tools/call`, `tools/list` never watched for a handler registered *after* `instrument()`.

## Fix
Monkeypatch `setRequestHandler` so a `tools/list` handler registered after `instrument()` gets wrapped as it lands. Plus a regression test reproducing the MCP-Nest ordering.

### DIFF
--- a/.changeset/mcp-capture-late-tools-list.md
+++ b/.changeset/mcp-capture-late-tools-list.md
@@ -1,0 +1,5 @@
+---
+'@posthog/mcp': patch
+---
+
+Capture tool listings (and the injected `context` parameter) on MCP servers that register their `tools/list` handler after `instrument()` runs — e.g. `@rekog/mcp-nest`, which hands a bare server to `instrument()` and only then registers its handlers.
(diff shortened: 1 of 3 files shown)


pauldambra (approved):
one question about whether the surface we're touching could be simpler but :ship:

gesh:
> one question about whether the surface we're touching could be simpler but 🚢

I'm going to refactor how we monkeypatch `setRequestHandler` so we can have pretty common interface like:
```
instrument(server, {
     'initialize': initializeWrapperFunction
     'tools/call': toolsCallWrapperFunction
     ...
}
```

And then we just plug into these request handlers, both in place and if they are added dynamically.

PR #3983, opened 2026-06-25
fix(next): preserve Pages Router request context

dustinbyrne:
## Problem

`getServerSidePostHog(ctx)` awaited the shared node-client cache and then called `enterContext()` before returning the client. Because `enterContext()` uses `AsyncLocalStorage.enterWith()`, setting context after an `await` inside the helper does not propagate back to the caller's `getServerSideProps` continuation.

That means Pages Router server-side flag evaluation/capture could miss the request's cookie/tracing identity even though the helper looked scoped.

## Changes

- Mirror the App Router `getPostHog()` request-scoping approach for `getServerSidePostHog()`.
- Return a proxy that wraps method calls in `client.withContext(contextData, ...)` instead of calling `client.enterContext()` after async initialization.
- Keep disabled/no-key and opted-out paths returning the raw client without context wrapping.
- Update tests to assert method calls after `await getServerSidePostHog(ctx)` are wrapped with the expected request context.
- Add a patch changeset for `@posthog/next`.

Validation:

```bash
pnpm --filter @posthog/next test:unit -- getPostHog.test.ts getServerSidePostHog.test.ts
pnpm --filter @posthog/next lint
```

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

This was investigated with Pi and an oracle subagent. The subagent confirmed the async-boundary issue with `AsyncLocalStorage.enterWith()`, and this PR chooses the same proxy + `withContext()` pattern already used by App Router `getPostHog()`.

### DIFF
--- a/.changeset/next-pages-context-boundary.md
+++ b/.changeset/next-pages-context-boundary.md
@@ -1,0 +1,5 @@
+---
+'@posthog/next': patch
+---
+
+Fix Pages Router server clients to apply request context after async initialization.
--- a/packages/next/src/pages/getServerSidePostHog.ts
+++ b/packages/next/src/pages/getServerSidePostHog.ts
@@ -1,3 +1,4 @@
+import { isFunction } from '@posthog/core'
 import type { GetServerSidePropsContext } from 'next'
 import type { PostHogOptions, IPostHog } from 'posthog-node'
 import { getOrCreateNodeClient } from '../server/clientCache.node.js'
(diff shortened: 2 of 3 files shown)


dustinbyrne:
Folded into #4091 (commit `1584e5afa`): `getServerSidePostHog` now applies context via a shared `withRequestContext` proxy helper (same mechanism as App Router) instead of `enterContext()`, with the tests from this PR carried over. #4091 rewrites this file anyway for the `createPostHog` factory, so landing the fix there avoids a guaranteed conflict. Closing as superseded.

PR #3985, opened 2026-06-25
ci: validate Next example packaging

dustinbyrne:
## Problem

We want to stop relying on Vercel preview deployments as CI coverage for the `@posthog/next` demo app while preserving the useful packaging/build validation they provided.

## Changes

- Extends the existing `Package tarballs` workflow job to install and build `examples/example-next-app-router` from the freshly generated local tarballs.
- Adds the same install/build validation for `examples/example-nextjs` so the `@posthog/nextjs-config` demo app path is represented in GitHub Actions as well.
- Disables automatic Vercel git deployments for `examples/example-next-app-router` via `vercel.json`.
- `posthog-js` and `posthog-nextjs-config` Vercel deploys disconnected (these two were managed in Vercel)

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi was used to inspect the existing Vercel-backed example projects and wire equivalent tarball install/build validation into GitHub Actions. The PR intentionally keeps the deploy-removal behavior limited to `example-next-app-router`; the newly added `example-nextjs` build may surface existing packaging issues and is included to make that validation explicit in CI.

### DIFF
--- a/.github/workflows/library-ci.yml
+++ b/.github/workflows/library-ci.yml
@@ -319,7 +319,7 @@ jobs:
 
       # Replaces Vercel preview deploy validation for @posthog/next.
       - name: Install Next.js App Router example dependencies from local tarballs
-        run: pnpm install
+        run: pnpm install --lockfile=false
         working-directory: examples/example-next-app-router
 
       - name: Build Next.js App Router example
@@ -331,7 +331,7 @@ jobs:
           NEXT_PUBLIC_POSTHOG_HOST: https://us.i.posthog.com
 
       - name: Install Next.js config example dependencies from local tarballs
-        run: pnpm install --no-frozen-lockfile
+        run: pnpm install --lockfile=false
         working-directory: examples/example-nextjs
 
       - name: Build Next.js config example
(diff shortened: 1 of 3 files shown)


.github/workflows/library-ci.yml
```diff
@@ -317,6 +317,32 @@ jobs:
       - name: Build package tarballs
         run: pnpm package
 
+      # Replaces Vercel preview deploy validation for @posthog/next.
```
marandaneto, line 320: Can this be its own job so runs in parallel?

marandaneto (approved):

PR #3986, opened 2026-06-26
feat(browser): capture $device_model on Android Chromium

### LINKED ISSUE #21955: Add Device Make/Model for insights/web analytics

We are trying to build a more targeted test routine based on the mobile devices visiting our site. With a wide variety of makes and models out there now, a simple iOS/Android dichotomy or mobile/tablet/desktop breakdown is limited in its usefulness, especially with performance and screen resolutions to take into account.

I'd like the ability to see the make and model tracked as an attribute to an event, or as a general report within web analytics. Ideally it would be good to determine that 27% of my users are using iPhone 13 devices, 10% are using iPhone 12 Mini devices, 13% are using Google Pixel 6 devices, 11% are using Samsung Galaxy S21 devices, and so on.

GA4 appears to have these metrics available, though some of it seems limited at the moment or we have instrumented incorrectly. I've heard that Flurry may also have this information, but their site seems iffy.

## Additional context

GA4 lists their user dimensions available out of box here (they replace "make" with "brand": https://support.google.com/analytics/answer/9268042?hl=en&ref_topic=11151952&sjid=18183633266358039648-NC 

### Debug info

```shell
- [ ] PostHog Cloud, Debug information: [please copy/paste from https://us.posthog.com/settings/project-details#variables]
```

ioannisj:
## Problem

posthog-js sets `$device_type`, `$os`, and `$browser` from User-Agent parsing, but never sets `$device_model`. The native iOS and Android SDKs already send `$device_model`, and the server supports it end-to-end (taxonomy, property definitions, and the materialized ClickHouse `mat_$device_model` column), so web is the one surface leaving that column empty on mobile-web traffic. This makes the property an inconsistent dimension across SDKs and blocks device-model breakdowns for web.

Resolves PostHog/posthog#21955 (Add Device Make/Model for insights/web analytics): the server already supports `$device_model` end-to-end, so adding it to the web SDK is all that was missing.

## Changes

Adds `$device_model` capture to the browser SDK:

- New `packages/browser/src/utils/device-model-utils.ts` — an async `getDeviceModel()` resolver that calls `navigator.userAgentData.getHighEntropyValues(['model'])`. It is browser-only by design (kept out of `@posthog/core`, which React Native shares and which has its own native `$device_model`).
- During `init`, the value is resolved **once** (fire-and-forget, off the capture hot path) and registered as a `$device_model` super-property so it rides on subsequent events. The first event may legitimately miss it; that is acceptable for a stable per-device breakdown dimension, so there is no buffering.
- New opt-out config `disable_device_model` (default `false`).
- Corrected a stale comment in `utils/blocked-uas.ts` (the old comment feared a permission prompt; `getHighEntropyValues` does not prompt).

Behavior notes:
- Chromium-only and only meaningful on Android: resolves to `undefined` on Safari/Firefox and an empty string on desktop — nothing is registered in those cases (no `''` cardinality).
- Errors (including `NotAllowedError` from a `ch-ua-high-entropy-values` Permissions-Policy block) are caught and resolve to `undefined`; a debug-level log makes silent iframe/reverse-proxy blocks diagnosable.
- The raw OEM model code is sent (e.g. `Pixel 7`), with no client-side marketing-name mapping. Property name is exactly `$device_model`, matching the native SDKs, server taxonomy, and ClickHouse column. No property allowlist needed updating (the browser SDK uses an opt-out `property_denylist`).
- Confirmed end-to-end on the Pixel_8 emulator. `navigator.userAgentData.getHighEntropyValues(['model'])` only exists in a secure context, so had to run through `http://localhost:3000` and not `http://10.0.2.2:3000` for it to work

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with Claude Code. The on-by-default opt-out (`disable_device_model`) defers the privacy/product call to review; if on-by-default is not wanted, the same code flips to an opt-in `capture_device_model`.


packages/browser/src/utils/device-model-utils.ts
```diff
@@ -0,0 +1,28 @@
...
+    if (!uaData?.getHighEntropyValues) {
+        return undefined
+    }
+
+    try {
+        const hints = await uaData.getHighEntropyValues(['model'])
+        const model = hints?.model
+        return isString(model) && model.length > 0 ? model : undefined
+    } catch (e) {
+        logger.info('Unable to resolve $device_model from userAgentData.getHighEntropyValues', e)
```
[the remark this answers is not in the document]
ioannisj: I don't see logger.debug being used in the browser source, so .info should be fine

packages/browser/src/posthog-core.ts
```diff
@@ -838,6 +841,17 @@ export class PostHog implements PostHogInterface {
...
+        // Not awaited — the first event may miss $device_model, which is fine for a stable per-device dimension.
+        if (!this.config.disable_device_model) {
+            getDeviceModel()
+                .then((model) => {
+                    if (model) {
+                        this.register({ [DEVICE_MODEL]: model })
+                    }
+                })
+                .catch(__NOOP)
+        }
```
marandaneto: we do this only once right?

packages/browser/src/types.ts
```diff
@@ -152,6 +152,19 @@ export type PostHogInterface = Omit<BasePostHogInterface, 'config' | 'init'>
...
+     *
+     * When enabled (the default), PostHog resolves the hardware model once during init via
+     * `navigator.userAgentData.getHighEntropyValues(['model'])` and registers it as the raw OEM
+     * code (e.g. `Pixel 7`). This is Chromium-only and only meaningful on Android — it resolves to
+     * `undefined` on Safari/Firefox and to an empty string on desktop, in which cases nothing is
+     * registered.
+     *
+     * @default false
+     */
+    disable_device_model?: boolean
```
marandaneto: whats the benefit of having this? we probably dont need that since its not a sensitive value or anything

marandaneto (approved):

packages/browser/src/posthog-core.ts
ioannisj: yes this is once at init (non-blocking) and also survives a reset()

packages/browser/src/types.ts
ioannisj: No strong opinion here. Kept it to be inline with the disable_* opt-out family along with the fact that [getHighEntropyValues](https://developer.mozilla.org/en-US/docs/Web/API/NavigatorUAData/getHighEntropyValues) is still experimental. I can drop if you think that this increases the public api surface with no aded value

marandaneto: since its Experimental, makes sense, worth adding a comment about that then for disable_device_model jsdocs, missed that point

packages/browser/src/types.ts
```diff
@@ -152,6 +152,22 @@ export type PostHogInterface = Omit<BasePostHogInterface, 'config' | 'init'>
...
+     * code (e.g. `Pixel 7`). This is Chromium-only and only meaningful on Android — it resolves to
+     * `undefined` on Safari/Firefox and to an empty string on desktop, in which cases nothing is
+     * registered.
+     *
+     * This opt-out is offered because `getHighEntropyValues` is still experimental and Chromium-only
+     * (not yet a cross-browser standard); set it to `true` to skip the call entirely.
+     *
+     * @default false
+     */
+    disable_device_model?: boolean
```
marandaneto: Lets make this camel case since its the new convention for v2

marandaneto (approved):

packages/browser/src/types.ts
ioannisj: This will be the only camelCased top-level config though, will do

marandaneto: thats ok, so one less breaking change for v2
see @dustinbyrne RFC

PR #3987, opened 2026-06-26
feat(replay): add get_current_url override for client-side URL targeting

TueHaulund:
## Problem

Session replay V2 trigger groups, the replay URL blocklist, survey URL conditions, product tour URL conditions, and web experiment URL conditions all match against `window.location.href` directly — which ignores any `$current_url` rewritten in `before_send`. Electron/desktop apps served from a generated host (e.g. `https://<random>.skin/...`) can't match their own URL patterns, so these features silently stop targeting correctly for those sources. Reported via an enterprise support ticket (replay triggers stopped recording on desktop apps).

## Changes

Adds a `get_current_url` config hook to override the URL used for **client-side URL targeting** only (not the captured `$current_url` — that stays driven by `before_send`):

```js
posthog.init('<token>', { get_current_url: (defaultUrl) => '<logical-url>' })
```

- New `get_current_url?: (defaultUrl: string) => string` on `PostHogConfig`, default `undefined`.
- Shared helper `getTargetingUrl(instance)` resolves `config.get_current_url?.(href) ?? href`, with `applyUrlTargetingOverride` for callers that already have their own URL source. Falls back to the browser URL if the override throws or returns a non-string/empty value.
- Routed five URL-targeting sites through it:
  - replay URL triggers — `checkUrlTriggerConditions`
  - replay URL blocklist — `checkUrlBlocklist`
  - surveys — `doesSurveyUrlMatch`
  - product tours — `doesTourUrlMatch` (cache now keyed on the resolved URL)
  - web experiments — `webExperimentUrlValidationMap` via `_matchUrlConditions`

Unset = identical behavior to today, so no change for existing users.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — DRI: @TueHaulund

Built with Claude Code. Unit tests cover all five surfaces plus override edge cases (empty-string/throw fallback, tour cache invalidation); existing affected suites pass; typecheck + lint clean. No manual e2e testing.


fasyy612 (approved):
LGTM, just one question. Should we add get_currentl_url for autocapture-utils.ts line 15 as well?

TueHaulund:
@fasyy612 nice catch! Adding this now

PR #3994, opened 2026-06-26
feat: add official PostHog OpenFeature server provider (Node)

gustavohstrassburger:
Based on [PostHog/posthog-python#695](https://github.com/PostHog/posthog-python/pull/695), adds **`@posthog/openfeature-node-provider`** — the official PostHog provider for the [OpenFeature](https://openfeature.dev) **server** SDK (`@openfeature/server-sdk`), backed by [`posthog-node`](https://posthog.com/docs/libraries/node).

> **Split into two PRs** per review ([@marandaneto](https://github.com/marandaneto): _"could be 2 PRs tbh"_). This PR is the **server/Node** provider; the **web/browser** provider (`@posthog/openfeature-web-provider`, backed by `posthog-js`) is in **#4069**.

OpenFeature ships two SDKs with incompatible Provider contracts — the server SDK is **async** with a **per-call** evaluation context (distinct id from `targetingKey`), versus the web SDK's synchronous, static-context model. So the two providers are separate packages, each pulling in only its own paradigm's SDK. The Python parent PR is server-only (no browser runtime); this pair covers both.

## Behaviour

The server model is stateless and multi-user: the distinct id arrives per evaluation from the context's `targetingKey`, and resolution is async. Context mapping: `targetingKey` → `distinctId`, reserved `groups`/`groupProperties` → PostHog equivalents, every other attribute → `personProperties`.

Flag-type mapping (via `getFeatureFlagResult`): boolean → `enabled`, string → multivariate `variant`, number → parsed `variant`, object → JSON `payload`. A missing flag resolves to `FLAG_NOT_FOUND`; asking for a type the flag can't provide resolves to `TYPE_MISMATCH` (caller gets its default, per spec).

**One deliberate divergence from Python:** posthog-node's `FeatureFlagResult` has no `reason` field, so reason mapping simplifies to `enabled ? TARGETING_MATCH : DEFAULT` and there's no `posthog_reason` metadata.

## Options

- `sendFeatureFlagEvents` (default `true`) — control `$feature_flag_called` capture.
- `defaultDistinctId` — when set, evaluations without a `targetingKey` use it instead of raising `TARGETING_KEY_MISSING`.

## Publishing

Marked for release (changeset + entry in the `release.yml` publish matrix); created on npm under the `@posthog` scope on first publish, gated on the standard merge + `NPM Release` approval flow.

## Docs

Usage lives in the PostHog docs (kept out of the package README so it doesn't drift): **PostHog/posthog.com#18105** — `/docs/feature-flags/installation/openfeature-js` (covers both the node and web providers).

## Testing

- 20 unit tests (type mappings, error cases, context split, distinct-id resolution, initialize preload) + end-to-end tests through the real `@openfeature/server-sdk` client.
- ✅ Build (rslib esm + cjs + declarations) · ✅ Lint · ✅ Tests · ✅ `pnpm install --frozen-lockfile` consistent


packages/openfeature-web-provider/src/provider.ts
```diff
@@ -0,0 +1,125 @@
...
+        if (!subscribed) {
+          return
+        }
+        unsubscribe()
+        resolve()
+      })
+      subscribed = true
+      this._client.reloadFeatureFlags()
+    })
+  }
```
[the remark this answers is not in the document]
gustavohstrassburger: Good catch — fixed in 615ff92. \`_reloadFlags\` now has a timeout safety net: a new \`reloadTimeoutMs\` option (default 5000ms) wraps the wait so \`initialize()\`/\`onContextChange()\` resolve and the provider becomes READY even if posthog-js never delivers its flags callback (uninitialised SDK, silent network failure, etc.), instead of hanging in NOT_READY. On settle it clears the timer and unsubscribes. I chose resolve-on-timeout rather than reject so the provider serves whatever flags are cached rather than going to ERROR. Added a unit test covering the never-fires path.

packages/openfeature-provider/src/mapping.ts
```diff
@@ -0,0 +1,153 @@
...
+
+/**
+ * A `undefined` result means the flag does not exist (or was archived) —
+ * `getFeatureFlagResult` returns a populated result with `enabled: false` for a
+ * flag that exists but did not match. Surface the former as the OpenFeature
+ * `FLAG_NOT_FOUND` error so callers get their default value.
+ */
+function ensureResolved(result: PostHogFlagResult | undefined, flagKey: string): PostHogFlagResult {
+  if (result == null) {
+    throw new FlagNotFoundError(`Flag '${flagKey}' not found or disabled.`)
```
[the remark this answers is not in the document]
gustavohstrassburger: Agreed — fixed in 615ff92. The message is now `Flag '...' not found.` (dropped "or disabled"), matching the documented behaviour that `undefined` only means a missing/archived flag while a disabled flag returns `{ enabled: false }`.

packages/openfeature-provider/src/__tests__/server-provider.spec.ts
```diff
@@ -0,0 +1,230 @@
...
+    })
+
+    it('throws TypeMismatch when there is no object payload', async () => {
+      const { client } = makeClient({ key: 'flag', enabled: true, payload: 'a string' })
+      const provider = new PostHogServerProvider(client)
+      await expect(provider.resolveObjectEvaluation('flag', {}, CTX)).rejects.toMatchObject({
+        code: ErrorCode.TYPE_MISMATCH,
+      })
+    })
+  })
```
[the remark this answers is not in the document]
gustavohstrassburger: Done in 615ff92. Confirmed it.each/test.each is an established convention in this repo (29 of 69 spec files), so I consolidated the four resolution describe blocks into it.each tables — one happy-path table and one error table — in both server-provider.spec.ts and web-provider.spec.ts.

packages/openfeature-node-provider/src/mapping.ts
```diff
@@ -0,0 +1,151 @@
...
+  const resolved = ensureResolved(result, flagKey)
+  return { value: resolved.enabled, variant: resolved.variant, reason: reasonFor(resolved) }
+}
+
+export function resolveStringDetails(
+  result: PostHogFlagResult | undefined,
+  flagKey: string
+): ResolutionDetails<string> {
+  const resolved = ensureResolved(result, flagKey)
+  if (resolved.variant === undefined) {
```
haacked, line 118: `blocking`: A disabled or unmatched flag (`{ enabled: false, variant: undefined }`) throws `TypeMismatchError` instead of resolving to the caller's default. OpenFeature's SDK catches the throw, sets `errorCode: TYPE_MISMATCH` and `reason: ERROR`, and fires every registered error hook. So a consumer with error hooks wired to their error tracker gets paged on every ordinary disabled-flag read.

The [Python provider](https://github.com/PostHog/posthog-python/pull/695) that this PR mirrors returns the caller's default with `Reason.DEFAULT` and no error in exactly this case, only raising `TypeMismatchError` when `enabled` is true. The identical bug exists in [`packages/openfeature-web-provider/src/mapping.ts`](https://github.com/posthog/posthog-js/blob/ac4660a005ffb203c9281a9ae1045da5a6fdd6f7/packages/openfeature-web-provider/src/mapping.ts#L116). Neither `provider.spec.ts` has a test for disabled flags through these resolvers.

Add the `!resolved.enabled` guard before the throw in all three resolvers in both packages, and thread the caller's default value through:

```diff
 export function resolveStringDetails(
   result: PostHogFlagResult | undefined,
-  flagKey: string
+  flagKey: string,
+  defaultValue: string
 ): ResolutionDetails<string> {
   const resolved = ensureResolved(result, flagKey)
   if (resolved.variant === undefined) {
-    // A boolean flag has no string variant. Surface a type mismatch so the
-    // caller gets its default value (per the OpenFeature spec) rather than a
-    // surprising "true"/"false" string.
+    if (!resolved.enabled) {
+      return { value: defaultValue, reason: StandardResolutionReasons.DEFAULT }
+    }
+    // An enabled boolean flag has no string variant: this is a genuine mismatch.
     throw new TypeMismatchError(`Flag '${flagKey}' has no string variant (boolean flag).`)
   }
   return { value: resolved.variant, variant: resolved.variant, reason: reasonFor(resolved) }
 }
```

Apply the same guard in `resolveNumberDetails` (before the `variant === undefined` throw) and `resolveObjectDetails` (before the payload-type throw). Pass `_defaultValue` through at the call sites in both `provider.ts` files. Add disabled-flag cases to each `it.each` table asserting `reason: DEFAULT` and no `errorCode`.

packages/openfeature-node-provider/package.json
```diff
@@ -0,0 +1,74 @@
...
+    "openfeature",
+    "feature-flags",
+    "node",
+    "server"
+  ],
+  "peerDependencies": {
+    "@openfeature/core": "^1.11.0",
+    "@openfeature/server-sdk": "^1.7.0",
+    "posthog-node": ">=5.21.2"
+  },
```
haacked, line 62: `blocking`: The `@openfeature/server-sdk@^1.7.0` lower bound conflicts with the `@openfeature/core@^1.11.0` peer declared above. Every server-sdk version from 1.7.0 through 1.16.x requires `@openfeature/core` below 1.11.0, so no single installed version satisfies both constraints. A consumer on any server-sdk in that range hits an `ERESOLVE` failure on npm or yarn install; pnpm only warns unless `strict-peer-dependencies` is set.

`server-sdk@1.17.0` is the first version whose own `core` peer (`^1.6.0`) overlaps with `1.11.0`:

```diff
   "peerDependencies": {
     "@openfeature/core": "^1.11.0",
-    "@openfeature/server-sdk": "^1.7.0",
+    "@openfeature/server-sdk": "^1.17.0",
     "posthog-node": ">=5.21.2"
   },
```

packages/openfeature-web-provider/src/provider.ts
```diff
@@ -0,0 +1,151 @@
...
+    this._sendFeatureFlagEvents = options.sendFeatureFlagEvents ?? true
+    this._reloadTimeoutMs = options.reloadTimeoutMs ?? 5000
+  }
+
+  async initialize(context?: EvaluationContext): Promise<void> {
+    await this._reconcile(context)
+  }
+
+  async onContextChange(_oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
+    await this._reconcile(newContext)
```
haacked, line 84: `suggestion`: `onContextChange` reconciles on every call even when the context is unchanged. `@openfeature/web-sdk`'s `setContext` runs the provider's handler on every call with no equality check, and `_reconcile` doesn't diff against the previous context either. So a host that re-sets an equivalent context (a React integration passing a freshly-constructed object on every render, say) triggers a `$groupidentify` event and a `reloadFeatureFlags()` request on every call, since consecutive calls are typically more than 5ms apart and don't fold into the SDK's debounce.

Diff `newContext` against the previous context and skip reconciliation when nothing changed:

```ts
async onContextChange(oldContext: EvaluationContext, newContext: EvaluationContext): Promise<void> {
  if (contextsEqual(oldContext, newContext)) {
    return
  }
  await this._reconcile(newContext)
}
```

`groupProperties` is nested, so this needs a deep compare, not a shallow one. `initialize()` should keep calling `_reconcile` unconditionally since first load always needs it.

packages/openfeature-node-provider/src/provider.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  async initialize(): Promise<void> {
+    // Preload locally-evaluated flag definitions. reloadFeatureFlags() safely
+    // no-ops without a personalApiKey, and posthog-node catches its own poller
+    // errors, so a preload failure must not make the OpenFeature client
+    // un-ready — remote evaluation still works.
+    try {
+      await this._client.reloadFeatureFlags()
+    } catch {
+      // Intentionally ignored: remote evaluation remains available.
```
haacked: `suggestion`: This swallows every preload error with no log, not just the documented no-`personalApiKey` no-op case. A genuine misconfiguration during `reloadFeatureFlags()` (bad host, network failure) disappears silently, so a user whose local evaluation never warms up has no visibility. The Python reference that this PR mirrors logs a warning with the exception in the equivalent spot.

```diff
     try {
       await this._client.reloadFeatureFlags()
-    } catch {
-      // Intentionally ignored: remote evaluation remains available.
+    } catch (err) {
+      // Remote evaluation still works, so don't block readiness, but surface
+      // the failure so a real misconfiguration isn't invisible.
+      console.warn('[PostHogServerProvider] initialize() flag preload failed; remote evaluation still available.', err)
     }
```

`no-console` is already disabled for this package in `.eslintrc.cjs`.

packages/openfeature-web-provider/src/__tests__/provider.spec.ts
```diff
@@ -0,0 +1,225 @@
...
+        (p) => p.resolveObjectEvaluation('flag', {}),
+        { value: { a: 1 } },
+      ],
+    ])('resolves %s', (_name, result, resolve, expected) => {
+      const { client } = makeClient(result)
+      expect(resolve(new PostHogWebProvider(client))).toMatchObject(expected)
+    })
+
+    it.each<[string, FlagResult | undefined, Resolve, ErrorCode]>([
+      [
```
haacked, line 126: `suggestion`: This table only covers a string read on a boolean flag and a missing flag. The node package's equivalent table also has a "number from a non-numeric variant" case and an "object from a non-object payload" case, exercising `resolveNumberDetails`'s `!Number.isFinite` throw and `resolveObjectDetails`'s non-object-payload throw. Neither branch is exercised here, so a regression in either check on the web side ships silently.

Add the two missing rows, mirroring the node spec:

```ts
[
  'number from a non-numeric variant',
  { key: 'flag', enabled: true, variant: 'not-a-number' },
  (p) => p.resolveNumberEvaluation('flag', 0),
  ErrorCode.TYPE_MISMATCH,
],
[
  'object from a non-object payload',
  { key: 'flag', enabled: true, variant: 'x', payload: 'not-an-object' },
  (p) => p.resolveObjectEvaluation('flag', {}),
  ErrorCode.TYPE_MISMATCH,
],
```

packages/openfeature-web-provider/src/mapping.ts
```diff
@@ -0,0 +1,150 @@
+/**
```
haacked, line 1: `question`: This file is identical to [`openfeature-node-provider/src/mapping.ts`](https://github.com/posthog/posthog-js/blob/ac4660a005ffb203c9281a9ae1045da5a6fdd6f7/packages/openfeature-node-provider/src/mapping.ts#L1-L151) apart from two words in the header comment, and imports only from `@openfeature/core`, not from `posthog-js`, `posthog-node`, or any runtime SDK. So the dependency-isolation argument that justifies splitting the providers into two packages doesn't apply to the mapping layer: a shared internal module would peer on `@openfeature/core`, which both already require, adding no extra peer constraints. The concrete cost of duplication is drift: the disabled-flag type-mismatch bug exists identically in both files, and any fix lands twice and stays in sync.

Was a shared internal module considered and set aside deliberately? The likely counter-argument is that with `bundle: false` in both `rslib.config.ts`, extracting means either publishing a third `@posthog/openfeature-*` package or setting `bundle: true` here, which would deviate from the `@posthog/mcp` convention these packages otherwise follow. For 150 lines with two consumers that's a defensible tradeoff, just want to confirm it was weighed rather than copy-paste.

packages/openfeature-web-provider/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "lint": "eslint src",
+    "lint:fix": "eslint src --fix",
+    "build": "rslib build",
+    "dev": "rslib build --watch",
+    "test:unit": "jest",
+    "prepublishOnly": "pnpm lint && pnpm test:unit && pnpm build",
+    "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz"
+  },
+  "files": [
+    "dist/"
```
haacked, line 49: `nit`: No `engines` block here, while the node provider and every other sibling package (`@posthog/mcp`, `@posthog/nextjs-config`, `@posthog/nuxt`) declare one despite targeting the web. Without it, consumers on unsupported Node versions get no install-time warning when building or testing this package.

```diff
     "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz"
   },
+  "engines": {
+    "node": "^20.20.0 || >=22.22.0"
+  },
   "files": [
     "dist/"
   ],
```

packages/openfeature-web-provider/src/provider.ts
```diff
@@ -0,0 +1,151 @@
...
+    await this._reloadFlags()
+  }
+
+  private _reloadFlags(): Promise<void> {
+    return new Promise<void>((resolve) => {
+      let settled = false
+      let subscribed = false
+
+      const finish = (): void => {
+        if (settled) {
```
haacked, line 136: `nit`: `finish` references `timer` and `unsubscribe` before either `const` below it is declared. It works today only because `finish` never actually runs until after both are assigned, so a future edit that lets it run earlier would hit a temporal-dead-zone throw instead of a type error. Declaring `unsubscribe` and `timer` above `finish` would let the closure read top-to-bottom.

haacked (commented):
Nice addition! Two blocking issues inline: disabled flags throw instead of resolving to the default, and the node package's server-sdk peer range can't actually be satisfied. A few non-blocking suggestions too.

.changeset/openfeature-providers.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/openfeature-node-provider': minor
+'@posthog/openfeature-web-provider': minor
```
marandaneto: could be 2 PRs tbh

marandaneto:
worth adding new packages to https://github.com/PostHog/posthog-js/blob/main/.github/pull_request_template.md

marandaneto:
also append to https://github.com/PostHog/posthog-js#packages

marandaneto:
also append to https://github.com/PostHog/posthog-js/blob/main/CHANGELOG.md#packages

marandaneto:
can we add examples to https://github.com/PostHog/posthog-js/tree/main/examples so we know/can test this outside unit tests? or extend one of our examples for node/web

.changeset/openfeature-providers.md
gustavohstrassburger: Done — split into two PRs. This one (#3994) now ships only the **server** provider (`@posthog/openfeature-node-provider`); the **web** provider (`@posthog/openfeature-web-provider`) moved to #4069. Each has its own changeset and release-matrix entry.

packages/openfeature-node-provider/src/mapping.ts
gustavohstrassburger: Fixed in 7fffb7c (node) and 85f34db (web) — added the `!resolved.enabled` guard to all three resolvers, so a disabled/unmatched flag resolves to the caller's default with `reason: DEFAULT` instead of throwing. Threaded the default value through `resolveString/Number/Object` and their `provider.ts` call sites. Added disabled-flag test cases asserting `reason: DEFAULT` and no `errorCode`.

packages/openfeature-node-provider/package.json
gustavohstrassburger: Fixed in 7fffb7c — bumped the `@openfeature/server-sdk` peer floor to `^1.17.0`, the first version whose own `@openfeature/core` peer overlaps `^1.11.0`.

packages/openfeature-web-provider/src/provider.ts
gustavohstrassburger: Addressed on the web PR #4069 in 85f34db — `onContextChange` now deep-compares old vs new context (handling nested `groupProperties`) and skips reconciliation when nothing changed. `initialize()` still reconciles unconditionally.

packages/openfeature-node-provider/src/provider.ts
gustavohstrassburger: Fixed in 7fffb7c — `initialize()` now `console.warn`s on a preload failure (with the error) instead of swallowing it, so a genuine misconfiguration on a client set up for local evaluation is visible. Remote evaluation still works, so readiness isn't blocked.

packages/openfeature-web-provider/src/__tests__/provider.spec.ts
gustavohstrassburger: Addressed on the web PR #4069 in 85f34db — added the missing `number from a non-numeric variant` and `object from a non-object payload` cases (plus disabled-flag cases) to the parameterised tables.

packages/openfeature-web-provider/src/mapping.ts
gustavohstrassburger: Considered and set aside deliberately. With `bundle: false` (the `@posthog/mcp` convention both packages follow), sharing the mapping means either publishing a third `@posthog/openfeature-*` package or switching to `bundle: true` — both deviate from convention for ~150 lines. We accept the duplication; the drift risk you flagged (the disabled-flag bug) is now fixed in both packages in lockstep (7fffb7c / 85f34db). Happy to revisit if a third consumer appears.

packages/openfeature-web-provider/package.json
gustavohstrassburger: Addressed on the web PR #4069 in 85f34db — added an `engines` block (`^20.20.0 || >=22.22.0`) matching the node package.

packages/openfeature-web-provider/src/provider.ts
gustavohstrassburger: Addressed on the web PR #4069 in 85f34db — restructured `_reloadFlags` to collect cleanups in an array declared above `finish`, so the closure no longer forward-references `timer`/`unsubscribe`.

gustavohstrassburger:
@marandaneto addressed your follow-ups:

- **PR template**, **root README `#packages`**, **root CHANGELOG `#packages`** — added `@posthog/openfeature-node-provider` here (7fffb7c) and `@posthog/openfeature-web-provider` on #4069 (85f34db).
- **Examples** — added `examples/example-openfeature-node` here (7fffb7c) and `examples/example-openfeature-web` on #4069 (85f34db), so both providers can be exercised outside the unit tests.

packages/openfeature-node-provider/src/provider.ts
```diff
@@ -0,0 +1,148 @@
...
+
+  async initialize(): Promise<void> {
+    // Preload locally-evaluated flag definitions. reloadFeatureFlags() safely
+    // no-ops without a personalApiKey, and posthog-node catches its own poller
+    // errors, so a preload failure must not make the OpenFeature client
+    // un-ready — remote evaluation still works.
+    try {
+      await this._client.reloadFeatureFlags()
+    } catch (err) {
+      // Remote evaluation still works, so don't block readiness — but surface
```
haacked: `suggestion`: This catch can't fire for the misconfigurations the comment names. `reloadFeatureFlags()` awaits `featureFlagsPoller.loadFeatureFlags(true)`, which wraps its internal `_loadFeatureFlags()` in its own `.catch()` and only debug-logs. Every real failure (bad host, invalid key, network error, 401/403/429) routes through posthog-node's `onError` callback instead of a rejection, but that callback's event emitter drops the event when nothing's listening. So the promise here never rejects in production, and `console.warn` never runs for the cases this comment describes.

To actually surface those failures, listen on the channel posthog-node uses instead:

```diff
   async initialize(): Promise<void> {
-    // Preload locally-evaluated flag definitions. reloadFeatureFlags() safely
-    // no-ops without a personalApiKey, and posthog-node catches its own poller
-    // errors, so a preload failure must not make the OpenFeature client
-    // un-ready — remote evaluation still works.
-    try {
-      await this._client.reloadFeatureFlags()
-    } catch (err) {
-      // Remote evaluation still works, so don't block readiness — but surface
-      // the failure so a genuine misconfiguration (bad host, network error) on
-      // a client set up for local evaluation isn't invisible.
-      // eslint-disable-next-line no-console
-      console.warn('[PostHogServerProvider] initialize() flag preload failed; remote evaluation still available.', err)
-    }
+    // posthog-node swallows poller errors internally and surfaces the
+    // client/network-error class via its `error` event, so listen on that
+    // channel around the reload. Remote evaluation still works either way,
+    // so this never blocks readiness.
+    let preloadError: Error | undefined
+    const unsubscribe = this._client.on('error', (err: Error) => {
+      preloadError = err
+    })
+    try {
+      await this._client.reloadFeatureFlags()
+    } finally {
+      unsubscribe()
+    }
+    if (preloadError) {
+      // eslint-disable-next-line no-console
+      console.warn('[PostHogServerProvider] initialize() flag preload failed; remote evaluation still available.', preloadError)
+    }
   }
```

One caveat: posthog-node only routes `ClientError` (401/403/429, bad response) through that event, so a raw network/DNS failure on a bad `host` still won't surface either way.

packages/openfeature-node-provider/src/__tests__/provider.spec.ts
```diff
@@ -0,0 +1,248 @@
...
+
+  describe('initialize', () => {
+    it('preloads flags via reloadFeatureFlags', async () => {
+      const { client, reloadFeatureFlags } = makeClient(undefined)
+      const provider = new PostHogServerProvider(client)
+      await provider.initialize()
+      expect(reloadFeatureFlags).toHaveBeenCalledTimes(1)
+    })
+
+    it('does not reject when preloading fails', async () => {
```
haacked: `suggestion`: This test proves `initialize()` doesn't throw when `reloadFeatureFlags` rejects, but the real posthog-node client never rejects here (see [provider.ts:84](https://github.com/PostHog/posthog-js/blob/92704a5ade888439a351cb4d83a64f9ce8990f49/packages/openfeature-node-provider/src/provider.ts#L84)), so it doesn't cover the actual failure path.

If you switch to listening on the client's error event there, test that instead: emit an error on `'error'` and assert `console.warn` fires with it.

packages/openfeature-node-provider/src/mapping.ts
```diff
@@ -0,0 +1,165 @@
...
+  defaultValue: number
+): ResolutionDetails<number> {
+  const resolved = ensureResolved(result, flagKey)
+  if (resolved.variant === undefined) {
+    if (!resolved.enabled) {
+      return { value: defaultValue, reason: StandardResolutionReasons.DEFAULT }
+    }
+    throw new TypeMismatchError(`Flag '${flagKey}' has no variant to parse as a number.`)
+  }
+  const value = Number(resolved.variant)
```
haacked, line 144: `nit`: `Number('')` evaluates to `0`, not `NaN`, so a multivariate flag with an empty-string variant resolves to `0` instead of throwing `TypeMismatchError`. The `'not-a-number'` test case only covers the `NaN` branch; worth an added case if you want this edge nailed down, though real variant keys are unlikely to be empty strings.

examples/example-openfeature-node/index.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  // 2. Register the PostHog provider with OpenFeature.
+  await OpenFeature.setProviderAndWait(new PostHogServerProvider(posthog))
+  const client = OpenFeature.getClient()
+
+  // 3. Evaluate flags through the vendor-neutral OpenFeature API. The distinct
+  //    id comes from the evaluation context's targetingKey; other attributes
+  //    map to person/group properties. Swap these keys for real flags.
+  const context = { targetingKey: 'user_distinct_id', plan: 'enterprise' }
+  const result = {
```
haacked, line 28: `nit`: These three flag reads run sequentially even though they're independent. `Promise.all` would model the better pattern here since example code tends to get copy-pasted, though at 3 flags in a one-shot demo the actual cost is negligible.

haacked (approved):
Both blocking issues from before are fixed. One follow-up suggestion inline on `initialize()`'s error handling, plus a couple of nits.

packages/openfeature-node-provider/src/provider.ts
gustavohstrassburger: Good catch — confirmed `reloadFeatureFlags()` never rejects (poller `.catch()` at feature-flags.ts:748 only debug-logs), so the try/catch was dead code. Fixed in c2ad670 by listening on the client's `error` event around the reload, per your suggestion. Kept your caveat in the comment: only `ClientError` (401/403/429, bad response) routes through that event, so a raw network/DNS failure on a bad host still won't surface.

packages/openfeature-node-provider/src/__tests__/provider.spec.ts
gustavohstrassburger: Fixed in c2ad670 — the test now emits the client's `error` event while the reload is in flight and asserts `console.warn` fires with it, plus a companion case asserting no warning on a clean preload.

packages/openfeature-node-provider/src/mapping.ts
gustavohstrassburger: Fixed in c2ad670 — `resolveNumberDetails` now rejects a blank variant (`variant.trim() === ''`) before the `Number.isFinite` check, so an empty/whitespace variant throws `TypeMismatchError` instead of resolving to 0. Added an empty-string variant case to the throws table.

(The web package's mapping has the same line; since it's already merged I left it for a follow-up rather than reintroduce cross-package churn here.)

examples/example-openfeature-node/index.ts
gustavohstrassburger: Done in c2ad670 — the three independent reads now run via `Promise.all`.

packages/openfeature-node-provider/src/provider.ts
```diff
@@ -0,0 +1,161 @@
...
+    return this._client.getFeatureFlagResult(flagKey, distinctId, {
+      groups: Object.keys(groups).length > 0 ? groups : undefined,
+      personProperties:
+        Object.keys(personProperties).length > 0 ? (personProperties as Record<string, string>) : undefined,
+      groupProperties:
+        Object.keys(groupProperties).length > 0
+          ? (groupProperties as Record<string, Record<string, string>>)
+          : undefined,
+      sendFeatureFlagEvents: this._sendFeatureFlagEvents,
+    })
```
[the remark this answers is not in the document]
gustavohstrassburger: Good catch on the intent — you're right that the values must reach `getFeatureFlagResult` unchanged, since the flags API matches on the native JSON type (e.g. a numeric `age >= 21` filter), so serializing to strings would silently break comparisons.

The one wrinkle is that removing the casts outright doesn't compile: posthog-node's *public* option type narrows `personProperties` to `Record<string, string>` (and `groupProperties` to `Record<string, Record<string, string>>`), even though the internal type and the flags API accept any JSON value. So the assertion is needed purely to bridge that too-narrow upstream type — it changes nothing at runtime; the values are forwarded exactly as received.

Fixed in [`9f8a7fe`](https://github.com/PostHog/posthog-js/commit/9f8a7fe24c6ab9d30026fe8a9850554e7fc1fdd4): values are passed through unchanged with a documented comment, empty records collapse to `undefined` via a `nonEmpty()` helper, and a new test (`forwards non-string property values unchanged`) asserts that `age: 42`, `beta: true`, and `seats: 25` reach the client without coercion.

PR #3995, opened 2026-06-26
feat(mcp): add instrumentMutator for framework serverMutator hooks

lucasheriques:
## What

Adds `instrumentMutator(posthog, options?)` to `@posthog/mcp` — a point-free `(server) => server` helper for framework server-mutation hooks, most notably [`@rekog/mcp-nest`](https://github.com/rekog-labs/MCP-Nest)'s `serverMutator`:

```ts
McpModule.forRoot({
  serverMutator: instrumentMutator(posthog),
})
```

## Why

`instrument()` returns the analytics *handle*, not the server. So the natural-looking `serverMutator: (s) => instrument(s, posthog)` silently replaces the server with the handle and breaks the module — you have to remember `(s) => { instrument(s, posthog); return s }`. `instrumentMutator` returns the server for you, so the mutator is point-free and there's nothing to get wrong. It generalizes to any `(server) => server` hook, and it's the same shape Sentry exposes with `wrapMcpServerWithSentry`.

This came out of reviewing #3993 (the single `setRequestHandler` interceptor that makes nest's late-registered handlers get captured). The helper is the ergonomic layer on top.

## Notes

- The point-free form discards the analytics handle. Custom-event users (`analytics.capture(...)`) still call `instrument()` directly — documented in the JSDoc.
- Generic over the server type, so it returns the exact `McpServer`/`Server` it was handed.

## Tests

Two tests: it returns the same server instance (the contract), and the returned server actually captures tool calls when handlers are registered after the mutator (the nest-mcp shape). Full `@posthog/mcp` suite green (332). Changeset: minor.

## Follow-up

Once this ships, the context-mill `mcp-analytics` skill's nest path (PostHog/context-mill#204) gets simplified to inject `serverMutator: instrumentMutator(posthog)` instead of the multi-line arrow.

---

### DIFF
--- a/.changeset/mcp-instrument-mutator.md
+++ b/.changeset/mcp-instrument-mutator.md
@@ -1,0 +1,5 @@
+---
+'@posthog/mcp': minor
+---
+
+Add `instrumentMutator(posthog, options?)` — a point-free `(server) => server` helper for framework server-mutation hooks like `@rekog/mcp-nest`'s `serverMutator`. It instruments the server and returns it, so `serverMutator: instrumentMutator(posthog)` just works (no need to remember that `instrument()` returns the analytics handle, not the server).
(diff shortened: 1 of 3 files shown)


DanielVisca (approved):
Lets goo 👍🏽

Issue #3702, opened 2026-05-29
Feature request: (React-Native) add allowlist/ignorelist option for screen autocapture


ioannisj:
## Is your feature request related to a problem?

Currently, the only way to control which screens are tracked when `captureScreens: true` is to use a before_send hook

```ts
const TRACKED_SCREENS = new Set(['HomeScreen', 'ProductScreen'])

PostHog('<api-key>', {
  captureScreens: true,
  before_send: (event) => {
    if (event?.event === '$screen') {
      const screenName = event.properties?.['$screen_name']
      return TRACKED_SCREENS.has(screenName) ? event : null
    }
    return event
  },
})
```

This works but is unintuitive since users familiar with the web SDK's autocapture config naturally reach for an allowlist/ignorelist pattern

## Describe the solution you'd like

Proposed API:

```ts
// PostHogOptions
export type PostHogCaptureScreensOptions = {
    allowlist?: (string | RegExp)[]
    ignorelist?: (string | RegExp)[]
  }

captureScreens?: boolean | PostHogCaptureScreensOptions

// Usage
PostHog('<api-key>', {
  captureScreens: {
    allowlist: ['HomeScreen', 'ProductScreen'],   // only capture these
    ignorelist: ['AuthScreen', 'OnboardingScreen'], // or exclude these
  },
})
```

Both string exact-match and regex should be supported, mirroring the web autocapture convention. allowlist and ignorelist should be mutually exclusive or at least clearly documented if both are provided.

## Describe alternatives you've considered

use before_send as shown above

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

marandaneto:
or `denylist`, i'd just match what the JS SDK uses

ioannisj:
> or `denylist`, i'd just match what the JS SDK uses

I think it's `url_ignorelist` so I just dropped url part

cacheflow:
I'm willing to work on this.

cacheflow:
I have a PR that I'm still working on [here](https://github.com/PostHog/posthog-js/pull/3996).

[...]

Issue #3193, opened 2026-03-03
Implement survey popup delay (surveyPopupDelaySeconds) for React Native


adboio:
## 🚨 IMPORTANT

This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list here.

## Summary

The posthog-react-native SDK does not implement the survey popup delay (`surveyPopupDelaySeconds`). The sorting and delay logic is commented out.

## Current State

- The `surveyPopupDelaySeconds` field exists on the `SurveyAppearance` type with a default of `0`
- The `sortSurveysByAppearanceDelay` function is **commented out** in `packages/react-native/src/surveys/PostHogSurveyProvider.tsx:188-192`
- There is a TODO comment: `// TODO: sort by appearance delay, implement delay` (line 117)
- No actual delay is applied before rendering surveys

## Expected Behavior

When `survey.appearance.surveyPopupDelaySeconds` is set to a value > 0:
1. The survey popup should be delayed by that many seconds before being shown
2. Surveys should be sorted by their delay value (shorter delays first)
3. After the delay, the SDK should re-check URL/conditions before rendering

## Reference Implementation

See posthog-js browser: `packages/browser/src/extensions/surveys.tsx` — `_handlePopover()` method which uses `setTimeout` for the delay and re-checks conditions.

## Tracking

This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude using the `/survey-sdk-audit` skill._

Issue #3192, opened 2026-03-03
Implement survey wait period (seenSurveyWaitPeriodInDays) for React Native


adboio:
## 🚨 IMPORTANT

This issue is likely user-facing in the main PostHog app, see [`surveyVersionRequirements.ts`](https://github.com/PostHog/posthog/blob/master/frontend/src/scenes/surveys/surveyVersionRequirements.ts). If you delete or close this issue, be sure to update the version requirements list here.

## Summary

The posthog-react-native SDK does not implement the survey wait period (`seenSurveyWaitPeriodInDays`) filtering logic. The code is commented out in `packages/react-native/src/surveys/getActiveMatchingSurveys.ts` (lines 47-52).

## Current State

- The `seenSurveyWaitPeriodInDays` field is available on the `Survey` type (from `@posthog/core`)
- The filtering logic in `getActiveMatchingSurveys.ts` has the wait period check **commented out**
- No `lastSeenSurveyDate` storage is implemented

## Expected Behavior

When `survey.conditions.seenSurveyWaitPeriodInDays` is set, the SDK should:
1. Store the date when any survey was last shown to the user
2. During `getActiveMatchingSurveys()`, skip surveys if the wait period has not yet elapsed since the last seen survey date

## Reference Implementation

See posthog-js browser: `packages/browser/src/extensions/surveys/surveys-extension-utils.tsx` — `hasWaitPeriodPassed()` function and `LAST_SEEN_SURVEY_DATE_KEY` storage.
See also: `packages/browser/src/extensions/utils/matcher-utils.ts` — `hasPeriodPassed()` utility.

## Tracking

This is tracked in the survey SDK feature parity issue: https://github.com/PostHog/posthog/issues/45658

_This issue was generated by Claude using the `/survey-sdk-audit` skill._

PR #3996, opened 2026-06-26
feat(react-native): allow ignoring specific screen names during autoc…

cacheflow:
## Problem

React Native developers using PostHog navigation autocapture need a way to prevent specific screens (such as tab containers `(tabs)`, utility screens like `+not-found`, or sensitive views) from triggering automatic screen-view tracking. Doing this helps keep event logs clean and prevents capturing unwanted screens. References #3702.

## Changes

We have introduced the `ignoreScreenNames` array option to the `autocapture` configuration block:

* **`packages/react-native/src/types.ts`**: Added `ignoreScreenNames?: string[]` to the `PostHogAutocaptureOptions` interface.
* **`packages/react-native/src/PostHogProvider.tsx`**: Extracted the option from the `autocapture` prop and forwarded it to `useNavigationTracker`.
* **`packages/react-native/src/hooks/useNavigationTracker.ts`**: Updated the navigation tracker hook to compare current route names against normalized screen names in `ignoreScreenNames` (ignoring casing and non-alphanumeric characters) and skip tracking when a match is found.
* **`packages/react-native/test/PostHogProvider.spec.ts`**: Mocked React Navigation (`useNavigation` / `useNavigationState`) to add functional specs checking that ignored routes are successfully skipped and normal routes are still tracked.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context
**Autonomy:** Human-driven (agent-assisted with specs)
* **Agent:** Antigravity coding assistant.
* **Summary:** Antigravity helped write the initial unit tests.


packages/react-native/src/hooks/useNavigationTracker.ts
```diff
@@ -95,6 +96,17 @@ function _useNavigationTracker(
     const currentRouteName = options?.routeToName?.(name, params) || name || 'Unknown'
 
     if (currentRouteName) {
+       const normalizeRegex = /[^a-z0-9]/gi
+       const normalizedRouteName = currentRouteName.toLowerCase()?.replace(normalizeRegex, '')
+       const normalizedScreenNames = ignoreScreenNames.map((screenName) => screenName.toLowerCase()?.replace(
+          normalizeRegex, '')
+       )
+       const skipRouteTracking = normalizedScreenNames?.length && normalizedScreenNames?.includes(normalizedRouteName)
```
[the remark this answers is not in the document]
cacheflow: Where should I place: 

```
const normalizedScreenNames = ignoreScreenNames.map((screenName) => screenName.toLowerCase()?.replace(
          normalizeRegex, '')
       )
  ```

**Comment by marandaneto** [comment]
@cacheflow not a blocker but you could also drop screen events with https://posthog.com/docs/libraries/js/usage#amending-or-sampling-events

File: packages/react-native/src/hooks/useNavigationTracker.ts (hunk 1 above)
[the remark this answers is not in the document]
Comment (cacheflow) [reply packages/react-native/src/hooks/useNavigationTracker.ts]: Why though? If I useMemo at the beginning then I'm precomputing and caching something I may not need. Is this optimization really needed?

File: packages/react-native/test/PostHogProvider.spec.ts [hunk 2]
```diff
@@ -60,4 +68,156 @@ describe('PostHogProvider', () => {
...
+              ignoreScreenNames: ['$HOME'],
+            },
+          },
+          React.createElement(CaptureClient, { onClient })
+        )
+      )
+
+      jest.advanceTimersByTime(1)
+
+      expect(mockPostHog.screen).toHaveBeenCalledWith('home', undefined)
```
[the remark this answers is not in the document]
cacheflow: Updated.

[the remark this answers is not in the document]
cacheflow: How would that look? Isn't duplication better here than one huge it.each block?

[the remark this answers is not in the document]
cacheflow: That's one way to do it or they can add another test case. I'll keep a few in the each block and abstract the null case out to its own test case.

cacheflow:
Thanks for the suggestion @marandaneto.

packages/react-native/src/types.ts
```diff
@@ -48,6 +49,15 @@ export type PostHogAutocaptureOptions = {
    *
    * @default ['style', 'testID', 'accessibilityLabel', 'ph-label', 'children']
    */
+
+  ignoreScreenNames?: string[]
+   /**
+   * List of screen names to ignore during autocapture
+   *
+   * @default []
+   */
```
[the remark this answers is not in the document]
cacheflow: Thanks for the feedback. What should I update this JSDoc to then?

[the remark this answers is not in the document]
cacheflow: Hmmm that seems to go against convention here in the types files. All of the JSDoc blocks appear beneath the definition. 

Image

packages/react-native/src/hooks/useNavigationTracker.ts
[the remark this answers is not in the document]
cacheflow: Done.

turnipdabeets:
Thanks for your PR @cacheflow but I wonder why not use the proposed API suggestion in the issue #3702?
Cc @ioannisj

```
// PostHogOptions
export type PostHogCaptureScreensOptions = {
    allowlist?: (string | RegExp)[]
    ignorelist?: (string | RegExp)[]
  }

captureScreens?: boolean | PostHogCaptureScreensOptions

// Usage
PostHog('<api-key>', {
  captureScreens: {
    allowlist: ['HomeScreen', 'ProductScreen'],   // only capture these
    ignorelist: ['AuthScreen', 'OnboardingScreen'], // or exclude these
  },
})
```

cacheflow:
> Thanks for your PR @cacheflow but I wonder why not use the proposed API suggestion in the issue #3702?
> 
> Cc @ioannisj
> 
> 
> 
> ```
> 
> // PostHogOptions
> 
> export type PostHogCaptureScreensOptions = {
> 
>     allowlist?: (string | RegExp)[]
> 
>     ignorelist?: (string | RegExp)[]
> 
>   }
> 
> 
> 
> captureScreens?: boolean | PostHogCaptureScreensOptions
> 
> 
> 
> // Usage
> 
> PostHog('<api-key>', {
> 
>   captureScreens: {
> 
>     allowlist: ['HomeScreen', 'ProductScreen'],   // only capture these
> 
>     ignorelist: ['AuthScreen', 'OnboardingScreen'], // or exclude these
> 
>   },
> 
> })
> 
> ```

Thanks for the feedback. I considered that approach but I avoided it for a few reasons:

1. The request was related to auto capture and the proposed new API would require deeper changes to the options passed on init than I was comfortable with.

2. They're already passing autocapture options to the Posthog HoC in RN, so it made more sense to pass a new option for ignored screens there.

turnipdabeets:
> Thanks for the feedback. I considered that approach but I avoided it for a few reasons:
> 
> 1. The request was related to auto capture and the proposed new API would require deeper changes to the options passed on init than I was comfortable with.
> 2. They're already passing autocapture options to the Posthog HoC in RN, so it made more sense to pass a new option for ignored screens there.

Agreed that RN's captureScreens lives in the autocapture prop, not init. But we could still add a `screenAllowlist` and `screenIgnorelist`  to the existing autocapture options.

Maybe like this? But the main idea is to have an allowList and an ignoreList (that's what I think we want to follow from the JS SDK)

```
export type PostHogAutocaptureOptions = {
  captureScreens?: boolean
  // ...existing props...

  /**
   * Only capture screen views whose name matches one of these (string = exact, RegExp = pattern).
   * Mutually exclusive with `screenIgnorelist`; if both are set, `screenAllowlist` wins.
   * @default undefined (capture all)
   */
  screenAllowlist?: (string | RegExp)[]

  /**
   * Skip screen views whose name matches one of these (string = exact, RegExp = pattern).
   * @default undefined (ignore none)
   */
  screenIgnorelist?: (string | RegExp)[]
}
Usage:
<PostHogProvider autocapture={{ captureScreens: true, screenIgnorelist: ['About', /^\(tabs\)/, '+not-found'] }}>
```

packages/react-native/src/types.ts
```diff
@@ -48,6 +49,14 @@ export type PostHogAutocaptureOptions = {
    *
    * @default ['style', 'testID', 'accessibilityLabel', 'ph-label', 'children']
    */
+
+  ignoreScreenNames?: string[]
```
ioannisj: The two JSDoc blocks ended up swapped? the "List of prop names to capture from elements" comment is now stuck above `ignoreScreenNames`, and the "List of screen names to ignore during autocapture" comment is sitting above `propsToCapture`.

ioannisj: this could potentially belong inside `PostHogAutocaptureNavigationTrackerOptions` along with other navigation related options like`routeToName` and `routeToProperties`? We could add the allow/ignore list there as well?

ioannisj (commented):
Couple of comments and some lint issues that need to be addressed

packages/react-native/src/types.ts
cacheflow: That was my first thought but I wanted to minimize changes. I'll take another crack at it and see where I end up. Thanks for the feedback.

cacheflow:
I'm going to add the `screenIgnorelist` as suggested by @turnipdabeets.

packages/react-native/src/hooks/useNavigationTracker.ts
```diff
@@ -9,9 +9,12 @@ function _useNavigationTrackerDisabled(): void {
   return
 }
 
+const normalizeRegex = /[^a-z0-9]/gi
+
 function _useNavigationTracker(
   options?: PostHogAutocaptureNavigationTrackerOptions,
   navigationRef?: PostHogNavigationRef,
+  ignoreScreenNames: string[] = [],
```
turnipdabeets: `useNavigationTracker` is re-exported from the package index, so this signature is public. Slotting `ignoreScreenNames` before `client` means an existing `useNavigationTracker(opts, ref, client)` call now passes the client where the array is expected, and `ignoreScreenNames.map(...)` throws. Could it ride on the options bag instead — `PostHogAutocaptureNavigationTrackerOptions.ignoreScreenNames` — so the positional order stays stable?

packages/react-native/src/PostHogProvider.tsx
```diff
@@ -123,7 +123,7 @@ export const PostHogProvider = ({
   style,
   debug = false,
 }: PostHogProviderProps): JSX.Element | null => {
-  const captureAll = autocapture === true
+  const captureAll = autocapture === true;
```
turnipdabeets: Small thing — stray semicolon in a no-semicolon file; this is what's failing the Lint packages check.
```suggestion
  const captureAll = autocapture === true
```

packages/react-native/src/types.ts
```diff
@@ -48,6 +48,14 @@ export type PostHogAutocaptureOptions = {
...
```
turnipdabeets: Heads up — the JSDoc block above now attaches to `ignoreScreenNames` (it's the `propsToCapture` default list), and the new `List of screen names…` block below ends up documenting `propsToCapture`. Both comments landed on the wrong field. Moving the new doc block directly above this line fixes it (the closing `*/` is also indented one space short).

packages/react-native/test/PostHogProvider.spec.ts
```diff
@@ -60,4 +68,75 @@ describe('PostHogProvider', () => {
...
+      mockUseNavigationState.mockReset()
+      mockUseNavigation.mockReset()
+    })
+
+    afterEach(() => {
+      jest.useRealTimers()
+    })
+
+    it('ignores screen in ignoreScreenNames', () => {
+      const mockRoute = { name: routeName, params: {} }
```
turnipdabeets: `routeName` and `ignoreScreenNames` aren't defined in this test — they only exist in the `it.each` below — so this won't compile under ts-jest. And an ignored screen means `screen` shouldn't be called at all, so line 98 likely wants `expect(mockPostHog.screen).not.toHaveBeenCalled()` rather than `toHaveBeenCalledWith(null)`. Was this meant to be dropped, or wired up like the `it.each` cases?

packages/react-native/src/frameworks/wix-navigation.ts
```diff
@@ -9,6 +9,8 @@ import { OptionalReactNativeNavigationWix } from '../optional/OptionalReactNativ
 import { PostHog } from '../posthog-rn'
 import { PostHogAutocaptureOptions } from '../types'
 
+const normalizeRegex = /[^a-z0-9]/gi
```
turnipdabeets: nit: this normalize regex + skip logic is duplicated in `useNavigationTracker.ts`. Might be worth a small shared `shouldIgnoreScreen(name, ignoreScreenNames)` helper so the matching rule lives in one place.

packages/react-native/src/hooks/useNavigationTracker.ts
cacheflow: That's a great idea. I'll do that.

cacheflow:
@turnipdabeets I made a few changes including removing the PosthogProvider spec in favor of testing calling `posthog.screen` in isolation using the useNavigationTracker hook. I also fixed the JSDoc def. Let me know if that's what you had in mind. 😊

packages/react-native/src/PostHogProvider.tsx
```diff
@@ -50,7 +50,7 @@ function PostHogNavigationHook({
   options?: PostHogAutocaptureOptions
   client?: PostHog
 }): JSX.Element | null {
-  useNavigationTracker(options?.navigation, options?.navigationRef, client)
+  useNavigationTracker(options?.navigation, options?.navigationRef, options?.ignoreScreenNames, client)
```
turnipdabeets: This is the `TS2554` turning CI red — `useNavigationTracker` takes `(options, navigationRef, client)`, so `options?.ignoreScreenNames` lands in the `client` slot and the real client gets dropped.

The root issue is that the option now lives in two places and each path reads a different one: this hook reads the nested `navigation.ignoreScreenNames`, wix reads the top-level one, and this call site tries to bridge the two. I'd finish the move to the nested placement (what we discussed on the earlier threads) and make it the only place:

```suggestion
  useNavigationTracker(options?.navigation, options?.navigationRef, client)
```

plus: drop `ignoreScreenNames` from `PostHogAutocaptureOptions`, have wix read `options?.navigation?.ignoreScreenNames`, and update the example/tests to `autocapture={{ navigation: { ignoreScreenNames: [...] } }}`. That keeps one source of truth for the provider, direct hook callers, and wix — nothing to keep in sync.

packages/react-native/src/hooks/useNavigationTracker.ts
```diff
@@ -95,6 +96,16 @@ function _useNavigationTracker(
     const currentRouteName = options?.routeToName?.(name, params) || name || 'Unknown'
 
     if (currentRouteName) {
+      const normalizedRouteName = currentRouteName.toLowerCase()?.replace(normalizeRegex, '')
+      const normalizedScreenNames = options?.ignoreScreenNames.map((screenName) =>
```
turnipdabeets: This is the `TS18048` in CI — and at runtime, anyone who already passes `navigation: { routeToName }` without the new option throws here on every navigation.

```suggestion
      const normalizedScreenNames = (options?.ignoreScreenNames ?? []).map((screenName) =>
```

The new tests don't catch it because every case passes `ignoreScreenNames` explicitly (even `[]`), which no existing caller does — worth adding one case with options set but no `ignoreScreenNames`, and one with no options at all.

packages/react-native/src/utils.ts
```diff
@@ -43,3 +43,5 @@ export function isWindows(): boolean {
 }
 
 export const isHermes = () => !!GLOBAL_OBJ.HermesInternal
+
+export const normalizeRegex = /[^a-z0-9]/gi
```
turnipdabeets: The normalize-then-compare approach has some sharp edges: any entry that strips to an empty string acts as a wildcard — `ignoreScreenNames: ['設定']` (or `['-']`) normalizes to `''`, which then matches every screen whose name also strips to `''`, i.e. every non-Latin or symbol-only screen name in the app. Distinct names also collide (`user2` and `user_2` both become `user2`), which "list of screen names to ignore" doesn't imply.

I'd drop the regex and match on a case-insensitive exact compare instead:

```ts
const skip = ignoreScreenNames.some((name) => name.toLowerCase() === currentRouteName.toLowerCase())
```

`toLowerCase()` is Unicode-safe, so this works for every language, has no collisions or empty-string edge case, and it collapses the duplicated normalize logic in both call sites down to one trivial line. Route names come from the navigator config, so case-insensitivity should be all the forgiveness needed.

packages/react-native/src/frameworks/wix-navigation.ts
```diff
@@ -27,6 +28,20 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
 
     if (currentRouteName) {
       const properties = options?.navigation?.routeToProperties?.(currentRouteName, passProps || {})
```
turnipdabeets: nit: `routeToProperties` runs before the ignore check here, so a user's callback still executes for ignored screens — the hook does it in the other order. Moving the skip check above this line keeps them consistent (and skips potentially expensive user code).

examples/example-expo-53/app/_layout.tsx
```diff
@@ -27,6 +27,7 @@ export default function RootLayout() {
             autocapture={{
                 captureScreens: false, // expo-router requires this to be false and capture screens manually
                 captureTouches: true,
+                ignoreScreenNames: ['about'],
```
turnipdabeets: expo-router forces manual capture, no current RN example can exercise `ignoreScreenNames` (the other two don't use `PostHogProvider` at all). Worth also noting in the option's JSDoc that it applies to autocaptured screens only — manual `posthog.screen()` calls aren't filtered. Cc @ioannisj

packages/react-native/test/useNavigationTracker.spec.tsx
```diff
@@ -0,0 +1,186 @@
...
+
+  afterEach(() => {
+    jest.useRealTimers()
+    cleanup()
+  })
+
+  const wrapper = ({ children }: { children: React.ReactNode }) =>
+    React.createElement(PostHogContext.Provider, { value: { client: mockPostHog } }, children)
+
+  it('ignores screens in ignoreScreenNames', () => {
```
turnipdabeets: These tests inject the same `mockPostHog` twice — as the client arg and via the context wrapper — so they can't tell whether the client arg is actually plumbed through. More importantly, none of them go through `PostHogProvider`, which is where the wiring actually broke. A test that renders `PostHogProvider` with `autocapture={{ ignoreScreenNames: [...] }}` would have caught both CI failures and pins down which option placement is canonical.

turnipdabeets:
A couple other notes before we can merge: 

We'll also need a changeset - see RELEASING.md
Your commits need to be signed

packages/react-native/src/types.ts
cacheflow: Just for my understanding, we want this new definition placed where? Currently it's under `ignoreLabels` and above `propsToCapture`. I'd like to minimize the changes on this specific part if we can. 

Screenshot 2026-07-06 at 12 52 01 PM

packages/react-native/src/utils.ts
cacheflow: This is a very good point. I didn't think of i18n here. thanks!

examples/example-expo-53/app/_layout.tsx
cacheflow: Thanks. It looks like I typed this correctly and mentioned that it's only available during auto capture. Does that look right below? 

Screenshot 2026-07-06 at 1 04 25 PM

Screenshot 2026-07-06 at 1 04 12 PM

packages/react-native/src/frameworks/wix-navigation.ts
cacheflow: Thanks. Also, do we still need this in wix-navigation or can we keep it in `useNavigationTracker`?

packages/react-native/test/useNavigationTracker.spec.tsx
cacheflow: I'll work on that one next.

packages/react-native/src/frameworks/wix-navigation.ts
turnipdabeets: We need it in both, since wix apps don't go through useNavigationTracker. The duplication itself is fine to fix though: pull the comparison into a tiny shared helper in utils.ts and call it from both places, so the matching rule only lives in one spot.

packages/react-native/src/PostHogProvider.tsx
cacheflow: I see. So place `ignoreScreenNames` under the navigation property instead of directly under autocapture? Would we want that for every place that uses `PostHogAutocaptureOptions`?

packages/react-native/test/useNavigationTracker.spec.tsx
cacheflow: Are there any specs in the codebase for testing the PostHogProvider with autocapture I can use as reference?

packages/react-native/src/types.ts
```diff
@@ -1,6 +1,7 @@
 export type PostHogAutocaptureNavigationTrackerOptions = {
   routeToName?: (name: string, params: any) => string
   routeToProperties?: (name: string, params: any) => Record<string, any>
+  ignoreScreenNames?: string[]
```
turnipdabeets, line 10: The doc block got dropped in the move — the other fields in this type are undocumented too, but since this one is new and its two gotchas (case-insensitive matching, autocapture-only) aren't guessable from the type, let's keep its JSDoc:

```suggestion
  /**
   * List of screen names to ignore during autocapture. Matching is case-insensitive.
   * Only applies to autocaptured screens — manual `posthog.screen()` calls are not filtered.
   *
   * @default []
   */
  ignoreScreenNames?: string[]
```

packages/react-native/test/useNavigationTracker.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+  it('customizes route name using routeToName', () => {
+    const mockRoute = { name: 'user', params: { id: '123' } }
+    mockUseNavigationState.mockImplementation((cb) => cb({ routes: [mockRoute] }))
+    mockUseNavigation.mockReturnValue({ isReady: () => true, getCurrentRoute: () => mockRoute })
+
+    renderHook(
+      () =>
+        useNavigationTracker(
+          {
+            ignoreScreenNames: [],
```
turnipdabeets: Still missing the regression tests from the earlier thread: every case in this file passes `ignoreScreenNames` explicitly (even `[]`), which no pre-existing caller does. Two cases would pin the `?? []` fix so the crash can't come back:

1. this test with the `ignoreScreenNames: []` line simply removed — options set, field absent, the exact config that used to throw on every navigation
2. one calling `useNavigationTracker(undefined, undefined, mockPostHog)` and asserting `screen` still fires

packages/react-native/src/frameworks/wix-navigation.ts
```diff
@@ -26,7 +26,16 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
...
     if (currentRouteName) {
+      const ignoreScreenNames = options?.navigation?.ignoreScreenNames ?? []
       const properties = options?.navigation?.routeToProperties?.(currentRouteName, passProps || {})
+      const skipScreenCapture = ignoreScreenNames.some(
+        (name: string) => name?.toLowerCase() === currentRouteName?.toLowerCase()
+      )
+
+      if (skipScreenCapture) {
+        return
+      }
```
turnipdabeets: Following up on the ordering nit with the concrete change — decide the skip first, then run the user's `routeToProperties` callback only for screens we're actually capturing (matches the hook's order):

```suggestion
      const skipScreenCapture = ignoreScreenNames.some(
        (name: string) => name?.toLowerCase() === currentRouteName?.toLowerCase()
      )

      if (skipScreenCapture) {
        return
      }

      const properties = options?.navigation?.routeToProperties?.(currentRouteName, passProps || {})
```

turnipdabeets (commented):
This is looking really close. New inline items with suggestions below, plus need to run `pnpm generate-references` and commit the updated JSON. I'd recommend merging main into this since it's a bit behind and signing your other commits.

examples/example-expo-53/app/_layout.tsx
```diff
@@ -28,6 +28,9 @@ export default function RootLayout() {
                 captureScreens: false, // expo-router requires this to be false and capture screens manually
                 captureTouches: true,
                 customLabelProp: 'ph-my-label',
+                navigation: {
+                    ignoreScreenNames: ['about'],
```
turnipdabeets: nit: trailing whitespace here.
```suggestion
                    ignoreScreenNames: ['about'],
```

packages/react-native/src/hooks/useNavigationTracker.ts
```diff
@@ -95,6 +95,14 @@ function _useNavigationTracker(
     const currentRouteName = options?.routeToName?.(name, params) || name || 'Unknown'
 
     if (currentRouteName) {
+      const normalizedRouteName = currentRouteName.toLowerCase()
+      const normalizedScreenNames = (options?.ignoreScreenNames ?? []).map((screenName) => screenName?.toLowerCase())
```
turnipdabeets: The same ignore-name matching lives in `wix-navigation.ts` too (`ignoreScreenNames.some((name) => name?.toLowerCase() === currentRouteName?.toLowerCase())`). Worth pulling into a small shared helper like `shouldIgnoreScreen(routeName, ignoreScreenNames)` so the two implementations can't drift?

.changeset/dry-planes-teach.md
```diff
@@ -0,0 +1,5 @@
+    ---
```
turnipdabeets: The frontmatter opens with leading spaces, so changesets can't parse it — `posthog-react-native` reads as undeclared and won't get a version bump on release (this is what the changeset hygiene check is flagging). Dropping the indentation fixes it:
```suggestion
---
```

packages/react-native/src/frameworks/wix-navigation.ts
```diff
@@ -26,7 +27,14 @@ export const withReactNativeNavigation = (posthog: PostHog, options: PostHogAuto
       options?.navigation?.routeToName?.(componentName, passProps || {}) || componentName || 'Unknown'
 
     if (currentRouteName) {
+      const skipScreenCapture = shouldIgnoreScreen(currentRouteName, options?.navigation?.ignoreScreenNames)
+
```
turnipdabeets: nit: trailing whitespace on this blank line.
```suggestion

```

packages/react-native/src/autocapture/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const shouldIgnoreScreen = (currentRouteName: string, ignoreScreenNames?: string[]): boolean => {
+  const normalizedRouteName = (currentRouteName || '').toLowerCase()
+  const normalizedScreenNames = (ignoreScreenNames ?? []).map((screenName) => screenName?.toLowerCase()) || []
+
+  const screenMatch = normalizedScreenNames.some((name: string) => name?.toLowerCase() === normalizedRouteName)
```
turnipdabeets: Small thing — `normalizedScreenNames` is already lowercased by the `.map`, so the `.toLowerCase()` inside `.some` is redundant, and `.map(...)` always returns an array so `|| []` never fires. Could trim to:
```suggestion
  const normalizedScreenNames = (ignoreScreenNames ?? []).map((screenName) => screenName?.toLowerCase())

  const screenMatch = normalizedScreenNames.some((name) => name === normalizedRouteName)
```

turnipdabeets:
@cacheflow looking really good, minor nits left! 

I just want to get @ioannisj opinion on public API adding to `navigation.ignoreScreenNames` vs originally proposed  `captureScreens: boolean | { allowlist, ignorelist }` which is closer to web.

.changeset/dry-planes-teach.md
cacheflow: Odd because I ran prettier on these files. Let me try manually fixing them.

cacheflow:
Hey @turnipdabeets thanks again for your patience and working with me through this PR.

examples/example-expo-53/app/_layout.tsx
```diff
@@ -28,6 +28,9 @@ export default function RootLayout() {
                 captureScreens: false, // expo-router requires this to be false and capture screens manually
                 captureTouches: true,
                 customLabelProp: 'ph-my-label',
+                navigation: {
+                    ignoreScreenNames: ['about'], 
+                }
```
ioannisj: nit: config being showcased here is a no-op since captureScreens:false?

packages/react-native/references/posthog-react-native-references-4.46.30.json
```diff
@@ -2449,7 +2449,7 @@
       "name": "PostHogAutocaptureNavigationTrackerOptions",
       "properties": [],
       "path": "dist/types.d.ts",
-      "example": "{\n    routeToName?: (name: string, params: any) => string;\n    routeToProperties?: (name: string, params: any) => Record<string, any>;\n}"
+      "example": "{\n    routeToName?: (name: string, params: any) => string;\n    routeToProperties?: (name: string, params: any) => Record<string, any>;\n    ignoreScreenNames?: string[]\n}"
```
ioannisj: I think we should not be touching this file since it's chronological?

packages/react-native/src/autocapture/utils.ts
```diff
@@ -0,0 +1,8 @@
+export const shouldIgnoreScreen = (currentRouteName: string, ignoreScreenNames?: string[]): boolean => {
+  const normalizedRouteName = (currentRouteName || '').toLowerCase()
+  const normalizedScreenNames = (ignoreScreenNames ?? []).map((screenName) => screenName?.toLowerCase()) || []
```
ioannisj: nit: probably || [] not needed?


```suggestion
  const normalizedScreenNames = (ignoreScreenNames ?? []).map((screenName) => screenName?.toLowerCase())
```

packages/react-native/src/types.ts
```diff
@@ -1,6 +1,13 @@
 export type PostHogAutocaptureNavigationTrackerOptions = {
   routeToName?: (name: string, params: any) => string
   routeToProperties?: (name: string, params: any) => Record<string, any>
+  /**
+   * List of screen names to ignore during autocapture. Matching is case-insensitive.
+   * Only applies to autocaptured screens — manual `posthog.screen()` calls are not filtered.
+   *
```
ioannisj, line 7: Let's also document that this runs after `routeToName`

ioannisj (commented):
> @cacheflow looking really good, minor nits left!
> 
> I just want to get @ioannisj opinion on public API adding to `navigation.ignoreScreenNames` vs originally proposed `captureScreens: boolean | { allowlist, ignorelist }` which is closer to web.

Yeah I think this should be okay. Adding `allowScreenNames` later is additive and widening with RegExp should be non-breaking I think . However, I also think a simple hook `shouldCaptureScreen?: (screenName: string) => boolean` would have been far more flexible and simpler than both options. People could implement any logic they want (allow list, ignore list, contains, regex etc) and we wouldn't have to worry about normalizing for case-sensitivity either for example.

examples/example-expo-53/app/_layout.tsx
turnipdabeets: Might be good to add a comment to change the example app to true to test that?

packages/react-native/src/types.ts
cacheflow: How so? From the looks of it, it seems that [routeToName](https://github.com/cacheflow/posthog-js/blob/add-screen-ignoring-during-autocapture/packages/react-native/src/hooks/useNavigationTracker.ts#L96-L108) is used in the hooks to determine if it's an ignored screen.

packages/react-native/references/posthog-react-native-references-4.46.30.json
cacheflow: I don't think I changed this but I can remove it.

turnipdabeets:
Hey @cacheflow — first off, thank you so much for this. 🙏 You put real care into it: the shared shouldIgnoreScreen helper, coverage across both the React Navigation and Wix paths, the regression tests, and all the back-and-forth iterations. It was a pleasure to review.

After talking it through as a team, we've decided not to add a dedicated screen allow/ignore option to the public API, but I want to explain the reasoning, because it's not a reflection on your work at all.

The short version: PostHog's SDKs already ship a `before_send` hook, and it turns out to be a clean fit for exactly this. It runs on every event (including autocaptured $screen events), it composes (you can pass an array of functions), and returning null drops an event before it's sent. So you can filter screens however you like (an ignorelist, an allowlist, a prefix match, a regex, or logic based on the event's properties) without us adding new config surface to maintain.

The reason the web SDK has a dedicated `url_allowlist`/`url_ignorelist` (which is what the original issue was reaching for) is mostly historical: those landed back in 2022, before `before_send` existed. If we were adding that capability today, we'd likely just point people at `before_send`, so we'd rather not grow the mobile API in a direction web is itself moving away from. This old API may even be removed in web v2.

We think the real issue identified is discoverability, so we've fixed that with docs:

- Filtering autocaptured screens: https://posthog.com/docs/libraries/react-native#filtering-autocaptured-screens
- Common before_send patterns: https://posthog.com/docs/libraries/react-native#common-before_send-patterns

Genuinely grateful for the contribution and the time you spent. Please don't let this discourage you from opening more. 💙

cacheflow:
Thanks for the thoughtful explanation. I understand the reasoning around using `before_send`, but I disagree that it's the best long-term solution here.

The original intention of `before_send` was ["to edit or redact information before it is sent to PostHog"](https://github.com/PostHog/posthog.com/pull/9844/changes#diff-e9e9ef37e4e756e5cf3bc804e89db87a8e60864b4a39d43c61c311cd0ae13eb6R467), whereas the original request was for an  [intuitive API](https://github.com/PostHog/posthog-js/issues/3702)](https://github.com/PostHog/posthog-js/issues/3702) to configure screen autocapture behavior.

Using `before_send` to handle autocapture related events certainly works in the short-term, but it broadens the responsibility of the callback overall. Whereas a designated API would give us a clear separation of concerns and prevent `before_send` from becoming overburdened.

I also want to be candid about the review process. The PR went through several rounds of implementation feedback with dozens of review comments, changes across both navigation implementations, a shared helper to prevent divergence, regression tests (including language scenarios that weren't previously covered), documentation updates, and eventually reached the point where it was described as "looking really good" with only minor nits remaining.

I assumed the remaining work was fixing the remaining small issues and not revisiting whether the feature should exist at all.

So  it was disappointing to learn this late in the process that the team had decided against adding a dedicated API entirely, especially given the community's demand for it.

I hope I don't come across as angry because I appreciate everyone who took the time to review the PR and explain the final decision. Even though I still disagree with the outcome, I'm grateful for the thoughtful discussion and the feedback throughout the review.

cacheflow:
Hey @turnipdabeets, I'd still like to contribute to Posthog. Is there another issue you recommend tackling?

marandaneto:
> Hey @turnipdabeets, I'd still like to contribute to Posthog. Is there another issue you recommend tackling?

hey @cacheflow @turnipdabeets is on PTO today but i'll make sure this is seen and follow up, i agree with the feedback above, and sorry about that.

ioannisj:
Hey @cacheflow, I want to own how this went. You were told this was "looking really good, minor nits left," only for the PR to be closed a few days later. That decision should have been made on #3702 or internally before you got several rounds of review, each pulling in a different direction. That's on us, and your pushback here is completely fair.

To make matters worse, we somehow missed and left your feedback above unanswered. I'm sorry about that too.

I'd like to send you a PostHog merch code as a small thank-you for the time you put into this, and as an apology for the way we handled things. Could you drop me an email at yiannis @ posthog.com and I'll sort it out?

As for what to pick up next, I'd steer you toward the React Native survey gaps. You already know the package, and the behavior is defined on the browser sdk, so hopefully there's less API debate this time around:

- survey popup delay #3193
- survey wait period #3192

Again, apologies for how this was handled. We'll do better.

PR #4003, opened 2026-06-29
fix(browser): avoid Promise-dependent IE startup failures

marandaneto:
## Problem

The IE11 TestCafe job was failing before tests start when BrowserStack closes compressed API responses early. The TestCafe BrowserStack provider calls BrowserStack's Automate REST API through `node-fetch` without retries, so `Premature close` from the gzip stream fails the whole PR check before SDK tests execute.

BrowserStack's current Automate API docs still document `GET /automate/browsers.json`, and BrowserStack status reports all systems operational. The follow-up failure moved from `/automate/browsers.json` to `/automate/sessions/<id>.json`, which points to a BrowserStack API response compression/transport issue across Automate API calls rather than a documented endpoint change.

After BrowserStack startup was unblocked, the IE11 run exposed a real SDK issue: the ES5 bundle can execute Promise-dependent compiled helpers in a browser where `Promise` is unavailable.

## Changes

- Wrap the IE11 TestCafe command in a targeted three-attempt retry loop.
- Preload a narrow `node-fetch` patch for the TestCafe run that disables compression and sends `Accept-Encoding: identity` for BrowserStack API hosts only.
- Retry only when the TestCafe output references BrowserStack's API host, so real test failures still fail immediately.
- Add a Promise polyfill to the IE11/ES5 bundle.
- Avoid the Promise-dependent async compression path when `Promise` is unavailable, falling back to the existing synchronous encoding path.
- Guard the Segment integration Promise check with `typeof Promise`.
- Restored the TestCafe `is-affected` gate after validating the workflow path.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi investigated the IE11 workflow failure and found BrowserStack startup failures in `node-fetch` gunzip handling after BrowserStack closed compressed responses early. BrowserStack Automate API docs still document the browser-list endpoint. After avoiding BrowserStack API compression, the workflow reached IE11 and exposed a separate SDK `Promise` compatibility issue in the ES5 bundle, now covered by an IE11-only Promise polyfill and guarded async compression path. The TestCafe workflow passed once with the affected check bypassed, then the affected check was restored.

### DIFF
--- a/packages/browser/src/extensions/segment-integration.ts
+++ b/packages/browser/src/extensions/segment-integration.ts
@@ -31,7 +31,7 @@ export type { SegmentUser, SegmentAnalytics, SegmentContext, SegmentPlugin }
 const logger = createLogger('[SegmentIntegration]')
 
 const createSegmentIntegration = (posthog: PostHog): SegmentPlugin => {
-    if (!Promise || !Promise.resolve) {
+    if (typeof Promise === 'undefined' || !Promise.resolve) {
         logger.warn('This browser does not have Promise support, and can not use the segment integration')
     }
 
--- a/.changeset/fuzzy-ravens-promise.md
+++ b/.changeset/fuzzy-ravens-promise.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Avoid Promise-dependent async compression paths in browsers without Promise support.
--- a/.changeset/fuzzy-ravens-promise.md
+++ b/.changeset/fuzzy-ravens-promise.md
@@ -2,4 +2,4 @@
 'posthog-js': patch
 ---
 
-Avoid Promise-dependent async compression paths in browsers without Promise support.
+Include a Promise polyfill in the IE11 bundle and avoid Promise-dependent async compression paths when Promise support is unavailable.
--- a/.github/workflows/testcafe.yml
+++ b/.github/workflows/testcafe.yml
@@ -44,26 +44,22 @@ jobs:
         with:
           build: false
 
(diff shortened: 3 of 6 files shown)


packages/browser/src/extensions/segment-integration.ts
```diff
@@ -31,7 +31,7 @@ export type { SegmentUser, SegmentAnalytics, SegmentContext, SegmentPlugin }
 const logger = createLogger('[SegmentIntegration]')
 
 const createSegmentIntegration = (posthog: PostHog): SegmentPlugin => {
-    if (!Promise || !Promise.resolve) {
+    if (typeof Promise === 'undefined' || !Promise.resolve) {
```
ioannisj, line 34: Did not look deep into the implementation and follow the code oath but shouldn't this return a no-op SegmentPlugin instead of warn and continue?

ioannisj (approved):
Approve to unblock, left a comment

packages/browser/src/extensions/segment-integration.ts
marandaneto: i agree but not related to this PR, the goal here is to fix the flakiness for IE/testcafe
this issue already exists so not addressing rn

PR #4022, opened 2026-06-30
feat(mcp): emit $mcp_lib / $mcp_lib_version on MCP events

gesh:
## Summary

Stamp the `@posthog/mcp` SDK name + version onto every `$mcp_*` event as `$mcp_lib` (`@posthog/mcp`) and `$mcp_lib_version` (e.g. `0.5.1`), so you can tell which analytics SDK release produced the data.

The version was already resolved at runtime (`Event.sdkVersion`) but never mapped to a property; the name wasn't stored at all. This wires it the last mile.

### Why a namespaced `$mcp_lib` rather than reusing `$lib`
- posthog-node forcibly stamps `$lib` / `$lib_version` *after* event properties (`posthog-core-stateless.ts:382-385`), so the property bag can't override it — `$lib` correctly stays the transport SDK (`posthog-node`).
- Mirrors the sibling `@posthog/ai`, which keeps `$lib` and adds `$ai_lib` / `$ai_lib_version`.
- Avoids colliding with "MCP SDK" = `@modelcontextprotocol/sdk` (deliberately not tracked here).

## Changes
- `constants.ts` — `POSTHOG_MCP_LIB_NAME` + `McpLib` / `McpLibVersion` keys
- `posthog-events.ts` — stamp both on the regular and `$exception` events
- `posthog-mcp.ts` — populate `sdkVersion` on the server-agnostic `PostHogMCP` path
- tests in `posthog-events.test.ts` / `posthog-mcp.test.ts`

## Testing
- `pnpm test:unit` — 30 suites / 334 tests pass
- `pnpm lint` — clean
- Verified live in a `dummy-nest-mcp` server: `$mcp_tool_call` now carries `$mcp_lib: "@posthog/mcp"` + `$mcp_lib_version`

---

### DIFF
--- a/.changeset/mcp-lib-version-property.md
+++ b/.changeset/mcp-lib-version-property.md
@@ -1,0 +1,5 @@
+---
+'@posthog/mcp': minor
+---
+
+Emit `$mcp_lib` (`@posthog/mcp`) and `$mcp_lib_version` on every `$mcp_*` event (and the `$exception` sibling) so you can tell which analytics SDK release produced the data. The version was already resolved at runtime but never mapped to a property. Namespaced like `@posthog/ai`'s `$ai_lib` rather than overriding `$lib`, which stays the transport SDK (`posthog-node`).
--- a/packages/mcp/src/__tests__/posthog-events.test.ts
+++ b/packages/mcp/src/__tests__/posthog-events.test.ts
@@ -6,6 +6,7 @@ import {
 import { MCPAnalyticsEventType } from '../extensions/event-types'
 import { buildPostHogCaptureEvents, type PostHogCaptureEvent } from '../extensions/posthog-events'
 import type { Event } from '../types'
+import { version } from '../version'
 
 function makeEvent(overrides: Partial<Event> = {}): Event {
   return {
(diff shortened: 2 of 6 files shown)


pauldambra (approved):
nice

marandaneto:
if you want to get some analytics info (usage of this sdk, breakdown per version) etc you'd need to add that to the usage report, see example here https://github.com/PostHog/posthog/pull/65130/changes#diff-87e14b80cf403a45c1161ddcaed987b1addafdaeb20500d6979321495f74db10

> posthog-node forcibly stamps $lib / $lib_version after event properties (posthog-core-stateless.ts:382-385), so the property bag can't override it — $lib correctly stays the transport SDK (posthog-node).

maybe the correct approach here would be to actually set as $lib and $lib_version because those are materialized already, since this sdk is its own package/has a lot of features etc and its not simply an integration on top of the node sdk, but rather extends the node sdk functionality

we did the same with `$ai_lib`, the clickhouse team didnt want to materialize those and now we are introducing yet more non-materialized properties and most likely we'd like to query those as well

not a blocker, just food for thought

PR #4025, opened 2026-06-30
feat(mcp): report standard $lib=posthog-node-mcp (drop $mcp_lib)

gesh:
## What

MCP events now use the standard PostHog `$lib` / `$lib_version` schema (value **`posthog-node-mcp`**) instead of the namespaced `$mcp_lib` / `$mcp_lib_version` added in #4022.

## Why

`$lib` is the canonical, materialized "which SDK" dimension. posthog-node forcibly stamps `$lib`/`$lib_version` from `getLibraryId()`/`getLibraryVersion()` (spread last in the payload, after `before_send`), so they can't be overridden per-event via properties — overriding those two methods is the only lever.

## How

A shared `applyMcpLibIdentity` helper sets the identity on both emit paths:
- **`PostHogMCP`** — in its constructor.
- **`instrument(server, client)`** — on the host-supplied client (relabels every event that client sends; pass a client dedicated to MCP analytics).

Removed `addMcpLibProperties` and the `$mcp_lib` keys; replaced the changeset.

## Test

- 336/336 mcp unit tests pass; added lib-identity assertions for both paths.
- Validated end-to-end against a NestJS/`@rekog/mcp-nest` playground: `instrument()` and `PostHogMCP` both report `$lib=posthog-node-mcp` / `$lib_version=0.5.1`.

---

### DIFF
--- a/packages/mcp/src/__tests__/lib-identity.test.ts
+++ b/packages/mcp/src/__tests__/lib-identity.test.ts
@@ -1,5 +1,5 @@
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
-import { instrument, PostHog } from '../index'
+import { instrument, PostHog, PostHogMCP } from '../index'
 import { version } from '../version'
 
 /**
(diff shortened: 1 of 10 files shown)


packages/mcp/src/__tests__/lib-identity.test.ts
```diff
@@ -0,0 +1,26 @@
...
+
+      const server = new McpServer({ name: 'my-mcp', version: '1.0.0' }, { capabilities: { tools: {} } })
+      instrument(server, posthog)
+
+      expect(posthog.getLibraryId()).toBe('posthog-node-mcp')
+      expect(posthog.getLibraryVersion()).toBe(version)
+    } finally {
+      await posthog.shutdown()
+    }
+  })
```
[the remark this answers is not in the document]
gesh: Done in 62efa114 — consolidated both emit paths (PostHogMCP + instrument()) into a single `it.each` in lib-identity.test.ts and dropped the duplicate from posthog-mcp.test.ts.

marandaneto:
sdk name is fine by me
now you need to add this sdk name to usage report and most likely also for feat flags allow list (only if the node mcp sdk also uses feat flags), example https://github.com/PostHog/posthog/pull/62594

packages/mcp/src/extensions/constants.ts
```diff
@@ -11,10 +11,11 @@ export const DEFAULT_CONVERSATION_ID_DESCRIPTION =
...
 
-// Identifies the SDK that produced an `$mcp_*` event. Distinct from `$lib`
-// (posthog-node, the transport that actually sends the event) — this names the
-// higher-level analytics layer, mirroring `@posthog/ai`'s `$ai_lib`.
-export const POSTHOG_MCP_LIB_NAME = '@posthog/mcp'
+// The `$lib` identity stamped on every event @posthog/mcp sends. posthog-node
+// would otherwise report itself (`posthog-node`, the transport SDK); we override
+// `getLibraryId()` so MCP events self-identify the same way every other PostHog
+// SDK does. See `applyMcpLibIdentity` in `./lib-identity`.
+export const POSTHOG_MCP_LIB_NAME = 'posthog-node-mcp'
```
pauldambra, line 18: shall we echo the wrapped sdk names?  
  
so more like postog/node/mcp  
to sit alongside  posthog/node  
  
(we are slightly over designing this name but it is hard to change once out :)))

pauldambra, line 18:
```suggestion
export const POSTHOG_MCP_LIB_NAME = '@posthog/node/mcp'
```

gesh: Currently, we capture them as `posthog-node` / `posthog-python`. 
This one with the slashes was added by me today, which does not follow the convention. So we might want to stick with the dashes?

pauldambra: ah, so posthog-node-mcp in that case  
  
that way we can do startsWith posthog-node  
or equals posthog-node

gesh:
> sdk name is fine by me now you need to add this sdk name to usage report and most likely also for feat flags allow list (only if the node mcp sdk also uses feat flags), example [PostHog/posthog#62594](https://github.com/PostHog/posthog/pull/62594)

Thanks a lot for referencing these!! 🙌 

I think the MCP SDK is not gated by a feature flag now. Will check what's the case with both feature flags and usage report and will open a follow-up PR.

pauldambra (approved):

PR #4032, opened 2026-07-01
feat(mcp): add $mcp_error_type / $mcp_error_message to tool-call events

lucasheriques:
## Problem

On `$mcp_tool_call` the only failure signal is the `$mcp_is_error` boolean. To break failures down by reason on the MCP analytics dashboard you have to join to the separate `$exception` event — which can be disabled (`enableExceptionAutocapture: false`), and isn't emitted at all when a call is flagged as an error without a thrown value. So "which tools fail, and *why*" isn't answerable from the primary event.

## Changes

Stamp two structured properties directly onto `$mcp_tool_call` (and `$mcp_tools_list`) when a call fails:

- **`$mcp_error_type`** — a low-cardinality failure category. Defaults to the thrown error's type (the `$exception_list` entry's `type`, e.g. `TimeoutError`, `ZodError`). `captureToolCall` / `captureToolsList` also accept an explicit `errorType` label (e.g. `validation`, `permission`, `timeout`, `rate_limited`) so hosts that classify their own failures can override the generic type.
- **`$mcp_error_message`** — the error message, already bounded to `MAX_ERROR_MESSAGE_LENGTH` by the existing truncation pass.

Every server using `instrument()` gets a failure-reason breakdown with **no code change** — the type is derived automatically from the error the SDK already captures. The explicit `errorType` field is additive and optional.

### Libraries affected

- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with PostHog Code (Claude). The work started from a look at the dogfood MCP analytics data, which surfaced a large volume of errored tool calls whose cause wasn't visible on the primary event. This is the SDK half of a two-part change; the companion PostHog-server change stamps its own richer taxonomy via the explicit `errorType` field.

Design decisions:
- Derive `$mcp_error_type` generically from the normalized `$exception_list` so it works for every `instrument()` user without host changes, and let an explicit `errorType` override it for hosts with a domain taxonomy.
- Reuse the existing truncation of the exception message rather than re-capping in the event builder; the explicit `errorType` label is bounded to the metadata length limit.
- Kept `$mcp_error_message` on the primary event (same content already accepted on `$exception`); hosts can drop it via `beforeSend` if desired.

Tests: `pnpm test:unit` (337 pass, incl. new cases in `posthog-events.test.ts` and `posthog-mcp.test.ts`), `pnpm lint`, `pnpm build` all green. No manual testing performed.


packages/mcp/src/extensions/capture.ts
```diff
@@ -62,6 +62,7 @@ export function captureEvent(server: MCPServerLike, eventInput: McpEvent): Promi
     userIntentSource: eventInput.userIntentSource,
     isError: eventInput.isError,
     error: eventInput.error,
```
gesh, line 70: what's the difference between `error` and `error message`

gesh (commented):
Greptile comment seems legit

packages/mcp/src/extensions/capture.ts
lucasheriques: Good question. It's really three things sharing one failure, feeding two different consumers.

`error` is the full thrown value. It fans out into the separate `$exception` event, which is what gives us the stacktrace and the error-tracking grouping.

`errorType` (the line just below) is a coarse, low-cardinality category we keep on the `$mcp_tool_call` event itself, so the dashboard can break failures down by reason without having to join to `$exception`.

The message is derived from `error` onto that same tool-call event further downstream in `posthog-events.ts`. It's the only surviving copy when a server runs with `enableExceptionAutocapture: false`.

I added a comment at this spot in 116a7e1 spelling out the why.

packages/mcp/src/__tests__/posthog-events.test.ts
```diff
@@ -175,6 +175,40 @@ describe('buildPostHogCaptureEvents', () => {
...
+    expect(event.properties[PostHogMCPAnalyticsProperty.ErrorType]).toBe('validation')
+    expect(event.properties).not.toHaveProperty(PostHogMCPAnalyticsProperty.ErrorMessage)
+  })
+
+  it('omits error type/message properties on successful calls', () => {
+    const [event] = buildPostHogCaptureEvents(makeEvent({ isError: false }))
+
+    expect(event.properties).not.toHaveProperty(PostHogMCPAnalyticsProperty.ErrorType)
+    expect(event.properties).not.toHaveProperty(PostHogMCPAnalyticsProperty.ErrorMessage)
+  })
```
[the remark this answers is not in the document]
lucasheriques: Done, collapsed the four cases into a single `it.each` table in a0fcaa1. Kept the success case as a row where both expected values are absent. Agreed that parameterised is the right call here.

gesh (approved):

turnipdabeets:
@lucasheriques are we good to merge?

PR #4036, opened 2026-07-01
feat(logs): capture console calls made before the logs script loads

### LINKED ISSUE #3672: Logs console capture misses early startup errors

Yes. `posthog-js` browser Logs console capture can miss very early startup errors.

We enabled **Logs -> Capture console logs to Logs product** in PostHog project settings. Remote config correctly returns:

```js
logs: { captureConsoleLogs: true }
```

But the console wrapper only becomes active after remote config loads and `static/logs.js` is fetched. In our test, `static/logs.js` loaded around ~270ms after page start. Console errors before that were not backfilled into `/i/v1/logs`.

This matters for React hydration errors, which often happen during the initial startup window.

Either:

1. Buffer `console.*` calls as early as possible and flush them once Logs capture is ready, or
2. Provide a documented SDK option that installs console capture synchronously from initial config, before remote config finishes.

We considered adding our own early `console.error/warn` wrapper and flushing to `posthog.captureLog()`, but this is easy to get wrong and risks duplicate logs once PostHog's own console capture loads.

Session replay console logs can still contain these early errors, but then they are not available in the Logs product.

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

Test summary:

- Project setting enabled.
- Remote config contained `logs.captureConsoleLogs: true`.
- `static/logs.js` loaded successfully.
- Later `console.error` calls were sent to `/i/v1/logs` with 200.
- Immediate startup errors and immediate React hydration mismatch logs were not sent.
- Hydration mismatches triggered after the Logs hook loaded were sent successfully.

emmayusufu:
Fixes #3672

## Problem

Console autocapture only attaches after remote config returns **and** the external `logs.js` chunk is fetched — around 270ms in the report on #3672 — so everything a page writes to `console` during startup is lost.

Taking this over from @emmayusufu, who did the original investigation and the design work in the thread above. An earlier revision persisted the server-side enabled bit and loaded eagerly on the next visit; @dustinbyrne rejected that correctly, because a cached `true` had no stop path and a returning visitor kept emitting after the server had disabled capture. This is the buffer design agreed in that thread, rebuilt on current `main` after the logs core pipeline moved (#3865, #3895).

While verifying it in a real browser I found that `logs: { captureConsoleLogs: true }` passed to `init()` — the option the docs actually document — has never worked in any non-slim bundle. That is fixed here too; see "A second bug" below.

## Changes

The persisted bit is now only a hint that it is worth **recording**, never a licence to **emit**.

- `onRemoteConfig` persists `$logs_capture_enabled_server_side` when the server states an explicit `true` or `false`. A response omitting the `logs` key leaves the previous value alone. The key is registered `hidden` in `persistence-key-policy.ts` so it never reaches event properties.
- `initialize()` starts a small recorder when capture is enabled by either route — the local `captureConsoleLogs` opt-in, or that persisted hint. It wraps `console.*` and pushes `{ level, args, occurredAtMs, context }` into a capped array. Nothing loads, nothing is sent.
- On handover the recorder is stopped and `console` restored **before** `initializeLogs` runs, then the raw buffer goes to `replayConsoleBuffer` in the entrypoint, which runs it through the same `stringifyArgsSafely` as live capture.
- The buffer is dropped and `console` restored untouched when remote config disables capture, the config fetch fails, the script fails to load, the user opts out, `reset()` is called, or 30s elapse.

### A second bug, fixed here

`PostHogLogs` read `config.logs?.captureConsoleLogs` in its constructor, but the `PostHog` constructor builds the extension while `config` is still `defaultConfig()` — `init()` only applies the user's options afterwards. Every non-slim entrypoint imports `default-extensions` before creating the singleton, so the gate always read `undefined`. The documented public option produced no capture at all: no buffering, no live capture, and the `logs` chunk was never even requested. Only slim bundles, which construct the class inside `_initExtensions`, were unaffected. The opt-in is now read in `initialize()`.

Verified in Chrome against a real project, same page on both builds:

| | `main` | this branch |
|---|---|---|
| `logs.js` requested | no | yes |
| `/i/v1/logs` posts | 0 | 2 |
| records ingested | 0 | 4 |

### Details worth a reviewer's attention

- **Handover ordering.** The entrypoint assigns `console[level] = wrapped` directly rather than going through `patch()`, so a recorder left underneath it carries no layer marker and can never be spliced out — it would stay in the call path for the life of the page. Both steps are synchronous, so stopping first opens no capture gap.
- **`patch()` now walks `__rrweb_layer__` as well as `__posthog_layer__`.** rrweb's copy builds the identical `{ next }` layer under its own marker, and session replay's console plugin wraps the same methods. Without this, whichever of the two landed underneath leaked permanently. `patch.test.ts` covers the mixed chain.
- **Each buffered entry carries the SDK context and time from its own `console` call.** Core builds the OTLP record eagerly for exactly this reason; `identify()`, a session roll, an SPA navigation and the flags response all land inside the window this buffer covers. `PostHogLogs.captureLog` takes an optional `capturedAt` so a host can hand back what it snapshotted.
- **`timeUnixNano` and `observedTimeUnixNano` stay equal**, per the logs spec's *Log record data model* requirement. Confirmed on the wire, not just in unit tests.
- **The recorder is not started when nothing can confirm the hint** — with `advanced_disable_flags` and no preloaded config, `onRemoteConfig` never fires, so only the max-age backstop would have ended it.
- **Bounded twice**: `maxBufferSize` (100 by default) and a 30s age cap, so it cannot hold argument references indefinitely if `logs.js` never loads.
- **`BufferedConsoleEntry` lives in an internal `logs-types.ts`**, not `types.ts`, which is `export *`'d from the public entrypoints.

### Known limitation

Arguments are buffered by reference, so an object mutated between the `console` call and the replay serializes in its mutated state. Deep-copying at record time has a comparable set of failure cases; moving serialization into the core bundle is the real fix and is a follow-up. Documented on the type. This is the limitation discussed in the thread.

### Bundle size

`array.js` +661 bytes gzipped (85,263 → 85,924), measured from two forced clean rebuilds. `logs.js` +109 bytes. The `patch` helper was already in the main bundle.

### Follow-ups, all pre-existing

`array.full.no-external` inlines `initializeLogs` but ships no loader, so console capture is dead in that bundle; rrweb's own `patch` copy still walks one marker; `logger._log` reads `__rrweb_original__` one step instead of walking it; core has no same-millisecond timestamp bump; `'logs'` is missing from Playwright's `lazyLoadedJSFiles`, so `playwright/mocked/logs.spec.ts` stubs the extension and cannot fail.

### Libraries affected

- [x] posthog-js (web)
- [x] `@posthog/core` — optional `capturedAt` on `PostHogLogs.captureLog`; no behaviour change for posthog-node or posthog-react-native

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Anna took the PR over from @emmayusufu and drove the work; Claude Code did the implementation and verification under her direction.

- Three review passes with independent reviewer subagents (correctness, breaking-change/spec conformance, maintainability, test coverage), each finding triaged before anything was implemented. Findings were checked against the canonical logs spec in `sdk-specs` rather than taken at face value; two agent claims were rejected after being disproven.
- Test quality was driven by mutation testing rather than coverage: source mutations were applied and the suites re-run to confirm each new test actually fails when its production code is reverted. Several tests that looked like coverage but could not fail were rewritten.
- Verified in a real browser against a live project — built bundles served locally with the lazy chunk redirected to the local build, so the real loader, chunk and ingestion path all ran. The `main` baseline was built and run through the same page for the comparison above.
- The contributor's two commits are preserved unchanged; this pushed as a fast-forward, not a force-push.


emmayusufu:
Still keen to land this. It's the part-1 fix for #3672: persisting the console-capture enabled bit on remote config so log capture survives a page reload instead of resetting at init. CI is green, the only failing check is the Vercel preview, which can't build from a fork. Happy to rebase or adjust if someone can take a look.

dustinbyrne (commented):
Hey @emmayusufu, thanks for this. Can you take a look at the greptile comments? I agree that we should only persist explicit true/false, and not overwrite cached true for absent/nullish config.

emmayusufu:
Hey @dustinbyrne  thanks for the catch, I pushed a fix. It now only writes to persistence when captureConsoleLogs is explicitly set, so an absent or null/undefined config no longer overrides a previously cached true value. Also cleaned up the tests per the other two comments.

emmayusufu:
Keeping this open. The fix for the last review comment is pushed, just waiting on a re-review @dustinbyrne.

packages/browser/src/posthog-logs.ts
```diff
@@ -1,121 +1,39 @@
...
+    private _logger_instance: Logger | undefined
+
+    private _intervalLogCount: number = 0
+    private _intervalWindowStart: number = 0
+    private _droppedWarned: boolean = false
 
     constructor(private readonly _instance: PostHog) {
-        if (this._instance && this._instance.config.logs?.captureConsoleLogs) {
+        const localEnabled = !!this._instance?.config.logs?.captureConsoleLogs
+        const remoteEnabled = !!this._instance?.persistence?.props?.[LOGS_CAPTURE_ENABLED_SERVER_SIDE]
```
dustinbyrne: we're using persistence too early here - in most cases `PostHogLogs` is instantiated before persistence is defined

because of that, performing this check in `initialize` would be preferred

packages/browser/src/constants.ts
```diff
@@ -8,13 +8,13 @@
...
 export const DISTINCT_ID = 'distinct_id'
 export const DEVICE_ID = '$device_id'
-export const DEVICE_MODEL = '$device_model'
 export const ALIAS_ID_KEY = '__alias'
 export const CAMPAIGN_IDS_KEY = '__cmpns'
 export const EVENT_TIMERS_KEY = '__timers'
 export const AUTOCAPTURE_DISABLED_SERVER_SIDE = '$autocapture_disabled_server_side'
 export const HEATMAPS_ENABLED_SERVER_SIDE = '$heatmaps_enabled_server_side'
 export const EXCEPTION_CAPTURE_ENABLED_SERVER_SIDE = '$exception_capture_enabled_server_side'
+export const LOGS_CAPTURE_ENABLED_SERVER_SIDE = '$logs_capture_enabled_server_side'
```
dustinbyrne: this will need to be associated with a persistence key policy (packages/browser/src/persistence-key-policy.ts)

it should be `hidden`, e.g.,
```ts
 [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: { exposure: 'hidden' },
```

emmayusufu:
Still open and waiting on a re-review, the fix from the last round is pushed and green.

emmayusufu:
Both points addressed. The persisted bit is now read in initialize instead of the constructor, since persistence is not defined that early, and the key is registered as hidden in persistence-key-policy.ts so it stays out of event properties. Also updated the two policy test tables that track hidden keys. All three affected suites pass, 401 tests.

packages/browser/src/posthog-logs.ts
```diff
@@ -1,182 +1,72 @@
...
-                CONSOLE_SCOPE_NAME
-            )
-        }
-        return this._consoleCore
     }
 
     initialize() {
+        if (this._instance?.persistence?.props?.[LOGS_CAPTURE_ENABLED_SERVER_SIDE]) {
+            this._isLogsEnabled = true
+        }
```
dustinbyrne: Hmm - this is also problematic in that a cached `true` will immediately begin capturing logs, but when `onRemoteConfig` is called, if the response declares log capture disabled, it overwrites persistence and returns. Log capture continues.

dustinbyrne (dismissed):
I've been thinking about this, and I have a concern with using this cached value immediately.

If "capture console logs" is disabled server-side, users will expect to stop seeing console log data. However, this change would mean that returning users briefly send console logs upon return, until remote config is synchronized with the client.

In other cases where we perform client-side caching of server authoritative values (e.g., feature flags) this can often be a point of confusion when values change.

If we were to solve this, I think it'd be similar to what your PR description states: buffer console logs until remote config declares  whether to discard them or emit them. After all, there is an existing workaround: enable the client side config `config.logs.captureConsoleLogs`.

If we _were_ to buffer console logs on initialization, the constraint is that the solution would need to be mindful of its effect on bundle sizes, and remain local to the `PostHogLogs` extension.

I appreciate the work you've put in on this, and I hope you can understand where I'm coming from in regard to user experience.

emmayusufu:
That makes sense, the stale cache problem is real and the case you flagged confirms it, a cached true has no stop path once remote config says disabled. So I'll drop the persisted key entirely and go the buffer route. The shape I have in mind: on init, patch console with a tiny recorder that pushes method, args, and timestamp into a capped array, and nothing else loads. When remote config lands, enabled means load the extension and replay the buffer through the normal path, disabled means drop the buffer and restore console. The cap reuses the existing max buffer size so memory stays bounded, and all of it lives inside PostHogLogs. If that shape works for you I'll push it this week.

dustinbyrne:
That seems reasonable. Just a thought: we could use the new persistence key that you've added to know if we should begin buffering console logs. That way it's cleanly gated and doesn't need to run unconditionally.

marandaneto:
@emmayusufu moving this to draft until comments are addressed
cc @PostHog/logs

emmayusufu:
I pushed the buffer design we agreed on. The persisted key now only gates a small console recorder at init, so nothing gets captured from a cached value anymore. When remote config lands, enabled means the buffer replays through captureLog and then the extension loads, disabled means the buffer is dropped and the console is restored untouched. Also added tests for the recorder and reworded the changeset to match. Ready for another look @dustinbyrne @marandaneto.

marandaneto:
> I pushed the buffer design we agreed on. The persisted key now only gates a small console recorder at init, so nothing gets captured from a cached value anymore. When remote config lands, enabled means the buffer replays through captureLog and then the extension loads, disabled means the buffer is dropped and the console is restored untouched. Also added tests for the recorder and reworded the changeset to match. Ready for another look @dustinbyrne @marandaneto.

can you merge main into this branch and fix conflicts?
also sign your last commit https://github.com/PostHog/posthog-js/pull/4036/commits
we need signed commits

emmayusufu:
So it turned out to be more than a merge. Main refactored the whole logs module while this sat in draft (the core pipeline migration in #3865 and #3895), meaning my old commits patched code that no longer exists. I rebuilt the change on top of current main instead of dragging conflicts through a merge, and force-pushed a single signed commit, which also takes care of the unsigned June commit that was flagged.

Same design we agreed on above, adapted to the new architecture. The persisted key gates a small console recorder at init, nothing is captured from the cached value. When remote config lands, enabled means the buffer replays through _captureConsoleLog so replayed entries carry the console service name and scope, disabled means the buffer is dropped and the console is restored untouched. A failed config fetch also stops the recorder and drops the buffer.

The three affected suites are green locally. @marandaneto @dustinbyrne ready for another look.

packages/browser/src/posthog-logs.ts
```diff
@@ -41,6 +41,43 @@ const LOGS_SEND_TIMEOUT_MS = 90000
...
+        return JSON.stringify(arg) ?? String(arg)
+    } catch {
+        return String(arg)
+    }
+}
+
+const stringifyConsoleArgs = (args: any[]): string => {
+    const body = args.map(stringifyConsoleArg).join(' ')
+    return body.length > RECORDER_BODY_SIZE_LIMIT ? body.slice(0, RECORDER_BODY_SIZE_LIMIT) + '...' : body
+}
```
dustinbyrne: there's a lot of [existing logic in `logs.ts` to serialize safely](https://github.com/PostHog/posthog-js/blob/f6d5ef85ff19b5bed8f2922365fde9bc54021702/packages/browser/src/entrypoints/logs.ts#L293). there's a surprising number of failure cases that should be considered.

because of this, it'd be preferred if we can use that same logic. if possible, i recommend we just send the raw buffer to the logs entrypoint and let it perform the serialization. what do you think about this?

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +202,102 @@ export class PostHogLogs implements Extension {
...
+            if (!original) {
+                continue
+            }
+            this._originalConsoleMethods[level] = original
+            assignableWindow.console[level] = (...args: any[]) => {
+                if (args.length > 0 && this._consoleBuffer.length < maxBufferSize) {
+                    this._consoleBuffer.push({ level, args, timestamp: Date.now() })
+                }
+                original.apply(assignableWindow.console, args)
+            }
```
dustinbyrne: we should make sure the user hasn't opted out before buffering
```ts
 this._instance.is_capturing()
```

packages/browser/src/posthog-logs.ts
dustinbyrne: i suppose the downside would be that we're giving up frozen snapshots at the time of logging. deep copying everything likely has a similar set of failure cases.

the alternative is moving all the serialization into the core bundle and serializing immediately. i think it's okay to consider this a current limitation and improve it as a follow up item for later.

dustinbyrne: it's probably worth limiting the number of buffered calls by volume or by time.

e.g., imagine logs.js never loads and we're buffering / holding references indefinitely

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +202,102 @@ export class PostHogLogs implements Extension {
...
+        if (this._isRecordingConsole || !assignableWindow?.console) {
+            return
+        }
+        const maxBufferSize = resolveLogsConfig(this._instance?.config?.logs).maxBufferSize
+        for (const level of RECORDER_CONSOLE_LEVELS) {
+            const original = assignableWindow.console[level]
+            if (!original) {
+                continue
+            }
+            this._originalConsoleMethods[level] = original
```
dustinbyrne: take a look at [extensions/replay/rrweb-plugins/patch.ts](https://github.com/PostHog/posthog-js/blob/c708fbdc4da63e1b7db52919409e826454774b9a/packages/browser/src/entrypoints/logs.ts#L408-L415) - it handles patching functions like this while being able to remove patches out of installation order

https://github.com/PostHog/posthog-js/blob/c708fbdc4da63e1b7db52919409e826454774b9a/packages/browser/src/entrypoints/logs.ts#L408-L415
we should also set `console[level].__rrweb_original__` in the same manner (if it's not already set)

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +202,102 @@ export class PostHogLogs implements Extension {
...
+        // The recorder must hand the console back before the logs extension
+        // patches it, or a later restore would wipe the extension's wrapper.
+        const buffered = this._stopConsoleRecorder()
+
         if (isNullish(logCapture) || !logCapture) {
             return
         }
         this._isLogsEnabled = true
+        this._replayBufferedConsoleEntries(buffered)
         this.loadIfEnabled()
```
dustinbyrne: we're stopping buffering ahead of loading logs.js, so we'll lose console log data in the interim

dustinbyrne (commented):
hey @emmayusufu, thanks for revising this! this looks very promising. i've left some thoughts below

emmayusufu:
Thanks for the detailed pass. So the direction I'll take: keep the recorder running until logs.js has actually initialized, then hand it the raw buffer and let stringifyArgsSafely do the serialization in one place, meaning the interim gap and the duplicated serializer both go away. Also agreed on is_capturing before buffering, a time cap so the buffer can't hold references forever if logs.js never loads, and the __rrweb_original__ patch pattern so patches can come off out of order. On frozen snapshots, I'll treat it as the known limitation you described. I'll push the rework in the next day or two.

emmayusufu:
Pushed the rework. So the recorder now stays on until initializeLogs has run, meaning the window while the script loads is covered too, and the raw buffer goes to a new replayConsoleBuffer in the entrypoint so stringifyArgsSafely handles serialization in one place. Also added the is_capturing check before buffering, a 30 second cap so the buffer cannot hold argument references forever if the script never loads, and the recorder now goes through the layered patch helper with __rrweb_original__ set so patches come off cleanly in any order. Frozen snapshots stay a known limitation as discussed.

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +193,107 @@ export class PostHogLogs implements Extension {
...
+            this._consoleRecorderUnpatchers.push(
+                patch(assignableWindow.console, level, (next: any) => {
+                    const wrapped = (...args: any[]) => {
+                        if (
+                            this._isRecordingConsole &&
+                            args.length > 0 &&
+                            this._consoleBuffer.length < maxBufferSize &&
+                            this._instance?.is_capturing()
+                        ) {
+                            this._consoleBuffer.push({ level, args, timestamp: Date.now() })
```
dustinbyrne: **Agent-assisted review finding:** Each buffered entry stores raw arguments and a timestamp. It does not store the SDK context from the console call.

During replay, `_captureConsoleLog()` reads the current identity, session, URL, and feature flags. If `identify()` runs before replay, the SDK can assign an anonymous log to the identified user.

The normal logs pipeline saves this context before it delays processing. Please store `this._getSdkContext()` in each buffered entry. Add an internal replay method that accepts the stored context.

Keep the arguments raw so the lazy entrypoint can serialize them. Also use the stored timestamp as the OTLP event time.

packages/browser/src/posthog-logs.ts
```diff
@@ -253,9 +361,14 @@ export class PostHogLogs implements Extension {
         loadExternalDependency(this._instance, 'logs', (err) => {
             if (err || !phExtensions.logs?.initializeLogs) {
                 this._logger.error('Could not load logs script', err)
+                this._stopConsoleRecorder()
             } else {
                 phExtensions.logs.initializeLogs(this._instance)
```
dustinbyrne: **Agent-assisted review finding:** The code calls `initializeLogs()` before it stops the temporary recorder.

`initializeLogs()` installs a plain wrapper above the recorder. The unpatch function can only move through wrappers that have `__posthog_layer__`. It cannot reach and remove the recorder.

Stop the recorder and save its buffer before you call `initializeLogs()`. These operations run synchronously after `logs.js` loads. This order does not create a new capture gap.

The current test replaces `initializeLogs()` with a mock. Please add a test that uses the real entrypoint. Verify that the temporary recorder is removed. Also verify that console output and permanent capture each occur once.

packages/browser/src/types.ts
```diff
@@ -41,6 +41,13 @@ import type { PostHogMetrics } from './posthog-metrics'
 // Common types
 export type { Property, Properties, JsonType, JsonRecord } from '@posthog/types'
 
+// Args stay raw; the logs entrypoint owns their serialization.
+export interface BufferedConsoleEntry {
```
dustinbyrne: **Agent-assisted review finding:** `types.ts` is exported from the public `posthog-js` entrypoints. This makes `BufferedConsoleEntry` a public type.

This type only supports communication between the main bundle and the lazy bundle. Please move it to an internal module.

dustinbyrne (commented):
Human-driven, agent-assisted review.

The changeset describes the current buffer design. The PR title and description still describe the old eager-load design. Please update the PR title and description to match the current behavior.

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +192,167 @@ export class PostHogLogs implements Extension {
...
+    private _recordConsoleEntry(level: BufferedConsoleLevel, args: any[], maxBufferSize: number): void {
+        if (!this._isRecordingConsole || this._isRecordingConsoleEntry || args.length === 0) {
+            return
+        }
+        if (!this._instance?.is_capturing()) {
+            // Opting out mid-window releases the arguments already held, not just
+            // future ones.
+            this._stopConsoleRecorder()
+            return
+        }
```
dustinbyrne, line 419: this discards the buffer if we're opted out but only upon capturing one more log line. what do you think about immediately discarding the buffer from `opt_out_capturing` instead

packages/browser/src/posthog-logs.ts
```diff
@@ -161,31 +192,167 @@ export class PostHogLogs implements Extension {
...
+        // live argument graphs rather than serialized records, so the ceiling is core's
+        // flush threshold rather than its eviction cap.
+        const maxBufferSize = resolveLogsConfig(this._instance?.config?.logs).maxBufferSize
+        for (const level of BUFFERED_CONSOLE_LEVELS) {
+            const trueOriginal = originalConsoleMethod(assignableWindow.console[level])
+            if (!trueOriginal) {
+                continue
+            }
+            this._consoleRecorderUnpatchers.push(
+                patch(assignableWindow.console, level, (next: any) => {
```
dustinbyrne, line 386: agent found an annoying issue with `patch` here which is somewhat pre-existing and also sort of edge case-y. i'll have it add more information.

dustinbyrne (commented):
very close. i think the opt out case is something we should clean up before merging

packages/browser/src/posthog-logs.ts
dustinbyrne: The early logs recorder and rrweb use two copies of `patch()`.

Each copy stores its mutable `next` reference under a different name:

- The browser copy uses `__posthog_layer__`.
- The rrweb copy uses `__rrweb_layer__`.

The cleanup function follows these references to remove a wrapper from the middle of the chain.

This PR updates the browser cleanup to understand both names. The logs recorder can therefore remove itself when an rrweb wrapper is above it.

The rrweb cleanup still understands only `__rrweb_layer__`. Consider this order:

1. rrweb wraps `console.log`.
2. The early logs recorder wraps it with `__posthog_layer__`.
3. Session recording stops before the early logs recorder stops.

The rrweb cleanup sees `__posthog_layer__` above its wrapper. It cannot follow that reference, so it does not remove its wrapper. When the logs recorder later stops, it restores the rrweb wrapper that should have been removed.

This issue does not involve the permanent wrapper in the logs entrypoint. It only occurs if session recording stops during the early buffering window. The different property names are pre-existing, but this PR creates a mixed console-wrapper chain.

We can fix this by using one layer property, or by making both cleanup functions recognize both properties. A reverse cleanup-order test will cover this case.

packages/browser/src/entrypoints/logs.ts
dustinbyrne: pre-existing issue worth calling out, but not a blocker: it seems like we should also be calling `patch` here? the whole idea behind `patch` was that it can be unpatched in any order and retain the existing patch chain

packages/browser/src/posthog-logs.ts
dustinbyrne: i'm inclined to say it's non-blocking because changing the patch definition feels a little risky/scary given we use it for other things like fetch/XHR/history/etc.

packages/browser/src/posthog-logs.ts
turnipdabeets: Agreed — fixed in 535afa952 / a7653aecc.

`opt_out_capturing` wasn't the only route, so rather than hooking it directly I put the release in `_sync_opt_out_with_persistence`, which `opt_out_capturing`, `set_config({ opt_out_capturing_by_default })` and `clear_opt_in_out_capturing` all funnel through. It drops the buffer and unpatches `console` at that moment.

The `is_capturing()` check in `_recordConsoleEntry` stays as a backstop for anything that turns capture off without going through that funnel.

While adding the test I found the kill switch was leaking in a second way: it dropped the recorder buffer but left `_consoleQueue`/`_consoleCore`, so console records the chunk had already mirrored kept flushing after the server said no. That's fixed too, with a test — and a matching test pinning that `captureLog`/`logger` are *not* torn down, since that boundary was previously only stated in a comment.

packages/browser/src/posthog-logs.ts
turnipdabeets: Thanks — that write-up was precise enough to reproduce directly. Confirmed before fixing:

```
AFTER rrweb restore:    ["recorder","rrweb","base"]   ← rrweb wrapper survives its own restore
AFTER recorder restore: ["rrweb","base"]              ← and is then resurrected permanently
```

I took your "use one layer property" option rather than touching either walk, since you're right that changing the walk logic is risky with fetch/XHR/history on it. `patch` now publishes the same layer object under both `__posthog_layer__` and `__rrweb_layer__`. Purely additive — no removal semantics change, no walk change, and rrweb's copy is untouched — but its walk can now follow a layer this copy installed.

Added the reverse-cleanup-order test you asked for, as an `it.each` over both orders. Removing the second marker fails it.

Fixed in 535afa952.

packages/browser/src/entrypoints/logs.ts
turnipdabeets: Good catch — done in 2ab39e489. `initializeLogs` now installs through `patch()` instead of `assignableWindow.console[level] = wrapped`.

The old restore was `if (console[level] === wrapped) console[level] = original`, so anything wrapping `console` after the logs chunk loaded made `dispose()` a silent no-op — and it restored to the install-time value, which could resurrect a lower wrapper that had since been removed.

Test asserts the outer wrapper's `layer.next` moves from our wrapper to the real console on dispose; reverting the change fails it.

packages/browser/src/posthog-logs.ts
```diff
@@ -211,18 +259,40 @@ export class PostHogLogs implements Extension {
...
+            this._instance.persistence.register({ [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: !!logCapture })
+        }
+        if (isNullish(logCapture)) {
+            // No opinion from the server, so local config stands: release any hold from
+            // an older verdict, and stand down a recorder that only the hint started.
+            this._stopRecorderStartedByPersistedHint()
+            this.loadIfEnabled()
+            return
+        }
+        if (!logCapture) {
```
TueHaulund, line 263: The comment says the explicit `false` is deliberate, so this is a question about the behavior rather than the code: the server sends `false` unless the team turned console logs on in project settings, so anyone who sets `init({logs:{captureConsoleLogs:true}})` without that setting silently gets nothing. Is that what we want, and is it documented anywhere they'd look?

TueHaulund (approved):

packages/browser/src/posthog-logs.ts
turnipdabeets: You're right, and I have direct evidence for the premise rather than just the reasoning.

While testing this branch in a real browser against a test project where I had never touched the console-logs setting, the flags response came back:

```json
"logs": { "captureConsoleLogs": false }
```

So it's an explicit `false`, not an absent key — which means the kill switch I added treats "the team never turned it on" identically to "the team turned it off". With `init({ logs: { captureConsoleLogs: true } })` set, that run captured **zero** records and never even fetched the logs chunk.

On the documentation half: no. `docs/logs/installation/javascript.mdx` presents `logs: { captureConsoleLogs: true }` in `init()` as *the* way to enable console autocapture, with a warning about blast radius but nothing about a project setting. Searching the docs for the project-settings toggle turns up only the PII-scrubbing one. So today the documented path is the one that would silently stop working.

@dustinbyrne — you asked for the kill-switch semantics; does Tue's default-off case change it?

turnipdabeets: I think I was thinking about mobile like on replay where remote switch kills. But logs JS was written without that. and it's in the specs. https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/logs/spec.md

I'll go back to allowing local win.

turnipdabeets: Reverted in 3457ac661 — local opt-in wins.

`logs/spec.md` settles it:

> There is no remote kill-switch for the logs API today; the SDK SHALL NOT wait for a `logs.enabled`-style flag before capturing.

I'd leaned on the scenario below that ("when `captureConsoleLogs` is false the SDK does not mirror `console.*`") and read it as "the flag overrides local config". It isn't — that requirement is about *which surface* the flag gates (autocapture, not `captureLog`), not about whose setting wins. The replay precedent I reached for doesn't transfer either: `session-replay-ingestion-controls/spec.md` has an explicit "enabled **both** locally **and** remotely" requirement, which logs deliberately does not.

So now:

- `init({ logs: { captureConsoleLogs: true } })` captures, regardless of what the server reports. The documented path works.
- The remote flag still turns capture **on**, and still withdraws the persisted hint it granted — that's the only thing it ever granted, so nothing that used to work stops working.
- `captureLog` / `logger` remain ungated either way.

dustinbyrne: > @dustinbyrne — you asked for the kill-switch semantics; does Tue's default-off case change it?

i'm not sure what this is referring to - consent / opt out?

packages/browser/src/posthog-logs.ts
```diff
@@ -209,22 +250,41 @@ export class PostHogLogs implements Extension {
...
+            // The server reports `false` for every project that has not turned console
+            // capture on, so it cannot distinguish "not enabled" from "turned off" and
+            // does not override a local `captureConsoleLogs` opt-in. It does withdraw the
+            // persisted hint, which is the only thing it granted.
+            this._stopRecorderStartedByPersistedHint()
             return
         }
+        // The recorder keeps running until loadIfEnabled hands its buffer to the script.
         this._isLogsEnabled = true
```
dustinbyrne, line 271: Could we start the temporary recorder here before `loadIfEnabled()`?

On a first visit with no local option or persisted hint, `setup()` does not patch `console`. When the first remote `true` arrives, this starts loading `logs.js`, but the recorder remains off. Console calls made during that script request are still lost.

I reproduced this by holding the loader callback, writing to `console`, and then completing the load. `replayConsoleBuffer` received no entries.

A regression test can deliver remote `true` with no persisted hint, write to `console` before the loader callback, and verify that the entry is replayed.

packages/core/src/logs/index.ts
```diff
@@ -65,6 +65,17 @@ export class PostHogLogs {
...
   }
 
+  /**
+   * Drops every queued record. Credits the in-flight advance the same way a FIFO
+   * eviction does, so a batch already awaiting its response cannot slice records
+   * captured after this call.
+   */
+  clearQueue(): void {
+    const queue = this._instance.getPersistedProperty<BufferedLogEntry[]>(PostHogPersistedProperty.LogsQueue) ?? []
+    this._evictedSinceAdvance += queue.length
```
dustinbyrne: I think clearing this queue needs to invalidate the in-flight queue generation. I reproduced two cases:

1. `reset()` sets `_flushPromise` to `null` while the old request can still run. A new flush can then start. The old response can remove records captured after the reset. In my reproduction, the new send returned `retry-later`, but its record was already gone from the queue.
2. A pending send can return 413 after `clearQueue()`. `_flushInner()` retries from its old local `queue`, so it sends a purged console record again after opt-out.

Could we give each cleared queue a new generation? An old batch should stop before it retries or advances the queue when its generation is no longer current. We should also keep flushes serialized across `reset()`, or make their bookkeeping generation-local.

dustinbyrne (commented):
Human-driven, agent-assisted review.

packages/browser/src/posthog-logs.ts
turnipdabeets: Confirmed and fixed in 9c3c330.

You're right — on a first visit with no local option and no persisted hint, `setup()` starts no recorder, so the window between the remote `true` and `replayConsoleBuffer` was unguarded. `onRemoteConfig` now starts the recorder before `loadIfEnabled()`, guarded on `!this._isLoaded` so a later callback can't re-patch `console` behind the entrypoint (that buffer would never be collected and would pin argument graphs until the max-age backstop).

Regression tests added for both: `should record console calls made while the script loads after a first remote enable` and `should not start a recorder once the script is capturing live`. Both mutations (dropping the start, dropping the `_isLoaded` guard) are killed by them.

packages/core/src/logs/index.ts
turnipdabeets: Confirmed both cases and fixed in 9c3c330 — thanks, these were real.

I reproduced each as a failing test before changing anything:

**2 (413 after `clearQueue`)** — exactly as you describe. `_flushInner` holds a local `queue` and the `too-large` branch `continue`s without re-reading it, so the retry sliced from the stale copy. The test asserted a single send and got `[['purged one','purged two'], ['purged one']]`.

**1 (`reset()` nulling `_flushPromise`)** — the second `flush()` really did start a concurrent `_flushInner`, and the two then fought over `_evictedSinceAdvance`: it's a single field that each flush zeroes per batch, so the new one wiped the credit the old one needed. Test asserted `_sendLogsBatch` called once, got two.

Fix is the generation you suggested:

- `_queueGeneration`, bumped by `clearQueue()`.
- `_flushInner` captures it on entry and checks it immediately after `await _sendLogsBatch`, returning on mismatch. That one check covers both the 413 retry and the advance.
- `reset()` no longer nulls `_flushPromise`, so flushes stay serialized across it and there is no concurrent bookkeeping to get wrong.

One deliberate difference from your sketch: `reset()` does *not* bump the generation. The queue is host-owned and `reset()` doesn't clear it, so retiring an in-flight batch there would assume a clear that may not have happened and would re-send on the next flush. The browser's `reset()` calls `clearQueue()` first, which is what actually bumps. That also let the `_evictedSinceAdvance` credit `clearQueue()` used to add go away — the generation subsumes it.

Mutation-tested: dropping the generation check, dropping the bump in `clearQueue`, and re-adding `_flushPromise = null` to `reset()` are each killed.

packages/browser/src/posthog-logs.ts
turnipdabeets: Sorry — that was a stale question of mine from before the redesign, and it's since been answered.

It was about remote config as a kill switch, not consent. The question was whether a remote `captureConsoleLogs: false` should force console capture off even when the app passed `logs: { captureConsoleLogs: true }` to `init()`. It doesn't, and that's deliberate: the server sends `false` for every project that hasn't turned the setting on, so it can't distinguish "not enabled" from "turned off". Remote `false` withdraws the persisted hint (the only thing it granted) and leaves a local opt-in alone. That matches the logs spec.

Consent is a separate path — `_onOptOut()` stops the recorder, drops its buffer, and clears queued console records.

packages/browser/src/posthog-logs.ts
turnipdabeets: Correction to my previous message — that follow-up describes code that has since been replaced, in 134e3ae.

A further review pass showed the `!this._isRecordingConsole` guard I described was fixing the wrong layer, and it left two related holes:

- A **second** remote-config result can land while the logs script is still in flight — session recording asks for a fresh config of its own (`session-recording.ts` `_startCapture`, when `_isRemoteConfigFresh()` is false), and a failed fetch dispatches `{ok: false}` to the same subscription. That withdrew the hint-started recorder and dropped exactly the early lines this PR exists to preserve.
- A remote `false` after a `true` half-withdrew: the buffer was dropped, but the load it had already started still ran and console capture began live anyway.

Both now collapse into one rule in `_stopRecorderStartedByPersistedHint`: withdrawal applies only while `_isLogsEnabled` is unset. Once capture has actually been granted, the handover is coming and the hint is no longer the only thing holding the recorder up.

That made `_recorderStartedByHintOnly` redundant — at its only write site it was literally `!this._isLogsEnabled` — so the field is gone and `onRemoteConfig` no longer touches it. Net effect is less code than before the fix.

Tests: `should keep a hint-started buffer once remote config has granted capture` and `should replay the buffer when remote config flips to false after granting capture`. The test I named in my previous message is gone — it asserted the old behavior this commit deliberately corrects.

To be explicit about the behavior change, since it goes past restoring the prior state: previously, a hint-started recorder's buffer was dropped when a later remote result was `false` or failed. Now it survives, because the load is already in flight and the entrypoint captures live regardless.

packages/browser/src/posthog-logs.ts
```diff
@@ -39,9 +41,18 @@ export class PostHogLogs implements Extension {
...
+        // init on the next page load, not only after remote config returns.
+        if (this._instance?.persistence) {
+            this._instance.persistence.register({
+                [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: remoteEnabled,
+            })
+        }
+
+        if (!remoteEnabled) {
             return
         }
```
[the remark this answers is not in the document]
turnipdabeets: Good catch, and this is fixed — though the surrounding code has been rewritten since you looked, so the shape differs from the suggestion.

`onRemoteConfig` now derives a single nullable verdict and returns before touching persistence when there isn't one, so a response with no `logs` key (and a failed fetch, which we treat identically) can't overwrite a persisted `true`:

```ts
const logCapture = result.ok ? result.config.logs?.captureConsoleLogs : undefined
if (isNullish(logCapture)) {
    this._stopRecorderStartedByPersistedHint()
    return
}
this._instance?.persistence?.register({ [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: !!logCapture })
```

Pinned by `persists the server response so the next page load can buffer early console calls`, which walks `true` → `false` → absent and asserts `register` is not called for the last one. Mutation-checked: deleting the `isNullish` guard, inverting it, and dropping the early return are each caught.

packages/browser/src/__tests__/posthog-logs.test.ts
```diff
@@ -183,6 +184,56 @@ describe('posthog-logs', () => {
...
+                    siteApps: [],
+                    logs: { captureConsoleLogs: false },
+                }
+
+                logs.onRemoteConfig(response)
+
+                expect((mockPostHog as any).persistence.register).toHaveBeenCalledWith({
+                    [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: false,
+                })
+            })
```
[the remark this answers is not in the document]
turnipdabeets: [suggestion] Not taking this one, though not because of the convention — `it.each` is used 9 times in this file already.

The two tests you saw have since been merged, but into a single sequential test rather than a parameterised one, because the interesting assertion is order-dependent:

```
true → persists true
false → persists false
absent → register NOT called (the previous verdict survives)
```

That third case is the P1 bug from your other comment, and it only means something if a verdict was persisted first. Independent `it.each` cases each get a fresh mock, so the "must not overwrite" property can't be expressed that way.

Happy to split the two boolean cases out into `it.each` and keep the absent case separate if you'd prefer that shape, but it would mean asserting the same `register` call twice.

packages/browser/src/__tests__/posthog-logs.test.ts
```diff
@@ -183,6 +184,56 @@ describe('posthog-logs', () => {
...
+                        register: jest.fn(),
+                        props: { [LOGS_CAPTURE_ENABLED_SERVER_SIDE]: true },
+                    },
+                } as unknown as PostHog
+
+                const logsFromPersisted = new PostHogLogs(instanceWithPersistedBit)
+
+                expect((logsFromPersisted as any)._isLogsEnabled).toBe(true)
+            })
         })
```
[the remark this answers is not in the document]
turnipdabeets: Obsolete — resolving.

`initialize()` no longer exists on this class (0 occurrences in `posthog-logs.ts`); upstream #4533 replaced the extension lifecycle with `setup(client)` / `dispose()`, and the constructor test you were looking at is gone with it.

The behaviour you wanted covered is now asserted end-to-end rather than via a spy: `should buffer console entries instead of loading when the persisted bit is set` drives `setup()` and checks both that the loader was *not* called and that a console call lands in the buffer.

dustinbyrne (approved):

emmayusufu:
Thanks Anna, these are better than what I had. The `_sync_opt_out_with_persistence` funnel is cleaner than hooking `opt_out_capturing` directly, and the patch-splice write-up was a great read. Good call on the kill-switch revert too, the spec read settles it.

PR #4045, opened 2026-07-01
ci: skip heavy CI for non-impacting changes

marandaneto:
## Summary
- add a shared changed-files classifier that fails open on pushes, detection/API failures, empty file lists, unknown paths, local action changes, and non-safe workflow/source/config changes
- use a generic global safe-file set instead of SDK/package/folder category allowlists: markdown, docs, templates/CODEOWNERS/dependabot, and selected lightweight workflows only
- guard required heavy CI jobs after checkout but before pnpm/setup with generic `skip_heavy` / `run_heavy` outputs so required contexts still start and succeed for safe PR-only changes
- consider both `filename` and `previous_filename`, so renames from unsafe paths still run full CI

## Validation
- Ruby YAML parse for changed-files action and updated workflows
- local classifier simulations for release-only => skip, release + packages/node => run, unsafe previous_filename rename => run, push => run, and API failure => run
- git diff --check

### DIFF
--- a/.github/actions/changed-files-classifier/action.yaml
+++ b/.github/actions/changed-files-classifier/action.yaml
@@ -1,0 +1,170 @@
+name: Changed files classifier
+
+description: Classifies PR changed files so required CI jobs can skip heavyweight setup safely.
+
+inputs:
+  github-token:
+    description: Token used to read pull request files.
+    required: true
+
+outputs:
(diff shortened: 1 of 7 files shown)


marandaneto:
Closing this for now — I'll investigate other options. We don't want hardcoded paths this way.

PR #4048, opened 2026-07-01
fix(core): honor caller-supplied feature-flag props over cached values

turnipdabeets:
## Problem

When you pass a `$feature/<flag>` property explicitly to `capture()`, the shared core drops it in favor of its own cached flag value. This blocks passing server-evaluated flag values onto client exposure events (e.g. paywall screen views) as a safety net for when the client's cached flags are stale or a server-side person property changed evaluation.

The flag properties are injected in two places, each spreading `getCommonEventProperties()` **after** the caller's properties, so the cached value wins on collision:

- `PostHogCore.enrichProperties()` (`packages/core/src/posthog-core.ts`)
- `PostHogCoreStateless.buildPayload()` (`packages/core/src/posthog-core-stateless.ts`) — the final, authoritative merge

This also made the SDKs inconsistent. Precedence of an explicit `$feature/foo` passed to `capture()`:

| SDK | Behavior |
| --- | --- |
| posthog-js (web) | caller wins |
| posthog-android | caller wins |
| posthog-ios | caller wins ([posthog-ios#690](https://github.com/PostHog/posthog-ios/pull/690)) |
| **posthog-react-native / `@posthog/core`** | **SDK-cached wins** (this PR flips it to caller wins) |
| posthog-flutter | delegates to native |

## Changes

Added a shared `applyCallerFeatureFlagOverrides()` helper in `@posthog/core` and applied it at both injection sites so caller-supplied `$feature/*` and `$active_feature_flags` win over the cached values. All other property precedence is unchanged. `PostHogCoreStateless` (used by posthog-node) has no cached flag props to override, so this is a no-op there.

Added a regression test in `posthog.featureflags.spec.ts` asserting an explicit `$feature/*` and `$active_feature_flags` survive into the sent event over the loaded flags.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Anna directed this work with Claude Code following a customer (Portola) escalation about server/client feature-flag values, alongside the sibling iOS fix ([posthog-ios#690](https://github.com/PostHog/posthog-ios/pull/690)). The agent traced the precedence across all PostHog SDKs (web, iOS, Android, React Native core, Flutter) and found the target behavior "caller wins" already shipped in web and Android, so this aligns RN rather than inventing new behavior.

The non-obvious part: flags are injected twice (`enrichProperties` then `buildPayload`), so the override had to be applied at both sites — the deeper `buildPayload` is the authoritative one. Scope was deliberately limited to `$feature/*` and `$active_feature_flags`. Verified locally: full `@posthog/core` suite (690 tests) passes and eslint is clean.

### DIFF
--- a/.changeset/feature-flag-prop-precedence.md
+++ b/.changeset/feature-flag-prop-precedence.md
@@ -1,6 +1,5 @@
 ---
 "@posthog/core": patch
-"posthog-react-native": patch
 ---
 
 fix(core): feature-flag properties (`$feature/*` and `$active_feature_flags`) passed explicitly to `capture()` now take precedence over the SDK's cached flag values, matching posthog-js (web) and posthog-android
--- a/.changeset/feature-flag-prop-precedence.md
+++ b/.changeset/feature-flag-prop-precedence.md
@@ -1,5 +1,7 @@
 ---
 "@posthog/core": patch
+"posthog-react-native": patch
+"posthog-js-lite": patch
 ---
 
-fix(core): feature-flag properties (`$feature/*` and `$active_feature_flags`) passed explicitly to `capture()` now take precedence over the SDK's cached flag values, matching posthog-js (web) and posthog-android
+fix: feature-flag properties (`$feature/*` and `$active_feature_flags`) passed explicitly to `capture()` now take precedence over the SDK's cached flag values, matching posthog-js (web) and posthog-android
--- a/.changeset/feature-flag-prop-precedence.md
+++ b/.changeset/feature-flag-prop-precedence.md
@@ -1,0 +1,6 @@
+---
+"@posthog/core": patch
+"posthog-react-native": patch
+---
+
+fix(core): feature-flag properties (`$feature/*` and `$active_feature_flags`) passed explicitly to `capture()` now take precedence over the SDK's cached flag values, matching posthog-js (web) and posthog-android
--- a/packages/core/src/__tests__/posthog.featureflags.spec.ts
+++ b/packages/core/src/__tests__/posthog.featureflags.spec.ts
@@ -900,6 +900,28 @@ describe('PostHog Feature Flags v4', () => {
         })
       })
 
+      it('should let caller-supplied feature flag properties override cached values', async () => {
+        posthog.capture('test-event', {
+          '$feature/feature-1': 'server-value',
+          $active_feature_flags: ['server-flag'],
+        })
+
+        await waitForPromises()
+
+        expect(parseBody(mocks.fetch.mock.calls[1])).toMatchObject({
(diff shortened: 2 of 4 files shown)


packages/core/src/__tests__/posthog.featureflags.spec.ts
```diff
@@ -900,6 +900,28 @@ describe('PostHog Feature Flags v4', () => {
...
+              event: 'test-event',
+              properties: {
+                '$feature/feature-1': 'server-value',
+                $active_feature_flags: ['server-flag'],
+                '$feature/feature-2': true,
+              },
+            },
+          ],
+        })
+      })
```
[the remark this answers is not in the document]
turnipdabeets: Good call — addressed in `test(core): parameterise FF-override test over string, false, null`. It's now an `it.each` covering a string variant, boolean `false` (the client-side disable case), and `null`, each asserting the caller value survives both injection sites while a non-overridden cached flag stays intact.

ioannisj (approved):

.changeset/feature-flag-prop-precedence.md
```diff
@@ -0,0 +1,5 @@
+---
+"@posthog/core": patch
```
marandaneto, line 2: you need to add the SDKs that will benefit from this change, eg RN, Node, etc? otherwise this patch isnt available right away for customers that upgrade the sdk

turnipdabeets: Good call — added `posthog-react-native` and `posthog-js-lite` to the changeset (both extend the stateful `PostHogCore` that caches `$feature/*`, so they're the SDKs whose behavior actually flips to "caller wins"). They now get their own readable changelog entry rather than just an opaque `Updated dependencies → @posthog/core` line.

Left `posthog-node` off the explicit list on purpose: it derives `$feature/*` per-call from explicitly-passed flags rather than a cached store, so `applyCallerFeatureFlagOverrides` is a genuine no-op there (as noted in the PR). It'll still auto-cascade a `@posthog/core` dependency bump, so upgraders get the new core — just without a misleading behavioral entry. `changeset status` confirms the three-package patch bump.

PR #4049, opened 2026-07-01
ci(upgrade): post accurate alert when the upgrade PR fails (not "Failed to release")

turnipdabeets:
## Problem

The `🚨 Failed to release posthog-js@X 🚨` alerts in **#alerts-posthog-js** are misleading: the release actually **succeeds**. They fire from the downstream **PostHog Upgrade** workflow (`posthog-upgrade.yml`), which runs *after* publish to open a dependency-bump PR in the `PostHog/posthog` monorepo.

Worked example — `posthog-js@1.396.4` (2026-07-01):
- Release run [`28531105556`](https://github.com/PostHog/posthog-js/actions/runs/28531105556) published to npm + tagged + created the GitHub release at 17:03:12; its failure-notification steps were **skipped**. ✅
- PostHog Upgrade run [`28534364662`](https://github.com/PostHog/posthog-js/actions/runs/28534364662) failed at 17:04:05 in its `pnpm i` step:
  ```
  ERR_PNPM_EXOTIC_SUBDEP  Exotic dependency "uWebSockets.js" (resolved via url) is not allowed
  in subdependencies when blockExoticSubdeps is enabled
    ...while installing the dependencies of ultimate-express@2.0.9
  ```
- Both `release.yml` and `posthog-upgrade.yml` emitted the **same** `posthog-js-github-release-workflow-failure` event, and the CDP function ([project 11213](https://us.posthog.com/project/11213/functions/019ae9c0-03ee-0000-2f32-971f64be8ffe)) renders any of them as "Failed to release."

`concurrency: group: release` serializes releases, so there is **no** concurrent-publish race — an earlier revision of this PR chased that wrong diagnosis and has been reverted (thanks to the reviewer who flagged it).

## Changes

Scoped to `posthog-upgrade.yml`; `release.yml` is untouched (genuine release failures still alert exactly as before):

- The failure event is renamed to a distinct **`posthog-js-github-upgrade-pr-failure`**, so the release-failure CDP alert no longer fires for an upgrade failure. It's kept (best-effort) so upgrade-failure analytics are preserved.
- A new step posts an **accurate alert** to #alerts-posthog-js via `slackapi/slack-github-action` (the same pinned action the org's `slack-thread-reply` wraps):
  > ⚠️ The PostHog/posthog dependency bump for `posthog-js@X` failed — this is the downstream monorepo upgrade, not the posthog-js release.

The `posthog-js-github-release-workflow-failure` event is **not lost** — `release.yml` still emits it for real release failures.

**Assumption to confirm before merge:** the `SLACK_CLIENT_LIBRARIES_BOT_TOKEN` bot must be a member of **#alerts-posthog-js** (`C07HTMN9X47`) for `chat.postMessage` to succeed. The channel can be overridden with a `SLACK_ALERTS_POSTHOG_JS_CHANNEL_ID` repo var.

## Follow-up (separate, out of scope)

The recurring trigger is the `ultimate-express@2.0.9` → `uWebSockets.js` exotic subdep blocked by `blockExoticSubdeps` in `PostHog/posthog`; the upgrade `pnpm i` keeps failing until that's resolved. This PR only fixes the misleading alert, not the underlying install failure.

### Libraries affected

CI-only change — no library code changes, no version bump, no changeset.

## Checklist

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Anna directed this while investigating a false `Failed to release posthog-js@1.396.4` alert.

Claude Code first diagnosed a concurrent-publish race and proposed idempotent publish/tag/release guards; a reviewer pushed back (releases are concurrency-1). Re-investigation via the Actions API found the true source — the PostHog Upgrade run failing at `pnpm i` and emitting the shared release-failure event. The idempotency commits were reverted and the branch repurposed. Iterations settled on: a distinct event name (stops the false CDP alert without losing the release event, which `release.yml` still emits) plus a direct, accurate Slack alert (self-contained, no CDP dependency, no alerting gap). Not manually tested — workflow behavior is only observable on a real release, and Slack bot channel membership still needs confirming.

### DIFF
--- a/.github/workflows/posthog-upgrade.yml
+++ b/.github/workflows/posthog-upgrade.yml
@@ -178,5 +178,5 @@ jobs:
                   payload: |
                       {
                         "channel": "${{ vars.SLACK_ALERTS_POSTHOG_JS_CHANNEL_ID || 'C07HTMN9X47' }}",
-                        "text": ":warning: Could not open the PostHog upgrade PR for `${{ github.event.inputs.package_name }}@${{ github.event.inputs.package_version }}` — the release itself succeeded. <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|View workflow run>"
+                        "text": ":warning: The PostHog/posthog upgrade for `${{ github.event.inputs.package_name }}@${{ github.event.inputs.package_version }}` failed — the release itself succeeded. <https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}|View workflow run>"
                       }
--- a/.github/workflows/posthog-upgrade.yml
+++ b/.github/workflows/posthog-upgrade.yml
@@ -143,19 +143,23 @@ jobs:
               run: |
                   gh pr edit "$PR_NUMBER" --repo PostHog/posthog --add-reviewer PostHog/client-libraries-approvers --add-reviewer PostHog/team-client-libraries
 
-            # https://us.posthog.com/project/11213/functions/019ae9c0-03ee-0000-2f32-971f64be8ffe
+            # This job runs after posthog-js is already published; it only opens the
+            # dependency-bump PR in the PostHog/posthog monorepo. A failure here means
+            # the upgrade PR could not be opened, not that the release failed, so it
+            # emits its own event rather than the release-failure one.
+            # CDP function: https://us.posthog.com/project/11213/functions/019ae9c0-03ee-0000-2f32-971f64be8ffe
             - name: Send failure event to PostHog
               if: ${{ failure() }}
               uses: PostHog/posthog-github-action@58dea254b598fb5d469c0699c98af8288a7f7650 # v1.2.0
               with:
                   posthog-token: '${{ secrets.POSTHOG_PROJECT_API_KEY }}'
-                  event: 'posthog-js-github-release-workflow-failure'
+                  event: 'posthog-js-github-upgrade-pr-failure'
                   properties: >-
                       {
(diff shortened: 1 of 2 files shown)


ioannisj (approved):

.github/workflows/posthog-upgrade.yml
```diff
@@ -146,17 +146,30 @@ jobs:
             # https://us.posthog.com/project/11213/functions/019ae9c0-03ee-0000-2f32-971f64be8ffe
             - name: Send failure event to PostHog
               if: ${{ failure() }}
+              continue-on-error: true
               uses: PostHog/posthog-github-action@58dea254b598fb5d469c0699c98af8288a7f7650 # v1.2.0
               with:
                   posthog-token: '${{ secrets.POSTHOG_PROJECT_API_KEY }}'
-                  event: 'posthog-js-github-release-workflow-failure'
+                  event: 'posthog-js-github-upgrade-pr-failure'
```
marandaneto, line 153: if you change the event name
https://us.posthog.com/project/11213/functions/019ae9c0-03ee-0000-2f32-971f64be8ffe will stop working

marandaneto:
see thread https://posthog.slack.com/archives/C07HTMN9X47/p1782996221979129?thread_ts=1782925444.358909&cid=C07HTMN9X47

marandaneto (changes requested):
see thread https://posthog.slack.com/archives/C07HTMN9X47/p1782996221979129?thread_ts=1782925444.358909&cid=C07HTMN9X47

PR #4052, opened 2026-07-02
fix: skip update_referrer_info when user has registered custom $referrer

### LINKED ISSUE #4208: save_referrer() overwrites posthog.register() values for $referrer and $referring_domain in SPA + iframe contexts
*Written with Posthog AI after investigation*

**Describe the bug**

When using PostHog inside a Single Page Application (SPA) embedded as an iframe, calling posthog.register({ $referrer: '...', $referring_domain: '...' }) to set a custom referrer (e.g. the parent integrator's domain) does not persist across events. Every call to posthog.capture('$pageview') triggers an internal save_referrer() call that overwrites the registered values with document.referrer, which in an iframe context resolves to the iframe's own domain rather than the embedding site's domain.

The register() documentation states that registered super properties are "included with all events", but this is not the case for $referrer and $referring_domain because the SDK itself re-registers those same keys on every pageview.

**To reproduce**

Embed a PostHog-instrumented SPA as an iframe on an external site (e.g. blabla.fr embeds app.example.com)
On init, register a custom referrer: posthog.register({ $referrer: 'https:/blabla.fr/', $referring_domain: 'blabla.fr' })
Trigger a manual pageview as recommended for SPAs: posthog.capture('$pageview', { $current_url: currentUrl })
Observe the $referring_domain property on the captured event and subsequent events

**Expected behavior**

$referring_domain should be blabla.fr on all events, as registered via posthog.register().

**Actual behavior**

$referring_domain is app.example.com (the iframe's own domain) on all events after the first capture('$pageview') call, because save_referrer() overwrites the persistence store with document.referrer.

**Root cause**

Internally, save_referrer() calls this.register({ $referrer: document.referrer, $referring_domain: ... }). Since it runs after the user's register() call, it wins. In an iframe, document.referrer is the iframe's origin, not the parent page's domain.

This was a latent issue, but became visible after enabling full autocapture: the increased frequency of pageview detection means save_referrer() is called much more often, reducing the window during which the user-registered value survives in persistence to near zero.

**Workaround**

Pass $referrer and $referring_domain directly in every capture() call (direct properties have priority over persistence), and immediately call posthog.register() again after each capture('$pageview') to restore the correct persistence value for subsequent autocaptured events:

```js
posthog.capture('$pageview', {  
 $current_url: currentUrl,  
 $referrer: 'https://blabla.fr/',  
 $referring_domain: 'blabla.fr',  
})  
posthog.register({  
 $referrer: 'https://blabla.fr/',  
 $referring_domain: 'blabla.fr',  
})  
```

This workaround is brittle and requires wrapping every capture() call in the application. It would not be necessary if register() truly guaranteed super property precedence over internal SDK mechanisms.

**Environment**

posthog-js version: [your version]
Framework: Next.js (SPA, App Router)
Deployment context: iframe embedded on external sites
capture_pageview: false (manual pageview capture via SPA router)
Full autocapture enabled

**Related**
Issue PostHog/posthog#1442

tsushanth:
Closes PostHog/posthog-js#4208

## Problem

`posthog.register({ $referrer: '...', $referring_domain: '...' })` registers properties in `this.persistence`. However, on every `posthog.capture('$pageview')`, `update_referrer_info()` writes `document.referrer` to `this.sessionPersistence`.

In `calculateEventProperties`, the merge order is:

```ts
properties = extend(
    {},
    infoProperties,
    this.persistence.properties(),
    this.sessionPersistence.properties(),   // ← wins for shared keys
    properties
)
```

Because `sessionPersistence` is merged **after** `persistence`, the SDK-set `$referrer` (from `document.referrer`) silently overwrites the user-registered value on every pageview. This breaks SPAs embedded as iframes, where `document.referrer` is the iframe's own origin rather than the embedding page's domain.

## Fix

Before calling `update_referrer_info()`, check whether `$referrer` or `$referring_domain` already exists in `this.persistence.props`. If the user has explicitly registered custom values, skip the automatic update so the user-set values survive.

This is a targeted guard at the call site and requires no changes to `PostHogPersistence`.

## Behaviour

| Scenario | Before | After |
|---|---|---|
| No manual `register()` call | `document.referrer` set on every pageview ✓ | unchanged ✓ |
| User called `posthog.register({ $referrer })` | `document.referrer` overwrites on every pageview ✗ | user value preserved ✓ |
| User called `posthog.unregister('$referrer')` | `document.referrer` overwrites ✓ | `document.referrer` used ✓ |

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1352,17 +1352,16 @@ export class PostHog implements PostHogInterface {
             this.sessionPersistence.update_campaign_params()
         }
         if (this.config.save_referrer) {
-            // posthog.register() writes to this.persistence (regular persistence).
-            // update_referrer_info() writes to this.sessionPersistence, which is
-            // merged AFTER persistence in calculateEventProperties — so session values
-            // silently win for shared keys.  Skip the SDK update when the user has
-            // explicitly registered a custom $referrer or $referring_domain via
-            // posthog.register(), so their value is not overwritten on every pageview
-            // (e.g. in SPA-in-iframe scenarios where document.referrer is the iframe
-            // origin rather than the embedding page).
+            // Always update session referrer so fresh page loads reflect document.referrer.
+            // Then restore any value the user has explicitly registered via posthog.register()
+            // on a per-property basis, so a custom $referrer doesn't suppress an auto-detected
+            // $referring_domain (and vice-versa).
+            this.sessionPersistence.update_referrer_info()
             const persistenceProps = this.persistence.props
-            if (!('$referrer' in persistenceProps) && !('$referring_domain' in persistenceProps)) {
-                this.sessionPersistence.update_referrer_info()
+            for (const key of ['$referrer', '$referring_domain'] as const) {
+                if (key in persistenceProps) {
+                    this.sessionPersistence.register({ [key]: persistenceProps[key] })
+                }
             }
         }
 
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -1352,7 +1352,18 @@ export class PostHog implements PostHogInterface {
             this.sessionPersistence.update_campaign_params()
         }
         if (this.config.save_referrer) {
-            this.sessionPersistence.update_referrer_info()
+            // posthog.register() writes to this.persistence (regular persistence).
+            // update_referrer_info() writes to this.sessionPersistence, which is
+            // merged AFTER persistence in calculateEventProperties — so session values
+            // silently win for shared keys.  Skip the SDK update when the user has
+            // explicitly registered a custom $referrer or $referring_domain via
+            // posthog.register(), so their value is not overwritten on every pageview
+            // (e.g. in SPA-in-iframe scenarios where document.referrer is the iframe
+            // origin rather than the embedding page).
+            const persistenceProps = this.persistence.props
+            if (!('$referrer' in persistenceProps) && !('$referring_domain' in persistenceProps)) {
+                this.sessionPersistence.update_referrer_info()
+            }
         }
 
         if (this.config.save_campaign_params || this.config.save_referrer) {


packages/browser/src/posthog-core.ts
```diff
@@ -1352,7 +1352,18 @@ export class PostHog implements PostHogInterface {
...
+            // update_referrer_info() writes to this.sessionPersistence, which is
+            // merged AFTER persistence in calculateEventProperties — so session values
+            // silently win for shared keys.  Skip the SDK update when the user has
+            // explicitly registered a custom $referrer or $referring_domain via
+            // posthog.register(), so their value is not overwritten on every pageview
+            // (e.g. in SPA-in-iframe scenarios where document.referrer is the iframe
+            // origin rather than the embedding page).
+            const persistenceProps = this.persistence.props
+            if (!('$referrer' in persistenceProps) && !('$referring_domain' in persistenceProps)) {
+                this.sessionPersistence.update_referrer_info()
```
dustinbyrne: Could we resolve precedence per property during the final merge rather than skipping `update_referrer_info()` here? If session persistence already contains a referrer before `register()` is called, skipping the update leaves that stale value in place, and it still wins because session properties are merged after regular persistence.

This also suppresses both automatic properties when only one was registered. Consider continuing to populate session referrer data, then omitting each session key from `calculateEventProperties()` only when that same key exists in regular persistence.

To echo what greptile said, it's also worth adding test coverage for registration after an initial capture, single-key overrides, and unregister().

tsushanth:
Good point on both. Reworked to per-property precedence:

- Always call `update_referrer_info()` so fresh page loads capture the current `document.referrer`
- After the update, loop over `['$referrer', '$referring_domain']` and for each key the user has in `persistence.props`, write that value back into `sessionPersistence`

This way a user who registers only `$referrer` doesn't accidentally suppress the auto-detected `$referring_domain`, and a stale session value from a previous page can't survive because we always overwrite with the fresh `document.referrer` first.

turnipdabeets:
Thanks @tsushanth for surfacing this and kicking off the design discussion. We're consolidating on PostHog/posthog#4127, which implements the precedence-at-merge approach from the review thread here along with tests and a changeset, so I'm closing this one in its favor.

PR #4061, opened 2026-07-02
fix(convex): gate feature-flag refresh loop on local-eval being enabled

### LINKED ISSUE #4046: @posthog/convex: make feature flag refresh opt-in instead of running on every install

Yes. In `@posthog/convex`, the feature-flag refresh machinery runs for **every** install, unconditionally — including projects that only capture events, track exceptions, or evaluate flags remotely, and never use local flag evaluation at all.

Three things run on a schedule regardless of whether flags are in use:

1. A **supervisor cron** on a fixed 5-minute interval (`src/component/crons.ts` → `ensureRefreshLoop`).
2. A **self-rescheduling `refreshLoop` chain** that defaults to **60 seconds** (`src/component/lib.ts`, `DEFAULT_INTERVAL_SECONDS = 60`).
3. A **`refreshFlagDefinitions` action** scheduled on *every* loop tick.

The problem is where the gating lives. Neither the cron, nor `ensureRefreshLoop`, nor `refreshLoop` checks whether feature flags are actually being used. The only gate is *inside* `refreshFlagDefinitions`, which returns `{ status: 'skipped', reason: 'missing-keys' }` when the project token or personal API key is absent. But by then the scheduled mutation and action have already executed, logged, and counted toward usage. The code says as much:

> The cron registers unconditionally (see `crons.ts`); this is its runtime gate.

So the loop starts and keeps rescheduling itself even when local evaluation is impossible.

**Concrete cost at defaults, with no personal API key set** (i.e. a project not using local flags at all):

- ~1,440 `refreshLoop` mutation ticks/day
- ~1,440 `refreshFlagDefinitions` action invocations/day
- 288 `ensureRefreshLoop` supervisor runs/day

That's roughly **3,000+ scheduled function executions per day (~95k/month)** whose only outcome is an immediate `skipped` return. Every one of them lands in the Convex logs and counts against scheduled-function and action usage.

The impact is twofold:

- **Log noise.** These lines dominate the Convex deployment logs and bury the events, identifies, and exceptions that are the actual reason the component was installed. Debugging real activity means scrolling past a wall of flag-refresh churn.
- **Usage.** Scheduled function + action invocations that scale with *nothing the application does*. For projects that *do* set `POSTHOG_PERSONAL_API_KEY` but don't need per-minute freshness, it's worse — each tick fires an outbound `/flags/definitions` request every 60 seconds.

There is currently no way to turn this off.

Make feature-flag refresh **opt-in, off by default.**

When it isn't opted in:

- `refreshLoop` never starts and never reschedules itself.
- `ensureRefreshLoop` early-returns without scheduling anything.
- No `refreshFlagDefinitions` action is scheduled.

Net result: no per-minute loop, no per-tick action, and no log lines for projects that don't use local flags.

**Implementation shape.** Expose an explicit switch — either a component option (e.g. `app.use(posthog, { featureFlags: true })`) or an env var (e.g. `POSTHOG_ENABLE_LOCAL_FLAGS`). Because a cron's interval is fixed at registration and Convex only forwards component env vars at runtime — the exact constraint that #3957 / #4021 already worked around by moving to a runtime-driven loop — the gate belongs at runtime in `ensureRefreshLoop`, mirroring the pattern `refreshFlagDefinitions` already uses. When disabled, `ensureRefreshLoop` returns immediately and never spawns the chain, leaving only the cheap 5-minute supervisor no-op as residual cost.

**Lowest-surface variant, worth calling out.** Local evaluation *already* requires `POSTHOG_PERSONAL_API_KEY` and is a guaranteed no-op without it (`refreshFlagDefinitions` returns `missing-keys`; `getFlagDefinitions` reports `localEvalConfigured: false`). Gating the loop on the presence of that key would make the feature opt-in via the switch that *already* enables it — eliminating the churn for everyone who hasn't set a key, with **zero new configuration**. The remote `evaluateFlag` / `evaluateFlagPayload` / `evaluateAllFlags` methods are per-call actions, not part of the loop, so they'd be entirely unaffected either way.

- **Raising `POSTHOG_FLAGS_POLLING_INTERVAL_SECONDS`.** Not an off switch. It only throttles the `refreshLoop` chain; the 5-minute supervisor cron is a fixed floor that keeps firing (the `crons.ts` comment itself notes that "at very long intervals the supervisor itself dominates the cost"), and the loop still runs when no key is set — so the skipped executions and the log lines persist regardless of the interval.
- **Filtering the logs downstream.** Treats the symptom, not the cause. The scheduled-function and action usage is still incurred; only the visibility changes.
- **Not installing the component, or forking it.** Loses event capture and exception tracking too, or creates ongoing maintenance burden just to delete a few scheduled functions.
- **Gating the loop on `POSTHOG_PERSONAL_API_KEY`** (see above) — the low-effort path if a dedicated opt-in flag is more than the team wants to add right now.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context

Relevant source (observed on `@posthog/convex` v2.0.31):

- `packages/convex/src/component/crons.ts` — registers the 5-minute `ensureRefreshLoop` supervisor unconditionally.
- `packages/convex/src/component/lib.ts` — `ensureRefreshLoop`, `refreshLoop`, `refreshFlagDefinitions`, `DEFAULT_INTERVAL_SECONDS` (60), and `readPollingIntervalSeconds`. `refreshLoop` and `ensureRefreshLoop` have no key/usage gate; only `refreshFlagDefinitions` gates at runtime and still executes as a scheduled action first.

The self-rescheduling loop + supervisor shipped in **2.0.31** (PR #4021, fixing #3957) specifically so the runtime polling interval would be honored. This request builds on that: gate the *same* machinery behind an explicit opt-in so it doesn't run for projects that don't use local flags. It's purely additive and preserves current behavior for anyone who enables it.

turnipdabeets:
## Problem

The `@posthog/convex` component ran its feature-flag refresh loop on **every** install, unconditionally — including projects that only capture events or evaluate flags remotely and never use local evaluation. With no personal API key set, that's ~3,000 scheduled function executions per day whose only outcome is an immediate `skipped` return: pure log noise and usage, with no way to turn it off. (Closes #4046)

## Changes

- **Gate the loop on local evaluation actually being enabled.** `refreshLoop` and `ensureRefreshLoop` now early-return unless `POSTHOG_PERSONAL_API_KEY` is set — the key that already enables local eval. Event-only / remote-flag installs schedule no ticks at all; only the cheap 5-minute supervisor cron remains, and it no-ops. Behavior is unchanged when a key is configured.
- **New `POSTHOG_DISABLE_LOCAL_EVALUATION` off-switch** for installs that *do* set a key but don't want the background poll (e.g. they only use the remote `evaluateFlag*` actions). Accepts `true`/`1`/`yes`/`on`; unset by default.
- **Tests** for both gates (no key, and key + disabled, via both entry points) and the env-var parser.

Deliberately left untouched: `getFlagDefinitions.localEvalConfigured` still reflects only key presence — folding the disable flag into it would make the client throw a misleading "set your API key" error. The three key checks are intentionally distinct.

### Libraries affected

- [x] @posthog/convex

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — directed by @turnipdabeets.

Built with Claude Code. Reviewed by a parallel panel of specialized review agents (correctness, testing, maintainability) plus a Plannotator pass.

Key decisions:
- Followed the maintainers' call to keep polling **on by default when a key is set** (the issue's low-surface variant), rather than the issue title's "off by default" — gating on the existing key means zero new config for the common event-only case, plus an explicit env-var off-switch for the rest.
- Kept scope strictly to #4046; the separate `personalApiKey` → `secretKey` rename is out of scope here.

Verification: 172 unit tests pass; `tsc` and `eslint` clean. No manual/device testing performed.


packages/convex/src/component/refresh-loop.test.ts
```diff
@@ -148,4 +149,55 @@ describe('self-rescheduling flag refresh loop', () => {
...
+    })
+
+    test('key set with POSTHOG_DISABLE_LOCAL_EVALUATION=false: supervisor still schedules', async () => {
+      process.env.POSTHOG_DISABLE_LOCAL_EVALUATION = 'false'
+      const t = convexTest(schema, modules)
+
+      await t.mutation(internal.lib.ensureRefreshLoop, {})
+
+      expect(pending(byName(await scheduledJobs(t), 'refreshLoop'))).toHaveLength(1)
+    })
```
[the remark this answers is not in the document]
turnipdabeets: Good call — parameterised both pairs into `test.each` over the no-key / disabled setups in 874428edd. Kept the `=false` case as its own test since it's a positive control asserting the opposite (the loop *does* still schedule).

turnipdabeets:
Thanks @greptile — fixed in 874428edd. The `refreshLoop` JSDoc now reads "while local evaluation is enabled it queues a successor" instead of "unconditionally", so the fork warning still holds for the enabled case (the normal one) without being contradicted by the new gate.

packages/convex/src/component/refresh-loop.test.ts
```diff
@@ -148,4 +149,46 @@ describe('self-rescheduling flag refresh loop', () => {
...
+    )
+
+    test('key set with POSTHOG_DISABLE_LOCAL_EVALUATION=false: supervisor still schedules', async () => {
+      process.env.POSTHOG_DISABLE_LOCAL_EVALUATION = 'false'
+      const t = convexTest(schema, modules)
+
+      await t.mutation(internal.lib.ensureRefreshLoop, {})
+
+      expect(pending(byName(await scheduledJobs(t), 'refreshLoop'))).toHaveLength(1)
+    })
```
haacked, line 192: `suggestion`: There's no test ensuring that `getFlagDefinitions.localEvalConfigured` stays `true` when a key is set and `POSTHOG_DISABLE_LOCAL_EVALUATION=true`. This separation is essential for the PR: without it, the client throws a misleading "set your API key" error. But `localEvalConfigured` is only tested in `client/index.test.ts`, which mocks `runQuery`'s return value and never runs the real `env`-based computation.

If a future change accidentally makes the disable flag affect `localEvalConfigured` (easy to do since both read `POSTHOG_PERSONAL_API_KEY`), it flips to `false` and the client throws "local feature flag evaluation is not configured" at users who only meant to turn off the background poll, with nothing here to catch it.

```typescript
test('key set with DISABLE=true keeps localEvalConfigured true', async () => {
  process.env.POSTHOG_DISABLE_LOCAL_EVALUATION = 'true'
  const t = convexTest(schema, modules)
  const result = await t.query(api.lib.getFlagDefinitions, {})
  expect(result.localEvalConfigured).toBe(true)
})
```

packages/convex/src/component/lib.ts
```diff
@@ -370,13 +381,14 @@ async function recordLoopJob(ctx: MutationCtx, jobId: GenericId<'_scheduled_func
...
- * manual call forks the chain into two loops that run forever and double the cadence — the
- * supervisor can't detect the fork. Use `ensureRefreshLoop` to (re)start it.
+ * Never invoke directly (e.g. dashboard "Run function"): while local evaluation is enabled it queues
+ * a successor, so a manual call forks the chain into two loops that run forever and double the
+ * cadence — the supervisor can't detect the fork. Use `ensureRefreshLoop` to (re)start it.
  */
 export const refreshLoop = internalMutation({
   args: {},
   handler: async (ctx) => {
+    if (!localEvaluationEnabled()) return
```
haacked, line 391: `nit`: The JSDoc on [`refreshFlagDefinitions`](https://github.com/PostHog/posthog-js/blob/874428edd535058965db45ff4427dc34ff3e5d9d/packages/convex/src/component/lib.ts#L421) still says it's called automatically "when `POSTHOG_PERSONAL_API_KEY` is set," but the loop now also requires `POSTHOG_DISABLE_LOCAL_EVALUATION` not be set. Update it to "when local evaluation is enabled" to match the condition.

haacked (approved):
Nice fix! A couple small suggestions inline, nothing blocking.

packages/convex/src/component/refresh-loop.test.ts
turnipdabeets: Great catch — added exactly this in 6ed4460ab. It runs the real `env`-based `getFlagDefinitions` through `convexTest` (not the mocked `runQuery` in `client/index.test.ts`) and asserts `localEvalConfigured` stays `true` under `DISABLE=true` + key set. So if anyone later folds the disable flag into that computation, this fails loudly instead of silently throwing "set your API key" at users who only turned off the poll.

packages/convex/src/component/lib.ts
turnipdabeets: Fixed in 6ed4460ab — updated to "when local evaluation is enabled" to match the gate.

Issue #3522, opened 2026-05-04
Web and React Native: Dark Mode Support for Surveys


shahzore-qureshi-usmobile:
## Is your feature request related to a problem?

There is no official dark mode support for surveys on web and React Native. It would be nice to implement support for this out-of-the-box instead of overriding the colors locally.

## Describe the solution you'd like

Add a dark mode color palette to the PostHog Survey configuration on the PostHog Dashboard. 

## Describe alternatives you've considered

For now, we override the appearance with dark mode colors defined in the RN app. Here is example source code:

```typescript
import { SurveyPosition } from '@posthog/core';
import PostHog, { PostHogProvider, PostHogSurveyProvider, SurveyAppearance } from 'posthog-react-native';

export const posthog = new PostHog(POSTHOG_KEY, { host: POSTHOG_HOST });

export function PHProvider({ children }: { children: React.ReactNode }) {
  const isDarkMode = useIsDarkMode();
  const theme = isDarkMode ? DarkTheme : LightTheme;
  const { colors } = theme;

  const surveyAppearance: SurveyAppearance = useMemo(
    () => ({
      backgroundColor: colors.cardBg,
      textColor: colors.text,
      submitButtonColor: colors.mainBlue,
      submitButtonTextColor: colors.dynamicButtonText,
      ratingButtonColor: colors.fieldBg,
      ratingButtonActiveColor: colors.mainBlue,
      inputBackground: colors.fieldBg,
      inputTextColor: colors.text,
      borderColor: colors.cardStroke,
      position: SurveyPosition.Center,
    }),
    [colors],
  );

  return (
    <PostHogProvider client={posthog} autocapture={false}>
      <PostHogSurveyProvider
        client={posthog}
        overrideAppearanceWithDefault
        defaultSurveyAppearance={surveyAppearance}
      >
        {children}
      </PostHogSurveyProvider>
    </PostHogProvider>
  );
}
```

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [x] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

marandaneto:
cc @PostHog/team-surveys

PR #4066, opened 2026-07-03
feat(surveys): support dark mode appearance

Max-Reisinger:
## Problem

Surveys currently support a single appearance palette. This makes it difficult for apps to support dark mode without overriding survey colors locally.  #3522

## Changes

- Added dark mode appearance fields to the shared survey appearance type.
- Updated web survey rendering to resolve the effective appearance before applying CSS variables.
- Supports explicit `dark`, explicit `light`, and system-based dark mode behavior.
- Added unit coverage for dark mode appearance selection.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

### DIFF
--- a/packages/browser/src/posthog-surveys-types.ts
+++ b/packages/browser/src/posthog-surveys-types.ts
@@ -30,7 +30,7 @@ export interface SurveyEventWithFilters {
 
 // Extends core SurveyAppearance with browser-specific fields
 // Omit 'position' from core because browser's SurveyPosition has additional values (e.g., NextToTrigger)
-export interface SurveyAppearance extends Omit<CoreSurveyAppearance, 'position' | 'widgetType'> {
+export interface SurveyAppearance extends Omit<CoreSurveyAppearance, 'position' | 'widgetType' | 'darkModeAppearance'> {
     // Browser-specific fields not in core
     /** @deprecated - not currently used */
     descriptionTextColor?: string
@@ -49,6 +49,8 @@ export interface SurveyAppearance extends Omit<CoreSurveyAppearance, 'position' 
     // Browser's SurveyPosition has more options than core (e.g., NextToTrigger)
     position?: SurveyPosition
     widgetType?: SurveyWidgetType
+    darkModeColorScheme?: 'light' | 'dark' | 'system'
+    darkModeAppearance?: Omit<SurveyAppearance, 'darkModeColorScheme' | 'darkModeAppearance'>
 }
 
 export type SurveyQuestion = BasicSurveyQuestion | LinkSurveyQuestion | RatingSurveyQuestion | MultipleSurveyQuestion
--- a/packages/browser/src/posthog-surveys-types.ts
+++ b/packages/browser/src/posthog-surveys-types.ts
@@ -7,6 +7,7 @@
 import type { Properties, PropertyMatchType } from './types'
 import type {
     SurveyAppearance as CoreSurveyAppearance,
+    SurveyAppearanceColorScheme,
     SurveyQuestionTranslation,
     SurveyResponseValue as CoreSurveyResponseValue,
     SurveyTranslation,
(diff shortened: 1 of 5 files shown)


Max-Reisinger:
Addressed the review feedback and tests are passing. Could you take another look when you have a chance?

packages/core/src/types.ts
```diff
@@ -628,6 +630,11 @@ export type SurveyAppearance = {
   widgetColor?: string
 }
 
+export type SurveyAppearance = SurveyAppearanceBase & {
+  darkModeColorScheme?: SurveyAppearanceColorScheme
```
dustinbyrne, line 634: I think this may be misnamed. Something like `colorScheme` might be more fitting?

Otherwise there's ambiguity, e.g.,
```ts
darkModeColorScheme: 'light'
```

packages/browser/src/posthog-surveys-types.ts
```diff
@@ -49,6 +50,8 @@ export interface SurveyAppearance extends Omit<CoreSurveyAppearance, 'position'
     // Browser's SurveyPosition has more options than core (e.g., NextToTrigger)
     position?: SurveyPosition
     widgetType?: SurveyWidgetType
+    darkModeColorScheme?: SurveyAppearanceColorScheme
+    darkModeAppearance?: Omit<SurveyAppearance, 'darkModeColorScheme' | 'darkModeAppearance'>
```
dustinbyrne, line 54: This includes some additional types that are unused for dark mode styling, like:
```ts
 darkModeAppearance: {
     placeholder: 'Dark placeholder',
     backButtonText: 'Go back',
     position: SurveyPosition.TopLeft,
     thankYouMessageHeader: 'Dark thank-you',
     whiteLabel: true,
 }
```

dustinbyrne (commented):
Hey @jiuyige, thanks for putting this together! I appreciate the attention towards backwards compatibility.

I've left a couple of comments below and ran CI. There are some CI failures due to the fact that somewhere in the react native library we have a type which uses `Required<SurveyAppearance>`, so that will need to be addressed as well.

turnipdabeets:
Hey @jiuyige, thanks for taking this on — the schema shape and merge logic here are genuinely good work.

Unfortunately I don’t think we should merge this yet. The public API to a schema hasn't been designed yet. On web, survey.appearance comes entirely from the dashboard, which has no dark mode fields — so nothing would ever populate these, and meanwhile the field names and semantics would be locked into the public SurveyAppearance type, which we can't walk back once released. 

@PostHog/team-surveys can you help use with the dashboard schema? Do you have opinions on this? I believe we’ll need that first (in the main PostHog repo), then the SDKs can mirror it — web and React Native together.

In the meantime, both platforms have workarounds:

Web — override the CSS variables from page CSS (full list in addSurveyCSSVariablesToElement):

```
@media (prefers-color-scheme: dark) {
    [class*='PostHogSurvey'] {
        --ph-survey-background-color: #111827 !important;
        --ph-survey-text-primary-color: #f9fafb !important;
    }
}
```

React Native — the supported provider override with the reactive useColorScheme() hook:

```
const colorScheme = useColorScheme()
const appearance = useMemo(() => (colorScheme === 'dark' ? darkAppearance : lightAppearance), [colorScheme])

<PostHogSurveyProvider client={posthog} overrideAppearanceWithDefault defaultSurveyAppearance={appearance}>
```

Once the schema lands, much of this PR carries over — we'd love your help on the SDK side then. 🙏

turnipdabeets:
Closing as currently not planned

PR #4071, opened 2026-07-03
fix(ai): declare inclusive cache token reporting for OpenAI-compatible providers

### LINKED ISSUE #3615: Cache-aware cost calculation seems off for Claude via OpenRouter (still after #3022's fix)
## Bug description

On `@posthog/ai@7.16.13`, the cost PostHog reports for Claude generations via OpenRouter is ~3.4x what OpenRouter actually bills. I saw #3022 and the fix in #3433, but the numbers still come out high on my side. Might be missing something, happy to be corrected.

For one specific generation, side-by-side:

| Source | Value |
|---|---|
| **OpenRouter dashboard — final cost** | **$0.0342** |
| OpenRouter — subtotal | $0.106 |
| OpenRouter — cache read discount | −$0.0721 |
| OpenRouter — tokens | 32,611 → 561 |
| OpenRouter — provider cached | 27,929 prompt tokens |
| | |
| **PostHog event — `$ai_total_cost_usd`** | **$0.1146267** |
| PostHog — `$ai_input_cost_usd` | $0.1062117 |
| PostHog — `$ai_output_cost_usd` | $0.008415 |
| PostHog — `$ai_input_tokens` | 32,611 |
| PostHog — `$ai_cache_read_input_tokens` | 27,929 |
| PostHog — `$ai_output_tokens` | 561 |

Relevant event properties:

| Property | Value |
|---|---|
| `$ai_lib_version` | `7.16.13` |
| `$ai_provider` | `openai` |
| `$ai_model` | `anthropic/claude-sonnet-4.6` |
| `$ai_base_url` | `https://openrouter.ai/api/v1` |
| `$ai_cache_reporting_exclusive` | `true` |
| `$ai_cost_model_source` | `openrouter` |

## How to reproduce

1. Install `@posthog/ai@7.16.13` and `openai@^4`.

2. Create a client via `@posthog/ai`'s `OpenAI` wrapper, pointed at OpenRouter:
   ```ts
   import { OpenAI } from '@posthog/ai';

   const client = new OpenAI({
     baseURL: 'https://openrouter.ai/api/v1',
     apiKey: process.env.OPENROUTER_API_KEY,
     posthog, // posthog-node client
   });
   ```

3. Call `chat.completions.create` against a Claude model with `cache_control: { type: 'ephemeral' }` markers (streaming, with or without tool calls):
   ```ts
   await client.chat.completions.create({
     model: 'anthropic/claude-sonnet-4.6',
     messages: [...],   // with cache_control markers
     tools: [...],
     stream: true,
   });
   ```
4. Send the same request enough times to get a cache hit (so `prompt_tokens_details.cached_tokens > 0` on a subsequent call).
5. Compare PostHog's `$ai_total_cost_usd` against the matching generation in the OpenRouter activity dashboard. The PostHog number lands on OpenRouter's pre-discount subtotal rather than the final cost.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

## Additional context
Sharing what I traced back in case it's useful:

PostHog's input cost reproduces exactly if I assume the server is treating `$ai_input_tokens` and `$ai_cache_read_input_tokens` as two separate buckets.

| Component | Tokens | Rate | Cost |
|---|---:|---:|---:|
| `$ai_input_tokens` at full input rate | 32,611 | $3/M | $0.097833 |
| `$ai_cache_read_input_tokens` at cache rate | 27,929 | $0.30/M | $0.008379 |
| **Total** | | | **$0.106212** ← matches `$ai_input_cost_usd` |

But the 32,611 sent as `$ai_input_tokens` is OpenRouter's `prompt_tokens`, which for Anthropic models on OpenRouter already **includes** the 27,929 cached tokens. So the cached portion ends up counted twice — once at $3/M inside `$ai_input_tokens`, and once at $0.30/M via `$ai_cache_read_input_tokens`.

One guess on why this might not be caught by the previous fix: since I'm hitting OpenRouter via its OpenAI-compatible endpoint, `$ai_provider` on the event ends up as `openai` even though the underlying model is Claude. The cache-aware adjustment from #3433 may not cover that combination.

### Cross-refs

- #3022 (original cache-aware cost issue)
- #3433 (Vercel V3 fix shipped in 7.16.3)

DerGeraetK:
## Problem

Fixes #3615

Tracking Claude generations through OpenAI-compatible hosts (e.g. OpenRouter) reports ~3.4x inflated costs. This wrapper forwards OpenAI-convention usage — `prompt_tokens` (32,611) already *includes* cached tokens, and `prompt_tokens_details.cached_tokens` (27,929) is forwarded separately. Ingestion classifies any `claude*`/`anthropic*` model as Anthropic-convention (*exclusive*) and prices input at full rate *plus* cache reads: 32,611 × $3/M + 27,929 × $0.30/M = **$0.1062117** — exactly the `$ai_input_cost_usd` reported in the issue. The cached tokens are billed twice.

## Changes

- New thin seam `packages/ai/src/openai/capture.ts`: every capture from the OpenAI wrapper (chat ± streaming, Responses API, embeddings, Azure) now declares the documented property **`$ai_cache_reporting_exclusive: false`**. Ingestion honors the explicit flag (PostHog/posthog#49252) and performs the subtraction itself.
- `openai/index.ts` + `openai/azure.ts` switch one import each to the seam — covering all ~30 call sites; user-supplied `posthogProperties` can still override the flag.
- 3 regression tests (issue numbers as fixtures; streaming path; user override). They use plain `test()` rather than this file's `conditionalTest` because they are fully mocked and CI doesn't set `OPENAI_API_KEY` — gated tests would never run.

**Why not client-side subtraction (as #3433 did for the Vercel path):** subtraction needs Claude-model detection in the SDK (fragile across model-id formats like `us.anthropic.claude-…`), could double-subtract if ingestion later extends its inclusive-detection, and makes `$ai_input_tokens` diverge from the raw `$ai_usage`. The flag declares provenance — which only the SDK knows — and leaves token semantics untouched. Happy to switch to the subtraction approach if you prefer consistency with #3433. (The AI triage on the issue sketched the subtraction variant; it analyzed only this repo and couldn't see the ingestion-side flag.)

**Notes for reviewers:**
- Expected residual: even with correct accounting, totals land ~10% below OpenRouter's billed amount (route-specific pricing; cache *writes* have no field in OpenAI-shape usage). This PR fixes the double-billing, not reseller invoice parity — `costOverride` remains the path to exact parity.
- Same neighborhood, out of scope: `openai-agents/processor.ts` maps usage independently and may deserve the same declaration — happy to file a follow-up. Vercel/LangChain/Anthropic paths already handle Claude cache tokens.
- Self-hosted ingestion older than PostHog/posthog#49252 (03/2026) ignores the flag.

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code
- [x] Ran changeset generation (`.changeset/fix-openai-wrapper-cache-reporting-convention.md`)

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — DRI: @DerGeraetK

- Built with Claude Code. Root cause verified against ingestion source (`nodejs/src/ingestion/pipelines/ai/costs/input-costs.ts`) and reproduced arithmetically to 7 decimals from the issue's event values. An adversarial review pass challenged the diagnosis (alternative hypotheses: mispriced cache rate, model-name mismatch, missing `cached_tokens` — all refuted) and flipped the fix choice from client-side subtraction to convention declaration.
- Verified locally: jest for the openai suite (12 passed incl. the 3 new regression tests, `capture.ts` at 100% coverage) and eslint on the touched files. Full root build/lint/test runs in CI — the local Windows environment has pre-existing native-binding issues unrelated to this change.
- Agent-authored code, human-reviewed and directed; no self-merge.

### DIFF
--- a/packages/ai/src/openai/capture.ts
+++ b/packages/ai/src/openai/capture.ts
@@ -1,0 +1,21 @@
+import { PostHog } from 'posthog-node'
+import { captureAiGeneration as baseCaptureAiGeneration, CaptureAiGenerationOptions } from '../captureAiGeneration'
+
+/**
+ * OpenAI-compatible usage reports `prompt_tokens` INCLUSIVE of cached tokens
+ * (`prompt_tokens_details.cached_tokens` is a subset of it), unlike Anthropic's
+ * exclusive convention. Ingestion auto-classifies Claude-shaped models as
+ * exclusive regardless of provider, so events for Claude served through
+ * OpenAI-compatible hosts (e.g. OpenRouter) would get cache reads billed twice.
+ * Declaring the convention on every event from this wrapper lets ingestion
+ * normalize correctly (see PostHog/posthog#49252); for non-Claude models the
+ * flag is a no-op. Callers can still override it via posthogProperties.
+ */
+export const captureAiGeneration = (client: PostHog, options: CaptureAiGenerationOptions): Promise<void> =>
+  baseCaptureAiGeneration(client, {
+    ...options,
+    properties: {
+      $ai_cache_reporting_exclusive: false,
+      ...options.properties,
+    },
+  })
--- a/packages/ai/src/openai/capture.ts
+++ b/packages/ai/src/openai/capture.ts
@@ -2,6 +2,17 @@ import { PostHog } from 'posthog-node'
 import { captureAiGeneration as baseCaptureAiGeneration, CaptureAiGenerationOptions } from '../captureAiGeneration'
 
 /**
+ * The declared convention only describes the wrapper's own usage numbers, so
+ * when the caller passes any of these through posthogProperties the wrapper no
+ * longer knows the convention of the reported counts (callers working around
+ * the double-billing already pass exclusive ones) and must not declare it.
+ * Output/reasoning token overrides don't affect the input/cache relationship,
+ * so they don't suppress the declaration. Subset of the input/cache keys in
+ * `TOKEN_PROPERTY_KEYS` (../utils.ts) — keep in sync if that taxonomy grows.
+ */
+const INPUT_OR_CACHE_TOKEN_KEYS = ['$ai_input_tokens', '$ai_cache_read_input_tokens', '$ai_cache_creation_input_tokens']
+
+/**
  * OpenAI-compatible usage reports `prompt_tokens` INCLUSIVE of cached tokens
  * (`prompt_tokens_details.cached_tokens` is a subset of it), unlike Anthropic's
  * exclusive convention. Ingestion auto-classifies Claude-shaped models as
@@ -9,13 +20,23 @@ import { captureAiGeneration as baseCaptureAiGeneration, CaptureAiGenerationOpti
  * OpenAI-compatible hosts (e.g. OpenRouter) would get cache reads billed twice.
  * Declaring the convention on every event from this wrapper lets ingestion
  * normalize correctly (see PostHog/posthog#49252); for non-Claude models the
- * flag is a no-op. Callers can still override it via posthogProperties.
+ * flag is a no-op. Callers can still override it via posthogProperties, and
+ * when they pass through input or cache token counts themselves the flag stays
+ * unset unless they declare it explicitly.
  */
-export const captureAiGeneration = (client: PostHog, options: CaptureAiGenerationOptions): Promise<void> =>
-  baseCaptureAiGeneration(client, {
+export const captureAiGeneration = (client: PostHog, options: CaptureAiGenerationOptions): Promise<void> => {
+  const props = options.properties
+  // Own-property check, matching how getTokensSource detects passthrough and
+  // how the properties spread actually copies values into the event.
+  const callerReportsTokens =
+    props !== undefined && INPUT_OR_CACHE_TOKEN_KEYS.some((key) => Object.prototype.hasOwnProperty.call(props, key))
+  return baseCaptureAiGeneration(client, {
     ...options,
-    properties: {
-      $ai_cache_reporting_exclusive: false,
-      ...options.properties,
-    },
+    properties: callerReportsTokens
+      ? props
+      : {
+          $ai_cache_reporting_exclusive: false,
+          ...props,
+        },
   })
+}
--- a/.changeset/fix-openai-wrapper-cache-reporting-convention.md
+++ b/.changeset/fix-openai-wrapper-cache-reporting-convention.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+fix(ai): declare `$ai_cache_reporting_exclusive: false` on OpenAI wrapper events so ingestion no longer double-bills cached input tokens for Claude models served through OpenAI-compatible hosts (e.g. OpenRouter)
--- a/.changeset/fix-openai-wrapper-cache-reporting-convention.md
+++ b/.changeset/fix-openai-wrapper-cache-reporting-convention.md
@@ -2,4 +2,4 @@
 '@posthog/ai': patch
 ---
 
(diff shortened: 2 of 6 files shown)


DerGeraetK:
Done in 54eb6e2 — merged the two non-streaming tests into a single `test.each` (default declaration / user override), as suggested. The Claude-via-OpenRouter fixture from #3615 and the raw-token assertions now run for both rows; streaming keeps its own test since it exercises a distinct code path.

dustinbyrne:
change looks clean but i'm lacking context to approve, tagging @richardsolomou and @carlos-marchal-ph

packages/ai/src/openai/capture.ts
```diff
@@ -0,0 +1,21 @@
...
+ * OpenAI-compatible hosts (e.g. OpenRouter) would get cache reads billed twice.
+ * Declaring the convention on every event from this wrapper lets ingestion
+ * normalize correctly (see PostHog/posthog#49252); for non-Claude models the
+ * flag is a no-op. Callers can still override it via posthogProperties.
+ */
+export const captureAiGeneration = (client: PostHog, options: CaptureAiGenerationOptions): Promise<void> =>
+  baseCaptureAiGeneration(client, {
+    ...options,
+    properties: {
+      $ai_cache_reporting_exclusive: false,
```
richardsolomou: Could we leave `$ai_cache_reporting_exclusive` unset when `posthogProperties` overrides the input or cache token fields, unless the caller sets it explicitly? Token passthrough is supported today, and callers may already provide exclusive counts as a workaround: with 4,682 input tokens and 27,929 cached tokens, forcing `false` makes ingestion subtract the cache again and calculate −23,247 uncached tokens. I’d also add a regression test for this case.

richardsolomou (changes requested):
Could we add an unconditional mocked Azure assertion as well? Azure is separately switched to the helper, but the existing tests are credential-gated, so CI may not catch that path bypassing the default or override behavior.

DerGeraetK:
Done in 3150506 — both points addressed:

**Passthrough:** `$ai_cache_reporting_exclusive` now stays unset when `posthogProperties` passes through `$ai_input_tokens`, `$ai_cache_read_input_tokens` or `$ai_cache_creation_input_tokens`, unless the caller declares it explicitly (own-property check, matching how `getTokensSource` detects passthrough). Added two rows to the reporting-convention `test.each` with the numbers from your example: 4,682 exclusive input / 27,929 cached now asserts the property is absent entirely, and a passthrough-plus-explicit-declaration case guards the opposite direction. Output/reasoning-token overrides deliberately don't suppress the declaration since they don't affect the input/cache relationship.

**Azure:** added an unconditional mocked block to `azure-openai.test.ts` asserting the default declaration and the explicit override on the Azure path — deliberately not `conditionalTest`, so CI covers it without `AZURE_OPENAI_API_KEY`.

richardsolomou (approved):
Looks good to me, leaving final say to @carlos-marchal-ph. One optional follow-up: could we parameterize the passthrough coverage so `$ai_input_tokens`, `$ai_cache_read_input_tokens`, and `$ai_cache_creation_input_tokens` each independently prove that the reporting declaration stays unset? The current case always includes `$ai_input_tokens`, so it doesn’t exercise either cache key by itself.

DerGeraetK:
Done in 27fbddd — each of the three keys now independently proves the declaration stays unset when passed through alone. Also merged `main` back in (9464aee): #4076 and this branch both appended a describe block at the end of `azure-openai.test.ts`, which had started conflicting.

carlos-marchal-ph (approved):

PR #4072, opened 2026-07-03
fix(ai): capture OpenAI service_tier in $ai_model_parameters for accurate cost tracking

nandinitiw:
## What

Fixes PostHog/posthog#72429.

OpenAI's `service_tier` parameter (`flex`, `auto`, `priority`) affects the price of a request (e.g. `flex` is half the standard price), but it was not being captured in the event — so PostHog's backend had no way to apply the correct multiplier when calculating `$ai_total_cost_usd`.

## How

`getModelParams` in `src/utils.ts` extracts OpenAI request parameters into `$ai_model_parameters`. Added `service_tier` to the list of extracted keys so it flows through to the event automatically.

## Testing

Added 4 unit tests to `tests/utils.test.ts` covering:
- `service_tier` is included when set (`flex`, `priority`)
- `service_tier` is omitted when not provided
- `service_tier` is captured alongside other params (e.g. `temperature`)
- null params return `{}`

All existing tests continue to pass.

### DIFF
--- a/packages/ai/src/utils.ts
+++ b/packages/ai/src/utils.ts
@@ -126,6 +126,7 @@ export const getModelParams = (
     'language',
     'response_format',
     'timestamp_granularities',
+    'service_tier',
   ] as const
 
   for (const key of paramKeys) {
--- a/.changeset/ai-openai-service-tier-model-params.md
+++ b/.changeset/ai-openai-service-tier-model-params.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+fix(ai): capture OpenAI `service_tier` in `$ai_model_parameters` so PostHog can correctly attribute costs for flex and priority tier requests
--- a/packages/ai/tests/utils.test.ts
+++ b/packages/ai/tests/utils.test.ts
@@ -1,4 +1,4 @@
-import { toContentString, getTokensSource } from '../src/utils'
+import { toContentString, getTokensSource, getModelParams } from '../src/utils'
 
 describe('getTokensSource', () => {
   it.each([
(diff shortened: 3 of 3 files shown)


nandinitiw:
Still keen to get this merged — happy to address any feedback or make changes if needed. Let me know!

nandinitiw:
Hey @marandaneto @carlos-marchal-ph — would appreciate a review when you have a moment!

packages/ai/src/utils.ts
```diff
@@ -126,6 +126,7 @@ export const getModelParams = (
     'language',
     'response_format',
     'timestamp_granularities',
+    'service_tier',
```
ioannisj, line 129: Hey, thanks for this! This was flagged by my agent, worth a look? 

---

One thing worth thinking through: this captures the `service_tier` from the request, but for cost accuracy the backend really wants the tier the request was actually billed at, and OpenAI returns that in the response (and documents that it can differ from what you asked for).

From the OpenAI SDK types (`openai@6.25.0`, `chat/completions/completions.d.ts:183-190`, on the `ChatCompletion` response object):

> When the `service_tier` parameter is set, the response body will include the `service_tier` value based on the processing mode actually used to serve the request. This response value may be different from the value set in the parameter.

For an explicit `flex` or `priority` this is fine (request == billed), which covers the case in the issue. Where it drifts is `service_tier: 'auto'` (also the API default): the event records `'auto'`, which the backend can't map to a price multiplier, even though the request may have been served as `default`/`flex`/etc.

```ts
client.chat.completions.create({ model, messages, service_tier: 'auto' })
// OpenAI serves it as 'flex'; response.service_tier === 'flex'
// but $ai_model_parameters.service_tier === 'auto', so the backend can't map 'auto' to a multiplier
```

Nothing regresses here (an unset tier stays omitted, and explicit tiers are a strict improvement), so this is fine to ship as-is. But if we want full accuracy, would it be worth reading the response tier where it's available (`result.service_tier` on the non-streaming path, and the final chunk's `service_tier` when streaming) and preferring it over the request value? And if we'd rather keep request-capture for now, maybe a short note that `auto`/`default` won't be exact. What do you think?

ioannisj (approved):
Left a minor comment, but could also be a follow-up. Ah, this also needs a changeset item as well

carlos-marchal-ph (approved):

packages/ai/src/utils.ts
marandaneto: will follow up on this

marandaneto: Good point — addressed in follow-up PR #4306. It prefers the latest non-null tier reported by successful OpenAI/Azure OpenAI responses across streaming and non-streaming Chat Completions and Responses API paths, while retaining the requested tier as the fallback when response metadata is unavailable.

PR #4073, opened 2026-07-03
fix(rrweb): skip canvas frame encode when pixels are unchanged

pauldambra:
## Problem

The canvas capture worker (`image-bitmap-data-url-worker`) dedupes unchanged frames by fingerprinting the **encoded** bytes: every captured frame pays the full `convertToBlob` (webp/png encode) plus an `arrayBuffer` copy, and only then is an unchanged frame dropped. For a static canvas at the default 4 fps that's a continuous, pointless image-encode loop for the lifetime of the page — we found this pegging a desktop Electron renderer at most of a core, since WebGL canvases (which can't be mutation-observed) are snapshotted at `canvasFps` regardless of whether they changed.

Blank first frames were even worse: the worker encoded a transparent reference blob *and* the frame itself, just to compare their base64, per canvas size.

## Changes

The worker now fingerprints the raw RGBA pixels (`getImageData`, already drawn to the reusable offscreen canvas) *before* encoding:

- unchanged frame: hash compare, respond `{ id }`, **no encode at all**
- blank first frame: compared against a cached all-zero fingerprint per size, **no encode at all** (the transparent reference blob machinery is gone)
- changed frame: encoded and sent exactly as before

The hash walks 32-bit words with `Math.imul` so fingerprinting a multi-megapixel frame stays cheap, and the reusable 2d context is created with `willReadFrequently: true` since it's now read back every frame. Changed frames pay a small new `getImageData` + hash cost on top of the encode; unchanged frames — the common case for the static canvases this loop hits hardest — drop from a full encode to a readback + hash.

What gets transmitted is byte-for-byte the same protocol: first frames, changed frames, and content-to-blank transitions are still sent; unchanged and blank-first frames are still acknowledged with `{ id }` so the main thread clears `snapshotInProgressMap`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- Written with PostHog Code (Claude) while investigating a battery-drain report in the PostHog Code desktop app: an Activity Monitor sample showed this worker on-CPU for 74% of samples encoding frames of static (parked/hidden) xterm WebGL canvases at 3 fps.
- The test was written first against the unfixed worker to prove the cost: an unchanged frame cost 3 encodes (fails `toHaveBeenCalledTimes(1)`), a blank first frame cost 2 (fails `not.toHaveBeenCalled()`); the behavioral assertions (what gets posted back) passed before and after, pinning the protocol.
- Considered hashing on the main thread instead — rejected: same readback cost, worse thread to spend it on.

---


pauldambra:
* **#4073** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4073?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4073?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,51 +62,60 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
+      const fingerprint = fnv1aHash(pixels.data);
+      const lastFingerprint = lastFingerprintMap.get(id);
 
-      if (lastFingerprintMap.get(id) === fingerprint)
+      if (fingerprint === lastFingerprint) {
         return worker.postMessage({ id }); // unchanged
-      const base64 = encode(arrayBuffer);
-      worker.postMessage({ id, type, base64, displayWidth, displayHeight });
+      }
       lastFingerprintMap.set(id, fingerprint);
```
pauldambra: > [!NOTE]

**[convergent: qa-team/reliability + xp]** 🟠 HIGH

`lastFingerprintMap.set(id, fingerprint)` now runs before the encode, where on `main` it only ran after `convertToBlob` + `encode` succeeded. If `convertToBlob`/`arrayBuffer`/`encode` throws (memory pressure, RangeError on very large buffers), the catch replies `{ id }` with no payload but the fingerprint is already recorded — every subsequent identical frame is treated as "unchanged" and skipped, so a static canvas whose one encode failed is silently missing from the replay until its content next changes (potentially the whole session). Previously the failure was retried on the next frame.

Fix: move the `set` to after the successful `postMessage` in the encode path (keeping the intentional set in the blank-first-frame path, where nothing can throw after it), or delete the map entry in the catch block.

(xp independently flagged the set-before-guard ordering as the one spot a reader pauses on — fixing the failure ordering resolves both.)

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,51 +62,60 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
-          lastFingerprintMap.set(id, fingerprint);
-          return worker.postMessage({ id });
-        }
-        lastFingerprintMap.set(id, fingerprint);
-        worker.postMessage({ id, type, base64, displayWidth, displayHeight });
-        return;
-      }
+      // fingerprint the raw pixels so unchanged frames skip the expensive
+      // encode below entirely, instead of encoding first and deduping after
+      const pixels = reusableCtx!.getImageData(0, 0, width, height);
```
pauldambra: > [!NOTE]

Every frame now pays a `getImageData` readback plus a fresh `width*height*4` ImageData allocation (there is no buffer-reuse API for `getImageData`). For static canvases this is a clear win (readback+hash replaces encode+base64), but for constantly-animating canvases at the fps cap it is strictly additive work and multi-megabyte per-frame garbage (~33 MB/frame for a full-scale 4K canvas), plus a new failure mode where the allocation itself throws for huge canvases and the frame is dropped via the catch. `willReadFrequently: true` and running off the main thread mitigate this — worth acknowledging the animating-canvas regression profile in the changeset/PR description, since this worker's worst case is exactly the always-changing canvas.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,51 +62,60 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
         return worker.postMessage({ id }); // unchanged
-      const base64 = encode(arrayBuffer);
-      worker.postMessage({ id, type, base64, displayWidth, displayHeight });
+      }
       lastFingerprintMap.set(id, fingerprint);
+
+      // a canvas that starts out transparent isn't worth transmitting
+      if (
+        lastFingerprint === undefined &&
+        fingerprint === transparentFingerprint(width, height)
```
pauldambra: > [!NOTE]

**[convergent: paul + qa-team/compatibility + qa-team/data-integrity]** 🟡 MEDIUM

The blank-first-frame semantics narrowed, and the changeset slightly overstates "what gets transmitted is unchanged":

1. Old check: encode the frame and compare base64 against an encoded fully-transparent canvas — so a canvas with alpha-0 but non-zero RGB (e.g. WebGL `clearColor(1,0,0,0)`) that *encodes* to transparent was skipped. New check: raw all-zeros — that same canvas now transmits its first frame. Arguably more correct, but make it a conscious trade, not an accident.
2. Frames whose raw pixels differ but lossy-encode to identical bytes were previously dropped as unchanged; they now transmit.
3. The blank comparison went from exact base64 equality to a 32-bit FNV hash — a ~2^-32 collision now misclassifies real first content as blank, a false-drop class the old exact comparison could not hit.

All three move in the "sends more, never less" direction (modulo the negligible collision). Suggest softening the changeset claim to "no response shapes changed; unchanged-frame detection is equal or stricter" and noting the trade in a comment.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -1,23 +1,38 @@
...
   ImageBitmapDataURLWorkerResponse,
 } from '@posthog/rrweb-types';
 
 const lastFingerprintMap: Map<number, string> = new Map();
-const transparentBlobMap: Map<string, string> = new Map();
+const transparentFingerprintMap: Map<string, string> = new Map();
 
-function fnv1aHash(buffer: ArrayBuffer): string {
-  const view = new Uint8Array(buffer);
+function fnv1aHash(data: Uint8ClampedArray): string {
```
pauldambra: > [!NOTE]

**[qa-team/generalist]** 🟢 LOW

`fnv1aHash` has two latent traps for future callers: `data.byteLength >>> 2` silently ignores 1-3 trailing bytes when the length is not a multiple of 4 (two buffers differing only in the tail hash equal), and the `Uint32Array` constructor throws for a view whose `byteOffset` is not 4-aligned (e.g. a `subarray`). Both are impossible for `ImageData.data` today, but nothing in the signature says "ImageData only" — consider extending the existing comment to state the word-aligned/word-multiple contract. Note also word-wise FNV-1a has weaker per-byte diffusion than the byte-wise variant; adequate for last-frame dedupe, worth knowing if reused. (Separately, the `Math.imul` swap is a quiet correctness *improvement* over the old `(hash * 0x01000193) | 0`, which lost float precision — nice.)

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,51 +62,60 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
-
       if (
         !reusableCanvas ||
         reusableCanvas.width !== width ||
         reusableCanvas.height !== height
       ) {
         reusableCanvas = new OffscreenCanvas(width, height);
-        reusableCtx = reusableCanvas.getContext('2d')!;
+        reusableCtx = reusableCanvas.getContext('2d', {
+          willReadFrequently: true,
```
pauldambra, line 101: > [!NOTE]

**[paul]** 🟢 LOW

`willReadFrequently: true` is exactly right now we `getImageData` every frame — good catch. Lazily asking rather than checking: does forcing the software-backed 2d context have any downside for the `drawImage`/`convertToBlob` path that was previously (potentially) GPU-backed? Assumed a wash or a win for this read-heavy usage — confirm and this resolves.

packages/rrweb/rrweb/test/record/image-bitmap-data-url-worker.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      expect.objectContaining({ id: 1, base64: expect.any(String) }),
+    );
+
+    await onmessage(frame(1, CONTENT_A));
+    expect(postMessage).toHaveBeenLastCalledWith({ id: 1 });
+
+    expect(convertToBlob).toHaveBeenCalledTimes(1);
+  });
+
+  it.each([
```
pauldambra, line 127: > [!NOTE]

**[convergent: paul + xp + qa-team/generalist]** 🟢 LOW

Nicely parameterized, and the `convertToBlob` call-count assertions are the genuine regression guard (they go red against the old encode-then-dedupe code). Two cheap additions worth making:

1. A `BLANK` → `CONTENT_A` case on the same id — exercises the `lastFingerprint === undefined && transparent` skip followed by a real send. The existing changed-cases both start from content, so the transparent-first path only gets the "skips entirely" assertion, never the "then wakes up and sends" half.
2. An assertion that `dataURLOptions` are forwarded to `convertToBlob` — the spy records them but nothing asserts them, so silently dropping `{ type, quality }` would pass.

Also worth a short comment on the fakes: real `getImageData` returns an unpremultiplied *copy* and real `drawImage` composites onto a premultiplied backing store — the fakes deliberately don't model that, so this suite can't catch premultiplication divergence.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments; summary comment has the verdict.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 3 @ 0322f353, nits addressed @ 5f85b242)</sub>

Round 3 re-reviewed the dimension-tagging delta: all findings were NITs, security clean, all reviewers stamp. Two fixed in `5f85b242` (changeset example corrected — a solid-fill dimension swap is a visual no-op, now says "any equal-pixel-count resize"; blank-resize skip pinned with a test, which also guards the two-producer tag contract). Two declined with reasoning: the `${w}x${h}:` tag in two places sits where duplication beats abstraction (extract when a third producer appears), and the hash contract deliberately lives in the `hashPixels` name rather than a comment.

### Key findings (round 3)

- ⚪ **NIT** (paul) — blank-resize corner was the one described behavior without a test → test added
- ⚪ **NIT** (xp + qa-team) — changeset "swap" example misleading → corrected; tag-format duplication → declined per xp's own once-and-only-once analysis
- ⚪ **NIT** (qa-team) — hashPixels comment lost the trailing-bytes caveat → declined, contract is in the name by round-2 design

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Collapse sound, retry contract holds, pixel-count memo keying provably correct; dimension-blind dedupe was the one substantive find. |
| 👤 paul | Round-1 fixes landed well; seed collapse "genuinely nice"; naming-over-comment nudge on the hash helper. |
| 📐 xp | Once-and-only-once win, tests pin the collapsed behavior through the public interface. |
| 🛡 security-audit | No findings — same-origin boundary, fingerprint never leaves worker memory, no new reachability. |

<details>
<summary>Previous rounds (2)</summary>

round 1 @ ede08221 — ⚠️ REQUEST CHANGES: HIGH set-before-encode ordering (fixed in 9360d693 with red-first retry test), changeset honesty, documented trades, test gaps — all addressed.
round 2 @ e5a94c4b — ✅ APPROVE: MEDIUM dimension-blind fingerprint (fixed in ea77fdf4 with red-first resize test), blank-resize drift accepted, comment/naming nits fixed.

</details>

---

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Fixed in `9360d693` — `lastFingerprintMap.set` now runs only after the successful `postMessage` in the encode path (the blank-first-frame path keeps its set, where nothing can throw after it). Also added a regression test ("retries the encode on the next frame after a transient failure") and verified it fails against the pre-fix ordering and passes with it.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Acknowledged in `9360d693` — the changeset now states the trade explicitly: always-animating canvases do slightly more work per frame (readback + hash on top of the encode), while static canvases drop from a full encode per tick to a readback + hash. Not adding buffer pooling now: there is no reuse API for `getImageData`, and the allocation is off-main-thread; if profiling ever shows GC pressure here, an OffscreenCanvas-scale-down before readback is the follow-up.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Fixed in `9360d693` — the changeset claim is softened to "no response shapes changed... sends more, never less", and the transparent check now carries a comment stating the narrowing (raw all-zeros vs the old encodes-to-transparent) is deliberate. The 2^-32 collision class is accepted: same 32-bit fnv1a family as before, cost of a collision is one suppressed first frame.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Fixed in `9360d693` — the `fnv1aHash` comment now states the word-aligned / word-multiple contract (trailing bytes would be silently ignored). Leaving the signature as-is since the worker is the only caller.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Confirmed reasoning: with a `getImageData` per frame, an accelerated canvas would force a GPU sync + readback every tick — `willReadFrequently` keeps the raster CPU-side. The costs on the other paths are unchanged in kind: `drawImage(ImageBitmap)` becomes a CPU copy (the GPU-to-CPU transfer happens once at draw instead of at read), and `convertToBlob` encoding is CPU-bound regardless of backing. For a read-per-draw workload like this one it is a wash-or-win, which is exactly the case the hint exists for.

packages/rrweb/rrweb/test/record/image-bitmap-data-url-worker.test.ts
pauldambra: > [!NOTE]

Fixed in `9360d693` — added the `BLANK → CONTENT_A` case to the parameterized changes ("content after a blank first frame"), an assertion that `dataURLOptions` (`{ type, quality }`) are forwarded to `convertToBlob`, and a comment on the fake stating which real canvas semantics (unpremultiplied copy, source-over compositing) it deliberately does not model.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `9360d693` — `lastFingerprintMap.set` now runs only after the successful `postMessage` (the blank-first-frame path keeps its earlier set, where nothing can throw afterwards). A regression test covers the transient-failure retry, verified red against the pre-fix ordering. (Same finding was raised convergently by the QA Swarm review.)

packages/rrweb/rrweb/test/record/image-bitmap-data-url-worker.test.ts
```diff
@@ -0,0 +1,144 @@
...
+      expect(postMessage).toHaveBeenLastCalledWith(
+        expect.objectContaining({
+          id: 1,
+          base64: expect.any(String),
+          displayWidth: 4,
+          displayHeight: 4,
+        }),
+      );
+    },
+  );
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Added in `9360d693` as a third case in the parameterized block ("content after a blank first frame", `BLANK → CONTENT_A`). (Same gap was flagged convergently by two QA Swarm reviewers.)

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,50 +66,61 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
-        worker.postMessage({ id, type, base64, displayWidth, displayHeight });
-        return;
+      // fingerprint the raw pixels so unchanged frames skip the expensive
+      // encode below entirely, instead of encoding first and deduping after.
+      // an unseen canvas is compared against the transparent fingerprint, so
+      // one that starts out blank is skipped like an unchanged frame; raw
+      // all-zero pixels is deliberately narrower than the old "encodes to the
+      // same bytes as a transparent canvas" check (e.g. alpha-0 pixels with
+      // non-zero RGB now transmit) — we send more, never less
+      const fingerprint = fnv1aHash(
```
pauldambra: > [!NOTE]

**[convergent: qa-team/data-integrity + qa-team/generalist]** 🟡 MEDIUM

The raw-pixel fingerprint is dimension-blind, where the old encoded-blob fingerprint was not (webp/png headers embed width/height). A canvas resized with preserved pixel count and byte-identical RGBA content — realistically a solid/uniform-fill canvas whose width and height swap (e.g. 100x200 -> 200x100 on orientation change) — produces the same hash as the stale entry, so the frame is skipped and the replayer keeps the old-dimension snapshot indefinitely. Silent wrong replay visuals.

Fix: make the fingerprint dimension-aware (e.g. prefix `${width}x${height}:` onto both the content fingerprint and the transparent seed), and add a resize test — the fake canvas currently only ever runs at one size, so no existing test can catch this.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,50 +66,61 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
+        reusableCtx!.getImageData(0, 0, width, height).data,
+      );
+      const lastFingerprint =
+        lastFingerprintMap.get(id) ?? transparentFingerprint(width, height);
```
pauldambra, line 120: > [!NOTE]

Behavior drift from the seed collapse: a canvas that has never shown content and is resized while blank is now skipped at the new size (no map entry, so it compares against the transparent fingerprint at the new pixel count), whereas both the original code and the round-1 special-case version transmitted a blank frame after such a resize. Painting nothing for a still-blank canvas is arguably better, but it makes the changeset's "sends more, never less" claim technically false in this corner. Accept explicitly (soften the changeset wording) or note it in the code comment.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
```diff
@@ -67,50 +66,61 @@ let reusableCtx: OffscreenCanvasRenderingContext2D | null = null;
...
-        if ((await transparentBase64) === base64) {
-          lastFingerprintMap.set(id, fingerprint);
-          return worker.postMessage({ id });
-        }
-        lastFingerprintMap.set(id, fingerprint);
-        worker.postMessage({ id, type, base64, displayWidth, displayHeight });
-        return;
+      // fingerprint the raw pixels so unchanged frames skip the expensive
+      // encode below entirely, instead of encoding first and deduping after.
+      // an unseen canvas is compared against the transparent fingerprint, so
```
pauldambra, line 112: > [!NOTE]

**[convergent: qa-team/generalist + paul + xp]** ⚪ NIT

Three comment/naming nits that collapse into one edit:

1. The example "alpha-0 pixels with non-zero RGB now transmit" is factually wrong: `drawImage` premultiplies, zeroing RGB wherever alpha is 0, so `getImageData` on the reusable canvas reads all-zero and such frames are still skipped — same as before. The true "sends more" example is content that lossy-encodes to the same bytes as the transparent reference.
2. (xp) The 8-line block largely restates the changeset — trim it to the code-local surprising facts.
3. (paul) `fnv1aHash`'s word-alignment contract would be better carried by the name than the comment — `hashPixels` or similar. Naming over comment.

pauldambra (commented):
> [!NOTE]

QA Swarm round 2 complete (re-review after round-1 fixes + the seed-collapse refactor). See inline comments; summary comment updated.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Fixed in `ea77fdf4` — fingerprints are now dimension-tagged via a `frameFingerprint(width, height, data)` helper (`${width}x${height}:${hash}`), with the transparent seed carrying the same tag, so any resize retransmits even when raw bytes are identical. Added the resize regression test ("re-sends a resized canvas even when raw pixels are byte-identical") and verified it fails against the untagged version.

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Accepted explicitly in `ea77fdf4` — the changeset now names this exact corner: "a canvas that resizes while never having shown content stays skipped — it is blank at either size." With dimension-tagged fingerprints this is the only remaining sends-less case, and it is deliberate (a blank frame paints nothing).

packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts
pauldambra: > [!NOTE]

Fixed in `ea77fdf4` — the factually wrong alpha-0/non-zero-RGB example is gone (the changeset now carries the correct lossy-encode example), the comment block is trimmed to the code-local facts, and `fnv1aHash` is renamed to `hashPixels` so the contract lives in the name.

pauldambra:
@sortafreel one for review hog out of interest

arnohillen (approved):

PR #4077, opened 2026-07-04
fix(web): stop retrying log batches forever on status-0 failures

pauldambra:
## Problem

When a browser extension (ad blocker, privacy tool) blocks the `/i/v1/logs` endpoint, requests die before any HTTP response and surface as `statusCode: 0`. The logs pipeline classifies that as `retry-later`, so it buffers records and retries for the life of the page — futile network traffic every backoff window (up to ~3.2 min ceiling), forever, since a filter-list block is deterministic.

The event capture pipeline already handles this: its retry queue caps status-0 failures at `STATUS_CODE_ZERO_MAX_RETRIES = 3` instead of the default 10.

## Change

Give the logs pipeline the same status-0 budget, as a circuit breaker in the browser host (`posthog-logs.ts`) — the only layer that sees the status code:

- After **3 consecutive** status-0 failures while `navigator.onLine` is true, `_sendLogsBatch` short-circuits to a `fatal` outcome without touching the network, so core drops each batch instead of buffering indefinitely.
- Any real HTTP response (even 429/5xx) resets the count — a response proves the endpoint is reachable.
- Failures while the browser reports itself **offline don't count**: those records still queue for the existing `online` reconnect flush, preserving the offline story.
- The `online` event reopens the breaker, so whitelisting + connectivity change recovers.
- The breaker is shared across the programmatic and console cores (same endpoint, one verdict), and warns once with an ad-blocker-mentioning message for supportability.

The short-circuit (rather than only reclassifying to `fatal`) matters: `fatal` doesn't throw in core's flush loop the same way `retry-later` does, so without it the backoff would collapse to the 3s base interval and *increase* blocked traffic.

Deliberately unchanged: the `sendBeacon` unload path stays ungated — fire-and-forget, same as events on unload. The 90s send-timeout backstop also doesn't count toward the trip (not an ad-blocker signature).

## Tests

TDD: 8 new tests (11 cases with parameterization) in `posthog-logs.test.ts` covering the trip threshold, per-status reset, offline exclusion, `online` reopening, console-core sharing, single warning, and backstop exclusion. Full browser suite green (4825 tests), typecheck/lint/prettier clean. Changeset included (patch).

---


packages/browser/src/posthog-logs.ts
```diff
@@ -237,6 +248,13 @@ export class PostHogLogs implements Extension {
     private _sendLogsBatch(payload: OtlpLogsPayload): Promise<SendLogsBatchOutcome> {
         // eslint-disable-next-line compat/compat
         return new Promise((resolve) => {
+            if (this._consecutiveStatusZeroFailures >= MAX_CONSECUTIVE_STATUS_ZERO_FAILURES) {
```
pauldambra: > [!NOTE]

The tripped-breaker short-circuit fires regardless of `navigator.onLine`, so logs captured during a genuine offline period *after* the breaker has tripped are dropped via `fatal` instead of queueing for the reconnect flush. The changeset claims "genuine offline periods still queue for the reconnect flush", but that only holds pre-trip. Consider gating the short-circuit: `if (this._consecutiveStatusZeroFailures >= MAX && window?.navigator.onLine !== false)`. While offline the real send already returns status 0 without counting (per `_trackEndpointReachability`) and settles `retry-later`, so this preserves offline queueing at no extra network cost while online-and-blocked behavior is unchanged.

packages/browser/src/posthog-logs.ts
```diff
@@ -237,6 +248,13 @@ export class PostHogLogs implements Extension {
...
+                // Tripped: drop the batch without touching the network. `fatal`
+                // advances the queue so records don't pile up while blocked.
+                resolve({ kind: 'fatal', error: new Error('logs endpoint is unreachable, dropping batch') })
```
pauldambra, line 265: > [!NOTE]

`_flushViaTransport` (the `flushLogs('sendBeacon')` unload path and explicit-transport flushes) bypasses the tripped breaker and still fires the request, and its callback-less `_send_request` never feeds `_trackEndpointReachability` either. If this is a deliberate "last gasp" attempt at unload (sendBeacon is sometimes treated differently by blockers), a comment saying so would prevent someone later "fixing" it inconsistently; otherwise add the same breaker check there.

packages/browser/src/posthog-logs.ts
```diff
@@ -283,6 +302,25 @@ export class PostHogLogs implements Extension {
         })
     }
 
+    // Feeds the status-0 circuit breaker checked at the top of `_sendLogsBatch`.
+    private _trackEndpointReachability(statusCode: number): void {
+        if (statusCode === 0) {
+            // `onLine === false` is genuine offline: those records wait for the
+            // reconnect flush, so they don't count toward the trip.
+            if (window?.navigator.onLine !== false) {
```
pauldambra: > [!NOTE]

**[convergent: qa-team/compatibility + paul + xp]** 🟢 LOW

`window?.navigator.onLine !== false` treats a missing `window` as "online" (`undefined !== false` is truthy), so status-0 failures count toward the trip — but the constructor only attaches the `online` reopener `if (window)`. In a no-`window` environment the breaker can trip and never reopen, silently dropping logs for the life of the process. The retry-queue neighbour guards with `'onLine' in window.navigator` before trusting the value. Worth making the two `window` checks encode the same assumption (e.g. if there is no window, don't count status-0 toward the trip either).

packages/browser/src/posthog-logs.ts
```diff
@@ -283,6 +302,25 @@ export class PostHogLogs implements Extension {
...
+                this._consecutiveStatusZeroFailures++
+                if (this._consecutiveStatusZeroFailures === MAX_CONSECUTIVE_STATUS_ZERO_FAILURES) {
+                    this._logger.warn(
+                        'Log requests are failing before receiving an HTTP response; this can happen due to network issues, CORS, browser blocking, or ad blockers. Stopped sending logs; will try again when connectivity changes.'
```
pauldambra: > [!NOTE]

**[paul]** 🟡 MEDIUM

measure-before-you-limit instinct kicking in here. this is a path that _drops_ data, and every dropped batch is a log line someone won't see. we warn once when the breaker trips, which is good, but there's nothing that tells us in aggregate how often this fires in prod. i wonder if we want a counter (even just a captured metric / internal event) for "logs breaker tripped" and "batches dropped while tripped" so that if we've got the condition wrong we find out from a graph rather than from a customer saying "where did my logs go". not blocking the mechanism, but i'd sleep better shipping the drop with a way to see it.

pauldambra (commented):
> [!NOTE]

packages/browser/src/posthog-logs.ts
```diff
@@ -26,6 +26,12 @@ const CONSOLE_SCOPE_NAME = 'console'
 // settles via its callback first; this only fires on a genuinely callback-less
 // send (e.g. request enqueued before load, or a transport that never reports).
 const LOGS_SEND_TIMEOUT_MS = 90000
+// Mirrors the event retry queue's status-0 budget: a request that dies before any
+// HTTP response while the browser reports itself online is almost always
+// deterministically blocked (ad blocker, CORS, extension), so retrying forever
+// only burns network. After this many consecutive such failures we stop sending
+// and drop batches; the `online` event reopens the pipe.
+const MAX_CONSECUTIVE_STATUS_ZERO_FAILURES = 3
```
pauldambra, line 37: > [!NOTE]

**[convergent: xp + paul]** 🟡 MEDIUM

This is the second home for the same idea: `retry-queue.ts` already encodes "a status-0 request while online is a deterministic block; stop after N and warn about ad blockers/CORS" (`STATUS_CODE_ZERO_MAX_RETRIES` + a near-identical warn string). That's ThreeStrikesAndYouRefactor at strike two — extracting now would be premature (the mechanisms genuinely differ: retry-queue counts per-request retries, this counts consecutive batch failures across a shared endpoint — so "mirrors" in the comment is doing some load-bearing work), but a cross-reference comment in both places would stop the "3" and the warning copy silently drifting apart when someone edits one and not the other.

packages/browser/src/posthog-logs.ts
```diff
@@ -283,6 +302,25 @@ export class PostHogLogs implements Extension {
...
+            if (window?.navigator.onLine !== false) {
+                this._consecutiveStatusZeroFailures++
+                if (this._consecutiveStatusZeroFailures === MAX_CONSECUTIVE_STATUS_ZERO_FAILURES) {
+                    this._logger.warn(
+                        'Log requests are failing before receiving an HTTP response; this can happen due to network issues, CORS, browser blocking, or ad blockers. Stopped sending logs; will try again when connectivity changes.'
+                    )
+                }
+            }
+        } else {
+            // Any HTTP response proves the endpoint is reachable.
```
pauldambra: > [!NOTE]

**[xp]** ⚪ NIT

`_trackEndpointReachability` is a nicely named ComposedMethod — the callback stays readable. One nit: this else-branch comment ("Any HTTP response proves the endpoint is reachable") duplicates what `statusCode !== 0` plus the well-named method already say; the counter reset is self-explanatory. Consider dropping it. (The offline/`onLine === false` comment above is load-bearing and should stay.)

packages/browser/src/__tests__/posthog-logs.test.ts
```diff
@@ -1140,6 +1140,127 @@ describe('posthog-logs', () => {
...
+                    await flushWith(0)
+                }
+                Object.defineProperty(window.navigator, 'onLine', { value: true, configurable: true })
+
+                await flushWith(0)
+
+                expect(sendCount()).toBe(4)
+            })
+
+            it('reopens on the online event so recovery is possible', async () => {
```
pauldambra, line 1262: > [!NOTE]

**[xp]** 🟢 LOW

Nice TDD work — the parameterization is real (the `[1,2]` boundary, the `[200,429,503]` "any response resets" axis, the offline case). One gap: this reopen test asserts a request fires after the `online` event, but not that the counter actually reset to 0 — i.e. that a subsequent run of status-0 failures has to climb from scratch before re-tripping. Currently "reset to 0" is inferred, not pinned. Consider one assertion that after reopening, it takes 3 fresh failures to trip again.

pauldambra:
> [!NOTE]
>

## Verdict: ✅ APPROVE <sub>(round 4 @ bab84cc)</sub>

Quality loop converged. All actionable findings across four rounds are fixed and verified (the final round traced the exact-count assertions through core's flush loop and confirmed them); the only open items are two deliberately deferred design questions for a human (aggregate drop-observability metric; pre-init `fireCallbackOnDrop` interaction with the breaker).

### Key findings

- Round 3-4: two test NITs (comment accuracy, loose `toBeGreaterThan` assertions) — fixed in `bab84cc` and verified exact (queue holds 4 retained records; reconnect makes exactly 1 send).
- Deferred to human: 🟡 drop-observability metric (paul, design decision); 🟡 pre-init drops possibly feeding the breaker (codex bot, architectural).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | Delta traced correct end-to-end incl. core flush-loop and exact batch counts; regression test mutation-verified. APPROVE. |
| 👤 paul | All asks landed; `setOnline` helper and exact assertions welcomed; happy to stamp. |
| 📐 xp | OnceAndOnlyOnce wins throughout; tests now document the exact contract; ship it. |
| 🛡 security-audit | Clean both rounds — client-side only, no new sinks, strictly loss-reducing. |

<details>
<summary>Previous rounds (3)</summary>

round 1 @ 710953b — 💬 APPROVE WITH NITS: 3 MEDIUM + 3 LOW + 1 NIT; 5 fixed in 94a59b8/4e04a5c, 2 deferred to a human.
round 2 @ 4e04a5c — 💬 APPROVE WITH NITS: offline-gate fix unpinned by tests (mutation-verified); regression test added in 4bc32b8.
round 3 @ dd018b0 — ✅ APPROVE: NITs only (comment wording, loose assertions); fixed in bab84cc.

</details>

---

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Agreed — removed the comment in 94a59b82. The method name and the `statusCode !== 0` check already convey the intent.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Fixed in 94a59b82 — gated the short-circuit in `_sendLogsBatch` on `window?.navigator.onLine !== false` so genuine offline batches queue for the reconnect flush instead of being fatally dropped after the breaker trips.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Clarified in 94a59b82 — added a comment to `_drainQueueViaTransport` noting that the circuit-breaker bypass is intentional: this is a best-effort last-gasp send path (especially sendBeacon at unload) where the callback-less transport means we can't track outcomes anyway.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Fixed in 94a59b82 — changed `_trackEndpointReachability` to gate on `window &&` before testing `onLine`, aligning with the constructor which only attaches the `online` listener when `window` is present. In a no-window environment the breaker can no longer trip (the `online` event would never fire to reopen it anyway).

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Added cross-reference comments in 94a59b82 — both `MAX_CONSECUTIVE_STATUS_ZERO_FAILURES` in posthog-logs.ts and `STATUS_CODE_ZERO_MAX_RETRIES` in retry-queue.ts now point at each other with a note to keep the constant value and warning copy in sync.

packages/browser/src/__tests__/posthog-logs.test.ts
pauldambra: > [!NOTE]

Added in 94a59b82 — new test `counter resets to 0 on reconnect — needs 3 fresh failures to trip again` pins that after the online event resets the counter, the breaker requires a full 3 fresh status-0 failures before re-tripping, and the 4th attempt is dropped.

packages/browser/src/posthog-logs.ts
```diff
@@ -237,6 +250,18 @@ export class PostHogLogs implements Extension {
     private _sendLogsBatch(payload: OtlpLogsPayload): Promise<SendLogsBatchOutcome> {
         // eslint-disable-next-line compat/compat
         return new Promise((resolve) => {
+            if (
+                this._consecutiveStatusZeroFailures >= MAX_CONSECUTIVE_STATUS_ZERO_FAILURES &&
```
pauldambra: > [!NOTE]

The round-2 fix — gating this tripped-breaker short-circuit on `_isBrowserOnline()` so offline captures queue instead of being fatally dropped — has no test coverage. Mutation-verified: removing `&& this._isBrowserOnline()` from this condition passes the entire suite. The existing offline test only covers not-*counting* while offline, never the tripped-then-offline drop path. Add a test: trip the breaker (3 status-0 failures + 1 drop), set `navigator.onLine = false`, capture + flush with status 0, assert the send was attempted and the queue retains the records (retry-later, not fatal); then dispatch `online` and assert delivery. Without it a future refactor can silently reintroduce the offline data-loss bug this round fixed.

packages/browser/src/__tests__/posthog-logs.test.ts
```diff
@@ -1140,6 +1140,160 @@ describe('posthog-logs', () => {
...
+                // Reset the breaker via the online event.
+                assignableWindow.dispatchEvent(new Event('online'))
+
+                // Verify the counter was actually reset to 0.
+                expect((logs as any)._consecutiveStatusZeroFailures).toBe(0)
+
+                // The online event schedules a reconnect flush (empty queue, no send).
+                // The first flushWith after online resolves that lingering flush promise,
+                // so the second and third explicit flushes are the real first two failures.
+                await flushWith(0) // drains lingering online-reconnect flush promise
```
pauldambra, line 1292: > [!NOTE]

**[convergent: qa-team + paul + xp]** ⚪ NIT

This counter-reset test reaches into the private `_consecutiveStatusZeroFailures` and depends on subtle promise choreography (the first post-`online` `flushWith(0)` draining the lingering reconnect flush). The three-line comment is honest and load-bearing, and the test earns its white-box peek by proving the distinct "3 *fresh* failures to re-trip" semantics its black-box sibling can't — but the coupling means a future refactor of the reconnect-flush plumbing breaks this test one step off from its intent. Consider draining explicitly (e.g. `await (logs as any)._core.flush().catch(() => {})` right after dispatching `online`) so the three subsequent `flushWith(0)` calls unambiguously mean failures 1-3. Fine to ship as-is.

pauldambra (commented):
> [!NOTE]

QA Swarm round-2 re-review complete (delta since 710953b). See inline comments.

packages/browser/src/__tests__/posthog-logs.test.ts
pauldambra: > [!NOTE]

Acknowledged — the three-line comment is intentional and the white-box peek earns its place by proving the distinct "3 fresh failures to re-trip" semantics. Shipping as-is.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Fixed in `4bc32b8c` — added the regression test in `posthog-logs.test.ts`. The test trips the breaker (3 status-0 failures), goes offline, captures + flushes with status 0, asserts the send was attempted (not short-circuited) and the queue retains records, then dispatches `online` and asserts delivery. Mutation-verified: removing `&& this._isBrowserOnline()` from the guard causes the test to fail.

packages/browser/src/posthog-logs.ts
```diff
@@ -264,6 +282,7 @@ export class PostHogLogs implements Extension {
                 fireCallbackOnDrop: true,
                 callback: (response) => {
                     const status = response.statusCode
+                    this._trackEndpointReachability(status)
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `ac745595` — counting is now gated on `posthog.__loaded`, so the synthesized `{ statusCode: 0 }` from the pre-init `fireCallbackOnDrop` path never feeds the breaker. `__loaded` flips when `init` completes rather than on a successful request, so an ad-blocked-from-the-start page still trips as intended (no "must succeed once" catch-22). Covered by a red-first test: 3 pre-init drops, then load, then assert the next flush still attempts the network.

packages/browser/src/posthog-logs.ts
pauldambra: > [!NOTE]

Author's decision: no aggregate metric needed. The one-time console warn is the intended level of observability for this drop path.

packages/browser/src/posthog-logs.ts
```diff
@@ -283,6 +309,30 @@ export class PostHogLogs implements Extension {
         })
     }
 
+    // Returns true when the browser reports itself online (or the API is absent).
+    // False means the browser is explicitly offline — genuine network loss, not a
+    // blocker.  No `window` means no `online` event listener either (constructor),
+    // so the circuit breaker would never reopen; treat as indeterminate (false).
+    private _isBrowserOnline(): boolean {
+        return !!(window && window.navigator.onLine !== false)
```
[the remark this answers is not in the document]
pauldambra: > [!NOTE]

Fixed in `7083da24` — reworded so the two cases are explicit: missing `navigator.onLine` counts as online, but missing `window` returns false (no `online` listener means a tripped breaker could never reopen).

frankh (approved):

marandaneto (approved):

marandaneto:
Addressed the reset-path feedback in ba6432c52: `PostHogLogs.reset()` now clears the status-0 breaker counter, with a regression test proving a tripped breaker can send again immediately after reset.

PR #4081, opened 2026-07-06
fix(browser): drop redundant ver/compression params from capture URL

rafaeelaudibert:
The capture endpoint already recovers both without the query params: $lib_version is in the event payload, and gzip is self-describing via its magic-byte header (the endpoint sniffs it, plus base64 on legacy paths). Dropping them from the URL removes data that isn't needed and keeps the request simpler. Base64 still sends the compression hint, which non-legacy endpoints such as session replay require.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/.changeset/drop-capture-url-ver-compression.md
+++ b/.changeset/drop-capture-url-ver-compression.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: remove redundant `ver` and gzip `compression` query params from capture requests. `$lib_version` is already included in the event payload, and gzip is self-describing via its magic-byte header, so neither is required for ingestion — dropping them simplifies the request URL. Non-gzip codecs (base64) still send the `compression` hint, which non-legacy endpoints such as session replay require.
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -100,7 +100,7 @@ describe('request', () => {
             )
             expect(mockedXHR.open).toHaveBeenCalledWith(
                 'GET',
-                'https://any.posthog-instance.com/?_=1700000000000&ver=1.23.45',
+                'https://any.posthog-instance.com/?_=1700000000000',
                 true
             )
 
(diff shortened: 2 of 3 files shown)


rafaeelaudibert:
* **#4082** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4082?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#4081** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4081?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4081?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

marandaneto:
blocking: we cant remove `ver` and `compression` from all endpoints since its still used in some cases

Endpoint-by-endpoint:

 - /e, /batch, /i/v0/e, /s recordings
     - gzip compression query param: not needed; backend detects gzip by magic bytes.
     - ver: not functionally needed for ingestion.

 - /flags/?v=2
     - compression=base64: needed/expected for posthog-js flags, and PR keeps it.
     - gzip param is not relevant for posthog-js flags.
     - ver: backend uses it for canonical flag request logs (lib_version). Removing it likely won’t break flag
       evaluation, but loses browser SDK version in flag logs.

 - /i/v1/logs
     - gzip compression query param: needed today.
     - Backend has explicit middleware:
       rust/capture-logs/src/middleware.rs
       It translates ?compression=gzip-js into Content-Encoding: gzip so RequestDecompressionLayer can decode the body.
     - Without the query param, gzipped logs from posthog-js would reach the handler still compressed and fail
       JSON/protobuf parsing.
     - ver: ignored there.

PR as-is is too broad because buildRequestURL() affects logs too. Safer fix: only omit gzip compression and
 ver for capture/recording endpoints handled by rust/capture (/e, /batch, /i/v0/e, /s, etc.), but keep gzip compression
 for /i/v1/logs, and keep ver for /flags if we care about flag request SDK-version logging.

rafaeelaudibert:
@marandaneto Thank you! I checked the ingestion pipeline and couldn't find any references but I did miss the new `/logs` endpoint.

Is a better approach to simply make logs work like normal ingestion and then this can be done? Or would you rather we not bother about this?

marandaneto:
> @marandaneto Thank you! I checked the ingestion pipeline and couldn't find any references but I did miss the new `/logs` endpoint.
> 
> Is a better approach to simply make logs work like normal ingestion and then this can be done? Or would you rather we not bother about this?

yeah i think if all endpoints add support for detecting gzip w/o any hint, and flags, we could remove it
but maybe we cannot completely remove it to keep back compatibility, didnt check the details
rn compression and ver arent sensitive and arent dropping events via adblockers so not urgent rn
i did this btw https://github.com/PostHog/posthog/pull/68497 since capture does not use the ver query param, just flags

rafaeelaudibert:
Yeah, I agree it's not dropping events and it's not urgent.

I'll use this as learning, and just close this one. I'll let the logs team know they should probs change their implementation, and then we can rethink this, ty :)

PR #4085, opened 2026-07-06
fix(web): ignore unselecting text in dead-click detection

### LINKED ISSUE #1552: dead click detection should ignore _unselecting_ text
see https://x.com/AdrienBrault/status/1861177509901148615

if you click to select text we'd ignore that as your click would be _followed_ by a selection change

But if you click to unselect text we would count that as a dead click... 

A click that very closely follows a selection change probably isn't a dead click either

apoorva-01:
## Problem

Dead-click detection already ignores a click that *selects* text, because the `selectionchange` fires just after the click. But clicking to *unselect* text fires `selectionchange` just before the click, and that case slipped through, so unselecting text got logged as a dead click.

Closes #1552.

## Changes

The selection check only looked forward (`click.timestamp <= this._lastSelectionChanged`), so it caught selecting text but not unselecting it. Switched it to `Math.abs`, same as the visibility check right below. Now a selection change just before or just after the click both count, which also covers the "click right after a selection" case from the issue.

Left the threshold alone and reused `selection_change_threshold_ms` for both directions, like visibility does. Happy to split the before/after windows if you'd rather tune them separately, but matching visibility seemed cleaner.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Added a changeset (`.changeset/dead-click-ignore-unselecting-text.md`, `posthog-js` patch)


dustinbyrne (approved):
Hey @apoorva-01, thanks for the PR. This makes sense to me!

Issue #698, opened 2023-06-13
Issue when compiling typescript app


jondrummond-codeo:
We get the following error when compiling our app typescript using posthog-js.

```
node_modules/posthog-js/dist/module.d.ts:1:52 - error TS2307: Cannot find module 'types' or its corresponding type declarations.

1 import { SurveyCallback as SurveyCallback$1 } from 'types';
```

PashaStrayt:
Same problem.
I've install posthog-js@1.57.2 after this, it works for me

sdenardi:
Confirmed also getting this error with `1.68.2`.

pauldambra:
Hey all, sorry about this. We use posthog-js as a node module for building app.posthog.com which is working for us right now. I assume there's some difference between our environment and yours causing the problem.

Could you share a small reproduction or the versions of typescript and other tools you're using?

pauldambra:
So, for example, this isn't a "minimal" example really 🤣 but it's a typescript nextjs app which can compile with posthog present https://github.com/pauldambra/test-posthog-import

pauldambra:
From #712 

```
How to reproduce
install "posthog-js": "1.68.2"
npx tsc --noemit
Environment
"posthog-js": "1.68.2",
"typescript": "4.7.4",
```

pauldambra:
@jeremytenjo could you expand on your reproduction? E.g. posthog-js assumes window is available so it can't be used in a completely empty TS project.

I'm not sure what assumption we're making that means it works for us and others but is breaking it for you all 😓

pauldambra:
Ah, this comment https://github.com/PostHog/posthog/issues/16188#issuecomment-1604391443 💖 

I've reduced https://github.com/pauldambra/test-posthog-import to a minimal example now.

You see the error or not by switching between `skipLibCheck` true and false. 

`skipLibCheck: true` is the default for an empty TS project and does not error.

Switching to `skipLibCheck: false` demonstrates the error

pauldambra:
For context if I switch the github.com/posthog/posthog repo to `skipLibCheck:false` we get 200 errors including from React 🤔

pauldambra:
I forgot automation would close this 🙈 

I'll keep it open until someone else confirms this solves for them. Posthog-js 1.68.5 doesn't exhibit this error for me. 

If someone else can check and confirm that would be awesome 💖

jondrummond-codeo:
[...]

PR #4090, opened 2026-07-06
fix(react-native): repeating surveys show again when a new iteration starts

lucasheriques:
## Problem

Repeating surveys ("Repeat on a schedule") never repeat on React Native. The local seen-state is keyed by bare survey ID, so once a user answers or dismisses a survey on a device it stays hidden forever, even after the backend opens the next iteration. Web, iOS, and Android all key seen-state by iteration.

The bug existed because RN and web each maintain their own copy of this logic, and the copies drifted. Sharing it was blocked by a type mismatch: core uses TS enums, the browser package deliberately uses const-object literal unions, and TS enums don't accept equal string literals.

Found while auditing repeating-survey behavior across SDKs ([Slack thread](https://posthog.slack.com/archives/C07QD3LT8U9/p1783335567251989?thread_ts=1783335567.251989&cid=C07QD3LT8U9)). Companion PRs: [posthog#68735](https://github.com/PostHog/posthog/pull/68735) (backend), [posthog.com#18200](https://github.com/PostHog/posthog.com/pull/18200) (docs).

## Changes

1. **Fix: RN seen-state is now iteration-aware.** `id_iteration` when `current_iteration > 0`, bare ID otherwise, matching the other SDKs. Applied on both the read path (`getActiveMatchingSurveys`) and the write path (survey close/dismiss). Marking a survey seen also purges its stale keys from earlier iterations, so each survey holds exactly one slot in the bounded 20-entry list. Backward compatible: previously stored bare IDs simply stop matching a repeating survey's key, which correctly makes it eligible again. We deliberately do NOT migrate legacy bare IDs to the current iteration key: the old SDK never re-showed a survey after one answer, so a legacy bare ID almost always represents an answer from an old iteration — migrating it would wrongly suppress the current iteration for those users and delay this fix by a full cycle. The cost of not migrating is bounded to a one-time re-show for the rare user who answered the still-current iteration right before upgrading.

2. **The shared logic now lives in `@posthog/core/surveys`** (following the existing pattern there): `getSurveyIterationKey`, `doesSurveyActivateByEvent`, and `canSurveyActivateRepeatedly`. Web keeps its localStorage prefixes and its in-progress check as thin local layers; RN imports the core names directly. This is what prevents the two SDKs from drifting apart again.

3. **Core's survey enums are now const-object literal unions**, the pattern the browser package already uses. This removes the type mismatch that forced hand-rolled parameter types in shared helpers: they now use `Pick<Survey, ...>` from core's `Survey`, whose `current_iteration`/`schedule` are also nullable now, matching the API. Value usage (`SurveySchedule.Always`) and type annotations are unchanged for all consumers; runtime output is identical.

Deliberately **not** in this PR: unifying the browser's `posthog-surveys-types.ts` onto core's types. That's public API with real field drift and a known declaration-output landmine (#698), so it's a follow-up PR. This PR removes the blocker for it.

### Libraries affected

- [x] posthog-js (web)
- [x] posthog-react-native
- [x] @posthog/core

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Changesets added (all patch): `tidy-hogs-repeat.md` (RN fix), `shared-survey-state.md` (core/web refactor)

- New RN cases in `getActiveMatchingSurveys.spec.ts`: seen in a previous iteration → shows again; seen in the current iteration → stays hidden; bare pre-iteration key → shows again. The first fails against the old code.
- New `useSurveyStorage.spec.ts`: seen-list keeps one slot per survey and caps at 20 (guards the eviction edge).
- New core `keys.spec.ts`: `getSurveyIterationKey` for iterations 2/1/0/null/undefined, one canonical suite instead of per-SDK copies.
- Refactor and enum conversion covered by existing suites: core 703, RN 497 (full suites), browser surveys 307. Production-file typechecks clean in all three packages.
- Not done: manual device test on RN.

## 🤖 Agent context

Authored with PostHog Code (Claude) directed by Lucas, from a cross-repo audit of repeating-survey behavior. Iterated through a multi-agent review swarm (13 low-severity findings, all addressed, including the seen-list eviction edge) and Greptile comments (test parameterization, direct key coverage). The enum-to-const-object conversion was chosen over template-literal type workarounds to fix the sharing blocker at the root; full browser type unification was scoped out to a follow-up on purpose.

---

### DIFF
--- a/.changeset/shared-survey-state.md
+++ b/.changeset/shared-survey-state.md
@@ -3,4 +3,4 @@
 'posthog-js': patch
 ---
 
-chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. No behavior change.
+chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. Core's survey enums are now const-object literal unions (matching the web SDK's existing pattern), so the same values type-check across both SDKs. No behavior change.
--- a/.changeset/shared-survey-state.md
+++ b/.changeset/shared-survey-state.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+---
+
+chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. No behavior change.
--- a/.changeset/shared-survey-state.md
+++ b/.changeset/shared-survey-state.md
@@ -3,4 +3,4 @@
 'posthog-js': patch
 ---
 
-chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. Core's survey enums are now const-object literal unions (matching the web SDK's existing pattern), so the same values type-check across both SDKs. No behavior change.
+chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. Core's survey enums are now const-object literal unions (matching the web SDK's existing pattern), so the same values type-check across both SDKs. No behavior change. Type-level note: enum members no longer work as standalone type annotations (e.g. `SurveyType.Popover` as a type); use the exported union types instead. Runtime values are unchanged.
--- a/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
+++ b/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
@@ -8,7 +8,6 @@ import {
     SurveyEventProperties,
     SurveyPosition,
     SurveyQuestion,
-    SurveySchedule,
     SurveyType,
     SurveyWidgetType,
 } from '../../posthog-surveys-types'
@@ -22,7 +21,14 @@ import {
     SURVEY_IN_PROGRESS_PREFIX,
 } from '../../utils/survey-utils'
 import { isNullish, type SurveyResponses } from '@posthog/core'
-import { buildSurveyResponseProperties, getSurveyResponseKey, surveyHasResponses } from '@posthog/core/surveys'
+import {
+    buildSurveyResponseProperties,
+    canSurveyActivateRepeatedly,
+    doesSurveyActivateByEvent,
+    getSurveyResponseKey,
+    getSurveyStorageKey,
+    surveyHasResponses,
+} from '@posthog/core/surveys'
 
 import { propertyComparisons } from '../../utils/property-utils'
 import { getTargetingUrl } from '../../utils/url-targeting-utils'
@@ -574,17 +580,13 @@ export const getDisplayOrderQuestions = (survey: Survey): SurveyQuestion[] => {
 }
 
 export const hasEvents = (survey: Pick<Survey, 'conditions'>): boolean => {
-    return survey.conditions?.events?.values?.length != undefined && survey.conditions?.events?.values?.length > 0
+    return doesSurveyActivateByEvent(survey)
 }
 
 export const canActivateRepeatedly = (
     survey: Pick<Survey, 'schedule' | 'conditions' | 'id' | 'current_iteration'>
 ): boolean => {
-    return (
-        !!(survey.conditions?.events?.repeatedActivation && hasEvents(survey)) ||
-        survey.schedule === SurveySchedule.Always ||
-        isSurveyInProgress(survey)
-    )
+    return canSurveyActivateRepeatedly(survey) || isSurveyInProgress(survey)
 }
 
 /**
@@ -701,11 +703,7 @@ interface InProgressSurveyState {
 }
 
 const getInProgressSurveyStateKey = (survey: Pick<Survey, 'id' | 'current_iteration'>): string => {
-    let key = `${SURVEY_IN_PROGRESS_PREFIX}${survey.id}`
-    if (survey.current_iteration && survey.current_iteration > 0) {
-        key = `${SURVEY_IN_PROGRESS_PREFIX}${survey.id}_${survey.current_iteration}`
-    }
-    return key
+    return getSurveyStorageKey(SURVEY_IN_PROGRESS_PREFIX, survey)
 }
 
 export const setInProgressSurveyState = (
--- a/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
+++ b/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
@@ -13,7 +13,6 @@ import {
 } from '../../posthog-surveys-types'
 import { document as _document, window as _window } from '../../utils/globals'
 import {
-    doesSurveyActivateByEvent,
     getSurveyInteractionProperty,
     getSurveySeenKey,
     getSurveyAbandonedKey,
@@ -577,10 +576,6 @@ export const getDisplayOrderQuestions = (survey: Survey): SurveyQuestion[] => {
     }
 
     return reverseIfUnshuffled(survey.questions, shuffle(survey.questions))
-}
-
-export const hasEvents = (survey: Pick<Survey, 'conditions'>): boolean => {
-    return doesSurveyActivateByEvent(survey)
 }
 
 export const canActivateRepeatedly = (
--- a/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
+++ b/packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
@@ -13,9 +13,11 @@ import {
 } from '../../posthog-surveys-types'
 import { document as _document, window as _window } from '../../utils/globals'
 import {
+    doesSurveyActivateByEvent,
     getSurveyInteractionProperty,
     getSurveySeenKey,
     getSurveyAbandonedKey,
+    getSurveyStorageKey,
     SURVEY_LOGGER as logger,
     setSurveySeenOnLocalStorage,
     SURVEY_IN_PROGRESS_PREFIX,
@@ -24,9 +26,7 @@ import { isNullish, type SurveyResponses } from '@posthog/core'
 import {
     buildSurveyResponseProperties,
     canSurveyActivateRepeatedly,
-    doesSurveyActivateByEvent,
     getSurveyResponseKey,
-    getSurveyStorageKey,
     surveyHasResponses,
 } from '@posthog/core/surveys'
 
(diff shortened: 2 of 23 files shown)


packages/react-native/test/getActiveMatchingSurveys.spec.ts
```diff
@@ -354,6 +354,35 @@ describe('getActiveMatchingSurveys', () => {
...
+          id: 'repeating-survey',
+          schedule: SurveySchedule.Recurring,
+          current_iteration: 2,
+        }),
+      ]
+
+      const result = getActiveMatchingSurveys(surveys, mockFlags, ['repeating-survey_2'], mockActivatedSurveys)
+
+      expect(result).toHaveLength(0)
+    })
```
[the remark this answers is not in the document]
lucasheriques: Done in 039c7024a — collapsed into an it.each table with an extra row for the bare-id (pre-iteration) key.

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -228,6 +228,15 @@ export const canActivateRepeatedly = (survey: Survey): boolean => {
...
 }
 
+// Mirrors the web SDK's seen-key logic: keying by iteration lets a repeating survey
+// show again on this device when a new iteration starts.
+export const getSurveySeenKey = (survey: Pick<Survey, 'id' | 'current_iteration'>): string => {
+  if (survey.current_iteration && survey.current_iteration > 0) {
+    return `${survey.id}_${survey.current_iteration}`
+  }
+  return survey.id
+}
```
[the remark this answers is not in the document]
lucasheriques: Addressed in 039c7024a — the key builder moved to @posthog/core as getSurveyIterationKey with direct parameterized tests covering 2/1/0/null/undefined, so all SDK packages share one tested implementation.

packages/react-native/src/surveys/useSurveyStorage.ts
```diff
@@ -1,14 +1,29 @@
...
 }
 
+// To keep storage bounded, only keep the last 20 seen surveys
+const MAX_SEEN_SURVEYS = 20
+
+// One slot per survey: stale keys from earlier iterations (or the pre-iteration bare id)
+// can never match again and would otherwise evict other surveys' seen state.
+export function updateSeenSurveys(current: string[], survey: SurveyWithIteration): string[] {
+  const surveyKey = getSurveyIterationKey(survey)
+  return [surveyKey, ...current.filter((key) => key !== survey.id && !key.startsWith(`${survey.id}_`))].slice(
```
turnipdabeets: The hydrate guard only checks `parsedSeenSurveys[0]`, so a mixed array like `["a", 42]` makes it into state — and `.startsWith` then throws inside the close handler, on every close (the old `!==` filter didn't care about types). Worth a type guard here, or a `.filter((k) => typeof k === 'string')` where the list is parsed?

```suggestion
  return [surveyKey, ...current.filter((key) => typeof key === 'string' && key !== survey.id && !key.startsWith(`${survey.id}_`))].slice(
```

packages/core/src/surveys/keys.ts
```diff
@@ -0,0 +1,15 @@
...
+export type SurveyWithIteration = Pick<Survey, 'id' | 'current_iteration'>
+
+/**
+ * Iteration-qualified survey identifier ('id' or 'id_iteration'), used to key
+ * per-survey display state (seen, in-progress, ...). Keying by iteration lets a
+ * repeating survey become visible again when a new iteration starts.
+ */
+export function getSurveyIterationKey(survey: SurveyWithIteration): string {
+  if (survey.current_iteration && survey.current_iteration > 0) {
+    return `${survey.id}_${survey.current_iteration}`
```
turnipdabeets, line 20: nit: the `${id}_` grammar now lives here but gets re-derived in RN's `updateSeenSurveys` via `key.startsWith(`${survey.id}_`)`. Worth exporting something like `isSurveyIterationKeyFor(key, surveyId)` from this file so the format only exists in one place?

packages/core/src/surveys/activation.ts
```diff
@@ -0,0 +1,28 @@
...
+export function doesSurveyActivateByEvent(survey: SurveyForRepeatActivation): boolean {
+  return !!survey.conditions?.events?.values?.length
+}
+
+/**
+ * Platform-independent part of "can this survey show again after being seen":
+ * event-repeated activation or an 'always' schedule. SDKs may OR in
+ * platform-specific state (e.g. the web SDK's in-progress partial responses).
+ */
+export function canSurveyActivateRepeatedly(survey: SurveyForRepeatActivation): boolean {
```
turnipdabeets, line 23: keys.ts got its own spec in this PR but activation.ts didn't — these helpers are only covered through the browser/RN wrapper tests. Worth a small activation.spec.ts (Always, repeatedActivation false/undefined, null conditions) so coverage survives if those wrappers ever change?

.changeset/shared-survey-state.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+---
+
+chore: survey seen-key and repeat-activation helpers now live in @posthog/core, shared by the web and React Native SDKs. Core's survey enums are now const-object literal unions (matching the web SDK's existing pattern), so the same values type-check across both SDKs. No behavior change.
```
turnipdabeets: "No behavior change" is true at runtime, but external TS consumers who used enum members as types (`let x: SurveyType.Popover`) won't compile after this. Probably fine as a patch if `@posthog/core` is treated as internal — worth one sentence here acknowledging the type-level change?

packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -574,17 +580,13 @@ export const getDisplayOrderQuestions = (survey: Survey): SurveyQuestion[] => {
 }
 
 export const hasEvents = (survey: Pick<Survey, 'conditions'>): boolean => {
-    return survey.conditions?.events?.values?.length != undefined && survey.conditions?.events?.values?.length > 0
+    return doesSurveyActivateByEvent(survey)
```
turnipdabeets: nit: with `canActivateRepeatedly` now delegating to core, `hasEvents` has no production callers left — only its own test imports it. Delete, or keep deliberately as a public re-export?

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -214,20 +213,10 @@ export const getDisplayOrderQuestions = (survey: Survey): SurveyQuestion[] => {
   // return reverseIfUnshuffled(survey.questions, shuffle(survey.questions))
 }
 
-export const hasEvents = (survey: Survey): boolean => {
-  return survey.conditions?.events?.values !== undefined && survey.conditions.events.values.length > 0
-}
-
 // export const hasActions = (survey: Survey): boolean => {
```
turnipdabeets: nit: with its live neighbors gone, this commented-out `hasActions` block is now free-floating. Delete it? Git history keeps it if RN ever gets action-based activation.

turnipdabeets (commented):
LGMT

Also isRepeatableSurvey in packages/browser/src/utils/survey-event-receiver.ts is a third copy of canSurveyActivateRepeatedly — its "stays out of the core bundle" comment no longer applies since the helper lives in @posthog/core; worth delegating?

turnipdabeets:
@lucasheriques friendly ping :)

packages/react-native/src/surveys/useSurveyStorage.ts
lucasheriques: Good catch, fixed in 434b03cd4 — but at the parse boundary instead: hydrate now filters non-string entries, so state is always `string[]` and every downstream consumer is safe, not just this filter. Added a hydration test with a mixed persisted array (`['survey-a', 42, null, 'survey-b']`) that also exercises the close path.

packages/core/src/surveys/keys.ts
lucasheriques: Done in 434b03cd4 — added `isSurveyKeyForSurvey(key, surveyId)` to core's keys.ts (matches the bare id or any `id_n` key, which is what the call site needs, hence the slightly broader name than suggested) and `updateSeenSurveys` uses it now. keys.spec.ts covers it, including prefix collisions (`abc` must not match `abcd_1`).

packages/core/src/surveys/activation.ts
lucasheriques: Added in 434b03cd4 — parameterized activation.spec.ts covering both helpers: Always schedule, repeatedActivation true/false/undefined, repeatedActivation without event values, and null/undefined conditions and schedule.

.changeset/shared-survey-state.md
lucasheriques: Agreed — added a type-level note to the changeset in 434b03cd4. Keeping it a patch since @posthog/core is internal-first and the exported union types are the supported annotation; the note gives anyone affected the one-line migration.

packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
lucasheriques: Deleted in 434b03cd4 — no production callers and it's not exported from any public entrypoint, so no reason to keep a re-export. Its test coverage moved to core's new activation.spec.ts where the real implementation lives.

packages/react-native/src/surveys/surveys-utils.ts
lucasheriques: Deleted in 434b03cd4.

lucasheriques:
@turnipdabeets all seven addressed in 434b03cd4 (replies on each thread), including the one from your review body: `isRepeatableSurvey` in survey-event-receiver.ts now delegates to core's `canSurveyActivateRepeatedly` — you're right that the "stays out of the core bundle" rationale went stale the moment the helper moved to @posthog/core (survey-utils.ts in the main bundle already imports from there). Used the core helper rather than the web extension's `canActivateRepeatedly` wrapper on purpose: the wrapper ORs in in-progress state, which would change the receiver's consume-on-shown semantics.

One deliberate non-change worth flagging: legacy bare seen IDs are not migrated to the current iteration key. The old SDK never re-showed a survey after one answer, so a bare ID almost always represents an old iteration — migrating would wrongly suppress the current iteration and delay the fix by a cycle. Expanded the backward-compat paragraph in the PR description with this. Ready for another look!

turnipdabeets (approved):

PR #4091, opened 2026-07-06
feat(next): add createPostHog with server-side identity resolution

dustinbyrne:
## Problem

Apps need a way to provide user identity **server-side**. Today `@posthog/next` derives identity from the PostHog cookie and tracing headers, both of which are client-provided and spoofable — a problem for apps that can't rely on client identity (e.g. for GDPR or security reasons). There was no way to say "the current request is user X" from the server's source of truth (the auth session), the way posthog-rails resolves `current_user`.

## Changes

**`createPostHog()` is now the single server entry point** (the bare `getPostHog` / `getServerSidePostHog` exports are removed — pre-1.0, no deprecation cycle):

```ts
// lib/posthog.ts
import 'server-only'
import { createPostHog } from '@posthog/next'
import { auth } from '@/auth'

export const { getPostHog } = createPostHog({
    getDistinctId: async () => (await auth())?.user?.id,
})
```

- **`getDistinctId` resolver**: when it returns a non-empty string, that id overrides the client-provided distinct id for every event and flag evaluation. Session/device linkage (`$session_id`, `$device_id`, `$window_id`) stays client-provided so server events still correlate with the browser session (consistent with posthog-node and posthog-rails).
- **Precedence**: explicit call args > server-resolved id > tracing headers > cookie > anonymous/personless.
- **Safety**: the resolver is never called for opted-out users; Next.js control-flow errors (`redirect()`/`notFound()` — recognized by `digest`, version-tolerant across `next >= 13`) are rethrown; other resolver errors warn and fall back to client identity; numeric ids are coerced with a warning (matching posthog-js `identify()`).
- **Pages Router parity**: the same `getPostHog` works in both routers — call `getPostHog(ctx)` with the `GetServerSidePropsContext` inside `getServerSideProps`, and the resolver receives it so helpers like `getServerSession(ctx.req, ctx.res, ...)` work (mirrors next-auth v5's `auth()`, which takes the request context as an optional argument in the Pages Router).
- **Why a factory, not global registration**: exporting both a raw and a configured variant of the same function invites silent mis-imports that drop server identity. Identity-bearing Next.js SDKs converged on app-code factories (next-auth v5's `NextAuth()`, `@supabase/ssr`'s `createServerClient` utils, Vercel Flags SDK's `flag()`); global registration is confined to identity-free observability config (Sentry).

Verified end-to-end with a `next build` + `next start` app installing the packed tarball, and the App Router example app now uses the factory pattern (the tarball CI builds it).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [ ] Accounted for backwards compatibility of any changes (no breaking changes!) — **intentionally breaking**: `@posthog/next` is pre-1.0; the bare `getPostHog`/`getServerSidePostHog` exports are removed in favor of `createPostHog()` (see changeset)
- [x] Took care not to unnecessarily increase the bundle size — client/edge barrels unchanged; server-only code stays out of client bundles (enforced by packaging tests)

### If releasing new changes

## 🤖 Agent context

- Authored with Claude Code (Fable 5). Explored `posthog-rails`' request-context design and this repo's context architecture with subagents, then researched how identity-bearing Next.js SDKs (next-auth v5, `@supabase/ssr`, Vercel Flags SDK, Sentry, Clerk, LaunchDarkly) bind per-request identity via a fact-checked multi-agent research pass before choosing the factory shape.
- Key decisions: factory over `configureX()` global registration (silent-misconfiguration and bundle-duplication hazards; ecosystem precedent); factory-only exports after the human flagged the raw-vs-configured name-collision footgun; server id wins over client identity but session/device linkage is preserved.
- A multi-agent code review of the diff caught and fixed: swallowed `redirect()`/`notFound()` control-flow errors in the resolver, silent discard of non-string ids, and overstated JSDoc claims. The Pages Router export-resolution bug found during this work is split into #4092, which this PR stacks on.


dustinbyrne:
Docs for this API are staged on PostHog/posthog.com#17935 (draft).

packages/next/src/pages/getServerSidePostHog.ts
```diff
@@ -47,7 +38,16 @@ export async function getServerSidePostHog(
...
+        const contextData = buildContextData(tracing, state)
+
+        if (getDistinctId) {
+            const serverDistinctId = await resolveServerDistinctId(getDistinctId, ctx)
+            if (serverDistinctId) {
+                contextData.distinctId = serverDistinctId
+            }
+        }
+
+        client.enterContext(contextData)
```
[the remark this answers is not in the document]
dustinbyrne: this is fixed now that `withRequestContext` returns a `Proxy`

packages/next/src/shared/identity.ts
```diff
@@ -7,3 +8,83 @@ import { uuidv7 } from '@posthog/core'
...
+            return undefined
+        }
+        const trimmed = distinctId.trim()
+        return trimmed !== '' ? trimmed : undefined
+    } catch (error) {
+        if (isNextControlFlowError(error)) {
+            throw error
+        }
+        // eslint-disable-next-line no-console
+        console.warn('[PostHog Next.js] getDistinctId threw — falling back to client-provided identity', error)
```
turnipdabeets, line 87: [question] When the resolver throws, we fall back to the client-provided distinct id — the identity this feature exists to distrust. If someone can make the resolver throw (say, a malformed session cookie that crashes the auth lib), they get client-controlled attribution back. Was fail-open over fail-closed (dropping to anonymous) a deliberate call? A config option could let us offer both without changing the default.

packages/next/src/server/getPostHog.ts
```diff
@@ -5,33 +5,34 @@ import { getOrCreateNodeClient } from './clientCache.node.js'
...
+ * Wraps the shared client in a Proxy that applies request-scoped context
+ * to every method call. We can't use enterContext() here because
+ * AsyncLocalStorage.enterWith() doesn't propagate back to the caller
+ * across the await boundary of this async function.
  */
-export async function getPostHog(apiKey?: string, options?: Partial<PostHogOptions>): Promise<IPostHog> {
+export function withRequestContext(client: IPostHog, contextData: Parameters<IPostHog['withContext']>[0]): IPostHog {
+    return new Proxy(client, {
+        get(target, prop, receiver) {
+            if (prop === 'withContext') {
```
turnipdabeets, line 19: [question] The Proxy switch is a solid fix for the shared-context leak. One subtlety: if app code sets its own identity with `posthog.withContext({ distinctId: 'X' }, () => posthog.capture(...))`, the capture still goes out under the request identity — every proxied method call re-applies `contextData`, and posthog-node's context merge gives it precedence over the ambient one. This predates the PR on the App Router path, but it now carries the server-resolved id — is that the behavior we want? (Explicit `capture({ distinctId })` args still win.)

packages/next/src/server/createPostHog.ts
```diff
@@ -0,0 +1,100 @@
...
+ *     const flags = await posthog.getAllFlagsAndPayloads()
+ *     return { props: { posthogBootstrap: flags } }
+ * }
+ * ```
+ */
+export function createPostHog(config: CreatePostHogConfig = {}): CreatePostHogResult {
+    return {
+        getPostHog: (ctx?: GetServerSidePropsContext) =>
+            ctx
+                ? getServerSidePostHog(ctx, config.apiKey, config.options, config.getDistinctId)
```
turnipdabeets: [suggestion] In a Pages Router app, forgetting `ctx` sends us down the App Router branch into `cookies()` from `next/headers`, which throws a cryptic error there (and a wrong-but-truthy ctx crashes bare on `ctx.req.headers.cookie`) — so this path already breaks the host app today. Since we avoid throwing in SDKs, could we require ctx in the Pages factory's `getPostHog` signature so it's a compile error, and/or warn + return the context-less client as a graceful fallback (like the missing-apiKey path does)?

packages/next/tests/createPostHog.test.ts
```diff
@@ -0,0 +1,379 @@
...
+                expect.any(Function)
+            )
+            expect(warnSpy).toHaveBeenCalledWith(
+                '[PostHog Next.js] getDistinctId threw — falling back to client-provided identity',
+                expect.any(Error)
+            )
+            warnSpy.mockRestore()
+        })
+
+        it.each(['NEXT_REDIRECT;replace;/login;307;', 'NEXT_NOT_FOUND', 'NEXT_HTTP_ERROR_FALLBACK;404'])(
```
turnipdabeets: [suggestion] The digest-based check is a nice version-tolerant approach. `DYNAMIC_SERVER_USAGE` and `BAILOUT_TO_CLIENT_SIDE_RENDERING` are in the allowlist but not in this test though — a regression that drops those branches would swallow dynamic-rendering bailouts and still pass the suite.
```suggestion
        it.each([
            'NEXT_REDIRECT;replace;/login;307;',
            'NEXT_NOT_FOUND',
            'NEXT_HTTP_ERROR_FALLBACK;404',
            'DYNAMIC_SERVER_USAGE',
            'BAILOUT_TO_CLIENT_SIDE_RENDERING',
        ])(
```

packages/next/tests/createPostHog.test.ts
```diff
@@ -0,0 +1,379 @@
...
+            const getDistinctId = jest.fn(() => 'server-user')
+
+            const { getPostHog } = createPostHog({ getDistinctId })
+            await getPostHog()
+
+            expect(getDistinctId).not.toHaveBeenCalled()
+            warnSpy.mockRestore()
+        })
+
+        it('calls the resolver once per getPostHog call', async () => {
```
turnipdabeets, line 300: [suggestion] The invariant that motivated the Proxy switch — two interleaved requests on the shared client keeping separate identities — doesn't have a positive test yet. Two factories with different resolver ids + `Promise.all`, asserting each capture saw its own `distinctId`, would lock it in for us.

.changeset/olive-crabs-listen.md
```diff
@@ -0,0 +1,23 @@
+---
+'@posthog/next': minor
+---
+
+Replace the `getPostHog` and `getServerSidePostHog` exports with a `createPostHog()` factory. Configure PostHog once in a shared module — including an optional `getDistinctId` resolver that attributes server-side events and feature flags to the authenticated user, overriding the client-provided (spoofable) identity — and use the returned `getPostHog` everywhere. App Router server code imports `createPostHog` from `@posthog/next` and calls `getPostHog()` with no arguments. Pages Router server code imports `createPostHog` from `@posthog/next/pages` and calls `getPostHog(ctx)` with the `GetServerSidePropsContext`.
```
turnipdabeets: [nit] The `bootstrap` prop deprecation on `PagesPostHogProviderProps` isn't mentioned in either changeset — a line here would get the migration hint into the CHANGELOG.

packages/next/src/shared/identity.ts
dustinbyrne: current behavior is client trust by default, so this is deliberate.

when overriding the server-side `getDistinctId`, the server has full authority, so we're not relying on any user provided data. if we do fall back, the identity is only written to the request context. server-side calls to `capture`/`getFeatureFlag` can provide an explicit distinct id at the call site as the authoritative identity.

there's a case to be made for disabling server-side tracing headers (in all relevant SDKs) and cookie access, but i think it's out of scope for this PR

packages/next/src/server/getPostHog.ts
dustinbyrne: i could see this being a future enhancement where we update node's `withContext` to accept an optional parameter which only sets a context key if it doesn't already exist

otherwise i think this is fine for now - this integration largely manages request context automatically, so manual calls to `withContext({ distinctId: 'X' })` are somewhat unexpected

packages/next/src/server/createPostHog.ts
dustinbyrne: good catch, i've split this into pages / app router barrels so it's a compile time check

packages/next/tests/createPostHog.test.ts
dustinbyrne: added

packages/next/tests/createPostHog.test.ts
dustinbyrne: added

.changeset/olive-crabs-listen.md
dustinbyrne: for this change i'd prefer to keep it a silent deprecation - i plan on updating it further in another PR

turnipdabeets (approved):

PR #4094, opened 2026-07-06
fix(ai): wrap responses.stream() so $ai_generation events are captured

nandinitiw:
## What

Fixes #3583.

`responses.stream()` is a dedicated streaming helper on the OpenAI Responses API — separate from `responses.create({ stream: true })`. `WrappedResponses` overrode `create()` and `parse()` but not `stream()`, so any call through that path:

- Silently ignored `posthogTraceId`, `posthogDistinctId`, etc. — no `$ai_generation` event emitted
- Caused a 400 error from OpenAI if posthog params were passed at the top level (`Unknown parameter: 'posthog_trace_id'`)

This is the recommended streaming path for the Responses API in production, so this was a gap in the core use case.

## How

Added a `stream()` override to `WrappedResponses` in `packages/ai/src/openai/index.ts` that:

1. Calls `extractPosthogParams()` to strip PostHog params before forwarding to `super.stream()` — params never reach OpenAI
2. Iterates the `ResponseStream` in a background async task, collecting usage, stop reason, and completion ID from the `response.completed` event
3. Calls `captureAiGeneration` on completion (or error)
4. Returns the `ResponseStream` runner unmodified — callers see no behavioral change

## Testing

Added a `describe('responses.stream()')` block in `tests/openai.test.ts` that mocks `Responses.prototype.stream` with a fake `ResponseStream` (async iterable + `finalResponse()`) and verifies a `$ai_generation` event is captured with the correct `distinctId`, model, and token counts.

### DIFF
--- a/packages/ai/tests/openai.test.ts
+++ b/packages/ai/tests/openai.test.ts
@@ -71,6 +71,13 @@ jest.mock('openai', () => {
     parse() {
       return Promise.resolve({})
     }
+    // Simulates SDK's stream(): calls create({ stream: true }) and yields the result
+    async *stream(params: any) {
+      const result = await this.create({ ...params, stream: true })
+      for await (const chunk of result as any) {
+        yield chunk
+      }
+    }
   }
 
   // Mock Embeddings class
@@ -1599,6 +1606,68 @@ describe('PostHogOpenAI - Jest test suite', () => {
         },
       ])
       expect(properties['$ai_output_tokens']).toBe(0)
+    })
+  })
+
+  describe('responses.stream()', () => {
+    test('captures exactly one $ai_generation event with correct distinctId via responses.stream()', async () => {
+      // Mock at the network level (Responses.prototype.create) so the real
+      // ResponseStream machinery runs, catching any double-capture regression
+      // where stream() re-enters the wrapped create().
+      const chunks = [
+        {
+          type: 'response.output_text.delta',
+          delta: 'Hello!',
+        },
+        {
+          type: 'response.completed',
+          response: {
+            id: 'resp-stream-test',
+            model: 'gpt-4o',
+            status: 'completed',
+            output: [{ type: 'message', role: 'assistant', content: [{ type: 'output_text', text: 'Hello!' }] }],
+            usage: {
+              input_tokens: 10,
+              output_tokens: 5,
+              output_tokens_details: { reasoning_tokens: 0 },
+              input_tokens_details: { cached_tokens: 0 },
+            },
+          },
+        },
+      ]
+
+      const streamChunks = chunks
+      ;(openaiModule.Responses as any).prototype.create = jest.fn().mockImplementation((params: any) => {
+        if (params.stream) {
+          const mockStream = {
+            tee: jest
+              .fn()
+              .mockReturnValue([createMockAsyncIterator(streamChunks), createMockAsyncIterator(streamChunks)]),
+          }
+          return createMockAPIPromise(mockStream)
+        }
+        return createMockAPIPromise(mockOpenAiParsedResponse)
+      })
+
+      const runner = client.responses.stream({
+        model: 'gpt-4o',
+        input: 'Hello',
+        posthogDistinctId: 'test-stream-user',
+      } as any)
+
+      for await (const _ of runner) { /* consume */ }
+      await flushPromises()
+
+      // Must be exactly one capture — no double-fire from re-entrancy
+      expect(mockPostHogClient.capture).toHaveBeenCalledTimes(1)
+      const { distinctId, event, properties } = (mockPostHogClient.capture as jest.Mock).mock.calls[0][0]
+      expect(distinctId).toBe('test-stream-user')
+      expect(event).toBe('$ai_generation')
+      expect(properties['$ai_provider']).toBe('openai')
+      expect(properties['$ai_model']).toBe('gpt-4o')
+      expect(properties['$ai_input_tokens']).toBe(10)
+      expect(properties['$ai_output_tokens']).toBe(5)
+      expect(properties['$ai_completion_id']).toBe('resp-stream-test')
     })
   })
 


nandinitiw:
Still keen to get this merged — happy to address any feedback or make changes if needed. Let me know!

nandinitiw:
Hey @marandaneto @carlos-marchal-ph — would appreciate a review when you have a moment!

packages/ai/src/openai/index.ts
```diff
@@ -747,6 +748,116 @@ export class WrappedResponses extends Responses {
...
   }
+
+  public stream<Params extends ResponseStreamParams, ParsedT = ExtractParsedContentFromParams<Params>>(
+    body: Params & MonitoringParams,
+    options?: RequestOptions
+  ): ResponseStream<ParsedT> {
+    const { providerParams: openAIParams, posthogParams } = extractPosthogParams(body)
+    const startTime = Date.now()
+
+    const runner = super.stream<Params, ParsedT>(openAIParams, options)
```
ioannisj: I think it double-captures `$ai_generation` on every `responses.stream()` call.

`super.stream()` doesn't hit the network directly. OpenAI's `Responses.prototype.stream(body)` calls `this._client.responses.create({ ...body, stream: true })` under the hood (via `ResponseStream.createResponse`), and since `client.responses` is this `WrappedResponses` instance, that re-enters the already-wrapped `create()`, whose streaming branch also calls `captureAiGeneration`. So one `stream()` call emits two events. I confirmed the dispatch against openai@6.25.0.

It's a bit worse than a plain duplicate: the posthog params get stripped at line 756 before `super.stream()`, so the internal `create()` re-extracts from an already-stripped body and its event ends up with no `distinctId` and a fresh random `traceId`. The duplicate is orphaned.

`parse()` already guards this same re-entrancy by swapping `this.create` to the raw `super.create` (lines 687-690, restored at 746-748). One heads-up though: that swap won't transplant as-is, because `_run` defers the internal `create()` with `setTimeout(0)`, so a `finally`-restored swap is already gone by the time it runs.

A couple of ways out:
- forward the body untouched to `super.stream()` and let the internal wrapped `create()` be the single capture point. It strips the params, captures with the right context, and fixes the 400, so the whole background loop goes away and the override just needs to widen the type to `& MonitoringParams`
- or keep the loop but suppress the internal capture with a swap that survives the async deferral

The first is smaller and lines up with the `create({ stream: true })` behaviour. Is there a reason to prefer keeping the loop?

Would also be worth a test that mocks only the network-level `create` (instead of stubbing `Responses.prototype.stream` wholesale like the current one does) and asserts a single capture with the caller's `distinctId`, that's what would've caught this.

packages/ai/src/openai/index.ts
```diff
@@ -747,6 +748,116 @@ export class WrappedResponses extends Responses {
...
+        let usage: {
+          inputTokens?: number
+          outputTokens?: number
+          reasoningTokens?: number
+          cacheReadInputTokens?: number
+          webSearchCount?: number
+        } = { inputTokens: 0, outputTokens: 0, webSearchCount: 0 }
+        let rawUsageData: unknown
+
+        for await (const chunk of runner) {
```
ioannisj: The analytics loop and the caller iterate the same `ResponseStream` here, so a caller cancelling the stream early gets logged as a failed generation.

`ResponseStream` can't be `.tee()`'d, so when the caller breaks their `for await` early (a normal cancel, think "stop generating"), the shared `asyncIterator.return()` calls `abort()`. This loop then rejects with `APIUserAbortError` and the catch at line 837 emits an error `$ai_generation` with `output: []`, `latency: 0` and zero tokens.

The `create({ stream: true })` path uses `.tee()`, so its analytics stream is decoupled and still captures the full generation on an early break. I checked both against the real openai helpers: with tee the caller breaks after 2 chunks and analytics still sees all 6, with `stream()` analytics captures `{ error: 'Request was aborted.' }`. Full consumption and `finalResponse()`-only usage are both fine.

Could we treat `APIUserAbortError` in the catch as a cancel rather than an error, so skip the capture or record the partial `usage`/`completionId` the loop already has instead of zeroing it? If you take the first fix option from the double-capture comment, `create()`'s tee handles this one for free anyway.

ioannisj (commented):
Thanks for tackling this, it's a real gap and the direction is right, wrapping `stream()` so `$ai_generation` gets captured is what #3583 needs.

Two things I'd sort out before merge, both in the new `stream()` method:
- it double-captures the event, because `super.stream()` re-enters the wrapped `create()` (and one copy ends up orphaned with no `distinctId`)
- a caller cancelling the stream early gets logged as a failed generation, unlike the `create({ stream: true })` path

Details inline. The first is the blocker, the second is more of a parity thing that could even be a follow-up. 

This PR also needs a changeset item

nandinitiw:
Thanks for the detailed review! Both issues are addressed in the latest commit:

- **Double-capture**: replaced the background loop entirely with a simple pass-through — `super.stream()` calls `this.create()` internally so the wrapped `create()` is the single capture point (via its `tee()` path). No separate loop needed.
- **Early-cancel false error**: fixed for free by the same change — since we're no longer iterating in a background goroutine, cancelling the stream has no effect on analytics.
- Also updated the test to mock at the network level (`Responses.prototype.create`) instead of stubbing `stream()` wholesale, which would have caught the re-entrancy.
- Changeset added.

Happy to make any other changes!

ioannisj:
@nandinitiw can we solved merge conflicts here? I also think that now we avoid double capture but are are dropping props distinctId etc?

nandinitiw:
Hey @ioannisj — both issues are now addressed:

- **Merge conflicts**: rebased on latest main (including the just-merged #4072 changes)
- **`distinctId` dropping**: removed the `stream()` override entirely. The root cause was that the override stripped posthog params before calling `super.stream()`, which then called back into `WrappedResponses.create()` without them. Since the SDK's internal call chain already routes through `WrappedResponses.create()` with the full body, no override is needed — `posthogDistinctId` now flows correctly.

Also updated the test: added `stream()` to `MockResponses` in the Jest mock so it runs without a real API key, and switched from `conditionalTest` to `test` so it runs in CI.

marandaneto:
@nandinitiw your changes in packages/ai/src/openai/index.ts is just formatting? i dont see code changes to fix this issue at all

nandinitiw:
Good catch — you're right that the diff looked like just formatting, and I want to explain why.

I traced the delegation chain in the OpenAI SDK: `Responses.prototype.stream(body, options)` calls `ResponseStream.createResponse(this._client, body, options)`, which calls `client.responses.create({ ...body, stream: true }, options)`. Since `this._client` is the `PostHogOpenAI` instance and `client.responses` is our `WrappedResponses` instance, that `create()` call already routes through our overridden `WrappedResponses.create()` — which extracts and captures posthog params, including `distinctId`. **No override of `stream()` is needed at all** — the base class's inherited method already delegates correctly.

The earlier commits on this branch added a `stream()` override (to fix #3583), then patched bugs in that override (double-capture, distinctId stripping), before I ultimately realized the override itself was unnecessary and removed it — which is why the net diff against main is empty in `index.ts`.

I also found a real bug I introduced during a rebase: an earlier version of this branch had accidentally reverted some of the `service_tier` capture logic from #4072 due to a bad `git rebase --skip`. I rebuilt the branch clean from current `main` to fix that — it's now confirmed zero-diff against `main` on `index.ts`.

What this PR actually adds is test coverage: a new test that calls `client.responses.stream()`, mocks at the network level (`Responses.prototype.create`) so the real `ResponseStream` machinery runs, and asserts `distinctId` reaches the captured `$ai_generation` event with exactly one capture (no double-fire). This closes the gap — previously nothing exercised `responses.stream()` at all, so a regression here would have gone unnoticed.

Let me know if you'd still like some defensive code in `index.ts` even though it's not strictly necessary today, happy to add it if you think it's warranted for clarity/future-proofing.

marandaneto:
- Streaming instrumentation is already covered through responses.create({ stream: true }) in
   packages/ai/tests/openai.test.ts:687, including error behavior around :1164.
 - responses.stream() currently works because OpenAI’s implementation delegates back to client.responses.create({ stream:
   true }), which reaches WrappedResponses.create() in packages/ai/src/openai/index.ts:479.

nandinitiw:
That's right, and that's exactly the distinction I'd draw: the existing `openai.test.ts:687` coverage exercises the underlying `WrappedResponses.create()` path via `responses.create({ stream: true })`, but nothing previously exercised the `.stream()` **entry point** itself.

That distinction matters here specifically because #3583 was a bug in that entry point — a caller going through `client.responses.stream()` hit a 400 or lost `distinctId`, even though `create({ stream: true })` worked fine. The delegation from `stream()` to `create()` is an SDK implementation detail we don't own; if a future OpenAI SDK version changes that delegation, or someone reintroduces a `stream()` override on our side (as earlier commits on this branch did, before I removed it), the existing `create({stream:true})` test wouldn't catch it — only a test that actually calls `.stream()` would.

So I'd keep the new test as a regression guard on that specific entry point, even though today it's provably redundant with the create() path in terms of what code executes. Happy to trim it if you feel differently, but wanted to lay out the reasoning.

marandaneto:
https://github.com/PostHog/posthog-js/pull/4480

PR #4096, opened 2026-07-07
fix(browser): remove capture ver query param

marandaneto:
## Problem

The capture backend no longer uses the legacy `ver` query parameter for capture and session recording endpoints. The browser SDK was still appending it to every SDK request, including browser capture endpoints where it is now unnecessary, while feature flags still need it.

## Changes

- Stop adding `ver` to browser capture/session recording endpoints: `/e/` and `/s/`.
- Remove any pre-existing `ver` query parameter when sending to those endpoints, including configured/prefixed/proxied paths like `/i/v0/e/`, `/i/v0/s/`, and `/ingest/s/`.
- Keep sending `ver` for non-capture endpoints such as `/flags/`.
- Use the existing compat-safe URL parsing helper to identify request paths.
- Avoid `String.prototype.endsWith` so older browsers without that method do not fail remote config/request setup.
- Add request tests covering removed browser capture `ver`, proxied capture paths, existing `ver` removal, feature flag preservation, and missing `String.prototype.endsWith`.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi implemented the browser SDK follow-up to the already-merged backend capture change. The implementation intentionally limits `ver` removal to the browser SDK's capture/session recording endpoints and preserves feature flag behavior because feature flags still use the SDK version hint.

Verification run:

- `pnpm exec eslint src/request.ts src/__tests__/request.test.ts`
- `pnpm exec jest src/__tests__/request.test.ts --runInBand`
- `pnpm turbo --filter=posthog-js... build`
- `pnpm --filter posthog-js test:unit -- request.test.ts`

### DIFF
--- a/packages/browser/src/request.ts
+++ b/packages/browser/src/request.ts
@@ -1,7 +1,7 @@
 import { each, find } from './utils'
 import Config from './config'
 import { Compression, RequestWithOptions, RequestResponse } from './types'
-import { formDataToQuery, getQueryParam } from './utils/request-utils'
+import { convertToURL, formDataToQuery, getQueryParam } from './utils/request-utils'
 
 import { logger } from './utils/logger'
 import { AbortController, CompressionStream, fetch, navigator, XMLHttpRequest } from './utils/globals'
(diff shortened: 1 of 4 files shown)


marandaneto:
follow up https://github.com/PostHog/posthog/pull/68497

packages/browser/src/request.ts
```diff
@@ -388,10 +388,29 @@ const _sendBeacon = (options: RequestWithOptions) => {
...
+const VERSIONLESS_ENDPOINTS = ['/e/', '/batch/', '/capture/', '/track/', '/engage/', '/s/']
+
+const getURLPath = (url: string): string => {
+    const path = url
+        .split(/[?#]/)[0]
+        .replace(/^[a-z][a-z\d+.-]*:\/\/[^/]*(?=\/|$)/i, '')
+        .replace(/^\/\/[^/]*(?=\/|$)/, '')
+
+    return path ? (path[0] === '/' ? path : `/${path}`) : '/'
+}
```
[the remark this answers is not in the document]
marandaneto: Updated in abaf38a59 — I used the existing compat-safe `convertToURL` helper instead of `new URL()` because `new URL()` fails this package’s `compat/compat` lint target, and narrowed the endpoint list to the browser SDK capture endpoints.

packages/browser/src/request.ts
```diff
@@ -388,10 +388,27 @@ const _sendBeacon = (options: RequestWithOptions) => {
...
+    const parsedURL = convertToURL(url)
+    const path = parsedURL?.pathname || url.split(/[?#]/)[0]
+
+    return path ? (path[0] === '/' ? path : `/${path}`) : '/'
+}
+
+const isVersionlessEndpoint = (url: string): boolean => {
+    const path = getURLPath(url)
+
+    return VERSIONLESS_ENDPOINTS.some((endpoint) => path === endpoint || path.endsWith(endpoint))
```
ioannisj: No need for `path === endpoint` since it's covered by `path.endsWith(endpoint))`?

ioannisj (approved):

packages/browser/src/request.ts
marandaneto: have to change this cus i cant use endsWIth on IE11 anyway

marandaneto: Good catch. Fixed in f66ab40e4 — removed the explicit equality branch and replaced `endsWith` with a `slice`-based suffix helper so older browsers without `String.prototype.endsWith` do not throw.

PR #4106, opened 2026-07-07
feat(node): capture v1 wire types and event transform

eli-r-ph:
> Stacked on #4105. Review that first; this PR's diff is against the PR1 branch.

## Problem

Second PR in the Capture V1 stack for `posthog-node`. Before any transport work, we need the pure, well-tested translation from a queued (v0-normalized) event into the v1 wire shape (`POST /i/v1/analytics/events`). Keeping this a pure function makes the sentinel/coercion rules exhaustively unit-testable in isolation.

## Changes

- `capture-v1/types.ts`: `V1Event` / `V1Batch` / `V1EventOptions` wire types, plus the `V1BatchResponse` per-uuid result map and `V1ResultCode` union.
- `capture-v1/transform.ts`:
  - `buildV1Event(message)` / `buildV1Batch(messages, {createdAt, historicalMigration})`.
  - `coerceBool` / `coerceString` matching the server's coerce-or-omit semantics (booleans, `"true"/"1"/"false"/"0"`, nonzero numbers; strings and integer-to-decimal-string).
  - Lifts legacy `$`-sentinels into typed `options` (always stripping the property, even when coercion omits the option), promotes `$session_id`/`$window_id` to top-level strings (non-string dropped), strips `$lib`/`$lib_version` (server injects from `PostHog-Sdk-Info`), and relocates top-level `$set`/`$set_once` into `properties` (existing `properties` value wins on collision).
  - Never mutates the caller's queued message.

Not wired into any send path yet — this is inert building-block code consumed by PR3/PR4.

### Verification

- New `transform.spec.ts`: 57 tests, 100% statement/branch/line/function coverage of `transform.ts` (coercion matrices, each sentinel, purity, timestamp normalization, batch envelope).
- Node package builds cleanly (types generated); ESLint clean.

### Libraries affected

- [ ] posthog-node

No changeset here by design: this is internal plumbing with no user-facing behavior. The single `posthog-node` changeset lands in the wiring PR (PR4) so the release notes describe the shipped feature, not dead code.

## Checklist

- [x] Tests for new code

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8 agent). Sentinel mapping and coercion semantics were derived from the backend V1 parity standard and cross-checked against the posthog-go / posthog-rs / posthog-python v1 implementations. Kept as a pure function with no I/O so the mapping rules are fully covered by fast unit tests before the transport layer is introduced.

### DIFF
--- a/packages/node/src/capture-v1/transform.ts
+++ b/packages/node/src/capture-v1/transform.ts
@@ -68,7 +68,8 @@ function isRecord(value: JsonType | undefined): value is PostHogEventProperties 
  */
 function toRfc3339(timestamp: unknown): string {
   if (typeof timestamp === 'string') {
-    return timestamp
+    const asDate = new Date(timestamp)
+    return Number.isNaN(asDate.getTime()) ? new Date().toISOString() : timestamp
   }
   if (timestamp instanceof Date) {
     return timestamp.toISOString()
--- a/packages/node/src/capture-v1/transform.ts
+++ b/packages/node/src/capture-v1/transform.ts
@@ -29,18 +29,13 @@ export function coerceBool(value: JsonType): boolean | undefined {
 }
 
 /**
- * Coerce a value to a string the way the capture server does, or return
- * `undefined` if it cannot be coerced. Accepts native strings and integer
- * numbers (rendered as a decimal string); floats/booleans/objects/arrays fail.
+ * Accept only a native string, else return `undefined` (the option is then
+ * omitted). The backend's `product_tour_id` is `Option<String>`, so numbers,
+ * booleans, objects, and arrays are dropped rather than coerced — matching
+ * posthog-go, posthog-rs, and posthog-python exactly.
  */
 export function coerceString(value: JsonType): string | undefined {
-  if (typeof value === 'string') {
-    return value
-  }
-  if (typeof value === 'number' && Number.isInteger(value)) {
-    return String(value)
-  }
-  return undefined
+  return typeof value === 'string' ? value : undefined
 }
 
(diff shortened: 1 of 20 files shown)


packages/node/src/capture-v1/transform.ts
```diff
@@ -0,0 +1,159 @@
...
+ * fall back to "now" so a bad timestamp never rejects the whole batch.
+ */
+function toRfc3339(timestamp: unknown): string {
+  if (typeof timestamp === 'string') {
+    const asDate = new Date(timestamp)
+    return Number.isNaN(asDate.getTime()) ? new Date().toISOString() : timestamp
+  }
+  if (timestamp instanceof Date) {
+    return timestamp.toISOString()
+  }
```
dustinbyrne, line 76: we don't `NaN` check here - an invalid date will throw

packages/node/src/capture-v1/transform.ts
```diff
@@ -0,0 +1,159 @@
...
+
+/**
+ * Coerce a value to a boolean the way the capture server does, or return
+ * `undefined` if it cannot be coerced (the option is then omitted so the server
+ * falls back to its default). Accepts native booleans, the strings
+ * `"true"`/`"1"`/`"false"`/`"0"` (trimmed, case-insensitive), and any number
+ * (nonzero -> true).
+ */
+export function coerceBool(value: JsonType): boolean | undefined {
+  if (typeof value === 'boolean') {
```
dustinbyrne, line 13: `isYesLike`/`isNoLike` are similar to this

dustinbyrne (approved):

PR #4107, opened 2026-07-07
feat(node): Capture V1 transport sender with retry and error surfacing

eli-r-ph:
> Stacked on #4106 (which is stacked on #4105). Review those first; this PR's diff is against the PR2 branch.

## Problem

Third PR in the Capture V1 stack. With the pure transform in place (PR2), this adds the transport layer that actually talks to `POST /i/v1/analytics/events` and implements the v1 retry/response contract. Kept as a standalone, dependency-injected class so the full attempt loop is unit-testable without a live client or real timers.

## Changes

- `capture-v1/errors.ts`: `CaptureV1Error` carrying `requestId`, `drops` (uuid + server details), `retryExhausted` uuids, and an optional `cause`.
- `capture-v1/sender.ts`: `V1CaptureSender.sendV1Batch(messages)` owning the whole attempt loop:
  - **Request:** `POST {host}/i/v1/analytics/events`, `Authorization: Bearer`, and the six required headers. `PostHog-Request-Id` + `created_at` stable across attempts, `PostHog-Attempt` increments, `PostHog-Request-Timestamp` regenerated per attempt. No `api_key`/`sent_at` in the body.
  - **Partial retry:** each 2xx shrinks the next attempt to only `retry`-tagged events; `ok`/`warning`/unknown/absent are terminal successes; `drop` is a terminal failure.
  - **Status classification:** retryable `{408,500,502,503,504}`; **429 terminal**; other non-2xx terminal; **malformed 2xx terminal**; transport errors retried to the budget.
  - **Backoff:** `delay = max(exponential, min(retry_after, maxBackoff))`, 30s default ceiling. `Retry-After` parsed as delta-seconds or HTTP-date, treated as a minimum, clamped to the ceiling; past/non-positive ignored.
  - **Error surfacing:** drops and retry-exhausted events (with `details`) reach the error channel via `CaptureV1Error`; earlier-attempt drops are never swallowed by a later success. Handled failures resolve rather than throw (the internal budget replaces v0's queue-level retry).
  - gzip with uncompressed fallback; injectable `fetch`/`now`/`sleep`/`generateRequestId`/`compress` seams.

Not wired into the client yet — PR4 constructs the sender from client state and routes `sendBatch` to it.

### Verification

- New `sender.spec.ts`: 105 tests (with the transform suite), 100% line coverage of the `capture-v1` module. Covers the header/envelope contract, every result code, partial-retry batch shrinking, drop-accumulation-across-attempts, the full status matrix, transport-error retry/exhaustion, malformed-2xx, and the backoff/`Retry-After` min/clamp/date-parse rules.
- Node builds cleanly; ESLint clean.

### Libraries affected

- [ ] posthog-node

No changeset here by design (internal plumbing). The single `posthog-node` changeset lands in the wiring PR (PR4).

## Checklist

- [x] Tests for new code

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8 agent). The retry/backoff/response semantics follow the backend V1 parity standard (429-terminal, malformed-2xx-terminal, `Retry-After` as a clamped minimum, 30s ceiling) and were cross-checked against posthog-rs/go/python. The sender resolves (rather than throws) on handled failures so it slots cleanly behind the core `sendBatch` seam for both the batched and immediate paths without changing v0 queue semantics.

### DIFF
--- a/packages/node/src/capture-v1/sender.ts
+++ b/packages/node/src/capture-v1/sender.ts
@@ -1,0 +1,291 @@
+import {
+  type FetchLike,
+  type PostHogEventProperties,
+  type PostHogFetchOptions,
+  type PostHogFetchResponse,
+  gzipCompress,
+  safeJsonStringify,
+  safeSetTimeout,
+  uuidv7,
+} from '@posthog/core'
+
+import { CaptureV1Error, type V1DroppedEvent } from './errors'
+import { buildV1Batch } from './transform'
+import type { V1BatchResponse, V1Event } from './types'
+
+const V1_ANALYTICS_PATH = '/i/v1/analytics/events'
+const DEFAULT_MAX_BACKOFF_MS = 30_000
+
+/** HTTP statuses the v1 backend wants retried. 429 is intentionally terminal in v1. */
+const RETRYABLE_STATUSES = new Set([408, 500, 502, 503, 504])
+
+export interface V1CaptureSenderConfig {
+  host: string
+  apiKey: string
+  /** Canonical `$lib` identity, materialized by the server from `PostHog-Sdk-Info`. */
+  libraryId: string
+  libraryVersion: string
+  /** Value for the `User-Agent` header, if the runtime allows setting it. */
+  userAgent?: string
+  historicalMigration: boolean
+  /** When true, the body is gzip-compressed (falling back to uncompressed on failure). */
+  compressionEnabled: boolean
+  requestTimeoutMs: number
+  /** Total attempts, including the first (1 initial + N retries). */
+  maxAttempts: number
+  /** Base exponential backoff, doubled each retry and capped at `maxBackoffMs`. */
+  initialRetryDelayMs: number
+  /** Single ceiling for both the backoff schedule and the `Retry-After` clamp. */
+  maxBackoffMs?: number
+  isDebug?: boolean
+}
+
+export interface V1CaptureSenderHooks {
+  fetch: FetchLike
+  /** Surfaces partial/terminal delivery failures on the client error channel. */
+  onError: (error: Error) => void
+  now?: () => number
+  sleep?: (ms: number) => Promise<void>
+  generateRequestId?: () => string
+  compress?: (payload: string, isDebug?: boolean) => Promise<Blob | null>
+}
+
+/**
+ * Sends already-normalized events to the Capture V1 endpoint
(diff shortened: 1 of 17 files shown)


packages/node/src/capture-v1/sender.ts
```diff
@@ -0,0 +1,298 @@
...
+    // at least one attempt so a bad config surfaces a delivery error instead of
+    // silently dropping the batch without ever sending or calling onError.
+    const maxAttempts = Math.max(1, this.config.maxAttempts)
+    for (let attempt = 1; attempt <= maxAttempts; attempt++) {
+      const isLastAttempt = attempt === maxAttempts
+      const payload = safeJsonStringify({
+        created_at: createdAt,
+        ...(this.config.historicalMigration ? { historical_migration: true } : {}),
+        batch: pending,
+      })
```
dustinbyrne, line 112: minor optimization, but this looks like it could be moved out of the loop

packages/node/src/capture-v1/sender.ts
```diff
@@ -0,0 +1,298 @@
...
+  }
+
+  private surfacePartialDrops(requestId: string, drops: V1DroppedEvent[]): void {
+    if (drops.length > 0) {
+      this.onError(new CaptureV1Error({ requestId, drops, retryExhausted: [] }))
+    }
+  }
+
+  private async cancelBody(response: PostHogFetchResponse): Promise<void> {
+    await response.body?.cancel()?.catch(() => {})
```
dustinbyrne, line 296: good call on this

packages/node/src/capture-v1/sender.ts
```diff
@@ -0,0 +1,298 @@
...
+  /** Parse `Retry-After` (delta-seconds or HTTP-date). Non-positive/past values are ignored. */
+  private parseRetryAfter(response: PostHogFetchResponse): number | undefined {
+    const raw = response.headers?.get('Retry-After')
+    if (!raw) {
+      return undefined
+    }
+    const trimmed = raw.trim()
+    if (/^\d+$/.test(trimmed)) {
+      const seconds = parseInt(trimmed, 10)
+      return seconds > 0 ? seconds * 1000 : undefined
```
dustinbyrne: this _could_ return `Infinity` if the header was malicious

```suggestion
      const seconds = parseInt(trimmed, 10)
      return Number.isFinite(seconds) && seconds > 0 ? seconds * 1000 : undefined
```

dustinbyrne (approved):

PR #4109, opened 2026-07-07
test(node): Capture V1 SDK compliance harness job

eli-r-ph:
> Top of the Capture V1 stack: #4108 → #4107 → #4106 → #4105. Review those first; this PR's diff is against the PR4 branch.

## Problem

Final PR in the stack. With v1 wired into the client (PR4), this proves the wire behavior against the real contract by running the [SDK test harness](https://github.com/PostHog/posthog-sdk-test-harness) V1 suite (`POST /i/v1/analytics/events`) in CI, alongside the existing v0 job.

## Changes

- **Adapter (`compliance/node/adapter.js`)**: reads `POSTHOG_CAPTURE_MODE`, advertises `capture_v1` vs `capture_v0` accordingly (so the harness selects exactly one capture contract per process), and passes `captureMode` to the client. Also threads `historical_migration` through `/init`, translates the harness `EventOptions` into the SDK's sentinel properties (`cookieless_mode`→`$cookieless_mode`, `disable_skew_correction`→`$ignore_sent_at`, `process_person_profile`→`$process_person_profile`, `product_tour_id`→`$product_tour_id`), and shortens the v1 retry backoff so retry tests stay within their wait windows.
- **`Dockerfile.v1`**: bakes `ENV POSTHOG_CAPTURE_MODE=v1` — the reusable compliance workflow only passes `PORT` at runtime, so mode has to be baked into the image.
- **Workflow**: adds a second job (`test-node-sdk-v1`) building `Dockerfile.v1` with its own report name, kept alongside the v0 job while both modes ship dual.
- **compose / README**: `POSTHOG_CAPTURE_MODE=v1 docker-compose up …` for local runs.

### Why a separate image/job instead of one adapter serving both

A single adapter process advertises fixed capabilities and can't satisfy both the v0 (`/batch/`) and v1 (`/i/v1/analytics/events`) contracts at once, and the reusable workflow doesn't pass adapter env at runtime. Baking the mode per image and running two jobs is the clean split; `capture_tests.yaml` requires `capture_v0` and `capture_analytics_v1_tests.yaml` requires `capture_v1`, so each run picks up only its contract (the shared `feature_flags` suite runs in both).

### Verification

Ran the harness (v0.10.0) locally against the v1 adapter:

```
Total: 111 | 110 passed | 1 failed
```

The single failure — `partial_batch_handling.respects_retry_after_on_partial` — was a **real SDK bug**: the 200 partial-retry path ignored a `Retry-After` header on the 2xx response. Fixed in the transport commit on the PR3 branch (`fix(node): honor Retry-After on 200 partial-retry responses`, + unit test). With that fix the full v1 contract passes, including endpoint/method, all required headers, event format + `options` lifting, `$set`/`$set_once`/`$groups` passthrough, batch envelope, retryable-vs-terminal status matrix, partial-batch pruning, uuid/timestamp/request-id preservation across retries, backoff + `Retry-After`, compression, `historical_migration`, `$geoip_disable`, and the feature-flag side-effect suite.

### Libraries affected

- [ ] posthog-node

No changeset (CI/test-only; the feature changeset is in PR4).

## Checklist

- [x] Tests for new code (compliance harness contract + the surfaced transport fix)
- [x] Accounted for backwards compatibility of any changes (v0 job unchanged)

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8 agent). The dual-job approach (keep v0, add v1) follows the maintainer's "run both modes for smoke testing" direction. The local harness run caught a genuine `Retry-After` gap that unit tests had missed, which is exactly why the harness job earns its place.

### DIFF
--- a/compliance/node/Dockerfile.v1
+++ b/compliance/node/Dockerfile.v1
@@ -1,0 +1,41 @@
+FROM node:24-alpine
+
+WORKDIR /app
+
+# Copy the entire monorepo (needed for workspace dependencies)
+COPY package.json pnpm-lock.yaml pnpm-workspace.yaml ./
+COPY tooling ./tooling
+COPY packages/types ./packages/types
+COPY packages/core ./packages/core
+COPY packages/node ./packages/node
+
+# Install pnpm
+RUN npm install -g pnpm@11.7.0
+
+# Install dependencies
+RUN pnpm install --frozen-lockfile
+
+# Build the types, core, and node packages
+RUN pnpm --filter @posthog/types build
+RUN pnpm --filter @posthog/core build
+RUN pnpm --filter posthog-node build
+
+# Set up adapter in separate directory
+WORKDIR /app/adapter
+
+# Copy adapter code
+COPY compliance/node/adapter.js ./
+
+# Install adapter dependencies (express) in clean directory
+RUN npm init -y && npm install express
+
+# Run the adapter in Capture V1 mode so it advertises the capture_v1 capability
+# and the harness runs the /i/v1/analytics/events contract against it. The
+# reusable compliance workflow only passes PORT at runtime, so the mode is baked
+# into the image here.
+ENV POSTHOG_CAPTURE_MODE=v1
+
+EXPOSE 8080
+
+USER node
+CMD ["node", "adapter.js"]
--- a/.github/workflows/sdk-compliance-tests-node.yml
+++ b/.github/workflows/sdk-compliance-tests-node.yml
@@ -23,6 +23,7 @@ permissions:
   pull-requests: write
 
 jobs:
+  # Legacy /batch/ capture contract (capture_v0). Default capture mode.
   test-node-sdk:
     uses: PostHog/posthog-sdk-test-harness/.github/workflows/test-sdk-action.yml@02c049e529001d02f37a534745678e057d371fb0
     with:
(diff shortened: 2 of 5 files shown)


compliance/node/Dockerfile.v1
```diff
@@ -0,0 +1,41 @@
...
+COPY compliance/node/adapter.js ./
+
+# Install adapter dependencies (express) in clean directory
+RUN npm init -y && npm install express
+
+# Run the adapter in Capture V1 mode so it advertises the capture_v1 capability
+# and the harness runs the /i/v1/analytics/events contract against it. The
+# reusable compliance workflow only passes PORT at runtime, so the mode is baked
+# into the image here.
+ENV POSTHOG_CAPTURE_MODE=v1
```
dustinbyrne, line 36: this looks like the only change compared to the other Dockerfile

suppose it's just easier to do it this way than add a customizable `env` property to the runner?

dustinbyrne (approved):

compliance/node/Dockerfile.v1
eli-r-ph: good call, circling back to this 👍

eli-r-ph: I think DRYing this up will require a change to the SDK library first, will make a note for later, there's a couple rough spots there to update that would smooth a few SDK integrations 👍

PR #4110, opened 2026-07-08
feat(react-native): add macOS support to native plugin

ioannisj:
## Problem

Native crash autocapture (`errorTracking.autocapture.nativeCrashes`) is available on iOS and Android but not macOS, even though `posthog-ios` fully supports macOS error tracking. `@posthog/react-native-plugin` didn't build on react-native-macos targets:

- the podspec declared no `osx` platform, so the pod wasn't autolinked/compatible on a `platform :macos` target;
- the Swift referenced posthog-ios session-replay and surveys APIs that don't exist on the macOS slice of the SDK (they're `#if os(iOS)` / `@available(macOS, unavailable)` upstream), so it wouldn't compile there.

## Changes

- **Podspec:** declare `:osx => '10.15'` and `s.osx.deployment_target = '10.15'` (matching `posthog-ios`), so the pod autolinks and builds on react-native-macos targets.
- **Swift (`PosthogReactNativePlugin.swift`):** wrap every iOS-only posthog-ios symbol in `#if os(iOS)` — the session-replay config block, `config.surveys`, and the bodies of `isSessionReplayActive` / `startSessionRecording` / `stopSessionRecording`. On macOS, `isEnabled` resolves `false` and the recording controls are no-ops. The `.mm` bridge is unchanged and both platforms still resolve their promises.
- Add the `macos` keyword.

Session replay stays iOS-only (unsupported by posthog-ios on macOS); macOS gets native error tracking only. iOS and Android behaviour is unchanged — `#if os(iOS)` is a no-op there and the Android/podspec-ios surfaces are untouched.

## Testing

- Built the plugin on a real **react-native-macos 0.81** macOS target against **PostHog 3.63.x** — **BUILD SUCCEEDED**, with Swift + ObjC objects produced for arm64.
- Cross-checked every guarded symbol against posthog-ios 3.63.x availability annotations (session replay + surveys are `#if os(iOS)` / `@available(macOS, unavailable)` upstream; error tracking is `#if os(iOS) || os(macOS) || os(tvOS)`).
- Verified end-to-end on macOS: enabled `nativeCrashes`, triggered a native SIGSEGV, and confirmed posthog-ios captured it and reported it to the project on next launch (symbolicated via uploaded dSYMs).

> **Release ordering:** this plugin release must land **before** the companion `posthog-react-native` change, which raises its plugin peer-dependency floor to the version published here.

## Agent context

**DRI:** @ioannisj

### DIFF
--- a/examples/example-rn-native-plugin/macos/Podfile
+++ b/examples/example-rn-native-plugin/macos/Podfile
@@ -12,11 +12,8 @@ target 'PosthogReactNativePluginExample-macOS' do
   platform :macos, '14.0'
   use_native_modules!
 
-  # Flags change depending on the env values.
-  flags = get_default_flags()
-
   use_react_native!(
-    :path => '../node_modules/react-native-macos',
+    :path => "#{ws_dir}/node_modules/react-native-macos",
     :hermes_enabled => false,
     :fabric_enabled => ENV['RCT_NEW_ARCH_ENABLED'] == '1',
     # An absolute path to your application root.
--- a/examples/example-rn-native-plugin/macos/Podfile
+++ b/examples/example-rn-native-plugin/macos/Podfile
@@ -13,7 +13,7 @@ target 'PosthogReactNativePluginExample-macOS' do
   use_native_modules!
 
   use_react_native!(
-    :path => "#{ws_dir}/node_modules/react-native-macos",
+    :path => '../node_modules/react-native-macos',
     :hermes_enabled => false,
     :fabric_enabled => ENV['RCT_NEW_ARCH_ENABLED'] == '1',
     # An absolute path to your application root.
--- a/examples/example-rn-native-plugin/macos/Podfile
+++ b/examples/example-rn-native-plugin/macos/Podfile
@@ -1,0 +1,47 @@
+require 'pathname'
+
+ws_dir = Pathname.new(__dir__)
+ws_dir = ws_dir.parent until
+  File.exist?("#{ws_dir}/node_modules/react-native-macos/scripts/react_native_pods.rb") ||
+  ws_dir.expand_path.to_s == '/'
+require "#{ws_dir}/node_modules/react-native-macos/scripts/react_native_pods.rb"
+
+prepare_react_native_project!
+
+target 'PosthogReactNativePluginExample-macOS' do
+  platform :macos, '14.0'
+  use_native_modules!
+
+  # Flags change depending on the env values.
+  flags = get_default_flags()
+
+  use_react_native!(
+    :path => '../node_modules/react-native-macos',
+    :hermes_enabled => false,
+    :fabric_enabled => ENV['RCT_NEW_ARCH_ENABLED'] == '1',
+    # An absolute path to your application root.
+    :app_path => "#{Pod::Config.instance.installation_root}/.."
+  )
+
+  post_install do |installer|
+    react_native_post_install(installer)
+
+    # Xcode 26.4+ Clang rejects fmt's FMT_STRING consteval. fmt's base.h
+    # redefines FMT_USE_CONSTEVAL based on compiler detection, so a -D flag
+    # is not enough — patch the header directly to force it off. Mirrors the
+    # same workaround in the iOS Podfile.
+    # Upstream:
+    #   https://github.com/fmtlib/fmt/issues/4740
+    #   https://github.com/facebook/react-native/issues/55601
+    fmt_base_h = File.join(installer.sandbox.root, 'fmt', 'include', 'fmt', 'base.h')
+    if File.exist?(fmt_base_h)
+      contents = File.read(fmt_base_h)
+      patched = contents.gsub(/^(#\s*define FMT_USE_CONSTEVAL )1$/, '\10')
+      if patched != contents
+        File.chmod(0644, fmt_base_h)
+        File.write(fmt_base_h, patched)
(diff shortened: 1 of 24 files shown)


packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
```diff
@@ -126,49 +126,53 @@ class PosthogReactNativePlugin: NSObject {
             isReactNativeFatalJsError(event) ? nil : event
```
turnipdabeets, line 137: Once #4111 lands, this filter runs on macOS crashes too. The marker strings were picked from iOS crash reports — do we know what a real macOS native crash inside JSI frames puts in crash_info_message? If it happens to contain facebook::jsi::JSError, we'd drop a genuine crash that JS never captured. Could also check on #4111 instead if that's a better home. 

Also - the dedup comment at the top of the file still says Android + iOS only.

packages/react-native-plugin/posthog-react-native-plugin.podspec
```diff
@@ -40,6 +40,7 @@ Pod::Spec.new do |s|
     s.dependency 'PostHog', "~> #{posthog_ios_version}"
   end
   s.ios.deployment_target = '13.0'
+  s.osx.deployment_target = '10.15'
```
turnipdabeets, line 43: The example and CI build at macOS 14.0 only, so a consumer at 10.15–13.x would be the first to ever compile this pod there. Probably fine since the guards are all compile-time `#if` rather than `@available`, but did you get a chance to build below 14?

examples/example-rn-native-plugin/macos/Podfile
```diff
@@ -0,0 +1,47 @@
...
+require "#{ws_dir}/node_modules/react-native-macos/scripts/react_native_pods.rb"
+
+prepare_react_native_project!
+
+target 'PosthogReactNativePluginExample-macOS' do
+  platform :macos, '14.0'
+  use_native_modules!
+
+  # Flags change depending on the env values.
+  flags = get_default_flags()
```
turnipdabeets: nit: leftover from the template — `flags` is never used since `use_react_native!` gets explicit values below.
```suggestion
```

examples/example-rn-native-plugin/macos/Podfile
```diff
@@ -0,0 +1,47 @@
...
+
+  use_react_native!(
+    :path => '../node_modules/react-native-macos',
```
turnipdabeets: Small thing — lines 3–7 walk up the tree to find react-native-macos, but this hardcodes the one layout. Reusing the resolved dir keeps the two from disagreeing if node_modules ever hoists:
```suggestion
    :path => "#{ws_dir}/node_modules/react-native-macos",
```

packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
```diff
@@ -261,13 +269,17 @@ class PosthogReactNativePlugin: NSObject {
     func startRecording(
         resumeCurrent: Bool, resolve: RCTPromiseResolveBlock, reject _: RCTPromiseRejectBlock
     ) {
-        PostHogSDK.shared.startSessionRecording(resumeCurrent: resumeCurrent)
+        #if os(iOS)
+            PostHogSDK.shared.startSessionRecording(resumeCurrent: resumeCurrent)
```
turnipdabeets, line 284: Small thing — on macOS this resolves as success with nothing started and no log. A one-time `hedgeLog("Session replay is not supported on macOS")` in the non-iOS path (here and in `stopRecording`) would save someone a debugging session when recording "starts" but nothing arrives.

turnipdabeets (approved):
LGTM, mostly questions and nits

packages/react-native-plugin/posthog-react-native-plugin.podspec
ioannisj: Hey no did not build below 14 tbh, but this should be fine since it matches native dependency's floor? Built locally on macOS 11.0 (rn floor) and the build was green so I think we should be good

packages/react-native-plugin/ios/PosthogReactNativePlugin.swift
ioannisj: Checked this locally and seems that we drop true duplicaes on macOS as well. So a true native crash inside JSI frame isn't dropped unless the crashing exception is literally is a jsi::JSError

examples/example-rn-native-plugin/macos/Podfile
ioannisj: pulled this back cause CI was failing

ioannisj: You were right that the `require` and `use_react_native!` should resolve from the same place. I tried your snippet as written first and it broke `pod install` in CI: `use_react_native!` treats `:path` as relative to the Podfile, but `#{ws_dir}` is absolute, so React Native's script turned it into `./Users/runner/.../node_modules/react-native-macos` and couldn't find `package.json` there.

Same idea, just kept relative. The walk-up result lives in `rn_macos_dir` now, the `require` reads from it, and `use_react_native!` gets `rn_macos_dir.relative_path_from(Pathname.new(__dir__))`. Both point at the same directory so they can't drift, and the relative path recomputes on its own if `node_modules` ever hoists. Verified with a clean `pod install` locally in 1e4c138f; the CI build is running now.

PR #4111, opened 2026-07-08
feat(react-native): enable native error tracking on macOS

ioannisj:
## Problem

`errorTracking.autocapture.nativeCrashes` never initialized on macOS: the JS layer refused to load the native plugin there (`OptionalPlugin.ts` excluded `macos`), so native crash autocapture silently no-op'd even with a macOS-capable plugin installed.

## Changes

- **`OptionalPlugin.ts`:** load `@posthog/react-native-plugin` on macOS (previously iOS/Android only). The legacy session-replay-only fallback (`posthog-react-native-session-replay`) stays iOS/Android, since it has no macOS support.
- Update the `nativeCrashes` JSDoc to include macOS.
- Bump the `@posthog/react-native-plugin` peer-dependency floor to `>= 2.2.0` (the release that adds macOS support). It remains an optional peer, so iOS/Android consumers on an older plugin only see a warning, never a hard failure.
- Add `optional-plugin.spec.ts` covering platform-gated plugin loading.

The error-tracking-only orchestration path (`initializeNativePlugin` with session replay disabled) is already platform-agnostic, so no other JS changes are needed.

## Testing

- New `optional-plugin.spec.ts` (5 cases): macOS loads the primary plugin only and never the legacy fallback; iOS falls back to the legacy plugin when the primary is absent; web loads nothing.
- Full `posthog-react-native` suite (494 tests) and typecheck pass.
- Verified end-to-end on a real **react-native-macos 0.81** app: JS init → native `setup()` → posthog-ios crash handler installed → native crash captured and uploaded/symbolicated in the project.

> **Depends on #4110** (`@posthog/react-native-plugin` macOS support) — merge/release this only after that plugin's `2.2.0` is published to npm, since this PR's peer floor (`>= 2.2.0`) references it.

## Agent context

**DRI:** @ioannisj

### DIFF
--- a/packages/react-native/src/optional/OptionalPlugin.ts
+++ b/packages/react-native/src/optional/OptionalPlugin.ts
@@ -23,7 +23,7 @@ if (Platform.OS !== 'web') {
     OptionalReactNativePlugin = require('@posthog/react-native-plugin')
   } catch (e) {}
 
-  // The legacy fallback is session-replay only and has no macOS support, so it's iOS/Android only.
+  // The legacy fallback is session-replay only and has no macOS support, so it's skipped on macOS.
   if (!OptionalReactNativePlugin && Platform.OS !== 'macos') {
     try {
       OptionalReactNativePlugin = require('posthog-react-native-session-replay')
--- a/packages/react-native/src/optional/OptionalPlugin.ts
+++ b/packages/react-native/src/optional/OptionalPlugin.ts
@@ -18,12 +18,13 @@ export type PostHogReactNativePluginExtended = typeof PostHogReactNativePlugin &
 
 export let OptionalReactNativePlugin: PostHogReactNativePluginExtended | undefined = undefined
 
-if (Platform.OS !== 'macos' && Platform.OS !== 'web') {
+if (Platform.OS !== 'web') {
   try {
     OptionalReactNativePlugin = require('@posthog/react-native-plugin')
   } catch (e) {}
 
-  if (!OptionalReactNativePlugin) {
+  // The legacy fallback is session-replay only and has no macOS support, so it's iOS/Android only.
+  if (!OptionalReactNativePlugin && Platform.OS !== 'macos') {
     try {
       OptionalReactNativePlugin = require('posthog-react-native-session-replay')
     } catch (e) {}
--- a/packages/react-native/test/optional-plugin.spec.ts
+++ b/packages/react-native/test/optional-plugin.spec.ts
@@ -54,6 +54,14 @@ describe('OptionalPlugin loader', () => {
     expect(loadOptionalPlugin('ios', { primaryInstalled: false, legacyInstalled: false })).toBeUndefined()
   })
 
+  it('loads the primary plugin on Android', () => {
+    expect(loadOptionalPlugin('android')).toBe(PRIMARY)
+  })
+
+  it('falls back to the legacy plugin on Android when the primary is not installed', () => {
+    expect(loadOptionalPlugin('android', { primaryInstalled: false })).toBe(LEGACY)
+  })
+
   it('loads no native plugin on web', () => {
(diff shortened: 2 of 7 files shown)


packages/react-native/src/optional/OptionalPlugin.ts
```diff
@@ -18,12 +18,13 @@ export type PostHogReactNativePluginExtended = typeof PostHogReactNativePlugin &
 
 export let OptionalReactNativePlugin: PostHogReactNativePluginExtended | undefined = undefined
 
-if (Platform.OS !== 'macos' && Platform.OS !== 'web') {
+if (Platform.OS !== 'web') {
   try {
     OptionalReactNativePlugin = require('@posthog/react-native-plugin')
```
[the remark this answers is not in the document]
ioannisj: it degrades gracefully (silent no-op, no crash) and the JSDoc now documents the >= 2.2.0 macOS floor

packages/react-native/src/optional/OptionalPlugin.ts
```diff
@@ -18,12 +18,13 @@ export type PostHogReactNativePluginExtended = typeof PostHogReactNativePlugin &
 
 export let OptionalReactNativePlugin: PostHogReactNativePluginExtended | undefined = undefined
 
-if (Platform.OS !== 'macos' && Platform.OS !== 'web') {
+if (Platform.OS !== 'web') {
```
turnipdabeets, line 21: [question] Now that the plugin loads on macOS, `enableSessionReplay: true` there runs native `setup()`, which silently skips replay — but the JS side still marks replay initialized and logs "Session replay started…". Before this PR macOS got a clear "not installed" warning; now it claims success while recording nothing. Could we turn replay off on macOS in the JS layer with a one-time warn?

packages/react-native/src/optional/OptionalPlugin.ts
```diff
@@ -18,12 +18,13 @@ export type PostHogReactNativePluginExtended = typeof PostHogReactNativePlugin &
...
   try {
     OptionalReactNativePlugin = require('@posthog/react-native-plugin')
   } catch (e) {}
 
-  if (!OptionalReactNativePlugin) {
+  // The legacy fallback is session-replay only and has no macOS support, so it's iOS/Android only.
```
turnipdabeets: [nit] The fallback also runs on windows/tvOS (any non-web, non-macos platform — unchanged behavior), so "iOS/Android only" slightly misdescribes the condition.
```suggestion
  // The legacy fallback is session-replay only and has no macOS support, so it's skipped on macOS.
```

packages/react-native/test/optional-plugin.spec.ts
```diff
@@ -0,0 +1,48 @@
...
+
+  it('does not fall back to the legacy (session-replay-only) plugin on macOS', () => {
+    expect(loadOptionalPlugin('macos', { primaryInstalled: false })).toBeUndefined()
+  })
+
+  it('loads the primary plugin on iOS', () => {
+    expect(loadOptionalPlugin('ios')).toBe(PRIMARY)
+  })
+
+  it('falls back to the legacy plugin on iOS when the primary is not installed', () => {
```
turnipdabeets, line 47: [suggestion] Could we mirror these for `android` (primary loads + legacy fallback when primary is absent) and add the neither-package-installed case on iOS? The gate branches on `Platform.OS`, so a future tightening that forgets `android` would currently pass this suite.

turnipdabeets (approved):

packages/react-native/src/optional/OptionalPlugin.ts
ioannisj: Yeah, fixed and js slide no longer logs a false start

packages/react-native/test/optional-plugin.spec.ts
ioannisj: Done

PR #4112, opened 2026-07-08
fix(replay): stop iframe teardown from disconnecting shadow DOM observers

TueHaulund:
## Problem

Session replay silently stops recording shadow DOM mutations after an iframe is torn down.

One shared `ShadowDomManager` observes every shadow root, but `MutationBuffer.reset()` reset it — and that fires on any buffer teardown, so an iframe being removed or navigating away disconnected every shadow-root observer page-wide until the next full snapshot. Reproduced from a customer session where a widget in an open shadow root stopped recording ~1s in.

## Changes

- `MutationBuffer.reset()` releases only its own resources; global shadow teardown now runs from `takeFullSnapshot` and `record()` stop.
- Shadow restore handlers are keyed by owning document, and `ShadowDomManager.resetForDoc()` (called from the observer-bundle disposer) tears down exactly the departing document's shadow observers — so a removed iframe no longer leaves its own shadow observers connected until the next full snapshot.
- Regression + unit tests: shadow mutations keep flowing after an unrelated iframe is removed (fails without the fix), and `resetForDoc()` only tears down the given document's handlers.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/rrweb

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset`

## 🤖 Agent context

Root-caused and fixed with Claude Code from a customer replay export; iframe teardown was disconnecting the shared shadow observer. Verified with a puppeteer repro (removal + navigation + stop leak-check) before writing the committed test.

### DIFF
--- a/.changeset/replay-shadow-observer-iframe-teardown.md
+++ b/.changeset/replay-shadow-observer-iframe-teardown.md
@@ -1,0 +1,8 @@
+---
+'@posthog/rrweb': patch
+'posthog-js': patch
+---
+
+fix session replay silently dropping shadow DOM mutations after an iframe teardown
+
+The single shared ShadowDomManager observes every shadow root on the page, but MutationBuffer.reset() disconnected it. That reset fires whenever any one buffer is torn down, so an iframe being removed or navigating away disconnected every shadow-root observer page-wide. Shadow DOM content (for example a widget mounted in an open shadow root) then stopped recording until the next periodic full snapshot re-registered it. Buffer teardown now releases only its own resources; global shadow observation is reset by takeFullSnapshot and on recording stop.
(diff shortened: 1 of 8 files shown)


arnohillen:
One comment from Claude
- packages/rrweb/rrweb/src/record/index.ts:445 — takeFullSnapshot's onSerialize tags shadow roots inside same-origin iframes with the top-level document, so resetForDoc(iframeDoc) never matches them (CONFIRMED, found independently by 4 finders). rrweb-snapshot passes the same onSerialize into same-origin iframe recursion (snapshot.ts iframe branch), and the closure hardcodes addShadowRoot(root, document). When that iframe is later removed, resetForDoc(contentDocument) skips the entries and findAndRemoveIframeBuffer also misses the buffer (its bufferDoc() is the main document). Result: a connected MutationObserver and a dead MutationBuffer stay in the global mutationBuffers array (locked/unlocked on every snapshot, pinning the dead document) until the next full snapshot or stop. This turns the PR's over-teardown into under-teardown for exactly the "shadow root inside an iframe" case. Fix: tag with the node's owner document (n.ownerDocument) instead of the closed-over document, or better, derive the owner inside addShadowRoot from dom.host(shadowRoot).ownerDocument so callers can't get it wrong.

arnohillen (approved):

PR #4115, opened 2026-07-08
feat(metrics): add posthog.metrics count/gauge/histogram API (alpha)

DanielVisca:
## Problem

PostHog Metrics (alpha) currently requires standing up an OpenTelemetry SDK — a `MeterProvider`, a `PeriodicExportingMetricReader`, an OTLP exporter — before you can count a single thing. That's the right escape hatch for teams who already run OTel, but it's a heavy on-ramp for everyone else.  Adding wrapper for posthog sdk for zero setup beyond SDK init.

## Changes

Adds `posthog.metrics` — a statsd-style pre-aggregating metrics client, no OTel packages needed:

```ts
posthog.metrics.count('orders_created', 1)
posthog.metrics.gauge('active_connections', 42)
posthog.metrics.histogram('api_latency', 187, { unit: 'ms' })
```

**How it works**

- **Pre-aggregation, not per-call requests.** Samples fold into per-series aggregates in memory (counts sum, gauges keep the last value, histograms accumulate bucket counts + sum/min/max) and flush every 10s (configurable) as OTLP/JSON to `/i/v1/metrics` — one data point per series per window, however many calls were made. A `sendBeacon` drain runs on pagehide so the last window isn't lost.
- **Standard wire protocol.** The payload is a plain OTLP `ExportMetricsServiceRequest`, so the SDK path and the raw-OTel path land in the same product with the same series identity. Sums and histograms use **delta temporality**, which the metrics query layer already handles (each sample contributes its own value) — no cross-window state, counter resets are a non-issue.
- **Cardinality guardrails by design.** Attributes are scalars only; series identity is name + type + unit + attributes; a `maxSeriesPerFlush` cap (default 1000) drops new series past the cap with one warning per window. Deliberately unlike logs, no per-user context (distinct ID / session ID) is auto-attached — per-user series are the classic metrics-cardinality explosion 

**Layering** (mirrors logs)

- `@posthog/types`: public types (`Metrics`, `MetricSample`, `BeforeSendMetricFn`, OTLP metric wire types, `MetricsConfig`)
- `@posthog/core`: `PostHogMetrics` aggregator + OTLP payload builders — SDK-agnostic, so posthog-node and React Native can wire the same core later
- `posthog-js` (browser): `posthog.metrics` extension adapting the core to `_send_request` (batching, compression, retry outcome mapping: 2xx ok, 413 drop-too-large, 0/429/5xx retain-and-retry, other 4xx drop)

**Encoding notes pinned by the ingest**: nano timestamps as decimal strings, histogram `count`/`bucketCounts` as plain JSON numbers (string-encoded u64s were silently dropped by upstream opentelemetry-proto deserializers — opentelemetry-rust#3328).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (written by hand — same format)

## 🤖 Agent context

 Built TDD: core aggregator tests written and failing first (24 tests), then the implementation; same for the browser wiring (6 tests). Key decisions along the way: delta temporality over cumulative (verified the metrics query runner treats delta samples as their own increase, so no cross-window SDK state); no auto-attached distinct/session IDs (cardinality); numeric u64 encoding in OTLP/JSON (verified against the ingest's own parse tests, which document the string-u64 silent-drop bug); pre-aggregation with a series cap as the guardrail. The aggregator lives in @posthog/core so server SDKs can adopt it without new protocol work.

Verified: @posthog/core suite 722 tests green; posthog-js browser unit suite 119 suites / 4877 tests green including built-artifact checks against a fresh `tsc -b && rollup -c` build; typecheck and eslint clean.

Known scope cuts for the draft: browser only (node/RN wiring is a follow-up on the same core), no lazy-loaded extension chunk yet (the aggregator is small, but happy to split it into an on-demand entrypoint like logs' console capture if bundle review wants it), and 413 batches are dropped rather than split.


packages/browser/src/posthog-metrics.ts
```diff
@@ -0,0 +1,172 @@
...
+            // eslint-disable-next-line compat/compat
+            return Promise.resolve()
+        }
+        this._transportOverride = transport
+        try {
+            return this._core.flush().catch((err) => this._logger.error('PostHog metrics flush failed:', err))
+        } finally {
+            this._transportOverride = undefined
+        }
+    }
```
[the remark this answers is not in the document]
DanielVisca: Good catch — real race. Fixed in a5b00080 by removing the override field entirely: `flush(transport)` now calls a new synchronous `core.drainWindow()` (snapshot + build, bypassing the flush serializer) and hands the payload straight to `_send_request` with the forced transport in the same tick. This also fixes the deeper unload problem: the serializer could be awaiting an in-flight send that never completes during teardown, so even a correctly-scoped override would have deadlocked the beacon. Regression test added ('drains synchronously over the given transport even while another flush is in flight').

packages/core/src/metrics/index.ts
```diff
@@ -0,0 +1,367 @@
...
+    }
+
+    if (!filtered.name || typeof filtered.name !== 'string') {
+      this._logger.warn('Dropping metric with empty name')
+      return
+    }
+    if (typeof filtered.value !== 'number' || !Number.isFinite(filtered.value)) {
+      this._logger.warn(`Dropping metric '${filtered.name}': value must be a finite number`)
+      return
+    }
```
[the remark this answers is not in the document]
DanielVisca: Fixed in a5b00080 — the monotonic check now runs in `_capture` after the beforeSend chain (and the pre-check in `count()` is gone, so there's a single enforcement point). Test added for a beforeSend that mutates a count negative.

packages/browser/src/extensions/extension-bundles.ts
```diff
@@ -111,6 +112,11 @@ export const LogsExtensions = {
     logs: PostHogLogs,
 } as const satisfies ExtensionClasses
 
+/** The posthog.metrics API (count, gauge, histogram). */
+export const MetricsExtensions = {
+    metrics: PostHogMetrics,
+} as const satisfies ExtensionClasses
```
dustinbyrne, line 118: also export this in `src/extensions/extension-bundles.ts`

e.g.
```ts
export const MetricsExtensions = bundles.MetricsExtensions as typeof BundleTypes.MetricsExtensions
```

packages/core/src/metrics/index.ts
```diff
@@ -0,0 +1,410 @@
...
+
+    const outcome = await this._instance._sendMetricsBatch(this._buildPayload(window))
+    switch (outcome.kind) {
+      case 'ok':
+        return
+      case 'retry-later':
+        // Transient failure: merge the unsent window back so the data rides
+        // the next flush instead of being lost.
+        this._mergeWindowBack(window)
+        return
```
dustinbyrne, line 298: if no additional metrics are enqueued, are we only relying on shutdown to flush the unsent window back out?

packages/browser/src/posthog-core.ts
```diff
@@ -1100,6 +1106,7 @@ export class PostHog implements PostHogInterface {
         }
 
         this.logs?.flushLogs('sendBeacon')
+        void this.metrics?.flush('sendBeacon')
```
dustinbyrne: the `!this.config.request_batching` condition above would skip this should it evaluate true

perhaps this should happen unconditionally? tbh it looks to me like the others should move too, but that's a separate issue

packages/core/src/metrics/metrics-utils.ts
dustinbyrne, line 1: this file looks like it contains an actual null byte so git thinks it's a binary file - it should be escaped

packages/core/src/metrics/index.ts
```diff
@@ -0,0 +1,410 @@
...
+      }
+      state = {
+        name: filtered.name,
+        type: filtered.type,
+        unit: filtered.unit,
+        attributes: filtered.attributes,
+        windowStartMs: Date.now(),
+      }
+      this._series.set(key, state)
+    }
```
dustinbyrne, line 200: nit, because this is so edge-case-y: `filtered.attributes` is stored by reference but baked into the key, so a caller could mutate it after capturing a sample

maybe we copy it? maybe it's not worth?

packages/browser/src/extensions/extension-bundles.ts
DanielVisca: Added in 06360069 — it was already exported from `src/extensions/extension-bundles.ts` (that's where the bundle object is defined), the missing re-export was the `entrypoints/extension-bundles.es.ts` entrypoint, which now mirrors the `LogsExtensions` line.

packages/core/src/metrics/index.ts
DanielVisca: yes, it was. A 'retry-later' merged the window back but nothing re-armed the timer, so with no new captures the data sat until shutdown. Fixed in 06360069: `_mergeWindowBack` is now followed by `_armFlushTimer()`, with a regression test ('retries a merged-back window on the next interval even with no new samples').

packages/browser/src/posthog-core.ts
DanielVisca: Agreed — metrics are pre-aggregated client-side and don't ride the request queue, so the drain shouldn't depend on `request_batching` at all. Moved above the conditional in 06360069 so it runs unconditionally. Left the others (`logs?.flushLogs`, queue unloads) where they are per your note that that's a separate issue.

packages/core/src/metrics/metrics-utils.ts
DanielVisca: Fixed in 06360069 the stray NUL is gone and the file is plain UTF-8 now (the diff for that commit still renders as binary because the *old* blob was the one containing the byte; subsequent diffs will be text).

packages/core/src/metrics/index.ts
DanielVisca: Worth it, I think the attribute values are baked into the series key, so a post-capture mutation would silently desync key from stored attributes. 06360069 shallow-copies `filtered.attributes` on first capture (attributes are flat `string → primitive`, so shallow is sufficient), with a regression test ('snapshots attributes at capture time later caller mutation cannot corrupt the series').

turnipdabeets (approved):

PR #4116, opened 2026-07-08
feat(browser): opt-in snippet unload fallback + $sent_send_beacon tagging

pauldambra:
## Problem

On slow connections the snippet queues `posthog.capture(...)` calls in memory while `array.js` downloads. If the visitor bounces first, those events are silently lost — customers on high-bounce landing pages perceive this as data loss, and today there is no delivery marker to even measure it.

Separately, nothing marks events that left the browser via the `sendBeacon` fallback in the SDK's unload flush, so fire-and-forget deliveries are indistinguishable in analysis.

## Changes

**1. Opt-in snippet companion block: `packages/browser/snippet/unload-fallback.js`**

A self-contained ES5 block customers paste near the snippet (either order; double-paste is harmless). On `pagehide`, if `array.js` has not loaded, queued `capture` calls are sent with `navigator.sendBeacon` to `/e/` (base64 `data=` body — CORS-safelisted, no preflight at unload), marked `$sent_by_snippet_fallback_on_unload: true`, and spliced from the queue only after the browser accepts the beacon, so a late-arriving `array.js` can never double-process them (the drain and handler are both synchronous, so they cannot interleave).

It is deliberately **not** part of the default snippet — it adds ~1.3 KB gz to every page's HTML, and it trades away guarantees (identity joining, pipeline hooks) that only make sense for sites measurably losing events to bounces. The default snippet is size-guarded to stay untouched (869 B min / 517 B gz).

It refuses to send when the SDK would not, or when it cannot be sure:

- stored consent is no-like; consent pending with `opt_out_capturing_by_default`; a queued `opt_out_capturing()` (last `opt_in`/`opt_out` call wins, mirroring the real drain's replay order); DNT with `respect_dnt`; any `cookieless_mode`
- the pipeline is customized: `before_send`, `sanitize_properties`, `property_blacklist`/`property_denylist`, `request_headers`, or a queued `set_config` — bypassing redaction would be worse than losing events
- traffic looks automated (`navigator.webdriver`, bot/headless UA), honoring `opt_out_useragent_filter: true` like the SDK
- `disable_beacon`, no `sendBeacon`, payload over the 64 KB beacon limit (max 50 events; oversize left queued)

Identity: queued `identify` id > persisted `distinct_id` (localStorage/sessionStorage/cookie, token-sanitized key, `persistence_name` honored) > generated personless `snippet-` id with `$process_person_profile: false` so bouncing first-time visitors never create junk person profiles. It never writes storage, never logs, and every path is wrapped so it cannot throw into page code.

`snippet/README.md` documents installation, guarantees, and caveats — the artifact to share with customers.

**2. `$sent_send_beacon: true` on the SDK's beacon transport** (`src/request.ts`)

Tagged once at the transport level, covering all producers (`RequestQueue.unload()`, `RetryQueue.unload()`, explicit `transport: 'sendBeacon'` captures). Gated to `/e/` payloads so `/s/` snapshots and logs are untouched. Zero existing snapshots changed.

**3. Reference snippet + loader hardening**

`snippet/snippet.js` is a readable canonical copy of the classic snippet (tested as the exact file), now appending the loader `<script>` to `head` instead of `insertBefore` the first script element — works in script-less documents (e.g. programmatically created iframes). Production source of truth stays in posthog/posthog; grafting head-append there is a follow-up.

**Tests**: 70 jsdom tests execute the shipped files verbatim (`new Function`), covering the opt-in property in both directions, wire format, identity matrix, no-send matrix, queued-consent semantics, dedupe against the real `init_from_snippet()` drain, payload bounds, and minified-parity; a terser+gzip size test enforces byte budgets for both files; 2 chromium Playwright tests cover the slow-`array.js` fallback + no-double-processing end to end, and the `$sent_send_beacon` tag on real unload flushes.

**Why this lives in posthog-js rather than posthog/posthog**: the fallback's contract is entirely SDK internals (the `_i` tuple shape, `__loaded`, `_execute_array` replay order, persistence/consent key formats, the `/e/` wire shape), and the key dedupe test imports the real `init_from_snippet()` from source - a change to any of those internals breaks CI in the same PR here, instead of silently breaking pasted customer code and only surfacing against a released bundle. posthog/posthog gets the distribution surface (docs/UI embedding the minified block) and the production snippet head-append in a follow-up.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- Built with PostHog Code (Claude). Explored the snippet/transport/persistence/consent mechanics, planned, then implemented TDD (tests red first at each step).
- Direction changes during the session, driven by review + a threat-modeling pass: the fallback started embedded in the main snippet, then was extracted to a standalone opt-in block over size concerns and the risk that sites won't understand the identity/consent trade-offs; the threat model added the queued-opt-out handling, customized-pipeline skip, bot guard, and payload caps.
- Notable debugging: headless-browser events are dropped by the SDK's UA bot filter unless `opt_out_useragent_filter` is set (bit both the e2e fixture and, later, our own bot guard); Playwright served a stale `dist/` bundle predating the request.ts change; beacon bodies are only exposed via `postDataBuffer()` in Chromium.
- Commits are SSH-signed via the 1Password agent to satisfy the repo's verified-signature rule.

---


pauldambra:
* **#4116** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4116?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4116?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

pauldambra:
this isn't it

PR #4117, opened 2026-07-08
feat(metrics): wire posthog.metrics into posthog-node

DanielVisca:
## Problem

The Metrics product needs dogfooding traffic from real backend services, and the direction settled in #69428 (posthog repo) is that services should call the actual capture API rather than bridging Prometheus registries. That requires `posthog.metrics` in posthog-node — until now the API only existed in posthog-js (web, #4115). This is the SDK prerequisite for per-service adoption PRs in the main repo (logs services first).

## Changes

- `@posthog/core`: `PostHogCoreStateless` gains `_sendMetricsBatch` (same tagged-outcome contract and error classification as `_sendLogsBatch`, POSTing OTLP/JSON to `/i/v1/metrics`), which makes any core-based SDK a valid `MetricsHost`. `resolveMetricsConfig` moves from the browser package into core so both SDKs share one resolver (including the resourceAttributes-over-named-fields precedence rule), and the `Metrics` + `MetricsConfig` types are re-exported from the core barrel.
- `posthog-node`: lazy `client.metrics` getter (typed as the public `Metrics` interface) hosting core's pre-aggregating `PostHogMetrics`, a `metrics` client option (`serviceName`, `environment`, `flushIntervalMs`, `maxSeriesPerFlush`, `beforeSend`, ...), and a `_shutdown` hook that flushes the pending window and clears the timer. The flush timer is already unref'd by `safeSetTimeout`, so an un-shutdown process still exits.
- `posthog-js` (browser): no behavior change — it now imports the shared resolver instead of its local `metrics-defaults.ts` copy.

```ts
const client = new PostHog('phc_...', { metrics: { serviceName: 'billing-worker' } })
client.metrics.count('invoices.processed', 1, { attributes: { plan: 'pro' } })
client.metrics.gauge('queue.depth', 42)
client.metrics.histogram('job.duration', 187, { unit: 'ms' })
```

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types

`jest packages/node/src/__tests__/metrics.spec.ts` (5 tests). Each locks in a regression no existing test covers, since this wiring is new: end-to-end aggregation through the public `client.metrics` API into an OTLP POST with token and `service.name`; the retry-later path (network failure merges the window back and the next flush resends the summed counts); 413 dropping the batch instead of retrying forever; shutdown flushing the pending window; and the disabled gate sending nothing. Also re-ran the full core suite (728 passed) and browser metrics tests (7 passed) after moving `resolveMetricsConfig`, plus eslint/prettier on changed files and `pnpm generate-references` for the API report.

## 🤖 Agent context

Authored with PostHog Code (Claude). This PR was rebuilt on top of the merged #4115: compared to the first draft, config resolution moved into core (`resolveMetricsConfig`) instead of being duplicated inline (the inline version missed the resourceAttributes precedence rule), the getter returns the public `Metrics` interface rather than the concrete class, node API references were regenerated, and the test suite was extended with the retry-merge-back, 413, and disabled-gate cases. Direction context: a prom-registry-to-OTLP bridge for services was built and rejected twice (#69428, #69513 in the main repo) in favor of this SDK-first approach.


packages/node/src/client.ts
```diff
@@ -2381,6 +2409,12 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
 
     await this.featureFlagsPoller?.stopPoller(shutdownTimeoutMs)
     this.errorTracking.shutdown()
+    if (this._metrics) {
+      // Send whatever is aggregated in the current window, then clear the
+      // flush timer so it can't fire after teardown.
+      await this._metrics.flush().catch(() => {})
```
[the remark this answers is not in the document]
DanielVisca: Fixed in 4de7c49b — the shutdown metrics flush is now `Promise.race`d against `shutdownTimeoutMs` (default 30s when unset), so a hung transport can't hold teardown past the caller's deadline; regression test added (`bounds the shutdown metrics flush by the shutdown timeout`, hung-fetch + fake timers).

.changeset/metrics-node-wiring.md
```diff
@@ -0,0 +1,17 @@
+---
+'posthog-node': minor
+'@posthog/core': minor
```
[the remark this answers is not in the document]
DanielVisca: Fixed in 4de7c49b — added `'posthog-js': patch` to the changeset (the browser package's `metrics-defaults.ts` → core `resolveMetricsConfig` refactor rides this PR).

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1534,6 +1535,52 @@ export abstract class PostHogCoreStateless {
...
+          }
+          return isPostHogFetchRetryableError(err)
+        },
+      })
+      return { kind: 'ok' }
+    } catch (err) {
+      if (isPostHogFetchContentTooLargeError(err)) {
+        return { kind: 'too-large' }
+      }
+      if (err instanceof PostHogFetchNetworkError) {
```
dustinbyrne: [P1] Preserve a metrics window after exhausted retryable HTTP responses

The browser metrics adapter maps `429` and `5xx` responses to `retry-later`, and the Python implementation does the same. Here `fetchWithRetry()` retries those statuses internally, but after the retry budget is exhausted the final error is a `PostHogFetchHttpError`; this catch returns `retry-later` only for network errors, so the aggregator treats the final `429`/`503` as fatal and drops the window.

For example, with `fetchRetryCount: 0`:

```ts
client.metrics.count('jobs.processed', 3)
// first flush receives HTTP 503
client.metrics.count('jobs.processed', 2)
// second flush receives HTTP 200
```

The current second payload contains:

```text
asDouble: 2
```

Expected:

```text
asDouble: 5
```

This loses metric deltas during a transient ingestion outage and differs from both peer SDK paths.

Please classify `isPostHogFetchRetryableError(err)` as `retry-later` after excluding `413`, rather than checking only `PostHogFetchNetworkError`.

A regression test should cover exhausted `429` and `503` responses, then a successful flush, and assert that the successful payload includes both windows.

packages/core/src/metrics/config.ts
```diff
@@ -0,0 +1,23 @@
...
+export function resolveMetricsConfig(config: MetricsConfig | undefined): ResolvedPostHogMetricsConfig {
+  const resourceAttributes = config?.resourceAttributes
+  return {
+    serviceName: (resourceAttributes?.['service.name'] as string | undefined) ?? config?.serviceName,
+    serviceVersion: (resourceAttributes?.['service.version'] as string | undefined) ?? config?.serviceVersion,
+    environment: (resourceAttributes?.['deployment.environment'] as string | undefined) ?? config?.environment,
+    resourceAttributes,
+    beforeSend: config?.beforeSend,
+    flushIntervalMs: config?.flushIntervalMs ?? DEFAULT_FLUSH_INTERVAL_MS,
+    maxSeriesPerFlush: config?.maxSeriesPerFlush ?? DEFAULT_MAX_SERIES_PER_FLUSH,
```
dustinbyrne, line 21: [P1] Enforce `maxSeriesPerFlush` when merging a failed window back

This configuration promises a maximum number of retained series per flush window. Normal capture enforces it, but `_mergeWindowBack()` unconditionally inserts every absent series from the failed window into the live map. Python's equivalent merge path explicitly reapplies the cap.

A concrete timeline with `maxSeriesPerFlush: 2` is:

1. Capture series A and B and start a flush.
2. While the request is pending, capture new series C and D into the fresh live window.
3. Fail the request with a retryable network error.
4. Merge A and B back into C and D.

The next payload currently contains:

```text
4 data points: C, D, A, B
```

Expected:

```text
at most 2 data points
```

Repeated failed flushes with attribute churn can therefore grow the window beyond the advertised cardinality bound.

Please apply the cap and one-per-window drop warning inside `_mergeWindowBack()` before inserting a missing old series.

A regression test should use a deferred failing transport and the timeline above, then assert that the next payload never exceeds `maxSeriesPerFlush`.

packages/node/src/client.ts
```diff
@@ -2381,6 +2409,18 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+      // Send whatever is aggregated in the current window, then clear the flush
+      // timer so it can't fire after teardown. Raced against the shutdown budget:
+      // this flush runs before super._shutdown starts its own timeout, so an
+      // unresponsive transport must not be able to hold shutdown past the
+      // caller's deadline (its retry stack alone can take ~49s).
+      await Promise.race([
+        this._metrics.flush().catch(() => {}),
+        new Promise<void>((resolve) => safeSetTimeout(resolve, shutdownTimeoutMs ?? 30000)),
+      ])
+      this._metrics.reset()
```
dustinbyrne, line 2429: [P1] Prevent a timed-out metrics flush from restoring state after shutdown

`Promise.race()` lets shutdown continue while the original `metrics.flush()` is still running. `reset()` clears the current state, but if that in-flight request later resolves as `retry-later`, `_doFlush()` merges its captured window back and arms a new flush timer after teardown.

For example:

1. Capture one metric.
2. Start `shutdown(100)` with a fetch promise that remains pending.
3. Let the shutdown timeout win; shutdown returns and this `reset()` runs.
4. Reject the original fetch as a network failure.
5. Advance the metrics interval.

The executed reproduction observed:

```text
metrics calls when shutdown returned: 1
metrics calls after the late failure and timer: 2
```

Expected: no state or network activity can be recreated after shutdown returns. This matters in long-lived serverless workers where the process is reused after request teardown.

Please add a closed/generation state that makes late flush completion discard its window and prevents timer scheduling after reset/shutdown, or cancel/fully settle the in-flight flush before returning.

A regression test should use a deferred rejection after the timeout, advance fake timers, and assert that no second metrics request occurs and no series remain buffered.

packages/node/src/client.ts
```diff
@@ -488,6 +492,30 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+   * @example
+   * ```ts
+   * client.metrics.count('jobs.processed', 1, { attributes: { queue: 'default' } })
+   * client.metrics.gauge('queue.depth', 42)
+   * client.metrics.histogram('job.duration', 187, { unit: 'ms' })
+   * ```
+   *
+   * {@label Metrics}
+   */
+  public get metrics(): Metrics {
```
dustinbyrne, line 510: [P2] Add `metrics` to the exported `IPostHog` contract

The concrete `PostHogBackendClient` advertises this new public getter and implements the exported `IPostHog` abstraction, but `IPostHog` in `packages/node/src/types.ts` has no `metrics` member. Consumers that accept the SDK through its public interface cannot use the advertised API.

For example:

```ts
import type { IPostHog } from 'posthog-node'

const record = (client: IPostHog): void => {
  client.metrics.count('jobs.processed')
}
```

The current type check produces:

```text
TS2339: Property 'metrics' does not exist on type 'IPostHog'.
```

Expected: the public abstraction exposes the same metrics surface as concrete Node and edge clients.

Please import `Metrics` into `types.ts` and add `readonly metrics: Metrics` to `IPostHog`, or explicitly document that this interface intentionally excludes the new API.

A compile-time public-entrypoint test should type a `PostHog` instance as `IPostHog` and call `metrics.count()`.

packages/node/src/client.ts
```diff
@@ -488,6 +492,30 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
...
+    if (!this._metrics) {
+      // Lazy: `this` is the MetricsHost (isDisabled/optedOut/_sendMetricsBatch
+      // live on PostHogCoreStateless), so nothing is set up until first use.
+      this._metrics = new PostHogMetrics(this, resolveMetricsConfig(this.options.metrics), this._logger)
```
dustinbyrne, line 514: [P1] Guard malformed runtime metric attributes before `seriesKey()` throws

This exposes core `PostHogMetrics` directly to JavaScript callers. Although TypeScript restricts attributes to scalar values, runtime JavaScript and `any` inputs can still reach `seriesKey()`, which calls `JSON.stringify(attributes[k])` outside any guard. SDK capture methods should warn/drop malformed telemetry rather than throw into the host application.

For example:

```ts
client.metrics.count('jobs.processed', 1, {
  attributes: { jobId: 1n } as any,
})
```

The current call throws synchronously:

```text
TypeError: Do not know how to serialize a BigInt
```

Expected: the bad sample is dropped with a warning and application execution continues. A similar failure is possible with throwing getters or proxies.

Please validate/canonicalize runtime attributes inside the capture guard before generating the series key, and catch unexpected serialization failures.

Regression tests should keep one ordinary scalar-attribute case and add BigInt plus a throwing getter/proxy, asserting that capture does not throw and no malformed point is sent.

packages/node/src/client.ts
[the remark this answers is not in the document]
dustinbyrne: [P1] The added race bounds the metrics phase, but `shutdownTimeoutMs` still does not bound the complete shutdown

The public `shutdown(500)` contract is a maximum for teardown as a whole. The new metrics race can consume the full 500 ms, then `super._shutdown(shutdownTimeoutMs)` starts a fresh 500 ms timeout for queued events.

For example:

```ts
client.metrics.count('jobs.processed')
client.capture({ distinctId: 'u', event: 'queued-event' })
await client.shutdown(500) // both transports remain pending
```

The current timeline is:

```text
0-500 ms: metrics race
500-1000 ms: core event shutdown race
```

Expected:

```text
0-500 ms total maximum
```

This can overrun a serverless/process shutdown deadline even though the caller supplied a bound.

Please establish one absolute deadline and pass only the remaining budget to `super._shutdown()`, or run all shutdown work under one outer timeout.

A regression test should queue both a metric and an event with non-settling transports, advance fake timers, and assert that shutdown settles after one timeout budget rather than two.

packages/core/src/posthog-core-stateless.ts
DanielVisca: changed the catch to treat retryable http errors as retry later (413 still fatal), added a test covering 429/503 then a good flush, thanks

packages/core/src/metrics/config.ts
DanielVisca: added the cap check to the merge back path, same warning as capture, thanks

packages/node/src/client.ts
DanielVisca: added a generation counter so a stale flush gets discarded instead of merging back, thanks

packages/node/src/client.ts
DanielVisca: added metrics to IPostHog, thanks for catching

packages/node/src/client.ts
DanielVisca: wrapped the series key build in a try/catch, bad attributes now warn and drop instead of throwing, thanks

packages/node/src/client.ts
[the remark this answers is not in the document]
DanielVisca: good catch, shutdown now takes one deadline up front and passes the remainder to super, added the 600ms fake timer test, thanks

packages/core/src/metrics/config.ts
DanielVisca: Fixed in 87fb0fd0. `_mergeWindowBack` routes each non-colliding series from the failed window through `_admitNewSeries()`, enforcing `maxSeriesPerFlush` and the one-per-window cap warning before re-inserting; colliding keys merge in place (metrics/index.ts ~415).

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1534,6 +1535,55 @@ export abstract class PostHogCoreStateless {
...
+      })
+      return { kind: 'ok' }
+    } catch (err) {
+      if (isPostHogFetchContentTooLargeError(err)) {
+        return { kind: 'too-large' }
+      }
+      // Exhausted retries on a retryable failure (network error, 408/429/5xx)
+      // still classify as retry-later so the window rides the next flush; only
+      // non-retryable HTTP errors (and 413 above) drop the batch.
+      if (isPostHogFetchRetryableError(err)) {
```
dustinbyrne, line 1580: [P2 parity] Confirm the cross-SDK policy for HTTP 408

This transport uses `isPostHogFetchRetryableError`, so an exhausted HTTP 408 remains `retry-later` and the Node window is retained. The browser metrics adapter and Python implementation classify only network errors, 429, and 5xx as retryable, meaning both currently treat 408 as fatal and drop the window.

Treating 408 as retryable seems reasonable, but the three SDK surfaces should agree. Can we choose one policy, add a 408 regression case, and track the corresponding browser/Python alignment rather than silently shipping different delivery behavior?

dustinbyrne (commented):
Human-driven, agent-assisted follow-up review.

dustinbyrne (approved):
nice!

PR #4123, opened 2026-07-09
feat(mcp): stable sessions and client metadata on stateless / multi-pod MCP servers

### LINKED ISSUE #68635: Bug(MCP Analytics): $session_id not stable across multiple MCP server instances
### Bug Description

`$session_id` isn't stable when the MCP server runs as more than one instance (e.g. multiple pods behind a load balancer). Tool calls from the same client get scattered across instances, and each produces its own `$session_id` instead of sharing one. Without `identify`, `distinct_id` fragments the same way. This happens in stateless / multi-instance setups where the client re-initializes and no `mcp-session-id` is persisted — the SDK ties the session to the transport session id or to per-instance memory, neither of which is shared across instances.

## How to reproduce

1. Run 2 instances of an instrumented MCP server behind a round-robin load balancer (e.g. docker compose with 2 MCP containers and nginx)
2. Make a few tool calls from any client (e.g. Claude)
3. Inspect the captured events — each call carries a different $session_id, split across the two instances.

## Additional context

Sticky routing (`sticky sessions`) works around it but breaks on pod restarts/rebalancing, so it isn't a real fix. Would help to have a first-class way to supply a stable session id per request (e.g. derived from the auth principal or a client-provided id).

Mega issue: https://github.com/PostHog/posthog/issues/64016
https://posthog.slack.com/archives/C0B3E1Y576X/p1783434626337389

### Debug info

```shell
```

gesh:
## Problem

On **stateless** MCP servers (a fresh `Server` + transport per request — serverless, Cloudflare Workers, horizontally-scaled pods) every request became its own session, and `$mcp_client_name` / `$mcp_client_version` went missing after `initialize` (they're only sent in the init body, and `getClientVersion()` is empty on any instance that didn't handle init). A real customer saw `sessions == tool_calls` and a 100% "Other" client breakdown.

Fixes: https://github.com/PostHog/posthog/issues/68635

## Approach

The only value a spec-compliant client round-trips on **every** request is the `Mcp-Session-Id` header. So at `initialize` we mint it as a self-encoded token — `base64url(JSON{ sid, cn, cv })` — carrying the PostHog session id plus the client name/version. Clients replay it on every subsequent request, so **any pod decodes it and recovers the same `$session_id` + client metadata with no server-side store, no sticky routing, and no client changes.**

- `getSessionId` decodes the replayed token (stateless), or falls back to the transport session id (stateful) / an inactivity-rolling generated id.
- Auto-minting requires `enableJsonResponse: true` (response headers are built after the handler runs). SSE-mode servers set the header at the HTTP layer with the new `encodeSessionId` export; the SDK still decodes it.
- Fully backward compatible — stateful transports and stdio are untouched.

## New exports

`encodeSessionId`, `decodeSessionId`, `MCP_SESSION_HEADER`, `SessionTokenPayload`, `newSessionId`.

## Validation

- Unit tests: token codec round-trip, mint-on-initialize, decode-on-replay — **380 tests green**.
- Manually validated end-to-end on a single stateless mcp-nest server (JSON mode) **and a 2-pod docker-compose cluster behind round-robin nginx (no stickiness)** — `initialize` on pod-1, tool calls round-robined to pod-2, which recovered the same `$session_id` + client metadata despite never seeing the handshake.

## Follow-ups (separate PRs)

- Carry the resolved identity in the token (consistent person cross-pod + identify dedup).
- Filter transport-probe clients (e.g. `mcp-remote-fallback-test`).
- Attribute `tools/list` events to the identified person.

### DIFF
--- a/.changeset/mcp-stateless-session-token.md
+++ b/.changeset/mcp-stateless-session-token.md
@@ -1,0 +1,9 @@
+---
+'@posthog/mcp': minor
+---
+
+feat(mcp): stable sessions and client metadata on stateless / multi-pod MCP servers
+
+On stateless servers every request became its own session and `$mcp_client_name`/`$mcp_client_version` were missing after `initialize`. The SDK now mints the `Mcp-Session-Id` response header at `initialize` as a token carrying the session id and client name/version; clients replay it on every request, so any pod recovers both with no server-side store. Auto-minting requires `enableJsonResponse: true` on `StreamableHTTPServerTransport`; SSE-mode servers can set the header at the HTTP layer with the new exports.
+
+New exports: `encodeSessionId`, `decodeSessionId`, `MCP_SESSION_HEADER`, `SessionTokenPayload`, `newSessionId`.
--- a/packages/mcp/README.md
+++ b/packages/mcp/README.md
@@ -13,11 +13,33 @@ SDK usage examples and code snippets live in the official documentation so they 
 On stateless deployments the SDK mints the `Mcp-Session-Id` response header at `initialize`
 as a token carrying the session id and client name/version. Clients replay the header on
 every request, so any pod keeps `$session_id` and `$mcp_client_name`/`$mcp_client_version`
-stable with no server-side store. Works out of the box on `StreamableHTTPServerTransport`
-with `enableJsonResponse: true` and a fresh transport per request.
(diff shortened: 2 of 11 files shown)


pauldambra (approved):
elegant!

PR #4125, opened 2026-07-09
feat(replay): add session_recording.sampling for mousemove capture

### LINKED ISSUE #3982: Feature Request: replay disable mouse tracking

Add the ability to opt out of mouse tracking for replay, rrweb already exposes a mousemove and mouseInteraction sampling controls.

use rrwebs mousemove and mouseinteraction sampling controls

## Additional context
From: https://posthoghelp.zendesk.com/agent/tickets/59830 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/61149)

### Debug info

```shell
- [ ] PostHog Cloud, Debug information: [please copy/paste from https://us.posthog.com/settings/project-details#variables]
```

DerGeraetK:
## Problem

Closes #3982

Customers want to keep mouse movement out of session replay recordings, both for privacy and to cut recording payload size. rrweb already exposes sampling controls for this (`mousemove`, `mouseInteraction`), but posthog-js does not pass them through: the recorder's allowlist of rrweb options has no `sampling` key, and the canvas branch assigns `sampling` wholesale.

## Changes

- New narrow typed option `session_recording.sampling?: { mousemove?: boolean | number; mouseInteraction?: boolean }` (interface `SessionRecordingSamplingConfig` in `@posthog/types`, following the locally duplicated rrweb subset pattern of `MaskInputOptions` / `SlimDOMOptions`).
- The recorder allowlist passes only the `mousemove` and `mouseInteraction` keys through to `rrweb.record` (runtime filter, so untyped callers cannot smuggle in `canvas`, `scroll`, etc.).
- Canvas recording now merges its `canvas` fps into the sampling object instead of overwriting it, so user sampling and canvas recording coexist. The previous hard assignment would have discarded any other sampling key.

**Why this API shape:** it is what the issue triage suggested (a narrow typed `session_recording.sampling` option), and it follows the existing convention that rrweb passthrough options keep their rrweb names (`maskAllInputs`, `slimDOMOptions`, ...). The `canvas` key is deliberately not exposed because canvas fps is owned by the canvas recording config. The JSDoc explicitly disambiguates from `sampleRate` (controls whether a session is recorded at all) and from `posthog.startSessionRecording({ sampling: true })`. Happy to adjust naming or scope if the team prefers a different shape.

**Behavioral note:** with `mousemove: false`, rrweb also stops touchmove/drag capture, and pure mouse movement no longer counts as user activity for replay idle detection (clicks, scrolls and inputs still do). Both are documented in the JSDoc; the idle sources themselves are deliberately unchanged.

**Test strategy:** 6 new unit tests in `lazy-sessionrecording.test.ts` assert at the `rrweb.record` mock boundary: passthrough of both keys, numeric throttle, falsy `mousemove: 0`, merge with canvas sampling (regression test for the overwrite fix), filtering of non-allowlisted keys, and unchanged behavior when not configured. The existing exact-match default-options test stays untouched as a regression guard. Locally green: replay unit suite (656/656), `pnpm typecheck`, prettier.

Docs follow-up: I will open a separate posthog.com PR for the session replay docs once this is merged.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Planned and implemented with Claude Code. The repo was researched before writing any code: the recorder's allowlist mechanism, the existing rrweb subset type precedents (`MaskInputOptions`, `SlimDOMOptions`), the canvas sampling assignment, and the three existing "sampling" concepts in the codebase. A raw full `sampling` passthrough and a bespoke option name were considered and rejected (canvas ownership, naming collision with `sampleRate`). Code and tests were independently re-reviewed against the actual diff; unit tests, typecheck and formatting were run locally (Windows). I directed the work, reviewed the changes, and can explain every line.

### DIFF
--- a/.changeset/replay-mouse-sampling.md
+++ b/.changeset/replay-mouse-sampling.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+Add `session_recording.sampling` to disable or throttle mousemove capture (and optionally mouseInteraction) in session replay. Canvas recording now merges its canvas sampling with user-provided sampling instead of overwriting it.
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -3837,6 +3837,124 @@ describe('Lazy SessionRecording', () => {
         })
     })
 
+    describe('sampling passthrough', () => {
+        it('passes user sampling for mousemove and mouseInteraction to rrweb.record', () => {
+            posthog.config.session_recording.sampling = { mousemove: false, mouseInteraction: false }
+
+            sessionRecording.onRemoteConfig(
+                makeFlagsResponse({
+                    sessionRecording: {
+                        endpoint: '/s/',
+                    },
+                })
+            )
+
+            expect(assignableWindow.__PosthogExtensions__.rrweb.record).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    sampling: { mousemove: false, mouseInteraction: false },
+                })
+            )
+        })
+
+        it('passes a falsy numeric mousemove value of 0 to rrweb.record', () => {
+            posthog.config.session_recording.sampling = { mousemove: 0 }
+
+            sessionRecording.onRemoteConfig(
+                makeFlagsResponse({
+                    sessionRecording: {
+                        endpoint: '/s/',
+                    },
+                })
+            )
+
+            expect(assignableWindow.__PosthogExtensions__.rrweb.record).toHaveBeenCalledWith(
+                expect.objectContaining({
+                    sampling: { mousemove: 0 },
+                })
+            )
+        })
+
+        it('passes a numeric mousemove throttle to rrweb.record', () => {
+            posthog.config.session_recording.sampling = { mousemove: 250 }
+
+            sessionRecording.onRemoteConfig(
+                makeFlagsResponse({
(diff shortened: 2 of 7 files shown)


dustinbyrne (commented):
this change looks clean to me - @PostHog/team-replay any considerations before we merge?

TueHaulund (approved):

DerGeraetK:
The failing `config-snapshot` unit test is fixed in d19a41b (snapshot regenerated for the new `sampling` field, diff is exactly the 16 added lines). The CI workflows on that commit are awaiting maintainer approval to run.

arnohillen (approved):
Robot review:

The allowlist is the right shape, and merging canvas sampling instead of overwriting it fixes a real bug. Backwards compatible, since `{...undefined, canvas: fps}` equals the old value. The idle-detection note in the doc comment is the part people would otherwise get wrong.

Nit: a config with only non-allowlisted keys still assigns `sampling: {}` rather than leaving it undefined. Harmless, rrweb treats them the same.

marandaneto:
@PostHog/team-replay should this be merged? its approved but stale now

PR #4127, opened 2026-07-10
fix: don't overwrite registered $referrer with document.referrer

sarmah-rup:
## Problem

`$referrer` / `$referring_domain` set with `posthog.register()` are dropped on every capture in SPA and iframe setups. In an iframe the reported value becomes the iframe's own origin instead of the value the user registered.

Root cause: `register()` writes these to the main `persistence`, but `save_referrer` writes `document.referrer` into `sessionPersistence` via `update_referrer_info()`. In `calculateEventProperties()` the `sessionPersistence` props are spread after the main persistence, so they shadow the user's registered value on every event.

Fixes #[67448](https://github.com/PostHog/posthog-js/issues/4208)

## Changes

`update_referrer_info()` is left as-is and still populates the session referrer from `document.referrer` on every capture. The precedence is resolved per property at the `calculateEventProperties()` merge instead: for `$referrer` / `$referring_domain`, a value present in the regular persistence (i.e. one the user set via `register()`) wins over the session value.

This is the approach @dustinbyrne suggested on the earlier, now-stale PR PostHog/posthog#4052 (resolve precedence per key at the final merge rather than skipping the session update). It also handles the two edge cases raised there:

- A session referrer already stored before `register()` is called no longer shadows the registered value.
- Registering only one of the two keys does not suppress the other, which still falls back to `document.referrer`.

## Testing

Added unit tests in the existing `referrer` suite covering:

- a registered `$referrer` / `$referring_domain` surviving subsequent captures,
- `register()` called after an initial capture winning over the already-stored session value,
- a single registered key while the other stays automatic,
- `unregister()` restoring the `document.referrer` fallback.

All referrer unit tests pass (8/8, including the 4 new cases), and the full `posthog-core.test.ts` suite passes (35/35). Changeset included.

### DIFF
--- a/.changeset/referrer-register-guard.md
+++ b/.changeset/referrer-register-guard.md
@@ -1,0 +1,5 @@
+---
+"posthog-js": patch
+---
+
+Don't let save_referrer overwrite a $referrer / $referring_domain that was explicitly set via posthog.register(), so registered attribution values survive pageviews in SPA and iframe contexts
--- a/packages/browser/src/__tests__/posthog-core.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.test.ts
@@ -289,6 +289,65 @@ describe('posthog core', () => {
                 expect(laterProperties['$referrer']).toBe('https://blabla.fr/')
                 expect(laterProperties['$referring_domain']).toBe('blabla.fr')
             })
+
+            it('should let a register() called after a capture win over the already-stored session referrer', () => {
+                // arrange: a first capture stores document.referrer in session persistence
+                mockReferrer.mockReturnValue('https://iframe-origin.example.com/some/path')
+                const { posthog, beforeSendMock } = setup()
+                posthog.capture(eventName, eventProperties)
+                const { properties: firstProperties } = beforeSendMock.mock.calls[0][0]
+                expect(firstProperties['$referrer']).toBe('https://iframe-origin.example.com/some/path')
+
+                // act: register a custom referrer after the session value is already present
+                posthog.register({
+                    $referrer: 'https://blabla.fr/',
+                    $referring_domain: 'blabla.fr',
+                })
+                posthog.capture(eventName, eventProperties)
+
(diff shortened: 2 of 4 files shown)


sarmah-rup:
@dustinbyrne this implements the per-property precedence-at-merge approach you suggested on PostHog/posthog#4052 (which has gone stale), with the register-after-capture, single-key, and unregister() cases covered. Would appreciate a look when you have a moment.

sarmah-rup:
Bumping to keep this open - still relevant. 

Quick recap of the impact: if you register ({ $referrer }) yourself (we do this to preserve first-touch attribution), posthog-js overwrites it with document.referrer on the next capture, so the manually-set value is silently lost. This PR applies the per-property precedence-at-merge approach @dustinbyrne suggested on PostHog/posthog#4052, with the register-after-capture, single-key, and unregister() cases covered.

Checks are green and a changeset is included. 

@dustinbyrne  / @turnipdabeets  - could one of you take a look when you have a moment? Happy to rebase if anything's drifted.

turnipdabeets:
> Bumping to keep this open - still relevant.

Thanks for the tag and for addressing Dustin's suggestions! Generally LGTM. Looks like you may need a rebase to make CI happy. We also need verified signatures on your commits to merge. 

There's one caveat but it may be fine. Visitors still carrying a $referrer the SDK itself persisted before v1.53 (April 2023) will flip from per-session referrer back to that frozen first-touch value. I think it can be acceptable because it's their original pre-1.53 behavior with genuine data, the cohort is small and only shrinks, and unregister('$referrer') clears it. Curious what you think @dustinbyrne ?

sarmah-rup:
Rebased onto latest main and the commits are now signed and verified - CI should be green. Thanks @turnipdabeets ! Let me know if anything else is needed.

ioannisj (approved):

Clemog:
Thanks ! 🙏

Issue #3765, opened 2026-06-08
Track upstream rrweb changes to pull in


pauldambra:
## Track upstream rrweb changes to pull in

We vendor a fork of [`rrweb-io/rrweb`](https://github.com/rrweb-io/rrweb) under `packages/rrweb/*` (moved into this monorepo from the retired `PostHog/posthog-rrweb` in #3510). This issue tracks upstream changes worth pulling into our fork, and replaces the old split tracking in `PostHog/posthog-rrweb#155` (open PRs) and `#147` (merged PRs), which are now stranded in the old repo.

**Fork point:** upstream `2.0.0-alpha.18` (2024-12-06).
**Upstream now:** `2.1.0` (2026-06-27). The `2.0.0` stable release mostly re-published pre-`alpha.18` PRs we *already have*, `2.0.1` added only one fix (#1854), and `2.1.0` added only #1864 (rrweb) + #1865 (rrweb-snapshot) — all already triaged in the cherry-pick table below. So the genuinely-new surface = PRs **merged after 2024-12-06** plus the currently-open PRs.

**Prioritisation lens:** recording performance (CPU/memory cost on the *customer's* page while recording) outranks everything else — replay fixes, masking gaps, and framework compat all queue behind it. Recording-perf items are ranked first in the adoption plan (see comments).

This is a comprehensive scan: every merged-since-fork and open PR was triaged, and each "adopt"/"already-have" call was verified by grepping our fork. Items marked `*` are **PostHog-authored upstream PRs** that aren't in our fork yet — adopting them early means vendoring our own not-yet-landed work; they're also tracked in the contribute-back issue (#3766).

### Trustworthiness criteria
- **High** — core maintainers (Juice10, eoghanmurray, juliecheng, YunFeng0817), PostHog authors (pauldambra, d0ugal, daibhin, megboehlert, billyvg, benjackwhite, marandaneto, kevinatown), or APPROVED PRs.
- **Medium** — known/repeat contributors, or PRs with substantive review.
- **Low** — unknown/community/no-review (validate before adopting).

---

## Merged upstream since our fork — clean cherry-picks (not yet in our fork)

| PR | Author | Trust | Theme | Priority | Why it matters |
|----|--------|-------|-------|----------|----------------|
[...]

PR #4130, opened 2026-07-11
perf(rrweb): keep re-added nodes in DOM order in mutation addedSet (adopt upstream #1302)

pauldambra:
## Problem

Adopts upstream rrweb [#1302](https://github.com/rrweb-io/rrweb/pull/1302), tracked in #3765 (batch-1, recording performance).

`genAdds` skips nodes already in `addedSet`, leaving them at their original position in the set. The emit phase iterates the set in insertion order and pays for every node whose parent hasn't been emitted yet via its deferral list — on large batches where nodes are removed and re-added (virtual-DOM re-renders, list reordering), that stale ordering makes mutation processing markedly more expensive on the customer's page.

## Changes

- When `genAdds` re-encounters a node already in `addedSet`, it re-inserts it (delete + add) so the set's iteration order tracks the latest DOM order. 7 lines, record-path only.
- The integration test snapshot is updated: adds within a single mutation batch are now sequenced differently (and node ids shift accordingly), but every add carries `parentId`/`nextId` so the rebuilt DOM is identical — the replayer suite passes unmodified.

Refs #3765.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code (covered by existing suites: record 34/34, shadow-dom-manager 5/5, integration 50/50, replayer 42/42)

### If releasing new changes

## 🤖 Agent context

Ported as part of the batch-1 adoption plan on #3765 (PostHog Code session). Verified the emit-order change is the only observable difference by re-running the full record/integration/replayer suites; the single integration snapshot diff was inspected by hand (same nodes, same parent/next links, different batch sequence) before updating it.

---

### DIFF
--- a/packages/rrweb/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/rrweb/src/record/mutation.ts
@@ -799,7 +799,7 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n)) {
+    if (this.addedSet.delete(n)) {
       // re-insert n so the addedSet iteration order in the `emit` phase matches
       // the latest DOM order; a stale position makes emit's out-of-order
       // deferral list do far more work on large batches (upstream rrweb #1302).
@@ -807,7 +807,6 @@ export default class MutationBuffer {
       // positions and aren't re-appended here, but emit's addList deferral
       // re-derives parentId/nextId, so parent-before-child order still comes
       // out correct.
-      this.addedSet.delete(n);
       this.addedSet.add(n);
       return;
     }
--- a/packages/rrweb/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/rrweb/src/record/mutation.ts
@@ -799,7 +799,7 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.delete(n)) {
+    if (this.addedSet.has(n)) {
       // re-insert n so the addedSet iteration order in the `emit` phase matches
       // the latest DOM order; a stale position makes emit's out-of-order
       // deferral list do far more work on large batches (upstream rrweb #1302).
@@ -807,6 +807,7 @@ export default class MutationBuffer {
       // positions and aren't re-appended here, but emit's addList deferral
       // re-derives parentId/nextId, so parent-before-child order still comes
       // out correct.
+      this.addedSet.delete(n);
       this.addedSet.add(n);
       return;
     }
--- a/packages/rrweb/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/rrweb/src/record/mutation.ts
@@ -802,7 +802,11 @@ export default class MutationBuffer {
     if (this.addedSet.has(n)) {
       // re-insert n so the addedSet iteration order in the `emit` phase matches
       // the latest DOM order; a stale position makes emit's out-of-order
-      // deferral list do far more work on large batches (upstream rrweb #1302)
+      // deferral list do far more work on large batches (upstream rrweb #1302).
+      // this only moves n itself - already-present children keep their earlier
+      // positions and aren't re-appended here, but emit's addList deferral
+      // re-derives parentId/nextId, so parent-before-child order still comes
+      // out correct.
       this.addedSet.delete(n);
       this.addedSet.add(n);
       return;
--- a/packages/rrweb/rrweb/src/record/mutation.ts
+++ b/packages/rrweb/rrweb/src/record/mutation.ts
@@ -799,7 +799,15 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.has(n)) {
+      // re-insert n so the addedSet iteration order in the `emit` phase matches
+      // the latest DOM order; a stale position makes emit's out-of-order
+      // deferral list do far more work on large batches (upstream rrweb #1302)
+      this.addedSet.delete(n);
+      this.addedSet.add(n);
+      return;
+    }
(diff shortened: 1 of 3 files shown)


packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -799,7 +799,15 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.has(n)) {
+      // re-insert n so the addedSet iteration order in the `emit` phase matches
+      // the latest DOM order; a stale position makes emit's out-of-order
+      // deferral list do far more work on large batches (upstream rrweb #1302)
+      this.addedSet.delete(n);
```
pauldambra, line 810: > [!NOTE]

The early `return` fires before the child recursion, so only `n` moves to the set's tail — its already-present children keep their earlier positions and can iterate *before* their re-added parent. That's correct (emit's `addList` deferral re-derives parentId/nextId and fixes parent-before-child, exactly what upstream #1302 relies on), but it's the kind of thing a future reader might "fix" into a deep re-add and pay a real cost. Consider a half-sentence in the comment: children aren't re-appended; emit's deferral still fixes parent order. Not blocking.

packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -799,7 +799,15 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.has(n)) {
```
pauldambra, line 802: > [!NOTE]

nit picking / observability hat on — the whole optimisation is justified by "does far more work on large batches", but nothing measures the deferral-list cost, so the claim is taken on faith. if there's an upstream benchmark or issue thread to cite in the changeset, that would make it auditable later; and if a representative large-batch fixture exists, a quick before/after timing would back the "perf" in the title. ship as you see fit.

pauldambra (commented):
> [!NOTE]

QA Swarm review complete. See inline comments and the summary comment.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — skill not available in this environment)

Clean, minimal, faithful port of upstream rrweb #1302; no correctness or security issues found. The remaining findings are documentation/test-coverage suggestions.

### Key findings

- 🟡 **MEDIUM (convergent, test coverage)** — the correctness story rests entirely on a regenerated integration snapshot, which proves emit *order* changed but not the invariant the change depends on (rebuilt DOM equivalence / adds in DOM order). A small focused test — re-encounter a node in `genAdds`, assert adds come out in DOM order or that a rebuild matches — would lock in the property against future refactors of the emit deferral. (inline-commentable only as "general", so recorded here)
- 🟡 **MEDIUM (convergent, inline)** — the reorder is shallow (children keep earlier set positions); suggest extending the code comment so nobody "fixes" it into a deep re-add. See inline comment.
- ⚪ **NIT (convergent, inline)** — the perf claim is unmeasured; cite the upstream benchmark/issue or do a quick before/after on a large-batch fixture. See inline comment.

### Convergence

Both paul and xp independently flagged the shallow-reorder subtlety and the unmeasured perf claim; both independently verified the snapshot node-by-node and confirmed the rebuilt DOM is equivalent (same ids, same content, parentId/nextId anchors intact).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 👤 paul | solid little perf port, comfortable approving; read the emit phase end-to-end and the claim checks out; soft notes on observability only |
| 📐 xp | clean, minimal, faithful; scores well on all four rules of simple design; one nudge — a focused test for the DOM-order invariant |
| 🛡 security-audit | no findings — masking/blocking is applied per-node at serialization time and cannot be affected by iteration order; order-sensitive emit paths fail closed (fidelity loss, not disclosure) |
| 🔍 qa-team | skipped (skill unavailable) |

---

packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -799,7 +799,18 @@ export default class MutationBuffer {
     if (this.processedNodeManager.inOtherBuffer(n, this)) return;
 
     // if n is added to set, there is no need to travel it and its' children again
-    if (this.addedSet.has(n) || this.movedSet.has(n)) return;
+    if (this.addedSet.delete(n)) {
```
pauldambra: > [!NOTE]

Both reviewers independently question the simplify-pass change from `has(n)` to `if (this.addedSet.delete(n))`. It is correct (`Set.delete` returns true iff present) and saves one hash lookup — but (1) it diverges from upstream #1302's literal has→delete→add shape, creating a permanent conflict point on this hot line for every future rrweb re-sync of a vendored fork, (2) the guard now reads as "remove n" when the intent is "already present → reposition", and (3) the saving is unmeasurable — genAdds' cost is the tree walk and serialization, not one O(1) hash op. Recommendation: revert to upstream's exact shape unless a profile shows this lookup matters.

packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -799,7 +799,18 @@ export default class MutationBuffer {
...
+      // re-insert n so the addedSet iteration order in the `emit` phase matches
+      // the latest DOM order; a stale position makes emit's out-of-order
+      // deferral list do far more work on large batches (upstream rrweb #1302).
+      // this only moves n itself - already-present children keep their earlier
```
pauldambra, line 806: > [!NOTE]

**[xp]** 🟢 LOW

Re-raising from round 1: nothing pins the reorder behaviour itself — a focused test (large batch, node re-encountered in genAdds, assert emitted add order / rebuilt-DOM equivalence) would let the next person re-sync or re-simplify this line without fear. Counterpoint worth weighing: the regenerated integration snapshot *is* order-sensitive and fails if the reorder regresses, so it may already serve as the pin.

pauldambra (commented):
> [!NOTE]

QA Swarm round-2 review of the new commits (comment extension + simplify pass). See inline.

arnohillen (approved):

PR #4133, opened 2026-07-12
fix(web): make jsonStringify circular-safe so event serialization never throws

### LINKED ISSUE #4132: jsonStringify throws on circular references (DOM nodes with React fibers); with capture_exceptions the throw is recaptured in a loop
# `jsonStringify` throws on circular references; with `capture_exceptions` the throw is recaptured in a loop

## Environment
- `posthog-js` 1.396.5 (also present on `main` / 1.399.2)
- React 17, Chrome
- `capture_exceptions: true`, autocapture enabled

## Symptom
A burst of ~100 identical `$exception` events in ~2 seconds, all `handled: true`, with value:

```
'TypeError' captured as exception with message:
'Converting circular structure to JSON
  --> starting at object with constructor 'HTMLAnchorElement'
  | property '__reactFiber$…' -> object with constructor 'FiberNode'
  --- property 'stateNode' closes the circle'
```

The frames resolve into the SDK's request/serialization path, not application code.

## Root cause
`jsonStringify` in `packages/browser/src/request.ts` — the serializer every request body passes through — only guards `BigInt`:

```ts
return JSON.stringify(data, (_, value) => (typeof value === 'bigint' ? value.toString() : value), space)
```

When a captured property holds a circular value, `JSON.stringify` throws `Converting circular structure to JSON`. The most common source on a React app is a DOM node: React attaches `__reactFiber$…` / `__reactProps$…` to elements, and a fiber's `stateNode` points back at the element, so any element reference that reaches event properties is circular.

## Why it loops
With `capture_exceptions` enabled, the `TypeError` thrown by the serializer is itself autocaptured as a new `$exception`. If the original event is retried, or the condition recurs across elements, the SDK repeatedly reports its own serialization failure — turning one unserializable property into a flood of self-referential exceptions.

## Minimal repro
```js
const el = document.querySelector('a')      // any React-rendered element
posthog.capture('x', { el })                // el is circular via its React fiber
// → TypeError: Converting circular structure to JSON, recaptured as $exception
```

## Fix
Make `jsonStringify` fall back to a circular-safe pass when the fast path throws, so serialization never throws. PR attached. The fast (non-circular) path is unchanged, so shared-but-acyclic references keep their real values; only a genuine back-reference becomes `"[Circular]"`.

Two deeper follow-ups worth considering, out of scope for the fix:
1. Exclude errors originating from the SDK's own capture/serialization path from `capture_exceptions`, to prevent self-referential loops in general.
2. Prevent raw DOM nodes from entering event properties in the first place.

mikenicholls88:
## Problem

`jsonStringify` in `packages/browser/src/request.ts` — the serializer every request body passes through — only guards `BigInt`. When a captured property holds a circular value, `JSON.stringify` throws `Converting circular structure to JSON`.

The common trigger on a React app is a DOM node: React attaches `__reactFiber$…` / `__reactProps$…` to elements, and a fiber's `stateNode` points back at the element, so any element reference that reaches event properties is circular. With `capture_exceptions` enabled, the `TypeError` the serializer throws is itself autocaptured as a new `$exception` — and when the condition recurs it turns into a burst of `handled` self-referential exceptions. We saw ~100 identical `$exception`s in a ~2 second window in production.

Fixes #4132.

## Changes

`jsonStringify` now falls back to a circular-safe replacer **only when the fast path throws**, so serialization always succeeds instead of throwing:

- The common (non-circular) path is byte-for-byte unchanged, so shared-but-acyclic references keep their real values.
- Only a genuine back-reference is replaced with `"[Circular]"`, matching the convention already used in `entrypoints/logs.ts`.
- `BigInt` handling is preserved (extracted to a shared replacer used by both paths).

No public API change; not a breaking change.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

The bug was found while investigating an exception-tracking burst on a production React (17) / Next.js site running posthog-js: ~100 `handled` `$exception`s within ~2s, all `Converting circular structure to JSON` starting at an `HTMLAnchorElement` via its `__reactFiber$…` back-reference. Tracing the failing frame led to `jsonStringify` in `request.ts`, whose replacer guards `BigInt` but not circular references.

Approach considered and chosen: a whole-tree circular-safe replacer (WeakSet) on every call was rejected because it changes serialization of shared-but-acyclic references (a shared object would serialize as `"[Circular]"` on its second visit). Instead the fast path is left untouched and the circular-safe pass only runs after a throw, so normal payloads are unaffected and only genuinely-circular data is rewritten.

Authored with Claude Code (Claude). Directed and owned by @mikenicholls88.

### Test strategy

Added two cases to `request.test.ts` (body encoding) and ran the suite for the changed file:

```
pnpm --filter posthog-js exec jest src/__tests__/request.test.ts
→ Test Suites: 1 passed; Tests: 57 passed (55 existing + 2 new)
```

- **plain circular object** — asserts the exact `{"foo":"bar","self":"[Circular]"}` output (no throw).
- **circular DOM node** — a node whose `__reactFiber.stateNode` points back at the node (the shape that triggered this in production); asserts serialization does not throw and the request still sends.

Edge cases considered: `BigInt` handling is unchanged (existing test still passes); shared-but-acyclic references are unaffected because the fallback only runs after the fast path throws — verified separately that `{ a: shared, b: shared }` keeps both real values rather than emitting `[Circular]`.

Not done: I did not run the full monorepo suite or smoke-test in a browser example app. The change is confined to `jsonStringify` and covered by the unit tests above.

### DIFF
--- a/packages/browser/src/__tests__/request.test.ts
+++ b/packages/browser/src/__tests__/request.test.ts
@@ -713,6 +713,53 @@ describe('request', () => {
                     'application/x-www-form-urlencoded'
                 )
             })
+
+            it('does not throw on circular references and serializes them as [Circular]', () => {
+                const circular: any = { foo: 'bar' }
+                circular.self = circular
+                request(
+                    createRequest({
+                        url: 'https://any.posthog-instance.com/',
+                        method: 'POST',
+                        data: circular,
+                    })
+                )
+                expect(mockedXHR.send.mock.calls[0][0]).toMatchInlineSnapshot(`"{"foo":"bar","self":"[Circular]"}"`)
+            })
+
+            it('does not throw when a property is a circular DOM node (e.g. a React fiber back-reference)', () => {
+                // Mimics a DOM element that retains a React fiber which points back at the element —
+                // exactly what makes plain JSON.stringify throw "Converting circular structure to JSON".
+                const el: any = { tagName: 'A', nodeType: 1 }
+                el.__reactFiber = { stateNode: el }
+                expect(() =>
+                    request(
+                        createRequest({
+                            url: 'https://any.posthog-instance.com/',
+                            method: 'POST',
+                            data: { $el: el },
+                        })
+                    )
+                ).not.toThrow()
+                expect(mockedXHR.send).toHaveBeenCalledTimes(1)
+            })
+
+            it('keeps shared-but-acyclic references while replacing only true cycles', () => {
+                const shared = { n: 1 }
+                const data: any = { a: shared, b: shared }
+                data.self = data // the only real cycle
+                request(
+                    createRequest({
+                        url: 'https://any.posthog-instance.com/',
+                        method: 'POST',
+                        data,
+                    })
+                )
+                const body = JSON.parse(mockedXHR.send.mock.calls[0][0] as string)
+                expect(body.a).toEqual({ n: 1 })
+                expect(body.b).toEqual({ n: 1 })
+                expect(body.self).toBe('[Circular]')
+            })
         })
 
         describe('sendBeacon', () => {
--- a/packages/browser/src/request.ts
+++ b/packages/browser/src/request.ts
@@ -14,6 +14,7 @@ import {
     isGzipRequest,
     isNativeAsyncGzipError,
     isNativeAsyncGzipReadError,
+    safeJsonStringify,
 } from '@posthog/core'
 
 interface RequestWithEncodedBody extends RequestWithOptions {
@@ -90,12 +91,19 @@ export const extendURLParams = (url: string, params: Record<string, any>, replac
 }
 
 export const jsonStringify = (data: any, space?: string | number): string => {
-    // With plain JSON.stringify, we get an exception when a property is a BigInt. This has caused problems for some users,
-    // see https://github.com/PostHog/posthog-js/issues/1440
-    // To work around this, we convert BigInts to strings before stringifying the data. This is not ideal, as we lose
-    // information that this was originally a number, but given ClickHouse doesn't support BigInts, the customer
-    // would not be able to operate on these numerically anyway.
-    return JSON.stringify(data, (_, value) => (typeof value === 'bigint' ? value.toString() : value), space)
+    try {
+        // Fast path: convert BigInts to strings, since plain JSON.stringify throws on them.
+        // See https://github.com/PostHog/posthog-js/issues/1440.
+        return JSON.stringify(data, (_, value) => (typeof value === 'bigint' ? value.toString() : value), space)
+    } catch {
+        // A self-referential value — most commonly a DOM node that retains a React fiber pointing back
+        // at the element — makes JSON.stringify throw "Converting circular structure to JSON". With
+        // exception autocapture enabled that throw was recaptured as a new $exception, sometimes in a
+        // tight loop. Fall back to the shared circular-safe serializer (which also handles BigInt and
+        // Errors); it replaces only true cycles with "[Circular]", leaving shared-but-acyclic
+        // references intact. `space` formatting is dropped on this rare path.
+        return safeJsonStringify(data)
+    }
 }
 
 const encodeToDataString = (data: string | Record<string, any>): string => {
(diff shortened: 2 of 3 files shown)


mikenicholls88:
Extra evidence from the production instance where this surfaced (posthog-js 1.396.5, React 17 / Next.js 12):

- **108 `$exception` events in a ~2 second window**, all `mechanism.handled: true`, from a single browser session.
- Identical value on every one:
  ```
  'TypeError' captured as exception with message: 'Converting circular structure to JSON
    --> starting at object with constructor 'HTMLAnchorElement'
    | property '__reactFiber$…' -> object with constructor 'FiberNode'
    --- property 'stateNode' closes the circle'
  ```
- The failing frames resolved into the SDK's request/serialization chunk, not application code.

The shape (identical error, one session, sub-second cadence) is the `capture_exceptions`-recaptures-its-own-throw loop — one unserializable property becomes a flood. A `before_send` hook dropping `/Converting circular structure to JSON/` stops it as a consumer-side workaround, but the fix in this PR removes the throw at the source.

packages/browser/src/request.ts
```diff
@@ -89,13 +89,40 @@ export const extendURLParams = (url: string, params: Record<string, any>, replac
...
+    } catch {
+        // A circular reference (e.g. a DOM node that retains a React fiber, or any other
+        // self-referential value that ends up in a captured property) makes JSON.stringify throw
+        // "Converting circular structure to JSON". That throw previously escaped the send path, and
+        // with exception autocapture enabled it would be recaptured as a new $exception — sometimes
+        // in a tight loop. Retry with a circular-safe replacer so serialization always succeeds
+        // rather than throwing. Only the offending back-reference is replaced with "[Circular]"; the
+        // common (non-circular) path above is unaffected, so shared-but-acyclic references keep their
+        // real values.
+        const seen = new WeakSet<object>()
```
turnipdabeets: Could we make the catch just `return safeJsonStringify(data)` from `@posthog/core`? It was added in #4027 to fix this same bug in posthog-node and already guards the Node/RN capture paths, so this keeps one circular-serialization convention across SDKs (it also handles BigInt, and flattens Errors).

packages/browser/src/__tests__/request.test.ts
```diff
@@ -713,6 +713,36 @@ describe('request', () => {
...
+                expect(() =>
+                    request(
+                        createRequest({
+                            url: 'https://any.posthog-instance.com/',
+                            method: 'POST',
+                            data: { $el: el },
+                        })
+                    )
+                ).not.toThrow()
+                expect(mockedXHR.send).toHaveBeenCalledTimes(1)
```
turnipdabeets, line 744: [suggestion] Worth a third case with a shared-but-acyclic sibling inside a circular payload (`const shared = {n: 1}; const data: any = {a: shared, b: shared}; data.self = data`) asserting `a` and `b` both keep their real values — that pins down the "only genuine back-references" behavior the PR description promises.

turnipdabeets (commented):
Thanks for your contribution! A few comments below plus we need to make lint happy on CI and all of your commits need to be signed to merge. Otherwise LGTM

mikenicholls88:
Thanks for the review — all addressed:

- **`safeJsonStringify`**: the `catch` now returns `safeJsonStringify(data)` from `@posthog/core` and I dropped the local `WeakSet`. Good call — its ancestor-chain approach handles the shared-but-acyclic case correctly, which the previous `WeakSet` version would have gotten wrong, and it keeps one circular-serialization convention across SDKs.
- **Third test**: added `keeps shared-but-acyclic references while replacing only true cycles` — `{ a: shared, b: shared }` with a `self` cycle, asserting `a`/`b` both keep their real values and only `self` becomes `"[Circular]"`.
- **Lint**: `eslint src/request.ts src/__tests__/request.test.ts` is clean.
- **Tests**: `jest src/__tests__/request.test.ts` → 58 passed.
- **Commits**: squashed to a single SSH-signed commit (registering the signing key on my account so it verifies).

turnipdabeets:
@mikenicholls88 commits are signed but showing not verified, mind force pushing after registering on GitHub?

turnipdabeets (approved):
@mikenicholls88 thanks for your contribution! One request below and also could you

PR #4140, opened 2026-07-13
fix(browser): split over-quota beacons and fall back to fetch instead of dropping silently

turnipdabeets:
## Problem

`sendBeacon` returns `false` when a payload would exceed the page's shared ~64KiB in-flight keepalive quota ([Fetch spec](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch), [Beacon spec](https://w3c.github.io/beacon/#sec-processing-model)). The SDK ignored that return value, so an over-quota unload batch was dropped with no signal — no log, no callback, no retry. Follow-up to #4139, where base64 encoding (~33% larger bodies) made this slightly easier to hit.

## Changes

- `_sendBeacon` now checks the return value. On rejection:
  - a batch is split in half and each piece re-sent recursively, so the browser (the only reliable size oracle) re-judges each piece and as much as fits gets delivered;
  - below a 16KiB floor a rejection means the quota is exhausted rather than the payload being too big, so instead of splitting further the payload falls back to a non-keepalive fetch (keepalive would draw on the same exhausted quota) and logs a warning.
- Measured behavior backing the design: Chromium accepts 5×30KB beacons spaced out but only the first 2 fired synchronously; a single 70KB blob is rejected outright.

Harness A/B (30×~3KB events, compression inactive, bounce at 400ms): previous behavior delivers **0/30** (single rejected beacon, silent); with this change **30/30** arrive across 13 CORS-simple requests. On a slow network the floor is "first ~64KB delivers, in queue order" — events are drained before session recordings, so the most valuable data goes first.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Built with Claude Code, directed by Anna following review discussion on #4139. Chose reactive splitting (react to `sendBeacon` returning `false`) over proactive size estimation so the browser's quota accounting stays the single source of truth; considered and rejected a size-conditional `application/json` fallback to preserve the "beacon never preflights" invariant from #4139. Verified with unit tests plus a two-origin harness A/B against the pre-change bundle.

### DIFF
--- a/packages/browser/src/request.ts
+++ b/packages/browser/src/request.ts
@@ -10,6 +10,7 @@ import { gzipSync, strToU8 } from 'fflate'
 import { _base64Encode } from './utils/encode-utils'
 import {
     gzipCompress,
+    isArray,
     isGzipData,
     isGzipRequest,
     isNativeAsyncGzipError,
@@ -290,7 +291,7 @@ const xhr = (options: RequestWithOptions) => {
     req.send(body)
 }
 
-const _fetch = (options: RequestWithOptions) => {
+const _fetch = (options: RequestWithOptions & { _keepaliveDisabled?: boolean }) => {
     const encodedRequest = encodeRequest(options)
     if (!encodedRequest) {
         return
@@ -342,7 +343,9 @@ const _fetch = (options: RequestWithOptions) => {
         // so let's get the best of both worlds and only set keepalive for POST requests
         // where the body is less than 64kb
         // NB this is fetch keepalive and not http keepalive
-        keepalive: options.method === 'POST' && (estimatedSize || 0) < KEEP_ALIVE_THRESHOLD,
+        // _keepaliveDisabled: a beacon-rejected payload would fail a keepalive fetch too (shared quota)
+        keepalive:
+            options.method === 'POST' && !options._keepaliveDisabled && (estimatedSize || 0) < KEEP_ALIVE_THRESHOLD,
         body,
         signal: aborter?.signal,
         ...options.fetchOptions,
@@ -389,13 +392,16 @@ const _fetch = (options: RequestWithOptions) => {
     return
 }
 
+// below this size a rejection means the shared quota is exhausted, not that the payload is too big
+const BEACON_SPLIT_FLOOR_BYTES = 16 * 1024
+
 const _sendBeacon = (options: RequestWithOptions) => {
     // beacon documentation https://w3c.github.io/beacon/
     // beacons format the message and use the type property
 
     try {
         const { url, encodedBody } = encodePostDataSafely(options)
-        const { contentType, body } = encodedBody ?? {}
+        const { contentType, body, estimatedSize } = encodedBody ?? {}
         if (!body) {
             return
         }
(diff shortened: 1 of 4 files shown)


packages/browser/src/request.ts
```diff
@@ -389,21 +392,38 @@ const _fetch = (options: RequestWithOptions) => {
...
         // which can cause issues with proxies/WAFs that require it.
         const sendBeaconBody = body instanceof Blob ? body : new Blob([body], { type: contentType })
-        navigator!.sendBeacon!(url, sendBeaconBody)
+        if (navigator!.sendBeacon!(url, sendBeaconBody)) {
+            return
+        }
+
+        // rejected: over the page's shared ~64KiB in-flight keepalive quota
+        // (https://fetch.spec.whatwg.org/#http-network-or-cache-fetch) — halve so what fits still delivers
+        if (isArray(options.data) && options.data.length > 1 && (estimatedSize ?? 0) > BEACON_SPLIT_FLOOR_BYTES) {
```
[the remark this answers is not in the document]
turnipdabeets: Correct observation — a single-object payload can't be split at this layer: `_sendBeacon` has no knowledge of the object's internal structure, so halving is only possible for array batches. Non-array payloads get the same last-resort treatment as a single un-splittable event: warn + best-effort non-keepalive fetch. That's the designed floor, and it's strictly better than the previous behavior for the same payload (silently dropped with the boolean return ignored). If a specific producer (logs/metrics) ever hits the quota with a single object in practice, the right fix is structure-aware chunking in that module before it hands off to the transport — happy to file that as a follow-up if we see it.

packages/browser/src/request.ts
```diff
@@ -389,21 +392,38 @@ const _fetch = (options: RequestWithOptions) => {
...
+            _sendBeacon({ ...options, data: options.data.slice(0, mid) })
+            _sendBeacon({ ...options, data: options.data.slice(mid) })
+            return
+        }
+
+        logger.warn(`Beacon of ~${estimatedSize ?? 0} bytes was rejected by the browser, falling back to fetch`)
+        _fetch({ ...options, _keepaliveDisabled: true })
     } catch {
         // send beacon is a best-effort, fire-and-forget mechanism on page unload,
         // we don't want to throw errors here
```
ioannisj, line 429: should we warn/log here at least?

ioannisj (approved):

packages/browser/src/request.ts
turnipdabeets: Updated in [d14e919](https://github.com/PostHog/posthog-js/pull/4140/commits/d14e91975749dcf36d94b528b325920465751691)

PR #4143, opened 2026-07-14
feat(browser): stamp masked hostname as $snapshot_host on $snapshot events

robbie-c:
## Problem

I'm trying to beef up URL scrubbing (for the ML training data pipeline), and it'd be helpful if every $snapshot event contained the current domain so I could make sure that all references to it are scrubbed.

Claude code continues:

Mid-session `$snapshot` messages can contain zero URL-bearing events (rrweb Meta events and `$pageview`/`$url_changed` custom events only appear at navigation boundaries). Consumers that need to classify the recording's host per message — e.g. first-party vs external classification during scrubbing in the replay ingestion pipeline — have no signal on those messages and must fail closed, losing the distinction.

Capture-side `Origin`/`Referer` headers were considered and rejected: recordings can be batch-uploaded or relayed, and a wrong header-derived host would flip the fail-safe in the leak direction. A property stamped by the same runtime that produced the rrweb events stays consistent with the recording contents regardless of where the bytes are uploaded from.

## Changes

- `_flushBuffer` now stamps `$snapshot_host` on every flushed `$snapshot` message, computed once per flush and applied to every chunk `splitBuffer` produces, so each message carries its own host signal regardless of batch composition or ordering downstream.
- The value comes from a new `_currentMaskedHostname()` helper: `window.location.href` → `_maskReplayUrl` (the same masking pipeline applied to Meta hrefs and `$pageview`/`$url_changed` events, so the stamp cannot bypass a customer's URL masking config) → `new URL(masked).hostname`.
- If the URL is missing, masking removes it, or the masked result doesn't parse, the property is **omitted entirely** — consumers treat absence as "unclassifiable" and fail closed.
- Host only (no port, path, or query), minimizing added PII surface. Adds a few bytes per snapshot request.

The property is stamped in the recorder rather than generic property enrichment because `calculateEventProperties` deliberately early-returns for `$snapshot` to keep snapshot messages lean.

## Cross-repo contract

The ingestion consumer will read this property by the exact name `$snapshot_host` — the consumer-side change (folding it into the scrub context) is a separate task in the main repo.

## Testing

- Four new tests under `$snapshot_host stamping`: present on every flush, reflects the masked host when `maskCapturedNetworkRequestFn` rewrites the URL, absent when masking removes the URL, absent when the masked result doesn't parse.
- Updated 12 existing exact-match `$snapshot` property assertions to include the new property.
- All 656 replay unit tests pass; typecheck and lint clean.

---

### DIFF
--- a/.changeset/snapshot-host-property.md
+++ b/.changeset/snapshot-host-property.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': minor
+---
+
+Stamp the current hostname as `$snapshot_host` on every `$snapshot` event the session recorder sends. The value is derived from the page URL after it passes through the existing replay URL masking pipeline (`maskCapturedNetworkRequestFn` / deprecated `maskNetworkRequestFn`, hash stripping, personal-data query-param masking), so it cannot bypass a customer's masking config. When masking removes the URL or the masked result doesn't parse as a URL, the property is omitted entirely. This gives ingestion consumers a per-message host signal even for mid-session snapshot batches that contain no URL-bearing events.
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -1074,6 +1074,7 @@ describe('Lazy SessionRecording', () => {
                         $window_id: expect.any(String),
                         $lib: 'web',
                         $lib_version: '0.0.1',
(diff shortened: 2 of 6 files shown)


arnohillen (commented):
LGTM. Can you add this to changeset so the npm package is updated?

robbie-c:
@arnohillen done!

arnohillen (approved):

PR #4149, opened 2026-07-14
feat(dead-clicks): detect dead swipes as $dead_swipe events

pauldambra:
## Problem

Dead clicks autocapture catches clicks that go nowhere, but customers on touch-heavy web and mobile web apps also see users make swipe gestures that produce no result — a "failed navigation" (e.g. swipe-to-go-back or a carousel swipe that does nothing). There was no signal for those. This came up from a customer asking whether PostHog could detect false/dead swipes the way it detects dead clicks.

## Changes

Extends the existing dead clicks autocapture engine to also observe touch swipe gestures:

- A `touchstart`/`touchend` pair travelling beyond `swipe_threshold_px` becomes a swipe candidate and runs through the same dead-click heuristics (scroll / mutation / selection / visibility). If nothing changed, it is captured as a **`$dead_swipe`** event.
- Swipe events carry the same `$dead_swipe_*` timing properties as dead clicks, plus `$dead_swipe_direction` (`left`/`right`/`up`/`down`) and `$dead_swipe_distance_px`.
- New config on `capture_dead_clicks`: `capture_dead_swipes` (default `true`), `swipe_threshold_px` (default `30`), and `max_dead_swipes_per_page_load` (default `10`).
- Swipes over surfaces whose response the detector cannot observe (canvas/WebGL, video and other media/plugin elements under the finger, hit-tested via `elementsFromPoint`) are never captured, and `$dead_swipe` captures are capped per page load — so the feature can ship default-on without flooding events on pages it cannot judge.
- Swipes are only observed on the autocapture path, not the heatmaps path (which only tracks click points). Unlike clicks, swipes are not skipped on anchors, since a swipe isn't an anchor activation and a failed swipe-navigation is exactly the signal we want.

Tests were written first, modelled on the existing dead-clicks tests (`lazy-loaded-dead-swipes-autocapture.test.ts`), covering swipe detection, direction, thresholds, ignore rules, config gating, and capture output.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Built with the PostHog Slack app (Claude Code). The request was to add dead-swipe detection alongside dead clicks, primarily to catch failed navigations, writing tests first based on the existing dead-click tests.

Key decisions:
- Reused the existing `LazyLoadedDeadClicksAutocapture` engine rather than creating a parallel extension — swipes share the entire "did the screen change?" detection pipeline, so this avoids duplicating the controller, remote-config flag, and persistence wiring. A swipe is just a candidate with `type: 'swipe'`, and the event name/properties are derived from the candidate type.
- Kept swipe detection off the heatmaps path (gated on whether an external `__onCapture` was supplied) to avoid recording swipes as single-point heatmap "deadclick" entries and to avoid double-capture when both autocapture and heatmaps run.
- Widened `DeadClickCandidate.originalEvent` to `MouseEvent | TouchEvent`.

Verified: new + existing dead-click/heatmap/extension unit tests pass; lint clean on changed files. The remaining `@posthog/rrweb-*` typecheck errors are pre-existing and unrelated to this change.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0B7TE4R8GH/p1784045959537849?thread_ts=1784045959.537849&cid=C0B7TE4R8GH)*


pauldambra:
Adopted and completed a critical review pass.

Fixed a false-positive path where a successful swipe could scroll or mutate before `touchend`, before the candidate existed, and later be reported as `$dead_swipe`. Swipe candidates now use the gesture start timestamp and retain scroll activity observed during the gesture. Also fixed the touch test helper type and refreshed the config type snapshot.

Validation:
- dead-click/dead-swipe Jest suites: 81 passed, 1 skipped
- `@posthog/types` unit tests and snapshot: passed
- browser production TypeScript build: passed
- changed-file ESLint: passed

The browser test-only typecheck remains blocked locally by missing ambient `dompurify` and `dotenv` type definitions in this checkout; the production TypeScript build and Jest suites pass.

pauldambra:
Adopted for critical review. **One critical false-positive path found and fixed**, two smaller hardenings applied, and a few observations left as notes. Fixes are in 1b66ec0b (building on d7d8662b, which landed while I was reviewing and addressed part of the same issue).

## Critical: successful scroll gestures captured as `$dead_swipe` (fixed)

A swipe candidate is created at `touchend`, but the effect of the most common swipe — a drag that scrolls the page — happens **while the finger is down**, before the candidate exists. `_onScroll` only marks candidates already in `_clicks`, so a slow drag-scroll with no momentum tail (finger stops, then lifts) left `scrollDelayMs` unset, nothing else changed, and ~2.75s later the absolute timeout captured a `$dead_swipe` for a perfectly successful scroll. Same story for finger-following carousels that finish mutating at lift. On scrollable touch pages this would have been a flood of junk events for the exact customers the feature targets.

d7d8662b addressed this by stamping candidates with the gesture *start* time and carrying a during-gesture `scrollDelayMs`. That fixes quick drags, but makes the thresholds measure *gesture duration + response latency*, leaving two residual paths: a hold-then-drag whose first scroll lands ≥100ms after touchstart still captures as dead via `scroll_timeout`, and a JS scroll response shortly after lift does too (its delay is measured from touchstart, not the lift).

1b66ec0b keeps the in-flight `scrollDelayMs` tracking but changes the semantics: **a gesture that already scrolled, mutated the DOM, or grew a selection while the finger was down never becomes a candidate**, and surviving candidates keep the `touchend` timestamp so the post-gesture thresholds measure response-after-lift, exactly as they measure response-after-click. Regression tests cover scroll/mutation/selection during the gesture, plus a control asserting pre-gesture activity does *not* suppress.

## Smaller fixes

- `stop()` now clears the tracked touch origin — a gesture in flight across a remote-config stop/start cycle could pair a stale origin with a later touchend.
- The controller wrote `__onCapture` directly into the user's `capture_dead_clicks` config object (when it's an object). That object is shared between the autocapture and heatmaps controller instances, and `__onCapture`'s presence now also gates swipe observation — it's copied before writing now.

## Observations (not changed here)

- **Default-on is a volume/billing decision.** `capture_dead_swipes: true` means every project with dead clicks enabled (including via the remote-config flag) starts emitting a new event type on SDK upgrade. With the fix the events are meaningful, but idle thumb-flicks on non-scrollable layouts will still register — technically correct, potentially chatty. Worth a deliberate call on default-off vs. a server-side gate before release, and `$dead_swipe` needs product-side support (label/icon/docs) to be more than a raw event.
- **Pre-existing, now shared by swipes:** `visibilityChangedDelayMs` is `Math.abs(timestamp − lastVisibilityChange)` against *any* past visibility change, so once a tab has been re-foregrounded, `visibility_changed_timeout` is true for every candidate and capture happens at the first ~1s check instead of waiting the full 2.5s mutation window. A late-but-legitimate response (e.g. debounced navigation at 1.2s) can't save the candidate. Deserves its own fix since it changes dead-click behaviour too.
- Decisions that look right after scrutiny: single-finger gating with clearing on multi-touch/`touchcancel`, passive capture-phase listeners, not skipping anchors for swipes (defensible now that drag-scrolls over links are suppressed), keeping swipes off the heatmaps path with a belt-and-braces guard in `_onDeadClick`, and reusing the lazy-loaded engine rather than a parallel extension.

Verified: the four dead-clicks/heatmaps unit suites pass (150 tests), lint clean, types config snapshot unchanged, and a full `WRITE_MANGLED_PROPERTIES=1` build produces no mangle-cache drift (no new class fields were needed).

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -45,6 +60,8 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
         selection_change_threshold_ms: 100,
         mutation_threshold_ms: 2500,
         capture_clicks_with_modifier_keys: false,
+        capture_dead_swipes: true,
```
pauldambra, line 71: > [!NOTE]

**[convergent: paul + security-audit/business-logic-dos]** 🟡 MEDIUM

the default-on thing is the one i keep coming back to. `capture_dead_swipes: true` means the moment this lands, every touch device on every customer with dead clicks enabled starts emitting a brand new event type from a brand new heuristic — and swipe detection is notoriously noisy.

dead _clicks_ has a server-side kill switch (`DEAD_CLICKS_ENABLED_SERVER_SIDE`) so we can turn it off without a redeploy. swipes have nothing of their own. it's the knight capital thing (https://specbranch.com/posts/knight-capital/) — if this is 10x noisier than we expect, what's our lever?

the security pass independently traced the worst case: the "no observable change" heuristic is structurally blind to canvas/WebGL apps (maps, games — repaints aren't DOM mutations), video surfaces (UA shadow-root controls aren't observable), and boundary overscroll flicks (no scroll event when scrollTop can't change). ordinary panning on such pages emits a sustained stream of `$dead_swipe` events with full element payloads — billed events and mobile bandwidth — shipping silently on SDK upgrade to every project whose remote config already enables dead clicks. the 1s same-node dedup barely caps it since pans cross distinct child nodes.

not blocking, but i'd feel better if either (a) this defaulted to `false` for a release or two while we watch volume, or (b) it rode its own remote-config gate so we can pull it centrally — plus maybe a per-page-load cap on swipe captures. is there a reason it needs to be on by default on day one?

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -174,36 +205,114 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
...
+            swipeDirection: swipeDirection(dx, dy),
+            swipeDistancePx: Math.round(Math.sqrt(distanceSq)),
         })
+        if (!isNull(swipe) && !this._ignore(swipe)) {
+            this._clicks.push(swipe)
+        }
+        this._scheduleCheck()
+    }
+
+    private _ignore(candidate: DeadClickCandidate | null): boolean {
```
pauldambra, line 312: > [!NOTE]

**[xp]** 🟡 MEDIUM

`_ignore` grew an `isSwipe` flag that branches the policy in two places (modifier keys, and the anchor/`shouldSkipDeadClick` check). This is the deduplication-with-a-flag smell: click-ignore and swipe-ignore are genuinely different policies (swipes have no modifiers; a swipe on an anchor is the signal you *want*), sharing only the toolbar / already-seen / `shouldCaptureDeadClick` checks. Consider a small `_ignoreCommon(candidate)` called by separate `_ignoreClick` / `_ignoreSwipe` so each policy is expressed honestly rather than one method reading its own `type` to decide which rules apply. Not certain — it's close to the line — but right now you must hold both gesture types in your head to understand either.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -174,36 +205,114 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
...
+            // a short movement is a tap or a jitter, not a swipe
+            return
+        }
+
+        // a gesture that already did something while the finger was down — dragging scrolled
+        // the page, a finger-following carousel mutated the DOM, a text selection grew — is not
+        // dead, so it never becomes a candidate. surviving candidates keep the touchend
+        // timestamp: the thresholds in _checkClicks measure how quickly the page responds
+        // after the finger lifts, just as they measure the response to a click
+        const gestureCausedActivity = [this._lastScroll, this._lastMutation, this._lastSelectionChanged].some(
```
pauldambra, line 269: > [!NOTE]

**[paul]** 🟢 LOW

the docs say four guards, the code checks three. `gestureCausedActivity` looks at scroll/mutation/selection but not `_lastVisibilityChange`, yet the `capture_dead_swipes` docstring says a dead swipe has "no scroll, mutation, selection _or_ visibility change".

i went and looked — `_onVisibilityChange` only records the tab becoming _visible_ (coming back), not leaving — so i don't think there's a real correctness hole (and `_checkClicks` still applies the visibility check post-gesture). but the mismatch made me stop and check, and the next person will too. could we either add `_lastVisibilityChange` to the array for symmetry, or tweak the docstring so they match? not blocking, just readability.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -31,10 +44,12 @@ function checkTimeout(value: number | undefined, thresholdMs: number) {
 class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocaptureInterface {
     private _mutationObserver: MutationObserver | undefined
     private _lastMutation: number | undefined
+    private _lastScroll: number | undefined
     private _lastSelectionChanged: number | undefined
     private _lastVisibilityChange: number | undefined
     private _clicks: DeadClickCandidate[] = []
```
pauldambra, line 55: > [!NOTE]

**[xp]** 🟢 LOW

The `_clicks` / `_checkClicks` / `_checkClickTimer` / `_scheduleCheck` vocabulary now holds and processes swipes too, so `this._clicks.push(swipe)` reads oddly (Rule 2 — express intent). I would NOT rename the world here: `DeadClickCandidate` is a shared exported type and the churn would swamp the diff. But worth making the debt visible; if there's one easy win, renaming the array to `_candidates` repays itself fastest. Your call — flagging so it's conscious rather than drift.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -268,33 +377,31 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
...
 
-        if (this._clicks.length && isUndefined(this._checkClickTimer)) {
-            this._checkClickTimer = assignableWindow.setTimeout(() => {
-                this._checkClicks()
-            }, 1000)
-        }
+        this._scheduleCheck()
     }
 
     private _captureDeadClick(click: DeadClickCandidate, properties: Properties) {
```
pauldambra, line 451: > [!NOTE]

**[xp]** 🟢 LOW

`_captureDeadClick` now captures swipes and reaches for swipe-only fields (`$dead_swipe_direction`, `$dead_swipe_distance_px`) — same naming drift as the `_clicks` note. Emitting them always-but-undefined-for-clicks is fine: it matches how the existing `_dead_click_*_delay_ms` fields already lean on JSON dropping `undefined`, and the comment says so. No behaviour change needed; just noting the method name no longer tells the whole truth.

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -45,6 +60,8 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
         selection_change_threshold_ms: 100,
         mutation_threshold_ms: 2500,
         capture_clicks_with_modifier_keys: false,
+        capture_dead_swipes: true,
+        swipe_threshold_px: 30,
```
pauldambra, line 72: > [!NOTE]

**[paul]** ⚪ NIT

silly question... 30px — where'd that come from? it's configurable and measurable which is the bit i care about (thank you 🙏), so this is genuinely just curiosity. a one-line "// ~2-3mm on a typical mobile, below this it's a tap not a swipe" or wherever the number came from would stop future-me wondering. ship as you see fit.

packages/browser/src/__tests__/entrypoints/lazy-loaded-dead-swipes-autocapture.test.ts
```diff
@@ -0,0 +1,368 @@
...
+        it('swipe followed by a scroll after threshold, dead swipe', () => {
+            pushSwipeCandidate({ scrollDelayMs: 2501 })
+
+            lazyLoadedDeadClicksAutocapture['_checkClicks']()
+
+            expect(lazyLoadedDeadClicksAutocapture['_clicks']).toHaveLength(0)
+            expect(fakeInstance.capture).toHaveBeenCalledWith(
+                '$dead_swipe',
+                expect.objectContaining({
+                    $dead_swipe_absolute_delay_ms: -900,
```
pauldambra, line 388: > [!NOTE]

**[xp]** ⚪ NIT

`jest.setSystemTime(0)` against a candidate with `timestamp: 900` asserts `$dead_swipe_absolute_delay_ms: -900` — a negative delay production can't reach (time doesn't run backwards). It passes because the matcher just checks the number, but the fixture is slightly unreal. Set the candidate timestamp behind "now" (timestamp 0, then advance the clock) to read truer. Minor — the coverage itself is solid.

pauldambra (commented):
> [!NOTE]

pauldambra:
> [!NOTE]
>
> Multi-perspective review: paul-reviewer, xp-reviewer, security-audit (qa-team skipped — skill unavailable in this environment)

Well-engineered and well-tested; no correctness or security blockers found. The one substantive concern is rollout shape (default-on volume risk), flagged independently by two reviewers.

### Key findings

**🟡 MEDIUM**
- `capture_dead_swipes: true` default: a brand-new, inherently noisy touch heuristic ships enabled to every project with dead clicks on, with no swipe-specific kill switch — and the heuristic is structurally blind to canvas/WebGL/video surfaces and boundary overscroll, where ordinary panning emits a sustained `$dead_swipe` stream (billed events) on SDK upgrade. Consider default-off or a dedicated remote gate + a volume cap. *(convergent: paul + security-audit)*
- `_ignore` serves two gesture types via `isSwipe` branches — consider `_ignoreCommon` + per-type policies. *(xp)*

**🟢 LOW**
- Docstring promises a visibility-change guard the gesture-window check doesn't perform. *(paul)*
- `_clicks`/`_checkClicks` vocabulary now quietly carries swipes; `_candidates` would repay itself. *(xp)*
- `_captureDeadClick` name drift (captures swipes too). *(xp)*

**⚪ NIT** — unexplained 30px threshold; a test fixture asserting an impossible negative delay.

### Convergence

The default-on / volume-amplification concern was flagged independently by paul-reviewer (rollout-lever framing) and security-audit (traced billed-event amplification on canvas/video/overscroll surfaces) — highest-confidence finding of the review.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | skipped — skill unavailable |
| 👤 paul | approve with caveats: clean refactor, great parameterised tests; wants a rollout lever for the new default-on event before it hits every customer |
| 📐 xp | correctness sound, comments explain the why; main reservation is `_ignore`'s flag-parameter dedup and the `_clicks` naming debt |
| 🛡 security-audit | privacy stack fully reused (masking, ph-no-capture, sensitive-input exclusion); no XSS/leak/jank vectors; one Medium volume-amplification finding on the default-on path |

---

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — recording the author's decision, not written by a human

Author decision: `capture_dead_swipes` stays **default on** — this comes from a customer request, the expected volume is low, and the events are useful.

Both structural blind spots this thread traced are now guarded in c648a066437372d6993173980787c178ab04cee3:
- swipes over unobservable surfaces (canvas/WebGL, video/audio/embed/object) never become candidates — hit-tested via `elementsFromPoint` against the stack under the finger, so overlay-div canvas apps are caught; gated behind a cached document-level existence check so canvas-free pages never pay the hit-test
- `$dead_swipe` captures are capped per page load (`max_dead_swipes_per_page_load`, default 10, configurable), bounding any surface class we have not enumerated

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
pauldambra: > [!NOTE]
> 🤖 Automated comment by **PostHog Code** — recording a change made in response, not written by a human

Done in a3348c138d6359a537bdebf6ab3c6af8128b489d. `_ignore` no longer carries scattered `isSwipe` conditionals — it now null-guards then dispatches:

- `_ignoreCommon` — gates that apply to every candidate (toolbar, per-node 1s dedup, html/non-element, `css_selector_ignorelist`)
- `_ignoreClick` — click-only gates layered on top (modifier keys, and the anchor skip; a swipe that fails to navigate is the signal we want, so anchors are skipped for clicks only)

Swipes run the shared gates only; clicks run the click policy. All checks are pure predicates so the regrouping is behaviour-preserving — the full dead-click and dead-swipe suites are unchanged and green (157 passing), lint/typecheck clean, and the mangle-cache build adds only the two new method names.

packages/types/src/posthog-config.ts
```diff
@@ -337,6 +344,43 @@ export type DeadClicksAutoCaptureConfig = {
...
+     * Dead swipes are captured as `$dead_swipe` events. This only applies to the dead-click
+     * autocapture path, not the heatmaps path.
+     *
+     * Swipes over surfaces whose response cannot be observed — canvas, video and other
+     * media/plugin elements under the finger — are never captured, and capture is limited
+     * per page load (see `max_dead_swipes_per_page_load`).
+     *
+     * @default true
+     */
+    capture_dead_swipes?: boolean
```
dustinbyrne, line 364: in my testing this works pretty well

is it worth rolling this out at a smaller scale first _then_ switching the default to true? i think it's likely fine either way.

dustinbyrne (approved):
looks good!

PR #4155, opened 2026-07-15
fix(node): allow non-string person and group properties in flag options

### LINKED ISSUE #2133: Incorrect types on `personProperties` for `getAllFlags`
## Bug description

`personProperties` is typed as `Record<string, string>` whereas it should be `Record<string, string | number>`

## How to reproduce

Try to call `getAllFlags` with a numerical `personProperty` and you will get a type error

image

## Related sub-libraries

- [ ] All of them
- [ ] posthog-web
- [x] posthog-node
- [ ] posthog-react-native

## Additional context

`posthog-node` version `3.6.3`

ATKasem:
## Problem

Closes #2133.

`personProperties` is typed as `Record<string, string>`, so passing a numeric person property to `getFeatureFlag` / `getAllFlags` / `evaluateFlags` / `isFeatureEnabled` fails to type-check:

```ts
await posthog.getAllFlags('user_123', {
  personProperties: { age: 30 },
  //                  ~~~ Type 'number' is not assignable to type 'string'.
})
```

Local evaluation has always handled numbers — only the public types disagree:

- `matchProperty` accepts `propertyValues: Record<string, any>`, and the `gt`/`gte`/`lt`/`lte` branch explicitly checks `typeof overrideValue === 'number'` before comparing numerically.
- `personPropertiesForLocalEvaluation` already takes and returns `Record<string, any>`.
- `SendFeatureFlagsOptions.personProperties` is already `Record<string, any>`, so the same value is accepted on the capture path and rejected on the flag path.

The narrow type also shows up as friction in the existing suite, which casts around it:

```ts
} as unknown as Record<string, string>,
```

This matches what was agreed on the issue: types should be relaxed to accept numbers, and doing so isn't breaking since `Record<string, any>` still accepts everything `Record<string, string>` did.

## Changes

Widened `personProperties` to `Record<string, any>` and `groupProperties` to `Record<string, Record<string, any>>` on the flag evaluation options — `BaseFlagEvaluationOptions` (which `FlagEvaluationOptions` and `AllFlagsOptions` extend), the `IPostHog` declarations for `isFeatureEnabled` / `getFeatureFlag`, and their `client.ts` counterparts. `addLocalPersonAndGroupProperties` follows, since it now receives them.

Left alone deliberately:

- `groups?: Record<string, string>` — group keys, not properties.
- `extractPropertiesFromEvent` — it `String()`-coerces every value, so `Record<string, string>` is accurate there.

Types only; no runtime change.

This also removes a now-unnecessary `as unknown as Record<string, string>` cast from the existing "undefined property values" test — the narrow type was the only reason it was there.

### Why `Record<string, any>` and not a narrower union

`Record<string, string | number | boolean>` looks tighter, but it would be wrong here: `null`/`undefined` are legitimate person property values. `matchProperty` has an explicit null guard, and `NULL_VALUES_ALLOWED_OPERATORS` lets `is_not` / `is_set` match on them. A `string | number | boolean` union would reject those, and the existing test above would still need its cast to compile.

`any` is also the established shape for property bags here: `SendFeatureFlagsOptions` and `personPropertiesForLocalEvaluation` already use it, `matchProperty` takes `Record<string, any>`, and `@posthog/types` defines `Property = any` / `Properties = Record<string, Property>` (which the browser SDK re-exports). Using `Properties` by name would be equivalent — it resolves to exactly `Record<string, any>` — but `@posthog/types` isn't currently reachable from `posthog-node` (it comes in via `@posthog/core`, which doesn't re-export it), and adding that re-export felt like more blast radius than this fix warrants. Happy to do it if you'd rather have the named type.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Test strategy

Added `evaluates numeric person properties with comparison operators` to `feature-flags.spec.ts`, alongside the existing person-property tests. It defines a flag with a `gt: 21` condition on a person property and evaluates it locally with `personProperties: { age: 30 }` (expects `true`) and `{ age: 18 }` (expects `false`) — a positive and a negative case, so the change can't pass by making everything match.

Because jest strips types via babel, that test passes at runtime with or without the fix — the type error is what it's actually guarding. So I verified it against `tsc` in both directions:

- Reverting only `types.ts` + `client.ts` and keeping the test: `tsc --noEmit` fails with `error TS2322: Type 'number' is not assignable to type 'string'` on both new call sites.
- With the fix applied: no errors in `src/`.

Also ran the full `posthog-node` suite (794 passing). Two failures in `exception-autocapture.spec.ts` reproduce identically on a clean checkout of `main`, so they're unrelated to this change. `pnpm lint` is clean, and `packages/node/references/posthog-node-references-latest.json` is regenerated via `pnpm generate-references` — the only delta is the four widened type signatures.

Edge cases considered: `undefined` values still take the existing `null`-guard path in `matchProperty`; string-numeric values like `"10"` still compare numerically via the existing `parseFloat` fallback; `groups` is untouched.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — I'm the DRI. (The template asks for the driver as assignee; I can't set assignees as an external contributor, so noting it here instead — this PR isn't autonomous despite being unassigned.)

I used Claude Code (Opus 4.8) to investigate the issue and draft the change; I reviewed the diff and the reasoning before opening this.

Notes that may help review:

- The fix direction came from reading the runtime rather than the issue title — `matchProperty` already being `Record<string, any>` is what makes this a types-only change, and it's why I widened `groupProperties` alongside `personProperties` rather than only the field named in the issue.
- I first reached for a narrower `string | number | boolean` union before noticing the null guard in `matchProperty` ruled it out — the reasoning for `any` in the section above is the result of checking that, not a preference for looser types.
- `extractPropertiesFromEvent` looked like it needed the same treatment, but it string-coerces its output, so widening it would have made the type less accurate. Left it.

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -41,6 +41,9 @@ export type { CaptureLogOptions, LogAttributeValue, LogAttributes, LogSeverityLe
 // Re-export the shared error tracking rate-limiter config type so SDKs built on core
 // (e.g. posthog-node) don't have to depend on `@posthog/types` directly.
 export type { ExceptionRateLimiterConfig } from '@posthog/types'
+// Re-export the shared `Properties` type for the same reason, so SDKs can type
+// person/group property bags consistently without importing `@posthog/types`.
+export type { Property, Properties } from '@posthog/types'
 export {
   PostHogMetrics,
   buildOtlpMetricsPayload,
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -15,6 +15,7 @@ import {
   PostHogFlagsResponse,
   PostHogMetrics,
   PostHogPersistedProperty,
+  Properties,
   resolveMetricsConfig,
   RetriableOptions,
   safeSetTimeout,
@@ -1076,8 +1077,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     distinctId: string,
     options: {
       groups?: Record<string, string>
-      personProperties?: Record<string, any>
-      groupProperties?: Record<string, Record<string, any>>
+      personProperties?: Properties
+      groupProperties?: Record<string, Properties>
       onlyEvaluateLocally?: boolean
       sendFeatureFlagEvents?: boolean
       disableGeoip?: boolean
@@ -1343,8 +1344,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     distinctId: string,
     options?: {
       groups?: Record<string, string>
-      personProperties?: Record<string, any>
-      groupProperties?: Record<string, Record<string, any>>
+      personProperties?: Properties
+      groupProperties?: Record<string, Properties>
       onlyEvaluateLocally?: boolean
       sendFeatureFlagEvents?: boolean
       disableGeoip?: boolean
@@ -1414,8 +1415,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     matchValue?: FeatureFlagValue,
     options?: {
       groups?: Record<string, string>
-      personProperties?: Record<string, any>
-      groupProperties?: Record<string, Record<string, any>>
+      personProperties?: Properties
+      groupProperties?: Record<string, Properties>
       onlyEvaluateLocally?: boolean
       /** @deprecated THIS OPTION HAS NO EFFECT, kept here for backwards compatibility reasons. */
       sendFeatureFlagEvents?: boolean
@@ -1601,8 +1602,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     distinctId: string,
     options?: {
       groups?: Record<string, string>
-      personProperties?: Record<string, any>
-      groupProperties?: Record<string, Record<string, any>>
+      personProperties?: Properties
+      groupProperties?: Record<string, Properties>
       onlyEvaluateLocally?: boolean
       sendFeatureFlagEvents?: boolean
       disableGeoip?: boolean
@@ -2620,12 +2621,12 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
   private addLocalPersonAndGroupProperties(
     distinctId: string,
     groups?: Record<string, string>,
-    personProperties?: Record<string, any>,
-    groupProperties?: Record<string, Record<string, any>>
-  ): { allPersonProperties: Record<string, any>; allGroupProperties: Record<string, Record<string, any>> } {
+    personProperties?: Properties,
+    groupProperties?: Record<string, Properties>
+  ): { allPersonProperties: Properties; allGroupProperties: Record<string, Properties> } {
     const allPersonProperties = { ...(personProperties || {}) }
 
-    const allGroupProperties: Record<string, Record<string, any>> = {}
+    const allGroupProperties: Record<string, Properties> = {}
     if (groups) {
       for (const groupName of Object.keys(groups)) {
         allGroupProperties[groupName] = {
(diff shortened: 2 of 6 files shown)


packages/node/src/client.ts
```diff
@@ -1363,8 +1363,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
     matchValue?: FeatureFlagValue,
     options?: {
       groups?: Record<string, string>
-      personProperties?: Record<string, string>
-      groupProperties?: Record<string, Record<string, string>>
+      personProperties?: Record<string, any>
+      groupProperties?: Record<string, Record<string, any>>
```
dustinbyrne: Hey @ATKasem, thanks for putting this PR together! Good catch on this.

Can we swap out these types for the custom [`Properties` type definition](https://github.com/PostHog/posthog-js/blob/6326610e14c867c0db529aa1d919488e27ed2eef/packages/types/src/common.ts#L6)? E.g.:
```
personProperties?: Properties
groupProperties?: Record<string, Properties>
```

This matches other method signatures like `setPersonPropertiesForFlags` / `setGroupPropertiesForFlags` should we want to narrow that `any` type at some point.

ATKasem: Done — swapped to the shared `Properties` type: `personProperties?: Properties` and `groupProperties?: Record<string, Properties>`, matching `setPersonPropertiesForFlags`/`setGroupPropertiesForFlags`. Since posthog-node only depends on `@posthog/core`, I re-exported `Property`/`Properties` from core (next to the existing `ExceptionRateLimiterConfig` re-export) so we don't have to pull in `@posthog/types` directly. Also applied it to the internal local-eval helpers for consistency, regenerated the API references, and rebased onto main to clear the conflict.

packages/core/src/index.ts
```diff
@@ -40,6 +40,9 @@ export type { CaptureLogOptions, LogAttributeValue, LogAttributes, LogSeverityLe
 // Re-export the shared error tracking rate-limiter config type so SDKs built on core
 // (e.g. posthog-node) don't have to depend on `@posthog/types` directly.
 export type { ExceptionRateLimiterConfig } from '@posthog/types'
+// Re-export the shared `Properties` type for the same reason, so SDKs can type
+// person/group property bags consistently without importing `@posthog/types`.
+export type { Property, Properties } from '@posthog/types'
```
dustinbyrne, line 46: Can you add another changeset for this update to `@posthog/core`? I think this looks good to go once this is in

dustinbyrne (commented):
Linting has also failed https://github.com/PostHog/posthog-js/actions/runs/29608595187/job/88723355365?pr=4155#step:4:83

dustinbyrne:
Hey @ATKasem, just following up here. I can help get this fixed up if you'd prefer. Your commits will need to be signed in order to merge.

ATKasem:
Thanks @dustinbyrne — I've got it! Sorry for the delay. Pushing up the lint fix and re-signing the commits now.

ATKasem:
@dustinbyrne all set on my end — commits are signed and showing as Verified, and CI is green.

Two things were red:
- The `prettier/prettier` error in `packages/node` (the `personPropertiesForLocalEvaluation` signature fit on one line under the 120 print width).
- `Check changeset hygiene` — that one was flagging a real gap: `packages/core/src/index.ts` re-exports `Property`/`Properties` from `@posthog/types`, so `@posthog/core` needed declaring in the changeset. Added it as a patch alongside `posthog-node`. (The job was also erroring on the sticky-comment API call, which fork PRs can't make — declaring the package empties the report body so it no longer tries.)

Ready for review whenever you have a moment.

marandaneto (approved):

PR #4159, opened 2026-07-15
feat(flags): add $feature_flag_has_experiment to $feature_flag_called events

haacked:
## Problem

Part of a cross-SDK effort to shrink `$feature_flag_called` events. This first phase adds a `$feature_flag_has_experiment` boolean property to every `$feature_flag_called` event, reflecting the server's `has_experiment` signal: `metadata.has_experiment` in the `/flags?v=2` response and `has_experiment` on `/api/feature_flag/local_evaluation` flag definitions. When the server does not report the field (older deployments, bootstrapped flags), the property is omitted, so it is tri-state: `true`, `false`, or absent (unknown).

This lets ingestion distinguish experiment-linked flag events (which need the full property set for exposure analysis) from the rest, and lets us measure the split before a later phase strips the expensive properties (`$feature/<key>`, `$feature_flag_payload`, per-event system metadata) from non-experiment flag events. This PR intentionally minimizes nothing: no properties are removed, no config options are added, and dedupe behavior is unchanged.

## Changes

- `@posthog/core`: `has_experiment?: boolean` on `FeatureFlagMetadata`; `_getFeatureFlagResult` emits `$feature_flag_has_experiment` from `metadata.has_experiment ?? false`. This covers posthog-react-native and other core-based SDKs.
- `posthog-js` (browser): field added to `FeatureFlagMetadata` in `@posthog/types`; property added to the capture in `posthog-featureflags.ts` from the persisted v2 flag details.
- `posthog-node`: all three send paths covered: single-flag local evaluation (from the `/local_evaluation` definition), single-flag remote fallback (from `flagDetail.metadata`), and the `evaluateFlags()` snapshot path (`hasExperiment` threaded through `EvaluatedFlagRecord` for local and remote records).
- Bootstrapped/injected flags carry no server metadata, so they report `false` by design.
- Regenerated API reference files; changeset bumps `@posthog/core`, `@posthog/types`, `posthog-js`, `posthog-node`, `posthog-react-native` (minor).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native

### DIFF
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -1040,6 +1040,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         ...maybeAdd('$feature_flag_request_id', details?.requestId),
         ...maybeAdd('$feature_flag_evaluated_at', details?.evaluatedAt),
         ...maybeAdd('$feature_flag_error', featureFlagError),
+        $feature_flag_has_experiment: featureFlag?.metadata?.has_experiment ?? false,
       }
 
       this.capture('$feature_flag_called', properties)
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -1040,7 +1040,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         ...maybeAdd('$feature_flag_request_id', details?.requestId),
         ...maybeAdd('$feature_flag_evaluated_at', details?.evaluatedAt),
         ...maybeAdd('$feature_flag_error', featureFlagError),
-        $feature_flag_has_experiment: featureFlag?.metadata?.has_experiment ?? false,
+        ...maybeAdd('$feature_flag_has_experiment', featureFlag?.metadata?.has_experiment),
       }
 
       this.capture('$feature_flag_called', properties)
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -587,7 +587,7 @@ export type FeatureFlagMetadata = {
   description: string | undefined
   // Payloads in the response are always JSON encoded as a string
   payload: string | undefined
-  // Whether the flag is linked to an experiment. Absent on older servers.
+  /** Whether the flag is linked to an experiment. Absent when the server does not report it. */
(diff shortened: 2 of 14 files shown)


packages/core/src/posthog-core.ts
```diff
@@ -1040,6 +1040,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         ...maybeAdd('$feature_flag_request_id', details?.requestId),
         ...maybeAdd('$feature_flag_evaluated_at', details?.evaluatedAt),
         ...maybeAdd('$feature_flag_error', featureFlagError),
+        $feature_flag_has_experiment: featureFlag?.metadata?.has_experiment ?? false,
```
turnipdabeets: [question] Defaulting to `false` conflates "server didn't report `has_experiment`" (older servers, bootstrapped flags) with "flag has no experiment" — if a later phase strips properties from `false` events, experiment flags on older servers would lose their exposure data. Should we omit the property via `maybeAdd` (like `$feature_flag_request_id` above) when metadata is absent, or is always-`false` the settled cross-SDK contract?

packages/core/src/types.ts
```diff
@@ -587,6 +587,8 @@ export type FeatureFlagMetadata = {
   description: string | undefined
   // Payloads in the response are always JSON encoded as a string
   payload: string | undefined
+  // Whether the flag is linked to an experiment. Absent on older servers.
+  has_experiment?: boolean
```
turnipdabeets, line 591: [nit] Could we make this a `/** … */` doc comment so it shows in editor tooltips on the new public field? Same for the twins in `packages/types/src/feature-flags.ts` and `packages/node/src/types.ts`.

packages/core/src/posthog-core.ts
haacked: Ah yeah, we should omit if the field isn't in the response. I'll make that change to all my SDK PRs.

packages/core/src/types.ts
haacked: Done!

turnipdabeets (approved):

PR #4166, opened 2026-07-16
fix(docs): exclude internal members and tag deprecated ones in SDK references

turnipdabeets:
## Problem

Follow-up to #4165 (and PostHog/posthog.com#18577): the generated references published every member of object-shaped types as first-class options — `__`-prefixed internals and `@deprecated` members included, with no signal distinguishing them. A reader could pick `enable_heatmaps` with no hint that `capture_heatmaps` replaces it, or build on a `__preview_*` flag that may vanish in a patch release. The `@deprecated` migration text in the source JSDoc was dropped entirely by the pipeline.

## Changes

Two commits, second refines the first:

1. `60edb6ba5` excluded both categories.
2. `1ce63f145` (final behavior) keeps deprecated members but **tags** them, which preserves discoverability for people maintaining older integrations:

- **`__`/`_`-prefixed members are excluded** from both pipeline paths (checker-resolved aliases and interfaces). Internal surface shouldn't be documented at all.
- **`@deprecated` members stay, annotated**: they gain `releaseTag: "deprecated"` (the same field methods already carry) and their `@deprecated` tag text is appended to the description, e.g.:

  ```json
  {
    "name": "enable_heatmaps",
    "type": "boolean",
    "description": "Deprecated: Use `capture_heatmaps` instead.",
    "releaseTag": "deprecated"
  }
  ```

  The deprecation is visible as plain text today with no docs-site changes; `releaseTag` gives posthog.com a hook for strikethrough/badge rendering later.

- Regenerated references: browser `PostHogConfig` publishes 120 options (102 current + 18 tagged deprecated; 12 underscore internals removed). Tagged counts: 27 browser / 6 node / 5 react-native. Audit vs the pre-filter output: zero non-underscore members removed, schema change is the single additive `releaseTag` key.
- Every underscore/deprecated classification was verified against source JSDoc — including subtleties like `SurveyConfig.autoSubmit*` (deprecated only in the browser-local declaration, not the `@posthog/types` one of the same name).
- Tests (32 total) cover: tagging through both paths, summary + tag text combining, bare `@deprecated` with no message, deprecated enum members, underscore exclusion, and that `releaseTag` serializes only on deprecated members.

### Libraries affected

None — docs-generation tooling and regenerated reference JSONs only; no published package code changes, no version bumps.

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

No changeset: docs tooling and generated files only.

## 🤖 Agent context

Built with Claude Code as a follow-up to #4165, prompted by the public/internal/deprecated split in PostHog/posthog.com#18577. We first implemented exclusion for both categories, then prototyped tagging locally and compared real outputs; tagging won because the migration text (previously dropped) becomes visible with zero docs-site changes while keeping deprecated options discoverable. Every filtered/tagged member was individually verified against source JSDoc before committing.

### DIFF
--- a/packages/browser/references/posthog-js-references-latest.json
+++ b/packages/browser/references/posthog-js-references-latest.json
@@ -2372,6 +2372,12 @@
           "name": "selector"
         },
         {
+          "description": "Deprecated: Only `selector` should be used now.",
+          "type": "string",
+          "name": "tag_name",
+          "releaseTag": "deprecated"
+        },
+        {
           "description": "StringMatching.Exact",
           "type": "ActionStepStringMatching | null",
           "name": "text_matching"
@@ -2984,6 +2990,12 @@
           "name": "persistence_name"
         },
         {
+          "description": "Deprecated: - Use 'persistence_name' instead",
+          "type": "string",
+          "name": "cookie_name",
+          "releaseTag": "deprecated"
+        },
+        {
           "description": "List of custom property names that should be stored in cookies (in addition to the default ones)\nwhen using 'localStorage+cookie' persistence mode. This allows these properties to be shared\nacross subdomains when cross_subdomain_cookie is enabled.",
           "type": "readonly string[]",
           "name": "cookie_persisted_properties"
@@ -3004,6 +3016,12 @@
           "name": "save_campaign_params"
         },
         {
+          "description": "Deprecated: - Use `save_campaign_params` instead",
+          "type": "boolean",
+          "name": "store_google",
+          "releaseTag": "deprecated"
+        },
+        {
           "description": "Used to extend the list of campaign parameters that are saved by default.",
           "type": "string[]",
           "name": "custom_campaign_params"
@@ -3017,6 +3035,12 @@
           "description": "Determines whether PostHog should be in debug mode.\nYou can enable this to get more detailed logging.\n\nYou can also enable this on your website by appending `?__posthog_debug=true` at the end of your URL\nYou can also call `posthog.debug()` in your code to enable debug mode",
           "type": "boolean",
           "name": "debug"
+        },
+        {
+          "description": "Deprecated: Use `debug` instead",
+          "type": "boolean",
+          "name": "verbose",
+          "releaseTag": "deprecated"
         },
         {
           "description": "Determines whether PostHog should capture pageview events automatically.\nCan be:\n- `true`: Capture regular pageviews (default)\n- `false`: Don't capture any pageviews\n- `'history_change'`: Capture pageviews on the initial page load and on history API changes (pushState, replaceState, popstate)",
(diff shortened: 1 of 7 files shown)


scripts/docs/type-resolver.js
```diff
@@ -99,6 +99,17 @@ function classifyAlias(checker, declaration) {
...
 }
 
+// Underscore-prefixed and @deprecated members are internal surface and stay out of
+// the published reference; methods mirror the interface rendering, which lists only
+// property signatures
+function isDocumentedProperty(checker, prop) {
+    return (
+        !(prop.flags & ts.SymbolFlags.Method) &&
+        !prop.getName().startsWith('_') &&
+        !prop.getJsDocTags(checker).some((tag) => tag.name === 'deprecated')
```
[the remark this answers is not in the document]
turnipdabeets: This doesn't reproduce in the PR's final state — it looks reviewed against the intermediate commit (60edb6ba5, exclusion), which 1ce63f145 superseded with tagging. Two checks against the head:

- `PostHogAutocaptureElement` (react-native reference) is byte-identical to main: `tag_name` present with no `releaseTag`, so no tags leaked across same-named symbols in other types.
- A scripted diff vs main across all three reference JSONs shows zero non-underscore property removals. Deprecated members (e.g. `ActionStepType.tag_name`, which genuinely carries `@deprecated Only selector should be used now`) are kept and tagged, not dropped.

On the mechanism: `getJsDocTags` merges tags across a *symbol's own* declarations, but each type's members are distinct symbols — `tag_name` on `PostHogAutocaptureElement` and on `ActionStepType` don't share one. If a merged-declaration case ever appears, the serialization test (`releaseTag` only on deprecated members) and the audit would surface it.

packages/browser/references/posthog-js-references-latest.json
```diff
@@ -3220,8 +3204,10 @@
           "name": "opt_out_useragent_filter"
         },
         {
+          "description": "Deprecated: Use `consent_persistence_name` instead. This will be removed in a future major version. *",
```
ioannisj, line 3207: That `*` at the end doesn't look right?

packages/browser/references/posthog-js-references-latest.json
```diff
@@ -3415,12 +3400,10 @@
           "name": "get_current_url"
         },
         {
+          "description": "Deprecated: - use `before_send` instead",
           "type": "(properties: Properties, event_name: string) => Properties",
-          "name": "sanitize_properties"
-        },
-        {
-          "type": "(eventName: string, eventData: CaptureResult) => void",
-          "name": "_onCapture"
```
ioannisj, line 3423: `_onCapture` is excluded cause it's prefixed, but it's also deprecated. Maybe deprecated fields should always win?

https://github.com/PostHog/posthog-js/blob/fad6d9adae4163cd63859766916cdcbae629a110/packages/types/src/posthog-config.ts#L1747

ioannisj (approved):

PR #4172, opened 2026-07-16
feat(flags): minimize $feature_flag_called events for non-experiment flags

haacked:
## Problem

`$feature_flag_called` events carry the full enriched properties dict (super properties, context tags, custom event properties, system context) on every flag call, even for the vast majority of flags that aren't linked to an experiment. This PR minimizes those events for non-experiment flags when the server enables it, cutting event size and ingestion cost without losing anything needed for exposure analysis.

The event is trimmed to a strict allowlist iff the server-controlled gate is on (`minimalFlagCalledEvents` in the v2 `/flags` response, or `minimal_flag_called_events` in the local-evaluation payload) and the evaluated flag's `has_experiment` is exactly `false`. Any missing signal sends the full legacy shape unchanged. The reason for the server-gate is we can't just roll this out to everyone — it could break existing insights; we'll announce first and give people time to adjust. The main goal is to establish `$feature_flag_called` as a special system event.

Part of a cross-SDK rollout; reference implementation and fuller context: PostHog/posthog-python#748

## Changes

**`@posthog/core`** — the shared pieces:

- `MINIMAL_FLAG_CALLED_EVENT_PROPERTIES`: the canonical allowlist (flag identity, evaluation debug scalars, processing controls like `$groups`/`$process_person_profile`/`$geoip_disable`, debug location, and SDK identity incl. `$is_server`). It's the union across client and server SDKs; entries are inert where an SDK never sets them.
- `minimizeFlagCalledEventProperties()`: builds the minimal property set from the allowlist rather than deleting keys, so anything not explicitly allowlisted is structurally excluded.
- `PostHogCoreStateless.buildPayload` applies the filter for gated non-experiment `$feature_flag_called` events, behind a `protected isMinimalFlagCalledEventsEnabled()` hook (base: `false`, fail-safe).
- `PostHogCore` persists the gate alongside the cached flag details (`minimalFlagCalledEvents` on the stored flags format), overwriting it on every successful flags response so an absent field flips it off. Failed/quota-limited requests keep the stored gate, matching how they keep the cached flags.

**posthog-js (browser)** — the biggest payload win:

- The gate is parsed in `parseFlagsResponse` and persisted as a new `$minimal_flag_called_events` persistence key, registered with `exposure: 'hidden'` in the persistence key policy so it can never leak into event properties. Legacy v1 responses and locally injected flags always store `false`; bootstrap flags carry no gate, so they send full events until the first real `/flags` response.
- The filter runs in `capture()` after `calculateEventProperties` has merged persistence/context/super properties and after the last SDK-added property (the timestamp-override props), so the strip is structural. `before_send` runs after the filter and may deliberately re-add stripped properties. Transport-level keys the browser carries inside `properties` (`token`, `distinct_id`, `$cookieless_mode`) are preserved.
- Minimal events skip `$set`/`$set_once` handling; the `$set_once` calculation is skipped (not discarded) so the initial person props aren't marked as sent and still go out with the next full event.

**posthog-node** — dual-source gate:

- A single client-level gate, last-writer-wins across the two signal sources: each v2 `/flags` response (`getFeatureFlagResult`, `evaluateFlags` fallback) and the local-evaluation poller, which reports the top-level `minimal_flag_called_events` field of the definitions payload via a new `onMinimalFlagCalledEvents` callback whenever definitions are (re)loaded — from the API, from a `FlagDefinitionCacheProvider` (the field round-trips through `FlagDefinitionCacheData`), or on quota reset.
- The filter runs in `_prepareEventMessage` before `before_send`, so hooks can re-add properties; everything the SDK adds afterwards (`$groups`, `$lib`/`$lib_version`/`$is_server`, `$geoip_disable`) is itself allowlisted. `$is_server` is kept on minimal events; `locally_evaluated` is kept, `$feature_flag_definitions_loaded_at` is stripped.

**posthog-react-native** — no changes needed beyond `@posthog/core`: RN has no `$feature_flag_called` capture path of its own, so it inherits the gate persistence and minimization from `PostHogCore` (as does posthog-js-lite). Only the generated API reference changed.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Testing

Every package follows the same matrix from the cross-SDK contract: gated + no experiment → exactly the allowlist; gated + experiment → full; ungated / field absent / `has_experiment` absent / bootstrap-only → full; gate persisted across restart; gate flips off when a later response omits the field.

- **`@posthog/core`**: 7 new tests in `posthog.featureflags.spec.ts` — exact key-set assertion with registered super properties proving the strip, flip-off on a later response, and gate survival across a simulated restart on shared storage. Full suite: 44 suites, 776 passed.
- **posthog-js**: 3 gate-persistence unit tests (persisted, never event-visible, flips off on omission and on legacy v1 responses) plus 8 integration tests through the real capture path — exact key set including `$groups` from `posthog.group()` and the transport keys, no top-level `$set_once`, explicit-timestamp capture stays allowlisted (no `$event_time_override_*`), bootstrap-only → full, and a reload backed by the same persistence still sending minimal events. Full suite: 119 suites, 4919 passed.
- **posthog-node**: 9 new tests in `feature-flags.flags.spec.ts` (remote exact key set with `$groups`, `$is_server: true` and `$geoip_disable`, local-eval exact key set, gate read from cached definitions via a cache provider, remote flip-off sequence) and 1 in `evaluate-flags.spec.ts` (minimal vs experiment-full through the snapshot path). Full suite: 28 suites, 828 passed.
- **posthog-react-native** (515 passed) and **posthog-js-lite** (12 passed) suites run green against the core changes.

Lint clean in all touched packages; public API references regenerated.

---


packages/core/src/index.ts
```diff
@@ -4,6 +4,8 @@ export {
   getVariantFromValue,
   parsePayload,
   flagDetailsToResults,
+  MINIMAL_FLAG_CALLED_EVENT_PROPERTIES,
```
dustinbyrne: nit: it looks like `MINIMAL_FLAG_CALLED_EVENT_PROPERTIES` is unused as a package level export

dustinbyrne (approved):

packages/core/src/index.ts
haacked: Removed it.

PR #4173, opened 2026-07-16
feat(consent): add respect_gpc option honoring Global Privacy Control

turnipdabeets:
## Problem

`respect_dnt: true` reads the Do Not Track signal (`navigator.doNotTrack` and legacy variants), but DNT is deprecated as a web standard and no longer sent by all modern browsers. Its successor, Global Privacy Control (`navigator.globalPrivacyControl`), is what privacy-conscious users actually send today (Firefox setting, Brave, DuckDuckGo, extensions) — and unlike DNT it carries legal weight as a recognized opt-out signal under CCPA/CPRA. The SDK currently ignores it, so developers who enable `respect_dnt` for compliance reasonably believe they honor user opt-out signals but miss the one that matters now.

Surfaced by @dustinbyrne's review comment on PostHog/posthog.com#18601.

## Changes

- New `respect_gpc` config option (default `false`): when `true`, `ConsentManager` treats users whose browser sends the GPC signal as opted out — a `_getGpc()` check alongside the existing `_getDnt()`.
- Deliberately a **separate option** rather than folding GPC into `respect_dnt`: extending the existing option would silently change opt-out behavior (and event volume) for every current `respect_dnt` user. Separate and additive means zero behavior change unless explicitly enabled; the `respect_dnt` JSDoc now cross-references it.
- Never-throw: the check is a single optional-chained property read through `isYesLike`; absent/undefined resolves to `false`.
- Tests: GPC respected when enabled, ignored when not, and `respect_dnt` does not react to the GPC signal (28 consent tests pass).
- JSDoc flows into the generated `PostHogConfig` reference automatically (regenerated here) — no api.json churn since the property lives in `@posthog/types`, resolved cross-package by the reference pipeline.

Draft for the consent-behavior design review: mainly whether a separate `respect_gpc` (this PR) or extending `respect_dnt` is preferred, and whether other SDK platforms need parity (GPC is browser-only, so likely not).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Built with Claude Code from a review comment on the docs PR: while adding a DNT-deprecation caveat to the `respect_dnt` docs row, we checked `consent.ts` and found GPC unsupported. Chose the additive option over extending `respect_dnt` to avoid changing behavior for existing users; the PR is a draft to let the team weigh that trade-off.

### DIFF
--- a/.changeset/tidy-donuts-behave.md
+++ b/.changeset/tidy-donuts-behave.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': minor
+'@posthog/types': minor
+---
+
+Added the `respect_gpc` config option. When `true`, users whose browser sends the Global Privacy Control signal (`navigator.globalPrivacyControl`) are treated as opted out of capturing, like `respect_dnt` does for the deprecated Do Not Track signal. Defaults to `false`.
--- a/packages/browser/references/posthog-js-references-latest.json
+++ b/packages/browser/references/posthog-js-references-latest.json
@@ -3220,9 +3220,14 @@
           "name": "opt_in_site_apps"
         },
         {
-          "description": "Determines whether PostHog should respect the Do Not Track header when computing\nconsent in `ConsentManager`.",
+          "description": "Determines whether PostHog should respect the Do Not Track header when computing\nconsent in `ConsentManager`.\n\nNote the Do Not Track signal is deprecated and not sent by all modern browsers;\nsee `respect_gpc` for its successor.",
           "type": "boolean",
           "name": "respect_dnt"
+        },
+        {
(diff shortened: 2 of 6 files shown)


turnipdabeets:
Closing: adding a new public config option isn't warranted here — no user demand, and GPC's CCPA/CPRA weight targets sale/sharing of personal data rather than first-party analytics. The investigation is preserved in this PR if demand appears later.

PR #4189, opened 2026-07-17
fix(ai): report cache-write tokens for OpenAI-compatible providers

Fortiz2305:
## Problem

This completes the Claude-via-OpenRouter cache-cost fix from #3615. #4071 fixed the cache-**read** double-count (via `$ai_cache_reporting_exclusive`), but cache **writes** are still not reported for the OpenAI wrapper, so Anthropic's cache-write premium is never billed. Net effect after #4071: cost is now **under-reported** for cache-heavy Claude-via-OpenRouter generations.

The response usage carries both cache counts (Chat Completions shown; the Responses API uses `input_tokens_details`):

```
usage.prompt_tokens_details.cached_tokens       // cache reads  (already handled)
usage.prompt_tokens_details.cache_write_tokens  // cache writes (currently dropped)
```

`packages/ai/src/openai/index.ts` extracts `cached_tokens` → `cacheReadInputTokens` but never reads `cache_write_tokens`, so `$ai_cache_creation_input_tokens` is never set for this path.

Scope note on cost: PostHog ingestion only prices `$ai_cache_creation_input_tokens` in the **Anthropic branch** of `calculateInputCost`, so the cost correction here applies to Claude/Anthropic models served over OpenAI-compatible endpoints. The SDK forwards the field for any provider that returns it (harmless and useful for observability), but a direct OpenAI/GPT event's computed cost is unchanged until ingestion prices cache writes for those providers too.

## Changes

- Populate `cacheCreationInputTokens` from the response usage's `cache_write_tokens` and thread it into the `usage` passed to `captureAiGeneration`, mirroring the existing cache-read extraction. Covered on **both APIs and both wrappers**:
  - **OpenAI wrapper** (`openai/index.ts`) — Chat Completions (`prompt_tokens_details`): streaming + non-streaming; Responses (`input_tokens_details`): streaming, non-streaming, `parse()`, and background responses.
  - **Azure OpenAI wrapper** (`openai/azure.ts`) — the same six sites; it is a full duplicate of the OpenAI wrapper and had the same omission.
- Read the field through a small `extractCacheWriteTokens` helper in `packages/ai/src/openai/utils.ts` (mirroring `extractRequestId`), reused by both wrappers. `cache_write_tokens` is present in the pinned `openai@6.48.0` types on both `PromptTokensDetails` and `ResponseUsage.InputTokensDetails`.

The emission slot (`captureAiGeneration.ts` emits `$ai_cache_creation_input_tokens`) and the `$ai_cache_reporting_exclusive: false` declaration (`openai/capture.ts`, #4071) already exist — this only supplies the missing count.

- Unit coverage for `extractCacheWriteTokens` in `openai-utils.test.ts` (present / absent / null / undefined / non-object inputs), mirroring `extractRequestId`.
- Cache-write capture asserted in `openai.test.ts` (Chat streaming + non-streaming; Responses streaming, non-streaming, `parse()`) and in `azure-openai.test.ts` (Chat + Responses), checking `$ai_cache_creation_input_tokens`.
- `eslint`, `jest`, and the `@posthog/ai` build pass.

Two real Claude-Sonnet-via-OpenRouter generations, recomputed through the ingestion logic above (rates: input $3/M, cache-read $0.30/M, cache-write $3.75/M, output $15/M), match OpenRouter's billed cost to the cent:

| Generation | prompt / read / write / out | Before (write dropped) | This PR | OpenRouter billed |
|---|---|---|---|---|
| First msg, no prior cache | 24,423 / 0 / 24,423 / 57 | $0.0741 | **$0.0924** | $0.0924 |
| Reused cache | 33,400 / 29,580 / 3,820 / 572 | $0.0289 | **$0.0318** | $0.0318 |

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- Authored with Claude Code (Opus 4.8), directed by @Fortiz2305, following TDD — the cache-write tests (Chat + Responses, OpenAI + Azure) were written and watched fail before the fix landed.
- Review response (addresses @carlos-marchal-ph and @marandaneto): (1) extended the fix to the Responses paths after confirming `ResponseUsage.InputTokensDetails` exposes `cache_write_tokens`; (2) narrowed the changeset/description cost claim to Claude/Anthropic, since ingestion only prices this in the Anthropic branch; (3) added exhaustive `extractCacheWriteTokens` unit coverage; (4) extended the same fix to the Azure wrapper (`openai/azure.ts`), which duplicated the omission.
- Rebased onto the current `main` (the AI package refactor landed meanwhile); resolved the `index.ts` import conflict, re-applied the extraction on the new structure, and re-verified lint/tests/build.
- The cost/ingestion verification table was produced and verified by the human driver against `PostHog/posthog`; the agent did not independently reproduce it (that repo wasn't in the workspace).
- Follow-up left out of scope: the OpenAI Agents SDK tracing processor (`openai-agents/processor.ts`) drops cache-write on its Chat Completions path too — a separate integration.
- Agent-authored — requires human review.


marandaneto:
cc @PostHog/team-ai-observability

marandaneto (approved):

marandaneto:
I checked this against the OpenRouter docs, the OpenAI SDK types, and PostHog's ingestion logic. The underlying Chat Completions issue is real: OpenRouter returns `prompt_tokens_details.cache_write_tokens`, the wrapper currently drops it, and forwarding it lets the Anthropic cost branch correctly apply the cache-write rate.

There are two scope issues I think we should address before merging:

1. **Responses API also exposes cache writes.** The PR description says `InputTokensDetails` only exposes `cached_tokens` in `openai@6.48.0`, but [`ResponseUsage.InputTokensDetails` in v6.48.0 includes `cache_write_tokens`](https://github.com/openai/openai-node/blob/v6.48.0/src/resources/responses/responses.ts#L7260-L7270). OpenRouter also documents cache activity for [both Responses and Chat Completions](https://github.com/OpenRouterTeam/docs/blob/main/guides/best-practices/prompt-caching.mdx#L217-L221). The streaming, non-streaming, and `parse()` Responses extraction paths in `packages/ai/src/openai/index.ts` therefore appear to have the same omission and should use the helper too, with coverage.

2. **The OpenAI-direct cost claim currently overreaches.** In PostHog ingestion, `$ai_cache_creation_input_tokens` is only read and priced inside the Anthropic branch of `calculateInputCost`. For a direct OpenAI/GPT event, merely emitting this property does not yet change the calculated cost. We should either narrow that statement to Claude/Anthropic models served through OpenAI-compatible endpoints or pair it with an ingestion change that handles cache-write pricing for other providers.

So the core fix is valid, but the current implementation is incomplete for Responses and the PR description should narrow or qualify its OpenAI-direct claim.

marandaneto (commented):
check comment https://github.com/PostHog/posthog-js/pull/4189#issuecomment-5088740657

.changeset/openai-cache-write-tokens.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Report cache-write tokens for the OpenAI wrapper. OpenAI-compatible providers (e.g. Claude via OpenRouter) and OpenAI itself on newer models return `usage.prompt_tokens_details.cache_write_tokens` alongside `cached_tokens`, but the wrapper only surfaced cache reads. It now populates `$ai_cache_creation_input_tokens` for both streaming and non-streaming Chat Completions, so ingestion can bill the cache-write premium instead of under-reporting cost on cache-heavy calls. Completes the cache-aware cost fix started in #4071 (#3615).
```
carlos-marchal-ph: This reads as a general claim, but ingestion only prices `$ai_cache_creation_input_tokens` in the Anthropic branch of `calculateInputCost` in PostHog/posthog. For a direct OpenAI/GPT event this property is emitted but doesn't change cost yet. Could we narrow this to Claude-shaped models until ingestion prices cache writes for other providers too?

packages/ai/src/openai/index.ts
```diff
@@ -18,7 +18,7 @@ import type { ResponseCreateParamsWithTools, ExtractParsedContentFromParams } fr
 import type { FormattedMessage, FormattedContent } from '../types'
 import { sanitizeOpenAI, sanitizeOpenAIResponse } from '../sanitization'
 import { extractPosthogParams } from '../utils'
-import { isResponseTokenChunk, extractRequestId, buildProviderMetadata } from './utils'
+import { isResponseTokenChunk, extractRequestId, buildProviderMetadata, extractCacheWriteTokens } from './utils'
```
carlos-marchal-ph: OpenAI's own Responses `InputTokensDetails` in 6.48.0 does expose `cache_write_tokens` (the PR description says it doesn't), and OpenRouter documents cache activity for Responses too. Doesn't that mean the Responses streaming, non-streaming and `parse()` extraction sites below have the same omission, and should use `extractCacheWriteTokens` as well?

packages/ai/src/openai/utils.ts
```diff
@@ -27,6 +27,17 @@ export function extractRequestId(result: unknown): string | undefined {
...
 }
 
+/**
+ * Reads the `cache_write_tokens` field from a Chat Completions
+ * `usage.prompt_tokens_details` object. OpenAI-compatible providers (and OpenAI
+ * itself on newer models) report cache-creation tokens here, but the field is
+ * absent from the pinned SDK's types, so it has to be read through a cast. Used
+ * to populate `$ai_cache_creation_input_tokens`.
+ */
+export function extractCacheWriteTokens(promptTokensDetails: unknown): number {
```
carlos-marchal-ph: `extractRequestId` right above gets exhaustive `it.each` coverage in openai-utils.test.ts (null, undefined, wrong type, absent field). Since this is meant to mirror it, should it get the same treatment there?

carlos-marchal-ph (approved):
Approving to not block, but would appreciate a look at the issues below!

marandaneto:
moving to draft until comments are addressed

.changeset/openai-cache-write-tokens.md
Fortiz2305: Agreed. I'll narrow the changeset and description to Claude/Anthropic over OpenAI-compatible endpoints.

packages/ai/src/openai/index.ts
Fortiz2305: Correct. Sorry, I was checking 6.25.0's types when I did the PR, but it seems it is exposed in 6.48.0, as you said. I'll extend the Responses streaming, non-streaming and `parse()` sites to use `extractCacheWriteTokens` and fix the description.

packages/ai/src/openai/utils.ts
Fortiz2305: yes, it makes sense. I'll do it.

Fortiz2305: Will add the same `it.each` coverage for `extractCacheWriteTokens` in `openai-utils.test.ts` (null, undefined, wrong type, absent field), mirroring `extractRequestId`.

Fortiz2305:
Pushed the review changes:

- **Responses API** — extended the streaming, non-streaming, `parse()` and background-response paths to read `input_tokens_details.cache_write_tokens` via `extractCacheWriteTokens`, with coverage. Confirmed `6.48.0`'s `ResponseUsage.InputTokensDetails` exposes the field — the earlier description claim that it didn't was wrong and is now fixed.
- **Cost claim narrowed** to Claude/Anthropic in the changeset and description, since ingestion only prices `$ai_cache_creation_input_tokens` in the Anthropic branch.
- Added exhaustive `extractCacheWriteTokens` unit coverage mirroring `extractRequestId`.
- Rebased onto the current `main` (the AI package refactor had landed); resolved the `index.ts` conflict and re-verified lint/tests/build.

Moving back to ready for review.

Fortiz2305:
Extended cache-write reporting to the Azure OpenAI wrapper (`openai/azure.ts`), which duplicates the OpenAI wrapper and had the same omission — its six usage-extraction sites (Chat + Responses) now report cache-write, with coverage. Also corrected the helper's JSDoc (the field is present in the now-pinned `openai@6.48.0` types) and the PR description.

One related gap left as a follow-up: the OpenAI Agents SDK tracing processor (`openai-agents/processor.ts`) drops cache-write on its Chat Completions path too — a separate integration, out of this PR's scope.

marandaneto:
@Fortiz2305 

you have to sign and force push your commits (including old ones)
edit: ive done myself

Fortiz2305:
> @Fortiz2305
> 
> > Commits must have verified signatures.
> 
> you have to sign and force push your commits (including old ones) edit: ive done myself

Ups, didn't realize that, thanks for handling it! I've already set up signing on my side now so future commits are verified.

PR #4191, opened 2026-07-17
fix(autocapture): keep server opt-out when remote config fetch fails

turnipdabeets:
## Problem

Customer report ([Zendesk #62711](https://posthoghelp.zendesk.com/agent/tickets/62711)): projects with the project-level autocapture opt-out enabled (`autocapture_opt_out: true`) still received `$autocapture`/`$rageclick` events from some clients on 1.402.3/1.404.0.

Root cause: when the remote config request fails (network error, timeout, non-200 — both the `config.js` script and the JSON fallback), `RemoteConfigLoader` substitutes an empty config so extensions can initialize with defaults (#2895, shipped in 1.350.0; before that the failure path returned early). Autocapture reads the missing `autocapture_opt_out` key as "not opted out" (`!!undefined`), starts capturing, and persists `$autocapture_disabled_server_side: false` — so one failed fetch enables autocapture for the whole page *and* arms the next page load to capture from init. There is no in-session recovery because the refresh interval only reloads flags.

Confirmed in the customer's data: all offending events carry `$autocapture_disabled_server_side: false`, and the burst client had sent events with `true` earlier the same day — only the failed-fetch path can flip a persisted `true` back to `false`.

## Changes

- The config fetch outcome is now an explicit discriminated union (per review): `PostHog._onRemoteConfig` takes `RemoteConfigResult = { ok: true; config: RemoteConfig } | { ok: false }` — no more `{} as RemoteConfig` cast on the failure path.
- Extensions opt into failure semantics via an optional `Extension.onRemoteConfigFailed()`. Autocapture implements it: no key coercion, no persistence write, so `isEnabled` falls back to the last successfully received server value, or stays off if there has never been one. The intentional "missing key on a *successful* response = enabled" semantics are unchanged.
- Extensions without `onRemoteConfigFailed` receive `onRemoteConfig` with a structurally valid empty config, preserving their existing initialize-with-defaults failure behavior (verified: no handler distinguishes absent keys from the empty values).

Behavior matrix on config fetch failure:

| Persisted state | Before | After |
|---|---|---|
| never had a successful config | autocapture **on**, `false` persisted | stays **off**, nothing persisted |
| `$autocapture_disabled_server_side: true` (opted out) | flipped to **on**, `false` persisted | stays **off**, value kept |
| `$autocapture_disabled_server_side: false` (not opted out) | on | on (last known good — unchanged) |

**Trade-off (deliberate):** a *first-time* visitor on a project that is **not** opted out loses autocapture for the whole session if their config fetch transiently fails — there is no persisted last-known-good yet, and the refresh interval only reloads flags, never the config. Returning visitors are unaffected (persisted value wins). We prefer failing closed for one session over capturing on opted-out projects.

Verified with unit tests and an end-to-end Playwright repro against the built `array.js` using the customer's init config: config fetch blocked → clicks produced `$autocapture` before the fix, produce nothing and persist nothing after; opt-out-honoured control path unchanged.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- Investigated and authored with Claude Code (Fable 5): code/history tracing in this repo, a Playwright repro harness against the built bundle, and read-only verification of the affected project's events (via impersonated support session, plain GETs only).
- Considered guarding inside `Autocapture.onRemoteConfig` against any empty response instead of a marker; rejected because a legitimately empty successful config should keep the documented "enabled unless explicitly opted out" behavior — only genuine fetch failure must be treated as unknown.
- Repro gotchas for anyone extending the tests: Playwright's `navigator.webdriver` trips the bot filter (init with `opt_out_useragent_filter: true`) and events batch for ~3s (`request_batching: false` or wait).

### DIFF
--- a/packages/browser/src/extensions/types.ts
+++ b/packages/browser/src/extensions/types.ts
@@ -6,4 +6,10 @@ export type ExtensionConstructor<T extends Extension> = new (instance: PostHog, 
 export interface Extension {
     initialize?(): boolean | void
     onRemoteConfig?(config: RemoteConfig): void
+    /**
+     * Called instead of onRemoteConfig when the remote config could not be fetched.
+     * Implement when a server-controlled setting must not fall back to its default
+     * on failure; extensions without it receive onRemoteConfig with an empty config.
+     */
+    onRemoteConfigFailed?(): void
 }
--- a/packages/browser/src/remote-config.ts
+++ b/packages/browser/src/remote-config.ts
@@ -126,10 +126,9 @@ export class RemoteConfigLoader {
         // whether to show a survey), not during config processing. By the time a linked
         // flag is evaluated, flags have already loaded.
         //
-        // Even when config fails, we still notify extensions so they can initialize
-        // with their defaults, but the payload is marked as failed so extensions whose
-        // server setting must not fall back to a default (e.g. autocapture's opt-out)
-        // can keep their last known value instead.
+        // Even when config fails, we notify extensions so they initialize with their
+        // defaults — marked as failed so settings that must not fail open (e.g.
+        // autocapture's opt-out) can keep their last known value.
         this._instance._onRemoteConfig(config ?? ({ _configLoadFailed: true } as RemoteConfig))
 
         if (config?.hasFeatureFlags !== false) {
--- a/packages/browser/src/remote-config.ts
+++ b/packages/browser/src/remote-config.ts
@@ -127,9 +127,9 @@ export class RemoteConfigLoader {
         // flag is evaluated, flags have already loaded.
         //
         // Even when config fails, we notify extensions so they initialize with their
-        // defaults — marked as failed so settings that must not fail open (e.g.
-        // autocapture's opt-out) can keep their last known value.
-        this._instance._onRemoteConfig(config ?? ({ _configLoadFailed: true } as RemoteConfig))
+        // defaults — as an explicit failure, so settings that must not fail open
+        // (e.g. autocapture's opt-out) can keep their last known value.
+        this._instance._onRemoteConfig(config ? { ok: true, config } : { ok: false })
 
         if (config?.hasFeatureFlags !== false) {
             if (!this._instance.config.advanced_disable_feature_flags_on_first_load) {
--- a/packages/browser/src/remote-config.ts
+++ b/packages/browser/src/remote-config.ts
@@ -126,9 +126,11 @@ export class RemoteConfigLoader {
         // whether to show a survey), not during config processing. By the time a linked
         // flag is evaluated, flags have already loaded.
         //
-        // Even when config fails, we pass an empty object so extensions (autocapture,
-        // session recording, etc.) still initialize with their defaults.
-        this._instance._onRemoteConfig(config ?? ({} as RemoteConfig))
+        // Even when config fails, we still notify extensions so they can initialize
+        // with their defaults, but the payload is marked as failed so extensions whose
+        // server setting must not fall back to a default (e.g. autocapture's opt-out)
+        // can keep their last known value instead.
+        this._instance._onRemoteConfig(config ?? ({ _configLoadFailed: true } as RemoteConfig))
 
         if (config?.hasFeatureFlags !== false) {
             if (!this._instance.config.advanced_disable_feature_flags_on_first_load) {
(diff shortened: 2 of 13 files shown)


packages/browser/src/remote-config.ts
```diff
@@ -126,9 +126,10 @@ export class RemoteConfigLoader {
         // whether to show a survey), not during config processing. By the time a linked
         // flag is evaluated, flags have already loaded.
         //
-        // Even when config fails, we pass an empty object so extensions (autocapture,
-        // session recording, etc.) still initialize with their defaults.
-        this._instance._onRemoteConfig(config ?? ({} as RemoteConfig))
+        // Even when config fails, we notify extensions so they initialize with their
+        // defaults — marked as failed so settings that must not fail open (e.g.
+        // autocapture's opt-out) can keep their last known value.
+        this._instance._onRemoteConfig(config ?? ({ _configLoadFailed: true } as RemoteConfig))
```
dustinbyrne: hmm, this is a dangerous cast because it's _not_ a `RemoteConfig` - that type has a whole bunch of required properties

if this isn't raising an exception elsewhere already, it's just by chance (e.g., if an `onRemoteConfig` listener happens to iterate `supportedCompression`, `siteApps`, etc.)

we should be more explicit here, either update everything to accept an undefined `RemoteConfig` (indicating an error), or we union `RemoteConfig` with some other sort of type that indicates an error

dustinbyrne: for later: it'd be nice to ban explicit unsafe casts like this in eslint, because this shouldn't have gotten in to begin with

dustinbyrne: `(RemoteConfig & { ok: true }) | { ok: false }`?

turnipdabeets: Done in c8af746 — went with the union in this PR since it types the exact value this fix introduces: `_onRemoteConfig` now takes `{ ok: true; config: RemoteConfig } | { ok: false }` (wrapper rather than intersection so the discriminant doesn't leak into the config object handed to extensions), and extensions opt into failure semantics via an optional `onRemoteConfigFailed()` — autocapture implements it to keep its last known opt-out, everything else initializes with defaults from a structurally valid empty config. No `as RemoteConfig` casts remain on this path.

On the eslint idea: agreed — `@typescript-eslint/consistent-type-assertions` with `objectLiteralTypeAssertions: 'never'` would have caught this and the pre-existing `{} as RemoteConfig`. Keeping that as a separate PR since it flags existing violations repo-wide.

packages/browser/src/extensions/types.ts
```diff
@@ -6,4 +6,10 @@ export type ExtensionConstructor<T extends Extension> = new (instance: PostHog,
 export interface Extension {
     initialize?(): boolean | void
     onRemoteConfig?(config: RemoteConfig): void
```
dustinbyrne, line 8: i think it'd be preferred to force `onRemoteConfig` listeners to handle the failure case (instead of opt in to it), otherwise it's easy to ignore and unknowingly not get called if an error occurs

```ts
type RemoteConfigResult = (RemoteConfig | { ok: true }) | { ok: false };

// ...

interface  /* ... */ {
  onRemoteConfig?(config: RemoteConfigResult): void
}

// `onRemoteConfig` methods must now check config.ok before use
```

the current abstraction (`onRemoteConfigFailed`) would allow _other_ extensions which have issues similar to autocapture to remain (sort of; now they won't start). propagating the full `RemoteConfigResult` to the extension forces them to rectify.

re: the discriminant leaking into extensions, i think it's the feature not the bug

dustinbyrne (approved):
kind of getting into semantics so i'm approving (it's all internal anyway)

however - one thing i did not look at, which would be important to understand before merging if you choose to merge this as is: failed remote config requests now won't notify extensions (other than autocapture) at all. i'm _guessing_ this would be a behavioral change in some extensions, and might need to be resolved.

packages/browser/src/extensions/types.ts
turnipdabeets: sounds good, will push a follow up PR for this

PR #4194, opened 2026-07-17
refactor(browser): move shared utilities to browser-common

dustinbyrne:
## Problem

Reusable browser utilities live inside `posthog-js`, which prevents other browser packages from consuming them without depending on the full browser SDK. Keeping browser-local copies also makes shared behavior and singleton state harder to maintain.

## Changes

- Move reusable utility implementations into `@posthog/browser-common` and expose exact utility subpaths with CJS, ESM, and declaration outputs.
- Import browser-common utilities directly from `posthog-js`; remove browser utility facades and avoid custom Rollup resolution.
- Preserve shared browser config/logger state and explicitly mark side-effectful utilities.
- Canonicalize shared constants, cookie reading, and JSON serialization to avoid duplicate implementations.
- Keep browser-specific survey, product-tour, persistence, routing, and event-receiver behavior in `posthog-js`.
- Add browser-common unit coverage and validate the packed package's CJS, ESM, declarations, constants, and side-effect imports.
- Include browser-common in the browser SDK compliance Docker build and workflow path filters.

### Bundle review diff

For reviewer convenience, the following analysis-only diff resolves the baseline and current TypeScript sources through the same browser Rollup TypeScript → Babel pipeline with Terser and the Rslib pre-pass disabled. Generated binding names are normalized (for example, `globals_navigator` and `navigator$1` both become `navigator`) to remove compiler naming noise. The normalized files are for review only and are not executable artifacts.

<details>
<summary>Normalized, same-pipeline <code>array.js</code> git diff</summary>

```diff
diff --git a/baseline/array.js b/current/array.js
index 6e389aa..39c1039 100644
--- a/baseline/array.js
+++ b/current/array.js
@@ -1,27 +1,15 @@
 (function () {
     'use strict';
 
-    // only importing types here, so won't affect the bundle
-    // eslint-disable-next-line posthog-js/no-external-replay-imports
-
     /*
      * Global helpers to protect access to browser globals in a way that is safer for different targets
      * like DOM, SSR, Web workers etc.
      *
-     * NOTE: Typically we want the "window" but globalThis works for both the typical browser context as
-     * well as other contexts such as the web worker context. Window is still exported for any bits that explicitly require it.
-     * If in doubt - export the global you need from this file and use that as an optional value. This way the code path is forced
-     * to handle the case where the global is not available.
+     * Typically we want the browser `window`, but `globalThis` works for browser, workers, and SSR-ish targets.
+     * Export optional globals so callers are forced to handle absence.
      */
 
-    // eslint-disable-next-line no-restricted-globals
     var win = typeof window !== 'undefined' ? window : undefined;
-
-    /**
-     * This is our contract between (potentially) lazily loaded extensions and the SDK
-     * changes to this interface can be breaking changes for users of the SDK
-     */
-
     var global = typeof globalThis !== 'undefined' ? globalThis : win;
 
     // React Native polyfills for posthog-js compatibility
@@ -44,18 +32,12 @@
       return !!(win && win.navigator.onLine !== false);
     }
 
-    var version = "1.404.1";
-    var packageInfo = {
-    	version: version};
-
-    // overridden in posthog-core,
-    // e.g.     Config.DEBUG = Config.DEBUG || instance.config.debug
+    var packageVersion = '0.1.0' ;
     var Config = {
       DEBUG: false,
-      LIB_VERSION: packageInfo.version,
-      LIB_NAME: 'web',
-      /** The actual JS SDK version, unaffected by _overrideSDKInfo. Used for the `ver` request param. */
-      JS_SDK_VERSION: packageInfo.version
+      LIB_VERSION: packageVersion,
+      LIB_NAME: 'browser-common',
+      JS_SDK_VERSION: packageVersion
     };
 
     function asyncGeneratorStep(n, t, e, r, o, a, c) {
@@ -2155,6 +2137,18 @@
       loadScript(posthog, scriptUrl, callback);
     };
 
+    var version = "1.404.1";
+    var packageInfo = {
+    	version: version};
+
+    // overridden in posthog-core,
+    // e.g.     Config.DEBUG = Config.DEBUG || instance.config.debug
+    Config.DEBUG = false;
+    Config.LIB_VERSION = packageInfo.version;
+    Config.LIB_NAME = 'web';
+    /** The actual JS SDK version, unaffected by _overrideSDKInfo. Used for the `ver` request param. */
+    Config.JS_SDK_VERSION = packageInfo.version;
+
     /*
      * Constants
      */
@@ -2231,8 +2225,6 @@
     var INITIAL_REFERRER_INFO = '$initial_referrer_info';
     var INITIAL_PERSON_INFO = '$initial_person_info';
     var ENABLE_PERSON_PROCESSING = '$epp';
-    var TOOLBAR_ID = '__POSTHOG_TOOLBAR__';
-    var TOOLBAR_CONTAINER_CLASS = 'toolbar-global-fade-container';
 
     /**
      * PREVIEW - MAY CHANGE WITHOUT WARNING - DO NOT USE IN PRODUCTION
@@ -2241,7 +2233,6 @@
     var COOKIELESS_SENTINEL_VALUE = '$posthog_cookieless';
     var COOKIELESS_MODE_FLAG_PROPERTY = '$cookieless_mode';
     var WEB_EXPERIMENTS = '$web_experiments';
-    var SDK_DIST_CHANNEL = '$sdk_dist_channel';
     var SDK_DEBUG_EXTENSIONS_INIT_METHOD = '$sdk_debug_extensions_init_method';
     var SDK_DEBUG_EXTENSIONS_INIT_TIME_MS = '$sdk_debug_extensions_init_time_ms';
     var SDK_DEBUG_RECORDING_SCRIPT_NOT_LOADED = '$sdk_debug_recording_script_not_loaded';
@@ -2493,6 +2484,26 @@
       return config.name === TOOLBAR_INTERNAL_INSTANCE_NAME;
     }
 
+    var getCookieValue = name => {
+      if (!document) {
+        return;
+      }
+      try {
+        var nameEQ = name + '=';
+        var cookies = document.cookie.split(';').filter(cookie => cookie.length);
+        for (var i = 0; i < cookies.length; i++) {
+          var cookie = cookies[i];
+          while (cookie.charAt(0) == ' ') {
+            cookie = cookie.substring(1, cookie.length);
+          }
+          if (cookie.indexOf(nameEQ) === 0) {
+            return decodeURIComponent(cookie.substring(nameEQ.length, cookie.length));
+          }
+        }
+      } catch (_unused) {}
+      return null;
+    };
+
     /**
      * uuidv7: An experimental implementation of the proposed UUID Version 7
      *
@@ -2805,25 +2816,7 @@
       _error: function (msg) {
         logger.error('cookieStore error: ' + msg);
       },
-      _get: function (name) {
-        if (!document) {
-          return;
-        }
-        try {
-          var nameEQ = name + '=';
-          var ca = document.cookie.split(';').filter(x => x.length);
-          for (var i = 0; i < ca.length; i++) {
-            var c = ca[i];
-            while (c.charAt(0) == ' ') {
-              c = c.substring(1, c.length);
-            }
-            if (c.indexOf(nameEQ) === 0) {
-              return decodeURIComponent(c.substring(nameEQ.length, c.length));
-            }
-          }
-        } catch (_unused) {}
-        return null;
-      },
+      _get: getCookieValue,
       _parse: function (name) {
         var cookie;
         try {
@@ -3245,6 +3238,10 @@
       return isUndefined(defaultUrl) ? undefined : applyUrlTargetingOverride(instance, defaultUrl);
     }
 
+    var TOOLBAR_ID = '__POSTHOG_TOOLBAR__';
+    var TOOLBAR_CONTAINER_CLASS = 'toolbar-global-fade-container';
+    var SDK_DIST_CHANNEL = '$sdk_dist_channel';
+
     // Node.nodeType integer constants. We use the integers directly rather than
     // `Node.ELEMENT_NODE` etc. for browser portability (IE11) and to avoid a
     // runtime reference to the global `Node` in environments where it isn't defined.
@@ -4619,6 +4616,22 @@
       return undefined;
     };
 
+    var jsonStringify = (data, space) => {
+      try {
+        // Fast path: convert BigInts to strings, since plain JSON.stringify throws on them.
+        // See https://github.com/PostHog/posthog-js/issues/1440.
+        return JSON.stringify(data, (_, value) => typeof value === 'bigint' ? value.toString() : value, space);
+      } catch (_unused) {
+        // A self-referential value — most commonly a DOM node that retains a React fiber pointing back
+        // at the element — makes JSON.stringify throw "Converting circular structure to JSON". With
+        // exception autocapture enabled that throw was recaptured as a new $exception, sometimes in a
+        // tight loop. Fall back to the shared circular-safe serializer (which also handles BigInt and
+        // Errors); it replaces only true cycles with "[Circular]", leaving shared-but-acyclic
+        // references intact. `space` formatting is dropped on this rare path.
+        return safeJsonStringify(data);
+      }
+    };
+
     /**
      * IE11 doesn't support `new URL`
      * so we can create an anchor element and use that to parse the URL
@@ -4816,7 +4829,7 @@
     function _getCampaignParamsFromCookie() {
       var params = {};
       each(COOKIE_CAMPAIGN_PARAMS, function (kwkey) {
-        var kw = cookieStore._get(kwkey);
+        var kw = getCookieValue(kwkey);
         params[kwkey] = kw ? kw : null;
       });
       return params;
@@ -6745,21 +6758,6 @@
       }
       return baseUrl + "?" + updatedSearch.join('&');
     };
-    var jsonStringify = (data, space) => {
-      try {
-        // Fast path: convert BigInts to strings, since plain JSON.stringify throws on them.
-        // See https://github.com/PostHog/posthog-js/issues/1440.
-        return JSON.stringify(data, (_, value) => typeof value === 'bigint' ? value.toString() : value, space);
-      } catch (_unused) {
-        // A self-referential value — most commonly a DOM node that retains a React fiber pointing back
-        // at the element — makes JSON.stringify throw "Converting circular structure to JSON". With
-        // exception autocapture enabled that throw was recaptured as a new $exception, sometimes in a
-        // tight loop. Fall back to the shared circular-safe serializer (which also handles BigInt and
-        // Errors); it replaces only true cycles with "[Circular]", leaving shared-but-acyclic
-        // references intact. `space` formatting is dropped on this rare path.
-        return safeJsonStringify(data);
-      }
-    };
     var encodeToDataString = data => {
       return 'data=' + encodeURIComponent(typeof data === 'string' ? data : jsonStringify(data));
     };
```

</details>

### Validation

- `@posthog/browser-common`: build, lint, and 56 unit tests
- `posthog-js`: production TypeScript build, build, lint, all 120 unit suites (4,953 passed, 8 skipped), and both functional suites (7 passed)
- Frozen lockfile install
- Clean npm consumer installed the packed `posthog-js` and browser-common tarballs: CJS, ESM subpaths, deep `lib` shared-config identity, strict TypeScript declarations, and an executed browser-platform esbuild bundle
- Live US Cloud smoke using the packed npm dependency: `posthog.init()`, synthetic capture accepted by `/e/` (200), remote config loaded (200), feature flags loaded from `/flags/` (200), and `shutdown()` completed
- Isolated browser SDK compliance Docker context build sequence
- `array.js`: 227,749 bytes raw (+83 bytes), 73,175 bytes gzip (-55 bytes)

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. The browser now consumes the same standard package subpaths that are published, without compatibility facades, a utility barrel, or custom Rollup resolution. Shared constants, cookie reading, and JSON serialization are canonicalized to avoid duplicate runtime implementations. The normalized bundle diff above was generated solely as a reviewer aid.


.changeset/calm-otters-share.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/browser-common': minor
+'posthog-js': patch
```
turnipdabeets: [question] posthog-js publishes `lib/*` verbatim, so the moved source files (`lib/src/utils/globals`, `lib/src/uuidv7`, `lib/src/autocapture-utils`, …) disappear from the published package on the next release — and there are real deep-importers of e.g. `posthog-js/lib/src/utils/globals` in the wild. Should we keep thin re-export facades at the old paths (the way `src/request.ts` does for `jsonStringify`), or is dropping them intentional — in which case, worth stating here and maybe more than a patch?

packages/browser/src/customizations/before-send.ts
```diff
@@ -1,7 +1,7 @@
 import { BeforeSendFn, CaptureResult, KnownEventName } from '../types'
 import { includes, clampToRange } from '@posthog/core'
 import { appendArray, sampleOnProperty, updateThreshold } from '../extensions/sampling'
-import { logger } from '../utils/logger'
+import { logger } from '@posthog/browser-common/utils/logger'
```
turnipdabeets, line 4: [question] This file's published `lib/` output now imports `@posthog/browser-common/utils/*`, which only resolves via the exports map (no physical files, es2023 dist). Users on exports-map-unaware toolchains (webpack 4, older Metro/jest) who deep-import `posthog-js/lib/src/customizations` — which our docs recommend — would get "Module not found" on upgrade. Is dropping those toolchains intentional, or should browser-common ship physical stub dirs (the `customizations/package.json` pattern posthog-js already uses)? Either way, maybe we should also update the docs to point at the bundled `posthog-js/customizations` subpath, which stays safe on any resolver.

packages/browser-common/src/utils/globals.ts
```diff
@@ -0,0 +1,153 @@
...
+    markAsRead: (ticketId?: string) => Promise<any>
+    getTickets: (options?: any) => Promise<any>
+    requestRestoreLink: (email: string) => Promise<any>
+    restoreFromToken: (restoreToken: string) => Promise<any>
+    restoreFromUrlToken: () => Promise<any | null>
+    getCurrentTicketId: () => string | null
+    getWidgetSessionId: () => string
+}
+
+export interface PostHogExtensions {
```
turnipdabeets: [suggestion] The pre-move `globals.ts` typed this contract against `PostHog`/`SessionRecordingStatus`/`eventWithTime` and carried a warning that changes here can be breaking for SDK users; the moved copy is all `any`, so the compiler no longer checks the main-bundle ↔ lazy-bundle seam and the looser types flow into the published d.ts. Could we keep the strict shapes — via `@posthog/types` here, or a declaration-merge back in posthog-js?

packages/browser/src/customizations/setAllPersonProfilePropertiesAsPersonPropertiesForFlags.ts
```diff
@@ -1,12 +1,14 @@
 import type { PostHogConfig, PostHogInterface } from '../types'
+import '../config'
```
turnipdabeets, line 3: [suggestion] This bare import is load-bearing now: it's the only thing installing the `web`/real-version override into browser-common's `Config` before `getEventProperties` reads `$lib`/`$lib_version` in this standalone bundle — an "unused import" cleanup would silently ship `$lib: 'browser-common'`. Worth a comment so it survives:
```suggestion
// Side-effect import: overrides @posthog/browser-common Config with the 'web' LIB_NAME
// and real version before getEventProperties reads them. Do not remove as unused.
import '../config'
```

turnipdabeets (approved):
Left some questions but LGTM, approving to unblock

packages/browser-common/src/utils/globals.ts
dustinbyrne: yeah, this is so true - the introduction of `any` is a good indicator that there's some sort of mismatch here. i split the `assignableWindow` back out into `browser`

packages/browser/src/customizations/before-send.ts
dustinbyrne: turns out we don't have great support for webpack 4 anyhow due to `@posthog/core` imports. i spent some time looking into re-adding support, but considering it's _been_ broken, i'm going to ignore it

i'll fix the docs in a separate pr

.changeset/calm-otters-share.md
dustinbyrne: i'm feeling like needing to maintain everything in `lib` (literally every file) as public API is too onerous, especially considering this change is scoped strictly to util.

bumping to a minor seems reasonable though

PR #686, opened 2023-06-07
Update utils.ts to add Prerender bot string matching

simfish85:
Added prerender to blocked user agent string

## Changes

...

## Checklist
- [ ] Tests for new code (see [advice on the tests we use](https://github.com/PostHog/posthog-js#tiers-of-testing))
- [ ] Accounted for the impact of any changes across different browsers

PR #4195, opened 2026-07-18
fix(surveys): re-translate displayed surveys when user language changes

### LINKED ISSUE #4174: Re-translate displayed survey on language change
## Summary

The browser SDK resolves a survey's translation on every render attempt, but never re-translates a survey that's already on screen. If the `language` person property (or the browser language) changes while a popover or widget is displayed, the survey stays in the old language until the next fresh render.

## Current State

- Language resolution runs per render attempt in `_translateSurveyForRendering` (`packages/browser/src/extensions/surveys.tsx`): `override_display_language` → person property `language` → `navigator.language`.
- Once a popover is shown the poll loop skips it (`_surveyInFocus` guard), and `_handleWidget` early-returns when the widget is already mounted, so a displayed survey is never re-resolved.
- The gap is already called out as a TODO in `packages/browser/src/utils/survey-translations.ts`: "Consider adding dynamic language change detection in the future … Re-render survey when language changes mid-session."

## Expected Behavior

When the resolved display language changes while a survey is on screen (e.g. `identify()` / `setPersonPropertiesForFlags()` sets a new `language`, or the browser fires `languagechange`), the SDK should re-resolve the translation and update the rendered survey in place. `$survey_language` on subsequent survey events should match what's actually shown.

## Reference Implementation

posthog-ios: https://github.com/PostHog/posthog-ios/pull/686 subscribes to person-property changes, re-resolves the active survey's translation, and pushes the update to the rendering delegate, no-oping when the matched language is unchanged.

## Related

- posthog-js (React Native): https://github.com/PostHog/posthog-js/issues/4175
- posthog-android: https://github.com/PostHog/posthog-android/issues/628
- posthog-flutter: https://github.com/PostHog/posthog-flutter/issues/483

nandinitiw:
## What

When a survey popover is on screen, it was never re-translated if the user's language changed mid-session. This happened in two scenarios:
1. `posthog.identify()` / `setPersonPropertiesForFlags()` sets a new `language` person property
2. The browser fires a `languagechange` event (OS/browser language switch)

The TODO comment in `survey-translations.ts` already called this out as a known gap.

Also fixes a gap this surfaced: `$survey_questions[].question` and `$survey_language` were read at capture time, not answer time — if the language changed mid-survey, submitted events showed the *new* language's question text even for questions the user answered under the *old* language.

## How

In `SurveyManager`:
- Register a `window.languagechange` listener in the constructor
- Subscribe to `posthog.onFeatureFlags()` (which fires after `identify()` reloads flags)
- Both call `_onLanguageChange()`, which:
  - No-ops if no survey is in focus, or if the survey hasn't actually rendered yet (gates against a delayed survey — `surveyPopupDelaySeconds` — being forced on screen early by a mid-delay reload, skipping the eligibility recheck the delay timeout does)
  - Re-detects the resolved language
  - No-ops if the language is unchanged (idempotent, safe to call on every flag reload)
  - Otherwise updates `_currentLanguage` and re-renders the popover via `renderPopover()`
- Listeners are cleaned up on `destroy()`
- `_currentLanguage` is cleared when a survey is dismissed

**`questionSnapshots`**: `onNextButtonClick` now snapshots the currently-displayed question text into `questionSnapshots: Record<string, string>` (question id → text) before advancing, and persists it to `InProgressSurveyState` in localStorage. `buildSurveyResponseProperties` in `@posthog/core` now accepts this optional param and prefers snapshot text over the current (possibly re-translated) survey question text, so `$survey_questions` reflects what the user actually saw when they answered. `$survey_language` on `dismissed` now similarly prefers the in-progress language over the current display-time language.

The iOS reference implementation in [posthog-ios #686](https://github.com/PostHog/posthog-ios/pull/686) follows the same pattern.

## Tests

Unit tests covering:
- Re-renders when language differs
- Does not re-render when language is unchanged
- No-ops when no survey is in focus
- Does not re-render a survey that's in focus but not yet rendered (pending delay)
- Clears `_currentLanguage` on dismiss
- Removes `languagechange` listener on `destroy()`
- `questionSnapshots` recorded and used in place of current question text
- A real end-to-end render test: renders a popup via the unmocked `handlePopoverSurvey`, types an answer, fires a real `languagechange` event, and asserts both that the rendered question text updates and that the typed answer survives the re-render

Closes #4174


nandinitiw:
Hey @adboio @lucasheriques — would love a review on this when you get a chance! It closes the TODO in `survey-translations.ts` for dynamic language change detection.

turnipdabeets:
@ioannisj related to what you’re doing with survey as well?

packages/browser/src/extensions/surveys.tsx
```diff
@@ -131,11 +131,49 @@ export class SurveyManager {
...
+    private _onLanguageChange(): void {
+        if (isNull(this._surveyInFocus)) {
+            return
+        }
+        const surveys = this._posthog.get_property(SURVEYS) as Survey[] | undefined
+        const survey = surveys?.find((s) => s.id === this._surveyInFocus)
+        if (!survey) {
+            return
+        }
+        const { language: newLanguage } = this._translateSurveyForRendering(survey)
```
ioannisj: Nice, the re-translation itself reads well. One thing I think is off though: the events don't follow the same rule as the display. A `survey sent` or `survey dismissed` fired after a language switch stamps every `$survey_questions[].question` in the final language, including questions the user already answered in the previous one. 

I recently added this same live re-translation on iOS (https://github.com/PostHog/posthog-ios/pull/686), and I'm making the events report the answer-time language there, so it'd be good to keep the two SDKs aligned. 

I think the fix is to snapshot each question's displayed text at answer time into the in-progress state and build `$survey_questions` from that, and set `$survey_language` on sent/dismissed to the answer-time value.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -131,11 +131,49 @@ export class SurveyManager {
...
+        if (newLanguage === this._currentLanguage) {
+            return
+        }
+        this._currentLanguage = newLanguage
+        this.renderPopover(survey)
```
ioannisj: I think `renderPopover` drops some props on this re-render. It only forwards `survey`, `removeSurveyFromFocus`, and `surveyLanguage`, but `handlePopoverSurvey` originally renders the popover with `style` (the computed `NextToTrigger` position), `properties`, and `isSurveyCompleted` too. So re-rendering here snaps a `NextToTrigger`-anchored survey back to its default position and loses any custom event `properties` from later sent/dismissed events. Can you please validate if this is indeed the case?

packages/browser/src/extensions/surveys.tsx
```diff
@@ -131,11 +131,49 @@ export class SurveyManager {
...
+    private _languageChangeListener: (() => void) | null = null
+    private _unsubscribeFeatureFlags: (() => void) | null = null
 
     constructor(posthog: PostHog) {
         this._posthog = posthog
         // This is used to track the survey that is currently in focus. We only show one survey at a time.
         this._surveyInFocus = null
+
+        this._languageChangeListener = () => this._onLanguageChange()
+        window.addEventListener('languagechange', this._languageChangeListener)
```
ioannisj: I think here we should use the`addEventListener` wrapper the rest of this file uses. CI will prob trip on this

ioannisj (commented):
Left a few inline comments. The main ones is that the survey events don't follow the same language rule as the display after a mid-survey switch, and maybe a miss in survey style when switching language. everything else is smaller. Thanks for picking up that TODO!

We'll also need to add a changeset entry for this

nandinitiw:
Thanks for the review! Pushed fixes for all three:

1. **`addEventListener` wrapper** — switched from `window.addEventListener` to the `addEventListener` util used elsewhere in the file.
2. **Dropped render props** — now snapshots `style`, `properties`, and `isSurveyCompleted` from the initial `handlePopoverSurvey` call and restores them on language-change re-renders, so `NextToTrigger` positioning and custom event properties are preserved.
3. **Changeset** — added.

On the answer-time language question: since Preact's `render()` diffs against the existing shadow root rather than remounting, the component state (current question index, in-progress responses) is preserved across re-renders. Each subsequent `sendSurveyEvent` call passes `surveyLanguage` from the component's current prop, so events filed after a language switch correctly reflect the new language, and events filed before the switch reflect the old one. Let me know if you'd like a different approach here — happy to snapshot the displayed question text at answer time if that's the preferred alignment with iOS.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -131,11 +131,62 @@ export class SurveyManager {
...
+            window.removeEventListener('languagechange', this._languageChangeListener)
+            this._languageChangeListener = null
+        }
+        if (this._unsubscribeFeatureFlags) {
+            this._unsubscribeFeatureFlags()
+            this._unsubscribeFeatureFlags = null
+        }
+    }
+
+    private _onLanguageChange(): void {
```
ioannisj, line 166: Hey, nice fix. One small thing on the delayed path: for a survey with `surveyPopupDelaySeconds > 0`, `_addSurveyToFocus` sets `_surveyInFocus` before the actual render is deferred into the `setTimeout` (can you please validate if this is the case). 

So if a `languagechange` or an `onFeatureFlags` reload fires during that delay window with a different language (e.g. from `identify()`), `_onLanguageChange` sees the survey in focus and renders it before the delay elapses and without the `_shouldDisplaySurvey` recheck the timeout does.

That recheck is there specifically because `identify()` can flip targeting off mid-delay, and `onFeatureFlags` is exactly the hook this PR subscribes to, so the same reload can both make the survey ineligible and force it on screen early.

Could we gate `_onLanguageChange` so it only re-renders a survey that's already on screen?

packages/ai/src/utils.ts
```diff
@@ -126,6 +126,7 @@ export const getModelParams = (
     'language',
     'response_format',
     'timestamp_granularities',
+    'service_tier',
```
ioannisj, line 130: Looks unrelated. Probably need to merge latest main

nandinitiw:
Addressed both comments from the latest review:

**Delayed render race** (re: `_addSurveyToFocus` / `_onLanguageChange`): confirmed the race. Added a `_surveyIsRendered: boolean` flag — it's set to `true` only when the survey actually renders (immediately when `delaySeconds <= 0`, or inside the `setTimeout` callback after the delay elapses). `_onLanguageChange` now short-circuits if `!this._surveyIsRendered`, so any language or feature-flag reload that fires during the delay window is a no-op. The `_shouldDisplaySurvey` recheck in the timeout still runs before `_surveyIsRendered` is set, so the gate can't be bypassed.

**Unrelated `utils.ts` diff**: rebased on latest main, gone now.

nandinitiw:
Implemented answer-time language snapshotting to align with the iOS behavior:

**`$survey_questions[].question`**: `onNextButtonClick` now snapshots the current question text (from the currently displayed `survey.questions`) into `questionSnapshots: Record<string, string>` (question id → displayed text) before advancing. This is persisted to `InProgressSurveyState` in localStorage so it survives page restores. `buildSurveyResponseProperties` in `@posthog/core` now accepts an optional `questionSnapshots` param and prefers those over the current survey question text when building the array, so each entry reflects what the user actually saw.

**`$survey_language` on `sent`**: passed `questionSnapshots` through to `sendSurveyEvent`, and the `surveyLanguage` passed in is already the language at the time `onNextButtonClick` fires (the component prop at that moment), so this was correct.

**`$survey_language` on `dismissed`**: `dismissedSurveyEvent` already read `inProgressSurvey.surveyLanguage` via `_buildSurveyEventProperties`, but then overrode it with the current display-time prop. Now it prefers the in-progress language and only falls back to the current prop when there's no in-progress state (i.e. the user dismissed without answering anything).

ioannisj (approved):
LG, just this https://github.com/PostHog/posthog-js/pull/4195#discussion_r3655826219 here (probably left over from https://github.com/PostHog/posthog-js/pull/4072

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1645,4 +1646,104 @@ describe('surveys', () => {
...
+            ],
+        } as unknown as Survey
+
+        beforeEach(() => {
+            surveyManager = (surveys as any)._surveyManager
+            instance.get_property = jest.fn().mockReturnValue([frSurvey])
+            instance.onFeatureFlags = jest.fn().mockReturnValue(() => {})
+        })
+
+        it('updates _currentLanguage and re-renders when languagechange fires and language differs', () => {
```
lucasheriques, line 1677: these set `_surveyInFocus` / `_currentLanguage` / `_surveyIsRendered` directly and stub `_translateSurveyForRendering`, so they assert internal state rather than what the user sees. they would pass even if the popup never re-rendered.

the claim that answers survive the switch has no test. one that renders, types into a question, fires `languagechange`, then asserts the label changed and the typed value did not would cover it. `__tests__/extensions/surveys.test.ts:261` has a `describe('SurveyManager')` with fixtures if that is easier.

lucasheriques (commented):
the re-translate path holds up. checked whether a language switch wipes in-progress answers and it does not: `retrieveSurveyShadow` returns the existing shadow root and nothing re-keys, so preact patches in place.

four red checks, three are one-liners:

- lint: prettier at `surveys.test.ts:1682` and `:1696`, so `pnpm lint:fix`
- mangled names: 6 new private fields, so `cd packages/browser && pnpm write-mangled-property-names` and commit the file
- unit tests: `questionSnapshots` now rides along in the `sendSurveyEvent` call, and 2 pre-existing tests in `survey-popup.test.tsx` pin the exact object, so they need the key added
- hygiene: the script itself passed and its warning is real (`@posthog/core` changed, wants `"@posthog/core": patch`). the red x is ours though: that job posts a pr comment and fork prs get a read-only token, so it 403s on every external contribution. ignore it.

worth adding `questionSnapshots` to the description too, since it is why core changed.

one note inline.

nandinitiw:
Thanks for the thorough review! Addressed everything:

- **Lint**: fixed the two prettier failures in `surveys.test.ts` via `pnpm lint:fix`.
- **Mangled names**: ran `pnpm write-mangled-property-names` and committed `terser-mangled-names.json` — 7 new private fields registered (`_currentLanguage`, `_languageChangeListener`, `_onLanguageChange`, `_surveyIsRendered`, `_surveyPopupProps`, `_unsubscribeFeatureFlags`, plus `_surveyInFocus` picked up the same pass).
- **Unit tests**: added `questionSnapshots` to the two pinned `sendSurveyEvent` assertions in `survey-popup.test.tsx`.
- **Hygiene red X**: makes sense, ignoring — thanks for the context on the fork-token 403.
- **Changeset**: added `@posthog/core` alongside `posthog-js`, and updated both the changeset body and PR description to call out `questionSnapshots` as the reason core changed.
- **Inline comment**: you're right that the existing test only asserted internal state and would pass even if nothing rendered. Replaced it with a real end-to-end test — renders via the unmocked `handlePopoverSurvey`, types into the actual textbox, fires a real `languagechange` event (with `navigator.language` changed), and asserts both that the visible question text updates to the translation and that the typed answer is preserved across the re-render. Used the `describe('SurveyManager')` fixtures in `__tests__/extensions/surveys.test.ts` as suggested.

One thing I found along the way: the existing `SurveyManager` tests in that describe block never called `.destroy()` between tests, so each test's `languagechange` listener stayed attached to `window`. My new test's real `dispatchEvent` call was the first to actually trigger those stale listeners, which crashed on a `mockPostHog` whose mocks had been cleared. Added an `afterEach` that calls `surveyManager.destroy()` to fix the leak for the whole block, not just my test.

lucasheriques:
thanks fro the improvementes on the tests! @nandinitiw can you fix the last conflicts so we can get this merged?

nandinitiw:
Rebased on latest main and resolved the conflict in `surveys.tsx` — main had refactored the delayed-render path into `renderAfterDelay`/`renderIfStillEligible` helpers (for the delay-resumption-across-navigations work), so I merged `_surveyIsRendered = true` into the new `renderAfterDelay` function rather than the old inline callback. Verified the full survey test suite (184 tests) and lint pass, and regenerated `terser-mangled-names.json` against the rebased main — no further changes needed there. Ready for another look whenever you get a chance!

nandinitiw:
Reopening — this was auto-closed by the stale bot, not for cause. lucasheriques's last review was positive; picking this back up.

nandinitiw:
This was auto-closed by the stale bot after a week of inactivity, not for cause — the last review from @lucasheriques was positive and all requested changes had already been addressed. GitHub's reopen API is failing with an opaque validation error I couldn't resolve, so I've opened a fresh PR from the same branch with the identical diff: #4626

PR #4196, opened 2026-07-19
fix(ai): sanitize base64 content in $ai_trace/$ai_span input/output state

reclaim-admin:
## Problem

`_setLLMMetadata` runs generation input through `sanitizeLangChain` (BinaryContentRedactor), but `_popRunAndCaptureTraceOrSpan` captures `run.input` and `outputs` **raw** into `$ai_input_state` / `$ai_output_state`.

Since every `withStructuredOutput().invoke()` (and any chain) emits a `$ai_trace`/`$ai_span`, an image call embeds its full base64 data URLs in the trace event. In production we observed 1–4.5 MB events that ingest rejects with:

```
HTTP 413 — maximum event size exceeded: Event rejected by kafka during send
```

The whole batch POST is dropped after retries — **including unrelated `$exception` events queued alongside**, which silently degrades error tracking. In short-lived runtimes (Supabase edge functions) the timer-driven flush rejection also surfaced as an uncaught `event loop error`, terminating the worker.

Observed on `@posthog/ai` 8.2.1; the gap is still present on `main`.

## Fix

Apply `sanitizeLangChain` to `$ai_input_state` and `$ai_output_state` at capture time in `_popRunAndCaptureTraceOrSpan`, mirroring the generation path. Sanitizing at the capture site covers all sources that feed `_setTraceOrSpanMetadata` (chains, tools, retrievers, agent actions).

## Testing

- New regression test: a chain start/end carrying a base64 image data URL asserts both state properties contain `[base64 image/jpeg redacted]` (and not the payload) while normal text survives.
- `tests/callbacks.test.ts` (14 passed), `tests/sanitization.test.ts` + `tests/binary_content_redactor.test.ts` (90 passed).

### DIFF
--- a/.changeset/gentle-traces-rest.md
+++ b/.changeset/gentle-traces-rest.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Redact base64 content from LangChain trace and span input and output state.
--- a/packages/ai/src/langchain/callbacks.ts
+++ b/packages/ai/src/langchain/callbacks.ts
@@ -409,7 +409,7 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       $ai_lib: 'posthog-ai',
       $ai_lib_version: version,
       $ai_trace_id: traceId,
-      $ai_input_state: withPrivacyMode(this.client, this.privacyMode, run.input),
+      $ai_input_state: withPrivacyMode(this.client, this.privacyMode, sanitizeLangChain(run.input)),
       $ai_latency: latency,
       $ai_span_name: run.name,
       $ai_span_id: runId,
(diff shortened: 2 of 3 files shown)


marandaneto:
cc @PostHog/team-ai-observability

marandaneto (approved):

marandaneto:
cc @PostHog/team-ai-observability

carlos-marchal-ph (approved):
Hi @reclaim-admin, thanks for the contribution! We are actually working on a big refactor for the sanitisation plumbing ([already up for review for Python](https://github.com/PostHog/posthog-python/pull/762)), that should fix not only this issue, but many adjacent ones. Hoping to get it out early this week, merging this for the moment as it is the correct fix for this specific issue!

marandaneto:
@reclaim-admin 

can you force push your commits with verified signatures so we can merge this

reclaim-admin:
@marandaneto Done — rebased onto latest `main` and squashed into a single verified commit (7b76817), including your changeset and formatting fixes (kept you as co-author). Should be good to merge now. Apologies for the brief auto-close, that was a side effect of the force-push.

PR #4197, opened 2026-07-19
fix(browser): clear register_for_session props when PostHog session rotates

okxint:
## What does this PR do?

Fixes #3573

`register_for_session` stores properties in `sessionPersistence`, which uses `sessionStorage` under the hood. `sessionStorage` only clears when the browser tab closes — so if PostHog's idle timeout (30 min default) fires and generates a new `$session_id`, the session-registered props silently carry over into the next session. This is inconsistent with the documented behaviour that these properties are scoped to the current session.

### Root cause

`sessionPersistence` is backed by `sessionStorage` (or the same store as primary persistence when `persistence: 'sessionStorage'`). The PostHog session lifecycle (idle timeout, max length) and the browser's `sessionStorage` lifecycle are independent. When a new `$session_id` is generated the `sessionPersistence` blob is not cleared.

### Fix

- Added a `_sessionRegisteredPropKeys: Set<string>` instance field that tracks every key written via `register_for_session`.
- Added a `_clearSessionRegisteredProps()` helper that unregisters only those keys from `sessionPersistence`.
- Registered a `sessionManager.onSessionId` listener during `init_()` that calls `_clearSessionRegisteredProps()` when the change reason is `activityTimeout` or `sessionPastMaximumLength`. Cross-tab adoption and first-ever-session (`noSessionId`) are intentionally excluded.
- `unregister_for_session` now removes the key from the tracked set too.
- `reset()` already calls `sessionPersistence.clear()`, so it only needs to clear the set (no unregister loop needed).

### What is NOT affected

System-managed keys written directly to `sessionPersistence` (campaign params, referrer, search keyword) are untouched because they are not registered via `register_for_session` and are re-populated by the capture pipeline on the next event anyway.

## Tests

Four new unit tests covering:
1. Props cleared when `activityTimeout` fires
2. Props NOT cleared on first-ever session (`noSessionId`)
3. Only user-registered keys cleared, system keys preserved
4. Key removed from tracking set when `unregister_for_session` is called before rotation

```
register_for_session()
  ✓ clears session-registered props when PostHog session rotates due to activity timeout
  ✓ does not clear session-registered props on initial session creation
  ✓ clears only user-registered keys, not system-managed session storage keys
  ✓ removes keys from the tracked set when unregister_for_session is called
```

### DIFF
--- a/packages/browser/src/__tests__/posthog-core.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.test.ts
@@ -1,4 +1,4 @@
-import { defaultPostHog } from './helpers/posthog-instance'
+import { createPosthogInstance, defaultPostHog } from './helpers/posthog-instance'
 import type { PostHogConfig } from '../types'
 import { uuidv7 } from '../uuidv7'
 import { SurveyEventName, SurveyEventProperties } from '../posthog-surveys-types'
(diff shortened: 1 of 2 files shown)


okxint:
Heads up: the Mendral supply chain review is the only failing check — all other CI (tests, linting) pass. This PR introduces no new dependencies (pure TypeScript, no new packages). The Mendral check likely needs a team member to approve the scan result. Happy to answer any questions!

packages/browser/src/__tests__/posthog-core.test.ts
```diff
@@ -613,4 +613,82 @@ describe('posthog core', () => {
...
+            // Simulate a session rotation caused by activity timeout
+            const handlers: any[] = []
+            const realOnSessionId = posthog.sessionManager!.onSessionId.bind(posthog.sessionManager)
+            jest.spyOn(posthog.sessionManager!, 'onSessionId').mockImplementation((cb) => {
+                handlers.push(cb)
+                return realOnSessionId(cb)
+            })
+
+            // Fire the change handler directly as if a session timed out
+            ;(posthog as any)._clearSessionRegisteredProps()
```
ioannisj, line 634: Could we fire the real `_sessionIdChangedHandlers` with `{ activityTimeout: true }`, the way the `noSessionId` test just below does, so the gating logic is covered? As it stands a regression in that condition would still pass here.

packages/browser/src/posthog-core.ts
```diff
@@ -442,6 +442,8 @@ export class PostHog implements PostHogInterface {
     analyticsDefaultEndpoint: string
     version: string = Config.LIB_VERSION
     _initialPersonProfilesConfig: 'always' | 'never' | 'identified_only' | null
+    // Keys registered via register_for_session — cleared when the PostHog session rotates
+    _sessionRegisteredPropKeys: Set<string> = new Set()
```
ioannisj, line 446: _sessionRegisteredPropKeys is an in-memory Set and sessionPersistence defaults to sessionStorage, which survives a page reload within the tab. So if someone registers a prop, reloads, then the session rotates, the Set is empty and nothing gets clearedWould persisting the registered keys (e.g. a namespaced sessionStorage entry) so cleanup can rebuild them after a reload make sense here?

ioannisj:
@okxint friendly ping here

packages/browser/src/posthog-core.ts
```diff
@@ -705,6 +707,15 @@ export class PostHog implements PostHogInterface {
         if (!startInCookielessMode) {
             this.sessionManager = new SessionIdManager(this)
             this.sessionPropsManager = new SessionPropsManager(this, this.sessionManager, this.persistence)
+            // Clear user-registered session properties when a new PostHog session starts due to
+            // idle timeout or hitting the max session length. Properties registered via
+            // register_for_session are meant to be session-scoped, but sessionStorage persists
+            // until the browser tab closes — without this they leak into the next session.
+            this.sessionManager.onSessionId((_sessionId, _windowId, changeReason) => {
+                if (changeReason?.activityTimeout || changeReason?.sessionPastMaximumLength) {
```
marandaneto, line 715: blocking: A cross-tab adoption replaces this tab’s current PostHog session too, but this predicate leaves properties from the previous session registered. I reproduced this by firing the session handlers with `crossTabAdoption: true`; the old property remains. Could we clear for `changeReason.crossTabAdoption` as well and add coverage for that rotation path?

marandaneto:
Thanks @okxint for the original fix. This is now superseded by #4414, which preserves your authored commit and adds the reload persistence, cross-tab rotation, callback coverage, and changeset requested during review. Closing this PR in favor of the canonical replacement.

PR #4199, opened 2026-07-19
fix: use pending replay buffer for linked flag triggers

pauldambra:
## Problem

Linked-flag-only session recording waits in `buffering` state but uses the normal five-minute full-snapshot interval instead of the one-minute pending-trigger interval. This can retain more replay data in memory than intended while waiting for the flag.

## Changes

- Include linked flags in the existing V1 ANY/ALL trigger-status matcher.
- Add a regression test that fails on `main` before the fix, plus guards for ANY and ALL combinations.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

PostHog Code audited fixes removed from the configurable-buffer PR using the ReviewHog validation criteria. Only this linked-flag behavior reproduced independently on `main`; the regression test was committed while failing before the production fix was added.

---


packages/browser/src/extensions/replay/external/recording-strategies.ts
```diff
@@ -174,12 +174,13 @@ export class V1RecordingStrategy implements RecordingStrategy {
         this._sampleRate = isNumber(config.sampleRate) ? config.sampleRate : null
 
         // Setup trigger matching strategy (AND vs OR)
+        const triggerMatchers = [this._eventTriggerMatching, this._urlTriggerMatching, this._linkedFlagMatching]
```
pauldambra, line 178: > [!NOTE]

**[router]** 🟡 MEDIUM

Adding `_linkedFlagMatching` to `_triggerStatusMatcher` also changes the guard in `_activateTrigger` (`lazy-loaded-session-recorder.ts:854`), which early-returns when `hasPendingTriggers()` is false.

Under `triggerMatchType: 'any'` with a linked flag **plus** an event/url trigger: if the flag matches first, the OR matcher reports `trigger_activated`, so a later event/url trigger firing is no longer persisted to `SESSION_RECORDING_EVENT_TRIGGER_ACTIVATED_SESSION` and `_reportStarted('event_trigger_matched')` is skipped. If the flag then reverts mid-session (`linkedFlagSeen` is reset to `false` on every `onFeatureFlags` callback — see the existing test at `lazy-sessionrecording.test.ts:3618-3620` where status returns to `buffering`), the session stops recording even though an `any` trigger legitimately matched earlier; previously the persisted trigger kept it active. There is also a minor telemetry change ($recording_started reason no longer emitted for the second trigger).

Narrow edge, but real. Consider persisting the individual trigger activation regardless of the composite status (guard on that matcher's own status), or confirm this dedupe is intended.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -3543,6 +3543,61 @@ describe('Lazy SessionRecording', () => {
...
+        it('uses the pending trigger snapshot interval while waiting for a linked flag', () => {
+            sessionRecording.onRemoteConfig(
+                makeFlagsResponse({ sessionRecording: { endpoint: '/s/', linkedFlag: 'the-flag-key' } })
+            )
+
+            expect(sessionRecording.status).toEqual('buffering')
+            expect(sessionRecording['_lazyLoadedSessionRecording']['_fullSnapshotIntervalMillis']).toEqual(60_000)
+        })
+
+        it('stops rotating the buffer after one of several OR triggers activates', () => {
```
pauldambra: > [!NOTE]

**[router]** ⚪ NIT

Of the three new tests, only the first (`uses the pending trigger snapshot interval`) and third (`keeps rotating the buffer while an ALL trigger remains pending`) fail without the source change — verified by reverting `recording-strategies.ts` and re-running. `stops rotating the buffer after one of several OR triggers activates` passes both ways. That's fine as a regression guard against the OR-stays-pending failure mode; just noting it doesn't pin the fix itself.

pauldambra (commented):
> [!NOTE]

Multi-perspective review complete (cheap-first router pass; no delegation warranted for a 3-line source change). See inline comments.

pauldambra:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers as warranted

## Verdict: 💬 APPROVE WITH NITS <sub>(round 2 @ 1cff6e8)</sub>

Round 2 reviewed the round-1 test refactor (collapsing the two OR/ALL buffer tests into one `it.each`) and found it clean: assertions preserved per case, matches the file's existing `it.each` convention, lint + typecheck green. No new findings. The one substantive open item remains the round-1 MEDIUM below (deferred as a design decision).

### Key open findings

- 🟡 **MEDIUM** — `recording-strategies.ts:177`: side effect on the `_activateTrigger` guard under `triggerMatchType: 'any'` with a linked flag + event/url trigger. A trigger firing after the flag already activated is no longer persisted/reported, so the recording "entitlement" is lost if the flag reverts mid-session (plus a minor start-reason telemetry drop). **Convergent** — `chatgpt-codex-connector` flagged the same issue as P1 independently. Deferred: the fix is a genuine design decision (split matchers vs per-trigger guard vs accept the trade-off) in core session-recording logic — for the author to decide.

### Convergence

- `recording-strategies.ts:177` — QA Swarm router (MEDIUM) **+** chatgpt-codex-connector (P1), independently. Highest-confidence finding.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | R1: traced all three `hasPendingTriggers` consumers + both status-function paths, ran the linked-flag suite (10/10), confirmed 2/3 new tests fail against pre-fix source. R2: verified the test refactor preserves assertions, lint + tsc clean. Danger: MEDIUM→none, confidence HIGH. Delegated nothing. |

<details>
<summary>Previous rounds (1)</summary>

round 1 @ 1c8735f — 💬 APPROVE WITH NITS: MEDIUM `_activateTrigger` edge-case (deferred, convergent with codex P1) + NIT that one of three new tests doesn't pin the fix (resolved).

</details>

---

PR #4214, opened 2026-07-22
fix(browser): respect event/action trigger in survey eligibility

### LINKED ISSUE #2501: Event filtering does not work with posthog surveys
## Bug description
  
The survey display conditions do not respect the "User sends events" filter. When its set to X event, and from application using posthog-js SDK, the method "canRenderSurveyAsync" is returning visible property to true always, even when user has not sent X event yet.
  
  ## How to reproduce
  
  1. Create a survey and set Event filters to an event
  2. Send any event using SDK
  3. Call "canRenderSurveyAsync" method
  4. {visible: true} returned, regardless of the event that was being sent
  
  ## Related sub-libraries
  
  - [ ] All of them
  - [x] posthog-js (web)
  - [ ] posthog-node
  - [ ] posthog-react-native
  - [x] @posthog/react
  - [ ] @posthog/ai
  - [ ] @posthog/nextjs-config
  
  ## Additional context
Posthog host: https://eu.i.posthog.com
  
  #### _Thank you_ for your bug report – we love squashing them!

felipeatom:
## Problem
Surveys configured with a "User sends events" filter are reported as renderable by `canRenderSurvey` / `canRenderSurveyAsync` even before the trigger event has fired, returning `{ visible: true }` regardless. (Fixes #2501)

## Root cause
`canRenderSurvey` / `canRenderSurveyAsync` only run `SurveyManager.checkSurveyEligibility`, which validates running state, survey type, linked/targeting/internal feature flags, the wait period and the "already seen" check — but **not** the survey's event/action activation trigger. That trigger is evaluated separately, only inside the internal display loop (`_shouldDisplaySurvey` → `_hasActionOrEventTriggeredSurvey`). So the programmatic render check ignores it.

## Change
Include the existing `_hasActionOrEventTriggeredSurvey` check in `checkSurveyEligibility`, so the programmatic render check behaves like the display loop. Surveys without an event/action trigger are unaffected — the helper already returns `true` for them. The (now redundant) call in `_shouldDisplaySurvey` is harmless and left as-is to keep that predicate self-documenting.

Scope is intentionally limited to the event/action trigger, matching the issue. URL/device/selector conditions (`_isSurveyConditionMatched`) are a separate concern and left unchanged.

## Testing
Added unit tests in `surveys.test.ts`:
- not eligible until the trigger event fires,
- eligible once it fires,
- unaffected when there is no event/action trigger.

Local checks (`packages/browser`):
- `pnpm test:unit:surveys` → 461 passing
- `pnpm eslint` on changed files → clean
- `pnpm prettier --check` → clean
- `pnpm typecheck` (src) → clean

Fixes #2501

### DIFF
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -502,6 +502,38 @@ describe('SurveyManager', () => {
         })
     })
 
+    describe('respects the event trigger condition (issue #2501)', () => {
+        const EVENT_GATED_SURVEY_ID = 'event-gated-survey'
+
+        const makeEventGatedSurvey = (): Survey => ({
+            ...mockSurveys[0],
+            id: EVENT_GATED_SURVEY_ID,
+            conditions: { events: { values: [{ name: 'survey_trigger_event' }] } },
+        })
+
+        const setActivatedSurveys = (surveyIds: string[]): void => {
+            ;(mockPostHog.surveys as any)._surveyEventReceiver = { getSurveys: () => surveyIds }
+        }
+
+        it('is not eligible until the trigger event has fired', () => {
+            setActivatedSurveys([])
+            const result = surveyManager.checkSurveyEligibility(makeEventGatedSurvey())
+            expect(result.eligible).toBe(false)
+        })
+
+        it('becomes eligible once the trigger event has fired', () => {
+            setActivatedSurveys([EVENT_GATED_SURVEY_ID])
+            const result = surveyManager.checkSurveyEligibility(makeEventGatedSurvey())
+            expect(result.eligible).toBe(true)
+        })
+
+        it('is unaffected for surveys without an event/action trigger', () => {
+            setActivatedSurveys([])
+            const result = surveyManager.checkSurveyEligibility({ ...mockSurveys[0], conditions: null })
+            expect(result.eligible).toBe(true)
+        })
+    })
+
     test('callSurveysAndEvaluateDisplayLogic should handle popup surveys correctly', () => {
         const handlePopoverSurveyMock = jest
             .spyOn(surveyManager as any, 'handlePopoverSurvey')
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -515,21 +515,29 @@ describe('SurveyManager', () => {
             ;(mockPostHog.surveys as any)._surveyEventReceiver = { getSurveys: () => surveyIds }
         }
 
-        it('is not eligible until the trigger event has fired', () => {
+        it('is not renderable until the trigger event has fired', () => {
             setActivatedSurveys([])
-            const result = surveyManager.checkSurveyEligibility(makeEventGatedSurvey())
+            const result = surveyManager.checkSurveyRenderability(makeEventGatedSurvey())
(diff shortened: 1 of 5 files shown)


packages/browser/src/extensions/surveys.tsx
```diff
@@ -656,6 +656,12 @@ export class SurveyManager {
             return eligibility
         }
 
+        if (!this._hasActionOrEventTriggeredSurvey(survey)) {
```
turnipdabeets: `checkSurveyEligibility` is shared with the display path: `displaySurvey` goes through `canRenderSurvey` (and `ignoreConditions` defaults to `false`), so this check also gates explicit `posthog.displaySurvey('id')` calls. The trigger state only lives in memory (a reload clears it) and server-side events never set it, so those calls can now silently show nothing — the "not eligible" warning only prints in debug mode. Could we keep this method as it was and wrap it instead, so only the read-only checks pick up the trigger?

```ts
public checkSurveyRenderability(survey: Survey): { eligible: boolean; reason?: string } {
    const eligibility = this.checkSurveyEligibility(survey)
    if (eligibility.eligible && !this._hasActionOrEventTriggeredSurvey(survey)) {
        return { eligible: false, reason: 'Survey event/action trigger has not been fired yet' }
    }
    return eligibility
}
```

`canRenderSurvey`/`canRenderSurveyAsync` switch to the wrapper, and `displaySurvey` calls the eligibility check directly instead of going through `canRenderSurvey` — the check itself and everything else can stay as-is.

packages/browser/src/__tests__/extensions/surveys.test.ts
```diff
@@ -502,6 +502,38 @@ describe('SurveyManager', () => {
...
+            ...mockSurveys[0],
+            id: EVENT_GATED_SURVEY_ID,
+            conditions: { events: { values: [{ name: 'survey_trigger_event' }] } },
+        })
+
+        const setActivatedSurveys = (surveyIds: string[]): void => {
+            ;(mockPostHog.surveys as any)._surveyEventReceiver = { getSurveys: () => surveyIds }
+        }
+
+        it('is not eligible until the trigger event has fired', () => {
```
turnipdabeets: Could we also add one case asserting `checkSurveyEligibility` still returns eligible for an event-gated survey whose trigger hasn't fired? That's the regression guard that `displaySurvey`'s behavior didn't change.

turnipdabeets (commented):
Thanks @felipeatom this looks very good. I left one suggestion to prevent a regression.

felipeatom:
Thanks @turnipdabeets, this is a great catch and it's exactly the concern I had about `checkSurveyEligibility` being shared with the explicit display path — gating it on the in-memory trigger would make `displaySurvey('id')` silently render nothing, which is the opposite of what an explicit call should do.

I went with your wrapper approach:

- `checkSurveyEligibility` is back to its original behavior (no trigger check).
- New `checkSurveyRenderability` wraps it and adds the event/action trigger gate.
- `canRenderSurvey` / `canRenderSurveyAsync` now use `checkSurveyRenderability`.
- `displaySurvey` calls the eligibility check directly, so explicit displays are never suppressed by the trigger state.

I also added the regression test you suggested, asserting `checkSurveyEligibility` stays eligible for an event-gated survey whose trigger hasn't fired. Survey suites are green (142 tests). Let me know if you'd like anything else adjusted.

turnipdabeets (approved):

PR #4218, opened 2026-07-22
fix(browser): scope request timestamps by endpoint

marandaneto:
## Problem

The Browser SDK adds a timestamp under `_` to every request. Capture POST requests support numeric `sent_at` query timestamps, `/flags/` supports an ISO `sent_at` body field, and only freshness-sensitive GET requests need `_` for cache busting. Unrelated requests should not receive timestamps.

## Changes

- Add one mutually exclusive `timestampMode` request option for body or query timestamps.
- Send numeric `sent_at` query parameters for analytics and session recording capture POSTs without changing their existing request bodies.
- Send ISO 8601 `sent_at` in the existing `/flags/` request body.
- Keep `_` for freshness-sensitive GET requests for surveys, early access features, web experiments, product tours, and Conversations.
- Let remote config and unrelated requests use their declared cache behavior without a timestamp parameter.
- Preserve capture body shapes through batching, `sendBeacon` splitting, and compression fallback.
- Cover endpoint-specific query and body behavior, call sites, and functional requests with tests.

Depends on PostHog/posthog#72842. The backend compatibility change must be deployed before this SDK change is released. Older SDKs remain supported through the backend's query fallback.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent after auditing capture, session recording, feature flag, and dynamic GET backend handling. We kept capture timestamps in query parameters to preserve all existing analytics and recording body formats, while `/flags/` uses the new backend-compatible ISO body field. Read-only Pi reviewers checked batching, compression fallback, `sendBeacon` splitting, and endpoint coverage.


marandaneto:
send sent_at as body property instead of the _ query param for endpoints that actually use it

marandaneto:
| Method | Endpoint | Before | After | Reason |
|---|---|---|---|---|
| `POST` | `/e/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | Capture reads `sent_at` as Unix milliseconds and uses it for clock-skew correction. The event body remains unchanged. |
| `POST` | `/i/v0/e/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | Same shared capture handler as `/e/`. Using the explicit parameter name clarifies its timestamp semantics. |
| `POST` | `/batch/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | The capture handler uses the query value when the existing batch body has no valid top-level `sent_at`. The body format remains unchanged. |
| `POST` | `/capture/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | The legacy capture route actively reads and uses numeric query `sent_at`. |
| `POST` | `/track/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | The legacy capture route actively reads and uses numeric query `sent_at`. |
| `POST` | `/engage/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | The handler converts the payload into an identify event and uses query `sent_at` during timestamp processing. |
| `POST` | `/s/` | Numeric `_` query parameter | Numeric `sent_at` query parameter | Session recording already reads the numeric query timestamp. Its existing body format remains unchanged. |
| `POST` | `/flags/` | Numeric `_` query parameter | ISO 8601 `sent_at` property in the existing JSON body | The flags backend now accepts the body timestamp and uses it for cookieless processing. Body values take precedence over the legacy query fallback. |
| `GET` | `/array/{token}/config` | Numeric `_` query parameter | No timestamp parameter | This endpoint is intentionally cacheable with `Cache-Control: public, max-age=300`. Removing `_` enables the documented five-minute cache and canonical CDN purging behavior. |
| `POST` | `/i/v1/logs` | Numeric `_` query parameter | No timestamp parameter | The backend only uses the relevant token and compression metadata. `_` was ignored and POST requests do not need cache busting. |
| `POST` | `/i/v1/metrics` | Numeric `_` query parameter | No timestamp parameter | The backend does not assign meaning to `_`, and POST requests do not need cache busting. |
| `POST` | `/api/conversations/v1/widget/message` | Numeric `_` query parameter | No timestamp parameter | The endpoint consumes its JSON body and authentication header. `_` was unused. |
| `POST` | `/api/conversations/v1/widget/messages/{ticket_id}/read` | Numeric `_` query parameter | No timestamp parameter | The endpoint consumes its JSON body and path identifier. `_` was unused. |
| `POST` | `/api/conversations/v1/widget/restore` | Numeric `_` query parameter | No timestamp parameter | The endpoint consumes the restore token from its JSON body. `_` was unused. |
| `POST` | `/api/conversations/v1/widget/restore/request` | Numeric `_` query parameter | No timestamp parameter | The endpoint consumes the email and request URL from its JSON body. `_` was unused. |

marandaneto:
next step is to change the body shape for ingest endpoints so we can kill the sent_at query param in favor of sent_at body property similar to /batch

ioannisj (approved):

PR #651, opened 2023-05-29
feat: remove broken capture using img support

pauldambra:
## Changes

You can configure posthog-js to send capture requests by adding img tags to the DOM

It doesn't work... we've wanted to remove it for forever (see #155)

Technically removing this is a breaking change but since this doesn't really work. We should be safe to remove it

PR #4219, opened 2026-07-22
feat(react-native): Expo plugin dotenvFile option for upload credentials

ablaszkiewicz:
# Why

When wizard instruments react native app, it puts `.env` file in the react-native root directory. RN build is able to read this file and read CLI related variables just fine. This covers hermes source maps

However if we want to include native sources, we have to somehow instrument this file into native technologies build processes so they can read CLI related vars and upload symbol sets. We need this for dsyms and proguard maps

Recently I was working on ios and android raw support and in both of these technologies we can pass this file somehow. But here in react-native we don't have direct control over "native" config files so we have to find a way around this

# What

Adds a `dotenvFile` prop to the Expo config plugin. The path reaches every upload hook as `POSTHOG_CLI_DOTENV_FILE` (posthog-cli >= 0.8.4): iOS via an Xcode build setting (bundle + dSYM phases), Android via a `posthog.dotenvFile` entry in `gradle.properties` read by `posthog.gradle`.

For iOS it works out of the box we just need to instrument env variable. For android it wasn't so obvious because build script somehow strips env variables so I had to do this - https://github.com/PostHog/posthog-android/pull/651

Tested this locally and works great!

| Expo app - iOS - native crash |
|--------|
| CleanShot 2026-07-23 at 11 12 20 | 

| Expo app - android - native crash |
|--------|
| CleanShot 2026-07-23 at 11 12 52 | 


packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -314,6 +316,81 @@ export function disableUserScriptSandboxing(xcodeProject: any): void {
...
+// both gradle hooks (the SDK's posthog.gradle hermes upload and the
+// com.posthog.android mapping upload). Added when the prop is set, removed
+// when it isn't.
+export function updateDotenvFileGradleProperties(
+  properties: GradlePropertiesItem[],
+  dotenvFile?: string
+): GradlePropertiesItem[] {
+  const rest = properties.filter((item) => !(item.type === 'property' && item.key === POSTHOG_DOTENV_GRADLE_PROPERTY))
+  if (!dotenvFile) {
+    return rest
```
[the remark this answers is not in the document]
ablaszkiewicz: This is intentional — the plugin treats posthog.dotenvFile as a managed entry, same as the iOS build setting: set the prop and the entry is written, remove the prop and it's cleaned up. Without the filter, removing dotenvFile from app.json would leave a stale credential-file path exported to the upload hooks on every incremental prebuild.

The scenario described here is hand-editing android/gradle.properties, which is a generated file under CNG — any edit there is lost on prebuild --clean regardless of this code. Bare RN projects (where setting the property directly is the supported path) don't run the config plugin at all, so they're unaffected. If you're on Expo and want this property, the supported way is the dotenvFile prop in app.json.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -364,6 +441,23 @@ type PostHogPluginProps = {
...
   skipOnConflict?: boolean
+
+  /**
+   * Path to a dotenv file with POSTHOG_CLI_* credentials (API key, project id,
+   * optional host), relative to the project root — or absolute.
+   *
+   * The path reaches every upload hook as POSTHOG_CLI_DOTENV_FILE: on iOS as a
+   * build setting (Xcode exports it to the bundle and dSYM script phases), on
+   * Android as a `posthog.dotenvFile` entry in android/gradle.properties read
+   * by both gradle hooks. Process env always wins inside the CLI; a missing
```
ioannisj: This says "both gradle hooks" but only the `posthog.gradle` hermes upload actually gets `POSTHOG_CLI_DOTENV_FILE` in this PR. The R8 mapping upload runs inside the `com.posthog.android` plugin, which only reads `posthog.dotenvFile` once PostHog/posthog-android#651 lands, and that's still open. We pin the classpath to `1.2.0` (`POSTHOG_ANDROID_GRADLE_PLUGIN_VERSION`, line 10), and even the latest release `1.3.0` predates #651.

So as it ships, someone with creds only in the dotenv file gets working sourcemaps but the mapping upload runs without credentials?

Also, this needs #651 to land, cut a plugin release, bump the pinned version to it. Or, we can soften the changeset and this doc to "sourcemap upload today, mapping upload once the plugin ships it" and bump in a follow-up?

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -314,6 +316,81 @@ export function disableUserScriptSandboxing(xcodeProject: any): void {
...
+// setting feeds both the bundle-phase hermes upload and the dSYM upload phase;
+// posthog-cli (>= 0.8.4) reads POSTHOG_CLI_DOTENV_FILE itself, no script
+// changes needed. Removing the prop removes the setting again so changes take
+// effect without a clean prebuild.
+export function applyDotenvFileBuildSetting(xcodeProject: any, dotenvFile?: string): void {
+  const configurations = xcodeProject.pbxXCBuildConfigurationSection()
+  for (const key in configurations) {
+    const configuration = configurations[key]
+    if (configuration && configuration.buildSettings) {
+      if (dotenvFile) {
```
ioannisj, line 358: Should we trim here for parity with the gradle side? That guards with `!posthogDotenvFileProperty.toString().trim().isEmpty()` (`posthog.gradle:106`), but this only checks truthiness, so a whitespace-only `dotenvFile` (easy to get from a templated `app.config` value) writes `POSTHOG_CLI_DOTENV_FILE = "$(SRCROOT)/../   "` on iOS.

ioannisj (approved):
LG, couple of comments/questions

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: I'll wait for android release and bump it here and then merge

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: Addressed

PR #4222, opened 2026-07-22
feat(flags): add defaultValue option to isFeatureEnabled

turnipdabeets:
## Problem

`isFeatureEnabled` returns `undefined` when a flag has no value (flags not loaded yet, or no flag with that key), and callers have no way to supply a fallback. Android and Unity already expose a `defaultValue` parameter for this, and the sdk-specs server-side canonical signature includes `defaultValue` in its options — the JS SDKs did not.

## Changes

Adds an optional default value to `isFeatureEnabled`, returned whenever the flag has no value — flags not loaded, failed to load, or no flag with that key. When a flag *has* a value, it always wins over the default. Passing a default narrows the return type to `boolean` via TS overloads; without it, behavior and the `boolean | undefined` type are byte-for-byte unchanged.

- **`@posthog/core`** (inherited by `posthog-react-native` and `posthog-js-lite`): new `IsFeatureEnabledOptions` with `defaultValue?: boolean`. The implementation now calls `_getFeatureFlagResult` directly, using the legacy `missingFlagBehavior` only when no default is passed, so bare-call semantics and `$feature_flag_called` events are unchanged.
- **`posthog-js` (browser)**: same feature in its own implementation, as `defaultValue`, matching core/RN/js-lite (one option name across all JS SDKs). Applies to the `fresh`-miss, not-loaded, and missing-key paths.
- **`@posthog/types`**: `IsFeatureEnabledOptions` + interface overloads.
- **`posthog-react-native`**: override mirrors the overloads; JSDoc/docs updated.

Semantics match Android (`getFeatureFlagResult(key)?.value ?: defaultValue`) and Unity (`if (!flag.Value.HasValue) return defaultValue`): the default applies to *any* missing value, including a key absent from loaded flags. iOS and Flutter hard-code the equivalent of `defaultValue: false`.

Note for spec follow-up: the sdk-specs client-side canonical signature doesn't include `defaultValue` yet; it should be added once this ships.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- Written with Claude Code (Fable 5); Anna Garcia directed scope and decisions throughout.
- Key decisions: `defaultValue` lives in the existing options object (the positional slot is taken by `options` already shipped in stable; matches the sdk-specs server-side canonical shape); the default applies to any missing value, not just the not-loaded state (chosen for Android/Unity parity after comparing their implementations); bare calls keep returning `undefined` per the maintainer discussion on #3930.
- Verified against `sdk-specs/openspec/specs/is-feature-enabled/spec.md` and the Android, Unity, iOS, and Flutter implementations. Node/Python/PHP/Ruby are unaffected (separate implementations; Node's method is deprecated in favor of `evaluateFlags`).

### DIFF
--- a/packages/browser/src/posthog-core.ts
+++ b/packages/browser/src/posthog-core.ts
@@ -62,6 +62,7 @@ import {
     FeatureFlagsCallback,
     FeatureFlagOptions,
     FeatureFlagResult,
+    IsFeatureEnabledOptions,
     JsonType,
     OverrideConfig,
     PostHogConfig,
@@ -2008,7 +2009,8 @@ export class PostHog implements PostHogInterface {
      * Checks if a feature flag is enabled for the current user.
      *
      * @remarks
-     * Returns true if the flag is enabled, false if disabled, or undefined if not found.
+     * Returns true if the flag is enabled, false if disabled, or undefined if not found
+     * (unless `default_value` is given, which is returned instead of undefined).
      * This is a convenience method that treats any truthy value as enabled.
      *
      * {@label Feature flags}
@@ -2032,11 +2034,13 @@ export class PostHog implements PostHogInterface {
      * @public
      *
      * @param {string} key Key of the feature flag.
-     * @param {FeatureFlagOptions} [options] Optional lookup settings. If `{ send_event: false }`, we won't send a `$feature_flag_called` event to PostHog. If `{ fresh: true }`, we won't return cached values from localStorage - only values loaded from the server.
-     * @returns {boolean | undefined} Whether the feature flag is enabled, or undefined if the flag is unavailable.
+     * @param {IsFeatureEnabledOptions} [options] Optional lookup settings. If `{ send_event: false }`, we won't send a `$feature_flag_called` event to PostHog. If `{ fresh: true }`, we won't return cached values from localStorage - only values loaded from the server. If `{ default_value: false }`, we return that value instead of undefined when the flag has no value.
+     * @returns {boolean | undefined} Whether the feature flag is enabled; when the flag has no value, default_value if given, otherwise undefined.
      */
-    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean | undefined {
-        return this.featureFlags?.isFeatureEnabled(key, options)
+    isFeatureEnabled(key: string, options: IsFeatureEnabledOptions & { default_value: boolean }): boolean
+    isFeatureEnabled(key: string, options?: IsFeatureEnabledOptions): boolean | undefined
+    isFeatureEnabled(key: string, options?: IsFeatureEnabledOptions): boolean | undefined {
+        return this.featureFlags?.isFeatureEnabled(key, options) ?? options?.default_value
     }
 
     /**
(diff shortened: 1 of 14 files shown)


packages/browser/src/posthog-core.ts
```diff
@@ -2032,11 +2034,13 @@ export class PostHog implements PostHogInterface {
...
-     * @param {FeatureFlagOptions} [options] Optional lookup settings. If `{ send_event: false }`, we won't send a `$feature_flag_called` event to PostHog. If `{ fresh: true }`, we won't return cached values from localStorage - only values loaded from the server.
-     * @returns {boolean | undefined} Whether the feature flag is enabled, or undefined if the flag is unavailable.
+     * @param {IsFeatureEnabledOptions} [options] Optional lookup settings. If `{ send_event: false }`, we won't send a `$feature_flag_called` event to PostHog. If `{ fresh: true }`, we won't return cached values from localStorage - only values loaded from the server. If `{ default_value: false }`, we return that value instead of undefined when the flag has no value.
+     * @returns {boolean | undefined} Whether the feature flag is enabled; when the flag has no value, default_value if given, otherwise undefined.
      */
-    isFeatureEnabled(key: string, options?: FeatureFlagOptions): boolean | undefined {
-        return this.featureFlags?.isFeatureEnabled(key, options)
+    isFeatureEnabled(key: string, options: IsFeatureEnabledOptions & { default_value: boolean }): boolean
```
dustinbyrne: can we avoid the mixed camelCase and snake_case?

turnipdabeets: Went with camelCase in [a5960c6](https://github.com/PostHog/posthog-js/pull/4222/commits/a5960c6eb49e74ef3e072d4a10558ecb5dc206b2)

packages/core/src/posthog-core.ts
```diff
@@ -1152,12 +1153,15 @@ export abstract class PostHogCore extends PostHogCoreStateless {
...
-    if (response === undefined) {
-      return undefined
-    }
-    return !!response
+  isFeatureEnabled(key: string, options: IsFeatureEnabledOptions & { defaultValue: boolean }): boolean
+  isFeatureEnabled(key: string, options?: IsFeatureEnabledOptions): boolean | undefined
+  isFeatureEnabled(key: string, options?: IsFeatureEnabledOptions): boolean | undefined {
+    const result = this._getFeatureFlagResult(key, {
+      sendEvent: options?.sendEvent,
+      missingFlagBehavior: options?.defaultValue === undefined ? 'getFeatureFlag' : undefined,
```
dustinbyrne: assuming this was needed to make tests pass - this was a confusing addition for backwards compatibility

dustinbyrne (approved):
love the type narrowing here

packages/core/src/posthog-core.ts
turnipdabeets: Yea it is confusing, and not just for tests so refactored it a bit

PR #4223, opened 2026-07-22
fix(replay): terminate the canvas encode worker when recording stops

turnipdabeets:
## Problem

Stopping session recording with canvas capture enabled cancels the rAF capture loop but never terminates the dedicated encode worker. Dedicated workers are not cleaned up by becoming unreachable, so every stop/start cycle leaks a worker thread along with its capture-resolution `OffscreenCanvas` (~8MB of RGBA pixel buffer at 1080p) and its frame-fingerprint maps. Repeated stop/start — manual `stopSessionRecording()`/`startSessionRecording()`, or any integration that restarts recording — accumulates zombie worker threads and pinned memory for the page's lifetime.

The worker-load *error* path already terminates the worker; the normal teardown path in `CanvasManager.resetObservers` did not.

## Changes

Call `worker.terminate?.()` in the FPS observer's `resetObservers`, alongside the existing rAF cancellation and context-observer reset.

Safety notes for review:

- `Worker.terminate()` is idempotent per spec, so overlap with the `onerror`-path terminate is harmless.
- A snapshot that is mid-`createImageBitmap` at teardown resumes and posts to a terminated worker — a spec-defined silent no-op, and the call site is inside the existing per-canvas `try`.
- The mutation-observer variant of `resetObservers` is untouched (it owns no worker).

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [ ] Tests for new code — none added: exercising real worker-thread termination isn't reachable from the existing worker unit-test harness (it stubs `self`/`postMessage`, no thread exists). Existing worker suite passes (10/10); `@posthog/rrweb` builds clean.
- [x] Ran changeset (patch: `posthog-js`, `@posthog/rrweb`)

## 🤖 Agent context

Found by Claude Code during a performance review of the canvas capture path: tracing the recorder stop/start lifecycle showed `resetObservers` tearing down the rAF loop and context observer but leaving the worker thread alive, while the `onerror` path terminated it. The fix mirrors the error path's cleanup in normal teardown. Verified by code trace against the HTML spec (dedicated worker lifetime, terminate idempotency, post-to-terminated semantics) plus the existing worker unit suite and a clean `@posthog/rrweb` build — no live-browser measurement of the reclaimed memory was performed.

### DIFF
--- a/packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
+++ b/packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
@@ -371,6 +371,10 @@ export class CanvasManager {
     this.resetObservers = () => {
       canvasContextReset();
       cancelAnimationFrame(rafId);
+      // a dedicated worker is not terminated by becoming unreachable; without
+      // this, every recorder restart leaks a worker thread plus its
+      // capture-resolution OffscreenCanvas
+      worker.terminate?.();
     };
   }
 
--- a/.changeset/terminate-canvas-worker.md
+++ b/.changeset/terminate-canvas-worker.md
@@ -1,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/rrweb': patch
+---
+
+Terminate the canvas encode worker when session recording stops. Previously stopping a recording with canvas capture enabled cancelled the capture loop but left the dedicated worker running; dedicated workers are not cleaned up by becoming unreachable, so every stop/start cycle leaked a worker thread along with its capture-resolution OffscreenCanvas (~8MB of pixel buffer at 1080p) and frame-fingerprint state.
--- a/.changeset/terminate-canvas-worker.md
+++ b/.changeset/terminate-canvas-worker.md
@@ -1,5 +1,4 @@
 ---
-'posthog-js': patch
 '@posthog/rrweb': patch
 ---
 
--- a/packages/rrweb/rrweb/test/record/canvas-manager.test.ts
+++ b/packages/rrweb/rrweb/test/record/canvas-manager.test.ts
@@ -152,6 +152,25 @@ describe('CanvasManager FPS observer', () => {
     expect(rafCallbacks.size).toBe(0);
   });
 
+  it('should terminate the worker on normal teardown, exactly once', () => {
+    const win = {
+      document: { querySelectorAll: vi.fn(() => []) },
+      OffscreenCanvas: class {},
+    };
+
+    const manager = createCanvasManager(win);
+    manager.acquire();
+    const worker = workerControl.instances[0];
+    expect(worker).toBeDefined();
+
+    manager.reset();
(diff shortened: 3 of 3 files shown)


dustinbyrne (approved):
this looks reasonable - is it possible to test?

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
```diff
@@ -371,6 +371,10 @@ export class CanvasManager {
     this.resetObservers = () => {
       canvasContextReset();
       cancelAnimationFrame(rafId);
+      // a dedicated worker is not terminated by becoming unreachable; without
+      // this, every recorder restart leaks a worker thread plus its
+      // capture-resolution OffscreenCanvas
+      worker.terminate?.();
```
dustinbyrne, line 377: [P1] Normal worker teardown is the whole fix, but no test exercises this `terminate()` call.

The changeset promises that stopping recording terminates the canvas worker. The existing assertion in `canvas-manager.test.ts:133-152` only covers the separate `worker.onerror` path, while the reference-count tests replace `resetObservers` with a spy.

For example:

```ts
const manager = createCanvasManager(win)
manager.acquire()
const worker = workerControl.instances[0]
manager.reset()
manager.reset()

expect(worker.terminate).toHaveBeenCalledTimes(1)
expect(rafCallbacks.size).toBe(0)
```

Actual: removing this line leaves all existing assertions unchanged.

Expected: normal final reset is covered, including idempotent repeated cleanup.

Without that coverage, the worker-thread and retained-canvas leak can be silently reintroduced.

Consider adding this focused case using the existing worker and rAF mocks.

A regression test should acquire an FPS manager, reset it twice, and assert exactly one termination and no pending canvas rAF.

dustinbyrne (commented):
Agent-led review, human-reviewed before posting.

The change looks reasonable overall. I have one focused test-coverage suggestion:

packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts
turnipdabeets: Added in 6207cd6b8 — `should terminate the worker on normal teardown, exactly once`: `acquire()` → `reset()` → `reset()`, asserting `worker.terminate` called exactly once (the second reset exercises the `torndown` latch) and `rafCallbacks` empty, using the existing `workerControl`/`rafCallbacks` fixtures. Suite is 14/14.

Same push also narrows the changeset to `@posthog/rrweb` only — the hygiene bot was right that `posthog-js` gets its patch bump through the internal-dependency update, so declaring it directly was redundant.

turnipdabeets: added in [6207cd6](https://github.com/PostHog/posthog-js/pull/4223/commits/6207cd6b86c206a5b87a723b1dd36b72b7a1d7b7)

PR #4226, opened 2026-07-22
feat(rrweb): opt-in event loop yielding during seek fast-forward rebuilds

ksvat:
## Problem

Seeking in a long recording can freeze the tab hard enough that Chrome offers to kill the page. The cause is old and documented in the player code: a seek fast-forwards through every event between the last full snapshot and the target in one synchronous main-thread pass (`applyEventsSynchronously`). Scrubbing is the worst case — every jump pays a full rebuild, back to back.

We reproduced this on a ~6 hour production recording from a customer report, driving the real player with Playwright and watching `longtask` entries plus a CDP CPU profile:

- single seeks blocked the main thread for 0.4–3.7s each
- a burst of 6 quick seekbar jumps blocked it for ~4.8s straight — the clicks themselves queued for 4.8s waiting for the main thread, which is exactly the input starvation that triggers Chrome's "Page Unresponsive" dialog
- the largest share of non-idle CPU across the seek session landed in the rrweb chunk (~6s self time, plus most of ~7s of GC from rebuild allocation churn) — we confirmed which chunk is rrweb by grepping the live page scripts for `applyEventsSynchronously` / `rebuildFullSnapshot`

(There's a second contributor — the app re-rendering the inspector around seeks — that's an app-side follow-up, not addressed here.)

## Changes

New opt-in `seekYieldBudgetMs` player config, default `0` (today's fully synchronous behavior — callers that read the DOM right after `pause(t)` are unaffected). When set:

- the fast-forward applies events in time-budgeted chunks and yields to the event loop between chunks (`src/replay/fast-forward.ts`); seeks that finish within one budget stay fully synchronous
- a newer seek cancels a rebuild still in flight, so rapid scrubbing only pays for the last jump
- a superseded rebuild sends the new `RESET_LAST_PLAYED` machine event so the next seek does a full rebuild — the machine's skip-already-applied optimization can't trust a partially applied history. The scrub-supersession e2e test caught this: without the reset, a cancelled rebuild silently dropped a DOM node

Plumbing: the machine asset `applyEventsSynchronously(events)` becomes `applyEvents(events, onApplied)` — the applier emits `Flush` and the timer starts via the completion callback, with the paused/live completion cases netting out to the same timer state as the synchronous path.

Tests: 5 unit tests for the chunk scheduler, a machine test for `RESET_LAST_PLAYED`, and 4 new puppeteer tests (chunked pause end state, DOM equivalence vs a synchronous seek, scrub supersession, play after a chunked seek). The 46 pre-existing replayer e2e tests pass unchanged, which is the regression evidence for the default path.

The PostHog app will opt in separately (`seekYieldBudgetMs` in the player config, behind a flag).

### Review follow-ups (second commit)

A self-review pass found and fixed three gaps:

- **`addEvent` during a chunked rebuild lost the event from the playback schedule.** The machine only put future events on the timer `if (timer.isActive())`, and with chunking the timer doesn't start until the rebuild completes. Future events are now scheduled in the completion callback, from the live events array, so mid-rebuild arrivals play. Machine test included.
- **`Finish` could fire against a half-built frame.** The ~50ms finish timeout armed by casting the last event could beat a chunked rebuild's completion. It now re-arms while its own rebuild is still applying and drops itself when a newer seek supersedes it. E2e test asserts `finish` fires after the final `flush`.
- **`seekYieldBudgetMs` is optional** in `playerConfig` — a required key was a type-level break for consumers constructing a full config.

Plus new coverage: destroy mid-rebuild, `TO_LIVE` during a rebuild, and machine guards for the delta optimization and `RESET_LAST_PLAYED`.

A third commit closes the last unbounded block on the chunked paused path: when a chunked rebuild completes while paused, the completion no longer schedules every remaining future event onto the timer just to discard it (on a long recording seeked near the start, that was an O(all-future-events) synchronous closure-building burst). The paused branch nets the timer out directly; resuming recomputes the schedule via PLAY. The fully synchronous path keeps the legacy behavior — there the PAUSE only arrives after the PLAY transition completes, so the discard is unknowable at scheduling time.

Verifying those tests also surfaced a **pre-existing** seek boundary bug, unrelated to this PR's changes: pausing exactly at a FullSnapshot's timestamp classified the snapshot as a "future" event and skipped the rebuild. That was fixed separately in #4278 (merged, and this branch carries it); the regression test is now a passing one. The fullscreen scrub test still pins its `beforeOffset` strictly after the snapshot so it stays a scrub test rather than a boundary test.

### Review fixes (fourth round)

Review found the two-pass `play()` split had regressed two behaviors of the original single if/else, both on the default synchronous path, plus one doc/behavior mismatch:

- **Boundary snapshot double-apply**: a FullSnapshot/Meta at exactly the seek timestamp was applied synchronously *and* scheduled on the timer at delay 0, rebuilding the frame twice. The completion loop now mirrors the sync pass's boundary guard.
- **MouseMove batch spanning the baseline replayed**: the completion loop dropped the `event === lastPlayedEvent` skip, so an already-played MouseMove batch whose first position (negative offset) precedes the baseline was rescheduled. Skip restored under the same delta-optimization guard.
- **`getCurrentTime()` mid-rebuild**: `Timer.clear()` deliberately preserves the accumulated offset (plain pause reads it), so a seek exposed the previous play's offset until the rebuild completed. The play action now zeroes it next to `clear()` — behavior-identical on the synchronous path, and makes the `seekYieldBudgetMs` doc's claim true.

Each has a machine regression test verified red before its fix.

### Review fixes (fifth round)

A changes-requested review found three supersession gaps around going live, all now fixed with tests red before the fix:

- **`startLive()` stranded the virtual DOM**: a cancelled rebuild never emits `Flush`, so once a mutation chunk had engaged `usingVirtualDom`, every live mutation wrote to the abandoned virtual tree and never reached the iframe. `startLive()` now commits the partially applied frame via `Flush` before `TO_LIVE`.
- **`TO_LIVE` ignored from `playing`**: cancelling a `play(t)` rebuild left the machine `playing` with a timer that never started, and the machine only accepts `TO_LIVE` from `paused`. `startLive()` now routes through `PAUSE` in the cancellation case (the synchronous path is untouched).
- **Reentrant cancellation**: a cast's plugins/`event-cast` listeners can synchronously start a newer seek; the applier now checks cancellation after every cast and before completing/rescheduling, so a stale pass can't keep casting or call its old `onComplete`.

Also: `applyEvents` discards a superseded pass's undelivered flush buffers (stale scroll positions and friends) at the start of each rebuild, and the changeset now versions only `posthog-js` — rrweb packages release through it and are never versioned directly.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/rrweb (minor; @posthog/rrweb-replay and posthog-js are auto-bumped patch as workspace dependents)

## Checklist

- [x] Tests for new code
- [x] Ran `pnpm changeset` to generate a changeset file (authored manually: `.changeset/rrweb-seek-yield-budget.md`)

## 🤖 Agent context

Claude Code session, support-driven: a customer report of the player freezing on long recordings. The agent traced the freeze to the synchronous seek fast-forward, implemented the chunked applier, and verified the diagnosis against the reported recording by driving the production player with Playwright (playwriter) + a CDP profiler — numbers in the Problem section.

Decisions: opt-in config instead of default-on (headless replayer and rrweb-player read the DOM synchronously after seeks); time-budget chunking instead of event-count (short seeks stay byte-identical to today); cancellation via a generation counter. The `RESET_LAST_PLAYED` event wasn't in the original design — the scrub-supersession e2e test failed and exposed that `lastPlayedEvent` can't be trusted after a cancelled rebuild.


ksvat:
* **#4226** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4226?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4226?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

packages/rrweb/rrweb/src/replay/index.ts
```diff
@@ -771,32 +783,54 @@ export class Replayer {
...
+          // a PAUSE arrived while the rebuild was yielding. Skip onApplied:
+          // it would schedule every future event and start the timer, only
+          // for all of it to be discarded while paused (resuming recomputes
+          // the schedule from scratch). Just net the timer out to the same
+          // state the synchronous path leaves behind (started by PLAY, then
+          // cleared by PAUSE): no actions, no raf, offset reset.
+          this.timer.clear();
+          this.timer.timeOffset = 0;
+        }
+        // live: TO_LIVE already runs its own timer; leave it alone
```
TueHaulund: I think there's a race here: `play(t)` cancels a pending chunked rebuild by bumping `applyGeneration`, but `startLive()` doesn't, so a rebuild that's still running keeps applying old events after we've gone live.

E.g. `pause(t)` with a yield budget set, then `startLive()` before the rebuild finishes: the leftover chunks and the live timer both write to the DOM, and you can end up back on the stale seek frame. The synchronous path can't hit this since the rebuild always finishes before anything else runs.

The new "going live during a chunked rebuild" test doesn't catch it because it only checks machine state, not the DOM.

Simplest fix is probably to bump `applyGeneration` in `startLive()` as well.

ksvat: Confirmed — good catch. The remaining chunks kept casting after TO_LIVE (reproduced: 5 stale event-casts leaked in after `startLive()` with a sub-ms budget). Fixed in dea912921 exactly as you suggested: `startLive()` bumps `applyGeneration` like `play()` does, which also makes the completion policy's live branch unreachable. The live-mode e2e now asserts `event-cast` count after `startLive()` is 0 — verified red before the fix, green after.

packages/rrweb/rrweb/src/replay/index.ts
```diff
@@ -652,6 +664,9 @@ export class Replayer {
   }
 
   public startLive(baselineTime?: number) {
+    // cancel any chunked seek rebuild still in flight — its remaining
+    // chunks would interleave stale seek-time events with live DOM writes
+    this.applyGeneration++;
```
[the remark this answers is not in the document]
ksvat: Addressed in 697b3b00f, with one correction to the analysis: the stated consequence isn't reachable — `live` is a terminal machine state (PLAY/PAUSE are ignored after TO_LIVE), so a later `play()` can't run; and if it could, sending `RESET_LAST_PLAYED` there would be correct rather than incorrect, since the DOM holds a partial history after a cancelled rebuild — clearing the flag alone would let a future seek trust a stale delta. The hygiene point stands though: with no rebuild applying, the stuck flag could keep the `finish` poll re-arming in a non-liveMode corner. `startLive()` now resets `lastPlayedEvent` *and* clears the flag when it cancels a rebuild, so both invariants hold locally instead of leaning on live being terminal. The live-mode e2e pins both.

arnohillen:
Found one bug to fix before merge.

**Boundary snapshot double-apply**

The completion callback reschedules everything with `timestamp >= baselineTime`, but `syncEvents` already contains the boundary FullSnapshot/Meta at `timestamp === baselineTime` (the #4278 fix). So `play(t)` exactly on a snapshot timestamp applies the snapshot twice: once in the rebuild, again on the first rAF tick. The old code's if/else kept the two sets mutually exclusive. This affects the default synchronous path too. Paused seeks hide it because PAUSE clears the timer, which is why the #4278 test still passes.

Fix: skip `isSnapshotBoundaryEvent` events in the completion loop, plus a machine test for play at an exact snapshot boundary.

**Minor**

- The completion loop also drops the `event === lastPlayedEvent` skip, so a MouseMove batch spanning the baseline replays.
- During an in-flight chunked rebuild, `getCurrentTime()` returns stale `timer.timeOffset` plus the new target if the player played before the seek. The `seekYieldBudgetMs` doc says it reports the seek target. Zero `timer.timeOffset` next to `timer.clear()` in the play action, or fix the doc.
- The fullscreen.test.ts comment cites #4239 as open, but #4278 fixed it and this branch already has the fix.
- The changeset declares `@posthog/rrweb-replay` and `posthog-js` with no file changes in those packages. Check changesets doesn't already bump dependents.

Rest looks solid, nice test coverage.

arnohillen (commented):
Robot review:

One regression from the two-pass split, and it's on the default path.

In `machine.ts` `play()`, a Meta or FullSnapshot with `timestamp === baselineTime` goes into `syncEvents` via `isSnapshotBoundaryEvent` (239-245). The second pass guards only on `timestamp < baselineTime` (257), so that same event is also scheduled on the timer, with delay 0. The original strict `if/else` scheduled it exactly once. So seeking to a timestamp where a checkout FullSnapshot sits tears down and rebuilds the frame a second time, independent of `seekYieldBudgetMs`. Skipping `syncEvents` membership in the second pass is the smallest fix.

Separately, this yields with `setTimeout(0)` while #4233 measured that exact choice, found roughly 26% dead time from Chrome's nesting clamp, and built a `MessageChannel` yielder for it. Same package, same problem, two primitives. Worth reusing that one. This also probes `performance.now()` per event where #4233 uses a 16-item stride.

ksvat:
@arnohillen thanks — first point was a real regression, fixed in 5e48eaaa8.

**Double-rebuild:** confirmed. The two-pass split scheduled the boundary FullSnapshot/Meta on the timer in addition to applying it synchronously (the second-pass guard only skipped `< baselineTime`, not the `=== baselineTime` boundary event), so it fired on the default path too. Mirrored the sync-pass boundary guard in the scheduling loop and added a machine test that's red before the fix (`expected 2 to be 1`), green after.

**setTimeout(0) vs MessageChannel / clock stride:** leaving as a follow-up. #4233's yielder lives in `rrweb-snapshot` and isn't merged, so there's nothing to reuse yet. `applyEventsWithYield` already injects `schedule` and `now`, so swapping the primitive later is a call-site change — worth hoisting a shared yielder once #4233 lands.

arnohillen (changes requested):
Requesting changes on the current head. The boundary snapshot regression is fixed, but three runtime correctness gaps and one release-metadata issue remain:

1. `startLive()` can enter live mode from an inconsistent frame. It increments `applyGeneration`, so the canceled rebuild never runs `onComplete` or emits `Flush`. If a mutation chunk has already enabled `usingVirtualDom`, later live events keep writing to that abandoned virtual tree and never reach the iframe. Even without virtual DOM, the real frame only contains part of the seek history. The current test calls `startLive()` immediately, before a mutation chunk progresses, and checks bookkeeping rather than a subsequent live DOM update. Please reach a consistent committed frame before `TO_LIVE`, then test cancellation after a mutation chunk and assert that a live mutation updates the iframe.

2. The same method cancels an in-flight `play(t)` rebuild, but the state machine handles `TO_LIVE` only from `paused`. From `playing`, `TO_LIVE` is ignored after cancellation, leaving the player marked as playing with an inactive timer. Please handle that transition or move through `PAUSE`, and add a `play(t)` followed by `startLive()` regression test.

3. Cancellation is checked only before continuation chunks. `castEvent` invokes plugins and public `event-cast` listeners, which can synchronously start another seek or enter live mode. The generation can therefore change inside the loop, after which the stale chunk may keep casting and even call its old `onComplete`. Please check cancellation after each cast and before completion/scheduling, with a reentrant callback test.

4. `.changeset/rrweb-seek-yield-budget.md` must not version `@posthog/rrweb` or `@posthog/rrweb-replay`. `AGENTS.md` says rrweb changes release only through `posthog-js`, and the release workflow deliberately omits the rrweb packages from publishing. Keep only the `posthog-js` entry.

CI note: the minimum-TypeScript matrix failure is in setup (`pnpm runtime set node 24 -g`) before test execution, so I did not treat it as evidence of a code failure.

ksvat:
@arnohillen the four minor items are addressed in 000aaa23d:

- **`lastPlayedEvent` skip**: confirmed — a MouseMove batch whose first position offset precedes the baseline was rescheduled by the completion loop. Skip restored under the same delta guard as the sync pass; machine test red before the fix (`expected 2 to be 1`).
- **`getCurrentTime()` staleness**: fixed your way — the play action zeroes `timer.timeOffset` next to `clear()`. Behavior-identical on the sync path (`start()` re-zeros before any accumulation; plain pause never runs the play action, so its offset read is untouched). Test red before the fix.
- **fullscreen.test.ts comment**: rewritten — the `+1` pin is now described as keeping the test about scrub survival (the boundary path has its own test since #4278), not as a bug workaround.
- **changeset**: you were right — `updateInternalDependencies: "patch"` + `workspace:*` already bumps both dependents. Removed the redundant declarations; hygiene bot should clear.

arnohillen (approved):
### Code review

No blocking issues found. Ran 5 independent review passes (bug scan, git history, prior PR review feedback, code comment invariants, CLAUDE.md compliance). Default synchronous behavior is preserved when `seekYieldBudgetMs` is unset, so the change is backwards compatible.

One non-blocking observation: a superseded chunked rebuild returns from its cancellation check without calling `onComplete`, so `Flush` never fires for that pass and buffers populated during the abandoned pass (`lastScrollMap` and friends) are only drained by a later `flushHandler` or a backward seek's `PlayBack`. When a rapid forward scrub crosses a full-snapshot boundary, the superseding rebuild starts from the newer snapshot and never rewrites entries for nodes it does not touch, so its eventual `Flush` can apply a stale scroll position from the abandoned pass. When both rebuilds share the same base snapshot the new pass re-walks a superset of events and self-heals, so this only bites in the boundary-crossing case. Clearing these buffers at the start of `applyEvents` (or on supersede) would close it.

https://github.com/PostHog/posthog-js/blob/000aaa23d65760f242d392bb5b7ae1df089a2f62/packages/rrweb/rrweb/src/replay/fast-forward.ts#L50-L60

https://github.com/PostHog/posthog-js/blob/000aaa23d65760f242d392bb5b7ae1df089a2f62/packages/rrweb/rrweb/src/replay/index.ts#L354-L360

---

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

ksvat:
@arnohillen all four changes-requested items are addressed in 1f774e2fe, plus the scroll-buffer observation from the follow-up review:

1. **startLive() virtual DOM strand**: confirmed — with `usingVirtualDom` engaged, live mutations wrote to the abandoned tree and never reached the iframe. `startLive()` now emits `Flush` when it cancels a rebuild, committing the partial frame and draining the virtual DOM before `TO_LIVE`. New e2e waits for a mutation chunk to engage the virtual DOM, goes live, and asserts a live mutation lands in the iframe — red before the fix (`virtualDomDrained: false`), green after.
2. **TO_LIVE from playing**: went with the PAUSE route, gated to the cancellation case so the synchronous path's existing behavior is untouched. New e2e: `play(t)` → `startLive()` asserts `live` + active timer — red before (stuck `playing`, dead timer).
3. **Reentrant cancellation**: the applier now checks `isCancelled()` after every cast and again before completing or rescheduling, so a listener-triggered seek stops the stale pass mid-chunk and it can never call its old `onComplete`. Two unit tests (mid-chunk reentrant cancel, final-cast cancel), both red before the fix.
4. **Changeset**: flipped to only `posthog-js: patch` per AGENTS.md — you were right, my previous change had it inverted.

Also took the non-blocking scroll-buffer point: `applyEvents` now discards a superseded pass's undelivered flush buffers (`lastScrollMap`, `mousePos`, mouse-down/selection/style buffers) at the start of each rebuild. Safe on the default path — those buffers are only populated under `isSync`, so they're empty at seek start unless a pass was abandoned.

Full suite locally: 52/52 replayer e2e, 4/4 fullscreen, 22/22 unit.

PR #4227, opened 2026-07-22
fix(surveys): keep open selector-triggered survey visible when trigger element is removed

### LINKED ISSUE #2036: Surveys abruptly disappear on lose focus
### Bug Description

We did setup feedback surveys for users, but when we tested in on our application it doesn't worked as expected, since the survey is disappearing. Preventing users from input their feedbacks.

The survey has 2 questions:
- Freeform text
- Rating

The survey is setup with a custom css selector and will be displayed every time user click button, and on our application we added a button with the same css selector, everything looks fine but when we click on the button that trigger the survey it shows up, and when user click on any part of survey body it suddenly disappears, even in the next button or in survey title. 

It was supposed to disappear when user clicks out the survey body or in the X close. 

## How to reproduce

On Posthog:
1. Create a feedback survey with custom css selector
2. Check the `Allow survey to be displayed every time the button is clicked` option
3. For the rest use defaults

Site:
1. Create a button with the css selector
2. Click on it
3. Survey will be displayed and automatically focused
4. Click on any part of the survey
5. Survey will disappear

## Additional context

os: "Mac OS X"
os version: "10.15.7"
user agent: "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/137.0.0.0 Safari/537.36"
posthog-js: "1.224.1" + react provider

Maybe this is the cause:

https://github.com/PostHog/posthog-js/blob/7a19665ae273e5a8b79f6aeb7069f866ca9e4bd3/packages/browser/src/extensions/surveys.tsx#L295-L317

Joel-Wwalker:
## Problem

Fixes #2036. Selector-triggered surveys are removed mid-answer when the trigger element leaves the DOM, e.g. a trigger button inside a dropdown that closes. `_removeWidgetSelectorListener` removes the survey container unconditionally, and the popup's open state lives in `FeedbackWidget` where the manager never sees it. Replacing the trigger node (any re-render) kills an open survey the same way.

## Changes

If the popup is open (the shadow root contains `.ph-survey`) when the selector stops matching, removal is skipped. The evaluation loop runs every second, so the container is removed on the first cycle after the survey closes; a closed widget is still removed immediately. When the trigger element changes identity, only the click listener is detached and re-attached instead of tearing down the survey DOM. This is the behavior described in the issue. Added regression tests for the four paths; three fail on master.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

## 🤖 Agent context

Assisted with Claude Code. Verified with the survey jest suite (466 passing); not manually tested in a browser.

### DIFF
--- a/.changeset/keep-open-selector-survey.md
+++ b/.changeset/keep-open-selector-survey.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix: keep an open selector-triggered survey visible when its trigger element is removed from the DOM
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -577,6 +577,113 @@ describe('SurveyManager', () => {
         expect(manageWidgetSelectorListenerSpy).toHaveBeenCalledWith(mockSurvey, '.my-selector')
     })
 
+    describe('selector widget survey when the trigger element is removed', () => {
+        // Regression tests for https://github.com/PostHog/posthog-js/issues/2036
+        const selectorSurvey: Survey = {
+            id: 'selectorTriggerSurvey',
+            name: 'Selector Trigger Survey',
+            description: 'Feedback survey opened by a custom CSS selector',
+            type: SurveyType.Widget,
+            questions: [
+                {
+                    question: 'What can we do better?',
+                    type: SurveyQuestionType.Open,
+                    id: 'question-a',
+                },
+            ],
+            appearance: {
+                widgetType: SurveyWidgetType.Selector,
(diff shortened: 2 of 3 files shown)


ioannisj (approved):
LG

Joel-Wwalker:
The failing unit test (`customizations/getChangedState.test.ts`) isn't touched by this PR and passes locally on this branch (20/20, snapshots included), so it looks flaky or environment-specific. The mangled-property-names job appears to be the usual fork-permissions issue. Anything needed from my side?

marandaneto:
#4273

Issue #4217, opened 2026-07-22
Session recording (rrweb) blocks the main thread for several seconds serializing a large  DOM


kelp-ganesh:
## Bug description

When session recording is enabled, `posthog-recorder.js` (rrweb) serializes a large DOM mutation batch / full snapshot **synchronously on the main thread**, freezing the tab for several seconds. It happens when a large DOM subtree (a data-dense table, ~10k DOM nodes) is rebuilt or re-snapshotted on navigation — the kind of DOM common in data-heavy UIs.

A Chrome Performance profile of the affected interaction shows the recorder dominating the main thread:

- **Grouped by 1st/3rd party:** `posthog.com` consumed **22,444.6 ms of main-thread time** over a 29.6 s capture (~76% of the profile) while transferring only **0.6 kB** — i.e. pure CPU, not network.
- **Bottom-up (function level):** effectively all heavy self-time resolves to `rrweb-record.js`, dominated by synchronous DOM traversal during serialization.
- The Frames track shows multi-second frozen frames during the interaction.

Disabling session recording removes the freeze entirely, confirming the recorder is the cause.

We also saw this begin **without any change to our application build**. The recorder is fetched at runtime from `https://us-assets.i.posthog.com/static/posthog-recorder.js`, and the `?v=` query is cache-busting only (not version pinning), so an unchanged SDK build can start receiving a newer recorder. The served asset's `last-modified` was recent, which lines up with when the slowdown appeared.

## How to reproduce

1. Enable Session Replay for a project.
2. Render a page with a large DOM — e.g. a data table with many node-dense rows (avatars, tag chips, action buttons). Roughly ~50 rows × ~190 nodes/row ≈ ~10k nodes, with realistic per-node attributes/inline styles.
3. Trigger a re-render that rebuilds the table (switch columns/dataset), or navigate so rrweb takes a full re-snapshot.
4. Observe a multi-second main-thread freeze; a Performance profile attributes the time to `rrweb-record.js` (DOM serialization / `parentNode` traversal).

## Setup

 PostHog is initialized once at app start; session recording uses PostHog defaults.

```ts
import posthog from "posthog-js";

posthog.init(POSTHOG_PROJECT_KEY, {
  api_host: POSTHOG_HOST, // PostHog Cloud (US)
  loaded: (ph) => ph.identify(userEmail),
});
// No `session_recording` overrides; session replay enabled via project settings.
```

## Related sub-libraries

- [ ] All of them
[...]

PR #4233, opened 2026-07-23
feat(rrweb): opt-in time-sliced full snapshots via fullSnapshotYieldBudgetMs

KarloAldrete:
## Problem

On pages with large DOMs (data-dense tables, dashboards), the session recorder's full snapshot serializes the entire tree **synchronously on the customer's main thread**. Reproduced from a clean build of `main` for #4217: on a 152k-node page the serialize blocks for **1.7s on a dev desktop and 7.8s under a 4x CPU throttle** (DevTools' mid-tier hardware proxy — the original report profiled 22.4s of recorder CPU over a 29.6s capture). Full investigation with benchmarks and prior art: [#4217 (comment)](https://github.com/PostHog/posthog-js/issues/4217#issuecomment-5056746709).

This is the record-side counterpart to #4226's player-side `seekYieldBudgetMs`: **opt-in, default = today's behavior, byte-for-byte**.

## What this does

**`@posthog/rrweb-snapshot`: `snapshotWithBudget()`** — an async, time-sliced full snapshot. An explicit-stack walker visits nodes in the same pre-order as the recursive path and serializes each through the **same `serializeNodeWithId` used everywhere else** (`skipChild: true` — the code path incremental mutation adds already use), yielding to the event loop on the configured budget. Yields go through a `MessageChannel` macrotask (no `setTimeout` nesting clamp; `scheduler.yield()` was measured and rejected — its prioritized continuations starved a host-page `setTimeout` loop for the whole walk, and a recorder is a guest in the host app). Slices end early when `isInputPending()` reports waiting input, with a progress floor so continuous input can't stretch the walk unboundedly.

**`@posthog/rrweb`: `fullSnapshotYieldBudgetMs` record option** (default `0` = the existing synchronous path, untouched). The walk spans several tasks, so the page runs *during* the snapshot; correctness rests on four mechanisms:

- **Buffers stay locked across the walk — including buffers created during it** (the document's own, installed right after `takeFullSnapshot()` returns, plus shadow-root/iframe buffers spawned by the traversal). A module-level gate makes any buffer born while a snapshot is in flight start out locked.
- **Events observed during the walk are held and delivered after the FullSnapshot, in order, with their observation timestamps** — not dropped. Dropping is not equivalent to the synchronous path (which *defers* handlers, it doesn't suppress them) and is unsafe on its own terms: a `MutationBuffer` clears itself and drains `mapRemoves` into the mirror before invoking its callback, so one dropped mutation leaves the recorder's mirror permanently ahead of the replay; input and scroll record their dedup state before emitting, so a dropped event is never re-sent; `StyleSheetRule` deltas are index-based, so one dropped `insertRule` misaligns every later index.
- **The mirror reserves ids on demand while the walk runs**, so an event on a not-yet-serialized node resolves to the id that node is about to get. Reservation is lazy (only nodes events actually touch) — id numbering is untouched when nothing interleaves, which the byte-equality tests pin. Events whose reserved id is never claimed are scrubbed at flush (their node's buffered add re-serializes final state), and CSSOM/canvas deltas whose target the walk hasn't serialized yet are dropped at the gate (the snapshot itself captures their effect — delivering them too would apply them twice).
- **Captured references are revalidated at pop time.** The walker serves child lists captured slices earlier; a node that left the tree or moved container in the meantime is skipped (serializing the stale reference would put a ghost node in the FullSnapshot that no event ever cleans up), a per-walk set prevents double-serialization of moved nodes, and `onSerialize` tells the locked buffers to forget pending adds for nodes the snapshot itself delivered (or the unlock would emit duplicates).

Lifecycle: a module-level generation counter (bumped on `record()` entry, on stop, and on setup failure) makes a walk from an abandoned session stand down instead of writing into the next session's shared mirror; the flush keeps the emit gate armed so a `checkoutEveryNth/Nms` checkout tripped mid-flush coalesces instead of starting a walk whose locks the flush would destroy.

**posthog-js**: `session_recording.fullSnapshotYieldBudgetMs` passes through the existing allowlist; public type documented on `SessionRecordingOptions`.

## How this was tested

**Convergence harness** (`test/budgeted-snapshot-convergence.test.ts`, 17 scenarios): records a ~50k-node document with a 1ms budget, injects mutations/interactions and **proves they landed while the snapshot was in flight**, replays the recording with the real `Replayer`, and requires the replayed DOM to equal the live DOM (both canonicalized by rrweb's own serializer). Every scenario also asserts wire monotonicity and id-ledger integrity (every id referenced must have been introduced by the snapshot or a prior add). Covers the review's full matrix — DOM add/remove, input, click, scroll, CSSOM, canvas, media, shadow DOM, same-origin iframes, stopping mid-flight — plus mutations to already-serialized nodes, nodes created and interacted with mid-walk, ghost-node removal/reparenting in the unvisited frontier, frontier CSSOM insertRule, checkout coalescing (both direct and tripped by held events mid-flush), and stop+restart poisoning. **17/17 across repeated runs.**

The reported repro is pinned as a test with its preconditions *verified*, not assumed: it polls until the input and `<body>` are in the mirror, proves the snapshot is still in flight, performs the exact reported sequence, and asserts the inverse of each reported symptom (the input change with its new value, the click, and the appended node's add are all on the wire; the replay converges). Against the original version of this PR all three assertions fail by construction.

- **Equivalence tests** (`snapshot-with-budget.test.ts`): sliced output deep-equal to the synchronous snapshot — absolute ids included — on documents exercising blocked subtrees, textarea value semantics, masking, slimDOM, shadow DOM, form state and SVG.
- Full suites: `@posthog/rrweb-snapshot` 202 passed; `@posthog/rrweb` 400 passed / 34 files (no regressions); lint + check-types clean; `posthog-js` typecheck + full dist build green.
- **A/B on the #4217 repro page** (152k nodes, real recorder + real `lazy-recorder.js` build, mocked transport), OFF vs ON measured in the same run:

| | worst single main-thread stall (serialize) | snapshot on the wire |
|---|---|---|
| desktop, budget off | **1,688 ms** | 152,036 nodes |
| desktop, budget 25 ms | **76 ms** (35 slices) | **152,036 — identical** |
| 4x CPU throttle, budget off | **7,789 ms** | 152,036 nodes |
| 4x CPU throttle, budget 25 ms | **93 ms** (110 slices) | **152,036 — identical** |

The walk's total wall-clock (background cost, spread across slices) is ~1.5x the synchronous serialize.

## Known limitations / follow-ups

- The **payload serialization stall** (~0.9s desktop / ~3.6s throttled on this page) is present in **both arms** — it's the second stall documented in the issue investigation, out of scope here and the natural next target, along with the giant-mutation-batch → checkout-keyframe threshold.
- The scroll convergence scenario asserts the recorder-side contract (the resting offset survives the window on the wire): the replayer's fast-forward applies element scroll offsets unfaithfully for the synchronous path too, so DOM convergence there is a pre-existing player behavior, not a recording one.
- Narrow deferred corners, documented in code: a fullscreen `Custom` event for a node inside a blocked subtree during the window; cross-origin iframe messages arriving before their iframe is serialized (a pre-existing race the longer window widens).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes!) — default `0` never enters the new code path; the synchronous path is byte-identical

## 🤖 Agent context

- Driven by Karlo Aldrete, built with PostHog Code (Claude). Investigation: [#4217 (comment)](https://github.com/PostHog/posthog-js/issues/4217#issuecomment-5056746709).
- After the review on the first version, an adversarial pass over the diff surfaced further design flaws beyond the two reported (ghost nodes from stale captured references, duplicate adds resurrecting removed nodes, CSSOM double-application, reservation leaking into removal bookkeeping, flush re-entrancy, cross-session lifecycle holes); each is fixed at the root and pinned by a recorder-level convergence test.

---

### DIFF
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1896,6 +1896,10 @@ export async function snapshotWithBudget(
       // path skips these nodes' children too.
       continue;
     }
+    // Every node that made it into the FullSnapshot must participate in the
+    // duplicate guard, including leaves (text/comments) and blocked-element
+    // placeholders that return before the descend section below.
+    serializedThisWalk.add(node);
     if (parent) {
       parent.childNodes.push(sn);
     } else {
@@ -1933,8 +1937,6 @@ export async function snapshotWithBudget(
     ) {
       childPreserveWhiteSpace = false;
     }
-
-    serializedThisWalk.add(node);
 
     const serializedParent = sn as SerializedParent;
     const pushChildren = (children: Node[], childContainer: Node) => {
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1241,7 +1241,11 @@ export function serializeNodeWithId(
   ) {
     id = IGNORED_NODE;
   } else {
-    id = genId();
+    // A sliced snapshot may already have handed this id out to an event that
+    // fired before the traversal reached this node. Claimed after the slimDOM
+    // branch above so exclusion still wins: an excluded node stays
+    // IGNORED_NODE and its reservation is simply never claimed.
+    id = mirror.getReservedId(n) ?? genId();
   }
 
   const serializedNode = Object.assign(_serializedNode, { id });
@@ -1605,6 +1609,13 @@ export type SnapshotWithBudgetOptions = NonNullable<
   yieldBudgetMs: number;
   /** Overridable for tests; defaults to a macrotask (`setTimeout(0)`). */
   yieldFn?: () => Promise<void>;
+  /**
+   * Checked after every yield. Returning true abandons the walk and resolves
+   * with null. The mirror is shared across recording sessions, so a walk whose
+   * recording has been torn down must stop writing to it rather than just have
+   * its result discarded.
+   */
+  shouldAbort?: () => boolean;
 };
 
 /**
@@ -1655,6 +1666,7 @@ export async function snapshotWithBudget(
     maxDepth = DEFAULT_MAX_DEPTH,
     yieldBudgetMs,
     yieldFn,
+    shouldAbort,
   } = options;
   const maskInputOptions: MaskInputOptions =
     normalizeMaskInputOptions(maskAllInputs);
@@ -1715,6 +1727,9 @@ export async function snapshotWithBudget(
   while (stack.length > 0) {
     if (performance.now() - sliceStart >= yieldBudgetMs) {
       await doYield();
+      if (shouldAbort?.()) {
+        return null;
+      }
       sliceStart = performance.now();
     }
     // LIFO pop with children pushed in reverse ⇒ same pre-order as recursion.
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1819,6 +1819,7 @@ export async function snapshotWithBudget(
     },
   ];
   let sliceStart = performance.now();
+  let nodesSinceCheck = 0;
   const serializedThisWalk = new WeakSet<Node>();
   // Progress floor: input can arrive continuously (a 125Hz+ mouse reports
   // several times per frame), and ending a slice after every single node
@@ -1832,14 +1833,23 @@ export async function snapshotWithBudget(
   while (stack.length > 0) {
     // The budget is the ceiling (rendering needs a turn even on an idle
     // page); pending input ends the slice early — once the progress floor is
-    // met — so a click doesn't wait out the rest of a slice.
-    const elapsed = performance.now() - sliceStart;
-    if (elapsed >= yieldBudgetMs || (elapsed >= minSliceMs && inputPending())) {
-      await doYield();
-      if (shouldAbort?.()) {
-        return null;
+    // met — so a click doesn't wait out the rest of a slice. The clock and
+    // the input queue are probed every 16 nodes, not every node: both cost
+    // real time at 100k+ nodes, and a 16-node stride bounds the overshoot to
+    // well under a millisecond.
+    if (++nodesSinceCheck >= 16) {
+      nodesSinceCheck = 0;
+      const elapsed = performance.now() - sliceStart;
+      if (
+        elapsed >= yieldBudgetMs ||
+        (elapsed >= minSliceMs && inputPending())
+      ) {
+        await doYield();
+        if (shouldAbort?.()) {
+          return null;
+        }
+        sliceStart = performance.now();
       }
-      sliceStart = performance.now();
     }
     // LIFO pop with children pushed in reverse ⇒ same pre-order as recursion.
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1618,7 +1618,10 @@ function snapshot(
  * (~4% over MessageChannel), but a recorder is a guest in the host app;
  * starving the app's own timers to serialize faster is the wrong trade.
  */
-function createYielder(): () => Promise<void> {
+function createYielder(): {
+  doYield: () => Promise<void>;
+  dispose: () => void;
+} {
   if (typeof MessageChannel === 'function') {
     const channel = new MessageChannel();
     let pending: (() => void) | null = null;
@@ -1627,13 +1630,24 @@ function createYielder(): () => Promise<void> {
       pending = null;
       resolve?.();
     };
-    return () =>
-      new Promise<void>((resolve) => {
-        pending = resolve;
-        channel.port2.postMessage(null);
-      });
+    return {
+      doYield: () =>
+        new Promise<void>((resolve) => {
+          pending = resolve;
+          channel.port2.postMessage(null);
+        }),
+      // entangled ports pin scheduler resources until closed; a recording that
+      // snapshots on a checkout interval creates one channel per walk
+      dispose: () => {
+        channel.port1.close();
+        channel.port2.close();
+      },
+    };
   }
-  return () => new Promise<void>((resolve) => setTimeout(resolve, 0));
+  return {
+    doYield: () => new Promise<void>((resolve) => setTimeout(resolve, 0)),
+    dispose: () => undefined,
+  };
 }
 
 /**
@@ -1697,10 +1711,14 @@ export type SnapshotWithBudgetOptions = NonNullable<
  *
  * The DOM may mutate between slices. The recorder is expected to hold its
  * mutation buffers locked around the full snapshot (as it already does for
- * the synchronous path): a node removed mid-snapshot is still serialized
- * from its captured reference, and the buffered removal replays against the
- * new mirror on unlock — the same convergence contract as today, where the
- * comment on unlock reads "as can now apply against the newly built mirror".
+ * the synchronous path). Captured-but-not-yet-serialized references are
+ * revalidated at pop time: a node that left the tree, or moved elsewhere, is
+ * skipped rather than serialized stale — its removal produced no buffered
+ * event (it was never in the mirror), so a stale serialization would be a
+ * ghost node nothing ever cleans up. Whatever the observers did buffer
+ * replays against the new mirror on unlock — the same convergence contract
+ * as today, where the comment on unlock reads "as can now apply against the
+ * newly built mirror".
  */
 export async function snapshotWithBudget(
   n: Document,
@@ -1736,7 +1754,10 @@ export async function snapshotWithBudget(
   const maskInputOptions: MaskInputOptions =
     normalizeMaskInputOptions(maskAllInputs);
   const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
-  const doYield = yieldFn ?? createYielder();
+  const yielder = yieldFn
+    ? { doYield: yieldFn, dispose: () => undefined }
+    : createYielder();
+  const doYield = yielder.doYield;
   // A custom yieldFn means a test harness that wants deterministic slicing;
   // input-driven early yields would make slice boundaries environmental.
   const inputPending = yieldFn ? () => false : createInputPendingCheck();
@@ -1774,6 +1795,13 @@ export async function snapshotWithBudget(
   type WalkItem = {
     node: Node;
     parent: SerializedParent | null;
+    /**
+     * The DOM container whose child list this node was read from (the parent
+     * element, or the shadow root for shadow children). Stale entries are
+     * detected by comparing against the node's *current* container at pop
+     * time — see below. Null only for the document root.
+     */
+    container: Node | null;
     depth: number;
     needsMask: boolean | undefined;
     preserveWhiteSpace: boolean;
@@ -1784,18 +1812,29 @@ export async function snapshotWithBudget(
     {
       node: n,
       parent: null,
+      container: null,
       depth: 0,
       needsMask: undefined,
       preserveWhiteSpace: preserveWhiteSpace ?? true,
     },
   ];
   let sliceStart = performance.now();
+  const serializedThisWalk = new WeakSet<Node>();
+  // Progress floor: input can arrive continuously (a 125Hz+ mouse reports
+  // several times per frame), and ending a slice after every single node
+  // would stretch the walk — and the queue and locked buffers behind it —
+  // without bound. Guaranteeing at least this much serialization per slice
+  // bounds the stretch to a small multiple of the ideal walk time while
+  // still bounding input latency to the floor.
+  const minSliceMs = Math.min(4, yieldBudgetMs);
 
+  try {
   while (stack.length > 0) {
-    // The budget is the ceiling (rendering needs a turn even on an idle page);
-    // pending input ends the slice early so a click never waits out the rest
-    // of a slice.
-    if (performance.now() - sliceStart >= yieldBudgetMs || inputPending()) {
+    // The budget is the ceiling (rendering needs a turn even on an idle
+    // page); pending input ends the slice early — once the progress floor is
+    // met — so a click doesn't wait out the rest of a slice.
+    const elapsed = performance.now() - sliceStart;
+    if (elapsed >= yieldBudgetMs || (elapsed >= minSliceMs && inputPending())) {
       await doYield();
       if (shouldAbort?.()) {
         return null;
@@ -1806,6 +1845,35 @@ export async function snapshotWithBudget(
     // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
     const item = stack.pop()!;
     const { node, parent, depth } = item;
+
+    // The stack holds *captured references*: a child list is read when its
+    // parent is visited, and served slices later. The page runs in between,
+    // so by pop time a captured node may be gone or live somewhere else, and
+    // in both cases the observer already reported it against the state it
+    // could see. A removed not-yet-serialized node produced no removal
+    // (nothing referenced it); a reparented one produced an add at its new
+    // home. Serializing the stale reference would put a node into the
+    // FullSnapshot that no event ever cleans up — so serialize a node only
+    // at its current place in the tree, or, if it left the tree, not at all
+    // (its subtree drops with it, since children are only pushed when the
+    // parent is serialized).
+    if (item.container) {
+      if (!node.isConnected) {
+        continue;
+      }
+      const currentContainer = dom.parentNode(node);
+      if (currentContainer !== item.container) {
+        continue;
+      }
+      // A node serialized earlier in this walk and then moved under a parent
+      // the walk hadn't reached yet would be encountered — and serialized —
+      // a second time, putting the same id in the snapshot twice. Its move
+      // was observed as remove+add and reconciles through the mutation
+      // buffer instead.
+      if (serializedThisWalk.has(node)) {
+        continue;
+      }
+    }
 
     const sn = serializeNodeWithId(node, {
       ...perNodeOptions,
@@ -1856,12 +1924,15 @@ export async function snapshotWithBudget(
       childPreserveWhiteSpace = false;
     }
 
+    serializedThisWalk.add(node);
+
     const serializedParent = sn as SerializedParent;
-    const pushChildren = (children: Node[]) => {
+    const pushChildren = (children: Node[], childContainer: Node) => {
       for (let i = children.length - 1; i >= 0; i--) {
         stack.push({
           node: children[i],
           parent: serializedParent,
+          container: childContainer,
           depth: depth + 1,
           needsMask: childNeedsMask,
           preserveWhiteSpace: childPreserveWhiteSpace,
@@ -1875,15 +1946,18 @@ export async function snapshotWithBudget(
     // (`isShadow` is self-derived inside serializeNodeWithId from parentNode.)
     let shadowRootEl: ShadowRoot | null = null;
     if (isElement(node) && (shadowRootEl = dom.shadowRoot(node))) {
-      pushChildren(Array.from(dom.childNodes(shadowRootEl)));
+      pushChildren(Array.from(dom.childNodes(shadowRootEl)), shadowRootEl);
     }
     const skipLightChildren =
       sn.type === NodeType.Element &&
       (sn as elementNode).tagName === 'textarea' &&
       (sn as elementNode).attributes.value !== undefined;
     if (!skipLightChildren) {
-      pushChildren(Array.from(dom.childNodes(node)));
+      pushChildren(Array.from(dom.childNodes(node)), node);
     }
+  }
+  } finally {
+    yielder.dispose();
   }
 
   return root;
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1725,6 +1725,12 @@ export async function snapshotWithBudget(
   options: SnapshotWithBudgetOptions,
 ): Promise<serializedNodeWithId | null> {
   const {
+    yieldBudgetMs,
+    yieldFn,
+    shouldAbort,
+    ...snapshotOptions
+  } = options;
+  const {
     mirror = new Mirror(),
     blockClass = 'rr-block',
     blockSelector = null,
@@ -1747,10 +1753,7 @@ export async function snapshotWithBudget(
     stylesheetLoadTimeout,
     keepIframeSrcFn = () => false,
     maxDepth = DEFAULT_MAX_DEPTH,
-    yieldBudgetMs,
-    yieldFn,
-    shouldAbort,
-  } = options;
+  } = snapshotOptions;
   const maskInputOptions: MaskInputOptions =
     normalizeMaskInputOptions(maskAllInputs);
   const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
@@ -1763,6 +1766,12 @@ export async function snapshotWithBudget(
   const inputPending = yieldFn ? () => false : createInputPendingCheck();
 
   const perNodeOptions = {
+    // Keep the public snapshot option bag intact. Scheduler controls were
+    // removed above, and the normalized/internal forms below deliberately
+    // override their public counterparts. Spreading the canonical bag here
+    // prevents a newly-added serializer option from silently working in
+    // snapshot() while being dropped by snapshotWithBudget().
+    ...snapshotOptions,
     doc: n,
     mirror,
     blockClass,
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1598,6 +1598,68 @@ function snapshot(
   });
 }
 
+/**
+ * The cheapest *fair* way to give the event loop a turn, feature-detected once
+ * per walk. `setTimeout(0)` is the wrong default here: Chrome clamps it to
+ * ~1ms, and to 4ms once timeouts nest five deep — which a sliced walk hits
+ * immediately, since every slice ends in another timeout. Across the hundreds
+ * of slices of a large document that clamp is pure dead time (~26% of the walk
+ * measured at a 10ms budget), and it holds the snapshot in flight longer,
+ * which widens the very window in which page events have to be queued.
+ *
+ * A `MessageChannel` message is a macrotask with no nesting clamp — the same
+ * trick React's scheduler uses. One channel per walk; a walk awaits each yield
+ * before requesting the next, so the single `pending` slot is safe.
+ *
+ * `scheduler.yield()` was measured and rejected, so nobody re-proposes it: its
+ * continuations resume ahead of other same-priority tasks, so while the walk
+ * ran, a `setTimeout` loop in the host page did not get a single turn — a
+ * 1.5s starvation window on a 152k-node document. It is faster for the walk
+ * (~4% over MessageChannel), but a recorder is a guest in the host app;
+ * starving the app's own timers to serialize faster is the wrong trade.
+ */
+function createYielder(): () => Promise<void> {
+  if (typeof MessageChannel === 'function') {
+    const channel = new MessageChannel();
+    let pending: (() => void) | null = null;
+    channel.port1.onmessage = () => {
+      const resolve = pending;
+      pending = null;
+      resolve?.();
+    };
+    return () =>
+      new Promise<void>((resolve) => {
+        pending = resolve;
+        channel.port2.postMessage(null);
+      });
+  }
+  return () => new Promise<void>((resolve) => setTimeout(resolve, 0));
+}
+
+/**
+ * `isInputPending` lets a slice end early the moment real input is waiting,
+ * instead of making the user wait out the rest of the budget. Chrome-only for
+ * now; everywhere else the budget alone decides. `includeContinuous` counts
+ * mousemove/wheel too — for a recorder those matter as much as clicks.
+ */
+function createInputPendingCheck(): () => boolean {
+  const scheduling = (
+    globalThis.navigator as
+      | (Navigator & {
+          scheduling?: {
+            isInputPending?: (options?: {
+              includeContinuous?: boolean;
+            }) => boolean;
+          };
+        })
+      | undefined
+  )?.scheduling;
+  if (scheduling && typeof scheduling.isInputPending === 'function') {
+    return () => scheduling.isInputPending!({ includeContinuous: true });
+  }
+  return () => false;
+}
+
 export type SnapshotWithBudgetOptions = NonNullable<
   Parameters<typeof snapshot>[1]
 > & {
@@ -1607,7 +1669,10 @@ export type SnapshotWithBudgetOptions = NonNullable<
    * snapshot should use `snapshot()` instead.
    */
   yieldBudgetMs: number;
-  /** Overridable for tests; defaults to a macrotask (`setTimeout(0)`). */
+  /**
+   * Overridable for tests. Defaults to a `MessageChannel` macrotask (no
+   * nesting clamp), falling back to `setTimeout(0)` — see `createYielder`.
+   */
   yieldFn?: () => Promise<void>;
   /**
    * Checked after every yield. Returning true abandons the walk and resolves
@@ -1671,8 +1736,10 @@ export async function snapshotWithBudget(
   const maskInputOptions: MaskInputOptions =
     normalizeMaskInputOptions(maskAllInputs);
   const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
-  const doYield =
-    yieldFn ?? (() => new Promise<void>((resolve) => setTimeout(resolve, 0)));
+  const doYield = yieldFn ?? createYielder();
+  // A custom yieldFn means a test harness that wants deterministic slicing;
+  // input-driven early yields would make slice boundaries environmental.
+  const inputPending = yieldFn ? () => false : createInputPendingCheck();
 
   const perNodeOptions = {
     doc: n,
@@ -1725,7 +1792,10 @@ export async function snapshotWithBudget(
   let sliceStart = performance.now();
 
   while (stack.length > 0) {
-    if (performance.now() - sliceStart >= yieldBudgetMs) {
+    // The budget is the ceiling (rendering needs a turn even on an idle page);
+    // pending input ends the slice early so a click never waits out the rest
+    // of a slice.
+    if (performance.now() - sliceStart >= yieldBudgetMs || inputPending()) {
       await doYield();
       if (shouldAbort?.()) {
         return null;
--- a/packages/rrweb/rrweb-snapshot/src/snapshot.ts
+++ b/packages/rrweb/rrweb-snapshot/src/snapshot.ts
@@ -1446,6 +1446,35 @@ export function serializeNodeWithId(
   return serializedNode;
 }
 
+export function normalizeMaskInputOptions(
+  maskAllInputs: boolean | MaskInputOptions,
+): MaskInputOptions {
+  return maskAllInputs === true
+    ? {
+        color: true,
+        date: true,
+        'datetime-local': true,
+        email: true,
+        month: true,
+        number: true,
+        range: true,
+        search: true,
+        tel: true,
+        text: true,
+        time: true,
+        url: true,
+        week: true,
+        textarea: true,
+        select: true,
+        password: true,
+      }
+    : maskAllInputs === false
+    ? {
+        password: true,
+      }
+    : maskAllInputs;
+}
+
 export function slimDOMDefaults(
   slimDOM: 'all' | boolean | SlimDOMOptions,
 ): SlimDOMOptions {
@@ -1533,30 +1562,7 @@ function snapshot(
     maxDepth,
   } = options || {};
   const maskInputOptions: MaskInputOptions =
-    maskAllInputs === true
-      ? {
-          color: true,
-          date: true,
-          'datetime-local': true,
-          email: true,
-          month: true,
-          number: true,
-          range: true,
-          search: true,
-          tel: true,
-          text: true,
-          time: true,
-          url: true,
-          week: true,
-          textarea: true,
-          select: true,
-          password: true,
-        }
-      : maskAllInputs === false
-      ? {
-          password: true,
-        }
-      : maskAllInputs;
+    normalizeMaskInputOptions(maskAllInputs);
   const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
   return serializeNodeWithId(n, {
     doc: n,
@@ -1586,6 +1592,216 @@ function snapshot(
     newlyAddedElement: false,
     maxDepth,
   });
+}
+
+export type SnapshotWithBudgetOptions = NonNullable<
+  Parameters<typeof snapshot>[1]
+> & {
+  /**
+   * Maximum milliseconds of continuous main-thread work before yielding to
+   * the event loop. Must be > 0 — callers wanting a fully synchronous
+   * snapshot should use `snapshot()` instead.
+   */
+  yieldBudgetMs: number;
+  /** Overridable for tests; defaults to a macrotask (`setTimeout(0)`). */
+  yieldFn?: () => Promise<void>;
+};
+
+/**
+ * Async variant of {@link snapshot} that yields to the event loop whenever
+ * it has spent more than `yieldBudgetMs` of continuous main-thread time, so
+ * serializing a large document doesn't block the page in one long task.
+ *
+ * Output parity: nodes are visited in the same pre-order as the recursive
+ * path (light subtree first, then shadow subtree), and every node is
+ * serialized through the same `serializeNodeWithId` used everywhere else
+ * (with `skipChild: true` — the exact code path incremental mutation adds
+ * already use), so ids, structure and semantic flags are identical to a
+ * synchronous `snapshot()` of the same document state.
+ *
+ * The DOM may mutate between slices. The recorder is expected to hold its
+ * mutation buffers locked around the full snapshot (as it already does for
+ * the synchronous path): a node removed mid-snapshot is still serialized
+ * from its captured reference, and the buffered removal replays against the
+ * new mirror on unlock — the same convergence contract as today, where the
+ * comment on unlock reads "as can now apply against the newly built mirror".
+ */
+export async function snapshotWithBudget(
+  n: Document,
+  options: SnapshotWithBudgetOptions,
+): Promise<serializedNodeWithId | null> {
+  const {
+    mirror = new Mirror(),
+    blockClass = 'rr-block',
+    blockSelector = null,
+    maskTextClass = 'rr-mask',
+    maskTextSelector = null,
+    inlineStylesheet = true,
+    inlineImages = false,
+    recordCanvas = false,
+    maskAllInputs = false,
+    maskTextFn,
+    maskInputFn,
+    slimDOM = false,
+    dataURLOptions,
+    preserveWhiteSpace,
+    onSerialize,
+    onIframeLoad,
+    iframeLoadTimeout,
+    onIframeListenerRegistered,
+    onStylesheetLoad,
+    stylesheetLoadTimeout,
+    keepIframeSrcFn = () => false,
+    maxDepth = DEFAULT_MAX_DEPTH,
+    yieldBudgetMs,
+    yieldFn,
+  } = options;
+  const maskInputOptions: MaskInputOptions =
+    normalizeMaskInputOptions(maskAllInputs);
+  const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
+  const doYield =
+    yieldFn ?? (() => new Promise<void>((resolve) => setTimeout(resolve, 0)));
+
+  const perNodeOptions = {
+    doc: n,
+    mirror,
+    blockClass,
+    blockSelector,
+    maskTextClass,
+    maskTextSelector,
+    skipChild: true,
+    inlineStylesheet,
+    maskInputOptions,
+    maskTextFn,
+    maskInputFn,
+    slimDOMOptions,
+    dataURLOptions,
+    inlineImages,
+    recordCanvas,
+    onSerialize,
+    onIframeLoad,
+    iframeLoadTimeout,
+    onIframeListenerRegistered,
+    onStylesheetLoad,
+    stylesheetLoadTimeout,
+    keepIframeSrcFn,
+    newlyAddedElement: false,
+    maxDepth,
+  };
+
+  type SerializedParent = serializedNodeWithId & {
+    childNodes: serializedNodeWithId[];
+  };
+  type WalkItem = {
+    node: Node;
+    parent: SerializedParent | null;
+    depth: number;
+    needsMask: boolean | undefined;
+    preserveWhiteSpace: boolean;
+  };
+
+  let root: serializedNodeWithId | null = null;
+  const stack: WalkItem[] = [
+    {
+      node: n,
+      parent: null,
+      depth: 0,
+      needsMask: undefined,
+      preserveWhiteSpace: preserveWhiteSpace ?? true,
+    },
+  ];
+  let sliceStart = performance.now();
+
+  while (stack.length > 0) {
+    if (performance.now() - sliceStart >= yieldBudgetMs) {
+      await doYield();
+      sliceStart = performance.now();
+    }
+    // LIFO pop with children pushed in reverse ⇒ same pre-order as recursion.
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    const item = stack.pop()!;
+    const { node, parent, depth } = item;
+
+    const sn = serializeNodeWithId(node, {
+      ...perNodeOptions,
+      needsMask: item.needsMask,
+      preserveWhiteSpace: item.preserveWhiteSpace,
+      depth,
+    });
+    if (!sn) {
+      // slimDOM-excluded / ignorable whitespace / max depth — the recursive
+      // path skips these nodes' children too.
+      continue;
+    }
+    if (parent) {
+      parent.childNodes.push(sn);
+    } else {
+      root = sn;
+    }
+
+    // ---- descend decision: mirrors serializeNodeWithId's recursive section ----
+    if (sn.type !== NodeType.Document && sn.type !== NodeType.Element) {
+      continue;
+    }
+    if (
+      sn.type === NodeType.Element &&
+      _isBlockedElement(node as HTMLElement, blockClass, blockSelector)
+    ) {
+      // blocked elements record a placeholder only; children get no ids
+      continue;
+    }
+
+    // children inherit needsMask exactly as the recursive path computes it
+    let childNeedsMask = item.needsMask;
+    if (!childNeedsMask) {
+      const checkAncestors = childNeedsMask === undefined;
+      childNeedsMask = needMaskingText(
+        node as Element,
+        maskTextClass,
+        maskTextSelector,
+        checkAncestors,
+      );
+    }
+    let childPreserveWhiteSpace = item.preserveWhiteSpace;
+    if (
+      slimDOMOptions.headWhitespace &&
+      sn.type === NodeType.Element &&
+      (sn as elementNode).tagName === 'head'
+    ) {
+      childPreserveWhiteSpace = false;
+    }
+
+    const serializedParent = sn as SerializedParent;
+    const pushChildren = (children: Node[]) => {
+      for (let i = children.length - 1; i >= 0; i--) {
+        stack.push({
+          node: children[i],
+          parent: serializedParent,
+          depth: depth + 1,
+          needsMask: childNeedsMask,
+          preserveWhiteSpace: childPreserveWhiteSpace,
+        });
+      }
+    };
+
+    // Shadow children are pushed first and light children second: light pops
+    // first and its whole subtree drains before shadow surfaces, appending
+    // shadow-serialized children after the light ones — the recursive order.
+    // (`isShadow` is self-derived inside serializeNodeWithId from parentNode.)
+    let shadowRootEl: ShadowRoot | null = null;
+    if (isElement(node) && (shadowRootEl = dom.shadowRoot(node))) {
+      pushChildren(Array.from(dom.childNodes(shadowRootEl)));
+    }
+    const skipLightChildren =
+      sn.type === NodeType.Element &&
+      (sn as elementNode).tagName === 'textarea' &&
+      (sn as elementNode).attributes.value !== undefined;
+    if (!skipLightChildren) {
+      pushChildren(Array.from(dom.childNodes(node)));
+    }
+  }
+
+  return root;
 }
 
 export function visitSnapshot(
--- a/packages/rrweb/rrweb/src/record/index.ts
+++ b/packages/rrweb/rrweb/src/record/index.ts
@@ -1,6 +1,8 @@
 import {
   snapshot,
+  snapshotWithBudget,
   type MaskInputOptions,
+  type serializedElementNodeWithId,
   slimDOMDefaults,
   createMirror,
 } from '@posthog/rrweb-snapshot';
@@ -90,6 +92,7 @@ function record<T = eventWithTime>(
     emit,
     checkoutEveryNms,
     checkoutEveryNth,
+    fullSnapshotYieldBudgetMs = 0,
     blockClass = 'rr-block',
     blockSelector = null,
     ignoreClass = 'rr-ignore',
@@ -199,6 +202,15 @@ function record<T = eventWithTime>(
 
   let lastFullSnapshotEvent: eventWithTime;
   let incrementalSnapshotCount = 0;
+  // Budgeted (time-sliced) full snapshot state. While a budgeted snapshot is
+  // in flight the mirror is mid-rebuild, so id-bearing incremental events
+  // must not reach the wire (see wrappedEmit); non-id events are queued and
+  // flushed after the FullSnapshot lands.
+  let budgetedSnapshotInFlight = false;
+  let budgetedSnapshotQueued: { isCheckout: boolean } | null = null;
+  const budgetedSnapshotEventQueue: Array<
+    [eventWithoutTime, boolean | undefined]
+  > = [];
   // Set per id — one iframe id can collect several cleanups across loads.
   const iframeObserverCleanups = new Map<number, Set<listenerHandler>>();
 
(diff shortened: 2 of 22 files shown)


marandaneto:
cc @PostHog/team-replay

marandaneto:
The underlying performance issue is real, and time-slicing full snapshots is a credible direction. The current synchronous path recursively walks the entire DOM without yielding; in a simple independent check, recording a roughly 100k-node DOM blocked synchronously for ~213 ms, so the larger timings reported in #4217 are plausible.

However, I don't think this implementation is safe to merge as-is because the budgeted path can lose replay state and user interactions.

### Initial snapshot mutation buffers are not locked

During initialization, the code does:

```ts
takeFullSnapshot()
handlers.push(observe(document))
```

With the synchronous snapshot this ordering is safe. With `snapshotWithBudget()`, `takeFullSnapshot()` returns after the first yield, and only then is the document observer and its mutation buffer installed. That newly created buffer was not included in the earlier `mutationBuffers.forEach((buf) => buf.lock())`, so it can process mutations while the snapshot is in flight. `wrappedEmit` then drops the resulting incremental event.

I reproduced this against the PR by waiting until an input and `<body>` had mirror IDs, then changing the input, dispatching input/click events, and appending a node while the initial snapshot was yielding. The emitted recording contained only Meta + FullSnapshot. The FullSnapshot retained the old input value and omitted the appended node, while both changes existed in the live DOM.

Shadow-root or iframe mutation buffers created during traversal appear vulnerable to the same lock timing problem.

### Incremental events are not generally self-healing

The blanket drop here is also unsafe:

```ts
if (e.type === EventType.IncrementalSnapshot) {
    return
}
```

Clicks and selections are not self-healing. Input and scroll observers update their deduplication state before reaching `wrappedEmit`, so an event dropped here may not be emitted again unless the value/position changes further. CSSOM, canvas, and media changes can likewise happen after the corresponding node was serialized and then be absent from both the FullSnapshot and incremental stream.

A synchronous snapshot prevents these event handlers from running until after serialization. Allowing them to run between slices and then dropping their output is therefore not equivalent to synchronous behavior.

### Scope

This is still a useful optimization for the full-snapshot portion of #4217, but it remains opt-in and does not address the reportedly more expensive large structural-mutation path or the later payload `JSON.stringify` stall.

Before merging, I think this needs:

- Observers and mutation buffers installed and locked before the first possible yield, including buffers created for shadow roots/iframes during traversal.
- A strategy to safely queue/rebase incremental events rather than dropping all of them.
- Recorder-level tests that mutate already-serialized nodes during yields: DOM add/remove, input, click, scroll, CSSOM/canvas/media, shadow DOM/iframes, and stopping recording while a snapshot is in flight.

So: this addresses a real issue, but the current async lifecycle introduces a blocking replay-correctness problem.

KarloAldrete:
Thanks for the feedback, I'm gonna look into that right away and test/benchmark everything with real cases so we can avoid mistakes on prod 🫡

KarloAldrete:
Thanks for the review — you were right on both counts, and re-reading the code confirmed it was worse than "not equivalent": the synchronous path *defers* those handlers (the event loop is blocked; they run after serialization against the finished mirror), it doesn't suppress them. Dropping was simply wrong, and the "self-healing" claim didn't survive contact with the observers: input and scroll update their dedup state *before* the callback, so a dropped event is never re-sent, and a `MutationBuffer` clears itself and drains `mapRemoves` into the mirror before invoking its callback — one dropped mutation event leaves the recorder's mirror permanently ahead of the replay for the rest of the session.

The PR has been reworked around your three asks:

**Buffers locked before the first possible yield, including mid-walk ones.** `initMutationObserver` now consults a gate that's armed synchronously before the walk's first yield, so the document's own buffer (installed right after `takeFullSnapshot()` returns) and any shadow-root/iframe buffer spawned by the traversal are born locked and released with the rest.

**Queue instead of drop.** Everything observed during the walk is held with its observation timestamp and delivered in order after the FullSnapshot (which is stamped at walk start — that's the state the tree describes, and it keeps the wire monotonic). The mirror reserves ids on demand during the walk so an event on a not-yet-serialized node resolves to the id that node is about to get; reservation is lazy, so id numbering is byte-identical to the sync path when nothing interleaves. Two classes are intentionally *not* delivered, for cause: events whose reserved id was never claimed (the node was created mid-walk in visited territory — its buffered add re-serializes final state on unlock, so the held event's content is already on the wire), and CSSOM/canvas deltas whose target the walk hadn't serialized yet (the snapshot captures the post-change state; delivering the delta too would apply it twice and shift every later rule index).

**Recorder-level tests that mutate during yields.** `test/budgeted-snapshot-convergence.test.ts` runs your full matrix — add/remove, input, click, scroll, CSSOM, canvas, media, shadow DOM, same-origin iframes, stop mid-flight — plus already-serialized-node mutations, mid-walk node creation, checkout coalescing, and stop+restart. Each scenario records with a 1ms budget, proves the ops landed while the snapshot was in flight, replays with the real `Replayer`, and requires the replayed DOM to equal the live DOM, with wire monotonicity and id-ledger integrity asserted throughout. Your exact repro is pinned as a test with the preconditions verified rather than assumed (it polls `mirror.hasNode` for the input and `<body>`, then proves the snapshot is still in flight before mutating) and asserts the inverse of each symptom you observed; against the version you reviewed, all three assertions fail by construction.

Getting the harness to zero also flushed out flaws beyond the two you found, all now fixed and pinned: stale captured references serializing ghost nodes the replay could never remove (`processRemoves` skips never-serialized nodes, so nothing cleans them up — references are revalidated at pop time now); the walker and the locked buffers both delivering a node added mid-walk to an unvisited parent (duplicate add that could resurrect a node past its later removal — the walk now tells buffers to forget adds it delivered itself); `isSerialized()` being `getId() !== -1`, which id reservation silently broke; flush re-entrancy where a `checkoutEveryNth/Nms` checkout tripped by a held event mid-flush started a second walk whose locks the ongoing flush then destroyed; and the generation counter living per `record()` closure, so a walk from an abandoned session kept writing into the next session's shared mirror.

One measurement note for the perf numbers now in the description: `scheduler.yield()` looked like the obvious yield primitive and benchmarked ~4% faster than a `MessageChannel` macrotask, but its prioritized continuations starved a host-page `setTimeout` loop for the entire walk (~1.5s on the 152k-node page). A recorder is a guest in the host app, so it yields through `MessageChannel`, and that decision is documented in the code. With the same-run A/B on the repro page: worst serialize stall 1,688ms → 76ms on a dev desktop, 7,789ms → 93ms under a 4x CPU throttle, snapshot on the wire byte-identical (152,036 nodes) in every run.

Honest scope notes: the payload-serialization stall you'd still see on this page (~0.9s desktop / ~3.6s throttled) exists in both arms — it's the second stall documented in the issue investigation and the natural follow-up, together with the giant-mutation-batch → checkout threshold. The scroll scenario asserts the recorder-side contract only, because the replayer's fast-forward applies element scroll offsets unfaithfully for the synchronous path too. And two narrow corners are deferred with code comments: fullscreen on a node inside a blocked subtree during the window, and cross-origin iframe messages arriving before their iframe is serialized (a pre-existing race the longer window widens).

marandaneto:
thanks @KarloAldrete 
@ksvat will take a look at it

arnohillen (commented):
Robot review:

Two things worth fixing before this lands, plus the red CI.

**1. The double-serialization guard misses text nodes, comments and blocked elements.** `serializedThisWalk.add(node)` sits at `snapshot.ts:1920`, after the non-element `continue` (1890) and the blocked-element `continue` (1897), so those three kinds are serialized but never enter the set and the guard at 1866 can't protect them. A text node reparented mid-walk into a container the walk hasn't reached is popped again, passes `isConnected` and the container check, and gets serialized a second time. `serializeNodeWithId` reuses the mirror id for an already-known node (1220), so the same id lands at two places in the FullSnapshot. The buffered remove plus add only reconciles the copy the replayer's mirror points at, since the later `add` wins, leaving the other as a ghost nothing cleans up. Moving the `add` to directly after `if (!sn) continue` covers all three. The convergence matrix reparents elements only.

**2. The mutation.ts reorder applies with the flag off, so "default is byte-for-byte" isn't quite true.** Checking `addedSet.has(n)` before the `!isSerialized` bail changes `droppedSet` membership and deep-deletes descendants for add-then-remove-before-serialization, which is a common React pattern on every page, budget 0 included. Worth gating on the budget or landing separately with its own test. The other two shared-code changes do look equivalent: meta ids are never -1 so `hasNode` matches `getId(n) !== -1`, and `timestamp ??=` only differs if a caller pre-stamps.

**3. Unit tests are red on a stale `@posthog/types` config snapshot** from the new option, not on anything real. `test:unit -- -u`.

Three smaller ones:

- The held-event queue is uncapped. With `recordCanvas` on, a multi-second walk holds every canvas mutation in memory and then flushes them together.
- If the walk throws, `.finally` still flushes held events and unlocks buffers although no FullSnapshot was emitted, so claimed ids go out referencing a snapshot the replayer never received. Clearing the queue on the failure path seems safer.
- An abandoned walk returns early without clearing `newBuffersStartLocked`, so buffers created before the next snapshot start locked. Self-heals at the next unlock, but the gate outliving its walk is easy to miss.

KarloAldrete:
Addressed the review as root-cause fixes and rebased onto current `main` (the PR is mergeable again):

- Moved the per-walk serialization guard immediately after successful serialization, covering elements, text, comments, and blocked elements. Added explicit reparenting regressions for all three missed cases.
- Preserved budget=0 behavior: add/remove reorder bookkeeping now only applies while a snapshot transaction owns the mutation buffers.
- Replaced the loose sliced lifecycle with an owned transaction (token + generation): buffers created mid-walk inherit the lock; success commits; null/throw/stop/restart/overflow discards partial state and performs a synchronous recovery snapshot. Stale transactions cannot unlock a newer recording.
- Bounded held events at 4,096 events / 16 MiB retained estimate. Overflow aborts the partial walk and recovers synchronously instead of retaining/flushing an unbounded burst. The estimator is cycle-safe, typed-array-aware, and avoids spreading huge arrays.
- Updated the stale `@posthog/types` config snapshot that caused the red Unit tests job.
- Kept current-main canvas masking and iframe listener cleanup options in the shared snapshot options during conflict resolution.

Local validation on the rebased head:
- rrweb build + TypeScript: pass
- config snapshot Jest test: pass
- budgeted walker tests: 6/6 pass
- convergence/replay matrix: 19/19 pass repeatedly, including injected throw, overflow recovery, stop/restart, iframe/shadow/CSSOM/canvas/media cases
- full rrweb suite: 462 pass, 3 skipped; one WebGL screenshot differed once by 1.47%, then passed immediately in isolation (environmental rendering flake)
- focused recovery paths: 10 consecutive runs pass
- rrweb-snapshot non-integration suite: 155 pass, 1 skipped

Large-DOM browser probe (125,008 serialized nodes, 7,567,227-byte tree, 3 runs/arm): budget=0 median 513 ms with ~516 ms timer blockage; 10 ms budget median 1,021 ms total with typical max timer gap ~30 ms (one 66 ms outlier). Node count and serialized bytes were identical across 0/10/25 ms.

The new GitHub Actions runs currently show `action_required` with zero jobs, so they need maintainer approval to start (fork workflow policy). Once approved, the previously-red Unit tests snapshot should be covered by the green local Jest result above.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
```diff
@@ -1588,6 +1598,383 @@ function snapshot(
...
+  const slimDOMOptions: SlimDOMOptions = slimDOMDefaults(slimDOM);
+  const yielder = yieldFn
+    ? { doYield: yieldFn, dispose: () => undefined }
+    : createYielder();
+  const doYield = yielder.doYield;
+  // A custom yieldFn means a test harness that wants deterministic slicing;
+  // input-driven early yields would make slice boundaries environmental.
+  const inputPending = yieldFn ? () => false : createInputPendingCheck();
+
+  const perNodeOptions = {
```
TueHaulund, line 1768: `canvasMaskingConfigured` isn't in the destructure above or in `perNodeOptions`, so it never reaches `serializeNodeWithId`. It typechecks because the options type is derived from `snapshot`'s, so the value is just dropped.

Result: with a yield budget set, a canvas the customer masked gets serialized into the FullSnapshot.

Worth a test with canvas masking on, asserting budgeted output matches synchronous output. Also, rebuilding the option bag by hand will keep losing options as new ones land.

TueHaulund (commented):
Three blockers from a review pass. Rest of the findings are non-blocking, happy to follow up separately.

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -418,97 +634,319 @@ function record<T = eventWithTime>(
...
+
+  takeFullSnapshot = (isCheckout = false) => {
+    if (!recordDOM) {
+      return;
+    }
+    if (fullSnapshotYieldBudgetMs > 0) {
+      takeFullSnapshotBudgeted(isCheckout);
+      return;
+    }
+    takeFullSnapshotSynchronous(isCheckout);
```
TueHaulund: The return value is dropped here.

`takeFullSnapshotSynchronous` now returns `false` and resets the mirror instead of throwing. So on failure nothing aborts: no FullSnapshot goes out, observers stay installed, every mutation resolves to `-1`. Records forever, replays as nothing.

This is the default path, so it hits everyone. The budgeted path already handles it by bumping the generation and calling `stopRecording`.

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -226,7 +372,77 @@ function record<T = eventWithTime>(
   };
   wrappedEmit = (r: eventWithoutTime, isCheckout?: boolean) => {
     const e = r as eventWithTime;
-    e.timestamp = nowTimestamp();
+    // Stamped once, when the event is observed. An event held back for a sliced
+    // full snapshot therefore keeps the time it actually happened rather than
+    // the time it was released, and a caller that already knows the time an
+    // event belongs to (the FullSnapshot below) can set it itself.
+    e.timestamp ??= nowTimestamp();
```
TueHaulund: `wrappedEmit` also handles cross-origin iframe events, and those already carry the child frame's timestamp (`iframe-manager.ts`, `transformCrossOriginEvent`). They used to get restamped with the parent clock; now they don't.

So this changes the default path, and clock skew between frames can make the stream non-monotonic, which is what the new convergence tests check for.

Better to skip restamping only for events the snapshot path builds, rather than for everything.

packages/rrweb/rrweb-snapshot/src/snapshot.ts
KarloAldrete: Fixed in 34b3016a. `snapshotWithBudget` now removes only its scheduler controls (`yieldBudgetMs`, `yieldFn`, `shouldAbort`) and spreads the remaining canonical snapshot option bag into the per-node serializer options before overriding normalized/internal fields. This passes `canvasMaskingConfigured` today and prevents future public serializer options from being silently dropped by another hand-maintained list. Added a real serializer regression: with `recordCanvas: true` and masking configured, synchronous and budgeted trees are deep-equal and neither contains `rr_dataURL`. The test failed before the fix because only the budgeted tree contained the unmasked data URL; the complete budget suite now passes 7/7.

packages/rrweb/rrweb/src/record/index.ts
KarloAldrete: Fixed in 34b3016a and strengthened in 9efcaf52. The synchronous return value is now checked. `stopRecording` is installed before `init()` can take its first snapshot, and a `false` result invokes the full teardown; `init()` then returns without installing observers. This also covers later synchronous checkouts after observers are already active. Two recorder-level browser regressions inject real serialization failures via `maskTextFn`: one during the initial snapshot and one during a later checkout. Both assert no additional events are emitted and `addCustomEvent` is rejected afterward. The integrated convergence matrix passes 21/21.

packages/rrweb/rrweb/src/record/index.ts
KarloAldrete: Fixed in 34b3016a. Ordinary `wrappedEmit` calls again unconditionally stamp with the parent recorder clock, restoring the pre-PR behavior for transformed cross-origin iframe events. Timestamp preservation is now an explicit internal opt-in used only for recorder-owned budgeted Meta/FullSnapshot events and for events released from the held queue. Added a real cross-origin regression that sends an rrweb Custom event through `postMessage` with a deliberately skewed child timestamp of `1`; it failed before the fix by emitting `1` and now emits a parent-clock timestamp. The full cross-origin + canvas recorder suites pass 27/27, and the 21-scenario convergence matrix retains monotonic wire ordering.

marandaneto:
sorry misclicked the draft button
@KarloAldrete fix pr conflicts and make sure your commits are verified, https://github.com/PostHog/posthog-js/pull/4233/commits otherwise we cant merge it (its not)

KarloAldrete:
Thanks again for the thorough review and for catching these issues. I’ve rebased the PR onto `main`, resolved the conflicts, and verified all 15 commits. The focused test suites are passing (`52/52`).

I’ll keep an eye on CI and stay available to address anything else immediately. Hopefully we’re now ready for a **sasquatch and merge** 🤣

ryanbreen:
We're a customer on the other side of the escalation behind this PR, and we appreciate you looking into it. Ours was the trace with the single ~3.9s recorder task (99.6% CPU in posthog-recorder.js, landing mid-scroll) that got root-caused to the synchronous full-DOM snapshot, and we currently have session replay disabled across all of our properties because of it. Since this is the code we'd be re-enabling on, I read the PR at head (b0b7dad) end to end rather than taking the description on faith.

First, credit where due: this got genuinely strong over the review rounds. The transactional redesign — token-owned buffer locks, the recording-generation counter, lazy id reservation with unclaimed-id scrubbing, cleanup on walk failure — is the right shape, and hold-and-deliver over dropping is the right call for exactly the reasons stated (I checked the unsafety analysis of dropping against the observer code and it holds: the sync path defers, it doesn't suppress). The convergence harness is the best part of the PR — replayed-DOM-equals-live-DOM through rrweb's own serializer, a wire-level unknown-id ledger, failure injection through real serialization paths — well above the bar I usually see for rrweb changes. The scheduler homework (MessageChannel over setTimeout for the nesting clamp, `scheduler.yield()` measured and rejected because "a recorder is a guest") is documented in-code where it belongs. And default-0 with byte-identical `array.js` means nobody pays until they opt in.

That said, I traced a handful of things I'd want fixed before this is the build we validate against. None of these duplicate the earlier review rounds as far as I can tell.

**blocker: `commitMutationBuffers` can strand buffers locked forever, and both recovery paths are dead code when it happens.** [observer.ts#L103-L105](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/observer.ts#L103-L105) nulls the global owner token *first*, then iterates the live `mutationBuffers` array, and each `buffer.commit(token)` synchronously re-enters application code (`emit()` → `mutationCb` → `wrappedEmit` → the consumer's `emit` callback, which posthog-js wires straight to `onRRwebEmit` with no try/catch and no `errorHandler` registered). Interrupt that loop and the remaining buffers hold a `lockToken` nobody owns: `reset()` never clears it ([mutation.ts#L301-L304](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/mutation.ts#L301-L304)), `emit()` early-returns on `lockToken !== null` ([#L336](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/mutation.ts#L336)), and every future `lockMutationBuffers` fails on the stale token. The recovery paths can't save you because they gate on the global token that line 103 already cleared: the sync path's catch calls `discardMutationBuffers(bufferToken)`, which returns immediately at [observer.ts#L114](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/observer.ts#L114) without touching a buffer, and `discardActiveMutationBufferTransaction` (used by `stopRecording`) is dead the same way. The budgeted flush at [index.ts#L939](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L939) has no catch at all, so a throwing consumer callback escapes as an unhandled rejection while recording continues — silently emitting zero mutations for the rest of the session. Separately, iterating the live array while commit re-enters splice-capable code (the detached-iframe cleanup sweep) can shift entries under the `for...of` cursor and skip a buffer outright: same terminal state, no exception required. Pre-PR this was `mutationBuffers.forEach((buf) => buf.lock())`/`.unlock()` with no tokens, so a missed buffer self-healed on the next snapshot; token ownership converts a transient hiccup into permanent, silent recording death. Fix is cheap:

```ts
for (const buffer of [...mutationBuffers]) {
  try {
    if (!buffer.commit(token)) committed = false;
  } catch (e) {
    committed = false; // record it, keep releasing the rest
  }
}
activeMutationBufferLockToken = null; // only after every buffer is released
```

plus the same array-snapshot treatment for `lockMutationBuffers`/`discardMutationBuffers`, and a discard path that can force-release by token even when the global owner is already null. Worth two tests: a throw from the emit callback partway through a multi-buffer commit (assert the next checkout still acquires the lock), and an iframe teardown mid-commit with a later-created buffer present (assert that buffer still emits afterward).

**blocker: the pending-id gate drops CSSOM and masked-canvas deltas the FullSnapshot demonstrably does not contain.** The gate at [index.ts#L406-L417](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L406-L417) discards pending-id StyleSheetRule/StyleDeclaration/CanvasMutation events on the in-code assumption that the snapshot will carry the post-change state. Two paths where that's false:

- `serializeTextNode` bails out of reading `sheet.cssRules` when the `<style>`'s text node has siblings ([snapshot.ts#L634](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L634)) and again on the `hasEmptyShorthandLonghand` fallback ([#L646](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L646) — any sheet with a shorthand-plus-`var()`, which describes most design-system CSS), keeping the original `textContent` verbatim. An `insertRule()` landing while that element's id is a pending reservation is dropped *and* absent from the snapshot: permanent divergence, and every later index-based delta then applies against a sheet one rule short — the exact desync the comment above the gate says it exists to prevent. The frontier test at [budgeted-snapshot-convergence.test.ts#L903](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/test/budgeted-snapshot-convergence.test.ts#L903) uses a single-text-node sheet with no `var()` shorthand — precisely the one case `stringifyStylesheet` does cover — so it green-lights the gate without touching either hole.
- The FPS canvas worker commits its dedup state unconditionally after a send ([image-bitmap-data-url-worker.ts#L160-L161](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/workers/image-bitmap-data-url-worker.ts#L160-L161)), so a frame dropped at the gate is one the worker believes was delivered — and since masked canvases carry no `rr_dataURL` in the snapshot ([snapshot.ts#L839](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L839)), a static masked canvas replays blank until the 30s keyframe, or forever on a shorter session. (The 34b3016a fix that made `canvasMaskingConfigured` actually reach the budgeted walk is what makes this path live.)

There's also a mirror-image hole: the gate keys off the `<style>` *element's* id, but the CSS text is produced when its *text child* is serialized — a separate stack item behind the every-16-nodes yield check — so in the window where the element id is already claimed, the delta gets queued instead of dropped, `stringifyStylesheet` then reads the live rules including the insert, and the queued delta re-applies it after the snapshot. Double application, same index shift, opposite direction. We run text masking everywhere and our pages are CSS-heavy, so this cluster is squarely our configuration. Suggested direction: stop treating "pending reservation" as proof of snapshot coverage — queue these like every other held event and reconcile at claim time against the representation the walk actually produced (a sheet serialized from `textContent` needs its delta replayed, not dropped; canvas dedup should only advance on confirmed main-thread delivery, or force a masked re-send at claim).

**major: Meta emission and the held-event flush sit outside the transaction's recovery boundary.** Three positions, one pattern. On the sync path, `emitMetaEvent` at [index.ts#L749](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L749) runs after the lock but before the `try`. On the budgeted path — the worst one — `emitMetaEvent` at [#L819](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L819) runs after `budgetedSnapshotInFlight = true` but before the promise chain exists, so a throw there latches the flag with no `.finally()` ever created to clear it: every subsequent event diverts into a queue that never flushes, every future `takeFullSnapshot` coalesces into a follow-up that never runs, and the SDK's `_tryRRWebMethod` swallows the periodic-snapshot failures — total, silent recording death with no rollback. And in the flush itself ([#L929-L945](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L929-L945)), a throw from `wrappedEmit` skips `commitMutationBuffers` entirely while the `finally` resets only the flags, not the lock. This is reachable in the shipping integration: `wrappedEmit` calls the consumer `emit` unguarded, and on the Meta branch `onRRwebEmit` runs customer-supplied URL-masking callbacks with no try/catch — a throwing masking callback on a mid-session checkout is all it takes. Fix: bring lock → Meta → walk → flush → commit → finalize under one guard with unconditional cleanup on both paths. Note the existing failure-injection tests all throw from *inside the walk* via `maskTextFn`, which the catches genuinely cover — none injects a throwing `emit`, which is the boundary that isn't covered.

**major: backdating the FullSnapshot to `startedAt` turns `checkoutEveryNms ≤ walk duration` into a self-sustaining snapshot loop.** The budgeted snapshot is stamped `transaction.startedAt` ([index.ts#L851-L863](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L851-L863)) and becomes `lastFullSnapshotEvent`, but the mutations released at commit go through `wrappedEmit` with the default `preserveTimestamp = false`, i.e. flush time. So the checkout test ([#L496-L503](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L496-L503)) sees the entire walk duration on the first flushed mutation. Any walk longer than `checkoutEveryNms` trips a checkout on its own flush, which coalesces (in-flight is still true) and fires with zero gap at [#L947-L951](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L947-L951) — and the next walk's flush trips again if a single mutation landed in its window. Steady state: 100% snapshot duty cycle, buffers permanently cycling through locked, everything routed through the held queue until it overflows into the synchronous fallback. Pre-PR this couldn't happen — the snapshot was stamped after the blocking serialize, so the delta to the following mutations was ~0. Blast-radius honesty: posthog-js drops `checkoutEveryNms/Nth` at its allowlist, so SDK users can't hit it, but `@posthog/rrweb` ships both knobs beside the new one, and 5-15s checkout intervals are the standard rrweb recipe for error-replay ring buffers — exactly the colliding config. Keep the backdated wire timestamp (the ordering argument in the code comment is right) and track a separate wall-clock `lastFullSnapshotAt`, set at flush completion, used only for the `exceedTime` arithmetic. Current coverage is `checkoutEveryNth: 3` with a finite click burst asserting `>= 2` snapshots — structurally unable to catch a runaway.

**major: canvas FPS frames bypass the lock, and the overflow recovery path is the exact freeze this PR exists to remove.** `worker.onmessage` → `mutationCb` ([canvas-manager.ts#L217](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts#L217)) has no `locked` guard — in contrast to `flushPendingCanvasMutationFor` right below it ([#L472](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts#L472)) — and the rAF capture loop runs during the walk. So base64 webp frames land in the held queue, and on a *checkout* walk the mirror isn't reset first, so every canvas already has a claimed id and the pending-id gate catches nothing: 100% of frames during the walk are held. The cap added after the earlier queue feedback bounds the memory, but look at what overflow actually does ([index.ts#L442-L453](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L442-L453)): abort the transaction, throw away up to 4,096 held events — clicks, inputs, an entire walk-window of interactions, none of them self-healing by this PR's own correct argument — then fall back to `takeFullSnapshotSynchronous` at [#L894](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L894): a full unbudgeted serialize of the same huge document, mid-session. I see from the design doc that overflow-as-controlled-failure is deliberate, and it's admittedly a tail case (long walk plus large or multiple animating canvases) — but from the reporting customer's seat, "the mitigation occasionally schedules the original bug" is a hard sell. Two changes make it mostly moot: gate the FPS `mutationCb` on `this.locked` and drop superseded frames (periodic canvas frames are the one genuinely self-healing event class — the next frame wins), which removes the main realistic overflow driver; and on overflow prefer one budgeted retry, with telemetry, before a synchronous stall.

**moderate: the 4,096-event / 16 MiB cap never sees the two largest retainers, so a busy walk can still end in one giant synchronous emit.** The cap only measures events that reach `transaction.eventQueue` ([index.ts#L443](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L443)). Mutations never do — buffers are locked for the whole walk and `emit()` hard-returns while locked — and the commit itself is exempt from the gate because `budgetedSnapshotFlushing` is set before `commitMutationBuffers` runs, so the entire accumulated batch serializes in one synchronous pass immediately after the FullSnapshot. Concretely: 700 elements with 30k-char text nodes appended across seven separate timer tasks during a walk coalesced into a *single* mutation event with 1,400 adds and ~21 MB of JSON. Command-mode canvas (`sampling: 'all'`) accumulates behind the same lock, also unaccounted. Nothing downstream rescues it — `splitBuffer` only splits `data.length > 1`, and the mutation throttler filters only `attributes`. Total serialization work matches what the sync path would have done, granted; but the coalescing merges batches the sync path would have emitted separately, and it lands as another long task on exactly the data-dense pages this targets. Worth accounting buffer/canvas retention as it accumulates and cancelling (or splitting and yielding the commit) before constructing the payload; the existing overflow test drives the cap with 4,100 custom events, which is itself evidence the cap never sees the dominant retainer.

A few things I'd rather ask than assert:

1. **Is `fullSnapshotYieldBudgetMs` a hard bound or a cooperative target?** The TSDoc ([types.ts#L50-L55](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/types.ts#L50-L55)) reads as a hard bound, but the deadline is only checked every 16 nodes between *complete* `serializeNodeWithId` calls ([snapshot.ts#L1849](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L1849)), and a single call can stringify an entire stylesheet, run `toDataURL()`, or — via the iframe block at [#L1348-L1396](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L1348-L1396), which sits outside the `skipChild` gate — serialize a whole loaded same-origin iframe document through the fully recursive path as one task. With `yieldBudgetMs: 1` and a page of ~1ms-cssText stylesheets, the first yield can land near 50ms. For our pages, stylesheet cost *is* the freeze, so this is the difference between "fixed" and "shortened". Related: #4320 instruments the synchronous `snapshot()` while this path calls `serializeNodeWithId` directly — will #4320's stylesheet budget and telemetry actually be live on the budgeted path, or does that need explicit wiring?
2. **The changeset still describes the abandoned round-one semantics.** [.changeset/full-snapshot-yield-budget.md](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/.changeset/full-snapshot-yield-budget.md#L9) says id-bearing incremental events arriving mid-snapshot are *dropped* — the opposite of the shipped hold-and-deliver design — and that text becomes release notes support will answer questions like ours from. Also, `posthog-js: patch` for a new public `SessionRecordingOptions` field while the siblings take minor — intended?
3. The comment at [index.ts#L200-L205](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb/src/record/index.ts#L200-L205) ("the only mutation events that can be held are attach-iframe payloads") looks false under `recordCrossOriginIframes`: transformed child-frame mutations reach `wrappedEmit` directly with remapped texts/attributes/removes/adds, never passing a parent-side locked buffer — so a scrubbed `attachIframe` can be followed by delivered held events referencing ids the wire never introduced. Off by default and outside the posthog-js allowlist, so low stakes, but either extend the scrub across the full payload shape or fix the comment before someone builds on it.
4. Every budget test injects a custom `yieldFn`, and the browser harness undefines `MessageChannel` to force the setTimeout fallback — so the production `createYielder` branch ([snapshot.ts#L1621](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L1621)) is the one path every real user runs and no test executes. And nothing bounds a walk's wall clock: a host page that swaps in a non-delivering `MessageChannel` (the test suite itself demonstrates pages replace this global) leaves the walk suspended forever, buffers locked, no snapshot ever emitted. One integration run on the real yielder plus a wall-clock watchdog that aborts into recovery after N seconds?
5. Smallest one: the four working-notes docs under [docs/plans/](https://github.com/PostHog/posthog-js/blob/b0b7dadf98ec23dd7d363e37c2f37e461516733a/docs/plans/2026-07-29-budgeted-snapshot-review-fixes.md) (~650 lines, including process instructions like "Do not push or comment until explicitly requested") — deliberate? Your AI_POLICY.md points at the PR description's Agent context section as the home for that material; as PR narrative it's valuable, as permanent repo content it's noise.

Last, the ask that matters most from our side: a build we can test. We have a deterministic repro — live pages where the freeze lands mid-scroll on a trace — and replay stays off across our fleet until we can prove this out. Concretely: a pre-release (canary dist-tag is fine) with this plus #4320 once the blockers land; the config you'd recommend for us, in particular what budget value you consider sane for stylesheet-heavy pages given the granularity question above; and whether #4320's snapshot-duration telemetry will let us confirm in production that `maskAllText: true` stays the minor contributor you estimated, so we don't have to trade masking away on faith. Happy to validate against the real pages before this goes wide — we're arguably the ideal test case, since we're the reason it exists.

KarloAldrete:
I just saw those checks failing, give me a moment to clean them up and you'll have it asap 🙂‍↕️

arnohillen:
# Deep review at head `8cf3bbe6`

I did an independent deep review of this PR at head `8cf3bbe6` (post-rebase onto `990ee9ba3`). Method: every claim in ryanbreen's review was re-verified against the current head; a separate fresh-eyes pass ran across seven lenses (record-side state machine, snapshot walker, replay semantics, posthog-js integration, tests, benchmarks, API/release); the new test suites were executed locally (snapshot suite 211/211 incl. the 7 budget tests; convergence 21/21; record suites 81/81 — plus one failure detailed below). The review was agent-assisted with adversarial cross-checking; several findings were confirmed by building the dists and reproducing live against the real `Replayer` under puppeteer, and everything below was individually verified against the code before inclusion.

## Verdict

**Not mergeable as-is, and — more importantly — the PR's central safety claim ("default = today's behavior, byte-for-byte") is false at head.** All seven of ryanbreen's findings are confirmed against the current head. On top of them, this review found **two new blockers** (one a regression of a fix merged to main ~15h before the rebase, provable by running main's own test on the branch; one a live-reproduced permanent subtree loss in replay) and **five new majors**. The engineering direction is credible and much of the work is genuinely strong — but the test oracle is structurally blind to a whole bug class, the production yield path runs in zero tests, and none of the performance numbers are reproducible from the repo.

---

## Part 1 — ryanbreen's review, verified claim by claim: 7/7 confirmed

Reality is worse than stated in three places (default-path commit teardown; unsplittable coalesced mutation; allowlisted `recordCrossOriginIframes`) and two of the proposed fixes need amending (blocker 1 token ordering; major 2 marker timing). Take that review at full weight.

<details>
<summary><b>blocker 1 — <code>commitMutationBuffers</code> strands buffers forever: CONFIRMED, and worse than stated</b></summary>

- `observer.ts:99-111` nulls `activeMutationBufferLockToken` *before* the un-guarded `for...of` over the live `mutationBuffers` array. `buffer.commit()` synchronously re-enters consumer code (`mutation.ts:272-280` → `emit()` → `mutationCb` → `wrappedEmit` → `emit?.()` at `index.ts:471-472`, unguarded) and posthog-js's `onRRwebEmit` has no top-level try/catch and no `errorHandler` registered (it's not even in the options allowlist).
- Both recovery paths are provably dead once the token is nulled: `discardMutationBuffers` returns immediately (`observer.ts:113-125`), `discardActiveMutationBufferTransaction` no-ops. `reset()` never clears `lockToken`; `emit()` early-returns on it; every future `lockMutationBuffers` fails. Pre-PR the tokenless `forEach(lock)/forEach(unlock)` self-healed on the next snapshot.
- The splice-under-cursor mechanism is real too: the commit re-enters `runAndDetachIframeCleanup`/`removeIframeById` (`index.ts:532-541`) which reaches `findAndRemoveIframeBuffer` → `mutationBuffers.splice` (`observer.ts:488`); `shadow-dom-manager.ts:82` is a second splice site.
- **Aggravation:** the tokenized commit also runs on the **default sync path** (budget=0). A throw during commit reaches the catch at `index.ts:773` whose discard is a no-op, `takeFullSnapshot` returns false, and `stopRecording?.()` fires (`index.ts:963-966`) — silent recording death for *all* users, where main self-healed.
- **The proposed fix needs one amendment:** with the token nulled only *after* the loop, a buffer born from commit re-entry (iframe attached while a committed mutation's adds serialize) is locked by `initMutationObserver` (`observer.ts:164-167`) with a token the snapshotted `[...mutationBuffers]` loop never commits. The fix is only sufficient together with the force-release-by-token discard also asked for, or a post-loop sweep releasing any buffer whose `lockToken` still equals the token.
</details>

<details>
<summary><b>blocker 2 — pending-id gate drops CSSOM/masked-canvas deltas the snapshot doesn't contain: CONFIRMED on every sub-path</b></summary>

- `serializeTextNode` keeps raw `textContent` both when the `<style>`'s text node has siblings (`snapshot.ts:634`) and on the `hasEmptyShorthandLonghand` fallback (`snapshot.ts:646`, triggered by any shorthand-plus-`var()` — most design-system CSS). A gate-dropped `insertRule` is then permanently absent, and every later index-based delta misaligns.
- Masked canvases carry no `rr_dataURL` (`snapshot.ts:839`) while the FPS worker advances its dedup unconditionally after send — a gate-dropped frame is never re-sent.
- The double-apply mirror-image window is real: the gate keys off the *element's* id (claimed at `utils.ts:322-327` one walker iteration before the *text child* is serialized), so a delta landing on that slice boundary is held and re-applied over a snapshot that already inlined it.
- The frontier CSSOM test (`budgeted-snapshot-convergence.test.ts:904-927`) covers only the single-text-child, no-`var()` case — exactly the one case where dropping is correct.
- **Two aggravations:** unmasked static WebGL canvases are silenced forever (no masked keyframe without maskRegions); and the 30s dedup-reset keyframe cited as the recovery path is itself broken at head, because the rebase dropped `canvasManager.onFullSnapshot()` (see N1 below).
</details>

<details>
<summary><b>major 1 — Meta emission and the flush sit outside the recovery boundary: CONFIRMED (all five sub-claims)</b></summary>

Sync path: lock at `index.ts:745`, Meta at `749`, `try` at `751` — a Meta-emit throw leaks the lock, and this is a *regression vs main*, where Meta was emitted before the lock; it affects the default path. Budgeted path: flag set at `802`, Meta at `819`, promise chain (and its `.finally`) only exists from `833` — a throw latches `budgetedSnapshotInFlight` forever; every event diverts into a queue that never flushes; `_tryRRWebMethod` swallows the periodic-snapshot failures. In the flush, a throw at `935` skips `commitMutationBuffers` at `939`; the `finally` resets only flags. The unguarded chain to customer URL-masking callbacks on the Meta branch is real. All failure-injection tests throw from inside the walk via `maskTextFn`; none injects a throwing emit. One mitigation the review omits: a stop/restart heals the latch. The proposed single-guard fix is right.
</details>

<details>
<summary><b>major 2 — backdated FullSnapshot + <code>checkoutEveryNms</code> → self-sustaining snapshot loop: CONFIRMED, but the proposed fix doesn't break the loop</b></summary>

Mechanism verified end-to-end (`index.ts:806/851-863` stamping, `483-485` `lastFullSnapshotEvent`, `496-503` exceedTime on flush-time mutation stamps, `795-801` coalescing, `947-951` zero-gap follow-up). posthog-js's allowlist drops the checkout knobs so SDK users can't hit it, but `@posthog/rrweb` ships them. The existing test (`checkoutEveryNth: 3`, `>= 2` snapshots) is structurally blind to a runaway. **Correction:** "`lastFullSnapshotAt` set at flush *completion*" doesn't help — a coalesced walk starts in the same task the flush completes, so W > Nms still trips. The wall-clock marker must be captured at the current snapshot's emission/flush *start*.
</details>

<details>
<summary><b>major 3 — canvas FPS frames bypass the lock; overflow recovery is the original freeze: CONFIRMED</b></summary>

`worker.onmessage` → `mutationCb` (`canvas-manager.ts:230-272`) has no `locked` guard (unlike `flushPendingCanvasMutationFor` at `485-488`); the rAF loop runs between slices; on checkout walks the mirror isn't reset so ids are claimed and the gate passes 100% of frames into the held queue. Realistic overflow estimate: the 16MiB byte cap (base64 charged at 2 bytes/char) trips in roughly 2–8s of walk with 3–4 modest or one full-screen canvas at the default 4fps — squarely the feature's target population. One overstatement: the sync-fallback snapshot does re-serialize final DOM/input/scroll state, so the permanent loss is the click/selection/CSSOM-delta class, not literally everything. The proposed fixes (gate FPS frames on `locked` + drop superseded frames; one budgeted retry with telemetry before a sync stall) are the right shape.
</details>

<details>
<summary><b>moderate — the 4,096 / 16 MiB cap never sees the two largest retainers: CONFIRMED, and worse</b></summary>

Locked `MutationBuffer` accumulation never reaches the cap (`emit()` hard-returns while locked); the commit is exempt (`budgetedSnapshotFlushing` set at `929` before commit at `939`); cross-task coalescing into one giant mutation event is real (`mutation.ts:272-279, 520-590`). **Aggravation:** the coalesced event is unsplittable downstream (`splitBuffer` only splits `data.length > 1`; the throttler filters only attributes) — the sync path's separate events *were* splittable, so coalescing also creates an ingest/size-limit risk beyond the long task.
</details>

<details>
<summary><b>question cluster — all five valid</b></summary>

1. The budget is provably cooperative, not hard (16-node probe stride at `snapshot.ts:1849`; unbudgeted stylesheet stringify / `toDataURL` / whole same-origin-iframe recursive serialization inside one call), while both TSDocs say "Maximum milliseconds". And **#4320's stylesheet budget + telemetry arm only inside `snapshot()` — they are dead on this PR's direct-`serializeNodeWithId` path without explicit wiring.** For a stylesheet-heavy page, that's the difference between "fixed" and "shortened".
2. The changeset still describes the abandoned round-1 *drop* semantics — the opposite of the shipped design — and it violates the repo's changeset convention outright: AGENTS.md:189 says never to add changesets for `packages/rrweb/` modules (rrweb changes release via a `posthog-js` changeset only), yet `.changeset/full-snapshot-yield-budget.md` lists three of them.
3. The `index.ts:203-207` comment is false under `recordCrossOriginIframes` — and the review *understates* it: `recordCrossOriginIframes` **is** in the posthog-js allowlist, so this hole is reachable by SDK users (see N5 below).
4. Confirmed: every real-browser budget scenario undefines `MessageChannel`; the production yielder branch runs in no browser test (one Node-incidental unit test only); nothing bounds a stalled walk's wall clock.
5. AI_POLICY.md points AI-process material at the PR description's Agent context section; the four `docs/plans/*.md` files (~650 lines, including agent process instructions) shouldn't ship in the repo.
</details>

---

## Part 2 — New findings (not in any prior review round)

### N1 — blocker: the rebase dropped main's #4338 fix; main's own test fails on this branch

Main commit `c458807d3` (merged ~15h before the rebase) added `canvasManager.onFullSnapshot()` after the buffer unlock (main `record/index.ts:501`) so idle canvases re-emit one frame per snapshot epoch. On the branch the call is gone from *all three* snapshot completion paths — `grep onFullSnapshot packages/rrweb/rrweb/src/record/index.ts` returns nothing — while the method and its unit tests survived, so grep looks healthy. **Verified by running `test/record/full-snapshot-canvas-masking.test.ts` on the branch: 1 failed / 3 passed** ("resets canvas frame dedup on each full snapshot so idle canvases re-emit" — `postMessage` never called with `{resetFrameDedup: true}`). This regresses the default path — the rebase note about keeping current-main canvas masking during conflict resolution covered the shared snapshot *options*, but this call was lost — and means the "400 passed / no regressions" result predates the rebase and needs re-validating.

### N2 — blocker: moving a serialized ancestor mid-walk permanently loses its unserialized children in replay (live-reproduced)

`onSerialize` → `forgetAddedNode` (`record/index.ts:664-673`, `mutation.ts:268-270`) assumes a node the walk serialized has superseded its pending add. When an *ancestor* P (already serialized) is moved during the walk, P lands in `movedSet` with a pushed remove, while P's not-yet-serialized children land in `addedSet`; the walker then serializes those children into the FullSnapshot (pop-time revalidation passes — container is still P) and *forgets their adds*. Commit emits `remove(P)` + `add(P, skipChild: true)` with no child adds: the replayer drops the subtree and re-creates P childless, while the recorder's mirror keeps the children — every later event on them dangles forever. Reproduced with this PR's own harness pattern (built dists, puppeteer, replay-vs-live diff): 800-span div moved mid-walk → replay loses the children permanently.

<details>
<summary><b>N3 — major: same-parent reorders during the walk are never reconciled (live-reproduced, two variants)</b></summary>

Pop-time revalidation checks only disconnection and container *identity* (`snapshot.ts:1879-1895`) — a node repositioned within the same parent is serialized at its stale captured index, and the corrective pending add is deleted by `forgetAddedNode`. The removal record bails (`mutation.ts:805-810`, node not serialized-by-membership), so zero events describe the move. Replay shows the wrong sibling order until the next full snapshot — for sorted lists/tab strips/leaderboards, silently wrong content. Reproduced live (`insertBefore(lastChild, firstChild)` mid-walk; divergence permanent).
</details>

<details>
<summary><b>N4 — major: scrubbed <code>AdoptedStyleSheet</code> events are never re-sent — shadow-root styling permanently missing (live-reproduced)</b></summary>

The scrub's "the buffered add re-serializes final state" premise is false for adopted stylesheets, which aren't part of node serialization. Host created in visited territory mid-walk → `addShadowRoot` registers the root in the `shadowDoms` WeakSet and emits `adoptStyleSheets` on `setTimeout(0)` with the host's *reserved* id → scrubbed at flush. On unlock, `pushAdd`'s `onSerialize` calls `addShadowRoot` again but the WeakSet dedup returns early; the sheet is in `styleMirror`, so future adoptions of the same sheet emit a bare `styleId` with no rules. Standard Lit/FAST render-during-load pattern → unstyled shadow content for the rest of the recording. Reproduced live (zero source-15 events on the wire).
</details>

<details>
<summary><b>N5 — major: cross-origin iframe recording breaks three ways during a walk (and it IS allowlisted in posthog-js)</b></summary>

(a) Held child-frame events flush *before* `iframeManager.reattachIframes()` (flush order: FullSnapshot → held events → commit → `finishFullSnapshot` → reattach, `index.ts:931-940, 727-730`), so the replayer drops every one of them ("node not found") and child-side dedup never re-sends — typed values lost until the field changes again. On the sync path the same postMessages queue as macrotasks *after* reattach and apply cleanly — this is new. (b) A child FullSnapshot-attach landing under an unclaimed reserved parent id is scrubbed with no recovery — child recording permanently lost. (c) On overflow abort, the fallback's `mirror.reset()` re-keys every node and orphans `iframeManager.attachedIframes` — previously attached iframes render empty from then on.
</details>

### N6 — major: the default sync path now silently tears down recording on snapshot failure (backwards-compat break)

On main, a throw inside `takeFullSnapshot` propagated to posthog-js, whose `_tryRRWebMethod` catches and retries within 2s. At head, `takeFullSnapshotSynchronous` catches everything (`index.ts:773-779`), returns false, and `takeFullSnapshot` calls `stopRecording?.()` (`963-966`). No error reaches the SDK; `$sdk_debug_replay_rrweb_error` never set; status stays "active" while rrweb is dead. One transient `onRRwebEmit` throw (e.g. a localStorage quota error) during a routine 5-minute snapshot permanently truncates the session. This alone falsifies "default = today's behavior."

### N7 — major: a canary would be blind — zero production observability, no remote lever

All six degraded outcomes (budgeted-walk failure, null result, lock contention, sync-null, sync-throw, terminal `stopRecording`) surface only as `console.warn` or nothing; none route through rrweb's registered `errorHandler` or posthog-js's `_tryAddCustomEvent` channel (used for "recording paused" etc.). The held-queue overflow — "the feature gave up and re-ran the original stall" — produces zero ingested signal. And `fullSnapshotYieldBudgetMs` is client-init-only: remote config drives sampling/masking/canvas but not this option, so there is no kill switch and rollback requires a customer redeploy. Sibling PR #4320 ships telemetry for its budget; this PR — built to be canaried — ships none.

<details>
<summary><b>Moderates worth listing</b></summary>

- **Unload mid-walk loses everything**: Meta is on the wire but the FullSnapshot and all held events die with the page; short visits to exactly the target pages produce empty replays where sync mode captured a snapshot. No pagehide flush/abort exists.
- **No option validation**: `true` and `Infinity` pass the `> 0` gate (`true` coerces to a 1ms budget → minutes-long walks → overflow → the original freeze).
- **Test-harness blind spots** (see Part 4): duplicate-delivery blindness, timing-based frontier preconditions, overflow silence enshrined, timestamp preservation unasserted, zero SDK plumbing coverage.
- **Scrub losses beyond N4**: Selection `ranges` scrub has zero coverage and the id-ledger never reads `ranges`; Focus events on autofocused mid-walk modals are scrubbed and not reconstructable from the re-serialized add; fullscreen Custom events with reserved `payload.id` are delivered with an id the wire never introduced.
- **Unexamined interleavings**: a background tab frozen mid-walk pins buffer locks indefinitely (MessageChannel macrotask never fires; no watchdog); dual-recorder collision emits an orphan Meta before throwing; quirks-mode/XML documents have zero walker coverage; posthog-js's own cadence (5-min default, **1-min while triggers pending**) can approach a high walk duty cycle on huge DOMs even without checkout knobs.
</details>

---

<details>
<summary><b>Part 3 — Interactions with #4217, #4226, #4320</b></summary>

**#4217 (issue)** — the origin. The investigation comment there is high quality: reproducible numbers (linear in node count; serialization ≈98.5% of cost vs ~25ms for atomic id-assignment), the discovery of the *second* stall (~641ms payload `splitBuffer`+`JSON.stringify`), and a two-tier proposal. This PR is Tier 1. Worth internalizing: **that root-cause framing ("only id-assignment must be atomic with the live DOM") argues for Tier 2** (atomic capture + off-thread serialization, prototyped at 174ms blocked vs 1,723ms with zero long tasks) — where the entire transactional protocol this PR builds would be unnecessary. See Part 5.

**#4226 (merged)** — the player-side `seekYieldBudgetMs`. Precedent for the opt-in-yield-budget pattern and for the review dynamics: review there caught a boundary double-apply on the default path plus three minors, all fixed. Relevant here: hoisting a shared yielder was deferred "once #4233 lands" — so #4233's `createYielder` is expected to become shared infrastructure; its zero-browser-test status matters beyond this PR.

**#4320 (open)** — stylesheet-inlining budget + snapshot-cost telemetry. Two interactions matter:

1. **Its budget and telemetry are dead on #4233's budgeted path** (verified from #4320's diff: both arm only inside `snapshot()`, and #4233 calls `serializeNodeWithId` directly). For a customer whose freeze *is* stylesheet cost, #4233 alone shortens but does not bound the stall, and #4320 alone doesn't touch the budgeted path. If both merge, the combination needs explicit wiring plus an integration test; nobody owns that today.
2. **ryanbreen's #4320 review is prior art for this PR's own weak spots**: the #4320 blockers — deferral relocates rather than bounds a monolithic task; a default that ships in `recorder.js` while its off-switch ships in `array.js` — are the same two failure classes to re-check here. On the skew question #4233 is actually safe (default 0 lives in `record()`; old array.js simply never passes the option, so the feature can't engage during skew — the failure is only that an opt-in silently doesn't work, which the missing SDK plumbing test would let regress). On "longest single task is the metric," the #4320 point applies verbatim to #4233's unbudgeted per-node work and to the end-of-walk flush/commit stall.
</details>

---

## Part 4 — Testing and benchmarking: what clears the bar, and what doesn't

**What genuinely clears a high bar:** the convergence harness architecture (replayed-DOM ≡ live-DOM through rrweb's own serializer, wire-level unknown-id ledger, failure injection through real serialization paths, the reviewer's repro pinned with *verified* preconditions); byte-level equivalence tests for the no-interleaving case including absolute ids; documented scheduler homework (MessageChannel vs `scheduler.yield()`, measured and rejected for guest-code reasons); honest scope notes. All suites pass when run here (211/211, 21/21, 81/81). ryanbreen's "well above the bar I usually see for rrweb changes" is accurate.

<details>
<summary><b>Where "perfectly tested" fails</b></summary>

1. The oracle is final-state convergence + `>=`/`some()` counts — **structurally blind to duplicate delivery and idempotent double-application** (a flush that delivers every held event twice passes all 21 scenarios; the replayer tolerates duplicate adds by design). Exact-count assertions on the discrete dispatched events would close this cheaply.
2. **Main's own canvas test fails on the branch** (N1) — the "400 passed, no regressions" claim needs re-validating post-rebase.
3. The **production yield path runs in zero browser tests** (every scenario undefines `MessageChannel`), and nothing bounds a stalled walk.
4. Frontier preconditions are a bare `setTimeout(60)` — timing-based, unverified, and quietly vacuous if serialization gets faster.
5. Zero coverage of: the SDK plumbing (a renamed allowlist key would silently disable the feature for the exact customer validating it), throwing consumer emits (the boundary major 1/blocker 1 live on), runaway checkouts, Selection-ranges scrub, observation-timestamp preservation, non-Chromium browsers, quirks-mode/XML documents.
</details>

<details>
<summary><b>Where "perfectly benchmarked" fails</b></summary>

Every performance number lives in PR prose; the harness is on a fork, not the repo; no CI timing assertion exists (even a loose "no slice > 4× budget on the committed fixture"); the design doc's own acceptance metrics (slice p95, queue high-water, forced-failure recovery cost) were never executed or reported. The headline table has an unresolved arithmetic tension (35 slices × 25ms ≈ 875ms only reconciles with the 1,688ms baseline if the baseline includes the ~0.9s stringify stall the PR says is out of scope). Not measured at all: heap/GC of the walk window (the design's new cost), input latency during the walk, the held-queue estimator's accuracy, a stylesheet-heavy fixture (the reporting customer's actual page class), low-end hardware rather than DevTools throttle, and a soak run counting overflow→sync-fallback frequency.

Evidence worth adding: a committed runnable harness with per-slice p50/p95/max at 5/10/25ms on table-dense *and* stylesheet-heavy fixtures, whole-arm long-task traces OFF vs ON, heap/GC/queue high-water on an active page, input-latency during the walk on Chrome and one non-`isInputPending` browser, one real low-end device, and a soak with overflow counters.
</details>

---

## Part 5 — The strategic question

Four review rounds keep finding new interleavings, and they cluster in one place: the walk window converts rrweb's core atomicity assumption ("nothing runs during a snapshot") into a distributed transaction that **every** observer and manager must now participate in correctly — mutation buffers, canvas manager, stylesheet manager, shadow-dom manager, both iframe managers, the scrub/reconcile logic. Each confirmed bug is precisely a manager that wasn't taught the protocol (FPS canvas, adopted stylesheets, cross-origin iframes, reattach ordering, `onFullSnapshot`). That's not bad luck; it's the design's failure mode, and it will keep taxing every future observer anyone adds.

The #4217 investigation demonstrates the alternative boundary: atomic capture is cheap (~25ms for 152k nodes); serialization doesn't need the live DOM. Tier 2 (capture + off-thread serialize, prototyped at 174ms blocked / zero long tasks) makes the entire hold-and-deliver protocol unnecessary. **Before investing another fix round in Tier 1's protocol, it's worth deciding whether Tier 1 is a stopgap worth its permanent complexity (~1,100 lines of invariant-heavy core shipped unconditionally to every recorder user, no size gate on the browser package) or whether the effort should go to Tier 2 with a much smaller correctness surface.** If Tier 1 proceeds, the merge gate below applies.

## Part 6 — Suggested merge gate and canary plan

**Must fix before merge (default-path regressions first):**

1. Restore `canvasManager.onFullSnapshot()` in all snapshot completion paths (the failing test already exists on the branch) and re-run the full suite post-rebase (N1).
2. Revert the sync-path error contract: failure must reach posthog-js's retry, not `stopRecording()` (N6); move Meta emission back inside the guarded region on both paths (major 1).
3. Blocker 1 stranding fix: array-snapshot + per-buffer try/catch + force-release-by-token discard + post-loop sweep for buffers born mid-commit; tests for throwing emit mid-commit and iframe teardown mid-commit.
4. Blocker 2: stop treating pending reservation as proof of snapshot coverage — queue and reconcile at claim time against what the walk actually produced; canvas dedup advances only on confirmed delivery.
5. N2/N3: revalidate *position*, not just container, at pop — or don't forget adds whose buffered record carries reconciling information an emitted remove will need.
6. N4/N5 + scrub classes: `AdoptedStyleSheet`/attach-iframe/Selection/Focus/fullscreen must be delivered-at-claim or re-emitted, never silently scrubbed; flush held cross-origin events after `reattachIframes`.
7. Major 2 with the corrected marker (wall-clock captured at flush *start*); a real runaway-checkout test.
8. Major 3: gate FPS frames on `locked`, drop superseded frames per canvas, one budgeted retry before any sync fallback.
9. Lifecycle: pagehide/frozen-tab watchdog that aborts the walk and synchronously flushes held events; overflow and every fallback path emit an ingestable signal (align with #4320's telemetry); clamp/validate the option value.
10. Release hygiene: rewrite the changeset (the drop-semantics text is stale, and per AGENTS.md:189 it must not list `packages/rrweb/` packages at all — `posthog-js` changeset only), delete `docs/plans/*` per AI_POLICY.md, commit the benchmark harness + one CI slice assertion, add exact-count and duplicate-identity assertions to the convergence suite, add one real-`MessageChannel` browser run and one SDK plumbing test.

**Canary sequence** (the ask in the thread is reasonable and the reporting customer is the ideal test case): (1) fix the gate above; (2) dogfood on PostHog's own large-DOM surfaces with the new fallback/overflow telemetry at zero and replay-fidelity spot checks; (3) then a canary dist-tag at low `sampleRate` with a jointly agreed 10–25ms budget, the client-config kill switch pre-agreed, and longest-single-task as the shared acceptance metric; (4) be explicit that until #4320's stylesheet budget is wired into the budgeted path, stylesheet-heavy pages get "shortened", not "bounded" — that's the thread's number-one open question and it deserves a straight answer.

---
*Reviewed: head `8cf3bbe6`; all PR comments; issue #4217 + investigation; PR #4226 (merged) + review thread; PR #4320 + its review; local build & test execution; live puppeteer repros for N2/N3/N4. Agent-assisted review with adversarial cross-checking of every finding.*

marandaneto:
#4384

PR #4236, opened 2026-07-23
fix(ai): preserve Anthropic cache-write TTL breakdowns

gouveags:
## Problem

`@posthog/ai` can drop Anthropic's 5-minute and 1-hour cache-write breakdown before capturing an event.

The Anthropic streaming wrapper replaces the usage captured from `message_start` with the final usage delta, which may contain only output tokens. The LangChain callback also preserves the aggregate cache-write count but misses the TTL breakdown from both direct Anthropic responses and Bedrock Converse responses.

Bedrock Converse retains this data under `response_metadata.usage.cacheDetails` for non-streaming calls and `response_metadata.metadata.usage.cacheDetails` after streaming, but the PostHog callback did not read either shape.

Without those fields, ingestion cannot apply the TTL-specific pricing added in PostHog/posthog#71323.

## Changes

- Merge Anthropic streaming usage updates so fields from `message_start` survive later deltas that omit them.
- Read cache-write TTLs from direct Anthropic and Bedrock Converse LangChain metadata, including the aggregated streaming shape.
- Validate and sum repeated 5-minute and 1-hour entries before emitting both TTL counts alongside the existing aggregate.
- Keep the aggregate-only fallback when the breakdown is incomplete, invalid, or does not match the aggregate.
- Treat empty usage containers as missing while preserving explicit zero-valued counts.
- Preserve existing usage-source precedence and avoid subtracting Anthropic cache tokens twice.
- Add regression coverage for Anthropic streaming, direct LangChain usage, both Bedrock Converse shapes, TTL edge cases, aggregate-only payloads, and non-Anthropic precedence.

The Vercel integration already preserves raw usage and is unchanged.

The regression tests consume the mocked Anthropic stream through the wrapper and pass direct Anthropic and Bedrock Converse `AIMessage` usage shapes through the LangChain callback to the captured `$ai_generation` event.

- Focused LangChain callback suite — 31 passed
- `pnpm --filter @posthog/ai test:unit` — 517 passed, 84 skipped
- `pnpm --filter @posthog/ai build`
- `pnpm check:public-api`
- `pnpm build`
- `pnpm lint`
- `pnpm lint:playground`
- `pnpm test:functional`

The root `pnpm test:unit` run hit two unrelated five-second timeouts in `posthog-node` under parallel load. Both affected suites passed together in isolation: 31/31 tests.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Related work

- PostHog/posthog#71306 — original cache-write pricing report
- PostHog/posthog#71323 — ingestion support for TTL-specific pricing
- PostHog/posthog-python#745 — equivalent Python SDK gap
- PostHog/posthog-python#746 — companion Python SDK fix

## Deployment note

The new properties are optional, and the existing aggregate cache-write count remains available. The `@posthog/ai` release should follow PostHog/posthog#71323 so ingestion can price the TTL breakdown from the first event that includes it.

## 🤖 Agent context

Gabriel directed the change. I used Codex to inspect the provider payloads, implement the fix, review regression risks, and run validation; no public session link is available. We kept the change inside the existing Anthropic and LangChain integrations: Vercel is unchanged, and pricing remains an ingestion responsibility.


gouveags:
@greptile review

gouveags:
@richardsolomou here is the JS fix you asked for during the review on the Python PR.

I'm keeping this one as a draft until PostHog/posthog#71323 lands. Does this cover the JS side you had in mind, or is there anything else you'd like me to include?

gouveags:
@greptile review

packages/ai/src/langchain/callbacks.ts
```diff
@@ -725,6 +807,24 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       additionalTokenData.cacheWriteInputTokens = usage.input_token_details.cache_creation
     }
 
+    const cacheCreationTtl =
```
richardsolomou, line 850: Please validate the direct Anthropic TTL total against the aggregate before emitting the breakdown, and add a mismatch regression test. With an aggregate of 300 and TTL values totaling 200, this path replaces the valid aggregate with 200 and causes ingestion to underprice the cache write instead of using the aggregate fallback.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -802,40 +902,68 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
+    }
+
     const llmUsageKeys = ['token_usage', 'usage', 'tokenUsage']
 
     if (response.llmOutput != null) {
       const key = llmUsageKeys.find((k) => response.llmOutput?.[k] != null)
       if (key) {
-        llmUsage = this._parseUsageModel(response.llmOutput[key], provider, model)
+        const llmUsage = response.llmOutput[key]
+        return this._parseUsageModel(llmUsage, provider, model, key !== 'usage', llmUsage)
```
richardsolomou, line 1011: Please avoid treating an empty top-level usage object as authoritative. Adapters can emit an empty `llmOutput.tokenUsage` alongside valid generation metadata; this unconditional return captures zero tokens and skips the generation fallback that the previous implementation used.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -13,6 +13,9 @@ import { sanitizeLangChain } from '../sanitization'
 import { stringifyError } from '../serializeError'
 import { warnIfPostHogAiGateway } from '../gatewayWarning'
 
+const isRecord = (value: unknown): value is Record<string, unknown> =>
```
richardsolomou: Please reuse `isObject` from `packages/ai/src/typeGuards.ts`. It implements the same type predicate and checks, so using the shared guard avoids two copies drifting apart.

richardsolomou (changes requested):

packages/ai/src/langchain/callbacks.ts
gouveags: Good catch. I now compare the direct Anthropic TTL total with every valid aggregate before emitting the TTL fields. The mismatch regression keeps the aggregate at 300 and omits the 5m/1h fields when the breakdown totals 200.

packages/ai/src/langchain/callbacks.ts
gouveags: Fixed. Empty top-level usage objects are now skipped, so valid generation metadata remains the fallback. I added a regression with an empty `llmOutput.tokenUsage` and non-empty generation usage.

packages/ai/src/langchain/callbacks.ts
gouveags: Updated to use the shared `isObject` guard throughout the callback and removed the local duplicate.

gouveags:
@richardsolomou all fixed!

packages/ai/src/langchain/callbacks.ts
```diff
@@ -802,40 +916,70 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
-    if (llmUsage[0] === 0 && llmUsage[1] === 0 && response.generations) {
-      for (const generation of response.generations) {
-        for (const genChunk of generation) {
-          // Check other paths for usage information
-          if (genChunk.generationInfo?.usage_metadata) {
-            llmUsage = this._parseUsageModel(genChunk.generationInfo.usage_metadata, provider, model)
-            return llmUsage
-          }
+    const isAnthropic = provider?.toLowerCase() === 'anthropic' || model?.toLowerCase().includes('anthropic') === true
+    if (isAnthropic && normalizedGenerationUsage) {
```
richardsolomou: One last one! Please treat empty Anthropic generation usage as missing and add a regression for this fallback. If `message.usage_metadata` is `{}` while `response_metadata.usage` contains valid counts, this truthy check parses the empty object and emits zero tokens instead of using the raw generation usage.

richardsolomou (changes requested):

packages/ai/src/langchain/callbacks.ts
gouveags: No worries. Keep the reviews coming as needed! And sorry for missing this one. I just fixed it!

gouveags:
Done!

packages/ai/src/langchain/callbacks.ts
```diff
@@ -802,40 +916,70 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
+        const llmUsage = response.llmOutput[key]
+        if (!isObject(llmUsage) || Object.keys(llmUsage).length === 0) {
+          continue
         }
+        return this._parseUsageModel(llmUsage, provider, model, key !== 'usage', llmUsage)
       }
     }
 
-    return llmUsage
+    if (rawGenerationUsage) {
```
richardsolomou: Please treat empty raw generation usage as missing and add a regression for the Bedrock fallback. When `response_metadata.usage` is `{}` but `amazon-bedrock-invocationMetrics` contains valid counts, this branch parses the empty object and emits zero tokens instead of using the invocation metrics.

richardsolomou (changes requested):

packages/ai/src/langchain/callbacks.ts
gouveags: @richardsolomou You were right. The same empty usage check was repeated in a few branches, so I centralized it instead of only patching the Bedrock fallback. I also added regressions for the fallback, valid usage later in the same generation or a later generation, and explicit zero values. Does this cover the case you had in mind?

richardsolomou (approved):

gouveags:
@richardsolomou we still have some pending checks

richardsolomou:
@gouveags "Commits must have verified signatures" 😬😅

marandaneto:
i forced pushed signed commits
theres [this](https://github.com/PostHog/posthog-js/pull/4236#discussion_r3665974633) open question, can this be merged as is @richardsolomou @PostHog/team-ai-observability

gouveags:
Thanks for the help, @marandaneto! Things have been hectic on my side, so I haven’t really had time to look at this before.

gouveags:
@richardsolomou I resolved the conflicts with the latest main and reran the focused @posthog/ai checks. The branch is mergeable again, and the new commits are verified. Could you take another look when you have a chance?

gouveags:
@marandaneto maybe you can help

PR #4238, opened 2026-07-23
chore(browser): add fast array bundle size comparison

dustinbyrne:
## Problem

Checking whether browser SDK work changes `array.js` size currently requires the full Rollup build, which takes too long for quick iteration. Contributors need a fast, directional, apples-to-apples percentage comparison against `main` or another git baseline.

Sometimes I need to do this a lot. Maybe you might need to too. Now we can do it faster.

## Changes

- Add `pnpm bundle-size:array [baseline-ref]`, backed by esbuild rather than the production Rollup pipeline.
- Bundle the current working tree and a temporary git snapshot with identical options and installed dependencies.
- Report minified, gzip, and Brotli byte and percentage changes in about one second.
- Add contributor documentation and a project skill for invoking and interpreting the comparison.
- Keep browser package metadata out of the proxy bundle so unrelated `package.json` changes do not affect results.

Validated with `pnpm bundle-size:array HEAD`; all three comparisons reported `0.00%` (tested against a real branch too, but the robot didn't think you'd want to know that).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with the Pi coding agent. The implementation intentionally uses esbuild as a fast comparison proxy rather than reproducing the production Rollup/Terser pipeline; the command labels absolute production sizes as different. The branch is based directly on the latest `origin/main` and the original feature worktree was left clean.

### DIFF
--- a/.agents/skills/compare-array-bundle-size/SKILL.md
+++ b/.agents/skills/compare-array-bundle-size/SKILL.md
@@ -1,0 +1,40 @@
+---
+name: compare-array-bundle-size
+description: Quickly compare the posthog-js array.js bundle size in the current working tree against a git baseline using the repository's esbuild proxy. Use when asked about array.js size, bundle-size percentage changes, whether browser SDK changes increased the bundle, or for a fast size check that should not run the production Rollup build.
+compatibility: Requires a posthog-js checkout with dependencies installed.
+---
+
+# Compare `array.js` bundle size
+
+Use the repository's fast comparison command. Do not run TypeScript compilation, Terser, Rollup, Turbo, or the full build unless the user separately requests production bundle numbers.
+
+## Workflow
+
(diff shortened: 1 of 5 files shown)


dustinbyrne:
just dx improvements here, no changeset

turnipdabeets (approved):

PR #4265, opened 2026-07-24
feat(react-native): add autoPresentSurveys prop to defer popover survey presentation

### LINKED ISSUE #4242: React Native: no supported way to defer auto-presented popover surveys (collide with native-stack formSheet/modal)
**SDK:** `posthog-react-native`
**Type:** feature request (escape hatch) + interaction bug

### The gap (verified against `main`)
`PostHogSurveyProvider` auto-presents a popover survey the instant it activates (`setActiveSurvey(popoverSurveys[0])` → renders `<SurveyModal>` for any `SurveyType.Popover`), and there is **no supported way to defer or suppress that presentation**. The provider already sketches the escape hatch, but it is entirely commented out in `packages/react-native/src/surveys/PostHogSurveyProvider.tsx`: `// automaticSurveyModal?: boolean`, `// export const useActiveSurvey`, `// export const useFeedbackSurvey`.

### Why it matters — collision with native-stack sheets/modals
The survey renders as a transparent, full-screen RN `Modal` whose absolute-fill backdrop captures touches (`SurveyModal.tsx`: `<Modal transparent>` → `<View style={[styles.fill, styles.backdrop]} onTouchStart={Keyboard.dismiss}>`). When a route presented by a **native stack** as `formSheet` or `modal` (react-native-screens — e.g. Expo Router / React Navigation native-stack) is on top and a survey auto-activates, the survey `Modal` mounts over the native sheet and its backdrop can intercept touches intended for the frontmost native presentation, leaving the UI unresponsive on iOS.

This is **distinct from the `Modal.onDismiss` freeze fixed in 4.54.4** (`IOS_DISMISS_FALLBACK_MS`): that fix guarantees the survey `Modal` notifies its parent on *close*; it does not stop the survey from presenting *over* a native sheet in the first place. The missing piece is an app-controlled "not right now."

### Proposed fix (any one unblocks this)
1. Ship the commented-out `automaticSurveyModal={false}` + export `useActiveSurvey()` / `useFeedbackSurvey()`, so apps can render / gate the survey themselves; **or**
2. Add a `canShowSurvey?: boolean` (or `shouldShowSurvey?: () => boolean`) prop on `PostHogSurveyProvider` that **defers (re-checks later), not drops**, presentation while falsy — the survey presents once it flips back to true.

A userland workaround today is to patch the provider to add such a gate, but that requires `patch-package` against a minified build and breaks on every SDK bump.

### Related
- #3817 (manual `displaySurvey` API) — an imperative display API would share this solution space (opt out of auto-present, show on demand).

### Environment / repro
`posthog-react-native` (latest), iOS, `react-native-screens` native stack via Expo Router, `presentation: 'formSheet' | 'modal'`. A minimal Expo Router repro (one `formSheet` route + one popover survey) is available on request; happy to add a screen recording and confirm New Architecture (Fabric) on/off if useful for triage.

ioannisj:
Closes #4242

## Problem

React Native apps have no supported way to defer an auto-presented popover survey. `PostHogSurveyProvider` picks the first matching popover survey and renders its full-screen transparent `Modal` the moment the survey activates. When a route presented by a native stack as `formSheet` or `modal` (react-native-screens, so Expo Router / React Navigation) is on top, that Modal mounts over it and its absolute-fill backdrop can swallow touches meant for the native sheet, so the UI goes unresponsive on iOS.

## Changes

Adds an optional `autoPresentSurveys` prop to `PostHogSurveyProvider`, defaulting to `true` so nothing changes for existing apps. Set it to `false` to defer presentation, e.g. while a native-stack `formSheet` or `modal` is on top.

Deferral is display-only and non-destructive:

- An eligible popover survey stays armed while the prop is `false` and presents once it flips back to `true`. The prop is reactive, so the recheck is automatic.
- A survey already on screen isn't interrupted when the prop goes `false`. A latch keeps it mounted until the user closes it.
- `survey shown` still fires only when the modal actually displays, exactly once. Survey selection, fetch, and seen-tracking are untouched.

Tested with six new provider tests (deferred, presents once when enabled, no interruption when disabled mid-display, stays gated across re-renders, default auto-presents, latch clears on close). Also validated end to end in the `example-expo-53` app against a mock server: with `autoPresentSurveys={false}` the survey is armed but not shown, flipping to `true` presents it and fires exactly one `survey shown`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/.changeset/tidy-ships-nod.md
+++ b/.changeset/tidy-ships-nod.md
@@ -1,0 +1,5 @@
+---
+"posthog-react-native": minor
+---
+
+Add an `autoPresentSurveys` prop to `PostHogSurveyProvider`. Set it to `false` to defer automatic presentation of popover surveys, for example while a native-stack `formSheet` or `modal` is on top. Deferral is display-only: the survey stays armed and presents once the prop becomes `true` again, and a survey already on screen is never interrupted.
--- a/examples/example-expo-53/app/_layout.tsx
+++ b/examples/example-expo-53/app/_layout.tsx
@@ -1,6 +1,6 @@
 import { DarkTheme, DefaultTheme, ThemeProvider } from '@react-navigation/native'
 import { useFonts } from 'expo-font'
-import { Stack } from 'expo-router'
+import { Stack, usePathname } from 'expo-router'
 import { StatusBar } from 'expo-status-bar'
 import 'react-native-reanimated'
 
@@ -11,6 +11,15 @@ import { posthog } from './posthog'
 
 export default function RootLayout() {
     const colorScheme = useColorScheme()
+
+    // Demo of deferring popover surveys via `autoPresentSurveys`. Here we defer while the
+    // Surveys screen is open so an auto-popover doesn't fight that screen's demo UI, and
+    // let it present anywhere else. Real apps wire this to whatever should hold a popover
+    // back — e.g. a native formSheet/modal being on top — and are responsible for flipping
+    // it true again (here, on route change), or the survey stays deferred and never shows.
+    const pathname = usePathname()
+    const autoPresentSurveys = pathname !== '/surveys'
+
     const [loaded] = useFonts({
          
         SpaceMono: require('../assets/fonts/SpaceMono-Regular.ttf'),
@@ -31,7 +40,7 @@ export default function RootLayout() {
             }}
             debug={true}
         >
-            <PostHogSurveyProvider client={posthog}>
+            <PostHogSurveyProvider client={posthog} autoPresentSurveys={autoPresentSurveys}>
                 <ThemeProvider value={colorScheme === 'dark' ? DarkTheme : DefaultTheme}>
                     <Stack>
                         <Stack.Screen name="(tabs)" options={{ headerShown: false }} />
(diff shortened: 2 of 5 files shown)


turnipdabeets (approved):
LGTM.

Though we should probably document more loudly: the app fully owns flipping the prop back to true. If it goes false and the un-defer never happens (sheet dismissed via a path that doesn't refocus the provider, an unmount, an error boundary), the survey stays armed forever and survey shown never fires. No crash, but from the developer's prospective it just looks like "surveys are broken" and it'd be a hard one to debug. Worth a sentence in the JSDoc and, maybe even an Expo Router / React Navigation snippet wiring it to route focus so people have a correct pattern to copy in example app?

PR #4270, opened 2026-07-24
feat(replay): canvas mask regions for session replay canvas capture

turnipdabeets:
## Problem

Canvas capture records pixels, so DOM-based masking (`maskTextSelector`, `blockSelector`, input masking) structurally cannot mask text that an app has painted *into* a canvas. For apps that render entirely into canvas — Flutter web via CanvasKit being the reported case (PostHog/posthog-flutter#496) — every on-screen text ships to replay in plaintext, with no supported mitigation short of disabling canvas capture.

Needed for https://github.com/PostHog/posthog-flutter/pull/499

## Changes

**One** client-side option on `session_recording.canvasCapture` — the client-only capture-tuning bucket, alongside its precedent `resolutionScale` (#3885); `captureCanvas` stays reserved for remote-overridable settings:

**`maskRegionsFn?: (canvas) => CanvasMaskRegion[] | null | undefined`** — called once per canvas per captured frame, in the same task that snapshots the canvas. Returned CSS-pixel regions are validated, scaled to capture resolution (rounded **outward** to integers so anti-aliased mask borders can't blend content through), and painted black in the encode worker **before** the pixel-fingerprint dedup and webp encode — masked pixels never exist in any event payload. Re-read from live config every frame, so an embedding SDK that boots after recording starts (Flutter engine ≈1–3s) can swap in the real provider via `set_config` without a recorder restart.

The return value carries the whole contract — there is no second option:

| Return | Meaning | Result |
|---|---|---|
| *(option not set)* | no provider configured | recorded unmasked — **identical to main** |
| `[{…}]` | regions to hide | painted black |
| `[]` | "I looked, nothing to mask" | recorded as-is |
| `null` | "I could not compute this frame" | **frame skipped** |

An app whose real provider only exists once its runtime has booted chooses what happens in the gap purely by what it declares in `posthog.init`: a function covering the whole canvas blacks those frames out, `() => null` skips them, declaring nothing records them.

Client-only (a function can't come from remote config). **Without the option, canvas capture behavior is byte-identical to main** — the only new work is an optional-chain config read per canvas per tick.

### Why skip rather than black out

A frame that can't be masked is dropped before `createImageBitmap`, so no bitmap is created and nothing reaches the encode worker. Dropping rather than blacking out means a late-registering provider doesn't paint a black rectangle over the start of every recording — playback shows the page as it was (canvas untouched) until real frames arrive, and a transient mid-session failure holds the last good masked frame instead of flashing black. Strictly no weaker than blacking out: in both cases nothing unmasked is encoded.

### Three separable concerns (one branch)

- **Core masking** — types, `computeFrameMaskRegions` (extracted to `canvas-mask.ts` for unit testing), CanvasManager/record plumbing, worker painting, recorder config wiring, public types + doc comments, changeset, generated API reference.
- **30s masked-frame keyframe** — masks are painted before the fingerprint dedup, so a canvas whose changes are all under masks can go minutes without emitting a frame; the player only repaints canvas on the next canvas event (#3590), so seeking into such a gap showed blank canvas. An unchanged frame is re-sent every 30s for every canvas the provider answered — with regions or `[]` — since both lose their `rr_dataURL` snapshot still, bounding any seek to ≤30s from a repaint source. **Gated to provider-answered canvases — recordings without `maskRegionsFn` are unchanged.** Cost for masked idle canvases: ~1 webp frame/30s (≈1–2MB/hour idle at 1080p/q0.6). Isolated to `image-bitmap-data-url-worker.ts` — happy to pull it into a separate PR if you'd rather fix #3590 player-side first.
- **Full-snapshot canvas skip** — live payload testing on Flutter web (3.44.1 stable) refuted the assumption that CanvasKit canvases can't serialize into DOM snapshots: the DOM canvas is a `bitmaprenderer` canvas whose presented buffer *is* readable, so when the recorder loads after the engine boots (the common ordering — remote config + lazy recorder script vs ~1s engine boot), every interval full snapshot embedded a full-resolution unmasked screenshot via `rr_dataURL` (the serializer block is gated on `recordCanvas`, i.e. armed by enabling canvas capture). Now `canvasMasking.configured` — true iff `canvasCapture.maskRegionsFn` is set, evaluated at `record()` start (the embedding SDK's one-time registration restart re-evaluates it) — threads through to the serializer, which then skips canvas pixel serialization entirely: both the `'2d'` and unobserved-context branches, mutation-added canvases, and shadow-DOM/same-origin-iframe serialization, without ever reading pixels. Canvases repaint from the (masked) frame stream within ~250ms at the default 4 fps, and the 30s keyframe bounds seeks. Without the option the serializer is byte-identical to main.

(The related worker-leak fix was split out and already merged as #4223.)

### Reviewer attention — the always-on lines

Most of the diff is inert without the new option; these execute for every canvas-capture user and deserve the closest look:

- `lazy-loaded-session-recorder.ts`: the `regionsFn` wrapper must keep `undefined` (no provider configured) distinct from `null` (provider couldn't answer). Collapsing them with `?? null` would make **every** canvas without a provider fail closed and stop recording.
- `canvas-manager.ts`: `computeFrameMaskRegions` is called **before** `await createImageBitmap` (so regions and display dims describe the same frame if the canvas resizes mid-snapshot), and a skipped frame must clear `snapshotInProgressMap` or that canvas never captures again.
- `image-bitmap-data-url-worker.ts`: the dedup branch gained the keyframe check; the fillRect block is skipped when `maskRegions` is undefined.
- `rrweb-snapshot/src/snapshot.ts`: the canvas `rr_dataURL` block gained `&& !canvasMaskingConfigured` — the flag must default `false` (and does, at every layer) so non-masking users serialize exactly as on main.

### Fail-closed contract details

- Any malformed region element (non-object / non-finite x, y, width, height) invalidates the whole result → frame skipped. (A `[{}]` element would otherwise pass `Array.isArray` and become `fillRect(NaN,…)`, a spec no-op — i.e. a silent fail-open. Executed repro during internal review.)
- A provider that **throws** is treated the same as `null` — skipped, and the throw never escapes into capture.
- A computed **empty array** is a valid "nothing to mask" and is honored.
- Zero-area regions are dropped at the scaling step, so a provider returning only zero-area regions behaves exactly like `[]` — nothing painted (outward rounding would otherwise inflate a fractional-offset zero-width region into a 1px stripe), and like `[]` it still keyframes as a provider-answered canvas.
- Masked pixels feed the fingerprint, so changes entirely under masks dedup (less data, no leak).

### Not in this PR

- The consuming Flutter SDK change (separate PR in posthog-flutter; it feature-detects and degrades to current behavior on older posthog-js).
- A player-side seek-repaint for #3590 (the durable fix there; the keyframe commit only bounds the gap for provider-managed canvases).
- Decision: `@posthog/rrweb`'s `CanvasMasking.configured` and `regionsFn` stay independently optional by design; posthog-js always sets both, and `configured` cannot be inferred from `regionsFn` presence because the wrapper `regionsFn` is installed unconditionally.
- Deferred nits from internal review: a doc note that mutation-mode `sampling: 'all'` is unreachable via posthog-js. (Since resolved on this branch: the client-only option no longer structurally leaks into remote-shaped types — it lives in the client-only `canvasCapture` bucket; negative extents invalidate the whole result — fail closed; zero-area regions are filtered at scaling; and the full-snapshot `rr_dataURL` path is skipped when masking is configured.)

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code — `canvas-mask.test.ts` (13: scaling, outward rounding, canvas pass-through, no-provider vs provider-returned-nothing, empty-list, full unanswerable-provider matrix, zero-area filtering — mixed and only-zero-area lists), `canvas-manager.test.ts` (+1: a skipped frame creates no bitmap and never reaches the worker, then capture resumes — verified to fail without the guard), worker tests (+2: region painting with pixel assertions, `[]` recorded as-is, mask-only-change dedup, keyframe cadence, no keyframe without a provider, keyframe for an unchanged `[]`-answered canvas after the interval), recorder plumbing jest tests (live-config `regionsFn`, `undefined` vs `null`), snapshot-skip tests (serializer: both `rr_dataURL` branches × flag set/unset with no pixel reads when set; record-level: flag reaches `snapshot()` iff `canvasMasking.configured`; recorder: `configured` true iff `maskRegionsFn` is set, not with `recordCanvas` alone)
- [x] Ran changeset (minor: `posthog-js`, `@posthog/types`; the `@posthog/rrweb*` entries were intentionally dropped — they ship bundled as devDependencies and are delivered by the `posthog-js` bump, so the changeset-hygiene bot may post an expected informational comment listing them as undeclared)

## 🤖 Agent context

Built with Claude Code from the investigation of PostHog/posthog-flutter#496.

An earlier revision of this PR had a second option, `requireMaskProvider: boolean`, to opt into fail-closed behavior. It was removed: absence of a provider is ambiguous between "this app doesn't mask" and "this app masks but isn't ready", and the boolean existed only because the recorder's wrapper collapsed both into `null`. Keeping `undefined` distinct from `null` resolves the ambiguity at the source, so declaring the function — including as `() => null` — *is* the opt-in, and one public option carries the whole contract.

Design alternatives considered and rejected: screenshot-mode replay from the Flutter SDK (bypasses the recorder lifecycle, loses console/network/cursor), `before_send` rewriting (frames already webp-encoded and often gzip-compressed; hook is sync), and DOM overlays (can't be replay-visible while live-invisible). The chosen seam — regions painted in the encode worker pre-fingerprint — is the only point where raw pixels exist pre-network and geometry is injectable.


packages/types/src/posthog-config.ts
```diff
@@ -653,7 +658,47 @@ export interface SessionRecordingOptions {
...
+         * frame is encoded. Lets apps that render into canvas (e.g. Flutter web)
+         * mask content that DOM-based masking cannot see. Re-read from config on
+         * every frame, so the real provider can be swapped in after recording has
+         * started.
+         *
+         * Return `[]` for a frame with nothing to mask (recorded as is), or `null`
+         * if regions could not be computed — that frame is skipped rather than
+         * recorded unmasked. Anything other than an array — `null`, a thrown error,
+         * or an implicit `undefined` from an untyped caller — skips that frame.
+         * Not setting this at all records the canvas unmasked.
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — fixed in cfee65c06. Widened the return type to `CanvasMaskRegion[] | null | undefined`, matching the doc comment, the `fn(canvas) ?? null` runtime handling, and the `maskCapturedNetworkRequestFn` convention.

turnipdabeets:
@greptile-apps review pls

.changeset/canvas-mask-regions.md
```diff
@@ -0,0 +1,8 @@
+---
+'posthog-js': minor
+'@posthog/rrweb': minor
+'@posthog/rrweb-types': minor
+'@posthog/types': minor
```
marandaneto: those should not exist

dustinbyrne (approved):

dustinbyrne:
i've read through this a couple of times and it makes sense - @marandaneto's feedback still outstanding

.changeset/canvas-mask-regions.md
turnipdabeets: Right — dropped `@posthog/rrweb` and `@posthog/rrweb-types` in c10346242 per the AGENTS.md policy (rrweb modules are built with posthog-js). Kept `@posthog/types` since the PR adds public config types there and that package still takes its own changesets — happy to drop that one too if it's covered by the same policy.

arnohillen (approved):
Robot review:

Went looking for the null-versus-undefined trap and found the wrapper comment had already closed it. Fail-closed on throw, outward rounding so masks never under-cover, zero-area regions dropped before scaling. Tests are thorough.

One question: `configured` is evaluated once at `record()` start while `regionsFn` is read live. An SDK that registers a provider without restarting recording would keep serializing raw canvas pixels into DOM full snapshots while per-frame captures are masked. Is the restart guaranteed on the consumer side?

turnipdabeets:
> One question: `configured` is evaluated once at `record()` start while `regionsFn` is read live. … Is the restart guaranteed on the consumer side?

@arnohillen good question — two answers. On the consumer side: the Flutter plugin always restarts recording once at registration regardless (it has to — `blockSelector` is start-time-only), verified live in both of its opt-in flows. But you're right that the contract shouldn't rest on consumer convention, so 1ca7c0597 makes it symmetrical with `regionsFn`: `configured` is now a thunk evaluated at each canvas serialization (interval snapshots, mutation-added canvases, shadow DOM), so a provider installed via `set_config` mid-recording is honored at the next snapshot with no restart needed. Doc comment updated, and there's a regression test where the thunk flips between two snapshots.

.changeset/canvas-mask-regions.md
turnipdabeets: Confirmed the policy scope — AGENTS.md only excludes `packages/rrweb/` modules. `@posthog/types` versions independently via its own changesets (see its CHANGELOG, e.g. #4222), and this PR adds public config types there, so that entry stays.

turnipdabeets:
Re the ⚠️ backward-compatibility comment above: the run itself passed — **371/371** against published `posthog-js@1.407.7` (the suite includes the `session-recording/` specs, so the old-core + new-recorder path this PR changes is exercised). The ⚠️ is a results-parser issue in the workflow, not a test failure: `Process compat test results` treats an unparseable `compat-results.json` as "tests failed", and the same comment is currently appearing on unrelated open PRs (#4308, #4309, #4312) with identical 371-passed runs.

On the actual compatibility question: the no-option path is byte-identical to the published behavior — asserted by unit tests at the serializer/record/browser layers, and verified end-to-end in a live control run of this build without the option (it reproduces current behavior exactly). Version skew has no crash path in either direction: an old cached `array.js` with the new recorder evaluates `configured()` against a config that can't contain the option → false → serialization unchanged; a new `array.js` with an old cached recorder just stores an ignored config key. The boolean form of `configured` never shipped, so no third combination exists.

arnohillen:
One issue: `[]` frames never keyframe, but full snapshots still omit those canvases. After a full snapshot or seek, an unchanged canvas can stay blank indefinitely. Could we keyframe any defined result, including `[]`?

turnipdabeets:
> One issue: `[]` frames never keyframe, but full snapshots still omit those canvases. After a full snapshot or seek, an unchanged canvas can stay blank indefinitely. Could we keyframe any defined result, including `[]`?

Good catch — and agreed, that's exactly the right fix. It landed in 46eb0bf2e (pushed shortly before your comment): the keyframe is now gated on `maskRegions !== undefined`, so every provider-answered canvas — including `[]` — re-sends an unchanged frame every 30s, since the full-snapshot skip is page-wide and `[]`-answered canvases lose their `rr_dataURL` still too. There's a test pinning the `[]` case (`image-bitmap-data-url-worker.test.ts`), and the `maskRegionsFn` doc comment + changeset now state the ≤30s repaint bound applies to any answered canvas; only `null`-skipped canvases stay blank by design. Cost is ~1 webp frame per 30s per idle answered canvas.

PR #4271, opened 2026-07-24
fix(surveys): render inline survey instead of empty container on stale question index

felipeatom:
### Problem

Inline surveys could render an **empty container** instead of the survey. On resume, `Questions` restores `currentQuestionIndex` from the persisted in-progress record, and a stale index left over from a prior completion could point past the end of the questions array — the question renderer then bails and nothing is shown. `visitedIndices` had the same gap, so the Back button could navigate to a non-existent question and re-empty the container.

Closes #3575.

### Fix

- Read the persisted in-progress state once and **discard the whole record** when its index is *present but out of range* — the record is stale, so its `responses`/`visitedIndices` are dropped too rather than being kept alongside a reset index.
- A **missing / NaN / non-numeric** index predates the persisted-index feature: it is *not* treated as stale — the responses are kept and the survey starts at question 0 (`isNumber` guard).
- The current-index reader falls back to `0` for any non-usable index, so a resumed survey never renders an empty container.
- Restored `visitedIndices` are filtered to `[0, questions.length)` so Back can only ever land on a real question.

### Tests

- `survey-popup.test.tsx`: `it.each` over `[-1, length, length + 10, undefined, NaN]` asserting responses are kept only for an absent/in-range index and dropped for a present out-of-range one, and that question 0 always renders.
- `surveys.test.ts`: persisting an out-of-range trailing `visitedIndices` entry and clicking Back lands on a real question.


packages/browser/src/extensions/surveys.tsx
```diff
@@ -1327,7 +1327,12 @@ export function Questions({
...
         const inProgressSurveyData = getInProgressSurveyState(survey)
-        return previewPageIndex || inProgressSurveyData?.lastQuestionIndex || 0
+        // A stale persisted index (e.g. left over from a prior completion) can point past the
+        // end of the questions array, which would make the question renderer bail and leave the
+        // survey container empty. Clamp it back into range so an inline re-display still renders.
+        const savedIndex = inProgressSurveyData?.lastQuestionIndex ?? 0
+        const validSavedIndex = savedIndex >= 0 && savedIndex < survey.questions.length ? savedIndex : 0
+        return previewPageIndex || validSavedIndex
     })
     const [visitedIndices, setVisitedIndices] = useState<number[]>(() => {
```
ioannisj, line 1429: the clamp only covers `currentQuestionIndex`, but `visitedIndices` right below it is restored from the same persisted blob with no bounds check. 

Would it be worth clamping/filtering `visitedIndices` to `[0, survey.questions.length)` in the same initializer, or validating `previousIndex` in `onBackButtonClick` before setting it? A test that clicks Back with an out-of-range persisted `visitedIndices` would lock it in too.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1327,7 +1327,12 @@ export function Questions({
     } = useContext(SurveyContext)
     const [currentQuestionIndex, setCurrentQuestionIndex] = useState(() => {
         const inProgressSurveyData = getInProgressSurveyState(survey)
-        return previewPageIndex || inProgressSurveyData?.lastQuestionIndex || 0
+        // A stale persisted index (e.g. left over from a prior completion) can point past the
+        // end of the questions array, which would make the question renderer bail and leave the
+        // survey container empty. Clamp it back into range so an inline re-display still renders.
+        const savedIndex = inProgressSurveyData?.lastQuestionIndex ?? 0
```
ioannisj: Small thing on the fallback here: when `savedIndex` comes back out of range you reset `currentQuestionIndex` to 0, but the `responses` and `visitedIndices` restored just below from the same stale `inProgressSurveyData` are left untouched, so the re-displayed survey renders question 0 while still carrying the old progress state. Might be cleaner to clear the in-progress record (mint a fresh one) when the index comes back stale rather than clamp-and-keep. Is keeping the old responses intentional here?

ioannisj:
@felipeatom friendly ping

felipeatom:
Good catch on both, thanks @ioannisj — you're right that clamp-and-keep left the rest of the stale blob dangling.

Reworked it so an out-of-range persisted index is treated as a stale record rather than something to salvage: I now read the in-progress state once, and if `lastQuestionIndex` is out of range I clear the record and start fresh, so the stale `responses` and `visitedIndices` don't carry over either. Keeping the old responses was **not** intentional — this drops them.

On `visitedIndices` specifically, I also filter the restored array to `[0, questions.length)` on init, so `onBackButtonClick` can never pop an out-of-range entry and re-empty the container. Added a test that persists an out-of-range trailing visited index and clicks Back, asserting it lands on a real question.

ioannisj:
@felipeatom looks like some relevant checks are failing here. Mind having a look?

felipeatom:
Fixed @ioannisj — thanks for the nudge. The failing unit tests were in `survey-popup.test.tsx` and they were on me: my previous commit discarded the whole in-progress record whenever `lastQuestionIndex` wasn't a valid in-range number, but a persisted blob from before the index feature has no `lastQuestionIndex` at all. Reading that as "stale" wrongly dropped the restored responses on resume.

Now I default a missing index to 0 (the same fallback the reader already uses) and only discard when the index is actually present *and* out of range. Ran the full survey unit suite locally against the merged branch — 478/478 green, including the three `survey-popup` restore tests and the stale-index/back-navigation ones from this PR.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1344,13 +1345,31 @@ export function Questions({
...
+    // left over from a prior completion) can point past the end of the questions array, which
+    // makes the question renderer bail and leaves the survey container empty. When that happens
+    // the whole record is stale, so we discard it and start fresh rather than clamping the index
+    // while keeping the equally-stale responses and visited indices around.
+    const initialInProgressState = useMemo(() => {
+        const state = getInProgressSurveyState(survey)
+        if (!state) {
+            return null
+        }
+        const isIndexInRange = state.lastQuestionIndex >= 0 && state.lastQuestionIndex < survey.questions.length
```
lucasheriques: `state.lastQuestionIndex >= 0` is `false` when the field is absent, `NaN`, or non-numeric, so all of those take the clear path and lose the saved responses along with the index.

splitting the two cases covers it:

```ts
const hasIndex = typeof state.lastQuestionIndex === 'number' && !Number.isNaN(state.lastQuestionIndex)
const isIndexInRange = hasIndex && state.lastQuestionIndex >= 0 && state.lastQuestionIndex < survey.questions.length
if (hasIndex && !isIndexInRange) {
    clearInProgressSurveyState(survey)
    return null
}
```

then no index means start at 0 and keep the responses, while an out-of-range index still clears. the `?? 0` you already have at :1387 handles the first case. note this treats a negative index as corrupt rather than absent, so `-1` still clears.

lucasheriques (commented):
good find on the root cause, and clearing the record rather than clamping is the right call after the earlier thread.

the range check is what is turning ci red though. `undefined >= 0` is `false` in js, so a record that just omits `lastQuestionIndex` gets classified stale and `clearInProgressSurveyState` takes the saved responses with it. that is the 3 failures in `survey-popup.test.tsx` (`initializes with existing submissionId and responses from localStorage`, `clears localStorage on final submission`, `clears localStorage on dismissal`), all on `expect(textbox).toHaveValue('Previous answer')`. details inline.

worth saying those 3 mocks are incomplete rather than realistic: `lastQuestionIndex` is typed required and all three write sites set it, so no live user is hitting this. i would still fix it in the production code rather than the mocks, for two reasons. the code treats "no index recorded" and "index out of range" as the same thing when only the second one means the responses are stale. and fixing the check keeps you from editing three pre-existing tests to accommodate new code.

when you add a test for the missing-index case, `it.each([-1, questions.length, questions.length + 10, undefined, NaN])` would cover the whole range in about the same space as the two you have.

one small thing on the pr body: it still describes the original clamp-to-0 approach, so it reads as out of date next to the clearing behaviour now.

felipeatom:
Thanks @lucasheriques — adopted the split. You reviewed `459fc05`, and I'd already pushed `6eb4066` which fixed the 3 `survey-popup` failures via `?? 0`, but your `hasIndex` framing is the better model: it makes "no index recorded" vs "index out of range" explicit instead of collapsing them, and it also covers `NaN`/non-numeric which `?? 0` doesn't.

Pushed `8ca8769`:
- Guard is now `const hasIndex = isNumber(state.lastQuestionIndex)` (the repo lint rule wants `isNumber` over a raw `typeof`, and it already excludes `NaN`), clearing only when `hasIndex && !isIndexInRange`.
- Made the current-index reader fall back to 0 for any non-usable index, so the `NaN` case renders question 0 with responses intact rather than a `NaN` index.
- Added the `it.each([-1, length, length + 10, undefined, NaN])` you suggested.
- Rewrote the PR description so it matches the clearing behaviour instead of the old clamp-to-0 wording.

Full survey suite is green locally (483/483).

felipeatom:
The failing **Unit tests** check wasn't a test failure — all 5,155 tests passed. It tripped the post-run leaked-resources guard (`A worker process has failed to exit gracefully`).

Root cause: Open questions schedule a 100ms autofocus `setTimeout` on mount. The `@testing-library` tests auto-unmount and clear it, but my new #3575 Back-navigation test renders via `renderSurvey` (raw Preact mount, no teardown), leaving that timer orphaned — which can leave a Jest worker with a pending timer at exit under CI's parallelism.

Fixed in c6e6138 by wrapping the test in fake timers so the autofocus timer never becomes a real pending handle. Locally: `surveys.test.ts` 85/85, full survey suite 486/486, and the test alone under `--detectOpenHandles` reports zero open handles.

Could a maintainer re-run the workflows when convenient? Thanks!

lucasheriques (approved):

PR #4273, opened 2026-07-24
fix(surveys): keep open selector-widget survey when its trigger element unmounts

### LINKED ISSUE #2036: Surveys abruptly disappear on lose focus

felipeatom:
## Problem

Fixes #2036.

A **selector-widget** survey (custom CSS selector trigger) is abruptly removed while it is open if its trigger element leaves the DOM — for example a dropdown or menu that hosts the trigger closes. The user loses the survey mid-answer.

## Root cause

The ~1s display poll calls `_manageWidgetSelectorListener`, which on `!currentElement` (trigger no longer in the DOM) calls `_removeWidgetSelectorListener` → `_removeSurveyFromDom` **unconditionally**, tearing down the survey even when it is currently open.

This matches the diagnosis already reached in the issue thread, where @lucasheriques noted the fix should be that "the survey is only removed from the DOM if it's not open. Otherwise, this should happen only on close."

## Fix

Guard `_removeWidgetSelectorListener` with a new `_isWidgetSurveyOpen` check: while the survey popup (`.ph-survey`) is mounted in the shadow root, defer teardown and return early. A later poll retries the cleanup once the user has closed the survey, so nothing leaks.

## Testing

- Added a regression test that opens a selector-widget survey, unmounts its trigger element, runs the display poll and asserts the open survey stays in the DOM. It fails without the fix (the popup is removed) and passes with it.
- Full survey suites are green (139 tests). Lint, Prettier and the mangled-names list are updated and clean.


packages/browser/src/extensions/surveys.tsx
```diff
@@ -293,6 +293,13 @@ export class SurveyManager {
     }
 
     private _removeWidgetSelectorListener = (survey: Pick<Survey, 'id' | 'type' | 'appearance'>): void => {
+        // Defer teardown while the survey is open (issue #2036). The trigger element may have
+        // been unmounted mid-survey — e.g. a dropdown/menu that hosts it was closed — and tearing
+        // the survey down here would make the open survey abruptly vanish. Keep it in place; the
+        // next display poll retries this cleanup once the user has closed the survey.
+        if (this._isWidgetSurveyOpen(survey)) {
```
ioannisj, line 314: While the survey is open, `_removeWidgetSelectorListener` early-returns here, so when `_manageWidgetSelectorListener` hits the `currentElement !== existingListenerData.element` branch (the selector now resolving to a different live element), the old element's click listener and `PHWidgetSurveyClickListener` attribute are never removed, and the next lines overwrite `_widgetSelectorListeners` with the new element, so the reference needed to clean up the old one is gone. The old trigger keeps firing `show_survey_widget` for the page's lifetime.

Could you guard the element-swap branch with the same open check, so it skips re-attaching while the survey is open and lets a later poll do the swap once it closes?

packages/browser/src/__tests__/extensions/surveys.test.ts
```diff
@@ -577,6 +577,55 @@ describe('SurveyManager', () => {
...
+
+        // the trigger element is unmounted (e.g. a dropdown that hosts it closes) while the survey is open
+        document.querySelector('.my-selector')!.remove()
+
+        // the display poll runs again and finds the trigger gone
+        await act(async () => {
+            surveyManager.callSurveysAndEvaluateDisplayLogic()
+        })
+
+        // the open survey must still be in the DOM, not abruptly removed
```
ioannisj, line 665: This proves the survey survives the trigger unmounting, but not the other half of the fix, that the deferred teardown actually completes. Could you extend it (or add a sibling) that closes the survey and runs `callSurveysAndEvaluateDisplayLogic` again, asserting the click listener and `PHWidgetSurveyClickListener` attribute get removed once `_isWidgetSurveyOpen` flips false? Without it, a regression that defers cleanup forever would still pass.

ioannisj:
@felipeatom here too

felipeatom:
Thanks @ioannisj, both spot on.

1. The element-swap leak: I split the trigger cleanup out into a `_detachWidgetSelectorListener` helper that removes the click listener + `PHWidgetSurveyClickListener` attribute **without** touching the rendered survey DOM. The swap branch now calls that directly instead of going through `_removeWidgetSelectorListener` — the latter must keep deferring while the survey is open (so it doesn't tear down the open survey's DOM), but the swap only needs to unwire the old element, so it shouldn't defer. That stops the orphaned listener from firing `show_survey_widget` for the rest of the page's life.

2. Test coverage: extended the suite with a sibling test that closes the open survey and re-runs the display poll, asserting the deferred teardown actually completes (listener dropped + marker attribute removed once `_isWidgetSurveyOpen` is false). Also added a dedicated test for the swap case above.

packages/browser/src/__tests__/extensions/surveys.test.ts
```diff
@@ -617,6 +617,175 @@ describe('SurveyManager', () => {
         expect(manageWidgetSelectorListenerSpy).toHaveBeenCalledWith(mockSurvey, '.my-selector')
     })
 
+    it('does not tear down an open selector-widget survey when its trigger element unmounts (issue #2036)', async () => {
```
lucasheriques, line 620: nit: the three new tests each rebuild almost the same survey literal. a small `makeSelectorWidgetSurvey(overrides)` helper would cut most of that.

not `describe.each` though, since these are three different state transitions rather than one assertion over varied inputs.

lucasheriques (approved):
the element-swap case from the earlier thread is handled properly: the swap branch at :347 now calls the unconditional `_detachWidgetSelectorListener`, so the old element listener and `PHWidgetSurveyClickListener` attribute come off before the map entry gets replaced. both requested tests are there, and they drive the real poll entry point against a real shadow root, which is why i trust them over mock-based ones.

one thing left and it is mechanical. `_detachWidgetSelectorListener` is a new private member, and every `_`-prefixed class property has to be listed in `packages/browser/terser-mangled-names.json` or the `Write mangled property names` job fails on a `git diff --exit-code`. `_isWidgetSurveyOpen` from the earlier commit made it in, this one did not.

```
cd packages/browser && pnpm write-mangled-property-names
```

then commit the updated `terser-mangled-names.json`. should be one line, alphabetically between `_determineInitialView` and `_disabled`. that is the only thing between this and green.

thanks for sticking with this one through the review rounds.

felipeatom:
Pushed one more commit to get CI green: the new private `_detachWidgetSelectorListener` method gets mangled by terser in the production build, so it needed registering in `terser-mangled-names.json` (the mangled-property-consistency check was failing on the generated diff). No behaviour change — just the generated snapshot. Thanks for the review @lucasheriques!

PR #4274, opened 2026-07-25
Fix slim no-external extension bundle mangling

NVolcz:
## Problem

`module.slim.no-external.js` intentionally preserves private property names, while `extension-bundles.js` mangles them. Combining these artifacts could therefore break calls across the bundle boundary at runtime, including autocapture and feature flag paths, with errors such as `this.instance.Er is not a function` and `i._r is not a function`.

The existing consistency check covered `module.slim.js` but not the no-external variant, allowing incompatible property mappings to be published.

## Changes

- Reserve the 11 private properties exchanged directly between the slim core and extension bundles.
- Keep the cross-bundle ABI and known artifact-local duplicate names in one configuration file shared by Rollup and the post-build check.
- Extend the post-build check to parse generated bundles and source maps, verify both slim variants, and fail when a divergent private-property overlap is unknown or stale.
- Run the extension bundle Playwright matrix against both `module.slim.js` and `module.slim.no-external.js`, while retaining focused coverage for the originally reported extension combination and conversations identity callbacks.
- Include the ABI configuration, mangled-name snapshot, and build scripts in Turbo's build inputs so the compatibility check cannot be skipped through a stale cache hit.

## Verification

- Confirmed that a real Vite/Rolldown application using the published package fails with `this.instance.Er is not a function`.
- Rebuilt that application using this branch's `module.slim.no-external.js`; both the integration test and production build passed.
- Reproduced the feature flag failure (`i._r is not a function`) before reserving `_isFeatureFlagCacheStale`; the permanent bootstrapped-feature-flag Playwright case now passes against both slim variants.
- `pnpm turbo run build --filter=posthog-js` — passed, including 28 normal-slim mappings, 28 classified no-external overlaps, and 11 no-external ABI mappings.
- `pnpm --filter posthog-js test:unit -- --runInBand` — 5,056 passed.
- `pnpm --filter posthog-js test:functional -- --runInBand` — 7 passed.
- `pnpm --filter posthog-js lint` — passed.
- `pnpm exec playwright test playwright/mocked/slim-bundle.spec.ts --project=chromium --reporter=line` — 32 passed.
- Verified the checker rejects unknown divergent properties, stale allowlist entries, and duplicate or conflicting configuration.
- Verified that changing the ABI configuration changes the Turbo build hash.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for the impact of changes across different platforms
- [x] Accounted for backwards compatibility of changes (no breaking changes)

### If releasing new changes

- [x] Added a patch changeset for `posthog-js`

## 🤖 Agent context

Pi Coding Agent with OpenAI GPT-5.6 implemented and tested the change under human direction. Claude Code with Opus 5 independently reviewed two iterations, which led to broader feature flag coverage, a fail-closed ABI check, and Turbo cache input coverage. No public session links are available.

The implementation preserves the existing no-external property-mangling policy and reserves only properties proven to cross the core/extension boundary. Alternative approaches such as introducing a separate no-external extension artifact were rejected to avoid expanding the public artifact surface.


packages/browser/scripts/check-mangled-property-consistency.js
```diff
@@ -23,12 +23,26 @@ const { decode } = require('@jridgewell/sourcemap-codec')
 
 const DIST = path.resolve(__dirname, '../dist')
 
-function extractMangledPropertyNames(jsFile, mapFile) {
+// Private properties intentionally exchanged between the slim core and extension classes.
+// The no-external core preserves these names, so the extension bundle must preserve them too.
+// Keep this list in sync with the cross-bundle reserved section in rollup.config.mjs.
+const CROSS_BUNDLE_PRIVATE_PROPERTIES = [
```
ioannisj, line 29: The description says the cross-bundle ABI lives in one config file shared by Rollup and the post-build check, but the diff keeps two copies of the 7 names: the reserved array in `rollup.config.mjs` and `CROSS_BUNDLE_PRIVATE_PROPERTIES` here in `check-mangled-property-consistency.js`, synced only by a "keep this list in sync" comment.

If someone adds a new cross-bundle property to one list and forgets the other, the check quietly stops guarding that boundary, which is the runtime crash this PR is meant to prevent. I think pulling the 7 names into one shared module that both files import would make the fail-closed guarantee actually hold. Worth doing here?

marandaneto:
thanks @NVolcz 
superseeded by https://github.com/PostHog/posthog-js/pull/4305

PR #4277, opened 2026-07-27
fix(nuxt): delete sourcemaps before asset manifest

### LINKED ISSUE #4275: @posthog/nuxt: deleteAfterUpload leaves stale .js.map entries in the Nitro asset manifest, causing 500s
## Bug description

With `sourcemaps.deleteAfterUpload` (the default), `@posthog/nuxt` leaves the Nitro public-asset manifest referencing `.js.map` files it has already deleted. Every request for one then returns a **500**, not a 404, and is captured as an unhandled exception in error tracking.

The ordering is the problem:

1. Nitro bakes its public-asset manifest during the rollup build, while the `.js.map` files are still on disk.
2. `@posthog/nuxt` uploads and deletes them in its `close` hook — the last hook of the build:

```js
// packages/nuxt/src/module.ts
nuxt.hook("close", async () => {
  await cliRunner(getInjectArgs(serverDir, sourcemapsConfig));
  await cliRunner(getUploadArgs(outputDir, sourcemapsConfig)); // --delete-after
});
```

So the shipped manifest lists ~110 sourcemaps that no longer exist. At runtime `getAsset()` reports a hit, `readAsset()` calls `fs`, and the request dies:

```
[nitro] [request error] ENOENT: no such file or directory,
  open '/app/.output/public/_nuxt/0MZx2DBm.js.map'
  errno: -2, code: 'ENOENT', syscall: 'open',
  statusCode: 500, unhandled: true
```

If the entry simply weren't in the manifest, Nitro would take the path it already handles correctly — `getAsset()` misses, `isPublicAssetURL()` is true, clean 404.

This is **not** [#3058](https://github.com/PostHog/posthog-js/issues/3058). There is no `//# sourceMappingURL=` in the emitted chunks (`sourcemap: { client: 'hidden' }` is set and works correctly, verified on the deployed bundles). Nothing advertises these files — scanners discover them by appending `.map` to the chunk names in the HTML, which is enough to hit it thousands of times a week. In our case ~3300 requests over 7 days, each one a 500 plus a captured exception, making it the single largest source of noise in error tracking.

## How to reproduce

1. Nuxt app with `nitro.preset` = `node-server`, `sourcemap: { client: 'hidden' }`, and:

```ts
posthogConfig: {
  sourcemaps: {
    enabled: true,
    projectId: '...',
    personalApiKey: '...',
    releaseName: '...',
  },
},
```

2. `nuxt build`
3. Confirm the maps are gone but the manifest still lists them:

```sh
ls .output/public/_nuxt/*.js.map | wc -l                                   # 0
grep -c '"/_nuxt/[A-Za-z0-9_-]*\.js\.map":' .output/server/chunks/nitro/nitro.mjs   # 110
```

4. `node .output/server/index.mjs`, then request any chunk name with `.map` appended:

```sh
curl -o /dev/null -w '%{http_code}\n' http://localhost:3000/_nuxt/<chunk>.js      # 200
curl -o /dev/null -w '%{http_code}\n' http://localhost:3000/_nuxt/<chunk>.js.map  # 500
```

Removing the stale entries from the manifest and re-running gives `404` with no ENOENT, chunks and pages unaffected.

## Suggested fix

The maps need to leave `.output/public` before Nitro writes the manifest — e.g. move them somewhere the manifest doesn't cover and upload from there, or drop the entries after deletion. Either way the deletion currently happens too late in the build for Nitro to know about it.

## Environment

- `@posthog/nuxt` 1.7.77, also reproduced against 1.7.80
- `nuxt` 4.4.6, `nitropack` (node-server preset), Node 24
- Deployed behind Cloudflare, which surfaces the 500 as a 502

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

marandaneto:
## Problem

Nuxt builds with the default `sourcemaps.deleteAfterUpload` behavior delete public sourcemaps in the final `close` hook, after Nitro has generated its public-asset manifest. The resulting server bundle still lists the deleted `.js.map` files, so requests for discovered sourcemap URLs fail with `ENOENT` and return 500 instead of 404. Fixes #4275.

## Changes

- Upload and delete public sourcemaps in `nitro:build:public-assets`, before Nitro scans the public directory for its manifest.
- Keep the final whole-output upload so server and preset-specific sourcemaps are still handled.
- If the early public upload fails, disable deletion in the final fallback upload so manifest-listed files remain available rather than becoming stale.
- Preserve the existing single final upload when `deleteAfterUpload` is disabled.
- Add regression coverage for SSR and SPA builds, retained sourcemaps, and the failed-early-upload fallback.
- Add an `@posthog/nuxt` patch changeset.

A local node-server integration build confirmed that public maps and manifest entries are both removed and a request for `<chunk>.js.map` returns 404 without an `ENOENT` error.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent using the `karpathy-guidelines` and `pr` skills. The implementation was kept within the existing Nuxt lifecycle: early deletion is limited to the configuration that requests deletion, while the existing whole-output scan remains for preset compatibility. The fallback intentionally retains maps if early cleanup fails so Nitro cannot ship references to files deleted later.

### DIFF
--- a/.changeset/early-nuxt-maps.md
+++ b/.changeset/early-nuxt-maps.md
@@ -1,0 +1,5 @@
+---
+'@posthog/nuxt': patch
+---
+
+Delete uploaded public sourcemaps before Nitro generates its asset manifest to prevent stale entries from causing runtime 500 errors.
--- a/packages/nuxt/src/module.ts
+++ b/packages/nuxt/src/module.ts
@@ -140,6 +140,7 @@ export default defineNuxtModule<ModuleOptions>({
     })
 
     let isBuildProcess = false
+    let publicSourcemapsDeleted = false
 
     const posthogCliRunner = () => {
       const cliBinaryPath =
@@ -175,6 +176,11 @@ export default defineNuxtModule<ModuleOptions>({
         // Inject public sourcemaps
         // This cannot be done in the close hook. https://github.com/PostHog/posthog/issues/30957#issuecomment-2824545454
         await cliRunner(getInjectArgs(publicDir, sourcemapsConfig))
+        if (sourcemapsConfig.deleteAfterUpload ?? true) {
+          // Delete public sourcemaps before Nitro generates its asset manifest.
+          await cliRunner(getUploadArgs(publicDir, sourcemapsConfig))
+          publicSourcemapsDeleted = true
+        }
       } catch (error) {
         console.error('Failed to process public sourcemaps:', error)
       }
@@ -190,8 +196,13 @@ export default defineNuxtModule<ModuleOptions>({
         if (nuxt.options.ssr !== false) {
           await cliRunner(getInjectArgs(serverDir, sourcemapsConfig))
         }
-        // Upload all assets (public + any server output that exists)
-        await cliRunner(getUploadArgs(outputDir, sourcemapsConfig))
+        // Upload all assets (public + any server output that exists). If the early public upload failed,
+        // keep its sourcemaps so Nitro's manifest does not point to files deleted after it was generated.
+        const outputSourcemapsConfig =
+          (sourcemapsConfig.deleteAfterUpload ?? true) && !publicSourcemapsDeleted
+            ? { ...sourcemapsConfig, deleteAfterUpload: false }
+            : sourcemapsConfig
+        await cliRunner(getUploadArgs(outputDir, outputSourcemapsConfig))
       } catch (error) {
         console.error('Failed to process or upload sourcemaps:', error)
       }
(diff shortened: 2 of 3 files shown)


ioannisj (approved):

marandaneto:
@PostHog/team-error-tracking wanna double check this?

ablaszkiewicz (approved):
Ah yes nuxt is quite special... LGTM

PR #4283, opened 2026-07-27
fix(docs): restore browser-common types missing from the public API reference

ryanwaits:
## Problem

#4269 moved the browser config and survey types into `@posthog/browser-common` and re-exported them, so the
public API didn't change but the generated reference went from 107 types down to 67 at [`1abdbfd`](https://github.com/PostHog/posthog-js/commit/1abdbfd).

```bash
git show 1abdbfd^:packages/browser/references/posthog-js-references-latest.json | jq '.types | length'   # 107
git show origin/main:packages/browser/references/posthog-js-references-latest.json | jq '.types | length' #  67
```

All 40 are still public exports, and 13 are still referenced by types that stayed, so `Survey` shows up 9 times in the published reference with nothing defining it.

Nothing's wrong with the PR itself: api-extractor builds its doc model from one package's rollup, so a type re-exported from a sibling package leaves the model while staying in the public API. CI didn't catch it because `check-public-api.js` regenerates and diffs against what's committed, which compares the output to itself.

## Changes

One line in `packages/browser/api-extractor.json` plus the regenerated reference, which puts it back to 107.
This is what the other two packages already do: `packages/node/api-extractor.jsonc:70` and `packages/react-native/api-extractor.json:70` both bundle `@posthog/core`. Browser was the only one left on `[]`, which is why it was the one that broke.

```json
"bundledPackages": ["@posthog/browser-common"],
```

Checked on a clean checkout of main:

- restored set matches `1abdbfd^` exactly. 40 back, nothing extra
- the 67 that survived are unchanged apart from `PostHogConfig` property order
- classes and methods sections byte-identical
- `pnpm test:docs-scripts` 35 pass, `pnpm check:public-api` green, and a second regenerate is byte-identical
- nothing shipped changes. api-extractor only runs in `generate-references`, published types come from the rollup, and `references/` isn't in `files`

Two small things I noticed but didn't touch:

- the restored entries have `path` values like `../browser-common/dist/types/surveys.d.ts` where existing ones
  use `lib/src/types.d.ts`. If the docs site builds source links off `path`, that wants a mapping first. Say the word and I'll add it.
- `packages/browser-common/src/types/surveys.ts:270` reads `@default StringMatching.Exact`, but the type is `ActionStepStringMatching`. Pre-existing, just visible again now that `ActionStepType` is back in the reference. Left it alone since fixing it means bumping browser-common.

## Not in this PR

Same mechanism, older. 79 more public types re-exported from `@posthog/core` and `@posthog/types` are missing, and they aren't hypothetical: 43 of them were documented up through 1.314.0 and vanished in 1.315.0 (116 types → 72, nothing gained). `CaptureResult`, `AutocaptureConfig`, `BootstrapConfig`, `EarlyAccessFeature` and the rest have been undocumented ever since. So #4269 is the second time this has happened, not the first.

I did test it, for what it's worth: bundling all three gets 186 types, keeps all 107, and nothing degrades. But that's a much larger docs diff and a call about what the reference is meant to cover, so I'd rather leave it with you than tack it onto a regression fix. Happy to open another PR or write up what I found.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [ ] Tests for new code — no new code, the existing 35 docs-script tests pass unchanged
- [x] Accounted for the impact of any changes across different platforms — node and react-native configs untouched, their references regenerate identically
- [x] Accounted for backwards compatibility of any changes — additive to the reference, no published output changes
- [x] Took care not to unnecessarily increase the bundle size — no bundle impact, api-extractor doesn't run in the build

### If releasing new changes

- [ ] Ran `pnpm changeset` — skipped, nothing published changes. Can add an empty one if hygiene wants it.

### DIFF
--- a/packages/browser/api-extractor.json
+++ b/packages/browser/api-extractor.json
@@ -67,7 +67,7 @@
    *
    *   "bundledPackages": [ "@my-company/*" ],
    */
-  "bundledPackages": [],
+  "bundledPackages": ["@posthog/browser-common"],
 
   /**
    * Specifies what type of newlines API Extractor should use when writing output files.  By default, the output files
--- a/packages/browser/references/posthog-js-references-latest.json
+++ b/packages/browser/references/posthog-js-references-latest.json
@@ -2334,6 +2334,156 @@
   ],
   "types": [
     {
+      "id": "ActionStepStringMatching",
+      "name": "ActionStepStringMatching",
+      "properties": [],
+      "path": "../browser-common/dist/types/surveys.d.ts",
+      "example": "\"contains\" | \"exact\" | \"regex\""
+    },
+    {
+      "id": "ActionStepType",
+      "name": "ActionStepType",
+      "properties": [
+        {
+          "type": "string | null",
+          "name": "event"
+        },
+        {
+          "description": "StringMatching.Exact",
+          "type": "ActionStepStringMatching | null",
+          "name": "href_matching"
+        },
+        {
+          "type": "string | null",
+          "name": "href"
+        },
+        {
+          "description": "Property filters for action step matching.",
+          "type": "{\n        key: string;\n        value?: string | number | boolean | (string | number | boolean)[] | null;\n        operator?: PropertyMatchType;\n        type?: string;\n    }[]",
+          "name": "properties"
+        },
+        {
+          "description": "Pre-compiled regex pattern for matching selector against `$elements_chain`.",
+          "type": "string | null",
+          "name": "selector_regex"
+        },
+        {
+          "type": "string | null",
+          "name": "selector"
+        },
+        {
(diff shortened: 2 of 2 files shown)


marandaneto (approved):

marandaneto:
thanks @ryanwaits

PR #4291, opened 2026-07-28
fix(react-native): compose iOS source map upload wrappers

### LINKED ISSUE #4285: [React Native Expo] Combining PostHog and Sentry config plugins can silently ship iOS without main.jsbundle
## Summary

Using both `posthog-react-native/expo` and `@sentry/react-native/expo` in an Expo app can generate an invalid iOS “Bundle React Native code and images” command. The Release build succeeds, but `react-native-xcode.sh` never runs and the resulting IPA contains no `main.jsbundle`.

On device this presents as an indefinite native splash screen with no Hermes/React thread and no JavaScript/Sentry startup logs.

This is **not** caused by the PostHog runtime or `@posthog/react-native-plugin`. Removing only the PostHog Expo build plugin restores bundling while PostHog analytics and native session replay remain installed.

## Environment

- `posthog-react-native`: 4.58.0
- `@posthog/react-native-plugin`: 2.2.3
- `@sentry/react-native`: 7.11.0
- Expo SDK: 57.0.7
- React Native: 0.86.0
- iOS Release builds, reproduced in EAS Build and local `xcodebuild`
- Current package sources for `posthog-react-native` 4.61.0 and `@sentry/react-native` 8.20.0 appear to retain the same independent wrapper strategy

## Reproduction

Configure both plugins:

```json
{
  "expo": {
    "plugins": [
      [
        "@sentry/react-native/expo",
        {
          "url": "https://sentry.io/",
          "project": "example",
          "organization": "example"
        }
      ],
      "posthog-react-native/expo"
    ]
  }
}
```

Run:

```sh
npx expo prebuild --clean --platform ios
```

The generated Xcode bundle phase is effectively:

```sh
/bin/sh sentry-xcode.sh /bin/sh posthog-xcode.sh react-native-xcode.sh
```

## Root cause

Sentry’s wrapper sets:

```sh
REACT_NATIVE_XCODE="${1:-$REACT_NATIVE_XCODE_DEFAULT}"
```

With the composed command, `$1` is `/bin/sh`. The remaining PostHog wrapper and `react-native-xcode.sh` arguments are not preserved as a single command. Sentry therefore runs plain `/bin/sh`, which exits successfully, so Xcode reports success even though no JS bundle was generated.

`posthog-xcode.sh` never executes.

Reversing plugin order does not create a supported composition: PostHog’s current `*.sh` fallback can run the default RN script but bypasses Sentry’s wrapper/source-map upload.

## Evidence

- Downloaded EAS IPA: no `main.jsbundle`
- Matching local Release app: no `main.jsbundle`, indefinite splash, no Hermes/React runtime thread
- Remove only `posthog-react-native/expo`: generated command becomes:
  ```sh
  /bin/sh sentry-xcode.sh react-native-xcode.sh
  ```
- Corrected Release app embeds a ~15 MB `main.jsbundle`
- Five consecutive cold launches passed
- `posthog-react-native`, `@posthog/react-native-plugin`, the PostHog iOS pod, analytics, and session replay remain installed

## Related upstream work

This resembles #3682 / #3689, which fixed `posthog-xcode.sh` receiving `/bin/sh` from Expo. The remaining case is two third-party wrappers trying to own the same React Native bundle phase.

## Suggested fixes

Any of:

1. Detect another wrapper already owning the RN bundle phase and fail prebuild with a clear incompatibility error instead of nesting.
2. Provide an option to disable only the iOS bundle/source-map wrapper while retaining other platform configuration.
3. Add an explicit supported composite/upload phase that runs `react-native-xcode.sh` exactly once, then uploads the resulting map to PostHog.
4. Add a regression test with `@sentry/react-native/expo`.

At minimum, the generated Release build should not succeed when the embedded JS bundle is missing.

marandaneto:
## Problem

Expo apps using both `posthog-react-native/expo` and `@sentry/react-native/expo` can generate a nested iOS bundle command where Sentry receives `/bin/sh` instead of the next wrapper. The Release build can then succeed without running `react-native-xcode.sh`, producing an IPA without `main.jsbundle` and leaving the app stuck on its native splash screen. This fixes #4285.

## Changes

- Invoke `posthog-xcode.sh` directly so an outer wrapper receives the PostHog script as its command instead of `/bin/sh`.
- Preserve and execute the complete nested bundle command, while locating the underlying React Native script so its intermediate Hermes source map is retained.
- Resolve the React Native script through `NODE_BINARY` when an outer wrapper omits it and the default relative path is unavailable, supporting hoisted monorepos.
- Carry `skipOnConflict` through nested wrappers using an exported environment variable and migrate previously generated shell-prefixed PostHog commands.
- Declare the Xcode helper as a package bin target so package managers preserve its executable bit.
- Support both BSD and GNU `sed` when preserving React Native's intermediate Hermes source map.
- Add regression coverage for both PostHog/Sentry plugin orders, nested command execution, fallback execution, and existing-wrapper migration.
- Update the checked-in Expo 53 **Bundle React Native code and images** phase to use the composable PostHog command.
- Document in the changeset that projects with a checked-in `ios/` directory must rerun `npx expo prebuild --platform ios` after upgrading.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi resumed the [posthog-watcher investigation](https://github.com/PostHog/posthog-js/issues/4285#issuecomment-5099272970) and implemented the fix in a dedicated worktree. We rejected an upload opt-out because removing the Expo plugin already provides that behavior; instead, the wrappers now compose while preserving both vendors' source-map uploads. Focused tests, the full React Native unit suite (544 tests), package build, lint, shell syntax checks, and packed executable permissions were verified. A Linux CI failure in the new executable wrapper test exposed the existing BSD-only `sed -i ''` call; the script now selects the correct BSD/GNU syntax. Review feedback also led to separating legacy command migration from skip-on-conflict handling, resolving hoisted React Native fallbacks, and covering the combined legacy migration path.

The branch package was installed into `examples/example-expo-53`, the plugin was applied to its **Bundle React Native code and images** phase, and an XcodeBuildMCP Release simulator build completed successfully. The resulting app contains a 2.4 MB `main.jsbundle`, with the 11 MB source map retained in the build products. Xcode 26 required a temporary ignored-Pods workaround for an unrelated Expo 53/React Native 0.79 `fmt` consteval compilation failure; no workaround was checked in.

### DIFF
--- a/packages/react-native/package.json
+++ b/packages/react-native/package.json
@@ -5,6 +5,9 @@
     "url": "https://github.com/PostHog/posthog-js/issues"
   },
   "license": "MIT",
+  "bin": {
+    "posthog-react-native-xcode": "tooling/posthog-xcode.sh"
+  },
   "main": "dist/index.js",
   "types": "dist/index.d.ts",
   "exports": {
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -182,8 +182,8 @@ export function modifyExistingXcodeBuildScript(script: BuildPhase | undefined, s
   }
 
   if (script.shellScript.includes('posthog-xcode.sh')) {
-    const code = JSON.parse(script.shellScript)
-    script.shellScript = JSON.stringify(updatePostHogSkipOnConflictArg(code, skipOnConflict))
+    const code = migrateLegacyPostHogWrapperInvocation(JSON.parse(script.shellScript))
+    script.shellScript = JSON.stringify(updatePostHogSkipOnConflict(code, skipOnConflict))
     return
   }
 
@@ -205,10 +205,13 @@ const POSTHOG_SKIP_ON_CONFLICT_EXPORT = 'export POSTHOG_SKIP_ON_CONFLICT=1'
 const REACT_NATIVE_XCODE_LINE =
   /^([ \t]*)(?![A-Za-z_][A-Za-z0-9_]*=)([^\n]*(?:packager|scripts)\/react-native-xcode\.sh\b[^\n]*)$/m
 
-function updatePostHogSkipOnConflictArg(script: string, skipOnConflict: boolean): string {
+function migrateLegacyPostHogWrapperInvocation(script: string): string {
+  return script.replace(`/bin/sh ${POSTHOG_REACT_NATIVE_XCODE_PATH}`, POSTHOG_REACT_NATIVE_XCODE_PATH)
+}
+
+function updatePostHogSkipOnConflict(script: string, skipOnConflict: boolean): string {
   const skipArg = '--posthog-skip-on-conflict --'
   const lines = script
-    .replace(`/bin/sh ${POSTHOG_REACT_NATIVE_XCODE_PATH}`, POSTHOG_REACT_NATIVE_XCODE_PATH)
     .replace(new RegExp(`\\s*${skipArg}\\s*`, 'g'), ' ')
     .split('\n')
     .filter((line) => line.trim() !== POSTHOG_SKIP_ON_CONFLICT_EXPORT)
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -195,32 +195,43 @@ export function modifyExistingXcodeBuildScript(script: BuildPhase | undefined, s
   script.shellScript = JSON.stringify(addPostHogWithBundledScriptsToBundleShellScript(code, skipOnConflict))
 }
 
+// Invoked directly so another wrapper receives this script—not /bin/sh—as $1.
+// package.json declares it as a bin target so package managers preserve its executable bit.
 const POSTHOG_REACT_NATIVE_XCODE_PATH =
   "`\"$NODE_BINARY\" --print \"require('path').join(require('path').dirname(require.resolve('posthog-react-native')), '..', 'tooling', 'posthog-xcode.sh')\"`"
 
+const POSTHOG_SKIP_ON_CONFLICT_EXPORT = 'export POSTHOG_SKIP_ON_CONFLICT=1'
+
 const REACT_NATIVE_XCODE_LINE =
-  /^([ \t]*)(?![A-Za-z_][A-Za-z0-9_]*=)(?:\/bin\/sh\s+)?([^\n]*(?:packager|scripts)\/react-native-xcode\.sh\b[^\n]*)$/m
+  /^([ \t]*)(?![A-Za-z_][A-Za-z0-9_]*=)([^\n]*(?:packager|scripts)\/react-native-xcode\.sh\b[^\n]*)$/m
 
 function updatePostHogSkipOnConflictArg(script: string, skipOnConflict: boolean): string {
   const skipArg = '--posthog-skip-on-conflict --'
-  const withoutSkipArg = script.replace(new RegExp(`\\s*${skipArg}\\s*`, 'g'), ' ')
-  if (!skipOnConflict) {
-    return withoutSkipArg
+  const lines = script
+    .replace(`/bin/sh ${POSTHOG_REACT_NATIVE_XCODE_PATH}`, POSTHOG_REACT_NATIVE_XCODE_PATH)
+    .replace(new RegExp(`\\s*${skipArg}\\s*`, 'g'), ' ')
+    .split('\n')
+    .filter((line) => line.trim() !== POSTHOG_SKIP_ON_CONFLICT_EXPORT)
+
+  if (skipOnConflict) {
+    const commandIndex = lines.findIndex((line) => line.includes(POSTHOG_REACT_NATIVE_XCODE_PATH))
+    if (commandIndex !== -1) {
+      const indent = lines[commandIndex].match(/^[ \t]*/)?.[0] ?? ''
+      lines.splice(commandIndex, 0, `${indent}${POSTHOG_SKIP_ON_CONFLICT_EXPORT}`)
+    }
   }
-  return withoutSkipArg.replace(`${POSTHOG_REACT_NATIVE_XCODE_PATH} `, `${POSTHOG_REACT_NATIVE_XCODE_PATH} ${skipArg} `)
+
+  return lines.join('\n')
 }
 
 export function addPostHogWithBundledScriptsToBundleShellScript(script: string, skipOnConflict = false): string {
-  const postHogArgs = skipOnConflict ? '--posthog-skip-on-conflict -- ' : ''
-
   // Capture the full RN script invocation. Expo uses a backtick-wrapped
   // node --print command, so matching only up to react-native-xcode.sh cuts the
   // command substitution in half and leaves the generated shell invalid.
-  return script.replace(
-    REACT_NATIVE_XCODE_LINE,
-    (_match: string, indent: string, rnCommand: string) =>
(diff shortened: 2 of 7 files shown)


marandaneto:
The checked-in Expo 53 Xcode project now demonstrates the config plugin's bundle-phase migration.

**Before**

```sh
/bin/sh `"$NODE_BINARY" --print "require('path').join(require('path').dirname(require.resolve('posthog-react-native')), '..', 'tooling', 'posthog-xcode.sh')"` `"$NODE_BINARY" --print "require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'"`
```

**After**

```sh
`"$NODE_BINARY" --print "require('path').join(require('path').dirname(require.resolve('posthog-react-native')), '..', 'tooling', 'posthog-xcode.sh')"` `"$NODE_BINARY" --print "require('path').dirname(require.resolve('react-native/package.json')) + '/scripts/react-native-xcode.sh'"`
```

The command remains in **Bundle React Native code and images**. Re-running the Expo config plugin automatically migrates an existing shell-prefixed command to the new format, so users do not need to edit `project.pbxproj` manually. Projects with a committed native `ios/` directory must rerun `npx expo prebuild --platform ios` after upgrading.

I also verified the migrated phase with a Release simulator build of `examples/example-expo-53`; the app bundle and composed source map were generated, and the source-map artifacts appeared in PostHog.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -195,32 +195,43 @@ export function modifyExistingXcodeBuildScript(script: BuildPhase | undefined, s
...
 const POSTHOG_REACT_NATIVE_XCODE_PATH =
   "`\"$NODE_BINARY\" --print \"require('path').join(require('path').dirname(require.resolve('posthog-react-native')), '..', 'tooling', 'posthog-xcode.sh')\"`"
 
+const POSTHOG_SKIP_ON_CONFLICT_EXPORT = 'export POSTHOG_SKIP_ON_CONFLICT=1'
+
 const REACT_NATIVE_XCODE_LINE =
-  /^([ \t]*)(?![A-Za-z_][A-Za-z0-9_]*=)(?:\/bin\/sh\s+)?([^\n]*(?:packager|scripts)\/react-native-xcode\.sh\b[^\n]*)$/m
+  /^([ \t]*)(?![A-Za-z_][A-Za-z0-9_]*=)([^\n]*(?:packager|scripts)\/react-native-xcode\.sh\b[^\n]*)$/m
 
 function updatePostHogSkipOnConflictArg(script: string, skipOnConflict: boolean): string {
```
turnipdabeets: [suggestion] This function now also migrates the legacy `/bin/sh`-prefixed invocation, not just the skip-on-conflict setting — but the name is where a future reader would least expect the migration to live. Could we rename it (e.g. `migratePostHogWrapperInvocation`) or split the legacy-form cleanup from the env-var toggle?

packages/react-native/package.json
```diff
@@ -5,6 +5,9 @@
     "url": "https://github.com/PostHog/posthog-js/issues"
   },
   "license": "MIT",
+  "bin": {
+    "posthog-react-native-xcode": "tooling/posthog-xcode.sh"
```
turnipdabeets, line 9: [question] This puts `posthog-react-native-xcode` in every consumer's `node_modules/.bin` permanently — removing or renaming it later would be a breaking change. It's the standard way to guarantee the executable bit, so likely deliberate — just confirming we're happy committing to the name.

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -121,13 +131,23 @@ fi
...
-# Execute React Native Xcode script and check exit code
+# Execute the complete bundle command so other source-map wrappers keep their
+# script path and arguments. Fall back to the standard RN script when an outer
+# wrapper invokes posthog-xcode.sh without forwarding its remaining arguments.
 set +x +e # disable printing commands and allow continuing on error
-RN_XCODE_OUTPUT=$(/bin/sh -c "$REACT_NATIVE_XCODE" 2>&1)
+if [ "$#" -gt 0 ]; then
+  RN_XCODE_OUTPUT=$("$@" 2>&1)
+else
+  RN_XCODE_OUTPUT=$(/bin/sh "$REACT_NATIVE_XCODE_DEFAULT" 2>&1)
```
turnipdabeets, line 161: [suggestion] When Sentry is the outer wrapper this fallback always runs (sentry-xcode.sh only forwards its `$1`), and `../node_modules/react-native/...` doesn't exist in hoisted-workspace monorepos. Could we resolve the script via `"$NODE_BINARY" --print "require.resolve('react-native/package.json')"` when the default path is missing? Fine as a follow-up — this order previously broke outright, so it's strictly an improvement either way.

packages/react-native/test/expoconfig.spec.ts
```diff
@@ -153,19 +163,31 @@ describe('modifyExistingXcodeBuildScript', () => {
...
-    expect(parsed).toContain('--posthog-skip-on-conflict --')
+    expect(parsed).toContain('export POSTHOG_SKIP_ON_CONFLICT=1')
 
     modifyExistingXcodeBuildScript(script, false)
     parsed = JSON.parse(script.shellScript)
-    expect(parsed).not.toContain('--posthog-skip-on-conflict --')
+    expect(parsed).not.toContain('POSTHOG_SKIP_ON_CONFLICT')
+  })
+
+  it('migrates an existing shell-prefixed PostHog wrapper to a composable invocation', () => {
```
turnipdabeets, line 181: [nit] Worth one more case: the combined legacy form `/bin/sh <path> --posthog-skip-on-conflict -- <rn>` with `skipOnConflict: true`. The prefix strip, arg strip, and export insertion are each covered separately, but not all on one line.

turnipdabeets (approved):

packages/react-native/test/expoconfig.spec.ts
marandaneto: Added in c86f28995. The new test starts with the combined legacy `/bin/sh` prefix and `--posthog-skip-on-conflict --` argument, then verifies both are migrated and the exported setting is present.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Good catch. Fixed in c86f28995: when the outer wrapper omits the RN argument and the relative default is missing, the script now resolves `react-native/package.json` through `NODE_BINARY`. The executable regression test uses a hoisted fixture to cover this path.

packages/react-native/package.json
marandaneto: Yes, this is deliberate. Direct wrapper composition requires the installed helper to retain its executable bit, and declaring the bin target is the package-manager-supported way to guarantee that. We are treating `posthog-react-native-xcode` as a supported package command.

packages/react-native/src/tooling/expoconfig.ts
marandaneto: Good call. Fixed in c86f28995: legacy `/bin/sh` migration now lives in `migrateLegacyPostHogWrapperInvocation`, separate from skip-on-conflict updates.

PR #4294, opened 2026-07-28
fix(react-native): don't nest the Expo iOS bundle wrapper inside another tool's wrapper

### LINKED ISSUE #4285: [React Native Expo] Combining PostHog and Sentry config plugins can silently ship iOS without main.jsbundle

Mohith26:
Fixes #4285

## What

When the PostHog Expo config plugin and Sentry's plugin both wrap the iOS "Bundle React Native code and images" build phase, the PostHog wrapper could end up nested inside the other tool's wrapper. In that arrangement the bundle step can be skipped silently and the app ships without main.jsbundle.

The plugin now composes with existing build-phase content instead of re-wrapping it, so plugin order no longer matters.

## Tests

Fixture-based unit tests simulate a project where Sentry's plugin already modified the build phase, in both plugin orders. They fail on main (8 failures) and pass with the fix; the full react-native package suite shows no new failures.

Credit to @benschac for the report in #4285.

### DIFF
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -192,6 +192,29 @@ export function modifyExistingXcodeBuildScript(script: BuildPhase | undefined, s
   }
 
   const code = JSON.parse(script.shellScript)
+
+  // Another config plugin (e.g. @sentry/react-native/expo) may already wrap the
+  // react-native-xcode.sh invocation with its own script. Nesting a second wrapper
+  // produces a command like:
+  //   /bin/sh sentry-xcode.sh /bin/sh posthog-xcode.sh react-native-xcode.sh
+  // where the outer wrapper resolves its script argument to `/bin/sh`, exits
+  // successfully without ever bundling, and the Release app silently ships
+  // without main.jsbundle (issue #4285). Leave the phase untouched instead.
+  const foreignWrapper = findForeignBundleScriptWrapper(code)
+  if (foreignWrapper) {
+    console.warn(
+      `[posthog-react-native] Another tool's script ('${foreignWrapper}') already wraps the ` +
+        "'Bundle React Native code and images' build phase. Nesting a second wrapper can " +
+        'silently produce an iOS app without main.jsbundle ' +
+        '(https://github.com/PostHog/posthog-js/issues/4285), so the PostHog iOS sourcemap ' +
+        'upload wrapper was not installed. React Native bundling and the existing wrapper are ' +
+        'left untouched; PostHog analytics and session replay are unaffected. To upload iOS ' +
+        "sourcemaps to PostHog instead, remove the other plugin's bundle-phase integration " +
+        'and re-run prebuild.'
+    )
+    return
+  }
+
   script.shellScript = JSON.stringify(addPostHogWithBundledScriptsToBundleShellScript(code, skipOnConflict))
 }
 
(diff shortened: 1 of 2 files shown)


marandaneto:
already fixed

PR #4295, opened 2026-07-28
ci: install npm OIDC support only when publishing

marandaneto:
## Problem

The shared workspace setup action installs `npm@11.6.4` on every invocation, even though modern npm is only required for OIDC publishing. This repeats an unnecessary global npm installation across roughly 50 CI jobs.

## Changes

- Remove the npm upgrade from the shared setup action.
- Install `npm@11.6.4` in the release publish job only when the package version is new and will be published.
- Validate the modified action and workflow as YAML and with `git diff --check`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [ ] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with Pi in a dedicated worktree. The npm upgrade was moved rather than parameterized so it runs only after the release workflow confirms that a package has a new version to publish. No SDK package changes or changeset are required.

### DIFF
--- a/.github/actions/setup/action.yaml
+++ b/.github/actions/setup/action.yaml
@@ -28,11 +28,6 @@ runs:
         cache: false
         install: false
 
-    # Ensure latest npm with OIDC support (npm 11.5.1+)
-    - name: Ensure modern npm (OIDC support)
-      shell: bash
-      run: npm install -g npm@11.6.4
-
     - name: Install package.json dependencies with pnpm
       if: ${{ inputs.install == 'true' }}
       shell: bash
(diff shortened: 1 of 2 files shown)


turnipdabeets (approved):
Verified release.yml is the only workflow that publishes to npm and the new step lands in the same job, gated on is-new-version and ordered before the publish. Nice cleanup.

PR #4300, opened 2026-07-28
feat(browser-common): stream remote config outcomes

dustinbyrne:
## Problem

Shared browser extensions need to observe both successful and failed remote-config loads without racing an initial read against listener registration. The shared client contract currently only streams successful values and does not replay the latest outcome to late subscribers.

## Changes

- Stream `RemoteConfigResult` outcomes through the shared client contract.
- Replay the latest outcome immediately when a listener subscribes.
- Update the listener helper, test client, extension-authoring guidance, and README.
- Add a patch changeset for `@posthog/browser-common`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Prepared with the Pi coding agent and its code-review and SDK-review workflows in a local session. The shared contract was separated from the browser-v1 implementation so each package change can be reviewed and released independently.

### DIFF
--- a/.changeset/browser-common-remote-config-outcomes.md
+++ b/.changeset/browser-common-remote-config-outcomes.md
@@ -1,5 +1,5 @@
 ---
-'@posthog/browser-common': patch
+'@posthog/browser-common': minor
 ---
 
 Stream replayable remote config outcomes through shared browser extension clients.
--- a/.changeset/browser-common-remote-config-outcomes.md
+++ b/.changeset/browser-common-remote-config-outcomes.md
@@ -1,0 +1,5 @@
+---
+'@posthog/browser-common': patch
+---
+
+Stream replayable remote config outcomes through shared browser extension clients.
--- a/packages/browser-common/src/client.ts
+++ b/packages/browser-common/src/client.ts
@@ -4,7 +4,7 @@ import type { Properties } from '@posthog/types'
 import type { Disposable } from './disposable'
 import type { KeyValueStore } from './persistence'
 import type { Listener } from './pubsub'
-import type { RemoteConfig } from './types/remote-config'
+import type { RemoteConfigResult } from './types/remote-config'
 
 /** Recursively marks object properties as readonly while preserving callable values. */
 export type DeepReadonly<T> = T extends (...args: never[]) => unknown
@@ -103,10 +103,8 @@ export interface Client {
     /** Fires for every captured event through a deeply readonly view. */
     readonly onEvent: Listener<CapturedEventInfo>
 
-    /** Resolves with the current remote config, awaiting the first outcome when necessary. */
-    getRemoteConfig(): Promise<DeepReadonly<RemoteConfig> | undefined>
-    /** Fires when server-provided config arrives or changes successfully. */
-    readonly onRemoteConfig: Listener<DeepReadonly<RemoteConfig>>
+    /** Replays the latest remote-config outcome on subscription and fires for subsequent outcomes. */
+    readonly onRemoteConfig: Listener<DeepReadonly<RemoteConfigResult>>
 
     /** Public project token used to authenticate endpoint-specific requests. */
     readonly projectToken: string
--- a/packages/browser-common/src/types/remote-config.ts
+++ b/packages/browser-common/src/types/remote-config.ts
@@ -206,3 +206,6 @@ export interface RemoteConfig {
     /** Conversations widget configuration. */
     conversations?: boolean | ConversationsRemoteConfig
 }
+
+/** The latest outcome of loading remote configuration from PostHog. */
+export type RemoteConfigResult = { ok: true; config: RemoteConfig } | { ok: false }
--- a/packages/browser-common/tests/helpers/test-client.ts
+++ b/packages/browser-common/tests/helpers/test-client.ts
@@ -87,16 +87,9 @@ export class TestClient implements Client {
 
     readonly onEvent = this._eventPublisher.listener
     readonly onRemoteConfig: Client['onRemoteConfig'] = (handler) => {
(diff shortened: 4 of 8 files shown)


.changeset/browser-common-remote-config-outcomes.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/browser-common': patch
```
turnipdabeets: Removing `getRemoteConfig` and changing the `onRemoteConfig` payload breaks compilation for typed consumers of 0.2.x, and caret ranges auto-accept a 0.2.5 patch but not a 0.3.0 — so as a patch the break flows into dependents automatically. I think this needs to be a minor bump (this is also what greptile flagged):
```suggestion
'@posthog/browser-common': minor
```

packages/browser-common/tests/helpers/test-client.ts
```diff
@@ -79,14 +79,27 @@ export class TestClient implements Client {
...
+    private _remoteConfigResult: RemoteConfigResult | undefined
     private _requestResponse: ApiResponse
     private _dynamicEventPropertyProducers: Array<() => Record<string, unknown>> = []
     private _eventPublisher = new Publisher<CapturedEventInfo>()
-    private _remoteConfigPublisher = new Publisher<RemoteConfig>()
+    private _remoteConfigPublisher = new Publisher<RemoteConfigResult>()
 
     readonly onEvent = this._eventPublisher.listener
-    readonly onRemoteConfig = this._remoteConfigPublisher.listener
+    readonly onRemoteConfig: Client['onRemoteConfig'] = (handler) => {
```
turnipdabeets, line 89: [suggestion] This replay + try/catch wrapper gets hand-rolled again almost verbatim in `BrowserClientAdapter` in #4301, with the replay semantics pinned only by doc comments. Could we encode it once in `pubsub.ts` — a `StatefulPublisher` or a `replayLatest` option — so both implementations share it and it's tested in `pubsub.spec.ts`?

packages/browser-common/src/client.ts
```diff
@@ -103,10 +103,8 @@ export interface Client {
     /** Fires for every captured event through a deeply readonly view. */
     readonly onEvent: Listener<CapturedEventInfo>
 
-    /** Resolves with the current remote config, awaiting the first outcome when necessary. */
-    getRemoteConfig(): Promise<DeepReadonly<RemoteConfig> | undefined>
-    /** Fires when server-provided config arrives or changes successfully. */
-    readonly onRemoteConfig: Listener<DeepReadonly<RemoteConfig>>
+    /** Replays the latest remote-config outcome on subscription and fires for subsequent outcomes. */
```
turnipdabeets, line 106: [suggestion] Replay-on-subscribe is only exercised by #4301's tests — if browser-common releases independently, nothing in-package asserts this promise. Could we add a small spec in this PR? (Falls out for free if replay moves into pubsub.)

packages/browser-common/tests/helpers/test-client.ts
```diff
@@ -79,14 +79,27 @@ export class TestClient implements Client {
...
+        const invoke = (result: RemoteConfigResult): void => {
+            try {
+                handler(result)
+            } catch (error) {
+                this.logger.error('Remote config listener failed', error)
```
turnipdabeets: [nit] `logger` defaults to `noopLogger`, so a handler that throws in a downstream package's test is swallowed silently. Worth rethrowing in the test helper (or defaulting to a loud logger) so failures stay visible?

packages/browser-common/src/types/remote-config.ts
```diff
@@ -206,3 +206,6 @@ export interface RemoteConfig {
     /** Conversations widget configuration. */
     conversations?: boolean | ConversationsRemoteConfig
 }
+
+/** The latest outcome of loading remote configuration from PostHog. */
+export type RemoteConfigResult = { ok: true; config: RemoteConfig } | { ok: false }
```
turnipdabeets, line 211: [question] Is `{ ok: false }` carrying no error/reason deliberate? Adding a field later is non-breaking, so starting lean seems right — just checking it's intentional and not something #4301 already wants.

packages/browser-common/tests/helpers/test-client.ts
dustinbyrne: it's written once in the shipped bundle and once in the test. given that, i don't think it's worth the bundled byte cost of adding an abstraction

packages/browser-common/tests/helpers/test-client.ts
dustinbyrne: will remove the try/catch

.changeset/browser-common-remote-config-outcomes.md
dustinbyrne: i've resolved greptiles comment because the consumer isn't introduced until the next pr, but will update anyway

packages/browser-common/src/client.ts
dustinbyrne: Added focused browser-common coverage for no initial outcome, replaying the latest failure to late subscribers, subsequent successful outcomes, and propagated listener failures in `packages/browser-common/tests/test-client.spec.ts`.

turnipdabeets (approved):

PR #4301, opened 2026-07-28
feat(browser): add shared extension client adapter

dustinbyrne:
## Problem

Shared browser extensions need a browser-v1 client implementation that routes analytics, persistence, requests, events, remote config, and lifecycle cleanup through the existing `PostHog` instance.

Stacked on #4300, which updates the shared remote-config contract.

## Changes

- Add a `BrowserClientAdapter` backed by one extension runtime per `PostHog` instance.
- Adapt capture options, identity and session state, persistence, transport requests, captured events, and replayable remote-config outcomes.
- Register dynamic event properties and clean up extensions synchronously during shutdown.
- Add focused coverage for adapter behavior, initialization, remote config, and shutdown.
- Add a patch changeset for `posthog-js`.

Fast esbuild bundle comparison against the stacked browser-common base: +4.69 KiB minified (+1.82%), +1.23 KiB gzip (+1.55%), and +0.99 KiB Brotli (+1.45%).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Prepared with the Pi coding agent and its code-review and SDK-review workflows in a local session. The adapter is stacked on the browser-common remote-config contract so the shared package and browser-v1 integration remain independently reviewable.


packages/browser/src/__tests__/extensions/replay/sessionRecording-onRemoteConfig.test.ts
```diff
@@ -147,6 +147,7 @@ describe('SessionRecording', () => {
                 },
             } as unknown as ConsentManager,
             register_for_session() {},
+            _onRemoteConfig: jest.fn(),
```
turnipdabeets, line 150: [nit] Couldn't find anything in the replay path that reads `_onRemoteConfig` off the instance — is this mock entry a leftover from an earlier iteration?

packages/browser/src/extensions/browser-client.ts
```diff
@@ -0,0 +1,195 @@
...
+    private readonly _runtime: ExtensionRuntime
+    private _latestRemoteConfigResult: RemoteConfigResult | undefined
+    private _disposed = false
+
+    constructor(readonly instance: PostHog) {
+        this._logger = logger.createLogger('[BrowserExtensions]')
+        this._latestRemoteConfigResult =
+            instance._lastRemoteConfig ?? (instance._shouldDisableFlags() ? { ok: false } : undefined)
+        this.kv = new BrowserExtensionKeyValueStore(instance)
+        this.onEvent = (handler) => {
```
turnipdabeets, line 64: [nit] `onRemoteConfig` returns a no-op disposable once `_disposed` and `handleRemoteConfig` early-returns, but `onEvent` has no `_disposed` guard — a post-dispose subscription still attaches to the live instance emitter and fires on later captures. Worth the same guard for consistency?

packages/browser/src/extensions/browser-client.ts
```diff
@@ -0,0 +1,195 @@
...
+} from '../types'
+
+class BrowserExtensionKeyValueStore implements KeyValueStore {
+    constructor(private readonly _instance: PostHog) {}
+
+    get<T = unknown>(prop: string): T | undefined {
+        return this._instance.persistence?.get_property(prop) as T | undefined
+    }
+
+    set(prop: string, value: unknown): void {
```
turnipdabeets, line 38: [question] Direct keys land in host persistence unprefixed, so extension state shows up as a property on every captured event (unless denylisted) and can collide with reserved keys like `distinct_id`. The exposure/collision test reads as deliberate — is "extensions namespace their own keys" the intent, or should the adapter enforce a prefix?

packages/browser/src/extensions/browser-client.ts
dustinbyrne: i'll probably end up cutting that from `onRemoteConfig`, i don't think we need to be so defensive and i've been fighting the models on this

`posthog-js` relies a lot on correctness over defensiveness which is (imo) deliberate to keep compiled bundle sizes down where possible

turnipdabeets (approved):

packages/browser/src/__tests__/extensions/replay/sessionRecording-onRemoteConfig.test.ts
dustinbyrne: This is reached indirectly by the stale persisted-config path: `SessionRecording._onScriptLoaded()` creates a `RemoteConfigLoader`, whose failure path calls `instance._onRemoteConfig(...)`. Removing the mock still leaves the tests green because that resulting `TypeError` is caught and logged by the loader, so I’m keeping it to avoid silently exercising an invalid mock.

PR #4306, opened 2026-07-29
fix(ai): prefer resolved OpenAI service tier

marandaneto:
## Problem

OpenAI can resolve a requested `service_tier: 'auto'` to a different processing tier such as `default` or `flex`. The request-only capture added in #4072 therefore cannot always identify the tier that was actually billed, which can lead to inaccurate cost attribution.

## Changes

- Prefer the latest non-null `service_tier` reported by successful OpenAI responses while retaining the requested tier as a fallback.
- Apply the behavior to OpenAI and Azure OpenAI Chat Completions and Responses API paths, including streaming, non-streaming, and parsed responses.
- Keep error captures on the requested tier when a completed response tier is unavailable.
- Add regression coverage for response precedence, streaming final-tier precedence, and null or missing response fallback.
- Add a dedicated `@posthog/ai` changeset for the resolved-tier behavior without modifying the changeset from #4072.

Follow-up to #4072 and discussion https://github.com/PostHog/posthog-js/pull/4072#discussion_r3636466279.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent and independently reviewed with its reviewer subagent. The implementation prefers response metadata only when non-null, keeps request metadata as the compatibility fallback, and covers both OpenAI and Azure OpenAI wrapper paths.

### DIFF
--- a/.changeset/ai-openai-service-tier-model-params.md
+++ b/.changeset/ai-openai-service-tier-model-params.md
@@ -2,4 +2,4 @@
 '@posthog/ai': patch
 ---
 
-fix(ai): capture OpenAI `service_tier` in `$ai_model_parameters`, preferring the tier reported by the response so PostHog can accurately attribute costs when `auto` resolves to another tier
+fix(ai): capture OpenAI `service_tier` in `$ai_model_parameters` so PostHog can correctly attribute costs for flex and priority tier requests
--- a/.changeset/ai-openai-service-tier-model-params.md
+++ b/.changeset/ai-openai-service-tier-model-params.md
@@ -2,4 +2,4 @@
 '@posthog/ai': patch
 ---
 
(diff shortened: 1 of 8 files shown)


dustinbyrne (approved):
makes sense

PR #4312, opened 2026-07-29
fix(surveys): keep hosted-survey prefill metadata and prevent thumbs loss

darkopia:
## Problem

Hosted (external) surveys opened via a prefilled deep link (e.g. a thumbs-up/down question set to "Automatically submit on selection" plus an optional follow-up, linked from an email with extra query params) had two related defects:

1. **Extra URL params were dropped on auto-submit.** The hosted survey page forwards non-question query params (things like `account_number`, `month`, `year`) to `renderSurvey` as the `properties` argument. When prefill auto-submitted the answer, those properties were missing from the resulting `survey sent` event — only the answer and `distinct_id` came through.
2. **The prefilled answer was cleared if the respondent then submitted manually.** After the auto-submit, if the user typed into the optional follow-up and pressed Submit, the follow-up + metadata were saved but the originally auto-captured answer disappeared from the response.

**Why:** both break the "capture immediately, but still let people add a comment" flow that hosted surveys are meant to support — the first response is recorded without its metadata, and completing the survey silently wipes the first answer.

## Changes

- Thread the caller-provided `properties` through `renderSurvey` → `_handleUrlPrefill` into the auto-submit `sendSurveyEvent`, so extra URL params land on the auto-submitted response (they already flowed through the manual-submit and `initialResponses` paths).
- Persist the auto-advanced (prefilled) question indices as `visitedIndices` in the in-progress survey state. The manual-submit path prunes the payload to "responses on the visited path"; because prefill never recorded those indices, the prefilled answer was pruned out, and since partial responses reuse the same `$survey_submission_id`, the merge cleared it. `calculatePrefillStartIndex` now also returns the skipped indices for this.
- Unit tests for both: auto-submit includes passed properties, and prefill persists the visited indices.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — DRI: Ben Lea (ben.l@posthog.com)

Investigated a support report about hosted-survey responses, traced the two failure modes to the URL-prefill path in `packages/browser/src/extensions/surveys.tsx`, and confirmed the mechanism by reading the source (auto-submit `sendSurveyEvent` missing `properties`; prefill not persisting `visitedIndices`). The behavior was reproduced end-to-end on a test external survey — the auto-submit event carried the answer but not the metadata, and the follow-up manual submit produced a second event under the same submission id with the answer cleared.

Fixes verified with the survey unit suites (`survey-url-prefill.test.ts`, `surveys.test.ts`, 156 passing) and eslint on the changed files. No manual browser click-through was run in the agent environment.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0A2AHK6CN8/p1785336747899189)*

### DIFF
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -7,7 +7,7 @@ import {
     useHideSurveyOnURLChange,
     usePopupVisibility,
 } from '../../extensions/surveys'
-import { retrieveSurveyShadow } from '../../extensions/surveys/surveys-extension-utils'
+import { getInProgressSurveyState, retrieveSurveyShadow } from '../../extensions/surveys/surveys-extension-utils'
 import {
     Survey,
     SurveyQuestionBranchingType,
@@ -1160,6 +1160,91 @@ describe('SurveyManager', () => {
                     $survey_completed: false,
                 })
             )
+        })
+
+        it('should include caller-provided properties in the auto-submitted prefill event', () => {
+            const survey: Survey = {
+                id: 'prefill-survey-props',
+                name: 'Prefill Survey Props',
+                type: SurveyType.Popover,
+                enable_partial_responses: true,
+                questions: [
+                    {
+                        id: 'q1',
+                        type: SurveyQuestionType.Rating,
+                        question: 'Rate us',
+                        scale: 10,
+                        skipSubmitButton: true,
+                    },
+                    {
+                        id: 'q2',
+                        type: SurveyQuestionType.Open,
+                        question: 'Any feedback?',
+                    },
+                ],
+                appearance: {},
+                conditions: null,
+                start_date: '2021-01-01T00:00:00.000Z',
+                end_date: null,
+                current_iteration: null,
+                current_iteration_start_date: null,
+                feature_flag_keys: [],
+                linked_flag_key: null,
+                targeting_flag_key: null,
+                internal_targeting_flag_key: null,
+            }
+
+            window.location.search = '?q0=8'
+            ;(surveyManager as any)._handleUrlPrefill(survey, null, { account_number: 'A12345', month: 'January' })
+
+            expect(mockPostHog.capture).toHaveBeenCalledWith(
+                'survey sent',
+                expect.objectContaining({
+                    $survey_id: 'prefill-survey-props',
+                    $survey_response_q1: 8,
+                    account_number: 'A12345',
+                    month: 'January',
+                })
+            )
+        })
+
+        it('should persist auto-advanced questions as visited so a later manual submit keeps prefilled answers', () => {
+            const survey: Survey = {
+                id: 'prefill-survey-visited',
+                name: 'Prefill Survey Visited',
+                type: SurveyType.Popover,
+                enable_partial_responses: true,
+                questions: [
+                    {
+                        id: 'q1',
+                        type: SurveyQuestionType.Rating,
+                        question: 'Rate us',
+                        scale: 10,
+                        skipSubmitButton: true,
+                    },
+                    {
+                        id: 'q2',
+                        type: SurveyQuestionType.Open,
+                        question: 'Any feedback?',
+                    },
+                ],
+                appearance: {},
+                conditions: null,
+                start_date: '2021-01-01T00:00:00.000Z',
+                end_date: null,
+                current_iteration: null,
+                current_iteration_start_date: null,
+                feature_flag_keys: [],
+                linked_flag_key: null,
+                targeting_flag_key: null,
+                internal_targeting_flag_key: null,
+            }
+
+            window.location.search = '?q0=8'
+            ;(surveyManager as any)._handleUrlPrefill(survey)
+
+            const state = getInProgressSurveyState(survey)
+            expect(state?.visitedIndices).toEqual([0])
         })
 
         it('should NOT auto-submit when skipSubmitButton is false, even with enable_partial_responses true', () => {
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -801,7 +801,7 @@ describe('SurveyManager', () => {
         })
     })
 
-    describe('renderSurvey with URL prefill that completes the survey', () => {
+    describe('renderSurvey with URL prefill', () => {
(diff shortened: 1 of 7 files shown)


darkopia:
## 🔍 QA team review — 💬 APPROVE WITH NITS

7 independent reviewers (2 generalists + data-integrity, compatibility, reliability, frontend, security). Overall risk **MEDIUM**, no blockers. All confirmed the two fixes are correct and address a real data-loss path.

**Findings**

| Priority | Finding | Status |
| --- | --- | --- |
| 🟡 Medium | Untrusted hosted-survey URL params flow into the auto-submit event and are spread *after* reserved `$survey_*` fields, so a crafted key can overwrite them on a public, page-load surface. Flagged independently by 3 reviewers. Pre-existing on the manual path; this PR extends it to auto-submit. | open — recommend follow-up |
| 🟢 Low | No end-to-end test for the manual-submit merge (the path the fix protects); current test covers the `visitedIndices` write side only. | open — recommend follow-up |
| 🟢 Low | Raw-index vs display-order coupling for `visitedIndices` (safe today; gated on `enable_partial_responses`). | fixed — comment added |
| 🟢 Low | Dead loop-guard in `calculatePrefillStartIndex` now also bounded the new push. | fixed — real counter |

The two 🟢 fixes are in 93cbf95. The 🟡 reserved-key hardening spans all submit paths in `sendSurveyEvent`, so it's better as its own change — happy to open one if wanted.

darkopia:
Addressed the QA follow-ups in 4d5e8bd:
- 🟡 **Reserved-key overwrite** — `sendSurveyEvent` now spreads caller `properties` *before* the reserved fields, so URL params can't clobber `$survey_*`/response keys on any submit path.
- 🟢 **Merge test** — added an end-to-end test (render after prefill → submit open question → assert the merged event keeps the prefilled answer). Verified it fails without the fix.

All 228 survey tests pass; lint clean.

darkopia:
Update: pulled the scope back to the ticket in ce9a00b. Reverted the `sendSurveyEvent` property-merge reorder (it touched the shared manual-submit path) and the unrelated loop-guard tweak, and dropped a redundant test. The reserved-key hardening is a real but pre-existing issue — better as its own small PR than bundled here. Net change is now just: thread `properties` into the prefill auto-submit, and persist auto-advanced indices as `visitedIndices`, with two regression tests.

packages/browser/src/__tests__/extensions/surveys.test.ts
```diff
@@ -889,6 +889,74 @@ describe('SurveyManager', () => {
                 expect(surveyDiv.getElementsByClassName('thank-you-message').length).toBe(0)
             }
         })
+
+        it('retains the auto-advanced prefilled answer through a later manual submit', async () => {
```
turnipdabeets, line 893: [suggestion] This test is the strongest guard in the PR — I verified it fails on main without the fix. #4313 (closed as a duplicate) also had a real-browser Playwright spec for this exact flow (`playwright/mocked/surveys/url-prefill.spec.ts`, auto-submit event + follow-up submit against the built bundle). Could we port it over as a follow-up commit so the flow is covered end-to-end too?

packages/browser/src/utils/survey-url-prefill.ts
```diff
@@ -152,9 +152,11 @@ export function calculatePrefillStartIndex(
...
     responses: Record<string, any>
-): { startQuestionIndex: number; skippedResponses: Record<string, any> } {
+): { startQuestionIndex: number; skippedResponses: Record<string, any>; skippedIndices: number[] } {
     let currentIndex = 0
     const skippedResponses: Record<string, any> = {}
+    // Auto-advanced questions, persisted as visitedIndices so a manual submit keeps their answers.
+    const skippedIndices: number[] = []
 
     const MAX_ITERATIONS = survey.questions.length + 1
     const iterations = 0
```
turnipdabeets: [nit] Pre-existing, not this PR: `iterations` is `const` and never incremented, so the `MAX_ITERATIONS` guard is a no-op and a cyclic branching config would loop forever. Fine to leave here — worth grabbing in the reserved-key follow-up PR.

packages/browser/src/__tests__/extensions/surveys.test.ts
darkopia: Ported it in 8cdc5ec — `playwright/mocked/surveys/url-prefill.spec.ts`, adapted to this branch's helpers. It renders after URL prefill, asserts the auto-submit event carries the prefilled answer + `account_number`, then submits the follow-up and asserts the merged event (same submission id) keeps the prefilled answer. The Playwright job will exercise it against the built bundle. (Left the `iterations` loop-guard nit for the follow-up per your note.)

marandaneto:
cc @PostHog/team-surveys pls review

lucasheriques (approved):
Approving, both fixes look right.

Heads up that `visitedIndices` is also the back-button stack - `canGoBack` is `visitedIndices.length > 0` (`surveys.tsx:1472`, rendered at `BottomSection.tsx:31`). So seeding it means a prefilled hosted survey with the back button enabled now gets one on the follow-up question. Probably fine (back = the thumbs they already answered, re-answering repopulates the index, no dupe) but there's a test pinning the old behavior at `back-button.test.tsx:173`. If you ever revisit this, `Object.keys(skippedResponses)` prunes the same set without touching the back stack.

Agree on reverting the reserved-key reorder, that hit the shared manual-submit path. The hosted page forwards every non-question param verbatim regardless, so it's already live there rather than something this adds. I'll do it separately in the monorepo, filtering `$` keys where the params get collected.

nit: new test at `surveys.test.ts:893` lives under "URL prefill that completes the survey" but doesn't complete.

packages/browser/src/utils/survey-url-prefill.ts
marandaneto: Good catch — fixed in 2ff17ba4d by incrementing the iteration counter, with a regression test covering cyclic branching.

marandaneto (approved):

marandaneto:
Addressed the test-scope nit from @lucasheriques in 2d130dfbe: renamed the enclosing block to `renderSurvey with URL prefill`, which covers both completion and continuation cases.

PR #4316, opened 2026-07-29
feat(browser-common): expand extension client context and persistence

dustinbyrne:
## Problem

Browser-common extensions need a host-neutral way to access the request semantics and identity context required by feature evaluation. Stateful extensions also need coherent batched persistence operations without depending on browser-v1 internals.

## Changes

- Add compression and sent-at request options plus device, library, and initial-person context to the shared `Client` contract.
- Implement the new context and transport forwarding in the browser-v1 client adapter.
- Allow key-value stores to initialize asynchronously, then expose synchronous reads and buffered writes while the host owns ordered durable flushing.
- Add batched KV reads, object writes, and multi-key removals to `KeyValueStore` and its browser/test implementations.
- Extend persistence-key policy checks to understand object-form KV writes.
- Document initialization and synchronous-buffer semantics for extension authors.

This is the prerequisite PR for the stacked feature-flags lifecycle migration.

### Validation

- `pnpm --dir packages/browser-common lint` — passed.
- `pnpm --dir packages/browser-common test:unit` — 14 suites passed, 76 tests passed.
- `pnpm --dir packages/browser-common build` — passed.
- Browser client/persistence focused tests — 2 suites passed, 436 tests passed.
- `cd packages/browser && pnpm typecheck` — passed.
- `pnpm bundle-size:array 3ad551fe0` — +0.51 KiB minified (+0.19%), +0.16 KiB gzip (+0.20%), +0.12 KiB Brotli (+0.17%).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react-native-plugin
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/openfeature-node-provider
- [ ] @posthog/openfeature-web-provider
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for the impact of any changes across different platforms
- [x] Accounted for backwards compatibility of any changes (no breaking changes!)
- [x] Took care not to unnecessarily increase the bundle size

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file

[...]

PR #4318, opened 2026-07-29
refactor(browser): migrate feature flags to the shared extension lifecycle

dustinbyrne:
## Problem

Browser feature flags are still owned through browser-v1-specific lifecycle, request, and persistence paths. This prevents them from using the shared extension lifecycle and keeps feature-flag state and event enrichment coupled to `PostHog` persistence internals.

This draft is stacked on #4316, which provides the shared client context, request semantics, and initialized buffered KV contract used here.

## Changes

- Refactor `PostHogFeatureFlags` into a shared-lifecycle extension with browser-v1 config mapping, eager enrollment, and lifecycle-managed setup/disposal.
- Initialize client KV during setup, then use its synchronous memory buffer directly as the sole feature-flag state source.
- Route feature-flag, remote payload, and early-access requests through the extension client.
- Make the latest `/flags` request authoritative through logical Promise cancellation; superseded transports may finish but cannot update state.
- Persist early-access flag and person state coherently in one batch before same-tick callbacks and capture.
- Move feature-flag event enrichment out of persistence and into the feature-flags extension while retaining the public browser feature-flag facade and persistence compatibility.
- Update feature-flag, core lifecycle, persistence-policy, and event-enrichment tests for the migrated ownership and asynchronous request path.

### Validation

- Affected 10-suite browser Jest run — 10 suites passed, 813 tests passed, 1 skipped.
- `cd packages/browser && pnpm typecheck` — passed.
- `cd packages/browser && pnpm write-mangled-property-names` — passed; generated names updated.
- Production build and cross-bundle mangled-property consistency check — passed.
- Focused Chromium slim-bundle feature-flag tests — 6 tests passed across external and no-external slim bundles.
- `git diff --check` and Prettier checks — passed.
- `pnpm bundle-size:array 79aef168d` — +2.75 KiB minified (+1.04%), +0.87 KiB gzip (+1.07%), +0.73 KiB Brotli (+1.06%).

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with Pi and delegated read-only review under direct human guidance. The shared client/KV prerequisite remains in #4316; this stacked PR contains only the dependent feature-flags lifecycle migration and its tests.


packages/browser/src/posthog-featureflags.ts
```diff
@@ -570,12 +661,20 @@ export class PostHogFeatureFlags implements Extension {
             return
         }
 
-        // Emit event so consumers know flags are being reloaded
-        this._instance._internalEventEmitter.emit('featureFlagsReloading', true)
+        this._requestInFlight = undefined
```
marandaneto: **Blocking:** Clearing `_requestInFlight` here removes the previous single-flight back pressure. Any reload more than 5 ms later can start another uncancelled `/flags` request while the slow request is still running, so common `identify()`, `group()`, or property updates can create unbounded concurrency. Please queue one follow-up request or abort the superseded transport while retaining latest-response-wins semantics.

marandaneto (commented):
Added inline review findings.

packages/browser/src/posthog-featureflags.ts
```diff
@@ -253,27 +314,78 @@ export class PostHogFeatureFlags implements Extension {
...
+    private _remove(keys: keyof FeatureFlagsState | readonly (keyof FeatureFlagsState)[]): void {
+        this._persist(() => this._client?.kv.remove(keys))
+    }
+
+    private _persist(operation: () => void): void {
+        try {
+            operation()
+        } catch (error) {
+            this._logger.error('Failed to update feature flag persistence', error)
+        }
```
marandaneto, line 378: should we swallow persistence failures here or in the caller?

marandaneto (approved):
left a comment, but lgtm to unblock

packages/browser/src/posthog-featureflags.ts
dustinbyrne: i'm going to leave this as is because the generic persistence is an interface and not guaranteed to handle exceptions (though it should)

Issue #4317, opened 2026-07-29
ai(otel): Vercel AI SDK v7 gaps: docs, peer range, serverless flushing


slshults:
Setting up AI observability for the Vercel AI SDK via OTel currently trips over four distinct problems. All were verified against `@posthog/ai@8.4.3`, current posthog-js main, and the live docs page on 2026-07-29. Two of them affect every AI SDK version, not just v7.

1. **Docs pass `apiKey`, but the published option is `projectToken`, and the mismatch is a silent no-op.** The [Vercel AI installation guide](https://posthog.com/docs/ai-observability/installation/vercel-ai) constructs `new PostHogSpanProcessor({ apiKey: '<ph_project_token>', ... })`. `PostHogSpanProcessorOptions` has required `projectToken` (published since at least 8.2.0), and a missing or blank token disables the processor as a defensive no-op with only a console.warn (#3703). TypeScript users get a compile error; plain JS users copying the docs get zero events and no visible failure.

2. **Docs use the pre-v7 telemetry API.** The guide's `experimental_telemetry` block (including `metadata.posthog_distinct_id`) no longer exists in AI SDK v7, which moved to `registerTelemetry(new OpenTelemetry())` from `@ai-sdk/otel`, per-call `telemetry: {...}`, and `runtimeContext` with `telemetry.includeRuntimeContext`. Ingestion already supports the v7 shapes (the vercel middleware reads both the `ai.telemetry.metadata.` and `ai.settings.context.` prefixes, including `posthog_distinct_id`), so this is purely a docs gap: v7 users have no way to discover the working recipe.

3. **`@ai-sdk/provider` peer range excludes v4.** `@posthog/ai` declares `"@ai-sdk/provider": "^2.0.0 || ^3.0.0"` (optional). AI SDK v7 ships provider v4, so v7 installs get unmet-peer warnings (npm can hard-fail resolution) even when only the provider-independent `/otel` subpath is used. Related: #3674 tracks v7 support for `withTracing`, which only handles `LanguageModelV2`/`V3`.

[...]

PR #4319, opened 2026-07-29
feat(ai): support Vercel AI SDK v7 OpenTelemetry

kdawgwilk:
## Problem

Vercel AI SDK v7 moved tracing to `@ai-sdk/otel` and `registerTelemetry`. `@posthog/ai` did not accept the v7 provider peer, while the legacy `withTracing` wrapper only implements the v5/v6 provider interfaces. The existing guide also used the wrong processor option name and did not provide a reliable request-scoped flush lifecycle.

## Changes

- Allow `@ai-sdk/provider` v4 as an optional peer for the OpenTelemetry integration.
- Keep `PostHogSpanProcessor` batched by default and document its existing public `forceFlush()` method for request-scoped/serverless lifecycles; no new processor mode is added.
- Reject provider-v4 models passed to legacy `withTracing` with a focused error directing users to `@ai-sdk/otel` and `@posthog/ai/otel`.
- Add an AI SDK v7 example using `@ai-sdk/otel`, `registerTelemetry`, `runtimeContext`, `enrichSpan`, explicit request-scoped flushing, and opt-in prompt/output recording.
- Map distinct ID, AI session ID, trace name, native groups, and scalar custom properties while preventing custom properties from overriding reserved attribution fields.
- Exercise the packaged public processor/exporter path against a local OTLP HTTP server in CI.
- Update the rendered installation guide in companion PR PostHog/posthog#75112. That docs PR remains draft until this package change is merged and released.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Validation

- `CI=true pnpm install --frozen-lockfile`
- `pnpm exec turbo build --filter=@posthog/ai` (all bundles and declarations; 5/5 tasks successful)
- `pnpm --filter @posthog/ai exec jest --watchman=false --runInBand` (25 suites passed, 1 skipped; 666 tests passed, 85 skipped)
- `pnpm install --lockfile=false`, `pnpm build`, and `pnpm test` in `examples/example-ai-vercel-ai-v7` using the locally packaged `@posthog/ai` tarball (1 OTLP integration test passed)
- `git diff --check upstream/main`

## 🤖 Agent context

Codex implemented and validated this change using current Vercel AI SDK package behavior and official documentation. The work was source-first: v7 telemetry, emitted span attributes, and PostHog OTLP ingestion expectations were checked before choosing the OTEL-only compatibility path. No proprietary AskElephant code, identifiers, tokens, or URLs are included.


kdawgwilk:
Resolves some issues with https://github.com/PostHog/posthog-js/issues/4317

packages/ai/src/otel/processor.ts
```diff
@@ -29,6 +35,14 @@ export interface PostHogSpanProcessorOptions {
    */
   host?: string
 
+  /**
+   * How completed spans are exported. Defaults to `batch`.
+   *
+   * Use `immediate` in request-scoped or serverless runtimes where the process may be
+   * suspended before a batch timer runs. Immediate export has more per-span overhead.
```
marandaneto: `SimpleSpanProcessor` removes the batch timer, but its `onEnd()` still starts an asynchronous export without giving the request a promise to await. A serverless runtime can therefore suspend while the HTTP export is still in flight. Could we document that callers must `await processor.forceFlush()` before returning (or attach it to a platform lifecycle hook such as `waitUntil`)? Otherwise `immediate` mode alone does not provide the reliability this recommendation implies.

examples/example-ai-vercel-ai-v7/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+The example uses `runtimeContext`, `telemetry.includeRuntimeContext`, and
+`enrichSpan` to add:
+
+- `posthog.distinct_id` for per-user attribution
+- `$session_id` for session correlation
+- custom properties
+- group IDs as `posthog.group.<type>` custom properties
+
+OpenTelemetry JavaScript attributes cannot contain the object-valued `$groups`
```
marandaneto: PostHog ingestion supports native groups from OpenTelemetry: it parses a JSON-string `$groups` attribute into the native groups object. Could this example emit `$groups: JSON.stringify(groups)` in `getPostHogSpanAttributes` and assert that attribute in the test? The object itself cannot be an OTel attribute, but the current text incorrectly concludes that native association is impossible.

examples/example-ai-vercel-ai-v7/integration.test.ts
```diff
@@ -0,0 +1,92 @@
...
+import { NodeSDK } from '@opentelemetry/sdk-node'
+import { InMemorySpanExporter, SimpleSpanProcessor } from '@opentelemetry/sdk-trace-base'
+import { generateText, registerTelemetry } from 'ai'
+import { MockLanguageModelV4 } from 'ai/test'
+
+import { getPostHogSpanAttributes } from './telemetry.js'
+
+test('AI SDK v7 emits enriched OpenTelemetry spans after provider startup', async () => {
+    const exporter = new InMemorySpanExporter()
+    const spanProcessor = new SimpleSpanProcessor(exporter)
```
marandaneto: This constructs OpenTelemetry's `SimpleSpanProcessor` directly, so it does not exercise `PostHogSpanProcessor({ exportMode: 'immediate' })`, including its selection, AI-span filtering, or exporter wiring. Also, `examples/**` is excluded from the root pnpm workspace and I couldn't find a workflow that runs this test, so normal CI does not verify the newly advertised provider-v4 path. Could we add CI coverage that goes through the public PostHog processor?

examples/example-ai-vercel-ai-v7/package.json
```diff
@@ -0,0 +1,26 @@
...
+        "start": "tsx index.ts",
+        "test": "tsx --test integration.test.ts"
+    },
+    "dependencies": {
+        "@ai-sdk/openai": "4.0.24",
+        "@ai-sdk/otel": "1.0.41",
+        "@opentelemetry/resources": "2.7.0",
+        "@opentelemetry/sdk-node": "0.218.0",
+        "@opentelemetry/sdk-trace-base": "2.7.0",
+        "@posthog/ai": "file:../../target/posthog-ai.tgz",
```
marandaneto: `target/` is gitignored and absent from a clean checkout, so the README's first `pnpm install` will fail because this tarball does not exist yet. Please add a setup step to run the repository's package command from the root before installing this example (as the other local-tarball examples do).

examples/example-ai-vercel-ai-v7/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+                attributes[`posthog.group.${groupType}`] = groupId
+            }
+        }
+    }
+
+    const properties = runtimeContext?.properties
+    if (properties && typeof properties === 'object' && !Array.isArray(properties)) {
+        for (const [key, value] of Object.entries(properties)) {
+            if (isAttributeValue(value)) {
+                attributes[key] = value
```
marandaneto: Because arbitrary properties are copied last, a property named `posthog.distinct_id`, `$session_id`, or one of the group attributes silently overwrites the trusted attribution value set above. Could we copy custom properties first, or explicitly skip reserved PostHog keys, so runtime attribution cannot be accidentally replaced?

marandaneto (commented):
A few inline findings from reviewing the AI SDK v7 and immediate-export changes.

packages/ai/src/otel/processor.ts
carlos-marchal-ph: There is a bigger cost here than per-span overhead. `SimpleSpanProcessor` sends one OTLP request per span, and the exporter's bounded queue throws `Concurrency Limit reached` once 30 are in flight (`concurrencyLimit` defaults to 30), which `onEnd` hands to `globalErrorHandler`, a no-op without a configured diag logger. No buffer, no retry, no output.

With a 300ms sink and 50 AI spans I got 50 delivered on batch and 30 on immediate. A v7 `generateText` with tools emits roughly 6 spans per call.

packages/ai/package.json
```diff
@@ -69,7 +69,7 @@
     "@opentelemetry/api": "^1.9.0",
     "@opentelemetry/exporter-trace-otlp-http": ">=0.200.0 <1.0.0",
     "@opentelemetry/sdk-trace-base": "^2.0.0",
-    "@ai-sdk/provider": "^2.0.0 || ^3.0.0",
+    "@ai-sdk/provider": "^2.0.0 || ^3.0.0 || ^4.0.0",
```
carlos-marchal-ph, line 72: This removes the unmet-peer warning that currently stops a plain JS user reaching `withTracing` with a v4 model, and that path has no version guard, so a v4 model falls through both `isV2Model` and `isV3Model`.

`middleware.ts:444` then skips the Anthropic branch, so `adjustAnthropicV3CacheTokens` never runs: on an identical Anthropic-shaped usage payload, changing only `specificationVersion`, I got `$ai_input_tokens: 200` for v3 and `1000` for v4, so cached tokens get billed as uncached input. `middleware.ts:471` also reports `$ai_framework_version: '5'`. Could we add an `isV4Model` guard that throws pointing at the OTel integration?

examples/example-ai-vercel-ai-v7/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    const groups = runtimeContext?.groups
+    if (groups && typeof groups === 'object' && !Array.isArray(groups)) {
+        for (const [groupType, groupId] of Object.entries(groups)) {
+            if (isAttributeValue(groupId)) {
+                attributes[`posthog.group.${groupType}`] = groupId
+            }
+        }
+    }
```
carlos-marchal-ph: `posthog.group.*` is not recognised anywhere in ingestion, only `posthog.distinct_id`, `posthog.geoip_disable`, `posthog.ai.debug` and `$groups` are, so these land as literal event properties named `posthog.group.company`. What ingestion actually parses is a JSON string `$groups`:

```suggestion
    const groups = runtimeContext?.groups
    if (groups && typeof groups === 'object' && !Array.isArray(groups)) {
        attributes.$groups = JSON.stringify(groups)
    }
```

examples/example-ai-vercel-ai-v7/telemetry.ts
```diff
@@ -0,0 +1,47 @@
...
+    runtimeContext: Record<string, unknown> | undefined
+}): Record<string, PostHogAttributeValue> {
+    const attributes: Record<string, PostHogAttributeValue> = {}
+
+    if (typeof runtimeContext?.distinctId === 'string') {
+        attributes['posthog.distinct_id'] = runtimeContext.distinctId
+    }
+
+    if (typeof runtimeContext?.sessionId === 'string') {
+        attributes.$session_id = runtimeContext.sessionId
```
carlos-marchal-ph: Should be `$ai_session_id`. `ai_events.session_id` is `JSONExtract(properties, '$ai_session_id')`, and both the sessions scene and the traces list read that column with no aliasing either way. `$session_id` only drives the replay button on the trace scene, and only when a real recording exists under that id, so `example-session` gives a dangling link and no session correlation. Same on README line 29.

examples/example-ai-vercel-ai-v7/README.md
```diff
@@ -0,0 +1,52 @@
...
+the v7 integration path.
+
+## Setup
+
+```bash
+pnpm install
+POSTHOG_PROJECT_TOKEN=phc_... OPENAI_API_KEY=... OPENAI_MODEL=... pnpm start
+```
+
+The OpenTelemetry `NodeSDK` must start before
```
carlos-marchal-ph: This is not true. `@ai-sdk/otel`'s constructor calls `trace.getTracer('gen_ai')`, which returns a `ProxyTracer` that resolves its delegate lazily per span, so swapping the two lines still produces all spans with enrichment intact. The real constraint, unstated, is that both happen before the first `generateText` call. Line 46's claim that the test proves the ordering does not hold either, it passes in either order.

examples/example-ai-vercel-ai-v7/README.md
```diff
@@ -0,0 +1,52 @@
+# Vercel AI SDK v7 + PostHog
+
+This example uses the AI SDK v7 OpenTelemetry integration. The legacy PostHog
+`withTracing` wrapper targets the AI SDK v5/v6 provider interfaces and is not
+the v7 integration path.
+
+## Setup
+
+```bash
+pnpm install
```
carlos-marchal-ph, line 16: `pnpm install` fails here on a clean checkout, for two separate reasons. `target/` is gitignored so `file:../../target/posthog-ai.tgz` does not exist until root `pnpm package` runs, and once it does, `pnpm-workspace.yaml`'s `minimumReleaseAge: 10080` rejects the committed lockfile with `ERR_PNPM_MINIMUM_RELEASE_AGE_VIOLATION` on 11 entries, since `ai@7.0.41` was published a day before this PR.

Adding `pnpmfile: ../.pnpmfile.cjs` fixes the second, which is what `example-convex` does; it also makes the hardcoded tarball path redundant, since `readPackage` already rewrites `@posthog/*` deps, and stops the lockfile churning on every local build. `example-convex/README.md:11` has the wording for the first.

examples/example-ai-vercel-ai-v7/integration.test.ts
carlos-marchal-ph: Agree with the concern

examples/example-ai-vercel-ai-v7/README.md
carlos-marchal-ph: +1

marandaneto:
moving to draft until comments are addressed

packages/ai/src/otel/processor.ts
kdawgwilk: Addressed. I removed the proposed immediate/Simple mode. `PostHogSpanProcessor` remains batched, its public `forceFlush()` is documented, and the request-scoped example awaits it before the lifecycle ends (with `waitUntil` noted as an alternative).

examples/example-ai-vercel-ai-v7/README.md
kdawgwilk: Addressed. The enrichment helper now emits `$groups` as a JSON string, the README describes its conversion to native PostHog groups, and the OTLP integration test asserts the attribute and group value.

examples/example-ai-vercel-ai-v7/integration.test.ts
kdawgwilk: Addressed. The test now constructs the public `PostHogSpanProcessor`, sends through the real `PostHogTraceExporter` to a local OTLP HTTP server, calls public `forceFlush()`, and asserts the exported payload. The package-tarball CI job now installs, builds, and runs this example.

examples/example-ai-vercel-ai-v7/package.json
kdawgwilk: Addressed. The README now runs root `pnpm package` before entering the example. The example uses the repository pnpmfile convention and an `@posthog/ai: "*"` dependency that is rewritten to the generated tarball.

examples/example-ai-vercel-ai-v7/telemetry.ts
kdawgwilk: Addressed. Custom scalar properties are copied first and reserved attribution keys are skipped. Trusted distinct ID, AI session ID, trace name, and groups are then applied. The OTLP test includes spoofed reserved properties and verifies none reach the payload.

packages/ai/src/otel/processor.ts
kdawgwilk: Agreed, and addressed by removing the proposed `SimpleSpanProcessor`/immediate option entirely. The supported path keeps batching and explicitly awaits public `forceFlush()` at the request lifecycle boundary, avoiding one request per span and the concurrency-limit data-loss mode.

packages/ai/package.json
kdawgwilk: Addressed. `withTracing` now checks `specificationVersion` up front and throws for v4 with guidance to use `@ai-sdk/otel` plus `@posthog/ai/otel`. A regression test covers the focused v7 error, so expanding the peer range no longer silently routes v4 through legacy accounting.

kdawgwilk: Addressed. `withTracing` now checks `specificationVersion` up front and throws for v4 with guidance to use `@ai-sdk/otel` plus `@posthog/ai/otel`. A regression test covers the focused v7 error, so expanding the peer range no longer silently routes v4 through legacy accounting.

examples/example-ai-vercel-ai-v7/telemetry.ts
kdawgwilk: Addressed exactly on the recognized ingestion shape: sanitized group entries are serialized into the `$groups` attribute. The example and real OTLP export test assert that field rather than emitting unsupported `posthog.group.*` properties.

examples/example-ai-vercel-ai-v7/telemetry.ts
kdawgwilk: Addressed. Session context now maps to `$ai_session_id` in the helper, example, README, and OTLP assertion. `$session_id` is no longer used for AI session correlation.

examples/example-ai-vercel-ai-v7/README.md
kdawgwilk: Addressed. The README now states the actual constraint: initialize both the OpenTelemetry SDK and AI SDK integration before the first AI call. It explicitly notes that their relative order does not matter because the tracer is lazy, and the test no longer claims to prove an ordering requirement.

examples/example-ai-vercel-ai-v7/README.md
kdawgwilk: Addressed. Setup now packages from the repository root first, `pnpm-workspace.yaml` uses `pnpmfile: ../.pnpmfile.cjs`, the dependency is `@posthog/ai: "*"`, and the churn-prone example lockfile is removed. A clean install/build/test against the generated tarball passes.

examples/example-ai-vercel-ai-v7/pnpm-workspace.yaml
```diff
@@ -0,0 +1,5 @@
+pnpmfile: ../.pnpmfile.cjs
+minimumReleaseAge: 10080
+allowBuilds:
+  esbuild: set this to true or false
```
marandaneto: [P1] `allowBuilds` requires boolean values, but these placeholder strings do not approve either build. The new package-tarball job currently fails here with `ERR_PNPM_IGNORED_BUILDS` for esbuild and protobufjs, so the v7 build and test never run. Please set both entries to the intended booleans.

packages/ai/src/otel/processor.ts
```diff
@@ -60,6 +60,10 @@ class NoopSpanProcessor implements SpanProcessor {
  * start with `gen_ai.`, `llm.`, `ai.`, or `traceloop.`) are exported;
  * all other spans are silently dropped.
  *
+ * Request-scoped and serverless runtimes should keep a reference to this
+ * processor and await {@link PostHogSpanProcessor.forceFlush} before the request
```
marandaneto, line 64: [P1] `BatchSpanProcessor.forceFlush()` only drains spans still in `_finishedSpans`; if the batch timer has already moved spans into an active HTTP export, the queue is empty and `forceFlush()` resolves before that request completes. In a long-running serverless request this can still suspend with spans in flight, so this lifecycle guarantee needs active-export tracking (or revised guidance) plus a delayed-export regression test.

examples/example-ai-vercel-ai-v7/index.ts
```diff
@@ -0,0 +1,80 @@
...
+            // them only when explicitly enabled.
+            recordInputs: captureContent,
+            recordOutputs: captureContent,
+        },
+    })
+
+    console.log(result.text)
+} finally {
+    // Request-scoped runtimes must await pending exports before their lifecycle ends.
+    await posthogSpanProcessor.forceFlush()
```
marandaneto, line 78: [P2] This can reject on an exporter error or timeout, which skips the following `sdk.shutdown()` and may mask the original generation error. Please use nested cleanup handling so shutdown always runs while preserving the relevant failure.

packages/ai/src/vercel/middleware.ts
```diff
@@ -42,6 +42,13 @@ function isV2Model(model: LanguageModel): model is LanguageModelV2 {
   return model.specificationVersion === 'v2'
 }
 
+function getSpecificationVersion(model: unknown): unknown {
```
marandaneto, line 45: [P3] This one-use helper duplicates the `isObject` guard already imported by this module. Please inline the lookup using that existing guard at the wrapper boundary rather than maintaining two subtly different object checks.

marandaneto (commented):
Inline findings from the AI SDK v7 review.

marandaneto:
@PostHog/team-ai-observability for a 2nd pass?

carlos-marchal-ph (approved):
Looks good!

PR #4328, opened 2026-07-30
fix(browser): identify anonymous users with matching IDs

marandaneto:
## Problem

When `identify()` is called with an ID that already matches the persisted `distinct_id` while the SDK still considers the user anonymous, the call enables person processing but emits no event and leaves `$user_state` anonymous. The person profile and stored attribution are therefore not updated until a later event happens to be captured.

## Changes

- Treat a matching-ID anonymous `identify()` call as an identity-state transition.
- Mark the user identified and emit one person-processed `$set` event, including any supplied properties and the SDK's automatic initial/session attribution.
- Avoid emitting `$identify` when the anonymous and identified IDs are equal.
- Reload feature flags for a same-ID transition only when the caller supplies properties that can affect flag evaluation. Preserve reported flag calls because the distinct ID did not change.
- Bypass a previously cached person-property hash for the required transition event, then cache the emitted values to preserve subsequent deduplication.
- Add regression coverage for a direct first touch followed by an anonymous `gclid` landing and identification, confirming current attribution reaches `$identify` without changing the original first-touch values.

Validation:

- `pnpm --dir packages/browser test:unit` — 123 suites passed, 5,079 tests passed.
- `pnpm --dir packages/browser lint`
- `pnpm --dir packages/browser exec tsc -b --pretty false`

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent and independently reviewed with Pi's reviewer subagent. We chose a person-processed `$set` instead of a same-ID `$identify`, because there is no anonymous ID to merge in this state. We also avoided reloading flags for a state-only transition because `$user_state` is not part of the `/flags` request. Campaign null semantics and SPA campaign refresh were intentionally kept out of scope.

### DIFF
--- a/.changeset/tidy-spiders-identify.md
+++ b/.changeset/tidy-spiders-identify.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Fix `identify()` creating a person profile when the supplied ID already matches an anonymously persisted distinct ID.
--- a/packages/browser/src/__tests__/personProcessing.test.ts
+++ b/packages/browser/src/__tests__/personProcessing.test.ts
@@ -265,6 +265,46 @@ describe('person processing', () => {
             })
         })
 
+        it('should carry a later gclid landing from an anonymous pageview onto identify without changing direct first-touch attribution', async () => {
+            // arrange: persist a direct first touch, then simulate a later browser session and SDK instance
+            const persistenceName = uuidv7()
+            const directUrl = 'https://example.com/landing'
+            mockReferrerGetter.mockReturnValue('')
+            mockURLGetter.mockReturnValue(directUrl)
+            const { posthog: firstVisitPosthog } = await setup('identified_only', undefined, persistenceName)
+            firstVisitPosthog.capture('direct first visit')
+            firstVisitPosthog.sessionManager!.resetSessionId()
+            firstVisitPosthog.sessionPersistence!.clear()
+            window.sessionStorage.clear()
+
+            const gclid = 'google-click-id'
+            mockURLGetter.mockReturnValue(`https://example.com/landing?gclid=${gclid}`)
+            const { posthog, beforeSendMock } = await setup('identified_only', undefined, persistenceName)
+
+            // act
+            posthog.capture('$pageview')
+            posthog.identify(distinctId)
+
+            // assert
+            const anonymousPageview = beforeSendMock.mock.calls[0][0]
+            expect(anonymousPageview.event).toBe('$pageview')
+            expect(anonymousPageview.properties.gclid).toBe(gclid)
+            expect(anonymousPageview.properties.$process_person_profile).toBe(false)
+            expect(anonymousPageview.$set_once).toBeUndefined()
+
+            const identifyCall = beforeSendMock.mock.calls[1][0]
+            expect(identifyCall.event).toBe('$identify')
+            expect(identifyCall.properties.gclid).toBe(gclid)
+            expect(identifyCall.properties.$process_person_profile).toBe(true)
+            expect(identifyCall.$set_once).toMatchObject({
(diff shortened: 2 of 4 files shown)


marandaneto:
### When does `shouldTransitionToIdentified` happen?

This is the case where the **distinct ID string stays the same, but its meaning changes from anonymous to identified**. It does not affect the normal flow where PostHog creates an anonymous UUID and login supplies a different user ID.

A realistic sequence is:

1. An SSR application already knows a candidate user ID and initializes PostHog with `bootstrap: { distinctID: "user-123" }`, but omits `isIdentifiedID: true` because client-side authentication has not been confirmed yet.
2. The SDK persists `distinct_id = "user-123"` with `$user_state = "anonymous"`.
3. The initial pageview is captured as anonymous and, with `person_profiles: "identified_only"`, does not update a person profile.
4. Client-side authentication completes and calls `posthog.identify("user-123")`.
5. The ID matches the persisted ID, so there are not two identities to merge. Previously this emitted no event and left `$user_state` anonymous.
6. This PR treats it as an identity-state transition: it marks the user identified and emits a person-processed `$set`, allowing stored attribution and supplied properties to reach the profile without sending an invalid same-ID `$identify`.

The same state can also result from an integration registering `distinct_id` before calling `identify()`, or legacy/custom persistence containing a known distinct ID without an identified user state. If bootstrap uses `isIdentifiedID: true`, initialization already marks the user identified and this branch is not used.

marandaneto:
@ioannisj since you added bootstrapping recently to the mobile sdks, its worth checking if they have the same bug, the user is identified but still cached as anon

ioannisj:
> @ioannisj since you added bootstrapping recently to the mobile sdks, its worth checking if they have the same bug, the user is identified but still cached as anon

I checked, and it seems the same bug is on mobile as well. It doesn’t seem related to bootstrapping itself, though, it’s more of an identity() issue, right? Bootstrapping just makes it more apparent?

marandaneto:
> > @ioannisj since you added bootstrapping recently to the mobile sdks, its worth checking if they have the same bug, the user is identified but still cached as anon
> 
> I checked, and it seems the same bug is on mobile as well. It doesn’t seem related to bootstrapping itself, though, it’s more of an identity() issue, right? Bootstrapping just makes it more apparent?

kinda, its the only way to start with an already identified user right so i dont see how else this would happen

ioannisj:
> > > @ioannisj since you added bootstrapping recently to the mobile sdks, its worth checking if they have the same bug, the user is identified but still cached as anon
> > 
> > 
> > I checked, and it seems the same bug is on mobile as well. It doesn’t seem related to bootstrapping itself, though, it’s more of an identity() issue, right? Bootstrapping just makes it more apparent?
> 
> kinda, its the only way to start with an already identified user right so i dont see how else this would happen

Does this also bite when you happen to identify() with the same anon id right? In that case anon id should be upgraded. At least that's my read

marandaneto:
@ioannisj 

> Does this also bite when you happen to identify() with the same anon id right? In that case anon id should be upgraded. At least that's my read

probably but the anon id is randomly generated (and its unique), so then i'd say this is a bug on the client side

marandaneto:
@ioannisj mind reviewing while you are here?

ioannisj (approved):

PR #4329, opened 2026-07-30
feat(ai): expose prompt config on fetched prompts

jurajmajerik:
## Problem

PostHog prompt versions can now store a `config` object next to the prompt content (PostHog/posthog#74264): model parameters, tools, or any agent configuration, versioned with the prompt and returned by the fetch endpoint. The SDK parses API responses into typed result objects, so the field is currently fetched and thrown away. Python counterpart: PostHog/posthog-python#801.

## Changes

`PromptRemoteResult` gains `config: Record<string, unknown> | null`, always present on api/cache/stale_cache results (null when the version has none). The fallback result declares `config: undefined`, matching how it handles `name` and `version`.

```ts
const result = await prompts.get('summarizer', { label: 'production' })
const config = result.config ?? {}
await openai.chat.completions.create({
  model: (config.model as string) ?? 'gpt-4o',
  temperature: (config.temperature as number) ?? 0,
  messages: [{ role: 'system', content: result.prompt }],
})
```

- Servers that omit `config` or send a non-object shape read as `null` (same tolerance as the existing "server ignored the label param" handling)
- The cache keeps its own copy and every result gets a `structuredClone`, so mutating `result.config` before spreading it into an LLM call cannot pollute later cache hits (this exact bug was caught by review on the Python PR; here it is prevented from the start)

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Claude Code authored this under my direction as the JS counterpart of PostHog/posthog-python#801. Four new tests (config through api and cache, stale-cache retention, cache-pollution guard, older-server matrix); the four existing full-equality metadata assertions now include `config: null`, pinning the "config is always present on remote results" contract. The changeset file was written manually in `.changeset/`.

### DIFF
--- a/.changeset/ai-prompt-config.md
+++ b/.changeset/ai-prompt-config.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': minor
+---
+
+`Prompts.get()` results now include `config`, the JSON object of model parameters or agent configuration stored with the prompt version in PostHog prompt management (`null` when the version has none). Config is carried through the client-side cache and the stale-cache fallback, and each result gets its own copy so mutating `result.config` cannot pollute later cache hits. The hardcoded `fallback` string has no config, so use defensive access like `(result.config ?? {}).temperature`.
--- a/packages/ai/src/prompts.ts
+++ b/packages/ai/src/prompts.ts
@@ -27,6 +27,19 @@ function normalizeHost(value?: unknown): string {
   return (normalizedHost || DEFAULT_PROMPTS_HOST).replace(/\/+$/, '')
 }
 
+/** Reads config from an API response, tolerating servers that don't send it. */
+function extractConfig(value: unknown): Record<string, unknown> | null {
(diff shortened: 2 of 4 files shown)


marandaneto:
the cache-pollution test mutates only top-level properties. Since configs commonly contain
 nested tools and response schemas, a nested mutation assertion would protect against accidentally replacing
 structuredClone with a shallow copy. The current implementation correctly handles nested mutation; the built-package
 smoke test confirmed this.

marandaneto (approved):
left a test suggestion but not a blocker, LGTM

jurajmajerik:
Fixed in [cf01aff](https://github.com/PostHog/posthog-js/pull/4329/commits/cf01aff900a1c20e948f66e763a9651bc0df2ce6). The test now also mutates a nested value, so swapping structuredClone for a shallow copy would fail the test

PR #4330, opened 2026-07-30
feat: expose getUnavailableReason on conversations addon

darkopia:
## Problem

The conversations API loads lazily in the browser. When a caller checks `posthog.conversations.isAvailable()` and gets `false`, that single boolean hides very different situations: the lazy bundle was blocked or failed to load, conversations are disabled for the project, remote config hasn't arrived yet, or it's still initializing. Anything falling back to another channel when conversations is unavailable can't tell these apart, so the fallbacks are undiagnosable.

## Changes

- Add `getUnavailableReason(): ConversationsUnavailableReason | null` to `PostHogConversations`. Returns `null` when available, otherwise one of `disabled_by_config`, `consent_opted_out`, `remote_config_pending`, `disabled_in_project`, `missing_token`, `extensions_unavailable`, `load_failed`, `initializing`, `not_loaded`.
- Track the last lazy-load error (`_lastLoadError`) so a failed load is distinguishable from "still initializing"; cleared on successful load and on `reset()`.
- Unit tests for the main reasons.

No behavior change to existing APIs. Purely additive.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (added manually)

## 🤖 Agent context

Directed by the PostHog support team from a Slack thread while debugging why users on the conversations flag still land in Zendesk. The alerting there collapses every cause into "extension unavailable at submit", which prompted exposing the underlying reason from the SDK.

Authored by the PostHog Slack app (Claude). I could not run the browser test suite in this environment (it needs `pnpm build` first), so the added Jest tests are unverified here — please run `jest .../conversations.test.ts` in review. The companion consumer is PostHog/posthog#(support form) which reads this getter with graceful degradation for older SDKs.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0AMXLYHZ8U/p1785402838172029?thread_ts=1785402838.172029&cid=C0AMXLYHZ8U)*


veryayskiy:
- The toolbar gate is missing. `loadIfEnabled` bails for the toolbar's internal instance (`isToolbarInstance`, matched on `config.name`), but the getter has no matching branch, so that case falls through to not_loaded. Low impact, but the two lists should stay in sync or the divergence should be commented.
- Branch order diverges from `loadIfEnabled`, which checks `__PosthogExtensions__` before remote config while the getter checks it after. With both missing you get `remote_config_pending`, not `extensions_unavailable`. Defensible as "most actionable cause first", but it's worth saying so explicitly since the two functions otherwise read as parallel.

+2 nits:
- `ConversationsUnavailableReason` isn't publicly exported. The entry points re-export `posthog-conversations-types`, not `extensions/conversations/posthog-conversations`, so a consumer can infer the union from the return type but can't import type it by name. It belongs in `posthog-conversations-types.ts`.
- `_lastLoadError` holds the message but only its truthiness is read. Either expose the message  or make it a boolean and save the bytes — relevant because this lands in the non-lazy `array.js` (+487 B) and the checklist claims bundle-size care.

marandaneto:
pushed a few fixes

marandaneto (approved):

PR #4332, opened 2026-07-30
fix(core): identify anonymous user when id already matches persisted distinct id

ioannisj:
## Problem

Calling `identify(id)` where `id` already equals the persisted distinct ID left the user cached as anonymous while core still considered them anonymous. No identity transition fired, no person-processed event was sent, and `$is_identified` stayed false.

This is easy to hit after a non-identified bootstrap: bootstrapping seeds the bootstrap ID as the anonymous ID, so a later `identify(sameId)` matches the stored distinct ID and the transition was skipped. The shared `@posthog/core` engine powers `posthog-react-native` and `posthog-node`, so both were affected.

Ports the browser fix from #4328 to the shared core.

## Changes

When the supplied ID matches the persisted distinct ID and the user is still anonymous, core now marks the user identified and captures one person-processed `$set` (no `$identify`, since there is no anonymous ID to merge). The property hash is cached after capture so deduplication cannot suppress the transition, and flags reload only when properties that affect evaluation are supplied. An argument-less `identify()` still does not upgrade an anonymous user, matching browser behavior.

### Libraries affected

- [ ] All of them
- [x] posthog-node
- [x] posthog-react-native

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj

### DIFF
--- a/.changeset/vivid-snakes-teach.md
+++ b/.changeset/vivid-snakes-teach.md
@@ -1,5 +1,8 @@
 ---
 '@posthog/core': patch
+'posthog-node': patch
+'posthog-react-native': patch
+'posthog-js-lite': patch
 ---
 
-Fix `identify()` leaving a user anonymous when the supplied ID already matches the persisted distinct ID (for example after a non-identified bootstrap seeded the same ID). The user is now marked identified and a person-processed `$set` event is captured. Ports the same fix from posthog-js (browser) to the shared core used by React Native and Node.
+Fix `identify()` leaving a user anonymous when the supplied ID already matches the persisted distinct ID (for example after a non-identified bootstrap seeded the same ID). The user is now marked identified and a person-processed `$set` event is captured. Ports the same fix from posthog-js (browser) to the shared core used by React Native, Node, and posthog-js-lite.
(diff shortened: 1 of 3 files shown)


.changeset/vivid-snakes-teach.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/core': patch
```
marandaneto, line 2: we need patches for node and rn then

marandaneto: and web lite

marandaneto (approved):

.changeset/vivid-snakes-teach.md
ioannisj: Was under the impression that changeset autobumps those when patch?

ioannisj: Added

marandaneto: it was in the past, i changed due to the number of releases even if the upper packages werent affected
i posted on our channel asking all of you before changing

PR #4340, opened 2026-07-30
fix(replay): avoid false-positive network body redaction

marandaneto:
## Problem

Session replay scans recorded request and response bodies for credit-card and SSN-shaped values before capture. The unanchored patterns can mistake timestamps and numeric fragments inside UUIDs for sensitive values, replacing otherwise safe GraphQL bodies with a redaction marker.

## Changes

- Require identifier boundaries and structurally valid groups when detecting SSN/ITIN-shaped values in network bodies.
- Detect payment card candidates as delimited numeric runs, then validate issuer shape and Luhn checksum while still finding cards next to numeric metadata.
- Preserve the existing conservative handling of compact, partially formatted SSNs, ITINs, and 13-digit Visa PANs.
- Add unit and replay integration coverage for the reported timestamp and UUID false positives and sensitive-data regression cases.
- Apply the fix directly without a new defaults version or configuration option.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent and reviewed with Pi's isolated `autoreview` workflow. Review findings led to retaining redaction for PANs next to numeric metadata, compact 13-digit Visa PANs, ITINs, and partially formatted SSNs. The final autoreview found no blocking issues. The change was verified with focused browser-common and replay tests, lint, dependency builds, and the fast array bundle-size comparison.

### DIFF
--- a/packages/browser-common/tests/utils/autocapture-utils.spec.ts
+++ b/packages/browser-common/tests/utils/autocapture-utils.spec.ts
@@ -55,12 +55,15 @@ describe('autocapture utils', () => {
             expect(shouldCaptureValue('{"ssn":"123456789"}', false)).toBe(false)
             expect(shouldCaptureValue('{"ssn":"123-456789"}', false)).toBe(false)
             expect(shouldCaptureValue('{"ssn":"12345-6789"}', false)).toBe(false)
+            expect(shouldCaptureValue('{"note":"SSN123-45-6789"}', false)).toBe(false)
+            expect(shouldCaptureValue('{"note":"123-45-6789SSN"}', false)).toBe(false)
             expect(shouldCaptureValue('{"itin":"912-70-1234"}', false)).toBe(false)
         })
 
         it('does not treat timestamps, UUID fragments, or invalid SSN groups as sensitive in network bodies', () => {
             expect(shouldCaptureValue('{"version":"1785400913428"}', false)).toBe(true)
             expect(shouldCaptureValue('{"id":"a2086e30-2564-40d2-b260-074641cd3b89"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"id":"prefix123-45-6789suffix"}', false)).toBe(true)
             expect(shouldCaptureValue('{"ssn":"000-12-3456"}', false)).toBe(true)
             expect(shouldCaptureValue('{"ssn":"666-12-3456"}', false)).toBe(true)
             expect(shouldCaptureValue('{"ssn":"123-00-4567"}', false)).toBe(true)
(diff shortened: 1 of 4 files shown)


marandaneto:
relates to https://us.posthog.com/project/2/support/tickets/65103

packages/browser-common/tests/utils/autocapture-utils.spec.ts
```diff
@@ -32,6 +32,41 @@ describe('autocapture utils', () => {
...
+            expect(shouldCaptureValue('{"itin":"912-70-1234"}', false)).toBe(false)
+        })
+
+        it('does not treat timestamps, UUID fragments, or invalid SSN groups as sensitive in network bodies', () => {
+            expect(shouldCaptureValue('{"version":"1785400913428"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"id":"a2086e30-2564-40d2-b260-074641cd3b89"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"ssn":"000-12-3456"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"ssn":"666-12-3456"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"ssn":"123-00-4567"}', false)).toBe(true)
+            expect(shouldCaptureValue('{"ssn":"123-45-0000"}', false)).toBe(true)
```
turnipdabeets, line 70: [suggestion] The identifier-boundary requirement also flips values where letters directly abut the number, e.g. `"SSN123-45-6789"` — previously redacted, now captured. Looks intentional (same rationale as the `prefix4242...suffix` card case), but it's not pinned by a test yet — could we document it here?
```suggestion
            expect(shouldCaptureValue('{"ssn":"123-45-0000"}', false)).toBe(true)
            expect(shouldCaptureValue('{"note":"SSN123-45-6789"}', false)).toBe(true)
```

turnipdabeets (approved):

packages/browser-common/tests/utils/autocapture-utils.spec.ts
marandaneto: Good catch! Rather than documenting that as intentional, I treated it as a privacy regression. Fixed in c81c400e4: SSNs with a letter prefix or suffix remain redacted, while UUID-like fragments are ignored only when identifier characters surround both sides. Added coverage for both cases.

PR #4345, opened 2026-07-31
fix(ai): treat LangGraph interrupts as control flow

willwearing:
## Problem

LangGraph implements resumable dynamic interrupts with a `GraphInterrupt` control-flow exception. The graph runtime catches it, checkpoints the run, and returns the pending interrupt to the caller.

The LangChain callback receives that exception through `handleChainError` or `handleToolError`. It currently adds `$ai_error` and `$ai_is_error`, which gives a successfully paused trace an error badge and includes it in AI error counts.

## Changes

- Mirror LangGraph's public, cross-package-safe `isGraphBubbleUp` predicate by checking `error.is_bubble_up === true`, covering `GraphInterrupt`, `NodeInterrupt`, `ParentCommand`, `GraphDrained`, and future subclasses.
- Keep capturing the interrupted trace or span, including the normal metadata cleanup, while omitting `$ai_error` and `$ai_is_error`.
- Preserve the existing behavior for every ordinary error.
- Cover chain interrupts, tool interrupts, the legacy name, and a regular failure.
- Add a patch changeset for `@posthog/ai`.

The guard deliberately mirrors LangGraph's public `isGraphBubbleUp` contract instead of importing LangGraph or using `instanceof`. This works when an application has multiple LangGraph package copies, covers new `GraphBubbleUp` subclasses automatically, and avoids changing the SDK's dependency graph or Zod compatibility.

### Local trace evidence

These traces use synthetic data. Both came from real checkpointed `StateGraph` runs using `MemorySaver` and `interrupt()`, sent through the dedicated AI capture endpoint and the local AI ingestion worker.

#### Before: published `@posthog/ai 8.6.0`

The graph invocation resolves with one pending interrupt, but the `approval` span is marked as an error.

before

#### After: local SDK build

The same interrupted span is preserved without an error classification.

after

### Verification

- Full `@posthog/ai` suite: 21 suites passed, 1 skipped; 528 tests passed, 84 skipped.
- Focused callback suite: 19 tests passed.
- `pnpm turbo build --filter=@posthog/ai`
- Real interrupt reproduction with the reported LangGraph `1.4.4` and LangChain core `1.2.0`.
- Real interrupt reproduction with the current LangGraph `1.4.8` and LangChain core `1.2.3`.
- Before and after events inspected in Kafka, ClickHouse, and the current local PostHog trace UI running through Flox.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Codex in the Codex desktop app assisted with upstream source research, implementation, tests, and local validation. The implementation mirrors LangGraph's `isGraphBubbleUp` predicate, keeps LangGraph out of the SDK dependency graph, and was checked with real interrupt runs against both the reported and current package versions.


packages/ai/src/langchain/callbacks.ts
```diff
@@ -13,6 +13,10 @@ import { sanitizeLangChain } from '../sanitization'
 import { stringifyError } from '../serializeError'
 import { warnIfPostHogAiGateway } from '../gatewayWarning'
 
+// Mirror LangGraph's cross-package-safe guard without adding LangGraph as a dependency.
+const isLangGraphInterrupt = (error: Error): boolean =>
+  error.name === 'GraphInterrupt' || error.name === 'NodeInterrupt'
```
carlos-marchal-ph: LangGraph's own predicate for "this is control flow, not a failure" is the broader `isGraphBubbleUp`, which is just `e.is_bubble_up === true`. `isGraphInterrupt` is a strict subset of it, and `is_bubble_up` is what the runtime itself checks to decide whether a task error is real.

I reproduced the same false error one class over. A subgraph node returning `new Command({ graph: Command.PARENT, goto: ... })`, the documented handoff pattern, resolves the graph perfectly happily and still flags three spans, with a `ParentCommand` payload that is an empty message and a stack. `GraphDrained` is the same story: cooperative drain on SIGTERM, checkpointed and resumable, so a busy deploy would spray fake errors across traces that later resume fine.

Would keying off `is_bubble_up` work instead? It's a prototype getter so it reads as `undefined` on an ordinary `Error`, it's exactly as safe as the name check across duplicated package copies, and it covers subclasses we haven't met yet.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -424,8 +428,10 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       eventProperties['$process_person_profile'] = false
     }
     if (outputs instanceof Error) {
-      eventProperties['$ai_error'] = stringifyError(outputs)
-      eventProperties['$ai_is_error'] = true
+      if (!isLangGraphInterrupt(outputs)) {
```
carlos-marchal-ph: We lose the interrupt payload completely here. Before this change `GraphInterrupt.interrupts` was being serialised into `$ai_error`, since `stringifyError` copies own enumerable keys, so the question actually posed to the human was at least visible somewhere. Now the span carries neither `$ai_error` nor `$ai_output_state`, which makes an interrupted span indistinguishable from a node that returned nothing, and "how many of my runs are waiting on a human?" unanswerable. LangChain doesn't hand `__interrupt__` to the callback either, I checked, so the trace doesn't recover it elsewhere.

Wouldn't it work to set `$ai_output_state` to `{ __interrupt__: outputs.interrupts }`, matching the key LangGraph hands back to the caller? That also routes it through `withPrivacyMode`, which the `$ai_error` path skips entirely, so it ends up better than the old behaviour rather than just restoring it.

packages/ai/tests/callbacks.test.ts
```diff
@@ -755,3 +755,94 @@ describe('LangChainCallbackHandler trace/span state sanitization', () => {
...
+    const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
+    expect(captureCall[0]).toMatchObject({
+      event: '$ai_trace',
+      properties: {
+        $ai_span_id: runId,
+      },
+    })
+    expect(captureCall[0].properties).not.toHaveProperty('$ai_error')
+    expect(captureCall[0].properties).not.toHaveProperty('$ai_is_error')
+    expect(captureCall[0].properties).not.toHaveProperty('$ai_output_state')
```
carlos-marchal-ph: Worth one positive assertion that the suppressed span still keeps `$ai_input_state`, `$ai_latency` and `$ai_span_name`. Every assertion here is a negative, so a regression that dropped those on the interrupt path would sail through.

packages/ai/tests/callbacks.test.ts
```diff
@@ -755,3 +755,94 @@ describe('LangChainCallbackHandler trace/span state sanitization', () => {
...
+  })
+
+  it('captures a legacy NodeInterrupt without error properties', () => {
+    const handler = new LangChainCallbackHandler({ client: mockPostHogClient })
+    const runId = 'node-interrupt-chain'
+
+    handler.handleChainStart(serializedChain, { messages: [] }, runId)
+    handler.handleChainError(makeLangGraphInterrupt('NodeInterrupt'), runId)
+
+    const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
```
carlos-marchal-ph, line 867: This one drops the `toHaveBeenCalledTimes(1)` assertion the other two have.

carlos-marchal-ph (approved):
Approving to unblock, but the findings below should be addressed!

willwearing:
> [!NOTE]
>
> Multi-perspective review: qa-team (2 generalists + 5 SDK specialists), paul-reviewer, xp-reviewer

## Verdict: ✅ APPROVE

All 9 perspectives found the implementation correct and appropriately scoped. The guard matches LangGraph's structural predicate, interrupt payloads use the existing privacy and sanitization path, and ordinary error telemetry remains intact.

### Key findings

- No actionable findings.

### Convergence

- Reviewers independently agreed the structural `is_bubble_up` guard is cross-package safe.
- Reviewers independently confirmed the callback cleanup, ordinary-error path, and focused tests are sound.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🔍 qa-team | 7 independent perspectives completed; no actionable findings survived synthesis. |
| 👤 paul | Clean implementation with no correctness or rollout concern. |
| 📐 xp | Cohesive change with justified test coverage and no worthwhile pre-merge refactor. |

PR #4346, opened 2026-07-31
fix(node): enforce Capture V1 response body timeouts

marandaneto:
## Problem

Capture V1 requests currently stop enforcing `requestTimeout` once fetch resolves with response headers. If a server, intermediary, or injected fetch implementation stalls while streaming the response body, the request can remain unresolved beyond the configured timeout and never enter the existing retry path. During shutdown, the separate `shutdownTimeout` guard eventually returns, but the unresolved send is abandoned and the batch can be lost.

This PR is deliberately scoped to the Capture V1 sender. The analogous timeout lifecycle in `@posthog/core` affects broader node and web request paths and should be addressed separately.

## Changes

- Keep the Capture V1 attempt deadline active through response body parsing.
- Treat timed-out 2xx response bodies as retryable transport failures within the existing retry budget.
- Preserve terminal HTTP status classification when an error response body stalls.
- Bound retryable-response body cancellation by the same attempt deadline before retrying.
- Retain explicit deadline races because `hooks.fetch` is user-injectable and custom implementations may ignore abort signals.
- Distinguish timeouts waiting for response headers from timeouts reading response bodies.
- Add focused coverage for stalled success bodies, exhausted retries, stalled error bodies, abort-aware custom fetches, late custom-fetch responses, and non-settling cancellation.

Validation:
- `CI=1 pnpm --filter posthog-node exec jest src/__tests__/capture-v1/sender.spec.ts --runInBand`
- `CI=1 pnpm --filter posthog-node lint`
- `CI=1 pnpm turbo --filter=posthog-node build`

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human selected this reviewed high-priority fix and directed its validation and submission. Pi coding agent was used for implementation review, focused validation, and PR preparation; this internal agent session has no public link.

The change keeps the existing timeout configuration and retry/status semantics, extending only the Capture V1 deadline lifetime through response consumption. Response cancellation is attempted before retry and bounded by that same deadline. Human review is required before merge.


packages/node/src/capture-v1/sender.ts
```diff
@@ -111,65 +118,71 @@ export class V1CaptureSender {
...
+            await this.sleep(this.backoffDelay(attempt, retryAfterMs))
+            continue
+          }
+          const httpError = await this.buildHttpError(response, status, captureAttempt.waitFor)
+          return this.surfaceBatchFailure(requestId, drops, pending, httpError)
+        }
+
+        let parsed: V1BatchResponse
+        try {
+          parsed = await captureAttempt.waitFor(this.parseResponse(response))
```
turnipdabeets: [question] A timed-out 2xx body is now classified as a retryable transport failure, so we re-send a batch the server already accepted — and since the deadline now also covers the body read, a merely slow (not stalled) success can convert into a re-send. uuids and `PostHog-Request-Id` stay stable across attempts, which is the right shape for dedup — do we know the v1 endpoint actually dedupes on event uuid? Worth confirming with whoever owns it before v1 becomes the default.

turnipdabeets (approved):

packages/node/src/capture-v1/sender.ts
marandaneto: Yes — ingestion deduplicates events by their event UUID. Capture V1 preserves each event UUID (and the batch request ID) across attempts specifically so an ambiguous timeout after server acceptance remains safe under at-least-once delivery. A slow or stalled 2xx body can therefore be retried without creating a second logical event. Thanks for calling this out; this is an important invariant to keep before V1 becomes the default.

eli-r-ph: 👋 Hey folks,

There's some nuance here. I definitely see the value of stall protection, but there's some additional context to consider:

* The capture (legacy and v1) backend already enforces stall timeouts on incoming requests
* Capture does _not_ dedup resent events - it will:
    * Fail an incoming batch with duplicate event UUIDs in it
    * Expect the caller (SDK) to prune events that weren't marked for retry in the backend response, prior to resending them
    * Trusts what it receives otherwise

ClickHouse storage _does_ dedup on event UUID in an eventually consistent way, but that's after capture. Enforcing strict UUID presence and compliance on incoming events across the SDKs (and clearly documenting we don't make event dedup guarantees in the product without it) is part of the v1 contract for every SDK that supports it, although we've made strides enforcing it in legacy capture too along the way.

eli-r-ph: One more thing - capture stall protection works fine outside the scope of a pod crash or unexpected hard fail, but during a deploy cutover or graceful shutdown (we try not to panic in the backend!) we also attempt to flush and resolve in-flight requests for a generous window of time. 

This isn't foolproof of course, but is relevant to how aggressive these protections and deadlines might need to be on the SDK side. Having some SDK-side stall protection still makes sense to me

packages/node/src/capture-v1/sender.ts
```diff
@@ -111,65 +118,71 @@ export class V1CaptureSender {
...
+      let captureAttempt: V1CaptureAttempt | undefined
       try {
-        response = await this.sendOnce(url, payload, attempt, requestId)
+        captureAttempt = await this.sendOnce(url, payload, attempt, requestId)
+        const { response, signal } = captureAttempt
+        const { status } = response
+        if (status < 200 || status >= 300) {
+          if (!isLastAttempt && RETRYABLE_STATUSES.has(status)) {
+            const retryAfterMs = this.parseRetryAfter(response)
+            this.cancelBody(response)
```
eli-r-ph: **1. Fire-and-forget `cancelBody` on the 5xx path trades a hang for connection-pool pressure.** Under undici an un-cancelled body keeps the connection checked out; sustained 5xx with keepalive can starve the agent pool. The hang it avoids is real, but this over-corrects — bound it instead of abandoning it, since the deadline is still armed there: `await captureAttempt.waitFor(this.cancelBody(response)).catch(() => {})`. Also, the comment asserts "some runtimes do not settle body cancellation" without naming one.

packages/node/src/capture-v1/sender.ts
```diff
@@ -182,11 +195,41 @@ export class V1CaptureSender {
     }
 
     const controller = new AbortController()
-    const timer = safeSetTimeout(() => controller.abort(), this.config.requestTimeoutMs)
+    const timeoutError = new Error(`Capture V1 request timed out after ${this.config.requestTimeoutMs}ms`)
+    timeoutError.name = 'AbortError'
+    let timer: ReturnType<typeof safeSetTimeout>
+    const deadline = new Promise<never>((_resolve, reject) => {
```
eli-r-ph, line 206: **2. Justify the `Promise.race` deadline or drop it.** Under spec-compliant fetch, aborting the signal errors the body stream, so the minimal fix is just keeping the timer armed through body consumption — no `deadline`/`waitFor` machinery. The machinery *is* justified, but only because `hooks.fetch` is user-injectable and a custom fetch may ignore `signal` for body reads. That reason appears nowhere.

packages/node/src/capture-v1/sender.ts
```diff
@@ -111,65 +118,71 @@ export class V1CaptureSender {
...
+        }
+        // Connection/timeouts, including a timeout while consuming the response body,
+        // are retried until the budget runs out.
         if (isLastAttempt) {
           return this.surfaceBatchFailure(requestId, drops, pending, transportError)
         }
         await this.sleep(this.backoffDelay(attempt))
-        continue
+      } finally {
+        await captureAttempt?.cleanup()
```
eli-r-ph: Smaller: `cleanup()` is called in five places (the backoff `sleep` has to run after the timer clears, so `finally` can't cover it) — restructure to yield a retry decision and sleep outside the try/finally; `await` in `finally` would replace a pending return if it ever rejected; and the timeout error can't distinguish "no headers" from "stalled body", which is exactly the distinction this PR exists to handle.

eli-r-ph (commented):
## The question I'd raise first

`@posthog/core`'s `_fetchWithRetry` has the **identical** defect — `posthog-core-stateless.ts:1632-1644` arms the timer then clears it in a `finally` right after `this.fetch(...)`, before any caller reads `res.text()`. That serves v0 capture and every other API call, in node **and** posthog-js web. Deliberate narrow scoping, or oversight? (Also: `shutdown()` already races `shutdownTimeoutMs` at `:1709`, so the hang never blocked shutdown indefinitely — it leaked the promise and silently lost the batch. The PR description overstates that one detail.)

packages/node/src/capture-v1/sender.ts
marandaneto: Good catch! Fixed in dcefffabc — retryable HTTP responses now await body cancellation through the still-armed attempt deadline. Cancellation is memoized so timeout cleanup does not invoke it twice.

packages/node/src/capture-v1/sender.ts
marandaneto: Addressed in dcefffabc — each attempt now computes its retry delay, runs one synchronous/idempotent cleanup in `finally`, and sleeps afterward. Timeout errors now distinguish waiting for response headers from reading the response body, with tests for both phases.

packages/node/src/capture-v1/sender.ts
marandaneto: Addressed in dcefffabc. I kept the explicit deadline race and documented why: `hooks.fetch` is user-injectable, so a custom implementation may ignore abort while resolving headers or consuming the body. The attempt also observes and cancels a custom-fetch response that resolves after the header deadline.

marandaneto:
Thanks for the broader review. The narrow scope here is deliberate: this PR fixes the new Capture V1 sender, while `@posthog/core`’s analogous timeout lifecycle affects broader node and web paths and should be handled separately. I also corrected the PR description: shutdown has its own `shutdownTimeout` guard, so the stalled send does not block shutdown indefinitely; instead, the guard returns while the unresolved send is abandoned and the batch can be lost. The updated implementation and description are in dcefffabc.

marandaneto:
@eli-r-ph are we good to go?

PR #4347, opened 2026-07-31
fix(core): preserve events added during queue flush

marandaneto:
## Problem

High-throughput SDK users can capture new events while a full queue is being flushed. If the queue reaches `maxQueueSize` and replaces the in-flight entries before the request completes, the flush previously removed entries by position and could discard those newly captured replacement events.

## Changes

- Remove flushed queue items by their snapshotted object identity or stable event UUID instead of slicing the live queue by batch length.
- Preserve compatibility with legacy queue items without UUIDs through the object-identity fallback.
- Add a regression test that fills the queue during an in-flight flush and verifies the replacement events remain queued and are sent by the next flush.
- Add patch changesets for `@posthog/core`, `posthog-node`, `posthog-react-native`, and `posthog-js-lite` so every SDK built on the shared stateless queue receives the fix.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

Additional library affected: `@posthog/core` (patch).

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human identified, prioritized, and reviewed the fix. A Pi coding agent reviewed the prepared diff, added release metadata, ran focused tests/lint/build validation, and opened this draft PR. The implementation uses UUID matching for persisted queue entries while retaining reference-identity matching for legacy entries without UUIDs.

### DIFF
--- a/.changeset/queue-flush-preserve-new-events.md
+++ b/.changeset/queue-flush-preserve-new-events.md
@@ -1,6 +1,8 @@
 ---
 '@posthog/core': patch
 'posthog-node': patch
+'posthog-react-native': patch
+'posthog-js-lite': patch
 ---
 
 Preserve events added to a full queue while an earlier batch is being flushed.
--- a/.changeset/queue-flush-preserve-new-events.md
+++ b/.changeset/queue-flush-preserve-new-events.md
@@ -1,0 +1,5 @@
+---
+'@posthog/core': patch
+---
+
+Preserve events added to a full queue while an earlier batch is being flushed.
--- a/.changeset/queue-flush-preserve-new-events.md
+++ b/.changeset/queue-flush-preserve-new-events.md
@@ -1,5 +1,6 @@
 ---
 '@posthog/core': patch
+'posthog-node': patch
 ---
 
 Preserve events added to a full queue while an earlier batch is being flushed.
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -1454,7 +1454,25 @@ export abstract class PostHogCoreStateless {
 
       const persistQueueChange = async (): Promise<void> => {
         const refreshedQueue = this.getPersistedProperty<PostHogQueueItem[]>(queueKey) || []
-        const newQueue = refreshedQueue.slice(batchItems.length)
+        // The live queue may have overflowed while this batch was in flight. Remove only
+        // snapshotted items: UUID survives persistence, while reference identity supports
+        // legacy queue items without a UUID.
+        const remainingBatchItems = [...batchItems]
+        const newQueue = refreshedQueue.filter((item) => {
+          const itemUuid = item.message?.uuid
+          const batchItemIndex = remainingBatchItems.findIndex(
+            (batchItem) =>
+              batchItem === item ||
+              (typeof itemUuid === 'string' && itemUuid.length > 0 && batchItem.message?.uuid === itemUuid)
+          )
+
+          if (batchItemIndex === -1) {
+            return true
+          }
+
+          remainingBatchItems.splice(batchItemIndex, 1)
+          return false
+        })
         this.setPersistedProperty<PostHogQueueItem[]>(queueKey, newQueue)
         queue = newQueue
         this._dequeuedMessagesCount += batchItems.length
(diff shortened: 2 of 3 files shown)


marandaneto:
To make the race and the before/after behavior concrete, consider this reduced example:

```ts
{
  flushAt: 3,
  maxQueueSize: 3,
  flushInterval: 0,
}
```

The queue starts full:

```text
Queue:     [initial-1, initial-2, initial-3]
In flight: [initial-1, initial-2, initial-3]
```

While that HTTP request is pending, three more events are captured. Because the queue is full, each capture evicts its oldest current entry:

```text
capture replacement-1 -> [initial-2, initial-3, replacement-1]
capture replacement-2 -> [initial-3, replacement-1, replacement-2]
capture replacement-3 -> [replacement-1, replacement-2, replacement-3]
```

### Before this change

When the request succeeds, the flush removes `batchItems.length` entries by position:

```text
[replacement-1, replacement-2, replacement-3].slice(3) -> []
```

The `initial-*` events were successfully sent, but all three `replacement-*` events are removed without ever being sent.

### After this change

The flush removes only entries matching the snapshotted batch by UUID (or object identity for legacy entries). Since none of the `initial-*` entries remain in the live queue, nothing is removed:

```text
Queue after first flush: [replacement-1, replacement-2, replacement-3]
```

Those replacement events remain available for the next flush.

A partial-overflow case is also fixed. For example, if `[A, B]` is in flight, the full live queue changes from `[A, B, C, D, E]` to `[B, C, D, E, F]`, the old positional removal would remove both `B` and unsent `C`. The new logic removes only sent event `B`, preserving `[C, D, E, F]`.

This does not change the intentional `maxQueueSize` overflow policy: if enough events arrive, genuinely unsent oldest events can still be evicted. It specifically prevents flush cleanup from removing different events than the ones it snapshotted and sent.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1454,7 +1454,25 @@ export abstract class PostHogCoreStateless {
...
+        // The live queue may have overflowed while this batch was in flight. Remove only
+        // snapshotted items: UUID survives persistence, while reference identity supports
+        // legacy queue items without a UUID.
+        const remainingBatchItems = [...batchItems]
+        const newQueue = refreshedQueue.filter((item) => {
+          const itemUuid = item.message?.uuid
+          const batchItemIndex = remainingBatchItems.findIndex(
+            (batchItem) =>
+              batchItem === item ||
+              (typeof itemUuid === 'string' && itemUuid.length > 0 && batchItem.message?.uuid === itemUuid)
```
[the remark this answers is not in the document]
marandaneto: We’re treating event UUIDs as the stable, unique identity of an event. The SDK generates a fresh UUID when one is not supplied; a caller-provided UUID is expected to identify that same logical event and must not be reused for a different queued event. Under that invariant, UUID matching cannot remove a distinct replacement event, while remaining necessary to identify persisted queue entries whose object identity may not survive serialization. We’re therefore keeping the implementation as-is.

.changeset/queue-flush-preserve-new-events.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-node': patch
```
turnipdabeets, line 3: Should we include these as well? Or not needed?
```suggestion
'posthog-node': patch
'posthog-react-native': patch
'posthog-js-lite': patch
```

turnipdabeets (approved):

.changeset/queue-flush-preserve-new-events.md
marandaneto: Good catch — yes, both are direct consumers of the shared stateless queue. Added patch releases for `posthog-react-native` and `posthog-js-lite` in 7d7ad5108, alongside the existing `@posthog/core` and `posthog-node` entries.

PR #4348, opened 2026-07-31
fix(node): schedule feature flag polling after requests settle

marandaneto:
## Problem

Node SDK clients using local feature flag evaluation schedule the next definitions poll before the current request finishes. When a definitions request takes longer than the configured polling interval, the timer fires while the original request is still in flight and is deduplicated, leaving no timer to trigger a later poll. The old scheduling point also calculates the delay before response-driven backoff state is updated.

## Changes

- Schedule each feature flag definitions poll in the request's `finally` block so the delay starts after the request settles and uses the latest backoff state.
- Track poller shutdown so an in-flight request cannot schedule another timer after `stopPoller()`.
- Add focused tests for slow requests, shutdown during an in-flight request, and response-driven backoff scheduling.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human directed and reviewed this high-priority fix; Pi's coding agent prepared the changeset, reran focused validation, and opened this PR. The local agent runner does not expose a shareable session URL.

The implementation keeps the existing polling and backoff model, moving only the timer scheduling point and adding a shutdown guard. Focused feature flag tests and Node package lint pass. The dependency-aware build was also attempted, but its declaration generation resolved `@posthog/core` through another linked review worktree in the shared local dependency setup and failed on unrelated `getInjectedReleaseId` declarations.

### DIFF
--- a/.changeset/feature-flag-poll-after-settlement.md
+++ b/.changeset/feature-flag-poll-after-settlement.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+Schedule local feature flag polls after each request settles.
--- a/packages/node/src/__tests__/feature-flags.spec.ts
+++ b/packages/node/src/__tests__/feature-flags.spec.ts
@@ -6521,6 +6521,82 @@ describe('ETag support for local evaluation polling', () => {
   })
 })
 
+describe('local evaluation poll scheduling', () => {
+  let posthog: PostHog
+
+  jest.useFakeTimers()
+
+  afterEach(async () => {
+    jest.useRealTimers()
+    await posthog.shutdown(100).catch(() => undefined)
+    jest.useFakeTimers()
+  })
+
+  it('schedules the next poll after a slow fetch completes', async () => {
+    let resolveFetch!: (response: any) => void
+    const deferredFetch = new Promise<any>((resolve) => {
+      resolveFetch = resolve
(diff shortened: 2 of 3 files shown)


turnipdabeets (approved):
Verified against main: the poller schedules its next tick at request start (`_loadFeatureFlags`) and the dedup wrapper swallows the tick if a definitions request outlasts the interval, so polling dies permanently and flags silently freeze — reachable with default config. Moving the scheduling into `finally` with the `pollerStopped` guard covers every exit path, and both new tests fail on main.

One note: please land #4349 with or before this one — without it a stalled fetch means the `finally` (and thus the reschedule) can be delayed ~300s by undici's default timeouts, since the request abort timer on this path doesn't work until #4349.

gustavohstrassburger (approved):

PR #4349, opened 2026-07-31
fix(node): keep feature flag definition timeout active

marandaneto:
## Problem

Feature flag definition requests configured with a timeout could still hang indefinitely. The timeout was cleared as soon as the fetch promise was returned, rather than when that promise settled, so a stalled definitions endpoint could block explicit reloads and local feature flag evaluation.

## Changes

- Await the feature flag definitions fetch inside its existing `try` block so the abort timer remains active until the request settles.
- Add a regression test with a hanging fetch that verifies both explicit reload and on-demand local evaluation abort and settle after the configured timeout.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human directed this reviewed fix, with implementation and PR preparation assisted by the Pi coding agent. The minimal `async`/`await` change was chosen to preserve the existing request and cleanup flow while keeping the existing abort timer alive until fetch settlement. Validation covered the complete Node feature flag test file, Node lint, and the dependency-aware Node build. Agent session link is unavailable from this environment.

### DIFF
--- a/.changeset/feature-flag-definition-body-timeout.md
+++ b/.changeset/feature-flag-definition-body-timeout.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+Apply the feature flag definition request timeout while consuming the response body.
--- a/.changeset/feature-flag-definition-timeout.md
+++ b/.changeset/feature-flag-definition-timeout.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+Keep the feature flag definition request timeout active until the request settles.
--- a/packages/node/src/__tests__/feature-flags.spec.ts
+++ b/packages/node/src/__tests__/feature-flags.spec.ts
@@ -6175,6 +6175,64 @@ describe('fetch context handling', () => {
   })
 })
 
+describe('feature flag definition request timeout', () => {
+  let posthog: PostHog
+
+  jest.useFakeTimers()
+
+  afterEach(async () => {
+    jest.useRealTimers()
+    await posthog.shutdown()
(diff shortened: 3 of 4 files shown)


turnipdabeets (approved):
Verified against main: the sync `finally` in `_requestFeatureFlagDefinitions` clears the abort timer before the fetch settles, so `requestTimeout` has never applied to this endpoint at all — a stalled definitions fetch pins `loadingPromise`, hanging local evaluation and killing the poller. The `async`/`return await` fix is exactly right and the regression test fails on main.

Follow-up worth filing: `res.json()` on this path is still outside the abort window (same body-phase gap #4346 fixes for the capture sender).

marandaneto:
Good follow-up — I kept this PR focused on the fetch-phase timeout and implemented the response-body phase separately in #4373. The follow-up keeps the timeout active through both `json()` and `text()` consumption, preserves the response body, and covers stalled/slow bodies plus poller recovery. Its focused suite passes all 233 tests and autoreview is clean.

gustavohstrassburger (approved):

PR #4350, opened 2026-07-31
fix(node): preserve unhandled rejection semantics

marandaneto:
## Problem

Exception autocapture currently installs an `unhandledRejection` listener unconditionally. In Node.js, the presence of that listener changes runtime semantics: fatal `throw`/`strict` rejections can stop terminating the process, while other modes can have their warning and exit-code behavior altered.

Node.js users need exception autocapture to observe failures without changing the process behavior configured by Node or the application.

## Changes

- Snapshot the effective unhandled-rejection mode from `process.execArgv` and `NODE_OPTIONS` when the autocapture module loads, preserving command-line precedence and last-option-wins behavior while ignoring later mutations before PostHog construction.
- In fatal `throw` mode, avoid installing an `unhandledRejection` listener and capture the rejection after Node promotes it through `uncaughtException`, while preserving the `unhandledRejection` mechanism tag.
- In `strict` mode, observe promoted rejections through `uncaughtExceptionMonitor` without intercepting Node's fatal exit. Because Node does not wait for asynchronous capture to finish, fatal strict-mode rejections are no longer captured.
- If an application or third-party `unhandledRejection` listener handles the event in `throw` mode, Node does not promote it through `uncaughtException`, so PostHog no longer captures that rejection.
- Keep the existing Sentry-derived `uncaughtException` listener-counting approach rather than adding custom listener-participation tracking.
- In `warn` and `none` modes, capture without changing process completion.
- In `warn-with-error-code`, capture nothing rather than reimplementing Node's warning and exit-code behavior inside the SDK.
- Add child-process coverage that compares behavior with and without the SDK across the supported Node modes.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human directed this high-priority fix and its PR scope. Pi coding agent was used to apply the reviewed changes, run the focused Jest suite plus lint/build, and prepare this draft PR; the local agent session is not publicly linkable.

The implementation intentionally follows Node's configured rejection mode instead of imposing one SDK behavior. Child-process tests compare SDK-enabled behavior against native Node behavior. The scope deliberately matches established SDK approaches instead of adding custom handling for listener-ordering edge cases. This PR requires human review and will not be self-merged or auto-approved.


marandaneto:
## TL;DR

This PR makes Node.js exception autocapture **observational instead of behavior-changing**.

Today, enabling exception autocapture always installs an `unhandledRejection` listener. In Node.js, the mere presence of that listener changes runtime behavior: a rejection that would normally terminate the process in the default/`throw` mode can instead be treated as handled and allow the process to continue. It can also affect warnings and exit codes in other unhandled-rejection modes.

The goal is that running with PostHog produces the same process exit code, warnings, and application-listener behavior as running without PostHog—plus a captured exception event.

To achieve that, this PR:

- avoids installing an `unhandledRejection` listener in fatal `throw`/`strict` modes and captures the rejection after Node promotes it to `uncaughtException`;
- preserves `warn`, `warn-with-error-code`, and `none` behavior;
- respects application `unhandledRejection`/`uncaughtException` listeners, including `once`, listener ordering, and listeners added or removed during dispatch; and
- adds child-process tests comparing native Node behavior with and without the SDK.

In short: **enabling PostHog exception autocapture should never change whether or how the user's Node process terminates.**

marandaneto:
@PostHog/team-error-tracking not sure if this PR makes sense but this caught my attention, we should preserve the ecosystem behaviour rather than changing it, wdyt?

marandaneto:
## Scope and tradeoff update

After comparing this with Sentry Node and OpenTelemetry Node, I narrowed the implementation to avoid maintaining behavior that neither SDK attempts to fully reproduce:

- `throw` / `strict`: no `unhandledRejection` listener; capture after Node promotes the rejection through `uncaughtException`.
- `warn` / `none`: capture through `unhandledRejection`, where installing the listener does not change process termination.
- `warn-with-error-code`: intentionally skip capture because installing a listener suppresses Node's native warning and exit code, and reproducing those semantics inside the SDK adds policy and maintenance complexity.
- Keep the simpler Sentry-derived current-listener count for `uncaughtException`; do not add custom tracking for `once` listeners or listeners added/removed during dispatch.

### Why mode detection remains

Unlike Sentry, the PR derives the effective mode from `process.execArgv` and `NODE_OPTIONS`. Sentry always installs its rejection listener and applies its own mode (`warn` by default), which can override Node's configured `throw` behavior. PostHog instead treats Node's documented configuration as authoritative so it can capture in the safe `warn`/`none` modes without installing a behavior-changing listener in the other modes.

The alternative would be the simpler OpenTelemetry-style approach of never installing `unhandledRejection`, but then PostHog would not capture rejections in `warn` or `none`. Mode detection is kept because it relies on documented Node options and directly enables safe capture; this is a smaller and more stable surface than reconstructing EventEmitter participation or Node's warning/exit implementation.

### Accepted tradeoff

`warn-with-error-code` rejections are not captured, and uncommon `once`/dynamic-listener ordering cases retain the same limitations as the Sentry-style approach. This keeps the PR focused on the primary regression: exception autocapture must not turn default fatal `throw`/`strict` rejections into continuing processes.

Autoreview flagged the known preceding-`once` listener limitation as a blocker. That finding is valid but consciously accepted here because addressing it is the complexity deliberately removed in this scope decision.

.changeset/preserve-node-unhandled-rejection-semantics.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+Preserve Node.js unhandled rejection behavior when exception autocapture is enabled.
```
turnipdabeets: The trade-off is real and test-pinned but not disclosed: in default `throw` mode, a rejection consumed by any application or third-party `unhandledRejection` listener (pm2, Sentry, ...) is no longer captured at all, and under `--unhandled-rejections=warn-with-error-code` nothing is captured. Defensible design, but users upgrading will silently lose `$exception` events in a common setup — I think we need to disclose it. This line also doesn't say what an external dev actually observes (the process now crashes); something like:
```suggestion
Fix `enableExceptionAutocapture` suppressing Node's default crash on unhandled promise rejections; rejections handled by another `unhandledRejection` listener are no longer captured.
```
with the fuller detail (`warn-with-error-code` capturing nothing, which listeners count as handling) spelled out in the PR body's Changes section.

packages/node/src/extensions/error-tracking/autocapture.ts
```diff
@@ -4,7 +4,88 @@
...
+const UNHANDLED_REJECTION_OPTION_NAMES = ['--unhandled-rejections', '--unhandled_rejections']
+const UNHANDLED_REJECTION_MODES = new Set<UnhandledRejectionMode>([
+  'throw',
+  'strict',
+  'warn',
+  'warn-with-error-code',
+  'none',
+])
+
+function splitNodeOptions(nodeOptions: string): string[] {
```
turnipdabeets, line 21: [nit] This diverges from Node's actual NODE_OPTIONS parser (Node only treats `"` as a quote, honors backslash escapes only inside quotes, and splits on spaces only) — a misparse here could re-install the listener in throw mode. Exotic inputs only, but a short comment noting the divergence (or aligning with Node's rules) would help.

packages/node/src/__tests__/extensions/exception-autocapture.child.cjs
```diff
@@ -0,0 +1,35 @@
...
+    const exception = event.properties.$exception_list[0]
+    process.stdout.write(`posthog-capture:${exception.mechanism.type}:${exception.value}\n`)
+    return Promise.resolve()
+  }
+}
+
+setTimeout(async () => {
+  await posthog?.shutdown()
+  process.stdout.write('completed\n')
+}, 50)
```
turnipdabeets: [nit] The 50ms timer races the rejection handling — on a loaded CI box, if shutdown finishes late the child still prints `completed` and the `not.toContain('completed')` assertions flip. Might be worth a longer delay or an explicit signal instead of a fixed timer.

marandaneto: Good catch! Fixed in ca6192f02 by replacing the fixed 50ms timer with beforeExit-based shutdown/completion, so the child waits for Node's event-loop lifecycle rather than wall-clock timing.

.changeset/preserve-node-unhandled-rejection-semantics.md
marandaneto: Good catch! Updated the changeset in ca6192f02 to disclose the observable crash and capture tradeoff, and expanded the PR body's Changes section to explicitly document that application/third-party unhandledRejection listeners prevent capture in throw mode and that warn-with-error-code captures nothing.

packages/node/src/extensions/error-tracking/autocapture.ts
marandaneto: Good catch! Fixed in ca6192f02 by aligning the parser with Node's NODE_OPTIONS rules: double quotes only, escapes only inside quoted strings, and splitting only on literal spaces. I also added coverage for tabs and single quotes.

.changeset/preserve-node-unhandled-rejection-semantics.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+Fix `enableExceptionAutocapture` suppressing Node's default crash on unhandled promise rejections; rejections handled by another `unhandledRejection` listener are no longer captured.
```
ioannisj: [suggestion] Missing `warn-with-error-code` here? `addUnhandledRejectionListener` returns early for that mode too, and there's no `uncaughtException` promotion to fall back on

ioannisj (approved):

.changeset/preserve-node-unhandled-rejection-semantics.md
marandaneto: Good catch! Updated the changeset in f7f9c062a to explicitly disclose that rejections under `warn-with-error-code` are not captured.

hpouillot:
Automated review flagged those:

[P2] strict mode still changes rejection semantics
    packages/node/src/extensions/error-tracking/autocapture.ts:96-105
    PostHog’s uncaughtException listener handles the promoted rejection, allowing Node to subsequently emit unhandledRejection before the asynchronous fatal shutdown exits.
    Without the SDK, strict mode terminates without invoking that listener. This allows application side effects that should never run. Reproduced on Node 20 and 22. Add coverage
    for strict plus an application unhandledRejection listener.

[P2] Live option reads can select the wrong runtime mode
    packages/node/src/extensions/error-tracking/autocapture.ts:69-75
    process.execArgv and process.env.NODE_OPTIONS are mutable after startup. If an application changes NODE_OPTIONS for child processes before constructing PostHog—for example,
    setting it to warn while the current process actually started in default throw mode—the SDK installs a listener and suppresses the crash this PR intends to preserve. The
    effective startup options should be snapshotted rather than read when autocapture starts.


I think we should start using `uncaughtExceptionMonitor` to avoid changing exit behavior (https://nodejs.org/api/process.html#event-uncaughtexceptionmonitor)

marandaneto:
@hpouillot Good catches — addressed in 8c23918a3.

- `strict` mode now uses `uncaughtExceptionMonitor`, so PostHog no longer handles the promoted rejection or allows application `unhandledRejection` side effects before exiting.
- The effective mode is snapshotted when the autocapture module loads, so later `process.execArgv` or `NODE_OPTIONS` mutations before constructing PostHog cannot change listener selection.
- Added child-process regression coverage for strict mode with an application rejection listener and for both option-mutation cases.
- Documented the strict-mode tradeoff: Node does not wait for asynchronous capture to finish, so fatal strict-mode rejections are no longer captured.

Validated with the focused exception-autocapture suite on Node 24, the posthog-node build, and lint.

hpouillot (approved):
🚀

PR #4352, opened 2026-07-31
fix(node): bound stack source context file reads

marandaneto:
## Problem

Node exception enrichment reads source files named by parsed stack frames. The current implementation does not verify the file type or size, so a special file or a large newline-free file can block or consume excessive memory while source context is collected.

Source files outside `process.cwd()` and symlinked workspace files are legitimate in monorepos, serverless layers, and services launched from a different working directory, so the safeguards must preserve that existing behavior.

## Changes

- Normalize relative filenames against the working directory captured for each enrichment pass, keeping successful and failed caches isolated when the working directory changes.
- Continue supporting regular source files anywhere, including absolute paths outside the working directory and symlinked files.
- Open files with nonblocking read flags, validate from the opened descriptor that they are regular files no larger than 10 MiB, and read from that same descriptor so path replacement cannot change what is consumed.
- Bound the stream itself to 10 MiB and close both streams and descriptors on completion, errors, and initialization failures.
- Emit a debug-gated diagnostic when an oversized regular source file is skipped.
- Add coverage for normal, outside-working-directory, symlinked, replaced, oversized, and special files; working-directory changes/removal; cache isolation; and cleanup after early completion and initialization failure.

Validation performed:

- `cd packages/node && ./node_modules/.bin/jest src/__tests__/extensions/error-conversion.spec.ts --runInBand` (17 tests passed)
- `./node_modules/.bin/eslint packages/node/src/extensions/error-tracking/modifiers/context-lines.node.ts packages/node/src/__tests__/extensions/error-conversion.spec.ts` (passed)
- Prettier and `git diff --check` (passed)
- Autoreview against `origin/main` (clean; no actionable findings)
- A direct TypeScript check reported existing errors elsewhere in the package but none in the changed files.
- A direct Rslib build emitted ESM/CJS bundles, then declaration generation hit the worktree's stale `@posthog/core` output (`getInjectedReleaseId` missing); this is unrelated to these changes.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human reviewed the original security-hardening approach and chose a narrower reliability-focused design. Pi coding agent implemented and independently reviewed the revision. The final implementation deliberately avoids treating `process.cwd()` as a security boundary, preserving source context for legitimate outside-root and symlinked files while bounding and validating descriptor-based reads.

### DIFF
--- a/packages/node/src/__tests__/extensions/error-conversion.spec.ts
+++ b/packages/node/src/__tests__/extensions/error-conversion.spec.ts
@@ -2,7 +2,7 @@ import { ErrorTracking as CoreErrorTracking } from '@posthog/core'
 import { constants, type ReadStream } from 'node:fs'
 import { mkdtemp, open, rename, rm, symlink, truncate, writeFile } from 'node:fs/promises'
 import { tmpdir } from 'node:os'
-import { join, relative } from 'node:path'
+import { join, relative, sep } from 'node:path'
 import { createModulerModifier } from '@/extensions/error-tracking/modifiers/module.node'
 import { addSourceContext, MAX_CONTEXTLINES_FILE_SIZE } from '@/extensions/error-tracking/modifiers/context-lines.node'
 import { createRelativePathModifier } from '@/extensions/error-tracking/modifiers/relative-path.node'
@@ -161,6 +161,21 @@ describe('error conversion', () => {
       await addSourceContext([frame])
 
       expect(frame.context_line).toBe('linked context')
+    })
+
+    it('preserves filesystem semantics for paths traversing symlink parents', async () => {
+      const sourceRoot = await makeTemporaryDirectory(tmpdir())
+      const linkedTarget = await makeTemporaryDirectory(sourceRoot)
+      await writeFile(join(sourceRoot, 'source.ts'), 'target context\n')
+      const linkRoot = await makeTemporaryDirectory(tmpdir())
+      await writeFile(join(linkRoot, 'source.ts'), 'lexically normalized context\n')
+      const linkedDirectory = join(linkRoot, 'linked-directory')
+      await symlink(linkedTarget, linkedDirectory, 'dir')
+      const frame = makeFrame(`${linkedDirectory}${sep}..${sep}source.ts`)
+
+      await addSourceContext([frame])
+
+      expect(frame.context_line).toBe('target context')
     })
 
     it('reads from the validated descriptor when the source path is replaced', async () => {
--- a/packages/node/src/__tests__/extensions/error-conversion.spec.ts
+++ b/packages/node/src/__tests__/extensions/error-conversion.spec.ts
@@ -1,6 +1,10 @@
 import { ErrorTracking as CoreErrorTracking } from '@posthog/core'
+import { constants } from 'node:fs'
+import { mkdtemp, open, realpath, rename, rm, symlink, truncate, writeFile } from 'node:fs/promises'
+import { tmpdir } from 'node:os'
+import { join, relative } from 'node:path'
 import { createModulerModifier } from '@/extensions/error-tracking/modifiers/module.node'
-import { addSourceContext } from '@/extensions/error-tracking/modifiers/context-lines.node'
+import { addSourceContext, MAX_CONTEXTLINES_FILE_SIZE } from '@/extensions/error-tracking/modifiers/context-lines.node'
 import { createRelativePathModifier } from '@/extensions/error-tracking/modifiers/relative-path.node'
 
 describe('error conversion', () => {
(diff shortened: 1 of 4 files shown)


packages/node/src/extensions/error-tracking/modifiers/context-lines.node.ts
```diff
@@ -92,25 +111,88 @@ export async function addSourceContext(
...
+ * Opens a bounded regular file without blocking on special files such as FIFOs.
+ * File validation and content reads use the same descriptor so path replacement cannot change what is read.
+ */
+async function openRegularSourceFile(path: string, openSourceFile: OpenSourceFile): Promise<FileHandle | undefined> {
+  let fileHandle: FileHandle | undefined
+  let isValid = false
+  try {
+    fileHandle = await openSourceFile(path, constants.O_RDONLY | constants.O_NONBLOCK)
+    const fileStat = await fileHandle.stat()
+    if (!fileStat.isFile() || fileStat.size > MAX_CONTEXTLINES_FILE_SIZE) {
```
turnipdabeets: [suggestion] Regular source files over 10 MiB now silently lose context and land in the failed-read cache. Could we emit a debug log when the size gate rejects a file, so that's diagnosable in the field?

packages/node/src/__tests__/extensions/error-conversion.spec.ts
```diff
@@ -54,4 +58,205 @@ describe('error conversion', () => {
...
+    it('reads from the validated descriptor when the source path is replaced', async () => {
+      const directory = await makeTemporaryDirectory(tmpdir())
+      const sourceFile = join(directory, 'application.ts')
+      const openedFile = join(directory, 'opened.ts')
+      const replacementFile = join(directory, 'replacement.ts')
+      await writeFile(sourceFile, 'original context\n')
+      await writeFile(replacementFile, 'replacement context\n')
+      const openSourceFile = jest.fn(async (path: string, flags: number) => {
+        expect(path).toBe(sourceFile)
+        expect(flags & constants.O_NONBLOCK).toBe(constants.O_NONBLOCK)
```
turnipdabeets: [nit] On win32 `constants.O_NONBLOCK` is `undefined`, so this becomes `0 !== undefined` and fails. CI is ubuntu-only today, so no rush — but guarding it like the device tests below would keep the suite portable.

packages/node/src/__tests__/extensions/error-conversion.spec.ts
```diff
@@ -54,4 +58,205 @@ describe('error conversion', () => {
...
+      await truncate(sourceFile, MAX_CONTEXTLINES_FILE_SIZE + 1)
+      const frame = makeFrame(sourceFile)
+
+      await addSourceContext([frame])
+
+      expect(frame.context_line).toBeUndefined()
+    })
+
+    if (process.platform !== 'win32') {
+      it.each(['/dev/null', '/dev/zero'])('does not read special file %s', async (specialFile) => {
```
turnipdabeets, line 258: [nit] The FIFO-blocks-forever scenario is the strongest motivation for the non-blocking open, but it's only covered indirectly — an `mkfifo` test inside this non-win32 guard would pin it directly.

turnipdabeets (approved):

packages/node/src/__tests__/extensions/error-conversion.spec.ts
marandaneto: Added in 43c5ec3b0. The Unix-only tests now create a real FIFO with `mkfifo` and verify source-context enrichment returns without blocking.

packages/node/src/__tests__/extensions/error-conversion.spec.ts
marandaneto: Fixed in 43c5ec3b0. The `O_NONBLOCK` flag assertion is now guarded on non-Windows platforms.

packages/node/src/extensions/error-tracking/modifiers/context-lines.node.ts
marandaneto: Good suggestion — fixed in 43c5ec3b0. Oversized regular files now emit a debug-gated diagnostic through the client logger before being skipped.

hpouillot (approved):

PR #4355, opened 2026-07-31
fix(mcp): preserve user-defined tool arguments

marandaneto:
## Problem

MCP server authors can define legitimate `context` or `conversation_id` tool arguments. High-level instrumentation always removed those names before invoking the callback, including when the corresponding analytics feature was disabled or schema injection was skipped because the tool already owned the parameter.

## Changes

- Determine ownership of `context` and `conversation_id` from each tool's current input schema.
- Strip only analytics-owned arguments while preserving user-defined fields and honoring disabled analytics features.
- Add end-to-end coverage for disabled features, schema collisions, analytics-owned fields, and non-object Zod schemas.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

The Pi coding agent implemented the human-directed scoped fix and used GitHub CLI plus the repository's local Jest, ESLint, and Rslib binaries. No public agent-session link is available from this environment.

The implementation uses each tool's current schema as the ownership boundary so analytics removes only parameters it injected. It deliberately preserves existing behavior for analytics-owned arguments and the missing-capability tool while adding regression coverage for user-owned arguments and non-object Zod schemas.

### DIFF
--- a/packages/mcp/src/__tests__/reserved-arguments.test.ts
+++ b/packages/mcp/src/__tests__/reserved-arguments.test.ts
@@ -3,6 +3,7 @@ import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
 import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
 import { CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
 import { z } from 'zod'
+import { DEFAULT_CONTEXT_PARAMETER_DESCRIPTION, DEFAULT_CONVERSATION_ID_DESCRIPTION } from '../extensions/constants'
 import { instrument } from '../index'
 import { fakePostHog } from './test-utils'
 
@@ -65,54 +66,36 @@ describe('high-level reserved analytics arguments', () => {
     }
   })
 
-  it('recomputes ownership after a schema-only tool update', async () => {
-    const server = new McpServer({ name: 'updated-reserved-arguments', version: '1.0.0' })
-    const receivedArgs: Record<string, unknown>[] = []
+  it('strips analytics arguments injected into a non-object Zod schema', async () => {
+    const server = new McpServer({ name: 'record-schema-reserved-arguments', version: '1.0.0' })
+    let receivedArgs: Record<string, string> | undefined
 
-    const registeredTool = server.registerTool(
-      'updated_schema',
-      { inputSchema: z.object({ value: z.string() }).passthrough() },
-      async (args) => {
-        receivedArgs.push({ ...args })
-        return { content: [{ type: 'text', text: 'ok' }] }
-      }
-    )
+    server.registerTool('record_schema', { inputSchema: z.record(z.string()) }, async (args) => {
+      receivedArgs = { ...args }
+      return { content: [{ type: 'text', text: 'ok' }] }
+    })
 
     const { client, cleanup } = await connect(server)
     try {
       instrument(server, fakePostHog(), { context: true, enableConversationId: true })
 
-      const argumentsWithReservedFields = {
-        context: 'supplied context',
-        conversation_id: 'supplied conversation',
-        value: 'kept',
-      }
-      await client.request(
-        { method: 'tools/call', params: { name: 'updated_schema', arguments: argumentsWithReservedFields } },
-        CallToolResultSchema
-      )
-
-      registeredTool.update({
-        paramsSchema: {
-          context: z.string(),
-          conversation_id: z.string(),
-          value: z.string(),
-        },
-      })
+      const listResult = await client.request({ method: 'tools/list' }, ListToolsResultSchema)
+      const tool = listResult.tools.find((candidate) => candidate.name === 'record_schema')
+      expect(tool?.inputSchema.properties?.context).toBeDefined()
+      expect(tool?.inputSchema.properties?.conversation_id).toBeDefined()
 
       await client.request(
-        { method: 'tools/call', params: { name: 'updated_schema', arguments: argumentsWithReservedFields } },
-        CallToolResultSchema
-      )
-
-      expect(receivedArgs).toEqual([
-        { value: 'kept' },
         {
-          context: 'supplied context',
-          conversation_id: 'supplied conversation',
-          value: 'kept',
+          method: 'tools/call',
+          params: {
+            name: 'record_schema',
+            arguments: { context: 'analytics context', conversation_id: 'analytics conversation', value: 'kept' },
+          },
         },
-      ])
+        CallToolResultSchema
+      )
+
+      expect(receivedArgs).toEqual({ value: 'kept' })
     } finally {
       await cleanup()
     }
@@ -126,7 +109,11 @@ describe('high-level reserved analytics arguments', () => {
     server.registerTool(
       'existing_context',
       {
-        inputSchema: z.object({ context: z.string().describe('Application context'), value: z.string() }).passthrough(),
+        inputSchema: {
+          properties: z.string().optional(),
+          context: z.string().describe('Application context'),
+          value: z.string(),
+        },
       },
       async (args) => {
         receivedArgs.set('existing_context', { ...args })
@@ -159,11 +146,26 @@ describe('high-level reserved analytics arguments', () => {
       instrument(server, fakePostHog(), { context: true, enableConversationId: true })
 
       const listResult = await client.request({ method: 'tools/list' }, ListToolsResultSchema)
-      for (const name of ['existing_context', 'existing_conversation_id', 'analytics_owned']) {
-        const tool = listResult.tools.find((candidate) => candidate.name === name)
-        expect(tool?.inputSchema.properties?.context).toBeDefined()
-        expect(tool?.inputSchema.properties?.conversation_id).toBeDefined()
-      }
+      const listedTools = new Map(listResult.tools.map((tool) => [tool.name, tool]))
+      expect(listedTools.get('existing_context')?.inputSchema.properties?.context).toMatchObject({
+        description: 'Application context',
+      })
+      expect(listedTools.get('existing_context')?.inputSchema.properties?.conversation_id).toMatchObject({
+        description: DEFAULT_CONVERSATION_ID_DESCRIPTION,
+      })
+      expect(listedTools.get('existing_context')?.inputSchema.properties?.properties).toBeDefined()
+      expect(listedTools.get('existing_conversation_id')?.inputSchema.properties?.context).toMatchObject({
+        description: DEFAULT_CONTEXT_PARAMETER_DESCRIPTION,
+      })
+      expect(listedTools.get('existing_conversation_id')?.inputSchema.properties?.conversation_id).toMatchObject({
+        description: 'Application conversation',
+      })
+      expect(listedTools.get('analytics_owned')?.inputSchema.properties?.context).toMatchObject({
+        description: DEFAULT_CONTEXT_PARAMETER_DESCRIPTION,
+      })
+      expect(listedTools.get('analytics_owned')?.inputSchema.properties?.conversation_id).toMatchObject({
+        description: DEFAULT_CONVERSATION_ID_DESCRIPTION,
+      })
 
       const suppliedArguments = {
         context: 'supplied context',
--- a/packages/mcp/src/__tests__/reserved-arguments.test.ts
+++ b/packages/mcp/src/__tests__/reserved-arguments.test.ts
@@ -1,0 +1,193 @@
+import { Client } from '@modelcontextprotocol/sdk/client/index.js'
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
+import { z } from 'zod'
+import { instrument } from '../index'
+import { fakePostHog } from './test-utils'
+
+async function connect(server: McpServer) {
+  const client = new Client({ name: 'reserved-argument-test-client', version: '1.0.0' })
+  const [clientTransport, serverTransport] = InMemoryTransport.createLinkedPair()
+  await Promise.all([client.connect(clientTransport), server.server.connect(serverTransport)])
+
+  return {
+    client,
+    async cleanup() {
+      await clientTransport.close?.()
+      await serverTransport.close?.()
+    },
+  }
+}
+
+describe('high-level reserved analytics arguments', () => {
+  it('passes legitimate context and conversation_id fields to the callback when both features are disabled', async () => {
+    const server = new McpServer({ name: 'disabled-reserved-arguments', version: '1.0.0' })
+    let receivedArgs: Record<string, unknown> | undefined
+
+    server.registerTool(
+      'reserved_fields',
+      {
+        inputSchema: z.object({
+          context: z.string(),
+          conversation_id: z.string(),
+          value: z.string(),
+        }),
+      },
+      async (args) => {
+        receivedArgs = { ...args }
+        return { content: [{ type: 'text', text: 'ok' }] }
+      }
+    )
+
+    const { client, cleanup } = await connect(server)
+    try {
+      instrument(server, fakePostHog(), { context: false, enableConversationId: false })
+
+      await client.request(
+        {
+          method: 'tools/call',
+          params: {
+            name: 'reserved_fields',
+            arguments: { context: 'tool context', conversation_id: 'tool conversation', value: 'kept' },
+          },
+        },
+        CallToolResultSchema
+      )
+
+      expect(receivedArgs).toEqual({
+        context: 'tool context',
+        conversation_id: 'tool conversation',
+        value: 'kept',
+      })
+    } finally {
+      await cleanup()
+    }
+  })
+
+  it('recomputes ownership after a schema-only tool update', async () => {
+    const server = new McpServer({ name: 'updated-reserved-arguments', version: '1.0.0' })
+    const receivedArgs: Record<string, unknown>[] = []
+
+    const registeredTool = server.registerTool(
+      'updated_schema',
+      { inputSchema: z.object({ value: z.string() }).passthrough() },
+      async (args) => {
+        receivedArgs.push({ ...args })
+        return { content: [{ type: 'text', text: 'ok' }] }
+      }
+    )
+
+    const { client, cleanup } = await connect(server)
+    try {
+      instrument(server, fakePostHog(), { context: true, enableConversationId: true })
+
+      const argumentsWithReservedFields = {
+        context: 'supplied context',
+        conversation_id: 'supplied conversation',
+        value: 'kept',
+      }
+      await client.request(
+        { method: 'tools/call', params: { name: 'updated_schema', arguments: argumentsWithReservedFields } },
+        CallToolResultSchema
+      )
+
+      registeredTool.update({
+        paramsSchema: {
+          context: z.string(),
+          conversation_id: z.string(),
+          value: z.string(),
+        },
+      })
+
+      await client.request(
+        { method: 'tools/call', params: { name: 'updated_schema', arguments: argumentsWithReservedFields } },
+        CallToolResultSchema
+      )
+
+      expect(receivedArgs).toEqual([
+        { value: 'kept' },
+        {
+          context: 'supplied context',
+          conversation_id: 'supplied conversation',
+          value: 'kept',
+        },
+      ])
+    } finally {
+      await cleanup()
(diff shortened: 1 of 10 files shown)


marandaneto:
@PostHog/team-mcp-analytics unless removing both props were intentional

turnipdabeets (approved):

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -44,6 +45,47 @@ function isCallbackUpdate(value: unknown): value is { callback: ToolCallback } {
...
+    return false
+  }
+
+  if (schema.properties && typeof schema.properties === 'object') {
+    return !Object.prototype.hasOwnProperty.call(schema.properties, parameterName)
+  }
+
+  // A non-object Zod schema is converted to a complex JSON schema that the
+  // list instrumentation deliberately leaves alone.
+  if ('_def' in schema || '_zod' in schema || '~standard' in schema) {
```
gesh, line 70: 🔴 **Blocking — this is a behavioural regression vs `main`, not a theoretical one.**

The premise in the comment ("a non-object Zod schema is converted to a complex JSON schema that the list instrumentation deliberately leaves alone") doesn't hold for `z.record(...)`: it converts to `{ type: 'object', additionalProperties: {...} }` with no `oneOf`/`allOf`/`anyOf`, so `addContextParameterToTool` and `addConversationIdToTool` **do** inject. This branch then says analytics doesn't own those parameters, so nothing strips them — and a record schema validates the injected keys straight through into the tool's arguments.

Verified with SDK 1.29, `registerTool('rec', { inputSchema: z.record(z.string()) }, …)`; `tools/list` advertises injected `context` + `conversation_id` on **both** branches:

```
main:     tool receives {"value":"v"}                                        ✅ stripped
this PR:  tool receives {"value":"v","context":"ctx","conversation_id":"cid"}  ❌ leaked
```

`z.object().catchall(...)` is fine — still a ZodObject, so `getObjectShape` resolves and stripping stays correct. It's specifically the schemas that fall through to this `_def` branch while still accepting unknown keys.

Root cause: ownership is inferred from the **registered Zod schema** while injection decides from the **converted JSON Schema**, and the two can't be kept in sync by sniffing. Cleanest fix is one shared predicate over the same converted schema (the `oneOf`/`allOf`/`anyOf` + `properties.<name>` rules already in `context-parameters.ts` / `conversation-id.ts`), so "we injected it" and "we strip it" can never disagree. Cheaper fix: make this branch default to *owned* (strip) unless the schema really is complex — which is effectively what `main` did.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -44,6 +45,47 @@ function isCallbackUpdate(value: unknown): value is { callback: ToolCallback } {
...
+  if (shape) {
+    return !Object.prototype.hasOwnProperty.call(shape, parameterName)
+  }
+
+  const schema = inputSchema as Record<string, unknown>
+  if (schema.oneOf || schema.allOf || schema.anyOf) {
+    return false
+  }
+
+  if (schema.properties && typeof schema.properties === 'object') {
```
gesh: Same root cause as my comment on line 69, smaller blast radius: this JSON-Schema sniff runs *before* the raw-Zod-shape fallback at line 74, so a raw shape containing a parameter literally named `properties` (or `oneOf`/`anyOf`/`allOf`, or `_def`) is misclassified. Shape `{ properties: z.string(), context: z.string() }` takes this branch, `properties.hasOwnProperty('context')` is false, and the user's own `context` gets stripped — exactly the bug this PR is fixing.

If you go with the shared-predicate fix this disappears on its own. Otherwise: detect a raw shape first — the SDK's own `isZodRawShapeCompat` just does `Object.values(obj).some(isZodTypeLike)`.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -78,18 +120,21 @@ function setupListenerToRegisteredTools(server: HighLevelMCPServerLike): void {
...
+            const getCurrentInputSchema = () => value.inputSchema
+            const nextValue = addTracingToToolCallbackInternal(value, property, server, getCurrentInputSchema)
 
             if (typeof nextValue.update === 'function') {
               const originalUpdate = nextValue.update
               nextValue.update = function (...updateArgs: unknown[]) {
                 if (updateArgs[0]) {
                   const updateObj = updateArgs[0]
                   if (isCallbackUpdate(updateObj)) {
+                    const updatedTool = createWrappedTool(nextValue, updateObj.callback)
```
gesh: 🔴 **Blocking (latent):** this hands the *raw* user callback back, unwrapped.

`createWrappedTool` does `{ ...originalTool, [key]: fn }`, and object spread copies enumerable **symbol** keys — so `updatedTool` carries `MCP_ANALYTICS_PROCESSED` (set at line 253 by plain assignment, hence enumerable). `addTracingToToolCallbackInternal` then hits the guard at line 201, logs "already processed" and returns the tool untouched, so line 139 assigns the unwrapped callback onto the update: no `context`/`conversation_id` stripping and no `extra.__mcp_analytics_error` capture for that tool.

Before this PR the bare `{ callback: updateObj.callback }` object carried no symbol and was wrapped correctly.

Simplest fix is to keep the old shape — `getCurrentInputSchema` is already passed explicitly, so `createWrappedTool` isn't buying anything here:

```ts
const wrappedTool = addTracingToToolCallbackInternal({ callback: updateObj.callback }, property, server, getCurrentInputSchema)
```

To be fair on severity: I couldn't make this observable, because the whole `update()` interception is currently dead code (see line 190) — the SDK hands users the original tool object, not the wrapped copy stored in the registry, so this patched `update` never runs. It's still wrong as written, and it turns into a live regression the day that's fixed.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -141,12 +186,12 @@ function setupListenerToRegisteredTools(server: HighLevelMCPServerLike): void {
 function addTracingToToolCallbackInternal(
   tool: RegisteredTool,
   toolName: string,
-  server: HighLevelMCPServerLike
+  server: HighLevelMCPServerLike,
+  getCurrentInputSchema: () => unknown = () => tool.inputSchema
```
gesh, line 192: 🟡 **Follow-up (pre-existing) + one small ask on this PR.**

"Recompute ownership after registered tool schema updates" doesn't hold end-to-end. Verified with SDK 1.29: register `t` with `z.object({ value })`, `instrument()`, then `t.update({ paramsSchema: { context, conversation_id, value } })`, then list + call:

```
LIST SCHEMA:   properties = { value, context (analytics description), conversation_id }   // old schema
TOOL RECEIVED: {"value":"v"}                                                             // Zod stripped them anyway
```

The client still sees the **old** schema and the SDK validates against it, so `getCurrentInputSchema()` reading the live object can't change the outcome. The new test only shows a difference because it uses `.passthrough()`.

Root cause is pre-existing and well outside this PR: instrumentation stores a *copy* of the registered tool in `_registeredTools`, while `registerTool()` returns the original object whose `update`/`enable`/`disable` closures mutate that original. Confirmed identical on `main` and this branch — after `instrument()`, `update({callback})`, `update({description})` and `disable()` are **all** silently ignored:

```
CALLS: [["orig", …]]     // swapped callback never runs
DESC:  [["t", null]]     // description update invisible to clients
AFTER DISABLE: ["t"]     // still listed
```

Worth its own PR (mutate in place, or keep a live reference instead of a copy) — it'd also make the `update({callback})` wrapper on line 132 real again. **The only ask here:** soften or drop that bullet from the PR description/changeset, since as shipped it doesn't reach clients. Keeping `getCurrentInputSchema` is fine — it becomes correct once the copy issue is fixed.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -170,15 +215,17 @@ function addTracingToToolCallbackInternal(
...
-    const cleanedArgs = toolName === missingToolName ? args : stripConversationId(removeContextFromArgs(args))
+    const inputSchema = getCurrentInputSchema()
+    const cleanedArgs = stripOwnedAnalyticsArguments(args, {
+      context:
+        toolName !== missingToolName &&
+        isContextEnabled(options?.context) &&
+        analyticsOwnsParameter(inputSchema, 'context'),
+      conversationId:
+        toolName !== missingToolName &&
+        options?.enableConversationId === true &&
```
gesh, line 228: 🟡 **Follow-up:** ownership is applied to what the *tool* receives, but not to what *analytics* consumes — so the same idea is half-applied. Verified on this branch with a tool declaring its own `context` (required) and `conversation_id` (optional), agent omitting the latter:

```
TOOL GOT:  {"context":"internal app state blob","value":"v"}                                  ✅ this PR
EVENT:     $mcp_tool_call  intent="internal app state blob"  conversation_id="019fb873-…"      ❌
RESULT:    […, "[SERVER]: Reuse conversation_id=019fb873-… on every subsequent tool call…"]    ❌
```

So (1) `intent.ts:59` records the tool's own `context` value as `$mcp_intent` — arbitrary, potentially sensitive application data showing up as user intent; and (2) `resolveConversationId` mints an id and the prompt-back tells the agent to send `conversation_id` on the next call, straight into the tool's own parameter. A user-supplied value is likewise adopted as the analytics conversation id.

Not a blocker — the tool-facing fix here is still a strict improvement — but the same ownership predicate wants threading into `resolveToolCallIntent` and `resolveConversationId` in a follow-up.

Tiny nit while you're here: `options?.enableConversationId === true` reads the option differently from `instrumentation.ts:94` (`?? false`) and `instrumentation.ts:363` (truthy). An `isConversationIdEnabled(options)` helper next to `isContextEnabled` would keep all three honest.

packages/mcp/src/__tests__/reserved-arguments.test.ts
```diff
@@ -0,0 +1,193 @@
...
+
+      expect(receivedArgs).toEqual({
+        context: 'tool context',
+        conversation_id: 'tool conversation',
+        value: 'kept',
+      })
+    } finally {
+      await cleanup()
+    }
+  })
```
gesh, line 67: This case passes only because the initial schema is `.passthrough()` — with a plain `z.object` the SDK strips unknown keys during validation and both calls look identical, so it isn't really exercising the recompute. And per my comment on `instrument-highlevel.ts:190`, the client-visible schema and the validation schema never actually change here, so the scenario asserts behaviour no real consumer can observe.

The coverage I'd most like to see added in this PR is a tool whose `inputSchema` is a non-object Zod schema that still accepts unknown keys — `z.record(z.string())` — asserting the injected `context`/`conversation_id` do **not** reach the callback. That's the regression flagged on line 69, and it's currently green because no test covers that shape.

Follow-up coverage (fine after this PR): `update({ callback })` interception, analytics-side assertions for tool-owned params, and the low-level-server equivalent.

packages/mcp/src/__tests__/reserved-arguments.test.ts
```diff
@@ -0,0 +1,193 @@
...
+      { inputSchema: z.object({ value: z.string() }).passthrough() },
+      async (args) => {
+        receivedArgs.set('analytics_owned', { ...args })
+        return result
+      }
+    )
+
+    const { client, cleanup } = await connect(server)
+    try {
+      instrument(server, fakePostHog(), { context: true, enableConversationId: true })
```
gesh, line 146: Nit: for `existing_context` / `existing_conversation_id` these assertions are satisfied by the tool's *own* declared property, so they'd still pass if injection clobbered it. The meaningful assertion is that injection was **skipped** — e.g. that the description is the tool's own (`'Application context'`), not `DEFAULT_CONTEXT_PARAMETER_DESCRIPTION`.

gesh (commented):
## Review — verified locally against a real MCP SDK 1.29 server

Checked out the branch, ran `packages/mcp` jest (**35 suites / 400 tests pass**) and `eslint src` (clean), then drove a real `McpServer` over `InMemoryTransport` on **`main` vs this branch** so every claim below is observed behaviour, not inference.

**The core fix works.** Tool declaring its own `context` / `conversation_id` in the schema:

```
main:     callback receives {"value":"v"}                                        ❌ user data destroyed (the reported bug)
this PR:  callback receives {"context":"ctx","conversation_id":"cid","value":"v"}  ✅ fixed
analytics-owned tool (no such params declared): still stripped correctly           ✅ no leak
```

### 🔴 Must fix before merge — one genuine regression + one latent defect

**R1. Analytics-injected args now leak into tools whose `inputSchema` is a non-object Zod schema** (`instrument-highlevel.ts:69`). For `z.record(...)`, `tools/list` still injects `context`/`conversation_id` (its JSON Schema has no `oneOf`/`allOf`/`anyOf`), but the new `_def` branch declares analytics *doesn't* own them, so they're no longer stripped and land in the tool's arguments. `main` stripped them:

```
inputSchema: z.record(z.string())   —  both branches advertise injected context + conversation_id
main:     {"value":"v"}                                        ✅
this PR:  {"value":"v","context":"ctx","conversation_id":"cid"}  ❌
```

**R2. `update({ callback })` re-registers an *unwrapped* callback** (`instrument-highlevel.ts:132`). `createWrappedTool` spreads the enumerable `MCP_ANALYTICS_PROCESSED` symbol onto the new object, so the "already processed" guard short-circuits and the raw callback is handed back. Not observable today because that branch is unreachable (see F1) — but it's wrong as written and becomes a live regression the moment F1 is fixed.

### 🟡 Possible follow-ups — all pre-existing, none of them should block this PR

**F1. Instrumentation makes `RegisteredTool.update()` / `enable()` / `disable()` no-ops.** `_registeredTools` ends up holding a *copy* of each tool while the SDK's closures mutate the original the user holds. Verified identical on `main` and this branch: after `instrument()`, a callback swap, a description change and `disable()` are all silently ignored. This is also why the *"Recompute ownership after registered tool schema updates"* bullet in the description doesn't hold end-to-end — see the comment on line 190. Only ask for this PR: soften or drop that bullet so the changelog doesn't promise something the current architecture can't deliver.

**F2. Ownership is applied tool-side but not analytics-side** (`instrument-highlevel.ts:226`). A tool that owns `context` still has its value captured as `$mcp_intent`; a tool that owns an optional `conversation_id` still gets an analytics-minted UUID plus the `[SERVER]: Reuse conversation_id=…` prompt-back, which then feeds that UUID into the tool's own parameter. Same ownership predicate wants threading into `resolveToolCallIntent` / `resolveConversationId`.

**F3. The low-level path still has the original bug** (no inline comment — file isn't in the diff). `instrument-lowlevel.ts` → `captureToolCall` → `cloneRequestWithoutConversationId` strips `conversation_id` from the request unconditionally when the feature is on, so a low-level server that owns that parameter still loses it.

**F4. `z.object().strict()` tools fail *every* call when context injection is on:** `MCP error -32602: Input validation error: … Unrecognized key(s) in object: 'context'`. Clearing `additionalProperties` in the advertised JSON Schema (`context-parameters.ts:91`) doesn't help, because Zod — not the JSON Schema — does the validation. Probably deserves its own issue.

### Test gaps
- A `z.record` / non-object-schema tool (would have caught R1).
- `update({ callback })` interception (would catch R2 once F1 is fixed).
- Analytics-side assertions — `$mcp_intent`, conversation id, prompt-back — for tool-owned params (F2).
- Low-level-server equivalent of the new suite (F3).

Overall: real fix for a real bug and I'd like to see it land — just R1 and R2 first; the rest are happily follow-ups.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Good catch! Fixed in 18a45645d — ownership now shares the same injection predicate, and non-object Zod schemas such as z.record default to the empty object schema behavior used by the MCP SDK. Added an end-to-end regression test.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Fixed in 18a45645d — callback updates are wrapped from a bare callback object, so the processed symbol can no longer short-circuit wrapping.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Agreed that analytics-side ownership (and the low-level equivalent) should be handled in a follow-up. I kept this PR scoped to the tool-facing fix and left this thread open for that follow-up.

packages/mcp/src/__tests__/reserved-arguments.test.ts
marandaneto: Fixed in 18a45645d — replaced the schema-update test with z.record(z.string()) coverage that verifies context and conversation_id are advertised but stripped before the callback.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Fixed in 18a45645d — raw Zod shapes are detected before JSON Schema inspection. The end-to-end collision test now uses a raw shape with a parameter literally named properties.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Agreed. I removed the schema-update claim from the PR description and replaced the unobservable schema-update test in 18a45645d. The pre-existing registry-copy issue remains follow-up work.

packages/mcp/src/__tests__/reserved-arguments.test.ts
marandaneto: Fixed in 18a45645d — list assertions now verify the original application descriptions are preserved and analytics-owned fields use the SDK defaults.

gesh (approved):
## Re-reviewed at `18a45645` — both blockers verified fixed ✅

Re-ran the probes against a real SDK 1.29 server rather than taking the fixes on trust. `35 suites / 402 tests` pass, `eslint src` clean.

**R1 (`z.record` leak) — fixed.** Ownership and injection now share `canInjectAnalyticsParameter`, and the non-object-schema branch flipped to *owned*:

```
z.record(z.string()):   tool receives {"value":"v"}                                   ✅ (was leaking both params)
tool-owned context/cid: {"context":"ctx","conversation_id":"cid","value":"v"}          ✅ core fix intact
```

I also checked the new branch's premise holds for every non-object schema I could construct — `z.record`, `z.union`, `.refine()`, `.optional()` are all advertised as `EMPTY_OBJECT_JSON_SCHEMA` by the SDK (`mcp.js:75`), so injection and stripping can't disagree and no `anyOf` mismatch is reachable. Zod v4 object schemas resolve through `getObjectShape` correctly too (not misdetected as raw shapes).

**R2 (processed-symbol short-circuit) — fixed** by going back to the bare `{ callback }` object.

**Nits — fixed, and better than suggested:** raw shapes are detected before JSON-Schema inspection via the new `isZodRawShapeCompat`, with the end-to-end tool actually declaring a `properties` parameter to lock it in; the list assertions now verify the application's own descriptions survive versus `DEFAULT_*` for injected ones. Nice catch on `hasOwnProperty` vs truthiness — `properties: { context: false }` counting as declared is a case I'd missed, and it's covered in both unit suites.

Thanks for dropping the schema-update claim from the description and swapping the unobservable test for the `z.record` regression test — that's exactly the resolution I was after.

### Two residual notes, neither blocking

- `getCurrentInputSchema` is still plumbed through but now has no test, and stays inert until the stale-registry-copy issue is fixed. Worth a short comment saying so, or dropping it until then.
- Worth one line under *Known sharp edges* in `docs/ARCHITECTURE.md`: with a non-object schema (e.g. `z.object({ context, value }).refine(...)`) the SDK advertises no parameters at all, so a tool that declares its own `context` there still won't receive it. Consistent with what's advertised, but silent.

### Still open as agreed follow-ups

- Analytics-side ownership: `$mcp_intent` from a tool-owned `context`, plus conversation-id minting and the prompt-back.
- The low-level path — `cloneRequestWithoutConversationId` still strips unconditionally.
- Pre-existing: `z.object().strict()` tools fail every call while context injection is on.

LGTM.

marandaneto:
@gesh Thanks for the re-review! Addressed both residual notes in a5ddee719: documented why the lazy schema lookup remains despite the current stale-registry limitation, and added the non-object/refined Zod schema behavior under Known sharp edges.

PR #4356, opened 2026-07-31
fix(mcp): preserve real missing-capability tools

marandaneto:
## Problem

MCP server authors can legitimately register a real tool with the same name configured for `reportMissing` (by default, `get_more_tools`). The SDK previously treated every matching call as the virtual missing-capability tool even when descriptor injection was skipped because of the collision, so the real handler could be bypassed.

## Changes

- Determine virtual-tool ownership statelessly at call time from the server's raw tool listing, so missing-capability reporting works before `tools/list` and across stateless pods.
- Warn on default or custom name collisions while preserving the real tool descriptor and handler.
- Fail open to the server's real dispatcher when ownership cannot be determined, and apply normal context/conversation parameter handling to colliding real tools.
- Preserve tool-call analytics when low-level servers register their dispatcher after `instrument()`.
- Add high- and low-level coverage for disabled reporting, default/custom collisions, pre-list and cross-pod calls, late handler registration, and normal virtual-tool behavior.
- Document the ownership and collision behavior in the MCP architecture notes.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human directed the scope and submission. Pi's coding agent was used to validate and submit the existing fix. The implementation determines virtual-tool ownership statelessly from the server's raw tool listing; collisions warn and preserve the real descriptor and handler rather than intercepting it.


marandaneto:
**TL;DR:** When `reportMissing` is enabled, PostHog exposes a virtual tool (by default `get_more_tools`) for reporting capabilities that an MCP server does not provide. If the server already had a real tool with that name, the SDK previously skipped advertising the duplicate virtual tool but still intercepted calls by name, so the real handler never ran. This PR makes PostHog claim and intercept the name only after it successfully injects the virtual tool into `tools/list`. On a collision, it warns and leaves the real descriptor and handler untouched; when there is no collision, missing-capability reporting continues to work as before.

packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -360,15 +360,27 @@ async function getTracedToolsList(
...
+      if (data.options.reportMissing) {
+        const missingToolName = resolveMissingCapabilityToolName(data.options)
+        const alreadyPresent = tools.some((tool) => tool?.name === missingToolName)
+        if (alreadyPresent) {
+          log(
+            `Warning: Cannot inject missing-capability tool "${missingToolName}" because a real tool already uses that name. The real tool will not be intercepted.`
+          )
+        } else {
+          tools.push(getReportMissingToolDescriptor(missingToolName))
+          data.missingCapabilityToolInjected = true
```
turnipdabeets: Persisting the injection fact in per-instance state breaks `reportMissing` for the stateless/multi-pod deployments ARCHITECTURE.md documents ("which a stateless pod resets on every request"): pod A serves `tools/list` and advertises `get_more_tools`, the `tools/call` lands on pod B where this flag is still false, and the agent gets `MCP error -32602: Tool get_more_tools not found` instead of the canned reply — reproduced with a fresh instrumented instance receiving `tools/call` without a prior `tools/list` (works on main, fails on this branch, and no `$mcp_missing_capability` event is captured). I think the collision check needs to be stateless — e.g. treat the name as virtual when `reportMissing` is on and no real tool owns it at call time (high-level can consult the registered tools; low-level could try the original handler first and only intercept unknown-tool) — rather than remembering whether this instance injected the descriptor.

marandaneto: Good catch — fixed in 4e2d711117f342a930eda78f187a694d63ede2bf. Missing-capability ownership is now determined statelessly from the server’s raw tool listing, with real collisions delegated and listing failures failing open. Added separate-pod, high/low-level collision, disabled-tool, late-registration, and reused/frozen-list regression coverage.

ioannisj:
@marandaneto some conflicts need to be resolved here

packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -276,14 +294,51 @@ export function patchRequestHandlers(server: MCPServerLike, patches: Record<stri
...
+
+  try {
+    const response = (await listHandler({ method: 'tools/list', params: {} }, extra)) as ListToolsResult
+    if (!response || !Array.isArray(response.tools)) {
+      return undefined
+    }
+    // Match the page the current list instrumentation can expose. Pagination
+    // passthrough is handled separately from missing-capability ownership.
+    return response.tools.some((tool) => tool?.name === toolName)
+  } catch {
```
ioannisj: Worth a log here

ioannisj (approved):

packages/mcp/src/extensions/instrumentation.ts
marandaneto: Good call — fixed in 8ba3ce260. A failed raw tools/list ownership probe now emits a warning through the configured logger before failing open to the server; it is not captured as an SDK exception. Added regression coverage for the warning.

lucasheriques (approved):
Approving — the collision fix is right and the coverage is genuinely thorough. The pod-A-advertises / pod-B-serves case and the `[...tools]` copy in `getTracedToolsList` (a server returning a cached array was getting a duplicate `get_more_tools` appended on every listing) are both good catches.

I dug into this fairly deeply and everything I found is non-blocking — we'll pick it up in follow-ups rather than hold this. Two things only you can adjust, both zero-cost:

- The description still describes the `missingCapabilityToolInjected` design that the second commit replaced with the live `isToolAdvertised` probe, so "leave calls before virtual-tool advertisement with the server's real handler" now reads backwards — a pre-list call reaching the virtual tool is the whole point of the stateless fix. `ARCHITECTURE.md` is accurate; just the description drifted.
- The changeset only mentions the collision fix, but adding `'tools/call'` to `patchRequestHandlers` fixes something bigger: previously a low-level `Server` that registered its dispatcher *after* `instrument()` — the documented `@rekog/mcp-nest` / `serverMutator` path — had its analytics wrapper replaced and emitted **zero** `$mcp_tool_call` events. Your new test covers it. Worth its own changelog line, since affected users will see tool-call volume appear out of nowhere on upgrade.

marandaneto:
@lucasheriques Thanks — addressed both follow-ups. I updated the PR description to reflect stateless `isToolAdvertised` routing and added the late low-level handler analytics fix to the changeset in f63fb2ba6.

marandaneto:
@ioannisj Resolved — I merged the latest `main`, resolved the MCP conflicts, and pushed the result in 95534b48e.

PR #4357, opened 2026-07-31
fix(mcp): prevent concurrent attribution leaks

marandaneto:
## Problem

Overlapping MCP requests can yield while resolving identity, metadata, or tool results. Attribution was then read from shared mutable server session state at publish time, allowing one request to inherit another request's identity, client, or protocol metadata.

## Changes

- Snapshot session, client, protocol, and identity attribution before request instrumentation yields, then carry that snapshot through event publication.
- Preserve request-scoped attribution for tool calls, tool listings, identify events, and initialization while retaining the existing fallback for custom capture calls.
- Add regression coverage for concurrent generated-session identities and session-token client/protocol metadata.
- Add a patch changeset for `@posthog/mcp`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A human directed the scoped fix and PR submission. Pi's GPT-5.6 Sol coding agent assisted using local Git, Jest, ESLint, and Rslib tooling; the agent session was a local run without an externally accessible session link.

The implementation keeps the existing custom-capture fallback and scopes the attribution snapshot only to instrumented request paths. Concurrency regression tests cover both generated identities and token-carried client/protocol metadata.


marandaneto:
One important scope clarification: this fixes request-scoped attribution for auto-captured MCP events, but custom events emitted through `analytics.capture()` remain concurrency-unsafe. The public `McpAnalytics.capture` / `CaptureEventData` API still provides no request-scoped attribution context, so `captureCustomEvent()` continues to read the shared mutable `trackingData.sessionId` and falls back to the current shared session/identity/client/protocol state. Until the public API can accept request attribution, the PR description and changeset should explicitly call out that custom captures are not covered by this fix.

packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -137,10 +153,14 @@ async function prepareToolCallEvent(
...
   conversation: ConversationIdResolution,
   eventType: MCPAnalyticsEventType
-): Promise<McpEvent | null> {
+): Promise<PreparedToolEvent | null> {
   try {
     const sessionId = getSessionId(server, extra)
-    await handleIdentify(server, data, sessionId, request, extra)
+    // Snapshot token/client/protocol metadata synchronously, before identify or
+    // metadata callbacks can yield and let another request replace shared state.
+    const sessionInfo = getSessionInfo(server, data, sessionId)
```
turnipdabeets: [suggestion] This snapshot is the same object `getSessionInfo` stores as `data.sessionInfo`, so during the `handleIdentify` await a concurrent request's `getSessionId` can still mutate it in place — reproduced: request A (token `client-a`) parks in its identify callback, request B (`client-b`) runs to completion, and A's `$mcp_tool_call` publishes `$mcp_client_name: "client-b"` (pre-existing on main too, but the comment above overclaims that the snapshot beats the yield). A shallow copy closes the window; the same treatment in `handleListToolsRequest`/`handleInitializeRequest`, plus dropping the in-place `identifyActor*` writes in `handleIdentify`, also resolves the tools/list corruption flagged above.
```suggestion
    const sessionInfo = { ...getSessionInfo(server, data, sessionId) }
```

packages/mcp/src/extensions/internal.ts
```diff
@@ -167,17 +170,31 @@ export async function handleIdentify(
       const shouldPublish = changed || (firstSeen && !announcedAtInitialize)
 
       data.identifiedSessions.set(sessionId, mergedIdentity)
+      data.sessionInfo.identifyActorGivenId = mergedIdentity.distinctId
+      data.sessionInfo.identifyActorData = mergedIdentity.properties || {}
+      data.sessionInfo.identifyActorGroups = mergedIdentity.groups
```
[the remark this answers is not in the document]
marandaneto: Fixed in 973c5f8d440eec823aad2fbd424d8d5abd4fc1a7. Removed the shared identifyActor mutations and added request-local snapshots plus deterministic concurrent tool/list/initialize regression tests.

packages/mcp/src/extensions/instrumentation.ts
marandaneto: Good catch — fixed in 973c5f8d440eec823aad2fbd424d8d5abd4fc1a7. Tool, tools/list, and initialize paths now take request-local session snapshots before yielding; initialize token upgrades use that snapshot, and identify uses the pre-await session source. Added deterministic interleaving regressions.

ioannisj (approved):

packages/mcp/src/extensions/capture.ts
```diff
@@ -53,9 +60,9 @@ export function captureEvent(server: MCPServerLike, eventInput: McpEvent): Promi
     // request from another client can't misattribute this one.
     clientName: eventInput.clientName ?? sessionInfo.clientName,
     clientVersion: eventInput.clientVersion ?? sessionInfo.clientVersion,
-    identifyActorGivenId: sessionInfo.identifyActorGivenId,
-    identifyActorData: sessionInfo.identifyActorData,
-    groups: sessionInfo.identifyActorGroups,
+    identifyActorGivenId: eventInput.identifyActorGivenId ?? sessionInfo.identifyActorGivenId,
```
lucasheriques: these three `eventInput.* ??` fallbacks look unreachable. the only thing that sets `identifyActorGivenId` / `identifyActorData` / `groups` on an event is posthog-mcp.ts:252, and that goes through `#emit -> sink.capture` without touching `captureEvent`, which also is not exported from index.ts.

so this advertises an event-level override nothing can actually use. `sessionInfo.*` on its own would say what is true.

packages/mcp/src/extensions/internal.ts
```diff
@@ -163,21 +167,32 @@ export async function handleIdentify(
...
     }
+
+    log(`Warning: Supplied identify function returned null for session ${sessionId}`)
   } catch (error) {
     log(`Error: User supplied identify function threw an error while identifying session ${sessionId} - ${error}`)
   }
+  return identityBeforeRequest
+}
+
+function withIdentity(sessionInfo: SessionInfo, identity: UserIdentity): SessionInfo {
```
lucasheriques: `withIdentity` is now defined twice, here and instrumentation.ts:138, differing only in whether `identity` is optional. both added in this pr, so probably worth exporting one before they drift.

packages/mcp/src/__tests__/concurrent-attribution.test.ts
```diff
@@ -0,0 +1,384 @@
...
+    })
+    expect(captureB?.properties).toMatchObject({
+      $session_id: 'ses_b',
+      $mcp_client_name: 'client-b',
+      $mcp_client_version: '2.0.0',
+      $mcp_protocol_version: '2025-06-18',
+    })
+  })
+
+  it('keeps tools/list anonymous while another request finishes identifying', async () => {
```
lucasheriques, line 183: this one passes on main, so it is guarding the in-place mutation from the intermediate commit rather than the bug being shipped. fine to keep, just noting it is not evidence for this fix the way the other five are.

two gaps i would fill at some point: no concurrent case on the error paths, which is where attribution is easiest to lose (`publishFailedToolEvent`, and the `getTracedToolsList` catch). and the `await Promise.resolve()` flushes here differ from the `setTimeout(r, 50)` the rest of the suite uses. that works today because the sink pipeline is microtask-only, but one `await` behind a timer in `processMcpEvent` and these either fail intermittently or pass for the wrong reason.

lucasheriques (approved):
the snapshot-before-first-await approach holds everywhere i checked: `getSessionId` and `getSessionInfo` are both called with no await before the shallow copy at instrumentation.ts:158, :321 and :580, `identityBeforeRequest` is read in the synchronous prologue at internal.ts:136, and the error paths carry the same `requestAttribution`. 5 of the 6 new tests fail on main for specific reasons, so they are real regression tests rather than restatements.

also confirming greptile p1 is dealt with: no writes to `data.sessionInfo.identifyActor*` remain anywhere, only the clientName/clientVersion/protocolVersion ones in session.ts and instrumentation.ts, which are not what it was about.

one thought that does not fit on a line in this diff. the root hazard is `getSessionInfo` mutating `data.sessionInfo` and also handing back the live reference, and this pr works around that at three call sites with a spread. `return { ...sessionInfo }` in session.ts:126 would fix it once and mean a fourth caller cannot forget. worth it while this is fresh, since the three spreads are load-bearing and silent if one goes missing.

three small things inline.

packages/mcp/src/extensions/capture.ts
marandaneto: Good catch — fixed in c83f922c6. `captureEvent` now takes identity and groups exclusively from the request/session attribution snapshot; the unreachable event-level identity fallbacks were removed.

packages/mcp/src/__tests__/concurrent-attribution.test.ts
marandaneto: Addressed in c83f922c6. I kept the existing regression, added deterministic concurrent failure coverage for both `publishFailedToolEvent` and the `getTracedToolsList` catch path, and replaced the microtask-only flushes with timer-backed capture flushing.

packages/mcp/src/extensions/internal.ts
marandaneto: Fixed in c83f922c6. `withIdentity` is now exported once from `internal.ts` and reused by instrumentation.

marandaneto:
Addressed the remaining review-summary suggestion in c83f922c6 as well: `getSessionInfo` now returns a shallow request-local snapshot centrally, and the three load-bearing caller spreads were removed. Also merged the latest `main` and preserved the intervening MCP logger, reserved-argument, and missing-tool changes.

PR #4359, opened 2026-07-31
fix(mcp): isolate loggers per server

marandaneto:
## Problem

Instrumenting multiple MCP servers in one process configured logging through shared module state. A later server could therefore redirect earlier servers' logs, while a server without a logger could inherit another server's or the legacy logger. This breaks server isolation and can leak operational details across server boundaries.

## Changes

- Bind a safe logger to each instrumented server and store it with that server's tracking data.
- Route instrumentation, compatibility, context, conversation, intent, capture sink, and tool logs through the server-bound logger.
- Keep the exported legacy `setLogger`/`log` behavior available for backwards compatibility, while loggerless instrumented servers remain no-ops.
- Avoid serializing identified user data into log messages.
- Add concurrent multi-server and loggerless-server regression coverage.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

- A human directed the scoped fix and PR submission; the Pi coding agent used repository-local Jest, ESLint, and Rslib binaries plus GitHub CLI.
- The existing per-server logger design was retained. The agent added the requested patch changeset, validated the complete MCP package, and merged the latest `origin/main` without rewriting branch history before opening this PR.


.changeset/calm-logs-stay.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/mcp': patch
+---
+
+Isolate logger configuration per instrumented MCP server.
```
turnipdabeets, line 5: [suggestion] This also changes public `setLogger` behavior: on main a loggerless instrumented server fell back to the global logger, so someone debugging with `setLogger(console.error)` loses all `instrument()` logs in a patch release. Could the entry say that?
```suggestion
Scope internal SDK logs to each instrumented server's own `logger` option — servers without one no longer fall back to the logger set via `setLogger`.
```

packages/mcp/src/extensions/logger.ts
```diff
@@ -21,12 +21,25 @@ export function setLogger(logger: LoggerFn | undefined): void {
...
       // never let logging blow up the tracking pipeline
     }
   }
 }
+
+/**
+ * Creates an isolated, safe logger for one instrumented server. In particular,
+ * an omitted logger stays a no-op rather than falling back to `activeLogger`.
+ */
+export function createLogger(logger: LoggerFn | undefined): LoggerFn {
```
turnipdabeets, line 40: [question] The cross-server leak (server A's `options.logger` becoming module state) is clearly a bug, but a loggerless server consulting an explicitly-set `setLogger` logger was the documented legacy configuration rather than leakage. Should `createLogger` fall back to `activeLogger` when the server has no logger of its own? That would fix the leak and keep the legacy debugging path — either outcome seems defensible, just checking the no-op is deliberate.

.changeset/calm-logs-stay.md
marandaneto: Good catch! Fixed in 80030f229 by preserving the legacy `setLogger` fallback for loggerless servers. The changeset can therefore remain focused on per-server isolation without documenting a behavior removal.

packages/mcp/src/extensions/logger.ts
marandaneto: Agreed — the no-op fallback was not necessary. Fixed in 80030f229: `createLogger` now keeps an explicitly supplied server logger isolated and otherwise delegates to the legacy global logger. The regression test verifies that the global logger receives loggerless-server messages without receiving another server’s logs.

packages/mcp/src/extensions/compatibility.ts
```diff
@@ -22,8 +22,8 @@ type ServerRecord = Record<string, unknown>
  */
 
 // Function to log compatibility information
-export function logCompatibilityWarning(): void {
-  log(
+export function logCompatibilityWarning(logger: LoggerFn = log): void {
```
ioannisj: [suggestion] Worth making `logger` required on these instead of defaulting to `log`? Every call site on the `instrument()` path already passes the server-bound logger explicitly, so the default never fires today anw. `captureEvent` already requires the param, so it'd be consistent with that.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -46,18 +47,19 @@ function isCallbackUpdate(value: unknown): value is { callback: ToolCallback } {
...
+  logger: LoggerFn
 ): Record<string, RegisteredTool> {
   return Object.fromEntries(
-    Object.entries(tools).map(([name, tool]) => [name, addTracingToToolCallbackInternal(tool, name, server)])
+    Object.entries(tools).map(([name, tool]) => [name, addTracingToToolCallbackInternal(tool, name, server, logger)])
   )
 }
 
-function setupListenerToRegisteredTools(server: HighLevelMCPServerLike): void {
+function setupListenerToRegisteredTools(server: HighLevelMCPServerLike, logger: LoggerFn): void {
```
ioannisj, line 66: [nit] Small one: this takes a `logger` param and also reads `data.logger` a few lines down for the same server.

ioannisj (approved):

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -141,20 +144,20 @@ function setupListenerToRegisteredTools(server: HighLevelMCPServerLike): void {
 function addTracingToToolCallbackInternal(
   tool: RegisteredTool,
   toolName: string,
-  server: HighLevelMCPServerLike
+  server: HighLevelMCPServerLike,
+  logger: LoggerFn
```
lucasheriques, line 161: when you rebase: #4355 added a 4th positional param to this same function on main (`getCurrentInputSchema: () => unknown`), and both sides pass a 4th arg at the same two call sites. the merged signature needs both, so five params.

worth doing by hand because `() => unknown` is assignable to `LoggerFn`. if the schema getter lands in the logger slot it compiles clean, logs go nowhere, and ownership falls back to the stale schema, undoing the #4355 fix. tsc catches the reverse but not that direction.

packages/mcp/src/index.ts
```diff
@@ -51,19 +51,18 @@ import type {
  * ```
  */
 function instrument(server: unknown, posthog: PostHog, options: MCPAnalyticsOptions = {}): McpAnalytics {
+  const logger = createLogger(options.logger)
```
lucasheriques: this is now outside the try/catch that used to wrap all of `instrument()`, so `instrument(server, ph, null)` from untyped js throws a TypeError into the caller instead of the old no-op. `options?.logger` covers it.

lucasheriques (approved):
isolation looks right. went looking for a missed call site and a residual cross-server leak and found neither, and tests 1 and 3 in `logger-isolation.test.ts` do fail on main, so they are real regression tests.

the rebase is the only thing blocking. `context-parameters.ts` and `conversation-id.ts` are mechanical: take main's collapsed guard and swap `log` for `logger`. `instrument-highlevel.ts` needs care, inline.

worth re-running `reserved-arguments.test.ts` and `error-capture.test.ts` after, since both came from #4355 and that is what you are merging against.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Fixed in f4a43bc9b. The listener now consistently uses its server-bound `logger` parameter for logging while retaining tracking data for the server metadata caches.

packages/mcp/src/extensions/compatibility.ts
marandaneto: Good suggestion. Fixed in 38cf7fc30: compatibility helpers now require an explicit logger, and direct test call sites pass one explicitly.

packages/mcp/src/extensions/instrument-highlevel.ts
marandaneto: Handled carefully in f4a43bc9b while merging current `main`. `addTracingToToolCallbackInternal` now takes both required parameters, with `getCurrentInputSchema` before `logger`, and every call site passes all five arguments. The reserved-arguments and error-capture regression tests pass.

packages/mcp/src/index.ts
marandaneto: Fixed in 38cf7fc30 with `options?.logger`, plus regression coverage confirming untyped JavaScript can pass `null` options without throwing.

PR #4360, opened 2026-07-31
fix(ai): preserve provider promise helpers

marandaneto:
## Problem

Instrumented provider requests could lose SDK-specific promise behavior, breaking callers that use raw responses or response metadata.

## Changes

- Preserve OpenAI, Azure, and Anthropic promise helpers while transforming monitored results.
- Keep lazy parsing and raw response access intact.
- Add regression coverage against the real provider SDK promise contracts.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


.changeset/calm-promises-wait.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+Preserve provider SDK promise helpers when instrumenting requests.
```
carlos-marchal-ph: This also widens the constructor options for `PostHogAnthropic` and `PostHogAzureOpenAI` to the full SDK `ClientOptions`, so callers can now pass `fetch`, `timeout`, `maxRetries` and friends. That is the user visible half of the change, worth a line here too.

packages/ai/src/anthropic/index.ts
```diff
@@ -330,7 +332,7 @@ export class WrappedMessages extends AnthropicOriginal.Messages {
         }
       ) as APIPromise<Message>
```
carlos-marchal-ph: This intermediate cast is redundant now: `preserveProviderPromise` takes a plain `Promise<Output>` and the result is cast again on the next line. It also mistypes the value, since the `.then()` result is not an `APIPromise`. The non-Azure OpenAI wrapper leaves these uncast.

packages/ai/src/openai/azure.ts
```diff
@@ -346,7 +349,7 @@ export class WrappedCompletions extends AzureOpenAI.Chat.Completions {
         }
       ) as APIPromise<ChatCompletion>
```
carlos-marchal-ph: Redundant cast, same at 552 and 685. `preserveProviderPromise` takes a plain `Promise<Output>` and the result is cast again on the next line.

packages/ai/src/openai/azure.ts
```diff
@@ -555,54 +562,63 @@ export class WrappedResponses extends AzureOpenAI.Responses {
     const { providerParams: openAIParams, posthogParams } = extractPosthogParams(body)
     const startTime = Date.now()
 
-    const parentPromise = super.parse(openAIParams, options)
+    const originalCreate = super.create.bind(this)
+    const originalSelfRecord = this as Record<string, unknown>
+    const tempCreate = originalSelfRecord['create']
+    originalSelfRecord['create'] = originalCreate
```
carlos-marchal-ph: This block is copied verbatim from `packages/ai/src/openai/index.ts` and neither copy says why it exists. It took me a while to work out that it is because `Responses.parse` calls `this._client.responses.create`, which resolves back to the wrapped `create` and would capture twice. Since you are already extracting a shared helper for the promise side, could this move next to it? The OpenAI copy at least keeps the `// Restore our wrapped create method` comment, this one dropped it.

packages/ai/src/providerPromise.ts
```diff
@@ -0,0 +1,42 @@
...
+/**
+ * Keep the provider SDK helpers on a promise whose resolved value is instrumented.
+ * OpenAI's parse helpers compose create calls through `_thenUnwrap`, while both
+ * OpenAI and Anthropic expose the raw response through `asResponse` and
+ * `withResponse`.
+ */
+export function preserveProviderPromise<Input, Output>(
+  parentPromise: ProviderPromise<Input>,
+  wrappedPromise: Promise<Output>
+): ProviderPromise<Output> {
```
carlos-marchal-ph: Returning `ProviderPromise<Output>` rather than `APIPromise<Output>` pushes an `as APIPromise<...>` onto all 14 call sites, where the old helper kept the single unavoidable cast inside itself. Could the return type be parameterised so the cast stays here?

packages/ai/src/providerPromise.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (withResponse) {
+    providerPromise.withResponse = async () => {
+      const [response, data] = await Promise.all([withResponse(), wrappedPromise])
+      return { ...response, data }
+    }
+  }
+
+  if (typeof parentPromise._thenUnwrap === 'function') {
+    providerPromise._thenUnwrap = <Unwrapped>(transform: (data: Output) => Unwrapped) =>
+      preserveProviderPromise<Output, Unwrapped>(providerPromise, wrappedPromise.then(transform))
```
carlos-marchal-ph: The real `_thenUnwrap` wraps its result in `addRequestID(transform(...), props.response)` and this shim does not. `defaultParseResponse` attaches `_request_id` as a non-enumerable property, and `parseChatCompletion` returns `{ ...completion, choices }`, so the spread loses it and the SDK re-attaches it after the transform. `chat.completions.parse` is not overridden in either wrapper, so it routes through `create` and lands on this shim, which means parsed completions come back without `_request_id` on both the OpenAI and Azure clients. Our own `extractRequestId` reads exactly that property, and it is not in the SDK's public types so nothing will flag it.

Separately, `transform` is typed `(data: T) => Result` while the SDK's contract is `(data, props)`, and `.then()` only ever passes one argument, so `props` would always be `undefined`. Nothing in the SDK uses it today.

packages/ai/tests/provider-promises.test.ts
```diff
@@ -0,0 +1,217 @@
...
+      messages: [{ role: 'user', content: 'Hello' }],
+      max_tokens: 32,
+    })
+
+    expect(typeof promise.asResponse).toBe('function')
+    expect(typeof promise.withResponse).toBe('function')
+    const rawResponse = await promise.asResponse()
+    const { data, request_id } = await promise.withResponse()
+
+    expect(rawResponse.status).toBe(200)
```
carlos-marchal-ph, line 222: `asResponse()` is only checked for its status. The wrapper attaches `.then()` at `create` time, so the SDK's `parse()` has already called `response.json()` on this same instance, and I would like to know what a caller can actually do with the response they get back. Worth asserting either way. Nothing here covers `_request_id` surviving on a parsed result either, which is what hides the `_thenUnwrap` gap.

ioannisj (approved):

carlos-marchal-ph (approved):
Approving to unblock but I think some of the comments are worth addressing!

.changeset/calm-promises-wait.md
marandaneto: Good catch! Fixed in ba0ace4f2 — the changeset now also documents support for the full Anthropic and Azure OpenAI client options.

packages/ai/src/anthropic/index.ts
marandaneto: Fixed in ba0ace4f2 — removed the redundant intermediate Anthropic APIPromise cast.

packages/ai/src/openai/azure.ts
marandaneto: Fixed in ba0ace4f2 — removed the redundant Azure casts at all noted call sites.

packages/ai/src/openai/azure.ts
marandaneto: Fixed in ba0ace4f2 — extracted the temporary original-create swap into a documented shared helper used by both OpenAI and Azure, including exact property restoration.

packages/ai/src/providerPromise.ts
marandaneto: Fixed in ba0ace4f2 — provider-specific overloads now keep the unavoidable facade cast inside the shared helper, so the call-site result casts are gone.

packages/ai/src/providerPromise.ts
marandaneto: Fixed in ba0ace4f2 — `_thenUnwrap` now receives the complete provider response metadata, preserves non-enumerable `_request_id`, and composes without parsing the response body twice.

packages/ai/tests/provider-promises.test.ts
marandaneto: Fixed in ba0ace4f2 — tests now assert parsed `_request_id` preservation and non-enumerability, `_thenUnwrap` response props, raw response headers/body state, and create-method restoration.

PR #4361, opened 2026-07-31
fix(ai): preserve stream cancellation semantics

marandaneto:
## Problem

Eager telemetry stream consumption could defeat backpressure and continue provider work after the caller cancelled.

## Changes

- Add a synchronized monitoring tee that reads only when the caller reads.
- Forward iterator return, abort, and stream errors to the provider source.
- Cover cancellation, concurrent reads, empty streams, and provider failures.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


packages/ai/src/stream.ts
```diff
@@ -0,0 +1,233 @@
...
+          settleMonitorTerminal()
+          settleCallersTerminal()
+        } else if (monitorPending) {
+          monitorPending.resolve(result)
+          monitorPending = undefined
+          cancellationPromise = undefined
+        } else {
+          bufferedMonitorResult = result
+          cancellationPromise = undefined
+        }
```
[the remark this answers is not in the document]
marandaneto: I traced this through the helper invariants and don’t believe the proposed `pump()` calls can unblock a reachable state. If `callerQueue` and `monitorPending` are both populated, `pump()` starts synchronously and clears `monitorPending`; if the monitor is between chunks, its next `next()` call invokes `pump()`; and after clearing/buffering a non-terminal return result, an immediate `pump()` is intentionally gated until the monitor requests that result to preserve lockstep backpressure. The suggested calls are therefore no-ops in the cited branches. Autoreview did uncover a related real cancellation gap: `return()` could wait behind an in-flight source `next()` without aborting the provider. Fixed in 7e090ff7c by aborting the shared controller before awaiting source `return()`, with a regression test for that exact race.

packages/ai/src/stream.ts
```diff
@@ -0,0 +1,238 @@
...
+      if (hasTerminalError) {
+        pending.reject(terminalError)
+      } else if (terminalResult) {
+        pending.resolve(terminalResult)
+      }
+    }
+  }
+
+  const pump = (): void => {
+    if (operationInFlight || callerQueue.length === 0 || (monitorActive && !monitorPending)) {
```
ioannisj, line 72: Does the monitor have any way to finish on its own here? `pump()` bails when `callerQueue.length === 0`, and the monitoring iterator's `next()` only registers `monitorPending` and calls `pump()`, so nothing advances the source except a caller read.

If a caller stops reading without cancelling (drops the iterator, or the request unwinds without a `break` and without an `abort()`), `monitorPending` never settles, the `for await` in the wrapper never finishes, and we never reach `captureAiGeneration`. So no `$ai_generation` event at all, and the accumulated chunks stay retained.

`break`/`return()` and `controller.abort()` are both handled and tested, this is the third path. I think `tee()` used to capture here since it pumped both branches eagerly, but I haven't verified that. Worth giving the monitor a way to terminate on its own, either flushing on caller finalization or settling terminally after an idle timeout and capturing a partial generation? Both keep the no-read-ahead property.

ioannisj (approved):
LG, one thing flagged which looks like could lead to dropped events?

packages/ai/src/stream.ts
marandaneto: You’re right that the monitor cannot finish if a consumer abandons the iterator without calling `return()` or aborting its controller. That abandonment is not observable through the async-iterator protocol: finalization would be nondeterministic, while an idle timeout could terminate a valid consumer that pauses between reads. The previous `tee()` happened to keep draining the provider eagerly, but that also meant an abandoned consumer could continue consuming and retaining the entire response. This helper deliberately requires explicit cancellation (`break`/`return()` or `abort()`) so it can preserve provider backpressure and capture the partial generation deterministically. I don’t think we can safely add monitor-initiated termination without reintroducing eager reads or an arbitrary timeout.

packages/ai/src/openai/azure.ts
```diff
@@ -439,11 +449,17 @@ export class WrappedResponses extends AzureOpenAI.Responses {
                   finalContent = chunk.response.output
                 }
                 if ('usage' in chunk && chunk.usage) {
+                  const chunkUsage = chunk.usage as {
```
carlos-marchal-ph: I this branch cannot fire. No member of `ResponseStreamEvent` declares a top-level `usage`; it lives at `chunk.response.usage`, which is what `openai/index.ts` reads in the equivalent loop. The chunk used to be `any` so this compiled silently, and now that it is typed the cast is the only thing keeping it alive

packages/ai/src/stream.ts
carlos-marchal-ph: Makes sense to me, but could we pin this in a test somewhere as deliberate?

carlos-marchal-ph (approved):

packages/ai/src/openai/azure.ts
marandaneto: Good catch — fixed in 15038f4ab by reading usage from `chunk.response.usage`, matching the Responses event shape and the non-Azure wrapper. I also added a streaming regression test covering input, output, reasoning, and cached token capture.

packages/ai/src/stream.ts
marandaneto: Added in 15038f4ab. The new regression test pins that an abandoned caller leaves the monitor pending without reading ahead; it then explicitly cancels during cleanup so the test does not leak pending work.

PR #4362, opened 2026-07-31
fix(ai): isolate provider calls from telemetry failures

marandaneto:
## Problem

PostHog capture failures could reject successful provider calls or replace the provider's original error.

## Changes

- Make telemetry capture best-effort across supported AI integrations.
- Preserve successful provider results and original provider errors.
- Add failure-isolation regression coverage for each affected integration.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


packages/ai/tests/captureAiGeneration.test.ts
```diff
@@ -98,6 +98,27 @@ describe('captureAiGeneration', () => {
     expect(client.captureImmediate).toHaveBeenCalledTimes(1)
   })
 
+  it('swallows synchronous capture delivery failures', async () => {
```
carlos-marchal-ph, line 105: Both of these fail delivery, which is the part now guarded. The two paths that still escape aren't covered: a client whose `captureException` throws while `enableExceptionAutocapture` is set, and an `input` or `output` that `JSON.stringify` rejects. If the `try` moves up to cover the whole body, those are the cases that pin it.

ioannisj (approved):

packages/ai/src/langchain/callbacks.ts
```diff
@@ -429,12 +429,16 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
+    try {
+      this.client.capture({
+        distinctId: this.distinctId ? this.distinctId.toString() : runId,
+        event: eventName,
+        properties: eventProperties,
+        groups: this.groups,
+      })
+    } catch {
+      // Telemetry delivery must never affect the LangChain callback lifecycle.
+    }
```
carlos-marchal-ph: This block and the one at line 552 are byte-identical, comment included. Would a small private helper (`_safeCapture(message)`) be better, so the "never break the callback lifecycle" rule lives in one place and it's obvious both capture paths are covered?

packages/ai/src/captureAiGeneration.ts
```diff
@@ -190,9 +190,13 @@ export const captureAiGeneration = async (client: PostHog, options: CaptureAiGen
...
-  } else {
-    client.capture(event)
+  try {
+    if (options.captureImmediate) {
+      await client.captureImmediate(event)
+    } else {
+      client.capture(event)
+    }
+  } catch {
+    // Telemetry delivery must never affect the instrumented provider call.
```
carlos-marchal-ph: The guard starts below the event construction, and everything above it is telemetry work that runs on the provider's happy path and can throw. `sanitizeValues` on line 107 is `JSON.parse(JSON.stringify(obj))`, which throws on circular structures and on `BigInt`, and when multimodal capture is enabled the per-provider redactor returns its input untouched so cycles aren't necessarily broken before this point. `client.captureException` on line 124 is also a call on the same client whose failures we're isolating, outside the guard.

Could the `try` wrap the whole function body instead? That is what `_captureEvent` in `openai-agents/processor.ts` already does, build and send inside one `try`.

carlos-marchal-ph: Nothing is logged, so a user whose events stop arriving has no signal at all. Would a `console.warn` save us some pain debugging in the futrue?

carlos-marchal-ph (approved):

packages/ai/tests/captureAiGeneration.test.ts
marandaneto: Good catch! Fixed in 1e82bcf74 — added regression coverage for throwing exception autocapture and circular input/output serialization failures.

packages/ai/src/captureAiGeneration.ts
marandaneto: Fixed in 1e82bcf74 — the guard now wraps the full `captureAiGeneration` body, including sanitization, event construction, and exception autocapture.

marandaneto: Fixed in 1e82bcf74 — telemetry failures now emit a `console.warn` with the original error while remaining non-throwing.

packages/ai/src/langchain/callbacks.ts
marandaneto: Fixed in 1e82bcf74 — extracted the duplicated LangChain capture guards into a private `_safeCapture` helper.

PR #4363, opened 2026-07-31
fix(ai): safely capture unknown telemetry values

marandaneto:
## Problem

Unknown generation values could throw during telemetry serialization or recurse indefinitely on circular and hostile objects.

## Changes

- Add bounded, cycle-safe conversion for unknown telemetry values.
- Handle BigInt, functions, hostile toJSON implementations, and invalid Date values safely.
- Keep privacy-mode redaction ahead of value inspection and add regression tests.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


marandaneto:
### Scope note

This PR intentionally applies bounded, JSON-safe normalization only to `$ai_input` and `$ai_output_choices`. Those are the arbitrary caller/provider content fields involved in this issue, and keeping the change there avoids changing serialization semantics for other event properties.

Other caller-owned fields—such as model parameters, raw usage, tools, provider metadata, and custom properties—may warrant similar hardening in a follow-up if we want a broader event-wide guarantee.

packages/ai/src/captureAiGeneration.ts
```diff
@@ -171,6 +173,8 @@ export const captureAiGeneration = async (client: PostHog, options: CaptureAiGen
     $ai_trace_id: traceId,
     $ai_base_url: options.baseURL ?? '',
     ...options.properties,
+    $ai_input: safeInput,
```
ioannisj: Should `$ai_input` and `$ai_output_choices` move back above `...options.properties`? They sit below the spread now, so a caller passing their own `$ai_input` through `properties` silently gets the converter's output instead, and that worked before the reorder.

ioannisj (approved):

packages/ai/src/captureAiGeneration.ts
marandaneto: Good catch! Fixed in 807888218 — restored the previous property spread precedence so custom `$ai_input` and `$ai_output_choices` values continue to override the generated values, with a regression test covering both fields.

packages/core/src/utils/json-utils.ts
```diff
@@ -0,0 +1,133 @@
...
+          for (; index < itemCount && state.remainingNodes > 0; index++) {
+            output.push(convert(current[index], depth + 1))
+          }
+          if (current.length > index) {
+            output.push(TRUNCATED_VALUE)
+          }
+          return output
+        }
+
+        const output: Record<string, unknown> = {}
```
Radu-Raicea, line 114: This turns some normally serializable objects into `{}`. Should we try their `toJSON()` method first and use this fallback if it throws?

Radu-Raicea (approved):

packages/core/src/utils/json-utils.ts
marandaneto: Good point — fixed in 4e3f8db94. The converter now tries non-Date `toJSON()` methods, sanitizes their returned values through the existing bounded conversion, and falls back to structural traversal if `toJSON()` throws. I also added coverage for successful, throwing, and self-referential results.

PR #4364, opened 2026-07-31
fix(ai): redact binary content across integrations

marandaneto:
## Problem

Binary and data-URL content could escape sanitization in several provider-specific input shapes.

## Changes

- Apply binary redaction consistently to OpenAI, Azure, Gemini, and Vercel payloads.
- Handle URL objects and media-type context without exposing embedded data.
- Add provider-specific regression coverage for the affected shapes.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


packages/ai/src/sanitization/binary_content_redactor.ts
```diff
@@ -42,7 +42,11 @@ export class BinaryContentRedactor {
   }
 
   private redactString(value: string, ctx: MediaTypeContext): string {
-    const minLength = ctx.signalsBinary() ? STRONG_CONTEXT_MIN_LENGTH : WEAK_CONTEXT_MIN_LENGTH
+    const minLength = ctx.hasExplicitMediaType()
+      ? 1
+      : ctx.signalsBinary()
+        ? STRONG_CONTEXT_MIN_LENGTH
+        : WEAK_CONTEXT_MIN_LENGTH
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 987d7aebf — explicit text media types now keep the normal recognition threshold, while short and line-wrapped binary MIME payloads remain redacted. Added regression coverage for both cases.

packages/ai/src/vercel/middleware.ts
```diff
@@ -113,10 +113,13 @@ const mapVercelPrompt = (messages: LanguageModelPrompt): PostHogInput[] => {
...
             if (contentData instanceof URL) {
-              fileData = contentData.toString()
+              fileData = redactBase64DataUrl(
+                contentData.toString(),
+                contentData.protocol === 'data:' ? c.mediaType : undefined
+              )
             } else if (isString(contentData)) {
               // Redact base64 data URLs and raw base64 to prevent oversized events
-              fileData = redactBase64DataUrl(contentData)
+              fileData = redactBase64DataUrl(contentData, c.mediaType)
```
ioannisj: I think the `tool-result` branch at L149 needs the same treatment this one just got. `output: c.output` goes through untouched, and nothing downstream catches it (`sanitizeValues` in `packages/ai/src/utils.ts` is a unicode round-trip, not a redactor)

ioannisj (approved):

packages/ai/src/vercel/middleware.ts
marandaneto: Good catch! Fixed in c66f0ba8b — Vercel tool-result outputs now go through recursive binary redaction before capture, while the provider payload remains unchanged. Added an integration regression test for nested media output.

packages/ai/src/sanitization.ts
```diff
@@ -2,14 +2,16 @@ import { BinaryContentRedactor } from './sanitization/binary_content_redactor'
...
 
-export function redactBase64DataUrl(str: string): string
-export function redactBase64DataUrl(str: unknown): unknown
-export function redactBase64DataUrl(str: unknown): unknown {
-  return redactor.redact(str)
+export function redactBase64DataUrl(str: string, mediaType?: string): string
+export function redactBase64DataUrl(str: unknown, mediaType?: string): unknown
+export function redactBase64DataUrl(str: unknown, mediaType?: string): unknown {
+  if (!mediaType) return redactor.redact(str)
+  return (redactor.redact({ data: str, mediaType }) as { data: unknown }).data
```
carlos-marchal-ph: The `as { data: unknown }` cast is not needed, `redact<T>(value: T): T` already hands back `{ data: unknown; mediaType: string }`.

More broadly, is fabricating a wrapper object purely so `MediaTypeContext` can find a sibling MIME the interface we want? Threading an optional media type into the walk would avoid the round trip.

packages/ai/src/sanitization/media_type_context.ts
```diff
@@ -89,6 +89,14 @@ export class MediaTypeContext {
     return undefined
   }
 
+  hasExplicitBinaryMediaType(): boolean {
+    if (!this.parent || !this.key || !STRONG_CONTEXT_KEYS.has(this.key)) return false
+    return MIME_HINT_KEYS.some((hint) => {
```
carlos-marchal-ph: Could this reuse `inferFromSiblingMime` instead of iterating `MIME_HINT_KEYS` again? As written the decision can come from one hint while `placeholderFor` takes the label from another, so `{ data, mediaType: 'text/plain', mimeType: 'image/png' }` redacts and then labels the result `text/plain`.

packages/ai/src/vercel/middleware.ts
```diff
@@ -113,10 +113,13 @@ const mapVercelPrompt = (messages: LanguageModelPrompt): PostHogInput[] => {
             const contentData: unknown = c.data
 
             if (contentData instanceof URL) {
-              fileData = contentData.toString()
+              fileData = redactBase64DataUrl(
```
carlos-marchal-ph: These four lines are now identical to the ones in `mapVercelOutput` (line 233), and the string branch matches too. Worth pulling out a small `redactFileData(data, mediaType)` helper so the two mappings cannot drift?

packages/ai/tests/binary_content_redactor.test.ts
```diff
@@ -62,8 +62,24 @@ describe('redactBinaryContent', () => {
...
+    it.each(['mediaType', 'media_type', 'mimeType', 'mime_type'])(
+      'redacts short base64 when sibling %s provides an explicit MIME',
+      (key) => {
+        expect((redactBinaryContent({ data: SHORT_B64, [key]: 'image/png' }) as any).data).toBe(
+          placeholder('image/png')
+        )
+      }
+    )
+
+    it('redacts line-wrapped base64 when an explicit binary MIME is present', () => {
```
carlos-marchal-ph, line 74: This passes on the strength of the first character alone, so it would go green for `Uhello world` too and does not really pin the line-wrapping behaviour. Same for the `SHORT_B64` cases above it. A value whose first character is base64 but which clearly is not base64 overall would say more about the intent.

packages/ai/src/openai/azure.ts
```diff
@@ -454,8 +454,10 @@ export class WrappedResponses extends AzureOpenAI.Responses {
                 ...posthogParams,
                 model: openAIParams.model ?? modelFromResponse,
                 provider: 'azure',
-                input: formatOpenAIResponsesInput(openAIParams.input, openAIParams.instructions),
-                output: finalContent,
+                input: sanitizeOpenAI(
```
carlos-marchal-ph: Azure now runs the redactor twice over Responses input, once inside `formatOpenAIResponsesInput` and once around it, while `openai/index.ts` only does the inner pass. `sanitizeOpenAI` and `sanitizeOpenAIResponse` are the same function.

carlos-marchal-ph (approved):

packages/ai/src/sanitization/media_type_context.ts
marandaneto: Addressed in 0264004e1. `hasExplicitBinaryMediaType` now reuses `inferFromSiblingMime`, so recognition and placeholder labeling consistently use the same MIME hint. Added regression coverage for conflicting sibling hints.

packages/ai/src/vercel/middleware.ts
marandaneto: Addressed in 0264004e1. Extracted a shared `redactFileData(data, mediaType)` helper and use it in both Vercel input and output file mappings.

packages/ai/src/sanitization.ts
marandaneto: Addressed in 0264004e1. `BinaryContentRedactor.redact` now accepts the optional media type directly and seeds the walk context without fabricating a wrapper object or requiring a cast.

packages/ai/tests/binary_content_redactor.test.ts
marandaneto: Addressed in 0264004e1. Explicit binary MIME content now validates up to the full first 64 normalized characters rather than only the first character, while still supporting CR/LF-wrapped base64. Added a negative `Uhello world` regression case.

packages/ai/src/openai/azure.ts
marandaneto: Addressed in 0264004e1. Removed the redundant outer `sanitizeOpenAI` calls from all Azure Responses input capture paths, matching the OpenAI implementation.

PR #4365, opened 2026-07-31
fix(ai): track background responses

marandaneto:
## Problem

OpenAI and Azure background Responses were not captured when completion happened through later retrieval, streaming retrieval, or cancellation.

## Changes

- Retain bounded monitoring context for pending background responses.
- Capture terminal polling, streaming, and cancellation results exactly once without delaying provider delivery.
- Preserve lazy APIPromise and raw response behavior with lifecycle regression tests.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


ioannisj (approved):

packages/ai/src/openai/index.ts
```diff
@@ -450,13 +454,44 @@ export class WrappedCompletions extends Completions {
...
+    await captureAiGenerationAfterSuccess(this.phClient, {
+      ...posthogParams,
+      model: openAIParams.model ?? result.model,
+      provider: 'openai',
+      input: formatOpenAIResponsesInput(sanitizeOpenAIResponse(openAIParams.input), openAIParams.instructions),
+      output: formatResponseOpenAI({ output: result.output }),
+      latency: (Date.now() - startTime) / 1000,
+      baseURL: this.baseURL,
+      modelParameters: getModelParams(openAIParams, result.service_tier),
+      httpStatus: 200,
```
carlos-marchal-ph, line 465: `httpStatus` is 200 and no `error` is passed, so `$ai_is_error` is never set and `result.error` / `result.incomplete_details` are dropped. A background job that fails comes back as a 200 retrieve with `status: 'failed'` and a populated `error`, and we would record it as a successful generation with `$ai_stop_reason: 'failed'`. Same shape in `azure.ts`, and there is no test for a `failed` or `incomplete` terminal status.

packages/ai/src/openai/index.ts
```diff
@@ -450,13 +454,44 @@ export class WrappedCompletions extends Completions {
...
+    context: BackgroundResponseState
+  ): Promise<void> {
+    const { openAIParams, posthogParams, startTime } = context
+    await captureAiGenerationAfterSuccess(this.phClient, {
+      ...posthogParams,
+      model: openAIParams.model ?? result.model,
+      provider: 'openai',
+      input: formatOpenAIResponsesInput(sanitizeOpenAIResponse(openAIParams.input), openAIParams.instructions),
+      output: formatResponseOpenAI({ output: result.output }),
+      latency: (Date.now() - startTime) / 1000,
```
carlos-marchal-ph: `startTime` is from create, so `$ai_latency` here is how long the app took to notice the response had finished rather than how long generation took. A worker polling every 30 seconds reports 30 seconds. Is there something better to derive it from, or would omitting it on this path beat reporting a number that tracks the polling cadence?

packages/ai/src/openai/index.ts
```diff
@@ -687,6 +723,81 @@ export class WrappedResponses extends Responses {
...
+      }
+
+      // Removing the context before capture makes concurrent or repeated
+      // terminal polls idempotent.
+      const context = this.backgroundResponses.take(responseID)
+      if (!context) {
+        return result
+      }
+
+      void this.captureBackgroundResponse(result, context).catch(() => undefined)
```
carlos-marchal-ph: On `create` the capture is awaited before the caller's promise resolves, so `posthogCaptureImmediate: true` means the event is flushed by the time it returns. Here it is fired and forgotten, so a serverless handler that awaits `retrieve` or `cancel` and then returns can be frozen before the event leaves. Non-blocking is clearly right for the stream path, but for non-streaming `retrieve` and `cancel` it quietly weakens the flag.

packages/ai/src/openai/index.ts
```diff
@@ -625,6 +660,7 @@ export class WrappedResponses extends Responses {
         async (result) => {
           if ('output' in result) {
             if (isPendingBackgroundResponse(openAIParams, result)) {
+              this.backgroundResponses.set(result.id, { openAIParams, posthogParams, startTime })
```
carlos-marchal-ph: Only this branch and `parse` register a context. The streaming create branch at line 523 captures when the stream reaches `response.completed`, so an interrupted background stream captures nothing and the follow-up streaming retrieve has no context to capture with either. Resuming an interrupted stream with `starting_after` is much of the point of background streaming, so is that gap intended for this PR?

packages/ai/src/openai/index.ts
```diff
@@ -687,6 +723,81 @@ export class WrappedResponses extends Responses {
...
+  ): APIPromise<OpenAIOrignal.Responses.Response | Stream<OpenAIOrignal.Responses.ResponseStreamEvent>>
+
+  public retrieve(
+    responseID: string,
+    query: ResponseRetrieveParamsBase = {},
+    options?: RequestOptions
+  ): APIPromise<OpenAIOrignal.Responses.Response | Stream<OpenAIOrignal.Responses.ResponseStreamEvent>> {
+    const parentPromise = super.retrieve(responseID, query, options)
+
+    if (query.stream) {
```
carlos-marchal-ph, line 777: `cancel` returns the upstream `APIPromise` untouched when the response was not created through this client, but `retrieve` always goes through `_thenUnwrap`, and every streaming retrieve gets re-wrapped in a fresh `Stream` whether or not a context exists. The re-wrap also drops the SDK's internal client reference that `Stream.tee()` propagates. Would the same `get` guard work at the top of `retrieve`?

packages/ai/src/openai/azure.ts
```diff
@@ -354,13 +372,42 @@ export class WrappedCompletions extends AzureOpenAI.Chat.Completions {
...
+      modelParameters: getModelParams(openAIParams, result.service_tier),
+      httpStatus: 200,
+      usage: {
+        inputTokens: result.usage?.input_tokens ?? 0,
+        outputTokens: result.usage?.output_tokens ?? 0,
+        reasoningTokens: result.usage?.output_tokens_details?.reasoning_tokens ?? 0,
+        cacheReadInputTokens: result.usage?.input_tokens_details?.cached_tokens ?? 0,
+        rawUsage: result.usage,
+      },
+      stopReason: result.status ?? undefined,
```
carlos-marchal-ph, line 417: `captureBackgroundResponse` sends `rawUsage` and `stopReason`, but Azure's `create` (line 551) and `parse` (line 695) do not, so background generations will carry `$ai_usage` and `$ai_stop_reason` while the synchronous ones will not. Should `create` and `parse` be brought in line?

carlos-marchal-ph (approved):

packages/ai/src/openai/index.ts
marandaneto: Fixed in 4fc4dea78 — background streaming creates now register context as soon as a response ID appears and retain it when the stream ends or is interrupted without a terminal event. Added an interrupted-create/starting_after resume test for both providers.

packages/ai/src/openai/index.ts
marandaneto: Good catch! Fixed in 4fc4dea78 — failed terminal responses now populate $ai_is_error/$ai_error, incomplete details are retained in provider metadata, and both OpenAI and Azure have failed/incomplete background tests.

packages/ai/src/openai/index.ts
marandaneto: Fixed in 4fc4dea78 — non-streaming retrieve and cancel now wait for terminal capture (including posthogCaptureImmediate) before resolving, while streaming capture remains non-blocking. Added deferred-flush and APIPromise/request-ID coverage.

packages/ai/src/openai/index.ts
marandaneto: Fixed in 4fc4dea78 — retrieve now returns the upstream APIPromise unchanged for untracked IDs, including streaming retrieves, so those streams are not re-wrapped. Added exact promise-identity coverage.

packages/ai/src/openai/index.ts
marandaneto: Fixed in 4fc4dea78 — background latency now comes from the provider’s completed_at - created_at timestamps. When no terminal timestamp is available, latency is omitted rather than reporting polling delay.

packages/ai/src/openai/azure.ts
marandaneto: Fixed in 4fc4dea78 — Azure Responses create and parse now include raw usage and stop reason, with assertions covering both synchronous paths.

PR #4366, opened 2026-07-31
fix(ai): capture Responses API failures

marandaneto:
## Problem

Failed, incomplete, and cancelled Responses API results could be treated as successful generations or lose their terminal status.

## Changes

- Classify terminal Responses results and stream events consistently for OpenAI and Azure.
- Capture provider error details and stop reasons without changing provider behavior.
- Add non-streaming and streaming regression coverage.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context


packages/ai/src/openai/utils.ts
```diff
@@ -31,19 +31,39 @@ export function extractRequestId(result: unknown): string | undefined {
...
+}
+
+export function getResponseFailure(
+  response: Pick<OpenAI.Responses.Response, 'id' | 'status' | 'error'> | null | undefined
+): unknown | undefined {
+  if (response?.status !== 'failed') {
+    return undefined
+  }
+
+  return response.error ?? new Error(`OpenAI response ${response.id} failed without error details`)
```
ioannisj: Worth returning a shallow copy of `response.error` here rather than the live reference? I think we are mutating this down the path

ioannisj (approved):

packages/ai/src/openai/utils.ts
marandaneto: Good catch! Fixed in 562a1dd68 — `getResponseFailure` now returns a shallow copy of the provider error so downstream exception capture cannot mutate the original response. Added regression coverage for the reference isolation.

packages/ai/src/openai/azure.ts
```diff
@@ -429,14 +436,18 @@ export class WrappedResponses extends AzureOpenAI.Responses {
...
-                  chunk.type === 'response.completed' &&
-                  'response' in chunk &&
-                  chunk.response?.output &&
-                  chunk.response.output.length > 0
-                ) {
-                  finalContent = chunk.response.output
+                  if (isTerminalResponse(chunk.response)) {
+                    terminalResponse = chunk.response
+                    finalContent = chunk.response.output
+                    if (chunk.response.status !== 'completed' && chunk.response.usage) {
```
carlos-marchal-ph: Why gate this on the status not being `completed`? The fallback just below reads a top-level `chunk.usage` rather than `chunk.response.usage`, so completed Azure streams never pick usage up at all, and the new streaming test locks that in with `status === 'completed' ? 0 : 11`. Reading `chunk.response.usage` for every chunk, the way `openai/index.ts` does, would fix completed streams too.

packages/ai/src/openai/azure.ts
```diff
@@ -462,7 +473,13 @@ export class WrappedResponses extends AzureOpenAI.Responses {
                 modelParameters: getModelParams(body, serviceTierFromResponse),
                 httpStatus: 200,
                 usage,
+                stopReason:
+                  terminalResponse?.status === 'completed' ? undefined : (terminalResponse?.status ?? undefined),
```
carlos-marchal-ph: This drops `$ai_stop_reason` for completed responses, but the OpenAI wrapper passes `result.status` straight through, so the same response yields `completed` there and nothing here. Anthropic and Gemini also pass the provider value through verbatim. Given the aim is consistent classification across both, could we keep the raw status in Azure too? The same ternary also appears three times in this file, so if the suppression is deliberate it would read better as a named helper alongside `isTerminalResponse`.

packages/ai/src/openai/index.ts
```diff
@@ -534,16 +534,10 @@ export class WrappedResponses extends Responses {
...
-                  chunk.type === 'response.completed' &&
-                  'response' in chunk &&
-                  chunk.response?.output &&
-                  chunk.response.output.length > 0
-                ) {
-                  finalContent = chunk.response.output
-                  if (chunk.response.status) {
+                  if (isTerminalResponse(chunk.response)) {
+                    terminalResponse = chunk.response
+                    finalContent = chunk.response.output
```
carlos-marchal-ph: Dropping the old emptiness guard is what makes failed and incomplete responses report their status, so that part is right, but a terminal chunk without `output` now assigns `undefined` and `$ai_output_choices` goes out as `undefined` rather than `[]`. Can we use `chunk.response.output ?? []`?

packages/ai/src/openai/index.ts
```diff
@@ -652,7 +650,11 @@ export class WrappedResponses extends Responses {
               stopReason: result.status ?? undefined,
               tools: availableTools,
               completionId: result.id,
-              providerMetadata: buildProviderMetadata({ requestId: extractRequestId(result) }),
+              providerMetadata: buildProviderMetadata({
+                requestId: extractRequestId(result),
+                incompleteDetails: result.incomplete_details,
+              }),
+              error: getResponseFailure(result),
```
carlos-marchal-ph, line 644: `httpStatus: 200` is passed a few lines up, so a failed response goes out with `$ai_is_error: true` and `$ai_http_status: 200`. The HTTP call really was a 200, so this may be exactly what we want, but anything keying off `$ai_http_status` will still read these as successful. Same shape in `azure.ts`.

packages/ai/src/openai/utils.ts
```diff
@@ -31,19 +31,41 @@ export function extractRequestId(result: unknown): string | undefined {
...
- * truthy value are included, and `undefined` is returned when there is nothing
+ * meaningful value are included, and `undefined` is returned when there is nothing
  * to report so the property can be omitted from the event entirely.
  */
 export function buildProviderMetadata(fields: {
   systemFingerprint?: string | null
   requestId?: string | null
-}): Record<string, string> | undefined {
-  const metadata: Record<string, string> = {}
+  incompleteDetails?: unknown
```
carlos-marchal-ph: Could this be `OpenAI.Responses.Response['incomplete_details']`? The neighbouring fields are typed precisely, and it would document what ends up under `$ai_provider_metadata.incomplete_details`.

packages/ai/src/openai/utils.ts
```diff
@@ -31,19 +31,41 @@ export function extractRequestId(result: unknown): string | undefined {
...
+
+const TERMINAL_RESPONSE_STATUSES = new Set(['completed', 'failed', 'cancelled', 'incomplete'])
+
+export function isTerminalResponse(response: { status?: string | null } | null | undefined): boolean {
+  return !!response?.status && TERMINAL_RESPONSE_STATUSES.has(response.status)
+}
+
+export function getResponseFailure(
+  response: Pick<OpenAI.Responses.Response, 'id' | 'status' | 'error'> | null | undefined
+): unknown | undefined {
```
carlos-marchal-ph: `unknown | undefined` collapses to `unknown`, so the union is not saying anything. Separately, the other exports in this file each carry a doc comment explaining the property they feed, and these two are the only ones without.

packages/ai/tests/openai.test.ts
```diff
@@ -1000,6 +1000,118 @@ describe('PostHogOpenAI - Jest test suite', () => {
...
+        output_tokens: 7,
+        total_tokens: 18,
+        input_tokens_details: { cached_tokens: 2 },
+        output_tokens_details: { reasoning_tokens: 3 },
+      },
+      error: status === 'failed' ? { code: 'server_error', message: 'provider response failed' } : null,
+      incomplete_details: status === 'incomplete' || status === 'cancelled' ? { reason: 'max_output_tokens' } : null,
+    })
+
+    test.each(terminalStatuses)('non-streaming %s response preserves terminal data', async (status) => {
```
carlos-marchal-ph, line 1057: `parse()` gained the same `error` and `incomplete_details` handling in both wrappers, but these cases only drive `create()`. Could we add one failed-status `parse()` case?

carlos-marchal-ph (approved):

packages/ai/src/openai/index.ts
marandaneto: Fixed in 86d056c38 — terminal stream output now falls back to [] in both OpenAI and Azure wrappers, with regression coverage for a terminal response that omits output.

packages/ai/src/openai/azure.ts
marandaneto: Fixed in 86d056c38 — Azure now preserves the raw response status as $ai_stop_reason consistently for streaming, create(), and parse() paths.

packages/ai/tests/openai.test.ts
marandaneto: Fixed in 86d056c38 — added failed-status parse() regression cases for both the OpenAI and Azure wrappers, covering the stop reason, error, and provider metadata.

packages/ai/src/openai/index.ts
marandaneto: The 200 is intentional here: $ai_http_status records the transport-level HTTP result, while $ai_is_error and $ai_error represent the provider-level failed Response returned over that successful transport. Using 500 would misreport the actual HTTP response.

packages/ai/src/openai/azure.ts
marandaneto: Good catch! Fixed in 86d056c38 — Azure streaming now reads chunk.response.usage for every response status, including completed responses.

packages/ai/src/openai/utils.ts
marandaneto: Fixed in 86d056c38 — getResponseFailure now returns unknown without the redundant union, and both terminal-response helpers have documentation comments describing their event roles.

packages/ai/src/openai/utils.ts
marandaneto: Fixed in 86d056c38 — incompleteDetails now uses OpenAI.Responses.Response["incomplete_details"] so the provider metadata shape is documented by the SDK type.

PR #4370, opened 2026-07-31
fix(ai): capture Agents Chat Completions usage

marandaneto:
## Problem

OpenAI Agents Chat Completions spans could omit model and token usage because those fields are nested in raw response data.

## Changes

- Extract Chat Completions model and usage from raw Agents span output when canonical fields are absent.
- Preserve canonical span fields when present.
- Add coverage for prompt, completion, cached, and reasoning token details.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

### DIFF
--- a/packages/ai/src/openai-agents/processor.ts
+++ b/packages/ai/src/openai-agents/processor.ts
@@ -483,7 +483,11 @@ export class PostHogTracingProcessor implements TracingProcessor {
     groupId: string | null,
     errorProperties: Record<string, any>
   ): void {
-    const usage = spanData.usage ?? {}
+    // OpenAI Agents 0.8 stores the raw Chat Completions response in output[0].
+    // Canonical generation fields take precedence when the SDK provides them.
+    const rawResponse = spanData.output?.[0]
+    const usage = spanData.usage ?? rawResponse?.usage ?? {}
+    const model = spanData.model ?? (rawResponse?.model as string | undefined)
     const inputTokens = (usage.input_tokens as number) || (usage as any).prompt_tokens || 0
     const outputTokens = (usage.output_tokens as number) || (usage as any).completion_tokens || 0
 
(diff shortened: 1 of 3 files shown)


ioannisj (approved):

packages/ai/src/openai-agents/processor.ts
```diff
@@ -513,6 +517,16 @@ export class PostHogTracingProcessor implements TracingProcessor {
...
     }
 
+    // Raw Chat Completions usage keeps token details under provider-specific fields.
+    const promptTokenDetails = (usage as any).prompt_tokens_details
+    const completionTokenDetails = (usage as any).completion_tokens_details
+    if (completionTokenDetails?.reasoning_tokens) {
+      properties.$ai_reasoning_tokens = completionTokenDetails.reasoning_tokens
+    }
+    if (promptTokenDetails?.cached_tokens) {
+      properties.$ai_cache_read_input_tokens = promptTokenDetails.cached_tokens
```
Radu-Raicea, line 534: Could we set `$ai_cache_reporting_exclusive: false` when using `rawResponse.usage`? Chat Completions’ prompt_tokens includes cached tokens, and without the flag Claude models on OpenAI-compatible providers can have their cache usage double-counted.

Radu-Raicea (approved):

packages/ai/src/openai-agents/processor.ts
marandaneto: Good catch! Fixed in 7f5c9b6b4 — raw Chat Completions usage now declares `$ai_cache_reporting_exclusive: false`, while canonical generation usage remains unchanged.

PR #4371, opened 2026-07-31
fix(ai): support current LangChain message metadata

marandaneto:
## Problem

Current LangChain AIMessage objects report usage and stop metadata in fields that the callback integration did not read.

## Changes

- Read usage_metadata and response_metadata from current LangChain messages.
- Retain fallbacks for legacy LangChain generation metadata.
- Add regression coverage for token details and finish reasons.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

### DIFF
--- a/packages/ai/src/langchain/callbacks.ts
+++ b/packages/ai/src/langchain/callbacks.ts
@@ -659,28 +659,17 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       return undefined
     }
     const gen = lastGeneration[0]
+    const messageResponseMetadata = (gen as any).message?.response_metadata
+    const generationResponseMetadata = gen.generationInfo?.response_metadata
+    const stopReason =
+      messageResponseMetadata?.finish_reason ||
+      messageResponseMetadata?.stop_reason ||
+      gen.generationInfo?.finish_reason ||
+      generationResponseMetadata?.stop_reason ||
+      generationResponseMetadata?.finish_reason ||
+      gen.generationInfo?.stop_reason
 
-    // Check generationInfo for finish_reason (OpenAI format)
-    if (gen.generationInfo?.finish_reason) {
-      return String(gen.generationInfo.finish_reason)
-    }
-
-    // Check generationInfo for response_metadata.stop_reason (Anthropic format)
-    if (gen.generationInfo?.response_metadata?.stop_reason) {
-      return String(gen.generationInfo.response_metadata.stop_reason)
-    }
-
-    // Check message response_metadata for finish_reason (common LangChain format)
-    if (gen.generationInfo?.response_metadata?.finish_reason) {
-      return String(gen.generationInfo.response_metadata.finish_reason)
-    }
-
-    // Check for stop_reason directly in generationInfo
-    if (gen.generationInfo?.stop_reason) {
-      return String(gen.generationInfo.stop_reason)
-    }
-
-    return undefined
+    return stopReason != null ? String(stopReason) : undefined
   }
 
   private _parseUsageModel(usage: any, provider?: string, model?: string): [number, number, Record<string, any>] {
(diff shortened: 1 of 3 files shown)


packages/ai/tests/callbacks.test.ts
```diff
@@ -79,6 +79,85 @@ describe('LangChainCallbackHandler', () => {
...
+        ],
+      ],
+    }
+    handler.handleLLMEnd(llmResult, runId)
+
+    const [captureCall] = (mockPostHogClient.capture as jest.Mock).mock.calls
+    expect(captureCall[0].properties['$ai_input_tokens']).toBe(15)
+    expect(captureCall[0].properties['$ai_output_tokens']).toBe(5)
+    expect(captureCall[0].properties['$ai_stop_reason']).toBe('end_turn')
+  })
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 673161157 by parameterizing the two LangChain metadata cases while preserving their provider-specific inputs and expectations.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -659,28 +659,17 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
       return undefined
     }
     const gen = lastGeneration[0]
+    const messageResponseMetadata = (gen as any).message?.response_metadata
+    const generationResponseMetadata = gen.generationInfo?.response_metadata
+    const stopReason =
+      messageResponseMetadata?.finish_reason ||
+      messageResponseMetadata?.stop_reason ||
+      gen.generationInfo?.finish_reason ||
```
ioannisj, line 667: [suggestion] Worth extending the new `it.each` with a case or two that carry only `generationInfo` and no `message`?

Nothing in `packages/ai/tests/callbacks.test.ts` touches `generationInfo` at all, before or after this PR, and the legacy tail here just went from explicit `if (…) return` branches to a single `||` chain. The ordering looks preserved exactly so it's not a regression today, but a future edit that reorders or drops one of those branches would still pass the suite.

packages/ai/src/langchain/callbacks.ts
```diff
@@ -823,18 +812,17 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
...
+          const messageResponseMetadata = message.response_metadata ?? {}
+          const generationInfo = genChunk.generationInfo ?? {}
+          const generationResponseMetadata = generationInfo.response_metadata ?? {}
           const chunkUsage =
-            responseMetadata['usage'] ??
-            responseMetadata['amazon-bedrock-invocationMetrics'] ??
-            messageChunk.usage_metadata
+            message.usage_metadata ??
+            messageResponseMetadata['usage'] ??
+            messageResponseMetadata['amazon-bedrock-invocationMetrics'] ??
```
ioannisj, line 822: [suggestion] `amazon-bedrock-invocationMetrics` is checked twice in this chain, here under `message.response_metadata` and again under `generationInfo.response_metadata` three lines down, and the key doesn't appear anywhere in the test suite. Bedrock models report usage only under it, so typo'ing or dropping either position would zero `$ai_input_tokens` and `$ai_output_tokens` for every Bedrock call with nothing failing.

`_parseUsageModel` already maps `inputTokenCount`/`outputTokenCount`, so it's a couple of lines per position. Worth dropping in, or better as a follow-up?

ioannisj (approved):

packages/ai/src/langchain/callbacks.ts
marandaneto: Added in 1ebad1447. The parameterized tests now cover `amazon-bedrock-invocationMetrics` under both `message.response_metadata` and `generationInfo.response_metadata`, including the Bedrock token field mappings.

packages/ai/src/langchain/callbacks.ts
marandaneto: Good suggestion! Added coverage in 1ebad1447 for both direct and nested `generationInfo` metadata without a `message`, including usage and stop-reason extraction.

Radu-Raicea (approved):

PR #4379, opened 2026-08-01
fix(mcp): respect reserved argument ownership

marandaneto:
## Problem

This is a follow-up to the review comments on [#4355](https://github.com/PostHog/posthog-js/pull/4355), especially the agreed follow-ups called out in the [re-review](https://github.com/PostHog/posthog-js/pull/4355#pullrequestreview-4829599659).

The original fix preserved tool-owned `context` and `conversation_id` arguments at callback time, but analytics could still consume those values as intent or conversation metadata. The low-level server path also stripped `conversation_id` without ownership information, while SDK-injected arguments could fail strict Zod validation before callback cleanup ran.

## Changes

- Track reserved-argument ownership from the effective high-level registry schema and low-level `tools/list` schemas.
- Preserve tool-owned `context` and `conversation_id` values without using them as analytics intent or conversation metadata.
- Strip only SDK-owned arguments before high-level Zod validation and low-level tool execution.
- Keep low-level behavior conservative before ownership is learned, and retain ownership across paginated tool listings.
- Skip injection for root `$ref` schemas and distinguish JSON Schema keywords from Zod internals.
- Add high-level, low-level, strict-schema, pagination, and schema-collision regression coverage.
- Document the ownership behavior and known schema edge cases.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common
- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

A Pi coding agent implemented the human-directed follow-up and used the repository's Jest, ESLint, Prettier, and Rslib tooling. The bundle-driven autoreview skill was run repeatedly; accepted findings added pagination-safe ownership caching, root `$ref` handling, and safer JSON Schema detection. The final review found the patch correct, with only a non-blocking note about an existing pagination cursor-forwarding gap handled separately.

### DIFF
--- a/packages/mcp/src/__tests__/instrument-lowlevel.test.ts
+++ b/packages/mcp/src/__tests__/instrument-lowlevel.test.ts
@@ -209,69 +209,39 @@ describe('Low-level Server tracing (e2e)', () => {
     }
   })
 
-  it('preserves tool-owned reserved arguments without consuming them as analytics metadata', async () => {
-    const { server, client, receivedCalls, connect, cleanup } = await setupLowLevelServer()
-    try {
-      instrument(server, fakePostHog(), { context: true, enableConversationId: true })
-      await connect()
-      await client.request({ method: 'tools/list', params: {} }, ListToolsResultSchema)
+  it.each(['context', 'conversation_id'] as const)(
+    'preserves a tool-owned %s argument without consuming it as analytics metadata',
+    async (reservedArgument) => {
+      const { server, client, receivedCalls, connect, cleanup } = await setupLowLevelServer()
+      try {
+        instrument(server, fakePostHog(), { context: true, enableConversationId: true })
+        await connect()
+        await client.request({ method: 'tools/list', params: {} }, ListToolsResultSchema)
 
-      const suppliedResult = await client.request(
-        {
-          method: 'tools/call',
-          params: {
-            name: 'owned_reserved',
-            arguments: {
-              context: 'application state',
-              conversation_id: 'application conversation',
-              value: 'first',
-            },
+        const suppliedArguments = { [reservedArgument]: 'application value', value: 'kept' }
+        const result = await client.request(
+          {
+            method: 'tools/call',
+            params: { name: 'owned_reserved', arguments: suppliedArguments },
           },
-        },
-        CallToolResultSchema
-      )
-      const omittedResult = await client.request(
-        {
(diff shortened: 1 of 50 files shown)


ioannisj (approved):

packages/mcp/src/__tests__/instrument-lowlevel.test.ts
```diff
@@ -193,6 +209,212 @@ describe('Low-level Server tracing (e2e)', () => {
     }
   })
 
+  it('preserves tool-owned reserved arguments without consuming them as analytics metadata', async () => {
```
lucasheriques: the reserved names are hardcoded as literals in here and in `reserved-arguments.test.ts` rather than iterated, so `context` and `conversation_id` each get their own near-duplicate block. `it.each(['context', 'conversation_id'])` would mean the next reserved parameter is one row instead of another copied test.

not blocking anything, it just gets more tempting to skip as the set grows.

lucasheriques (approved):
went looking for a silent-overwrite collision and could not build one. `analyticsOwnsParameter` returns false as soon as a tool declares `context` or `conversation_id` itself, and both call sites gate stripping and capture off that same boolean, so the user arg wins in both directions on both the high-level and low-level paths.

also checked the two deletions, since that is where this kind of change usually loses something. `cloneRequestWithoutConversationId` and the `removeContextFromArgs` closure both stripped unconditionally, which was the bug, and the ownership-gated path covers what they did. nothing dropped.

one note on the tests inline.

packages/mcp/src/__tests__/instrument-lowlevel.test.ts
marandaneto: Good suggestion — fixed in 9fb93ed23. The high- and low-level tool-owned reserved argument tests now use `it.each` for `context` and `conversation_id`, so each reserved name is covered as a separate row without duplicating the test flow.

PR #4380, opened 2026-08-01
fix(core): bound response body reads by request timeout

marandaneto:
## Problem

Follow-up to #4346, which fixed the equivalent response-body timeout behavior in the Node capture v1 transport.

`@posthog/core` stopped enforcing its request timeout once response headers arrived. A server or injected fetch implementation that stalled while reading or cancelling the response body could therefore block feature flag requests, retries, and shutdown indefinitely.

The capture UUID documentation also overstated the immediacy of server-side deduplication.

## Changes

- Apply one request deadline across both header retrieval and required response-body consumption.
- Retry required JSON reads that exceed the deadline while treating an already accepted write-only response as successful, avoiding duplicate capture, log, or metric writes.
- Preserve HTTP status classification when an error response body stalls.
- Cancel late responses from injected fetch implementations that ignore abort signals.
- Clarify that repeated capture UUIDs are eventually deduplicated in storage rather than strictly at ingestion time.
- Add regression coverage for stalled success, error, and feature flag response bodies.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent in a dedicated worktree. The response semantics were independently reviewed and the patch was checked with Pi autoreview. Focused Jest tests, affected-package lint, and the core build passed on Node.js 24. Human review is required before merge.


eli-r-ph:
> Clarify that repeated capture UUIDs are eventually deduplicated in storage rather than strictly at ingestion time.

One note about this - ingestion team is working on a service to sit behind capture that will do an initial dedup pass based on event UUIDs that will greatly improve the current ClickHouse process, and operate at ingestion time. The ClickHouse team is also working on improvements to its own compaction and deduplication background jobs.

Neither are shipped yet, but we can update the expectations here and docs once they are. I'll keep you in the loop about this

eli-r-ph (commented):
Reviewed the deadline mechanism against the head sources and the unchanged helpers it depends on (`retriable`, `isPostHogFetchRetryableError`, `isRetryableFlagsFetchError`, `categorizeRequestError`, `logFlushError`, `PostHogFetchResponse`, `safeSetTimeout`). The core mechanism is sound.

Things I specifically checked and found correct, so they are *not* flagged below:

- The `deadline` promise always has a handler attached (the first `Promise.race`), so no unhandled rejection.
- `reject()` before `ctrl.abort()` does make the deadline error win the race — `abort()` is synchronous and promise handlers are microtasks, so `ctrl.signal.aborted` is already `true` when the rejection handler runs.
- The late-response cancel (`fetchPromise.then`) and `finally`'s `cancelBody()` are mutually exclusive (`res` set ⟺ `responseAccepted`), so there is no double-cancel; `cancellation ??=` memoizes correctly.
- The timeout error keeps `name === 'AbortError'`, so `categorizeRequestError` still maps it to `{type:'timeout'}` and `isRetryableFlagsFetchError` still treats it as retryable.
- `successful-write` swallowing a stalled cancel does not re-send an accepted batch.
- `_sendLogsBatch` / `_sendMetricsBatch` previously never consumed the body at all — now they cancel it. Improvement.
- Overload resolution (`{type:'required'}` → `Promise<T>`, `{type:'successful-write'}` → `Promise<void>`) is correct.
- `let timer: ReturnType<typeof safeSetTimeout>` does not trip TS2454, but only because `safeSetTimeout` returns `any`. Fragile, not a compile error today.

Eight inline comments follow: 2 medium, 6 low.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1647,28 +1695,102 @@ export abstract class PostHogCoreStateless {
...
+          const isNoCors = options.mode === 'no-cors'
+          if (!isNoCors && (res.status < 200 || res.status >= 400)) {
+            const httpError = new PostHogFetchHttpError(res, reqByteLength)
+            try {
+              httpError.setResponseBodyText(await Promise.race([res.text(), deadline]))
+            } catch {
+              // A stalled error body must not replace its HTTP status classification.
+              httpError.setResponseBodyText('')
+            }
+            throw httpError
```
eli-r-ph: **Medium** — Error bodies are now read eagerly on *every* attempt, before the retry decision. A retryable status (408/429/5xx) whose body is slow or never ends now burns the full request deadline per attempt: at defaults (`requestTimeout` 10s, `fetchRetryCount` 3, `fetchRetryDelay` 3s) a flush goes from ~9s to ~49s, which blows past the 30s default `_shutdown` budget. Previously the status threw immediately and the body was read at most once, lazily, in `logFlushError`.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1647,28 +1695,102 @@ export abstract class PostHogCoreStateless {
...
+          // We only throw on HTTP errors if we're not in no-cors mode.
+          // https://developer.mozilla.org/en-US/docs/Web/API/Request/mode#no-cors
+          const isNoCors = options.mode === 'no-cors'
+          if (!isNoCors && (res.status < 200 || res.status >= 400)) {
+            const httpError = new PostHogFetchHttpError(res, reqByteLength)
+            try {
+              httpError.setResponseBodyText(await Promise.race([res.text(), deadline]))
+            } catch {
+              // A stalled error body must not replace its HTTP status classification.
+              httpError.setResponseBodyText('')
```
eli-r-ph: **Medium** — On a stalled error body the code records `''`, which is indistinguishable from a genuinely empty body. `err.text` then silently reports "no body" and `err.json` rejects with a `SyntaxError` from `JSON.parse('')` rather than surfacing that the read was abandoned. A sentinel or an `bodyReadTimedOut` flag would keep the diagnostic.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1647,28 +1695,102 @@ export abstract class PostHogCoreStateless {
...
+            throw httpError
+          }
+
+          if (responseHandling.type === 'successful-write') {
+            try {
+              await Promise.race([cancelBody(), deadline])
+            } catch (e) {
+              if (!ctrl.signal.aborted) {
+                throw e
+              }
```
eli-r-ph: **Low** — `if (!ctrl.signal.aborted) { throw e }` is unreachable: `cancelBody()` swallows its own rejections internally, so the race can only reject via `deadline`, which always aborts before the handler runs. Either drop the branch or add a comment saying it is defensive.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -62,11 +63,17 @@ class PostHogFetchHttpError extends Error {
...
-    return this.response.text()
+    return this.responseBodyText === undefined ? this.response.text() : Promise.resolve(this.responseBodyText)
   }
 
   get json(): Promise<any> {
-    return this.response.json()
+    return this.responseBodyText === undefined
+      ? this.response.json()
+      : Promise.resolve().then(() => JSON.parse(this.responseBodyText as string))
+  }
```
eli-r-ph, line 109: **Low** — The `responseBodyText === undefined` fallbacks in the `text`/`json` getters are dead. The only construction site calls `setResponseBodyText` on both the success and the catch path before throwing, so the field is always a string by the time the error escapes.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -62,11 +63,17 @@ class PostHogFetchHttpError extends Error {
...
+
+  setResponseBodyText(text: string): void {
+    this.responseBodyText = text
   }
```
eli-r-ph: **Low** — `setResponseBodyText` is a public mutator on an error object that is thrown to user code. Passing the body text through the constructor (or a `#private` field set by a factory) keeps the error immutable once thrown.

packages/core/src/posthog-core-stateless.ts
```diff
@@ -1647,28 +1695,102 @@ export abstract class PostHogCoreStateless {
...
+          }
+
+          // If we're in no-cors mode, we can't access the response status.
+          // We only throw on HTTP errors if we're not in no-cors mode.
+          // https://developer.mozilla.org/en-US/docs/Web/API/Request/mode#no-cors
+          const isNoCors = options.mode === 'no-cors'
+          if (!isNoCors && (res.status < 200 || res.status >= 400)) {
+            const httpError = new PostHogFetchHttpError(res, reqByteLength)
+            try {
+              httpError.setResponseBodyText(await Promise.race([res.text(), deadline]))
```
eli-r-ph: **Low** — `err.response`'s stream is now already consumed when the error reaches user code from a rejected `flush()`/`shutdown()`. `PostHogFetchHttpError` is not exported, but `response` is a public field, so anyone duck-typing `err.response.text()` in a catch block now gets a "body already used" `TypeError`. Worth a changeset note given the "no breaking changes" checkbox.

packages/core/src/__tests__/posthog.flush.spec.ts
```diff
@@ -12,6 +12,7 @@ describe('PostHog Core', () => {
         flushAt: 5,
         fetchRetryCount: 3,
         fetchRetryDelay: 100,
+        requestTimeout: 1000,
```
eli-r-ph: **Low** — `requestTimeout: 1000` was added to the shared `beforeEach` client, so it now applies to every test in the describe block rather than only the new ones. The new tests already build their own clients elsewhere in the file (line 255); doing the same here keeps the blast radius contained.

packages/core/src/__tests__/posthog.flush.spec.ts
```diff
@@ -177,6 +178,131 @@ describe('PostHog Core', () => {
...
+        json: () => Promise.resolve({ status: 'ok' }),
+        body: { cancel } as any,
+      })
+
+      posthog.capture('test-event-1')
+      await expect(posthog.flush()).resolves.toBeUndefined()
+
+      expect(cancel).toHaveBeenCalledTimes(1)
+      expect(jest.getTimerCount()).toBe(0)
+    })
```
eli-r-ph: **Low** — Coverage stops at the flags and write paths. `getRemoteConfig` (`posthog-core-stateless.ts:630`) and `getSurveys` (`:997`) were converted to the same `required` handling and are untested for a stalled body, as is the new `PostHogFetchHttpError` text/json caching. One parameterized case over the three `required` call sites would cover it cheaply.

packages/core/src/posthog-core-stateless.ts
marandaneto: Fixed in 47d2be389 — removed the unreachable branch and kept the catch focused on the deadline-only cancellation case.

packages/core/src/__tests__/posthog.flush.spec.ts
marandaneto: Fixed in 47d2be389 — moved the 1-second request timeout into a nested `response body deadlines` test scope, leaving the shared flush client at its default timeout.

packages/core/src/__tests__/posthog.flush.spec.ts
marandaneto: Fixed in 47d2be389 — added parameterized stalled-body coverage for remote config, flags, and surveys, plus coverage for cached HTTP error `text`/`json` access and the timeout diagnostic.

packages/core/src/posthog-core-stateless.ts
marandaneto: Fixed in 47d2be389 — the error response is no longer eagerly consumed. `err.response` remains available until the original request deadline; only `err.text`/`err.json` consumes it, and the deadline eventually aborts/cancels an ignored body. Since the prior observable access pattern is preserved, I did not add a breaking-change note.

packages/core/src/posthog-core-stateless.ts
marandaneto: Fixed in 47d2be389 — timed-out body reads now reject with an `AbortError`, expose `bodyReadTimedOut`, and `logFlushError` prints an explicit `<response body read timed out>` sentinel instead of treating it as an empty body.

packages/core/src/posthog-core-stateless.ts
marandaneto: Fixed in 47d2be389 — the getters now share one lazy, cached response-text promise, so the dead undefined fallbacks are gone and `json` parses the cached text.

packages/core/src/posthog-core-stateless.ts
marandaneto: Good catch! Fixed in 47d2be389 — HTTP error bodies are now read lazily, so retryable statuses retry immediately. Discarded retry responses are canceled without delaying the retry, while the final surfaced error remains readable until its original request deadline.

packages/core/src/posthog-core-stateless.ts
marandaneto: Fixed in 47d2be389 — removed `setResponseBodyText`; body reading and deadline cleanup are now encapsulated by the error from construction onward.

ioannisj (approved):

marandaneto:
@eli-r-ph are we good to go?

PR #4381, opened 2026-08-01
fix: clear completed lifecycle timeout handles

marandaneto:
## Problem

Several SDK lifecycle operations raced asynchronous cleanup against a timeout but left the losing timeout handle scheduled when cleanup completed first. Those handles unnecessarily kept the Node.js event loop alive and caused Jest to report open handles after otherwise successful test runs.

## Changes

- Retain and clear lifecycle timeout handles after core log shutdown and flush operations settle.
- Apply the same cleanup to Node.js metrics shutdown and feature flag cache-provider shutdown.
- Clear the React Native storage-drain timeout after pending writes finish.
- Preserve timeout-winning and late-rejection behavior.
- Add regression coverage for the successful fast path, including cache-provider shutdown, without leaving fake timers active during React Native shutdown.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent in a dedicated worktree after investigating an open-handle CI failure. The repeated timeout-race pattern was independently reviewed and the final patch passed Pi autoreview. The full React Native Jest suite, open-handle detection, affected-package lint, and affected builds passed on Node.js 24. Human review is required before merge.

### DIFF
--- a/packages/core/src/logs/index.ts
+++ b/packages/core/src/logs/index.ts
@@ -1,7 +1,7 @@
 import type { LogAttributeValue } from '@posthog/types'
 import { buildOtlpLogRecord, buildOtlpLogsPayload, buildResourceAttributes } from './logs-utils'
 import { Logger, PostHogPersistedProperty } from '../types'
-import { isArray, safeSetTimeout } from '../utils'
+import { isArray, raceWithTimeout, safeSetTimeout } from '../utils'
 import type { BufferedLogEntry, CaptureLogOptions, LogSdkContext, LogsHost, ResolvedPostHogLogsConfig } from './types'
 
 // Caps the retry backoff at 2^6 = 64× the flush interval.
@@ -383,17 +383,7 @@ export class PostHogLogs {
       await flushPromise
       return
     }
-    let timeoutHandle: ReturnType<typeof safeSetTimeout> | undefined
-    try {
-      await Promise.race([
-        flushPromise,
-        new Promise<void>((resolve) => {
-          timeoutHandle = safeSetTimeout(resolve, timeoutMs)
-        }),
-      ])
-    } finally {
-      clearTimeout(timeoutHandle)
-    }
+    await raceWithTimeout(flushPromise, timeoutMs)
   }
 
   /**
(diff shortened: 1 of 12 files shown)


packages/react-native/test/posthog.spec.ts
```diff
@@ -1895,20 +1895,30 @@ describe('PostHog React Native', () => {
...
+
+        await posthog.shutdown(5000)
+
+        // Both pipelines drained through the shared shutdown path. Logs use
+        // the smaller of the caller's shutdown budget and the configured
+        // `terminationFlushBudgetMs` (default 2000ms) — see _shutdown.
+        expect(logsShutdownSpy).toHaveBeenCalledWith(2000)
+        expect(sendLogsSpy).toHaveBeenCalled()
+
+        const drainTimeoutIndex = setTimeoutSpy.mock.calls.findIndex(([, delay]) => Number(delay) > 4000)
```
ioannisj: [suggestion] I think this assertion passes with the `posthog-rn.ts` change in this PR reverted, so a regression that reintroduces the drain-timer leak would still go green here.

`findIndex` takes the first setTimeout over 4000ms, and that's the core shutdown timer: `super._shutdown(5000)` is awaited inside the `Promise.all` above, so `posthog-core-stateless.ts:1731` arms its timer at exactly 5000 before the `finally` arms the drain one with `remainingMs`. And that handle already gets cleared in its own `finally`.

Would `findLastIndex` do it? The drain timer is the last one armed during shutdown since it's created in the `finally` that runs after `Promise.all` resolves, which also drops the dependence on the `> 4000` threshold happening to miss the core timer.

packages/core/src/logs/index.ts
```diff
@@ -383,7 +383,17 @@ export class PostHogLogs {
       await flushPromise
       return
     }
-    await Promise.race([flushPromise, new Promise<void>((resolve) => safeSetTimeout(resolve, timeoutMs))])
+    let timeoutHandle: ReturnType<typeof safeSetTimeout> | undefined
```
ioannisj: [suggestion] Worth pulling this into a `raceWithTimeout(promise, ms, onTimeout?)` next to `safeSetTimeout` in `packages/core/src/utils/index.ts`, so one place owns the arm/clear lifecycle?

The declare, assign-in-executor, clear-in-`finally` shape lands in six places once this merges (there's already one at `posthog-core-stateless.ts:1727` predating the PR). This PR exists because the un-cleared variant got copied five times, so six copies of the corrected one sets the next site up for the same drift.

The rejecting variant in `feature-flags.ts` and the `timedOut` flag in `flushWithTimeout` both fit as a callback. Could be a follow-up if you'd rather keep this PR tight.

ioannisj (approved):
LG 2 bot suggestions

packages/react-native/test/posthog.spec.ts
marandaneto: Good catch! Fixed in cc44328e3 — the test now uses the last timeout armed during shutdown, so it specifically asserts that the drain timer is cleared without relying on the delay threshold.

packages/core/src/logs/index.ts
marandaneto: Agreed and fixed in cc44328e3. Added a shared `raceWithTimeout` utility next to `safeSetTimeout`, migrated all six arm/clear lifecycle races to it (including rejecting and late-rejection callback variants), and added focused utility coverage for both race outcomes and rejection paths.

PR #4383, opened 2026-08-01
fix(browser): clear register_for_session properties on session rotation

nandinitiw:
Closes #3573

## Problem

`register_for_session`'s docstring states these properties are "cleared when the session ends," but nothing actually cleared `sessionPersistence` when `$session_id` rotates (activity timeout, max session length, etc.). Properties persisted in `sessionStorage` as long as the browser tab stayed open, requiring `unregister_for_session` as a manual workaround.

The gap was even anticipated in a comment in `sessionid.ts`, describing "session-scoped props" as a consumer that should react to session id changes but never did.

## Fix

Hooks into the existing `SessionIdManager.onSessionId` mechanism — the same pattern `SessionPropsManager` already uses for `$client_session_props`. A new handler, `_clearSessionPersistenceOnNewSession`, subscribes and clears `sessionPersistence` only on a genuine session id change.

Two edge cases handled:
- `onSessionId` fires immediately on subscription with the *current* session id. The handler treats its first call as a baseline recording rather than a clear, so properties registered before the handler is wired up (or across a page load within the same session) aren't wiped.
- When `config.persistence` is `'sessionStorage'` or `'memory'`, `sessionPersistence` and `persistence` are the same instance. Clearing in that case would wipe unrelated state (distinct_id, super properties, feature flags), so it's skipped.

## Testing

Added `posthog-core.session-persistence-reset.test.ts` covering:
- properties clear on a real session id rotation (via `resetSessionId()` + `checkAndGetSessionAndWindowId()`)
- properties are *not* cleared within the same session
- properties registered before the very first rotation aren't wiped
- shared-instance mode (`persistence: 'memory'`) doesn't wipe unrelated persisted state

### DIFF
--- a/.changeset/session-persistence-clear-on-rotation.md
+++ b/.changeset/session-persistence-clear-on-rotation.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Clear `register_for_session` properties when the `$session_id` rotates, matching the documented behavior that session super properties are cleared when the session ends.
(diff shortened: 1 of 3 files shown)


marandaneto:
#4197

PR #4384, opened 2026-08-02
feat(rrweb): opt-in time-sliced full snapshots, hardened (builds on #4233)

arnohillen:
## Problem

#4233 introduced opt-in time-sliced full snapshots (`fullSnapshotYieldBudgetMs`) to stop the recorder freezing large-DOM pages for seconds (#4217). Four review rounds on that PR kept finding the same class of correctness gap — subsystems that hadn't been taught the walk-window protocol — plus two regressions on the **default synchronous path**. This PR builds directly on #4233's branch (all original commits preserved) and addresses every finding from its review threads, without changing the feature's public surface: same option, same default-off behavior, opt-in only.

## Why

The customers who need this option are exactly the ones with huge, busy DOMs — where every one of the reported bugs has its highest probability. The feature has to be trustworthy before anyone re-enables replay on it, and the default path has to stay exactly as safe as `main`.

## Changes

**Default (budget = 0) path restored to `main` semantics** — two regressions fixed:
- restored `canvasManager.onFullSnapshot()` (dropped in a rebase; `main`'s canvas-masking test now passes again)
- a synchronous snapshot failure releases its locks and propagates to the SDK's catch-and-retry instead of silently tearing recording down; Meta is emitted before the buffer lock again

**Transaction safety** (review blocker "commit can strand buffers locked forever"):
- `lock/commit/discard` iterate an array snapshot with per-buffer error isolation, keep the token armed until every holder is released, sweep buffers born mid-commit, force-discard by token even when the global owner was already cleared; `reset()` clears a torn-down buffer's lock token
- Meta emission, the held-event flush and the buffer commit are all inside the recovery boundary now — a throwing consumer callback can no longer latch the recorder in-flight forever or leak locks

**Replay correctness during the walk window**:
- the walker now **skips nodes the locked buffers will re-add at commit** (added/moved mid-walk); the buffer is their single source of truth, carrying live positions and final state. This replaces the previous forget-pending-adds approach, which lost whole subtrees when a serialized ancestor moved before its children were reached, and froze same-parent reorders at stale positions
- **id reservation stays claimable through the commit** (with new-handout paused), so held events referencing mid-walk nodes are **deferred past the commit and delivered against the ids the commit claims** — instead of being scrubbed. This fixes lost adopted stylesheets on mid-walk shadow roots, cross-origin child-frame attaches, selections, focus and fullscreen events; only events whose target genuinely left the DOM are dropped
- **CSSOM deltas are reconciled against what the snapshot actually captured**: the serializer reports per stylesheet whether it inlined live CSSOM or kept raw author text, and each held delta is delivered or dropped based on that plus a serialization-order marker — closing both the lost-rule direction (raw text kept: `var()` shorthand, multi-text-node styles) and the double-apply direction
- cross-origin iframe reattach now happens **before** the held-event flush, and failure paths no longer reset the mirror (which orphaned attached iframe trees)
- the checkout clock (`checkoutEveryNms`) runs from when the snapshot reached the wire, not its backdated timestamp — a walk longer than the interval no longer self-sustains a snapshot loop

**Bounded degradation, visibly**:
- canvas FPS frames are dropped while the walk holds the lock (the restored dedup reset re-captures every canvas right after) instead of flooding the held queue
- the held-queue caps are joined by a locked-buffer backlog cap and a wall-clock watchdog; overflow/watchdog aborts get **one budgeted retry before the synchronous fallback**
- a hidden page finishes the walk in one task (nothing is watching, and background throttling/tab freeze would otherwise pin the locks); `pagehide` drains the walk synchronously so short visits still get their snapshot and held events
- every degraded outcome (retry, sync fallback, fallback failure) emits a replay custom event (`tag: 'budgeted-full-snapshot'`) with reason and drop counts, and routes through the registered `errorHandler` — no more console-only failures
- `fullSnapshotYieldBudgetMs` is validated: only a finite positive number turns the feature on (`true`/`Infinity`/`NaN` mean off, with a warning)

**Tests** (net new coverage; all suites pass):
- convergence harness upgraded with a duplicate-delivery oracle (an add of an already-attached id now fails any scenario — final-state convergence alone couldn't see double delivery)
- new scenarios: moved-ancestor subtree, same-parent reorders, mid-walk adopted stylesheets, both CSSOM reconciliation directions, checkout-runaway guard, throwing consumer emit mid-flush, the production MessageChannel yielder path (previously untested), pagehide flush, held-event timestamp preservation, mid-walk selections
- unit tests for the transaction helpers under hostile consumers (throwing/splicing/buffer-born-mid-commit), the walker controller (`flushSync`, hidden-page fast path, watchdog, skip predicate), option validation, and the SDK allowlist plumbing
- a committed long-task benchmark: on a ~35k-node fixture the synchronous arm must produce a long task and the budgeted arm must produce **none** (≥50ms, 5× the 10ms budget) through the real MessageChannel yielder — CI now fails if budget adherence regresses

**Hygiene**: changeset rewritten (posthog-js minor + @posthog/types patch only, per repo convention; the old text described abandoned drop semantics), working-notes docs removed from the tree, TSDoc now states the budget is cooperative (not a hard bound), that events are delayed by the walk, and that the option is client-config only.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes!) — budget=0 (default) now matches `main` more closely than the base branch did; the option remains opt-in

Refs #4217. Builds on #4233 (original commits preserved); addresses the findings from its review threads.

---

---

# Hardening round (2026-08-06)

This round addressed the findings from the customer review at head 63d47a2 — the blockers (reentrant rotation past the generation check; start-failure paths stranding a recorder), the page-death path, the parked-yield watchdog, abort losses, checkout runaways, scrub holes, the hold-gate bypass for SDK control events, commit outcome accounting, and success-path telemetry (`completed` diagnostic + `$sdk_debug_replay_budgeted_snapshot_*`).

# Second hardening round (2026-08-11)

`main` was merged in (signed merge commit, all history preserved). Beyond textual resolution, the merge **wires #4320's stylesheet budget into the budgeted path**: `snapshotWithBudget` now arms the snapshot cost tracking (it was dead on exactly the CSS-heavy-page path this feature targets), both paths pass `inlineStylesheetBudgetRules` and #4376's attribute-masking options, and both drain the deferred-stylesheet queue (the budgeted completion inlines skipped sheets across idle callbacks after its held window flushes, counted in the `completed` diagnostic; failed walks drop their queue so a retry cannot inline stale links). This answers the thread's open question: with both merged, stylesheet-heavy pages get **bounded**, not just shortened.

Every previously-listed follow-up that was a correctness gap is now closed, verified by an adversarial per-finding re-review (each fix traced against the code, each FIXED verdict attacked by an independent skeptic pass, surviving variants fixed too):

- **Rotation mid-commit** (skeptic variant of the reentrancy blocker): `commitMutationBuffers` only clears the module lock owner while it still holds it, so a rotation from a commit-window mutation delivery can no longer disarm the new session's buffer gate. Test rotates the recorder from the commit's own delivery.
- **Shadow-DOM scroll blind window**: known roots re-arm immediately after `init()` on a sliced walk, the registry is module-level so **session rotation** (stop+start on the same page) does not reopen the window for the new session's first walk, pruning is two-strike so a momentarily detached host (portal move) is not forgotten at the wrong instant, and an iframe's teardown still forgets its roots for good. Tests: mid-checkout shadow scroll, rotation shadow scroll.
- **Carried-event timestamps** (skeptic variant of abort losses): events carried across an abort are clamped to the retry snapshot's timestamp — their observation stamps predate it, and the replayer's timestamp sort would otherwise place them before the snapshot that introduces their ids and drop them all. The abort-carry test now asserts full wire monotonicity.
- **Dangling held interactions** (skeptic variant of the ordering position): a held event targeting a node removed mid-walk before the walker reached it is undeliverable (its id is absent from the new snapshot); it is now dropped and counted in `droppedHeldEventCount` instead of delivered for the replayer to drop silently. Remap-space ids pass through untouched.
- **Cross-origin abort carry + transitive attach deferral**: under `recordCrossOriginIframes`, an abort no longer blanket-drops held child-frame mutations (the retry cannot re-produce them; the frame replayed frozen for the session) — they carry like other held events; and a forwarded child event held after its frame's *deferred* attach is deferred transitively instead of overtaking the attach.
- **Command-mode canvas double-application**: `onSerialize` discards a canvas's pending commands exactly when the walk baked its pixels into `rr_dataURL` (post-bake commands survive). The abort-side residual for masked/WebGL canvases in command mode is named in the changeset (not reachable via posthog-js, which only configures FPS sampling).
- **Estimator and caps**: `pendingRecordCount` includes `mapRemoves`, the backlog cap counts pending canvas commands, `estimateRetainedSize` counts `Map`/`Set`/`Blob`/cross-realm buffers, survives throwing getters and hostile proxies, bounds its own traversal, and coerces spoofed sizes (a fake `Symbol.toStringTag` produced `NaN`, which disabled the byte cap for the rest of the walk). `shouldAbort` is probed at the stride inside `flushSync` and on the hidden fast path.
- **SDK pagehide ordering**: the SDK's pagehide handler now calls the recorder's new `record.drainPendingSnapshotForUnload()` before flushing, removing the listener-registration-order assumption (false whenever recording starts before DOMContentLoaded).

## Remaining design positions and out-of-repo items

- **Observation ordering across the walk window**: a locked `MutationBuffer` coalesces a window into one batch with no single observation time, so no stamping choice preserves both observation order and wire monotonicity. The changeset and the code state exactly what is guaranteed; final DOM state converges.
- **Remote kill switch**: `fullSnapshotYieldBudgetMs` is client-init-only; a remote lever needs server-side remote-config schema changes outside this repo. Documented on the public option; the option defaults to off and customers can gate it behind their own flag.
- The walker's stack is deliberately uncapped (O(serialized nodes), same order the mirror already retains; its lifetime is bounded by the watchdog and the backlog cap) — rationale now in code.

## Test state at head

All suites green at the pushed head: rrweb 47 files / 555+ tests (convergence 36 incl. the runaway guard and monotonicity assertions, lifecycle 35, failure-recovery, canvas masking), rrweb-snapshot suites incl. the budgeted-walk stylesheet-budget regression and walker abort probes (17 hardening tests), browser jest 773 (replay suites incl. the new pagehide-drain test), types config snapshot, repo-wide `turbo lint` 30/30, both packages typecheck clean. The long-task CI benchmark still enforces zero ≥50ms tasks on the budgeted arm.


ryanbreen:
Following up from #4233. We're the customer behind the escalation this started from — the trace with the single ~3.9s recorder task landing mid-scroll — and session replay is still disabled across all of our properties. I appreciate that every finding from our last review was confirmed by your own deep review rather than argued with, and that #4233 was closed in favor of this PR with its commits preserved. Since this one claims to close that whole round plus the two default-path regressions, and it's the code we'd be re-enabling on, I read the full diff at head 63d47a2 the same way as last time: mechanism by mechanism, against the checkout, not the description.

The headline is good news. I went through the #4233 findings one at a time and most are genuinely, verifiably fixed. The `commitMutationBuffers` blocker we filed is closed properly — array snapshot, per-buffer try/catch, the token cleared only after every buffer is released, plus the post-loop sweep for buffers born during the commit ([observer.ts#L107-L150](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/observer.ts#L107-L150)). The double-serialization guard covers text, comment and blocked nodes; add-then-remove reordering is gated on `lockToken !== null` so budget-0 stays byte-identical; `canvasMaskingConfigured` reaches the walker; cross-origin restamping is back; `forgetAddedNode` was deleted outright in favor of `shouldSkipNode`, which is a root-cause fix rather than a patch; same-parent reorders reconcile; the AdoptedStyleSheet loss became a real two-phase defer/recheck ([index.ts#L1162-L1178](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1162-L1178)); option validation landed with tests; `docs/plans` is gone; the changeset is `posthog-js: minor`. Plus caps, a 30s watchdog, a hidden-page drain, degradation diagnostics, a production-yielder scenario, a CI long-task A/B, and a 33-scenario convergence harness — and TSDoc that says out loud the budget is cooperative, which is the honest answer rather than the flattering one. That's a lot of hard, unglamorous work, and the design is in much better shape than two weeks ago.

What follows is what survived that pass. I've dropped everything I couldn't reproduce from the code as written at this head.

**blocker: a reentrant `record()`/`stop()` from inside the FullSnapshot callback runs past the transaction's only generation check.** `completeBudgetedWalk` checks generation exactly once, at entry ([index.ts#L976-L989](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L976-L989)), then hands the FullSnapshot to the consumer — and never re-checks after that call. If the consumer synchronously starts a new recording (or PostHog rotates the recorder from that callback), `record()` bumps `recordingGeneration`, discards the old token, resets the mirror and can begin its own reservation ([#L365-L370](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L365-L370), [#L1275-L1284](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1275-L1284)) — then control returns to the *old* completion, which carries on. That isn't stale-but-harmless, because `wrappedEmit`, `takeFullSnapshot` and `canvasManager` are module-level `let` bindings reassigned wholesale by each `record()` call rather than per-transaction captures, and `mirror` is one instance that is only ever `.reset()`, never replaced. So the old completion's `mirror.pauseReservationHandout()` ([#L1095](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1095)), its held-event flush through the bare `wrappedEmit` identifier, its `canvasManager.onFullSnapshot()`, and its unconditional `mirror.endIdReservation()` ([#L1185-L1199](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1185-L1199)) all resolve, at call time, to session B's objects and B's live reservation — which B may be mid-walk on. A's held clicks release through B's consumer, and B's events on not-yet-visited nodes start resolving to `-1`. The nearest test ("a stopped walk does not poison the next recording session", [convergence#L1025](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/test/budgeted-snapshot-convergence.test.ts#L1025)) calls `stop()`/`record()` from outside any dispatch while the first walk is parked on a yield, so the entry check fires and it passes. Fix: re-check the generation after every uncontrolled call, or capture the emit/canvas/mirror handles into the transaction so a superseded completion can't touch the live session. Test: `stop()` then `record()` from inside the FullSnapshot callback.

**blocker: two start-failure paths leave an attached recorder that never emits a FullSnapshot, and the SDK cannot tell.**

- The one budgeted retry emits Meta before any transaction state exists and outside any local guard ([#L1264-L1269](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1264-L1269)), and `wrappedEmit` calls the consumer `emit` unguarded. If that throws, the retry-start catch swallows it and the unconditional `return` at [#L1059](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1059) skips the whole synchronous-fallback block below it ([#L1062-L1085](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1062-L1085)). Reachable in the shipping SDK exactly as in #4233: `_maskUrl` calls a customer-supplied `maskCapturedNetworkRequestFn` with no try/catch ([lazy-loaded-session-recorder.ts#L748-L758](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L748-L758)), from `onRRwebEmit`'s Meta branch. Because the budgeted `takeFullSnapshot()` returns `true` unconditionally, observers are already installed and `recording` is already true — a live recorder, zero FullSnapshot, no `errorHandler` registered to see the `reportError`, and only the unrelated 5-minute periodic snapshot to recover it ([#L839-L849](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L839-L849)).
- On the **default (budget 0)** path, a null `snapshot()` or a lock conflict returns `false` with only a `console.warn` ([#L878-L889](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L878-L889)) — Meta having already gone out at [#L872](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L872). `init()` bails before `handlers.push(observe(document))` and before `recording = true` ([#L1637-L1641](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1637-L1641)), yet `record()` still returns the stop closure at [#L1685](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1685), so posthog-js's only failure check, `if (!this._stopRrweb)`, can never trip. No observer was installed, so no future `emit` fires to drain `_tryRRWebMethod`'s retry queue: the wire carries one Meta, forever, while the SDK believes replay is running. On `main`, `if (!node) return console.warn(...)` at least let `init()` install observers.

That second one also cuts against the "nobody pays until they opt in" argument I made last round: the default path now carries new behavior (this, plus the orphan-Meta case below) on top of the size report's +590 kB (+3.27%).

**major: hold-and-deliver is not observation-ordered, and the FullSnapshot is stamped for a moment it doesn't describe.** Three mechanisms, one theme. (1) Locked `MutationBuffer`s create no event until `commit()` unlocks and calls `emit()` ([mutation.ts#L292-L300](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/mutation.ts#L292-L300)), while clicks/inputs/scrolls go straight into `eventQueue` — and the flush drains every held event *first*, then commits ([index.ts#L1155-L1160](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1155-L1160)). Mutations also keep no observation timestamp; they emit with the default `preserveTimestamp = false`, i.e. commit time. A text mutation in task 1 followed by a click in task 2 reaches the wire as `[MouseInteraction, Mutation]`, and by timestamp too. (2) Within `eventQueue`, the two-pass algorithm reverses order: an event referencing a pending id is diverted to `deferred` while a *later* event on an already-claimed target emits immediately in the same loop, and the deferred one is re-stamped at flush time ([#L1162-L1178](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1162-L1178)) — typing into a new input then clicking an old button replays as click-then-type. (3) The FullSnapshot is stamped `transaction.startedAt` under a comment saying the tree "describes the document as it was at walk start," but `processNode` serializes each node at pop time from live DOM state, including `.value` on inputs — so a node visited 800ms in contributes its 800ms-in state to a t0-stamped snapshot, which can show a value that a correctly-timestamped input event later "introduces." Final-DOM convergence can't see any of this by construction, and the harness's only ordering assertion is monotonic timestamps, which all three satisfy. There's no timestamp-only fix here; it wants one sequence assigned at every observer callback — including when MutationRecords enter a locked buffer — and a merge respecting add-before-reference.

**major: the page-death path.** Four things, same handler pair:

- `completeWalkBeforePageHides` is wired to `visibilitychange → hidden`, not just `pagehide` ([#L1642-L1652](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1642-L1652)), and `flushSync`'s body is `while (stack.length > 0) processNode()` with no budget, watchdog or abort probe inside the loop ([snapshot.ts#L2030-L2032](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L2030-L2032)); `shouldAbort` short-circuits to `false` while draining. An ordinary tab switch 300ms into a walk serializes the rest of the document in one task, then replays the held window and commits — from our seat, the original stall with a different trigger.
- If the walk already aborted (a held-queue overflow earlier in that walk), `flushSync` returns `null` ([snapshot.ts#L2023-L2029](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L2023-L2029)) and the handler does nothing, since it only calls `completeBudgetedWalk` when `node` is truthy. The in-code justification — an aborted walk is left for its normal completion path — doesn't hold on a dying page, because that path is the parked yield. A short busy visit ends with Meta and nothing else, where synchronous mode would have captured a snapshot.
- Nothing in that chain catches. `processNode` pops before calling `serializeNodeWithId`, and neither `flushSync` nor `completeWalkBeforePageHides` has a try/catch — a throwing mask fn or `toDataURL` leaves `finished`/`aborted` both false with a node permanently popped, and the driver's `if (finished || aborted) break` at [snapshot.ts#L2083-L2088](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L2083-L2088) will resume from the damaged stack and emit a silently truncated tree.
- When the drain *does* work, it proves the events reached rrweb's callback, not the wire. FullSnapshots are marked for compression ([lazy-loaded-session-recorder.ts#L286-L291](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L286-L291)) and go through an async native-gzip queue; the only synchronous drain is `_onBeforeUnload` → `_drainCompressionQueueSync()` + `_flushBuffer()` ([#L1918-L1929](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L1918-L1929)), and there's no `pagehide` handler anywhere in that file. `beforeunload` fires *before* `visibilitychange`/`pagehide`, so the rescued FullSnapshot lands in the buffer after the SDK's only synchronous last-chance flush has run. The pagehide test's consumer is `window.snapshots.push` — no buffer, no compression queue, no `sendBeacon` assertion. This one needs an SDK-side fix: a `pagehide` handler registered after rrweb, or a "drain complete" hook the SDK flushes on.

**major: a parked yield is still unbounded — the watchdog can't fire from inside the loop it lives in.** `maxWalkWallClockMs` is evaluated only after `++nodesSinceCheck >= 16` *and* `elapsed >= yieldBudgetMs`, immediately before `await doYield()` ([snapshot.ts#L2052-L2074](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L2052-L2074)). If that promise never settles, nothing else runs: no watchdog, no `shouldAbort`, no `dispose`. There's no `setTimeout` or `setInterval` anywhere in `record/index.ts` as a backstop, and while parked, held events accumulate until overflow sets `abortRequested` — which only the parked loop reads. A follow-up `takeFullSnapshot` coalesces into `budgetedSnapshotQueued` rather than forcing an abort. Your own hardening test states it plainly ("a yield that never resolves parks the async driver for good… parked forever by design", [snapshot-with-budget-hardening.test.ts#L87-L118](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/test/snapshot-with-budget-hardening.test.ts#L87-L118)) — understood and accepted, but it's the precise scenario I raised last round and the 30s constant doesn't bound it. A `setTimeout(maxWalkWallClockMs)` armed at walk start that sets `aborted` and resolves the pending yield closes it. Smaller sibling: `flushSync` has no handle to settle or dispose the parked yielder, so ports and closure stay retained after a successful drain.

**major: abort is lossy in both directions — it deletes what it was holding and keeps what it should have dropped.** On overflow, `wrappedEmit` zeroes the queue ([index.ts#L505](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L505)) and the failure branch zeroes it again and discards every locked buffer ([#L1017-L1031](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1017-L1031)) — for *any* abort reason: overflow, mutation backlog or watchdog alike. The retry re-snapshots structure but can't resurrect clicks, selections, mouse batches or CSSOM deltas, none of which self-heal by this PR's own correct argument. Meanwhile `MutationBuffer.discard()` clears `mapRemoves` without draining it ([mutation.ts#L302-L323](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/mutation.ts#L302-L323)) — while `emit()` right below drains it through `mirror.removeNodeFromMap` first, under a comment explaining that reparent detection depends on that order. Every abort path therefore leaves dead nodes resolvable in `idNodeMap` forever: a growing leak on exactly the huge-DOM pages this targets, `mirror.has(id)` answering true for reaped iframes, `getNextId` resolving removed siblings. Separately, a consumer throw during the flush is caught per-event and the loop continues ([#L1125-L1136](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1125-L1136)) — liveness recovers, but that event is never retried, and scroll/input dedup advanced before the callback ran, so it's never resent; `MutationBuffer.emit()` likewise clears its arrays before calling `mutationCb`, and both `commitMutationBuffers` call sites discard the boolean it returns. The throwing-consumer test asserts only that a *later* mutation arrives, never that the `boom` event does. All of which makes the changeset's "every event observed while it is in flight is held and delivered after the FullSnapshot in observation order — nothing is dropped" false twice over, and that sentence is the release note support will answer questions like ours from.

**major: two checkout runaways remain.** The `checkoutEveryNms` fix is real for the case it addresses, but `lastFullSnapshotWallTime` initializes to `0` ([#L402](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L402)) and is only set when a FullSnapshot actually reaches the consumer. If retry and sync fallback both fail, or with `recordDOM: false` (where `takeFullSnapshot` returns `true` without emitting), the marker stays 0 and `e.timestamp - 0 > checkoutEveryNms` holds for every subsequent incremental event — a Meta-plus-walk storm with no snapshot ever landing. On `main` this was a loud `TypeError`; now it's a quiet loop. Second, the count path is untouched: `incrementalSnapshotCount` has no time component and held events replayed during the flush increment it ([#L560-L567](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L560-L567)), so a burst can trip `exceedCount` from inside the flush, coalesce, and fire again with zero enforced gap the instant `budgetedSnapshotInFlight` clears. Your own test demonstrates it — "coalesces checkouts triggered by held events during the flush" drives six clicks at `checkoutEveryNth: 3` and asserts `fullSnapshots >= 2` ([convergence#L1000-L1023](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/test/budgeted-snapshot-convergence.test.ts#L1000-L1023)) — framed as intended coalescing rather than what it becomes under sustained load. Guard `exceedTime` on `lastFullSnapshotWallTime !== 0`, and don't count held-window replays toward `incrementalSnapshotCount`.

**major: the pending-id fix is narrower than the matrix reads.** Top-document CSSOM sequencing is genuinely fixed, but `onStylesheetTextSerialized` is missing from both recursive option bags — the generic `bypassOptions` ([snapshot.ts#L1299-L1327](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L1299-L1327)) and the hand-built same-origin iframe bag ([#L1371-L1399](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/snapshot.ts#L1371-L1399)). `onSerialize` still seeds a `styleTargets` entry for every iframe `<style>` with `inlined: false` (an author-text `<style>` never carries `_cssText` on the element) and nothing corrects it, so `heldCssomDeltaCoveredBySnapshot` returns false for iframe sheets unconditionally ([index.ts#L924-L959](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L924-L959)). The failure is double application rather than a drop: a delta that landed *before* a successful live-CSSOM read is already baked into the snapshot text and gets replayed on top of it, shifting every later rule index. Second, command-mode canvas (`sampling: 'all'`) appends to `pendingCanvasMutations` regardless of the lock ([canvas-manager.ts#L155-L170](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/observers/canvas/canvas-manager.ts#L155-L170)) — only *emission* is gated — so if the walker reaches a 2D canvas after those commands ran, `rr_dataURL` bakes in the resulting pixels and the same commands flush again after unlock; on abort, `discardPending()` clears them with no bitmap fallback for masked or WebGL canvases, which carry no `rr_dataURL` at all. The one new canvas scenario declares pixel replay out of scope and asserts `>= 1` source-9 event, so neither is detectable. Third, the overflow tail still bottoms out in `takeFullSnapshotSynchronous` after two consecutive aborts ([#L1039-L1076](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1039-L1076)) — made rare and instrumented, which I accept, but not removed.

**major: three live holes in the scrub.** First, a newly created cross-origin iframe's attach mutation is correctly deferred (its `parentId` is a parent-mirror reservation), but the frame's *subsequent* child events carry ids minted by `CrossOriginIframeMirror`'s own counter, which never enter `pendingReservedIds` — so `scrubUnclaimedIds` can't flag them and they go out immediately, ahead of both the commit and the deferred attach. The replayer sees an input for a node that doesn't exist yet and drops it. That wants a transitive dependency tag — defer a frame's whole stream behind its attach — not a per-event id check. Second, `pauseReservationHandout()` documents itself as keeping existing reservations claimable, but `getId` gates the whole reserved-id branch on `reserveNextId` being truthy ([utils.ts#L219-L241](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb-snapshot/src/utils.ts#L219-L241)), so while paused it returns `-1` for a node that already holds one — and nested events during the flush bypass the hold gate and go straight out with `id: -1`. Returning `reservedIds.get(n)` before the `reserveNextId` check makes the code match its own comment. Third, the Custom branch reads `data.payload.id` for *every* custom event without inspecting the tag ([index.ts#L219-L235](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L219-L235)) — that special case exists for the internal fullscreen event, but `addCustomEvent('checkout', { id: 42, total: 100 })` is dropped whenever 42 happens to be an unclaimed reservation. Mirror ids are small monotonic integers, the same domain as order and user ids, so this is data-dependent loss that looks random across sessions. Restrict DOM-id scrubbing to registered internal schemas.

**major: resetting the shadow-DOM manager opens a scroll blind window.** Every checkout calls `shadowDomManager.init()` → `reset()` before walking, disconnecting each existing shadow root's scroll listener; `addShadowRoot` only reinstalls it when `onSerialize` reaches that host, which on a large walk can be seconds later. Scroll doesn't cross a shadow boundary, which is why that listener exists, and reinstalling emits no catch-up event. The snapshot does capture a live `rr_scrollTop` when the walker arrives, so end state converges; what's lost is the incremental trail — a probe against this head observed zero scroll events for a shadow-panel scroll inside that window.

**moderate: the caps still can't see the largest retainers, and the estimator is itself a hazard.** `pendingRecordCount()` counts records rather than bytes and omits `mapRemoves` and the canvas manager's `pendingCanvasMutations`; the walker's own stack (one `WalkItem` per child of every serialized node — ~150k live objects on the canonical target page) is in no cap at all; `shouldAbort` is reachable only after a yield, so it's dead for the whole of `flushSync` and on a hidden page. The commit is still one unsplittable synchronous emit — `splitBuffer` only splits an array of multiple top-level events, never the contents of one giant mutation event. And `estimateRetainedSize` ([#L161-L203](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L161-L203)) runs synchronously on the main thread for every held event with a ceiling up to 16 MiB, falls through to `Object.keys` — so `Map`, `Set`, `Blob` and cross-realm `ArrayBuffer` count as ~0 bytes — and invokes arbitrary enumerable getters, unguarded, from inside `wrappedEmit`.

**moderate: the canary can't be read from ingested data.** `emitBudgetedSnapshotDiagnostic` has exactly three call sites ([#L1044](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1044), [#L1062](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1062), [#L1079](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1079)), all in the failure branch. A successful walk emits nothing — no duration, no slice count, no held-queue high-water — so the "longest single task" number we'd compare against our current traces isn't computable from ingested data. The diagnostic also rides the Custom channel the SDK suppresses while idle, `grep budgeted-full-snapshot` across the test tree returns nothing so it can regress silently, and the option is client-init-only: a bad canary needs a customer redeploy rather than a remote flag.

Shorter ones, listed rather than argued — each reproduces at head:

- A reentrant `record.takeFullSnapshot()` from inside a buffer's `commit()` emits Meta at [#L872](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L872) and *then* fails to lock: an orphan Meta with no FullSnapshot behind it, at budget 0.
- `record()`'s outer catch bumps the generation but never calls `mirror.endIdReservation()` ([#L1686-L1692](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1686-L1692)), so a throwing `observe(document)` leaves the mirror in reservation mode for the life of the page.
- `stopRecording()` nulls `activeBudgetedSnapshot` without clearing `budgetedSnapshotInFlight` ([#L1596-L1608](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1596-L1608)) — exactly the state of the one silent, uncounted drop branch at [#L480-L483](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L480-L483).
- `canvasManager.onFullSnapshot()` and `adoptStyleSheets` sit after the inner `finally` but inside the outer `try` ([#L1185-L1190](https://github.com/PostHog/posthog-js/blob/63d47a21aeb0f0a836f3258adbb5dcce67345e24/packages/rrweb/rrweb/src/record/index.ts#L1185-L1190)), so a throw from the unguarded `reattachIframes()` above skips both: canvas dedup never resets, constructed-stylesheet documents replay unstyled. Same shape on the sync path via `finishFullSnapshot()`.
- `commitMutationBuffers`'s 4-sweep cap force-discards late-born buffers and returns `false`; all three call sites ignore the return value.
- `record.freezePage()` during a walk makes `commit()` a no-op that still returns `true`, after the held window has already gone out referencing ids those un-emitted adds were to introduce.
- A mixed-validity multi-range Selection is dropped whole (`ranges.some(...)`) while pointer `positions` one block above are filtered individually.
- `isIgnored` calls `getId` on every mutation target, minting reservations during a walk for nodes no event will reference.
- Held payloads are stored by reference and `eventProcessor`/`packFn` don't run until flush, so mutating a payload after `addCustomEvent` records a different value in budgeted mode than in sync mode — and the byte cap measures the pre-processed value.
- The held gate excludes only Meta/FullSnapshot, so SDK control events (`$session_id_change`, `$session_ending`, console logs) can arrive up to 30s late. The idle/not-started branch handles session rotation; the active branch doesn't.
- `scrubUnclaimedIds`'s docstring still claims it "Covers every id-bearing payload shape" while `removes[]`, `texts[]` and `attributes[]` are never inspected. Latent today, but it's the kind of comment the next author builds on.

A few things I'd rather ask than assert:

1. **Is `MAX_WALK_WALL_CLOCK_MS = 30_000` measured or a placeholder, and can it be remote-controlled?** For 30 seconds the buffers are locked, every event is held and the mirror is in reservation mode — a large blast radius for a safety limit, sitting inside a periodic-snapshot cadence that can be as tight as 60s. With the option client-init-only, a misbehaving canary has no lever short of a redeploy on our side.
2. **Does #4320's stylesheet budget and telemetry get wired into this path, and before or after a canary?** Same question as last round, and I don't think anything landed: the 16-node probe stride still lets one `serializeNodeWithId` overshoot without bound, and the same-origin iframe block still recursively serializes a whole child document as one unyielding task. The changeset says so out loud now, which is right — but for our pages stylesheet cost *is* the freeze, so this stays the difference between "fixed" and "shortened."
3. **`isInputPending`/`minSliceMs` run in zero tests.** `snapshot.ts` disables input-pending checks whenever a custom `yieldFn` is supplied and every deterministic test supplies one; the single production-yielder scenario never simulates pending input. That floor exists to stop continuous input stretching a walk unboundedly, which is upstream of the tab-switch and abort behavior above.
4. **What's the supported contract for plugin payloads carrying mirror ids?** Plugins get the mirrors, and `scrubUnclaimedIds` returns any non-Custom, non-Incremental event unchanged — so a Plugin event is both unscrubbed and emitted in the pre-commit pass, ahead of the ids the commit is about to introduce. Is the contract "plugins must not emit mirror ids during a sliced snapshot," or does this want a schema hook? No convergence scenario installs a plugin.
5. **What's the validation bar you're merging against?** Several frontier preconditions are still bare `setTimeout` waits rather than `mirror.hasNode()` polling, with `>=`/`some()` assertions; the walker has no quirks-mode or XML coverage and no non-Chromium run; the perf evidence is one 35k-node table fixture, one run per arm, with no per-slice p50/p95/max, no stylesheet-heavy fixture, no heap/GC or queue high-water. I'm not asking for all of it before merge — I'm asking which of it you consider load-bearing before a canary, so I know what our own testing has to cover.

Last, the ask that matters most from our side, unchanged in substance from #4233 but pointed at this PR: a build we can test. We still have the deterministic repro — live pages where the freeze lands mid-scroll on a trace — and replay stays off across our fleet until we can prove this out on them. Concretely: a pre-release off this branch (canary dist-tag is fine) once the two blockers land, ideally with #4320; the config you'd recommend for our shape of page, in particular a budget value you consider sane where stylesheet cost dominates; and enough success-path telemetry to confirm in production what the longest single recorder task actually became, rather than inferring it from the absence of failure diagnostics. Happy to run it against the real pages before this goes wide — we're the reason it exists, and I'd rather find the next one on our fleet than in your inbox.

PR #4387, opened 2026-08-02
fix: slim no-external extension bundle types

NVolcz:
## Problem

[#4305](https://github.com/PostHog/posthog-js/pull/4305) fixed the runtime ABI between `module.slim.no-external.js` and `extension-bundles.js`, but the corresponding TypeScript declarations remain incompatible.

`extension-bundles.d.ts` references types from `module.slim.d.ts`, while `module.slim.no-external.d.ts` contains an independently emitted copy of the same classes. TypeScript treats classes with private members nominally, so extension constructors from `extension-bundles` cannot be assigned to `PostHogConfig['__extensionClasses']` from `module.slim.no-external`. Consumers see errors such as:

```text
ScrollManager types have separate declarations of private property '_instance'
```

## Changes

- Preserve `module.slim`'s declaration-level re-export from `module.slim.no-external` instead of allowing `rollup-plugin-dts` to inline a second declaration graph.
- Rewrite the preserved `.es` declaration path to the published `./module.slim.no-external` path.
- Express the default export as a source-level re-export so the generated declaration remains a reference rather than pulling the nominal types back into `module.slim.d.ts`.
- Add a regression test that compiles the built declarations and verifies both extension constructor compatibility and default-export compatibility across the two slim entrypoints.

The generated `module.slim.d.ts` is now a pure re-export:

```ts
export * from './module.slim.no-external';
export { default } from './module.slim.no-external';
```

The runtime bundle remains unchanged in behavior and retains the `external-scripts-loader` side effect.

Verification:

- `pnpm turbo run build --filter=posthog-js`
- `pnpm --filter posthog-js test:unit -- --runInBand` — 5,122 passed
- Focused declaration regression test — 5 passed
- `pnpm --filter posthog-js lint`
- Prettier and `git diff --check`
- A real Vue consumer using `module.slim.no-external` and `extension-bundles` passes `vue-tsc --noEmit` with its declaration workaround removed

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi Coding Agent with OpenAI GPT-5.6 implemented and tested this change under human direction. Claude Code with Opus 5 independently reviewed the code changes. The implementation keeps `module.slim.no-external` as the canonical declaration graph and makes `module.slim` reference it, preserving compatibility for consumers of either entrypoint without adding a new public artifact. No public session links are available.

### DIFF
--- a/.changeset/fix-slim-no-external-extension-types.md
+++ b/.changeset/fix-slim-no-external-extension-types.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Share extension bundle types between the slim and slim no-external entrypoints.
--- a/packages/browser/rollup.config.mjs
+++ b/packages/browser/rollup.config.mjs
@@ -418,6 +418,7 @@ const typeTargets = entrypoints
     .map((file) => {
         const source = `./lib/src/entrypoints/${file.replace('.ts', '.d.ts')}`
         const isExtensionBundles = file === 'extension-bundles.es.ts'
+        const isSlimModule = file === 'module.slim.es.ts'
         // customizations types must reference the main module for the PostHog class —
         // an inlined duplicate would be nominally incompatible with the consumer's
         // `posthog` instance (same private-fields problem as extension-bundles).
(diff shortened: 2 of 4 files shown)


marandaneto (approved):

marandaneto:
thanks @NVolcz

PR #4399, opened 2026-08-03
fix(conversations): keep widget list markers under host CSS resets

christiaan-ph:
## Problem

Bullet and numbered lists sent from PostHog Conversations render in the support widget without their markers: the `<ul>`/`<ol>`/`<li>` structure is intact, but no bullets or numbers show, so list items read as plain indented lines.

The widget renders into the host page's DOM (no shadow DOM) and relies on inline styles to resist host CSS. The list elements in `RichContent.tsx` only inline `margin` and `padding-left`, so any host page with an aggressive CSS reset (for example Tailwind preflight's `ol, ul { list-style: none }`) strips the markers. This was latent since list rendering was added in #3000 and became visible once PostHog/posthog#73165 started delivering `bulletList`/`orderedList` nodes in `rich_content`.

## Changes

- Inline `listStyleType: 'disc'` / `'decimal'` and `listStylePosition: 'outside'` on the widget's `<ul>`/`<ol>`, so the markers survive host page resets that target `ol, ul`.
- Inline `listStyleType: 'inherit'` on `<li>` so resets that target `li` directly can't strip the marker either.
- Extended the existing bullet and ordered list tests to assert the inline `list-style-type` is present, which guards against this exact regression.

Verified with `pnpm jest src/__tests__/extensions/conversations/rich-content.test.tsx` in `packages/browser` (148 passed). Written and tested without a live widget embed; the DOM-level root cause was confirmed against a Tailwind host page in devtools before the change.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### DIFF
--- a/.changeset/widget-list-markers.md
+++ b/.changeset/widget-list-markers.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Conversations widget: bullet and numbered lists in a support reply now keep their markers on host pages with an aggressive CSS reset (for example Tailwind preflight's `ol, ul { list-style: none }`). The widget renders into the host page's DOM, so the list style is now set inline on `<ul>`, `<ol>`, and `<li>` rather than left to the page's own styles.
--- a/packages/browser/src/__tests__/extensions/conversations/rich-content.test.tsx
+++ b/packages/browser/src/__tests__/extensions/conversations/rich-content.test.tsx
@@ -715,7 +715,9 @@ describe('RichContent', () => {
 
             const { container } = render(<RichContent {...defaultProps} richContent={doc} />)
 
-            expect(container.querySelector('ul')).not.toBeNull()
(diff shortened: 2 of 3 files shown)


marandaneto (approved):

ioannisj:
This probably needs a changeset?

christiaan-ph:
@ioannisj added a changeset!

PR #4400, opened 2026-08-03
fix(browser): avoid relogging handled transport failures

marandaneto:
## Problem

The browser request layer already classifies expected failures that never receive an HTTP response (`statusCode: 0`) at warning severity. Several product consumers then logged those same handled failures at error severity, which could turn offline, blocked, CORS, timeout, and teardown failures into captured exceptions and duplicate logs.

## Changes

- Apply consistent status-zero logging ownership to Surveys, Product Tours, and Remote Config: request-layer failures are not re-logged, while bare XHR/synthetic status-zero responses receive one warning.
- Generalize Conversations handled errors across messages, mark-read, tickets, and restore flows so manager and UI catches do not duplicate known failures.
- Keep Conversations HTTP, malformed-response, unexpected-error, and existing rate-limit severities visible.
- Brand browser Logs status-zero failures so core retry/backoff continues without re-promoting them to errors; preserve error severity for 429, 4xx, 5xx, and other fatal failures.
- Add regression coverage for fetch-style and bare status-zero failures, HTTP failures, explicit/background Logs flushing, and Conversations UI catches.
- Update the generated Terser mangled-property list for the new private Logs helpers.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent, an independent reviewer subagent, and the isolated autoreview tool. The main design decision was to keep classification at browser product ownership boundaries instead of expanding public request response types. Logs handling remains browser-local so React Native and Node behavior is unaffected. The work follows the request-layer rule established by #4334: known status-zero failures have one logging owner, while HTTP and unexpected failures remain errors.


packages/browser/src/extensions/conversations/external/index.tsx
```diff
@@ -507,23 +514,20 @@ export class ConversationsManager implements ConversationsManagerInterface {
...
-
                     if (response.statusCode !== 200) {
-                        const errorMsg =
-                            response.json?.error ||
-                            response.json?.detail ||
-                            response.json?.message ||
-                            'Failed to restore conversations'
-                        reject(new Error(errorMsg))
+                        reject(
+                            createConversationsRequestError(
```
lucasheriques, line 532: the factory logs at `error` before it brands, so the `isConversationsError` guards at :490 and :444 end up suppressing the warns and keeping the error. i ran a 500 here against the existing `conversations-manager` harness:

```
main:        logger.error 0   logger.warn 1   ("Restore token exchange failed, retrying once")
this branch: logger.error 2   logger.warn 0   ("Failed to restore conversations" x2)
```

two things fall out. the path gets louder rather than quieter, so with `capture_console_errors: true` a failed restore now sends 2 captured exceptions where it sent 0. and `"retrying once"` is gone, so the two lines are identical and you cannot tell the retry from the first attempt.

moving the logging up into `_restoreFromTokenWithRetry` would give the retry one logging owner and keep that distinction.

packages/browser/src/posthog-surveys.ts
```diff
@@ -280,7 +280,13 @@ export class PostHogSurveys implements Extension {
                 const statusCode = response.statusCode
                 if (statusCode !== 200 || !response.json) {
                     const error = `Surveys API could not be loaded, status: ${statusCode}`
-                    logger.error(error)
+                    if (statusCode === 0) {
```
lucasheriques: nit: this reads easier flattened, since the inner `if` has a silent do-nothing path:

```ts
if (statusCode !== 0) {
    logger.error(error)
} else if (!response.error) {
    logger.warn(error)
}
```

`remote-config.ts:120` has the same shape if you want them consistent.

packages/browser/src/posthog-logs.ts
```diff
@@ -196,10 +220,16 @@ export class PostHogLogs implements Extension {
...
-            void this._core.flush().catch((err) => this._logger.error('PostHog logs flush failed:', err))
+            void this._core.flush().catch((err) => this._logFlushError(err))
         }
         if (this._consoleCore) {
-            void this._consoleCore.flush().catch((err) => this._logger.error('PostHog logs flush failed:', err))
+            void this._consoleCore.flush().catch((err) => this._logFlushError(err))
+        }
+    }
+
+    private _logFlushError(error: unknown): void {
```
lucasheriques, line 230: nit: the log-it side of `if (!isHandledLogsRequestError(error))` has no coverage. the only two `flushLogs()` tests are the empty-buffer one and the new one asserting nothing is logged, so flipping this to `if (isHandledLogsRequestError(error))` would keep the suite green.

one test with an unbranded `{statusCode: 500}` asserting `'PostHog logs flush failed:'` does show up would pin it.

lucasheriques (approved):
the brand-and-suppress mechanism is solid, and i like that unbranded throws, wrapped errors and non-`Error` throwables all still log. requiring an own `=== true` marker on the predicate is the right call.

one thing on the restore path that i think works against the goal of the pr, left inline on the reject. two nits below that, take or leave them.

packages/browser/src/extensions/conversations/external/index.tsx
marandaneto: Good catch! Fixed in 7c08996b3 — restore request errors are now classified without being logged by the factory, while the retry path owns a single retry-specific warning. Request-layer-handled network failures remain suppressed, and tests cover HTTP, rate-limit, invalid, and bare status-zero retries.

packages/browser/src/posthog-surveys.ts
marandaneto: Updated in 7c08996b3 — flattened the status-zero branch as suggested.

packages/browser/src/posthog-logs.ts
marandaneto: Added in 7c08996b3 — an explicit-flush test now verifies that an unbranded `{ statusCode: 500 }` failure still logs `PostHog logs flush failed:`.

PR #4402, opened 2026-08-03
fix: publish declarations for browser entrypoints

NVolcz:
## Problem

`posthog-js` publishes JavaScript entrypoints such as `dist/posthog-recorder.js`, `dist/exception-autocapture.js`, and `dist/web-vitals.js`, but it does not publish matching declarations at those public paths. Although TypeScript already emits declarations for their source entrypoints under `lib/src/entrypoints`, Rollup only moves declarations for `.es.ts` entrypoints into `dist`.

As a result, consumers get a missing declaration error for valid imports such as:

```ts
await import('posthog-js/dist/posthog-recorder')
```

and need to suppress it with `@ts-expect-error`.

## Changes

- Generate a root declaration for every TypeScript entrypoint that has a published JavaScript output in `dist`.
- Normalize format-specific source names so, for example, `module.es.ts` still produces `module.d.ts` and `main.cjs.ts` produces `main.d.ts`.
- Add a package-output regression test deriving the expected declarations from `src/entrypoints`, avoiding stale or code-split files in `dist`.
- Inline recorder types so the declarations do not require the development-only `@posthog/rrweb-record` package.
- Reference the canonical `PostHog` declaration from extension entrypoints instead of inlining nominally incompatible copies.
- Reduce total published declaration output by approximately 155 KB uncompressed (33 KB compressed) by deduplicating the canonical `PostHog` type; runtime JavaScript bundles are unaffected.
- Add TypeScript fixtures that resolve every generated entrypoint without ambient Node types, test the recorder in isolation, and compose extension APIs with the main `posthog` instance.
- Avoid leaking the `NodeJS.Timeout` type through the public surveys declaration.

### Alternatives and follow-ups

Adding a package `exports` map was considered as a way to distinguish supported imports from files that only need to be present for CDN or internal loading. However, introducing one would require defining every supported subpath and could break existing deep imports, so it would be better handled as a separate compatibility-focused change.

Because declarations are generated for every shipped entrypoint, this also covers paths that consumers probably do not need to import directly, such as array variants, `main`, and `external-scripts-loader`. There is currently no canonical list separating public extension entrypoints from these files, while `package.json` publishes all of `dist/*`. This PR therefore follows the mechanically enforceable rule that every published JavaScript entrypoint receives a declaration rather than introducing another allowlist.

Most side-effect-only entrypoints already produce 13-byte `export {}` declarations, while larger declarations now reference the canonical `PostHog` type rather than inlining the main declaration graph. Maintaining a narrower set of typed entrypoints or generating lightweight declarations that forward to the existing `lib` declaration tree was also considered, but those approaches add maintenance or declaration-graph compatibility concerns while the overall declaration output already decreases.

Consumers that currently suppress the missing-declaration diagnostic with `@ts-expect-error` will need to remove that directive after upgrading, because TypeScript can report it as unused once the declaration is available.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with OpenAI ChatGPT 5.6 using Pi Coding Agent. The scope and alternatives were reviewed with human direction, including how to handle the package's existing deep-import surface and whether declarations should preserve entrypoint exports or only provide module-resolution shims.

### DIFF
--- a/.changeset/browser-entrypoint-declarations.md
+++ b/.changeset/browser-entrypoint-declarations.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Publish TypeScript declarations for browser extension entrypoints under their public `dist` paths.
--- a/packages/browser/rollup.config.mjs
+++ b/packages/browser/rollup.config.mjs
@@ -411,9 +411,25 @@ const entrypointTargets = entrypoints.map((file) => {
 // consumers don't need a runtime dep on the re-exported package to resolve them.
 const inlineExternalTypesEntries = new Set([
     'extension-bundles.es.ts',
+    'posthog-recorder.ts',
+    'recorder.ts',
     'rrweb.es.ts',
     'rrweb-types.es.ts',
     'rrweb-plugin-console-record.es.ts',
+])
+
+// Entries whose declarations use PostHog must reference the canonical class in
+// module.d.ts instead of inlining a nominally incompatible copy.
+const mainModuleTypesEntries = new Set([
+    'conversations.ts',
+    'customizations.es.ts',
+    'dead-clicks-autocapture.ts',
+    'main.cjs.ts',
+    'product-tours-preview.es.ts',
+    'product-tours.ts',
+    'surveys-preview.es.ts',
+    'surveys.ts',
+    'tracing-headers.ts',
 ])
 
 // rrdom's dts drops the local `RRNodeType` alias declaration; the renderChunk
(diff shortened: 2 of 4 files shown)


marandaneto (approved):

marandaneto:
thanks @NVolcz

Issue #4202, opened 2026-07-20
Replay full snapshot remains on previous session after idle rotation in 1.404.1


tornikeo:
### Bug description

A long-lived background tab using `posthog-js` 1.404.1 rotated its session because `changeReason.activityTimeout` was true. The analytics session changed, but session replay never uploaded a playable full snapshot for the new session. The preceding session is playable.

This looks like a regression or incomplete edge case in the same failure family as #4104.

### Production diagnostics

All identifiers below are intentionally redacted; I can share the exact EU Cloud session privately if useful.

- SDK event `$lib_version`: `1.404.1`
- Browser: Chrome 152 on macOS
- `$session_recording_start_reason`: `session_id_changed`
- `$recording_status`: `active`
- `$sdk_debug_rrweb_start_attempted`: `true`
- `$sdk_debug_rrweb_attached`: `true`
- `$sdk_debug_replay_rrweb_error`: `false`
- `$has_recording`: `null`
- New-session replay buffer grew from 1 event / 465 bytes to 47 events / 164474 bytes
- New-session flushed size ultimately remained `0`
- `$sdk_debug_replay_full_snapshots` continued to list only the previous session ID, including a periodic full snapshot emitted after the analytics session had rotated
- The only replay blob ingested for the new session was a `$session_starting` lifecycle event; there was no rrweb FullSnapshot
- Normal analytics events continued to ingest for the new session for more than an hour
- The preceding session remained fully playable

The recorder script was loaded through PostHog Cloud EU. The proxy-served asset currently matches both the official EU asset and immutable `/static/1.404.1/recorder.js` byte-for-byte.

### Expected behavior

After an activity-timeout session rotation, a full snapshot should be associated with and flushed under the new session ID before incremental replay data.

### Additional context

The tab was idle/backgrounded and continued making an application live-stream request. Capturing a network failure event caused the analytics session rotation; there was no navigation or reload. This exact edge is operationally important because `get_session_replay_url()` returns a URL for the new session even though it is unplayable.

PR #4403, opened 2026-08-03
fix(replay): hold recordings for sessions with no user interaction

ksvat:
## Problem

#4224 fixed a real bug (#4202: sessions rotated in never-interacted tabs kept stamping snapshots with the stale session id and never flushed, so the new session was unplayable) but overshot: treating the `'unknown'` idle state as "not idle" made every activity-timeout rotation in a parked tab ship a Meta + FullSnapshot recording. A tab nobody touches now produces an unbounded chain of zero-interaction recordings, each billed as a full recording (`recording_count_in_period` counts them). Fleet-wide, zero-duration/zero-interaction recordings roughly tripled starting July 23; for idle-tab-heavy products, total recordings ~3x'd with no change on the customer side (Slack: https://posthog.slack.com/archives/C06UTAJJEHM/p1785770287662349).

Neither the pre-#4224 behavior ("never ship them", by accident of `_isIdle` truthiness) nor the post-#4224 behavior ("always ship them") was a deliberate policy. This PR makes the policy explicit.

Separately, confirmed-idle tabs skipped the session check entirely, so an idle session never rotated: we observed single recordings spanning 25-95 hours, sailing through `SESSION_LENGTH_LIMIT`.

## Changes

**Capture speculatively, ship on interaction.** While `_isIdle === 'unknown'` (no user interaction seen this recorder run):

- the recorder keeps capturing (so the rotated session's Meta + FullSnapshot exists, preserving the #4202 fix), but `_flushBuffer` holds instead of shipping
- the held buffer is pruned to the newest playable prefix (latest FullSnapshot, its Meta, and everything after), so an indefinitely parked tab holds bounded memory
- the first user interaction flips `_isIdle` to `false` and ships the held buffer, so the recording is playable from before the interaction
- `beforeunload` still force-ships (the long-standing one-bounded-recording-per-pageview behavior)
- a rotation's `stop()` finds the buffer held and clears it: a session nobody ever touched ships nothing

**Idle sessions respect the 24-hour cap.** `_updateWindowAndSessionIds` no longer early-returns while confirmed idle; the readOnly session check enforces `sessionPastMaximumLength` and hears cross-tab rotations, so idle sessions rotate instead of accreting multi-day recordings under one id.

**Behavioral decisions embedded here that reviewers should explicitly confirm:**

1. URL trigger activation no longer ships an uninteracted session's buffer (URL triggers scope where recording is allowed; they aren't evidence anyone cares). **Event triggers DO release the hold** (record-on-exception from an untouched tab is exactly what that config asks for), including when the activation itself short-circuits because a URL trigger already satisfied triggerMatchType 'any'. V2 trigger groups activate inside the strategy and don't set the evidence flag yet - known follow-up.
2. "Interaction" = rrweb ACTIVE_SOURCES (includes mouse move, scroll, touch), so passive-but-present viewers still record.
3. `stop()` discards a held buffer (including manual `stopRecording()` on a never-touched page).

Regression tests: unit tests assert a rotation chain in an untouched tab ships nothing, the held buffer prunes instead of flushing on the size cap, the hold releases on first interaction, and a confirmed-idle session rotates at the 24h cap; the #4224 Playwright spec now asserts hold-then-ship end to end in both compression modes.

Not run locally (env): Playwright specs, eslint, typecheck — relying on CI for those.

### Libraries affected

- [x] posthog-js (web)

## 🤖 Agent context

Drafted from the HeyGen replay-billing investigation (thread above): recordings 3x'd on July 23 with no customer-side change; root cause traced to #4224 shipping through the unpinned `posthog-recorder` lazy bundle, reaching SDKs pinned as far back as 1.390.0 within an hour of release.

### DIFF
--- a/.changeset/replay-hold-recordings-until-interaction.md
+++ b/.changeset/replay-hold-recordings-until-interaction.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Session replay no longer ships recordings for sessions that have seen no user interaction. A recorder in a tab nobody has touched (the 'unknown' idle state) now holds its buffer, bounded to the newest playable snapshot prefix, and ships it only on the first user interaction, on page unload, or not at all if the session rotates away untouched. This stops activity-timeout rotations in parked tabs from producing unbounded chains of zero-interaction billable recordings, while keeping rotated sessions playable from before the user's first interaction. Confirmed-idle tabs also now keep running the session check, so an idle session rotates at the 24-hour session cap instead of accreting a multi-day recording.
--- a/packages/browser/jest.config.js
+++ b/packages/browser/jest.config.js
@@ -25,6 +25,7 @@ module.exports = {
         '^preact/jsx-runtime$': path.join(preactRoot, 'jsx-runtime/dist/jsxRuntime.js'),
         '^preact/test-utils$': path.join(preactRoot, 'test-utils/dist/testUtils.js'),
         '^@testing-library/preact$': path.join(testingLibraryPreactCjs, 'index.js'),
+        '^@posthog/browser-common$': '<rootDir>/../browser-common/src/index.ts',
         '^@posthog/browser-common/config$': '<rootDir>/../browser-common/src/config.ts',
         '^@posthog/browser-common/constants$': '<rootDir>/../browser-common/src/constants.ts',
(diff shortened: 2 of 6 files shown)


ksvat:
@/private/tmp/claude-501/-Users-kimdugan-posthog/14c68ed4-3c5c-4578-ba3b-495a1fde4895/scratchpad/close4403.md

PR #4415, opened 2026-08-04
feat(react-native): add push notification support

ioannisj:
## Problem

PostHog Workflows can send push notifications to iOS and Android apps but not to React Native ones. The native SDKs already do all the hard parts (subscription state, retry and backoff, dedup, persistence, the HTTP calls) and `@posthog/react-native-plugin` already wraps both of them. Push was just never bridged, so a React Native app had no way to register a device token or report an opened notification.

## Changes

A thin wrapper over the native SDKs. No subscription state or delivery logic moves into JS.

In `@posthog/react-native-plugin` (Kotlin, Swift, ObjC++):

- `registerPushNotificationToken(deviceToken, appId)`, `unregisterPushNotificationToken()` and `capturePushNotificationOpened(properties)`, each a one-line forward to the native SDK.
- `reset(distinctId, anonymousId)`, so a logout unbinds the token from the old identity.
- `setPushIdentityProvider(provider)`. A callback can't cross the bridge, so native asks for a token over an event and JS replies through a bridge method keyed by request id. Native already runs a 10s watchdog and sends the request unauthenticated if nothing comes back, so there's no JS-side timeout to get wrong.

In `posthog-react-native`:

- The three push methods on `PostHog`, plus the `pushIdentityProvider` option.
- The native plugin now initializes when push alone is enabled. Previously only session replay or native error tracking would initialize it.
- `reset()` propagates to native, which unregisters the logged-out user's subscription and re-registers under the new anonymous id.
- Opt-out is respected in all three directions: an opted-out user registers no token, the opt-out state is forwarded into the native config (native defaults to opted-in and would otherwise keep registering), and `optOut()` unregisters a subscription that was already registered. Without that last one a token registered before consent withdrawal stays live, because the native handler keeps its own persisted record and retry loop and never sees the JS opt-out.

`capturePushNotificationSubscriptions` and `capturePushNotificationOpened` both default to `true`, matching the native SDKs. Setting either to `false` opts out. That default is deliberate, but it does mean an app that already has the plugin installed for session replay starts registering device tokens once it upgrades. The alternative was defaulting both to `false` in React Native only, which trades the silent upgrade for a documented divergence from the native SDKs, and I'd rather keep parity. Worth a second opinion though.

One known gap I've left for a follow-up: on iOS, `reset()` binds the push subscription to the wrong identity. `PostHogSDK.shared.reset()` re-registers push synchronously before it returns (`PostHogPushSubscriptionHandler.swift:399` snapshots the distinct id at call time), but the plugin can only write the JS `anonymousId` after `reset()` has returned, so the subscription lands under the anonymous id native minted for itself. Fixing it needs a way to seed the new anonymous id into `posthog-ios`'s `reset()`, so it belongs in a PR against that repo rather than this one.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -153,6 +163,25 @@ class PosthogReactNativePluginModule(
...
 
+              // Only set when present: the legacy start() path predates push, and there the
+              // native defaults (both true) must win, matching posthog-android on its own.
+              if (hasKey(pushConfig, "capturePushNotificationSubscriptions")) {
+                capturePushNotificationSubscriptions =
+                  getBoolean(pushConfig, "capturePushNotificationSubscriptions", true)
+              }
+              if (hasKey(pushConfig, "capturePushNotificationOpened")) {
+                capturePushNotificationOpened = getBoolean(pushConfig, "capturePushNotificationOpened", true)
+              }
```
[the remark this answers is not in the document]
ioannisj: Intended: the RN layer defaulting these on matches posthog-ios and posthog-android, which both ship `capturePushNotificationSubscriptions` and `capturePushNotificationOpened` as `true`, so forwarding them unchanged keeps the hybrid SDK on native parity instead of silently overriding it. The migration note this finding suggests is already in the changeset ("Both are on by default; opt out with `capturePushNotificationSubscriptions: false` or `capturePushNotificationOpened: false`"). Keeping as is.

packages/react-native/src/posthog-rn.ts
```diff
@@ -186,6 +187,63 @@ export interface PostHogOptions extends PostHogCoreOptions {
...
+   * send the request without an identity token.
+   *
+   * The token must be minted by your backend (HS256, with `sub` = distinctId,
+   * an `app_id` claim, and `aud` = `posthog:push_identity`), never in the app.
+   *
+   * The native SDKs cache the result per `(distinctId, appId)` and give the
+   * callback 10 seconds before falling back to an unauthenticated request, so
+   * a slow or throwing implementation degrades rather than blocking delivery.
+   */
+  pushIdentityProvider?: PostHogPushIdentityProvider
```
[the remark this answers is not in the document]
ioannisj: Same call as the sibling finding on the Kotlin side: default-on is deliberate parity with posthog-ios and posthog-android (both default these to `true`), it's documented as `@default true` on the options, and the changeset spells out the opt-out. An RN-only `false` default would make the hybrid SDK the odd one out, so keeping it.

.github/workflows/react-native-plugin-native-ci.yml
```diff
@@ -93,7 +93,7 @@ jobs:
 
   build-ios-static-lib:
     name: Build example for iOS (CocoaPods)
-    runs-on: macos-14-xlarge
+    runs-on: macos-26
```
ioannisj: Pods were failing on Xcode 16.4, see: https://posthog.slack.com/archives/C04MZFDA5KK/p1785915419494489

.github/workflows/library-ci.yml
```diff
@@ -140,26 +140,38 @@ jobs:
             exit 0
           fi
 
-          if ! jq -e '(.stats.unexpected | type == "number") and (.errors | type == "array")' compat-results.json > /dev/null; then
+          if ! node -e '
```
ioannisj, line 143: Bringing in some fixes here and in tests from https://github.com/PostHog/posthog-js/pull/4417. Should go away once that lands

packages/react-native/src/posthog-rn.ts
```diff
@@ -1966,14 +2080,229 @@ export class PostHog extends PostHogCore {
...
+   * manage subscriptions yourself.
+   *
+   * Not supported on web or macOS.
+   */
+  async unregisterPushNotificationToken(): Promise<void> {
+    return this._callPushPlugin(
+      'unregisterPushNotificationToken',
+      // No requireInit: if native never came up there is no subscription to remove, and
+      // booting it to delete nothing would defeat the opt-out this usually follows.
+      { allowMacOS: false, allowWhenOptedOut: true, requireInit: false, noop: 'nothing to unregister' },
```
marandaneto: blocking: This cleanup is silently skipped after a relaunch in manual-registration mode. With both capture flags disabled, a manual register forces native setup and persists the token on launch A. On launch B native intentionally is not initialized, so `_enqueueNative` returns at the `!requireInit` branch and unregister/reset/opt-out resolve without touching that persisted subscription. Workflows can therefore keep targeting a logged-out or opted-out device. Cleanup needs to initialize/access native persisted state rather than assuming no in-process initialization means no subscription.

packages/react-native/src/posthog-rn.ts
```diff
@@ -940,9 +1014,37 @@ export class PostHog extends PostHogCore {
...
     const result = super.optOut()
     void this._eventsStorage.waitForPersist()
+    // A device token registered before opt-out would otherwise survive consent withdrawal: the
+    // native subscription handler keeps its own persisted record and retry loop. unregister is
+    // deliberately allowed while opted out.
+    // Native gates unregister sends on its own opt-out flag, so the unregister has to reach
+    // native first; both share _nativeChain, so enqueueing it first is the ordering. A DELETE
+    // that still loses the race natively is persisted there and replayed after opt-in.
+    if (OptionalReactNativePlugin?.unregisterPushNotificationToken) {
+      void this.unregisterPushNotificationToken()
```
marandaneto, line 1031: blocking: The JS queue orders bridge dispatch, not the native DELETE. Android resolves this bridge call immediately after queueing `unregisterCurrent()` on its executor; `setOptOut(true)` can then set `config.optOut` before that executor runs, and `performUnregister` returns while opted out. iOS similarly re-checks consent after an asynchronous identity-token mint. The pending DELETE will not run for the rest of the opted-out period, leaving the server subscription active. Conversely, when deletion succeeds both native SDKs clear the stored token, while opt-in only flips consent/reinstalls a listener and does not refetch it, so the changeset's “re-arms it without an app restart” claim is not met. This needs a native consent transition that can delete while opted out and restore registration on opt-in.

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -153,6 +163,25 @@ class PosthogReactNativePluginModule(
...
               }
 
+              // Only set when present: the legacy start() path predates push, and there the
+              // native defaults (both true) must win, matching posthog-android on its own.
+              if (hasKey(pushConfig, "capturePushNotificationSubscriptions")) {
+                capturePushNotificationSubscriptions =
+                  getBoolean(pushConfig, "capturePushNotificationSubscriptions", true)
+              }
+              if (hasKey(pushConfig, "capturePushNotificationOpened")) {
+                capturePushNotificationOpened = getBoolean(pushConfig, "capturePushNotificationOpened", true)
```
marandaneto, line 174: blocking: Automatic Android open capture cannot observe the normal React Native cold start. `PostHogAndroid.setup` is called from JS after the launch Activity has already received `onActivityCreated`, while posthog-android 3.58.0 only registers `ActivityLifecycleCallbacks` during setup and inspects notification data from future `onActivityCreated` callbacks; it never inspects the current Activity/launch intent. Since the public docs delegate warm/foreground opens to the manual API and claim cold starts are automatic, the Android automatic path effectively captures none of the normal cases. The plugin needs to inspect the current launch intent or initialize before Activity creation.

packages/react-native/src/posthog-rn.ts
```diff
@@ -940,9 +1014,37 @@ export class PostHog extends PostHogCore {
...
+    this._propagateNativeOptOut()
     return result
   }
 
+  // Native persists its own consent flag and only reads the JS one at setup(), so runtime
+  // changes must cross the bridge or native's automatic push registration keeps running on
+  // its setup-time snapshot (e.g. re-registering an OS-refreshed token after optOut()).
+  private _propagateNativeOptOut(): void {
+    if (!OptionalReactNativePlugin?.setOptOut) {
+      // A plugin that predates push cannot auto-register tokens, so there is no consent to sync.
```
marandaneto, line 1042: blocking: This assumption is false for a dependency combination the package still supports. The peer range is `@posthog/react-native-plugin >= 2.2.0`, so 2.2.4 is valid; that plugin pins posthog-ios 3.69.0, whose push capture defaults are on, but its bridge ignores the new push config and exposes no `setOptOut`, unregister, or reset methods. The new RN layer can initialize that older setup for default-on push, then this branch skips consent propagation, so explicit push options and runtime opt-out can silently fail. Please bump the peer floor to the release containing this bridge or capability-gate push initialization.

packages/react-native/src/posthog-rn.ts
```diff
@@ -1679,15 +1789,19 @@ export class PostHog extends PostHogCore {
...
-      }
+    if (OptionalReactNativePlugin) {
+      // Queued, not inline: a same-tick reset() is already on _nativeChain and must land first,
+      // or its post-reset anonymous id gets overwritten. No init check here — gating on
+      // _isNativePluginInitialized() would lose every identify() issued while setup() is in
+      // flight, which is most startup identifies.
+      void this._enqueueNative('identify', (plugin) => {
+        const nativeDistinctId = String(this.getDistinctId() || previousDistinctId)
+        const anonymousId = String(this.getAnonymousId())
+        plugin.identify(nativeDistinctId, anonymousId)
```
marandaneto: suggestion: Return `plugin.identify(...)` from this callback. As written, `_enqueueNative` considers identify complete as soon as the bridge promise is created, so later reset/consent/push commands can advance before it settles and a rejection bypasses the queue's error handling. The current test only verifies invocation order, so it does not exercise promise settlement ordering.

ioannisj: Fixed in 49ec0e53e, the identify callback now returns the bridge promise so `_enqueueNative` only advances the chain once identify settles, and a rejection lands in the queue's catch instead of escaping. Added settlement-ordering and rejection tests.

packages/react-native/src/posthog-rn.ts
ioannisj: Bumped the floor to `>= 2.3.0` in 1135eb4, that's the minor this PR's changeset already publishes for the plugin, and the peer is optional so the not-yet-released version doesn't block anything at install. One correction on the framing though: the old-plugin combo was a graceful no-op rather than a silent failure, every bridge call is capability-gated (`posthog-rn.ts:1034`), so the bump mainly turns the mismatch into an install-time warning.

packages/react-native/src/posthog-rn.ts
ioannisj: Fixed in d4d50af, a manual `registerPushNotificationToken()` now persists a `PushRegistered` flag, and `unregisterPushNotificationToken()`/`reset()` escalate to `requireInit` when it's set, so a launch that never initialized native still boots it to reach the persisted subscription. The flag survives `reset()` (native re-binds the subscription rather than deleting it) and clears on unregister dispatch, both native SDKs persist the delete intent before their consent gates and own the retry from there. Added relaunch tests for the unregister and reset paths.

Worth a follow-up on posthog-android/ios: a native API that can touch the persisted subscription without a full setup would remove the boot entirely.

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
ioannisj: Fixed in 6b9b700, the Kotlin module now reads the launch Activity's intent right after `PostHogAndroid.setup()` and captures the cold-start tap itself, so the documented split (auto owns cold starts, manual owns warm/foreground) holds again. Recents relaunches are skipped via `FLAG_ACTIVITY_LAUNCHED_FROM_HISTORY` so a redelivered tray intent doesn't count as a new open, and the `google.message_id` extra is stripped afterwards so the native integration can't capture the same tap from that intent object.

The plugin has no Android test harness, so this is compile-verified only; worth a manual pass on the example app (tray tap from cold) before the release.

packages/react-native/src/posthog-rn.ts
ioannisj: You're right on both, and having looked closer the window isn't even narrow: `performUnregister` re-checks `config.optOut` after the identity-token mint (iOS does the same after its async mint), and `setOptOut(true)` lands milliseconds behind the unregister dispatch, so in the standard `optOut()` flow the DELETE probably parks most of the time. Bridge ordering can't reach it, both SDKs check consent inside their own async work, so I don't think this is solvable from the wrapper.

What softens it: both native SDKs persist the pending delete before their consent gates, so the DELETE isn't lost, it executes on the next opted-in launch. The exposure is the opted-out window itself, no capture happens during it, but the server subscription stays targetable.

Proposed split: filed PostHog/posthog-android#675 and PostHog/posthog-ios#746 for the native consent transition (delete while opted out, re-register on opt-in), and walked the changeset claim back in 6032ab8, it now discloses the limitation and links both issues. The RN side consumes the native fixes with plugin pin bumps only, no wrapper changes. Does that unblock this PR, or would you rather the native fixes land first?

marandaneto: idk what the side effects are of not unregistering so either take the call or check with the team (Workflows?)

ioannisj:
@marandaneto Thanx for all the review. Addressed almost everything here. Still pending decision on [this](https://github.com/PostHog/posthog-js/pull/4415#discussion_r3719629071) and CI is red cause https://github.com/PostHog/posthog-ios/pull/745 did not work as expected

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -333,15 +409,228 @@ class PosthogReactNativePluginModule(
...
+      appId: String,
+      completion: (String?) -> Unit,
+    ) {
+      val module = pushModule
+      val context = module?.reactApplicationContext
+      if (module == null || context == null || !context.hasActiveReactInstance()) {
+        declinePushIdentity(completion, "no React instance attached")
+        return
+      }
+      val requestId = UUID.randomUUID().toString()
```
marandaneto, line 583: we should always use uuid v7 unless query/sorting isnt important, not sure if thats the case, same for iOS
we already have a UUID v7 generator on the android sdk at least

packages/react-native-plugin/package.json
```diff
@@ -80,7 +80,6 @@
     "react-native": "*"
   },
   "jest": {
-    "preset": "react-native",
```
marandaneto: intentional? dont we need thid anymore?

packages/react-native/package.json
```diff
@@ -97,7 +97,7 @@
     "expo-device": ">= 4.0.0",
     "expo-file-system": ">= 13.0.0",
     "expo-localization": ">= 11.0.0",
-    "@posthog/react-native-plugin": ">= 2.2.0",
+    "@posthog/react-native-plugin": ">= 2.3.0",
```
marandaneto, line 100: not released so not sure if this will work? will ci fail during the release build?

marandaneto (approved):
left a few comments worth addressing/answering otherwise LGTM

packages/react-native/package.json
ioannisj: I think since this is an optional dependency, release should be okay. If it's okay this makes releasing RN + plugin ceremony way easier

marandaneto: when everything is green, its fine, if theres a partial success release, this would leave the sdk in a broken state, so its risky the least

sdk success
plugin fails

now the sdk is broken (its optional but what used to work does not work anymore eg session replay, error tracking, etc)

marandaneto:
> @marandaneto Thanx for all the review. Addressed almost everything here. Still pending decision on [this](https://github.com/PostHog/posthog-js/pull/4415#discussion_r3719629071) and CI is red cause [PostHog/posthog-ios#745](https://github.com/PostHog/posthog-ios/pull/745) did not work as expected

lets try this https://github.com/PostHog/posthog-ios/pull/747

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
ioannisj: For this specific case I think this is okay, just an identifier. never persisted or indexed

ioannisj:
> > @marandaneto Thanx for all the review. Addressed almost everything here. Still pending decision on [this](https://github.com/PostHog/posthog-js/pull/4415#discussion_r3719629071) and CI is red cause [PostHog/posthog-ios#745](https://github.com/PostHog/posthog-ios/pull/745) did not work as expected
> 
> lets try this [PostHog/posthog-ios#747](https://github.com/PostHog/posthog-ios/pull/747)

Should work, verified locally at least

packages/react-native/package.json
ioannisj: split the plugin into #4429 so 2.3.0 releases first. In general though, plugin + rn changes are always tightly coupled and would be easier if managed in same PR. Maybe we can think of sequencing the releases for these two packages?

packages/react-native/src/posthog-rn.ts
ioannisj: This will be tested internally first anw, so I would say safe to go now and we'll know more on the side effects of this. I'll communicate the two issues to workflows team

PR #4420, opened 2026-08-04
fix(ci): sign the commits these workflows create

Piccirello:
## Summary

Makes the workflows that commit in this repo produce **signed** commits.

Commits created with plain `git` are unsigned. PostHog is rolling the org-wide "Require signed commits" ruleset out to every repo, and this workflow would be rejected with `GH013` once it applies here. Routing the commit through the GitHub API instead means GitHub signs it with its own key, so it lands verified.

Found while inventorying which workflows across the org still create unsigned commits. `planetscale/ghcommit-action` is the pattern already in use in `posthog-js`, `posthog-roblox`, `brand`, `charts` and the SDK release fleet.

## Changes

`dependabot-changeset.yml` and `generate-references.yml` both commit through the API.

## Testing

**`dependabot-changeset.yml` is confirmed broken today**, not a precaution: run [30843671203](https://github.com/PostHog/posthog-js/actions/runs/30843671203) failed with `GH013: Commits must have verified signatures`. Of its last 100 runs, 98 were skipped and the single run that actually executed is that failure, so dependabot PRs have silently not been getting changesets. `changesets-dependencies-action` commits with plain git internally and has no API mode, so the fix lets its push fail via `continue-on-error` and commits the changeset it left in the workspace through the API. `generate-references.yml` is the latent case: `workflow_dispatch`-only, last ran 2025-12-20 before this repo was onboarded to the ruleset, so it has no failing run yet but would fail on the next dispatch.

### DIFF
--- a/.github/workflows/dependabot-changeset.yml
+++ b/.github/workflows/dependabot-changeset.yml
@@ -22,7 +22,10 @@ jobs:
       # the signed-commits ruleset on this repo rejects. continue-on-error keeps the
       # job alive through that rejected push; the changeset file it wrote is still in
       # the workspace, and the next step commits it through the GitHub API instead.
+      # The guard below makes sure this only tolerates the rejected push, not a real
+      # failure of the action.
       - name: Generate changeset
+        id: generate
         continue-on-error: true
         uses: the-guild-org/changesets-dependencies-action@f11b16181c79e07d62b112c2f32c9db534a9df09 # v1.2.2
         env:
@@ -42,8 +45,17 @@ jobs:
 
       - name: Check for a generated changeset
         id: changes
+        env:
+          GENERATE_OUTCOME: ${{ steps.generate.outcome }}
         run: |
           if [ -z "$(git status --porcelain -- .changeset)" ]; then
+            # The only failure we tolerate is the rejected push, which still leaves the
+            # changeset behind. Nothing to commit plus a failed step means the action
+            # itself broke, so surface it instead of passing a job that did nothing.
+            if [ "$GENERATE_OUTCOME" = "failure" ]; then
+              echo "::error::changesets-dependencies-action failed without writing a changeset"
+              exit 1
+            fi
             echo "No changeset generated"
             echo "changed=false" >> "$GITHUB_OUTPUT"
           else
--- a/.github/workflows/dependabot-changeset.yml
+++ b/.github/workflows/dependabot-changeset.yml
@@ -18,8 +18,34 @@ jobs:
         with:
           fetch-depth: 0
           token: ${{ steps.app-token.outputs.token }}
+      # The action writes the changeset and then commits it with plain `git`, which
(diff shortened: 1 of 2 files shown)


marandaneto:
The underlying signed-commit fix is needed, and using `ghcommit-action` is the right direction. I found one integrity issue in the Dependabot recovery path that I think we should address before merging.

After the generator's push fails, the workflow fetches the latest remote branch and runs `git reset --mixed` to that tip while preserving the older working tree. This can tolerate more than the intended unsigned-push failure:

- If generation fails after writing only some changesets, those partial results can still be committed.
- If the Dependabot branch advances between generation and the fetch, remote `.changeset` updates can appear locally as stale modifications or deletions and then be committed by `ghcommit-action`.

Could we validate the failed local commit before rewinding? On a failed generation step, I suggest requiring that:

1. local `HEAD` is exactly one commit ahead of the freshly fetched remote branch tip;
2. that commit changes only `.changeset/**`; and
3. the retained diff after resetting matches that commit.

If the remote advanced or those invariants do not hold, the job should fail and rerun against the new branch state rather than API-committing the preserved working tree. This keeps `continue-on-error` limited to the specific rejected-push case described in the workflow comments.

Piccirello:
Good catch — both cases are real, and the second one is worse than it reads. Fixed in a4e6088.

I reproduced the branch-advance case in a scratch repo. `reset --mixed` deliberately decouples HEAD from the working tree, and that is exactly what defeats the `expectedHeadOid` guard: after the reset, HEAD *is* the new tip, so the mutation is accepted rather than rejected. With the remote having deleted `.changeset/a.md` mid-run, the stale tree makes it reappear as untracked and `ghcommit` re-adds it:

```
$ git status --porcelain -uall -- .changeset
?? .changeset/a.md      <- deleted on the remote, about to be resurrected
?? .changeset/b.md
```

The partial-write case is real too: if generation dies after writing files but before `commitAll()`, there is no local commit, the reset is a no-op, and the old `GENERATE_OUTCOME` guard only caught the *empty* case — so partial results were committed.

I took your three invariants with one change. Rather than checking the retained diff separately (3), the reset now targets `HEAD^` instead of the fetched ref, which makes "retained tree == that commit's diff" true by construction:

```bash
remote=$(git rev-parse "refs/remotes/origin/$HEAD_REF")
[ "$(git rev-parse HEAD)"  = "$remote" ] && fail "generation failed without committing"
[ "$(git rev-parse HEAD^)" != "$remote" ] && fail "$HEAD_REF advanced during generation"
git diff --name-only HEAD^ HEAD | grep -qv '^\.changeset/' && fail "touches files outside .changeset/"
git reset --mixed HEAD^
```

The step is also now gated on `steps.generate.outcome == 'failure'`, so a successful push (if this repo ever stops requiring signatures) does not trip the one-commit-ahead check.

Verified against four scenarios: happy path allowed; partial-write-no-commit refused; branch-advanced refused; commit touching a file outside `.changeset/` refused.

marandaneto (approved):

PR #4430, opened 2026-08-05
feat(mcp): declare `_mcp_instructions` on tool output schemas

gesh:
**PR 1 of 3** in the `_mcp_instructions` stack. Sets up the fix for a measured zero-correlation bug; the follow-up does the writing.

## Problem

The `conversation_id` handle reaches the agent as a text block appended to `content`. Any client that reads `structuredContent` instead never sees it — which is what clients do whenever a tool declares an `outputSchema`.

Measured with Claude Code driving a real instrumented server over a 4-call dependent task:

| Tools declare `outputSchema` | Handle echoed back |
|---|---|
| none | **15/15 — 100%** |
| every tool | **0/15 — 0%** |

Not a degradation — zero. The SDK still emits the block correctly (verified with a raw MCP client); the client just never surfaces it. A stronger parameter description scored 0% too, because the message never arrives.

## Why the declaration is mandatory

[The spec](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#output-schema) makes both halves normative:

> If an output schema is provided:
> - Servers **MUST** provide structured results that conform to this schema.
> - Clients **SHOULD** validate structured results against this schema.

We are middleware that modifies results, so that MUST binds us: anything we add to `structuredContent` has to conform to the advertised schema. And since clients validate, a non-conforming key is not ignored — with `zod-to-json-schema` emitting `additionalProperties: false` for a plain `z.object`, it fails the **entire** tool result.

Declaring the property first is what brings the write into conformance.

## Why two PRs

The fix is mirroring the handle into `structuredContent`. That is only safe once the key is declared: the MCP client ajv-validates `structuredContent` against the schema from `tools/list`, and `zod-to-json-schema` emits `additionalProperties: false` for a plain `z.object`. An undeclared key is **not** ignored — it fails the entire tool result.

So this PR declares and ships alone; the next one writes. Shipping them in the other order breaks every result on a schema'd tool.

## Changes

New `output-instructions.ts`. When `enableConversationId` is on, `tools/list` adds an optional `_mcp_instructions` object to any tool advertising an `outputSchema`:

```jsonc
"properties": {
  "ok": { "type": "boolean" },
  "_mcp_instructions": {
    "type": "object",
    "properties": { "conversation_id": { "type": "string" }, "instructions": { "type": "string" } }
  }
},
"required": ["ok"],              // never added here — results predating this stay valid
"additionalProperties": false    // preserved; declaring inside `properties` is what makes the key legal
```

Skipped, both content-only and matching the policy on the input side: tools with **no** `outputSchema` (nothing to mirror into — `content` already works) and composed `oneOf`/`allOf`/`anyOf`/`$ref` schemas (no single `properties` bag). A tool that already declares the key keeps its own, with a warning.

Which tools got it is recorded on the existing per-tool ownership registry, so the follow-up only writes where writing validates.

**Inert on its own.** Nothing populates the field, and correlation on schema'd tools is still 0% after this PR.

## Tests

15 new tests (suite 451 → 466), covering the skip cases, non-mutation of the tool's schema, the ownership flag, and end to end through a real MCP client — including that **a result omitting the key still validates**, which is the regression that would catch declaring it as required.

## Not in scope

- Writing the key into `structuredContent` (PR 2)
- Re-confirming on every response, `isError` results, `get_more_tools` (PR 2)

- [x] @posthog/mcp

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility (optional property; no runtime behaviour change)

### Why `_mcp_instructions`

The 2026-07-28 reservation — *"any prefix where the second label is `modelcontextprotocol` or `mcp` is reserved for MCP use"* — governs **`_meta` key prefixes**, which are dot-separated labels followed by a slash (`io.modelcontextprotocol/`). This is a `structuredContent` property with no prefix grammar at all, so the rule does not reach it.

The name is deliberate rather than incidental: the field's job is to be read by a model and acted on. A key that reads as protocol-level guidance is treated as instructions; one that reads as vendor telemetry is plausibly skipped. The measured 100% echo rate is on this name, and nothing yet distinguishes the name's contribution from the channel's — so changing it would trade a measured configuration for an unmeasured one.

---

## Review follow-ups

- **The predicate now refuses to answer from a Zod value** (`1a1d057`). `canDeclareOutputInstructions` returns `false` when handed something schema-object-like (`_def` / `_zod` / `~standard`), mirroring what `analyticsOwnsParameter` already does for the input side. It could previously be fed a Zod schema — which has no `properties` bag, so every check passed vacuously and it answered `true` for a tool whose schema was actually skipped. That was the root cause of the #4431 bug; hardening it here makes the predicate impossible to misuse.
- **The e2e test now uses `client.callTool()`** instead of `client.request(..., CallToolResultSchema)`, which only parses the envelope. `callTool` ajv-validates `structuredContent` against the advertised output schema — the thing this PR exists to keep valid. The old assertion would have passed even with a wrong declaration.
- **Docstring names `additionalProperties: false`** as the reason an undeclared key fails the whole result, so the load-bearing rationale isn't only in this description.

### DIFF
--- a/.changeset/mcp-output-instructions-schema.md
+++ b/.changeset/mcp-output-instructions-schema.md
@@ -1,0 +1,8 @@
+---
+'@posthog/mcp': patch
+---
+
+Declare an optional `_mcp_instructions` property on the output schema of tools that
+advertise one, when `enableConversationId` is on. Inert by itself — it is the schema
+declaration that makes a later change able to mirror the conversation handle into
+`structuredContent` without failing client-side validation.
--- a/packages/mcp/src/__tests__/output-instructions.test.ts
+++ b/packages/mcp/src/__tests__/output-instructions.test.ts
@@ -1,0 +1,209 @@
+import { Client } from '@modelcontextprotocol/sdk/client/index.js'
+import { InMemoryTransport } from '@modelcontextprotocol/sdk/inMemory.js'
+import { McpServer } from '@modelcontextprotocol/sdk/server/mcp.js'
+import { CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
+import { z } from 'zod'
+import { instrument } from '../index'
+import { getAnalyticsParameterOwnership } from '../extensions/analytics-parameters'
+import {
+  MCP_INSTRUCTIONS_KEY,
+  addInstructionsToOutputSchema,
+  addInstructionsToOutputSchemas,
+  canDeclareOutputInstructions,
+} from '../extensions/output-instructions'
+import { fakePostHog } from './test-utils'
+
+/**
+ * Declaring `_mcp_instructions` on a tool's advertised output schema is what
+ * makes writing it into `structuredContent` safe: the MCP client ajv-validates
+ * that object against the schema under `additionalProperties: false`, so an
+ * undeclared key fails the whole tool result.
+ *
+ * This change only declares. Nothing writes the field yet.
+ */
+describe('_mcp_instructions output schema declaration', () => {
+  const objectSchema = () => ({
+    type: 'object' as const,
+    properties: { ok: { type: 'boolean' } },
+    additionalProperties: false,
+    required: ['ok'],
+  })
+
+  describe('canDeclareOutputInstructions', () => {
+    it('accepts a plain object schema', () => {
+      expect(canDeclareOutputInstructions(objectSchema())).toBe(true)
+    })
+
+    it('rejects a missing schema — there is nothing to mirror into', () => {
+      expect(canDeclareOutputInstructions(undefined)).toBe(false)
+      expect(canDeclareOutputInstructions(null)).toBe(false)
+    })
+
+    it('rejects composed schemas, which have no single properties bag', () => {
+      expect(canDeclareOutputInstructions({ oneOf: [] })).toBe(false)
+      expect(canDeclareOutputInstructions({ allOf: [] })).toBe(false)
+      expect(canDeclareOutputInstructions({ anyOf: [] })).toBe(false)
+      expect(canDeclareOutputInstructions({ $ref: '#/$defs/x' })).toBe(false)
+    })
+
+    it('rejects a schema that already declares the key', () => {
+      expect(
+        canDeclareOutputInstructions({ type: 'object', properties: { [MCP_INSTRUCTIONS_KEY]: { type: 'object' } } })
+      ).toBe(false)
+    })
+  })
+
+  describe('addInstructionsToOutputSchema', () => {
+    it('declares the key as an optional object property', () => {
+      const result = addInstructionsToOutputSchema({ name: 'get_issue', outputSchema: objectSchema() })
+      const schema = result.outputSchema as any
+
+      expect(schema.properties[MCP_INSTRUCTIONS_KEY].type).toBe('object')
+      expect(schema.properties[MCP_INSTRUCTIONS_KEY].properties.conversation_id.type).toBe('string')
+      // A result without the field must stay valid — every existing result lacks it.
+      expect(schema.required).toEqual(['ok'])
+    })
+
+    it('preserves the schema the tool advertised, including additionalProperties: false', () => {
+      const result = addInstructionsToOutputSchema({ name: 'get_issue', outputSchema: objectSchema() })
+      const schema = result.outputSchema as any
+
+      // Declaring inside `properties` is what makes the key legal; the closed
+      // schema itself is the customer's and stays untouched.
+      expect(schema.additionalProperties).toBe(false)
+      expect(schema.properties.ok).toEqual({ type: 'boolean' })
+    })
+
+    it('does not mutate the tool it was given', () => {
+      const outputSchema = objectSchema()
+      const tool = { name: 'get_issue', outputSchema }
+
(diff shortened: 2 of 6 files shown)


gesh:
Reviewed and ran this: stack tip 485 pass, and merged with #4428 it's 496 pass, `tsc --noEmit` clean, prettier clean. This PR itself looks correct to me — the declaration is genuinely inert, `required` is left alone, the deep copy protects a frozen/reused schema, and the skip cases match the input-side policy. I checked the regression you called out and a result omitting the key does still validate.

Two notes, neither blocking here.

**The predicate is correct, but it isn't the one that guards the write.** `canDeclareOutputInstructions` is a JSON-Schema predicate and gets a real JSON Schema in `getTracedToolsList`, so this PR uses it correctly. #4431 also feeds it `registeredTool.outputSchema`, which on the high-level path is a Zod value with no `.properties`, so it answers `true` unconditionally there. Details and a validated fix are in my comment on #4431 — mentioning it here only because the natural place to harden it is this file: an early `return false` when the argument looks like a schema object (`_def` / `_zod` / `~standard`) would make the predicate impossible to misuse, the way `analyticsOwnsParameter` already normalises Zod.

**The e2e test's client doesn't validate output schemas.** `client.request(..., CallToolResultSchema)` only parses the envelope; it does not ajv-validate `structuredContent` against the tool's output schema. `client.callTool()` does. So `a result omitting the key still validates` is currently asserting something weaker than it reads — it would pass even if the declaration were wrong. Worth switching, since client-side validation is the entire premise of splitting these two PRs.

Small thing: the `zod-to-json-schema` reasoning in the description is the load-bearing part of why this has to land first, and it's only in the PR body. A line of it in the module docstring would keep it discoverable — the existing docstring says an undeclared key "fails the entire tool result" but not that `additionalProperties: false` is what makes that true.

gesh:
Both taken.

**Predicate hardened here** (`1a1d057`) — `canDeclareOutputInstructions` returns `false` for anything schema-object-like (`_def` / `_zod` / `~standard`). You were right that this file is the place: it was the root cause of the #4431 bug, and refusing to guess makes it impossible to misuse rather than merely fixed at one call site.

**Test switched to `client.callTool()`** — good catch that the assertion read stronger than it was. It now ajv-validates, which is the entire premise of splitting these two PRs.

**Docstring** now names `additionalProperties: false` as the reason an undeclared key fails the whole result.

lucasheriques (approved):
the declare-then-write split reasoning is airtight, and the "result omitting the key still validates" test is exactly the right regression guard. approving. one guard edge worth fixing and one naming question.

**`output-instructions.ts:119` — should fix:** two edges in `canDeclareOutputInstructions`. a tool advertising `properties: true` (malformed but harmless before this) passes the guard, then the write throws a typeerror on a boolean inside the tools/list wrapper, which logs "an error posthog mcp analytics did not cause" (it did, this time) and rethrows, taking down the whole listing. i reproduced this in node. separately, `isSchemaObjectLike` scans `Object.values()`, so a json schema with a property literally named `_def` gets misclassified as zod and silently skipped. a `typeof schema.properties === 'object'` check covers the first, checking only the candidate itself covers the second.

**`output-instructions.ts:24` — question:** was `_posthog_instructions` considered? the 2026-07-28 schema reserves any prefix whose second label is `mcp` or `modelcontextprotocol` for spec use. that's written for `_meta` keys so it doesn't literally forbid this, but `_mcp_*` reads like spec surface, and once agents depend on the key we can't rename it. if `_mcp_` won for agent-compliance reasons that's a fine answer, just worth a line in the pr body.

gesh:
Both taken, both reproduced first (`7d3e2a4`).

**`output-instructions.ts:119` — fixed.** Your `properties: true` case throws exactly as you said:

```
TypeError: Cannot create property '_mcp_instructions' on boolean 'true'
```

and it surfaces inside the `tools/list` wrapper, so an annotation we could simply have skipped fails the whole listing. Non-object `properties` is now refused.

The `_def` case is fixed too, but not by checking only the candidate — that would have reintroduced the raw-shape hole the guard exists for, since a raw Zod shape has no `_def` at its top level either. It now settles "is this a JSON Schema" first, judged on the **value** of each marker rather than the key, because a raw shape is free to name a field `type` or `properties`:

| input | before | after |
|---|:--:|:--:|
| zod object | false | false |
| raw zod shape | false | false |
| raw shape with keys `type`/`properties` | false | false |
| plain json schema | true | true |
| json schema with a `_def` property | **false** | true |
| schema already declaring the key | false | false |
| `properties: true` | **true** | false |

Three tests added; verified they fail with either half of the guard reverted.

**`output-instructions.ts:24` — keeping `_mcp_instructions`, rationale now in the PR body.** The reservation you're thinking of covers `_meta` key *prefixes* — dot-separated labels followed by a slash, `io.modelcontextprotocol/`. This is a `structuredContent` property with no prefix grammar, so the rule doesn't reach it.

Worth being straight about the reason though: it isn't that `_mcp_` is safe, it's that the field exists to be read by a model and acted on, and a key reading as protocol guidance gets treated as instructions where one reading as vendor telemetry plausibly gets skipped. The 100% echo measurement is on this name, and nothing separates the name's contribution from the channel's — so a rename would swap a measured configuration for an unmeasured one. Agreed it's unrenameable later; that's the call being made now rather than an oversight.

PR #4431, opened 2026-08-05
feat(mcp): mirror the session handle into `structuredContent`

gesh:
**PR 2 of 3** in the `_mcp_instructions` stack. Stacked on #4430 — review that first; this diff shows only the top commit once it merges.

## Problem

[MCP 2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removes protocol-level sessions. Servers correlate calls via an explicit handle the agent threads back — for us, `conversation_id`.

That handle reaches the agent as a text block on `content`. But clients read `structuredContent` whenever a tool declares an `outputSchema`, and never see it:

| Tools declare `outputSchema` | Handle echoed back |
|---|---|
| none | 15/15 — 100% |
| every tool | **0/15 — 0%** |

Measured with Claude Code against a real instrumented server. The SDK emitted the block correctly the whole time; the client just never surfaced it.

## Why `structuredContent`, per the spec

[Stateful Tools](https://modelcontextprotocol.io/specification/2026-07-28/server/tools#stateful-tools) prescribes this exact pattern:

> MCP has no protocol-level session, so a server cannot rely on implicit per-connection state to relate one tool call to the next. Servers that need to maintain state across calls … should do so by **returning an explicit handle from a creation tool and accepting that handle as an argument on subsequent calls**.

and its worked example returns that handle in `structuredContent`:

```jsonc
// ← result
{
  "content": [{ "type": "text", "text": "Created basket bsk_a1b2c3" }],
  "structuredContent": { "basket_id": "bsk_a1b2c3" }
}
```

Two more rules make both channels necessary rather than either/or:

> Clients **SHOULD** validate structured results against this schema.

> For backwards compatibility, a tool that returns structured content **SHOULD** also return the serialized JSON in a TextContent block.

So the handle belongs in `structuredContent` where a structured-reading client will find it, while `content` stays the compatibility path. We were only doing the second.

## Change

Write the handle into `structuredContent` for tools whose schema #4430 declared `_mcp_instructions` on.

**Before** — handle present, agent never sees it:
```jsonc
{
  "content": [
    { "type": "text", "text": "{\"metric\":\"signups\",\"value\":4210}" },
    { "type": "text", "text": "[SERVER]: Reuse conversation_id=019fd23a-…" }   // ← dropped by the client
  ],
  "structuredContent": { "metric": "signups", "value": 4210 }                  // ← what it actually reads
}
```

**After**:
```jsonc
{
  "structuredContent": {
    "metric": "signups",
    "value": 4210,
    "_mcp_instructions": {
      "conversation_id": "019fd23a-363b-7c25-b26d-0565f728712a",
      "instructions": "Send this conversation_id as an argument on every subsequent tool call in this conversation."
    }
  }
}
```

Same measurement now: **15/15 — 100%**, each run on a single `$session_id`.

The structured copy rides **every** response, unlike the text block, so an agent that dropped the handle can read it back. The text block stays mint-only — repeating it would put a `[SERVER]:` line in front of the user on every tool result.

Never written where it wouldn't validate: only declared tools, only a plain-object `structuredContent`, never over a key the tool produced itself.

## Tests

9 new (suite 466 → 475), covering the skip cases, non-mutation, and end to end through a real MCP client. **No existing tests changed.**

## Not in scope

Both now in #4433, stacked on this:

- Prompt-back on `isError` results — separate channel, separate justification
- Including `get_more_tools`

- [x] @posthog/mcp

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility (optional field; only where declared)

---

## Review follow-ups

- **The `structuredContent` write is now gated on the listing** (`25100c8`). `outputInstructions` reads from `listed`, never from the override:

  ```ts
  const listed = toolName ? data.toolAnalyticsParameterOwnership.get(toolName) : undefined
  const ownership = override ?? listed
  // ...
  outputInstructions: data.options.enableConversationId === true && listed?.outputInstructions === true,
  ```

  Only the advertised JSON Schema can answer whether `tools/list` declared the key. The high-level override is built from the live registry, which holds Zod — so it answered `true` unconditionally, and we wrote an undeclared key into a tool that owned `_mcp_instructions` itself, failing the entire result with `-32602`. A tool never listed at all is now also left alone.
- **The reviewer's repro is now a regression test** — a tool declaring its own `_mcp_instructions`, asserting we skip the write.
- **The e2e tests use `client.callTool()`**, which ajv-validates the result. That's why a green suite missed this.
- The paired hardening (predicate refuses Zod values) is in #4430, where the predicate lives.

### DIFF
--- a/.changeset/mcp-mirror-structured-instructions.md
+++ b/.changeset/mcp-mirror-structured-instructions.md
@@ -1,0 +1,7 @@
+---
+'@posthog/mcp': patch
+---
+
+Mirror the `conversation_id` session handle into `structuredContent` for tools whose
+output schema declares `_mcp_instructions`. Clients that read structured results never
+saw the handle in `content`, so correlation for those tools was zero.
--- a/packages/mcp/src/__tests__/output-instructions.test.ts
+++ b/packages/mcp/src/__tests__/output-instructions.test.ts
@@ -221,6 +221,25 @@ describe('_mcp_instructions declaration over a real client', () => {
     }
   })
 
+  it('mirrors a key that validates against the real advertised schema', async () => {
+    // The claim this whole two-PR split exists to satisfy: the key we write is
+    // one the client's cached schema accepts. Only `callTool` ajv-validates
+    // `structuredContent`, so asserting the mirrored key anywhere else proves
+    // the write happened but not that it survives validation.
+    const { client, cleanup } = await connect({ enableConversationId: true })
+    try {
+      await client.listTools()
+
+      const result = await client.callTool({ name: 'get_issue', arguments: { issue_id: 'iss_7' } })
+
+      const mirrored = (result.structuredContent as Record<string, any>)[MCP_INSTRUCTIONS_KEY]
+      expect(mirrored?.conversation_id).toEqual(expect.any(String))
+      expect((result.structuredContent as any).ok).toBe(true)
+    } finally {
+      await cleanup()
+    }
+  })
+
   it('still accepts a result that omits the declared key', async () => {
     const { client, cleanup } = await connect({ enableConversationId: true })
     try {
--- a/packages/mcp/src/__tests__/output-instructions.test.ts
+++ b/packages/mcp/src/__tests__/output-instructions.test.ts
@@ -326,12 +326,15 @@ describe('mirroring over a real client', () => {
     const { client, cleanup } = await connectStructured({ enableConversationId: true })
     try {
       await client.request({ method: 'tools/list' }, ListToolsResultSchema)
-      const second = await call(client, 'conv-supplied')
+      // Shaped like a handle the SDK would have minted, which is what a real
+      // agent echoes — an arbitrary string is treated as no handle at all.
+      const echoed = '019fd2b0-1111-7111-8111-111111111111'
+      const second = await call(client, echoed)
 
       // The text block is mint-only; the structured copy rides every response.
       const hasText = (second.content ?? []).some((c: any) => String(c.text).includes('conversation_id='))
       expect(hasText).toBe(false)
-      expect((second.structuredContent as any)[MCP_INSTRUCTIONS_KEY].conversation_id).toBe('conv-supplied')
+      expect((second.structuredContent as any)[MCP_INSTRUCTIONS_KEY].conversation_id).toBe(echoed)
     } finally {
       await cleanup()
     }
--- a/packages/mcp/src/__tests__/output-instructions.test.ts
+++ b/packages/mcp/src/__tests__/output-instructions.test.ts
@@ -10,6 +10,7 @@ import {
   addInstructionsToOutputSchema,
   addInstructionsToOutputSchemas,
   canDeclareOutputInstructions,
+  mirrorInstructionsIntoStructuredContent,
 } from '../extensions/output-instructions'
 import { fakePostHog } from './test-utils'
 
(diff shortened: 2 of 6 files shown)


gesh:
Reviewed this stack end to end and ran the suites: #4428 alone 462 pass, this stack tip 485 pass, and #4428 merged into the stack tip auto-merges clean at 496 pass, `tsc --noEmit` clean, prettier clean. The design is right and the declare-then-write split is the correct call. One real bug in the gating, though.

## The `structuredContent` write is not actually gated on the declaration (high-level path)

`instrument-highlevel.ts:266` computes ownership from the live tool object:

```ts
getAnalyticsParameterOwnership(registeredTool.inputSchema, registeredTool.outputSchema)
```

But on the high-level path `registeredTool.outputSchema` is a **Zod value**, while `canDeclareOutputInstructions` is a **JSON-Schema** predicate. It looks for `.properties` / `.oneOf` / `$ref`, none of which exist on a Zod object, so it answers `true` for anything:

```
canDeclareOutputInstructions({ _mcp_instructions: z.string(), ok: z.boolean() })  -> true   // raw shape
canDeclareOutputInstructions(z.object({ ok: z.boolean() }))                       -> true   // zod object
canDeclareOutputInstructions({ type: 'object', properties: { _mcp_instructions: { type: 'string' } },
                               additionalProperties: false })                     -> false  // the only correct answer
```

What the SDK actually stores confirms it:

```
registeredTool.outputSchema keys: spa,_def,parse,safeParse,parseAsync,safeParseAsync,refine,refinement
has .properties?                  false
ownership.outputInstructions      true
```

So the comment at `instrumentation.ts:159-161` — "whether `tools/list` declared `_mcp_instructions` on this tool, i.e. whether writing that key into `structuredContent` would validate client-side" — does not hold on this path. The override never consults what the listing did.

### Failing repro

A tool that declares its own `_mcp_instructions`. `tools/list` correctly skips the declaration and logs `WARN: Tool "own_key" already declares '_mcp_instructions' ... Leaving it alone` — and then we write it anyway:

```ts
it('writes an undeclared key when the tool owns _mcp_instructions', async () => {
  const server = new McpServer({ name: 'p', version: '1.0.0' })
  server.registerTool(
    'own_key',
    {
      description: 'o',
      inputSchema: {},
      outputSchema: { _mcp_instructions: z.string().optional(), ok: z.boolean() },
    },
    async () => ({ content: [{ type: 'text', text: '{"ok":true}' }], structuredContent: { ok: true } })
  )
  instrument(server, fakePostHog(), { enableConversationId: true })

  const client = new Client({ name: 'test', version: '1.0.0' })
  const [c, s] = InMemoryTransport.createLinkedPair()
  await Promise.all([client.connect(c), server.connect(s)])
  await client.listTools()
  await client.callTool({ name: 'own_key', arguments: {} })
})
```

```
McpError: MCP error -32602: Structured content does not match the tool's output schema:
data/_mcp_instructions must be string
```

The entire tool result is rejected — exactly the failure #4430's description exists to prevent.

Two related observations from the same root cause:

- **Never-listed tools are written to.** Call a tool without the client ever issuing `tools/list` and we still write the key, despite nothing having been advertised.
- **The low-level path is correct.** Same tool shape against a low-level `Server` returns `OK {"ok":true}` with no write, because its flag comes from the listing cache. The intended design works; only the override bypasses it.

Worth noting the asymmetry inside the same function: `analyticsOwnsParameter` (used for `context` / `conversation_id`) carefully normalises Zod via `isZodRawShapeCompat` / `getObjectShape` / `_def` detection. The output-side predicate skipped that step.

### Proposed fix

The listing cache already holds the right answer, computed from the real JSON Schema in `getTracedToolsList` before the declaration is applied. Take `outputInstructions` from there unconditionally, and let the override stay input-side only:

```diff
 function getActiveAnalyticsParameterOwnership(
   data: MCPAnalyticsData,
   toolName: string | undefined,
   override: AnalyticsParameterOwnership | undefined,
   isMissingCapabilityTool: boolean
 ): AnalyticsParameterOwnership {
-  const ownership = override ?? (toolName ? data.toolAnalyticsParameterOwnership.get(toolName) : undefined)
+  const listed = toolName ? data.toolAnalyticsParameterOwnership.get(toolName) : undefined
+  const ownership = override ?? listed
   return {
     context: !isMissingCapabilityTool && isContextEnabled(data.options.context) && ownership?.context === true,
     conversationId: data.options.enableConversationId === true && ownership?.conversationId === true,
-    // Whether `tools/list` declared `_mcp_instructions` on this tool, i.e. whether
-    // writing that key into `structuredContent` would validate client-side.
-    outputInstructions: data.options.enableConversationId === true && ownership?.outputInstructions === true,
+    // Whether `tools/list` declared `_mcp_instructions` on this tool, i.e. whether
+    // writing that key into `structuredContent` would validate client-side. Only the
+    // listing can answer that: the high-level override carries the tool's Zod schema,
+    // not the JSON Schema the declaration was applied to.
+    outputInstructions: data.options.enableConversationId === true && listed?.outputInstructions === true,
   }
 }
```

Verified on this branch:

```
own_key   OK wrote=false     (was: -32602)
baseline  OK wrote=true      (unchanged)
no-list   OK wrote=false     (was: wrote=true)
485 existing tests still pass
```

Worth adding a guard in `canDeclareOutputInstructions` too, so feeding it a Zod value can't silently answer `true` again — return `false` when the argument looks like a schema object (`_def` / `_zod` / `~standard`), mirroring what `analyticsOwnsParameter` already does.

## The e2e tests bypass the validation they're asserting

This is why a green suite didn't catch the above. `client.request(..., CallToolResultSchema)` only parses the envelope — it does **not** validate `structuredContent` against the tool's output schema. `client.callTool()` does. So `delivers the session handle where a structured-only client will see it` and its siblings never exercise the ajv validation the whole two-PR split exists to satisfy.

Switching those three cases to `client.callTool({ name, arguments })` makes the bug above fail immediately, and makes the tests prove their actual claim.

## Rollout note

A client holding a `tools/list` from before the declaration gets the other side of it:

```
McpError: MCP error -32602: Structured content does not match the tool's output schema:
data must NOT have additional properties
```

Since #4430 and #4431 ship in the same release there's no cross-version window, and it self-heals once the client re-lists — but a long-lived agent connection spanning a server deploy will hit it, and the fix above doesn't help there (we can't see the client's cache). Probably just worth a line in the release notes.

## Also checked, all clean

Ordinary schemas advertise as plain object schemas and declare correctly, so the bug above needs a tool that literally names a field `_mcp_instructions` — unlikely in practice:

| output schema | advertised top level | declared | write |
|---|---|---|---|
| reused sub-schema | `type,properties,required,additionalProperties,$schema` | ✓ | ✓ |
| `.nullable()` object | same | ✓ | ✓ |
| discriminated-union field | same | ✓ | ✓ |
| `z.object(...)` | same | ✓ | ✓ |
| `.passthrough()` | same | ✓ | ✓ |
| top-level `z.intersection` | dropped by SDK 1.29 | ✗ | ✗ (consistent) |

Server-side ordering is also right: the mirror runs after `originalCallToolHandler`, so the SDK's own output validation sees the tool's untouched `structuredContent`.

gesh:
Confirmed and fixed (`25100c8`) — thanks, this was a real bug and the repro made it obvious.

Took your diff as proposed: `outputInstructions` reads `listed`, never the override. Your repro is now a regression test, and the three e2e cases use `client.callTool()` so they ajv-validate — which is why the suite was green over a `-32602`. The never-listed case is closed by the same change. Paired hardening (predicate refuses Zod values) went into #4430, where the predicate lives.

**Rollout note taken** — a client holding a pre-declaration `tools/list` across a server deploy hits the mirror image of this, and nothing server-side can see that cache. Will call it out in the release notes.

Also worth flagging: your table checking reused / `.nullable()` / discriminated-union / `.passthrough()` schemas is more coverage of the advertised shape than the PR had. That's the part I'd have been least confident about.

lucasheriques (approved):
the delivered-accounting (clearing ids nobody received off the event) is my favorite detail in the stack. approving. one question that i think matters a lot for stateless deployments.

**`instrumentation.ts:166` — question:** the write gate deliberately reads `listed`, and `listed` is a per-instance cache. on the stateless pattern this sdk explicitly supports (fresh server + `instrument()` per request), tools/list lands on pod a and every tools/call on a cold pod b, so the mirror never fires. permanently, not just "before any tools/list" like the comment says. and the new spec makes this worse: with `ttlMs` clients cache tools/list across connections, so even a long-lived instance may never serve a given client's listing. meanwhile the client's cached schema does declare the key (pod a declared it), so writing would validate fine. `reportMissing` solved this exact shape by recomputing from the raw listing instead of caching. same fix here? also the same root cause as my ownership question on #4428, one fix might cover both.

**`output-instructions.test.ts:200` — nit:** the ajv-validating path and the mirrored-key assertions never overlap. the only `callTool` test no longer omits the key now the write is live, and every key-present assertion uses `client.request()` which skips validation. one `callTool` test asserting the mirrored result validates against a real zod-derived schema would cover the exact claim in the pr body.

gesh:
You're right on both, and the first is worse than the comment admitted (`e4b1c7d`).

**`instrumentation.ts:166` — confirmed, and it's permanent, not transient.** Same server, same tool, only the instance differs:

```
warm  (tools/list then tools/call, one instance)   mirror fired: YES
cold  (tools/list on A, tools/call on B)           mirror fired: NO
```

The comment said "called before any `tools/list`", which reads as a startup window. It isn't — an instance that never serves a listing never mirrors at all. Comment now says that.

**Not taking the recompute, though, and I want to be precise about why the `reportMissing` analogy doesn't carry.** That one recomputes the *tool name* from options, which is available on any instance. Here the question is what JSON Schema the listing advertised, and on a cold instance nothing can answer it: the high-level registry holds Zod, and converting it ourselves means reimplementing what the MCP SDK does at listing time — with version drift producing a schema that differs from the one the client actually cached.

The asymmetry decides it. Writing a key the advertised schema didn't declare fails the **entire** tool result under `additionalProperties: false` — the failure #4430 exists to prevent. A wrong guess costs the caller their result; abstaining costs us one delivery channel while the `content` block still carries the handle. So it fails closed.

The real fix is the process-scoped ownership cache, so a listing served by *any* instance answers for the rest. That also covers your #4428 ownership question by the same route, which is the "one fix for both" you suspected — just at the cache's scope rather than at each call site. Both are recorded against that work.

Worth flagging one case it still won't reach, since you raised `ttlMs`: a process that never serves a listing at all because every client is working from a cross-connection cache. Process scoping doesn't help there, and I don't have an answer that doesn't involve guessing at the schema.

**`output-instructions.test.ts:200` — fixed.** Added a `callTool` test asserting the mirrored key alongside the tool's own field, so the ajv path and the key-present assertion finally overlap. Verified it fails when the mirror writes an undeclared key — which is the assertion the PR body was claiming all along and the suite wasn't making.

PR #4432, opened 2026-08-05
fix(replay): replay events captured during recorder lazy load through event trigger matching

arnohillen:
## Problem

Event trigger matching lives in the lazy-loaded recorder chunk, and its `eventCaptured` hook is only registered inside the recorder's `start()` (via `setupEventTriggerListeners`). Events captured before the chunk finishes loading are never seen by trigger matching, and there is no catch-up.

The initial `$pageview` of a pageload is captured at init, before remote config and the recorder script have loaded, so a `$pageview` event trigger can structurally never activate on the first page a user lands on. Sessions on single-page sites then buffer forever and are never flushed. This shipped as a customer-facing total recording failure (recordings worked on an app subdomain where users navigate, so route-change pageviews eventually fired after the recorder was up, and never on the marketing site where most visits are a single page).

## Change

- The eager `SessionRecording` extension registers an `eventCaptured` hook at construction and buffers events (capped at 100, keeping the earliest; `$snapshot` is excluded since it is recorder output with heavy payloads) until the recorder starts.
- The lazy recorder's `start()` consumes the buffer immediately after registering its live trigger listener, so nothing captured later in `start()` (e.g. a `$snapshot` from an override) can enter it, and replays the consumed events through the strategy's trigger callback at the end of `start()`, once the recorder is running. This covers both V1 event triggers and V2 trigger groups (including their property filters).
- Consumed events are filtered to the current session id, so a `posthog.reset()` or session rotation during the chunk load cannot activate recording for a session that never contained the trigger.
- The buffer hook is torn down (and the buffer released) on every terminal path where the recorder will not start: client `disable_session_recording`, consent opt-out, server-disabled remote config, a failed config fetch with nothing persisted to start from, stale persisted config after a failed refresh, recorder script load failure, a missing script loader (no-external bundles without the recorder imported), an adblocked recorder script, and `dispose()`. It stays armed through the pending states (before the first remote config, `AWAITING_CONFIG`, `LAZY_LOADING`).
- If a later remote config re-enables recording after a disabled-state teardown, buffering resumes before the lazy load begins, so that load window is covered too.
- `SessionRecording` now implements the optional `Extension.dispose()`, and the core calls it wherever it replaces or drops the instance (`_replaceExtension`, and the cookieless opt-out path). Previously those sites dropped the instance while leaving its `visibilitychange` listener attached, so a page that flipped consent accumulated one dead listener per flip.

Note on billing: a replayed pre-load event matches and scopes the trigger, but does not release the fresh-start interaction hold. Those events were captured before the recorder existed, so they are not interaction evidence, and releasing on them would ship a billable recording for a prefetched or background tab whose only event is the initial `$pageview`. A real interaction afterwards releases the hold through the usual path, so a session a user actually touched still records from `t=0`. Live trigger matches after the recorder has started are unaffected and still release.

## Backwards compatibility

The cross-bundle surface is one new public method on the eager extension, called with optional chaining from the recorder chunk:

- new recorder chunk + older bundled core: the method doesn't exist, replay is skipped, behavior is unchanged.
- older recorder chunk + new core: the buffer is never consumed and is cleared by the eager side after start.

The test harness mock of `posthog.on` previously returned one shared unsubscribe mock for every subscription, which the new second subscriber made ambiguous; it now returns a distinct unsubscribe per call, with the affected assertions updated to match.

## Testing

- 18 tests covering: trigger activation from a pre-load event, non-matching pre-load events keep buffering, previous-session events are not replayed, buffering stops after recorder start, the 100-event cap keeps the earliest events, `$snapshot` exclusion, buffer teardown on each terminal path (remote config disabled, client-side disable, opt-out, failed config fetch, script load failure, missing loader, adblocked recorder), buffering resuming when a later config enables recording, and a new recorder chunk running against an older core without the buffer API. Three more cover the fresh-start hold (a replayed pre-load event does not release it, a real interaction afterwards does, and a live post-start trigger match still does), and one covers listener disposal across consent flips.
- Each test was mutation-checked: reverting the corresponding fix flips exactly that test red.
- Full browser package suite green (126 suites, 5526 tests), typecheck, lint, and a full bundle build (including the cross-bundle private-property consistency checks) pass.

---


packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1241,6 +1247,17 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
         }
+
+        // events captured while this recorder chunk was loading never reached the trigger
+        // listener registered above, so an event trigger on e.g. the initial $pageview could
+        // otherwise never match on the first page of a pageload. Replay them through the
+        // matchers. Optional call: this chunk can run against an older bundled core.
+        const preStartEvents = this._instance.sessionRecording?.consumeEventsCapturedBeforeRecorderStarted?.() ?? []
+        if (eventTriggerCallback) {
+            for (const event of preStartEvents) {
+                eventTriggerCallback(event)
```
marandaneto: **[P2] Filter buffered events by session before replaying them.** If `posthog.reset()` or session rotation occurs while the recorder chunk loads, this callback uses the new current session ID for an event whose `properties.$session_id` belongs to the previous session. That stale event can therefore activate recording for a new session/user that never contained the trigger; clear or partition the buffer on session changes, or only replay events belonging to the current session.

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -288,6 +322,7 @@ export class SessionRecording implements Extension {
         if (response.sessionRecording === false) {
             this._persistRemoteConfig(response)
             this._discardRecording()
+            this._stopBufferingPreStartEvents()
```
marandaneto, line 358: **[P2] Re-enable buffering when remote config later enables recording.** This permanently unregisters the only pre-start hook. If a subsequent remote-config refresh changes `sessionRecording` from `false` to enabled, the recorder starts lazy loading without reinstalling the hook, so events captured during that load are still missed.

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -99,6 +110,12 @@ export class SessionRecording implements Extension {
         if (document?.addEventListener) {
             addEventListener(document, 'visibilitychange', this._onVisibilityChange)
         }
+
+        this._removePreStartEventBufferHook = this._instance.on?.('eventCaptured', (event) => {
```
marandaneto: **[P2] Release buffered payloads when recording cannot start.** This hook is installed unconditionally, but client-disabled recording, missing/failed config, and recorder dependency failures never consume or clear it. Those pages retain the first 100 complete `CaptureResult` payloads and continue invoking this listener for every event for the page lifetime; tear it down on terminal disabled/error paths.

marandaneto (commented):
Inline review findings from the lazy-load event buffer review.

packages/browser/src/extensions/replay/session-recording.ts
TueHaulund: The hook goes on here no matter what, but there are only three places that turn it off: `dispose()`, the `sessionRecording === false` branch, and the end of `_onScriptLoaded()`.

So it keeps buffering when:
- `disable_session_recording` is set
- the user opted out
- the server says `enabled: false` (these three all hit the `DISABLED` branch at L172)
- config is missing
- an adblocker blocks the recorder script
- the stale-config early return fires

Either clear the buffer in those paths too, or wait to add the hook until we know a start is coming. The second seems simpler.

Related: `stopRecording()` (L220) doesn't stop the buffering either, and nothing calls `dispose()` when the instance is thrown away (`_replaceExtension` at posthog-core.ts:465, `opt_out_capturing()` at 4066). The old instance sticks around because the hook still points at it. Calling `_stopBufferingPreStartEvents()` from `stopRecording()` would cover that case.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1133,10 +1134,15 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         // Let the strategy configure itself
         this._strategy.onRemoteConfig(config)
 
-        // Setup event trigger listeners via strategy
+        // Setup event trigger listeners via strategy. The strategy's callback is stashed so
+        // that events captured before this point can be replayed through it at the end of start().
+        let eventTriggerCallback: ((event: CaptureResult) => void) | undefined
         this._removeEventTriggerCaptureHook?.()
         this._removeEventTriggerCaptureHook = this._strategy.setupEventTriggerListeners(
```
TueHaulund, line 1207: The live listener starts here, but the buffer doesn't drain until L1255. In that gap, `_checkOverride` at L1164 can trigger a `$snapshot` capture, which lands in the buffer and gets replayed through the matchers. Draining first avoids it.

packages/browser/src/__tests__/extensions/replay/sessionRecording-onRemoteConfig.test.ts
```diff
@@ -151,7 +149,7 @@ describe('SessionRecording', () => {
             _internalEventEmitter: simpleEventEmitter,
             on: jest.fn().mockImplementation((event, cb) => {
                 const unsubscribe = simpleEventEmitter.on(event, cb)
-                return removePageviewCaptureHookMock.mockImplementation(unsubscribe)
+                return jest.fn().mockImplementation(unsubscribe)
```
TueHaulund, line 153: This file is missing the `posthog.sessionRecording = sessionRecording` line the other suite has, so it hits the skew path by chance rather than on purpose. Could also use a test for cleanup on a disabled path, and one for the new chunk with old core case.

TueHaulund (changes requested):
Good catch on the version skew between the chunk and core.

One blocker: the pre-start buffer never gets cleaned up unless recording actually starts. If it doesn't, the hook stays on the capture path and holds onto up to 100 events for the rest of the page. Two smaller things inline.

Also worth a line in the description: a buffered event now hits `_activateTrigger('event')`, which releases the fresh-start hold based on something captured before the recorder loaded. Looks intentional given the comment at lazy-loaded-session-recorder.ts:1066, but it changes which background loads end up billable.

marandaneto:
moving to draft until review is addressed

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: consumeEventsCapturedBeforeRecorderStarted now filters the returned events to those whose properties.$session_id matches the current session (read-only check), so a reset or rotation during the chunk load can only drop stale events, never replay them into the new session; covered by the "does not replay events that belong to a previous session" test.

packages/browser/src/extensions/replay/session-recording.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: the enabled branch of startIfEnabledOrStop re-installs the buffer hook before the lazy load begins, so a later remote config that turns recording on gets its whole load window buffered again; covered by the "resumes buffering when a later remote config enables recording" test.

packages/browser/src/extensions/replay/session-recording.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: the hook is torn down and the buffer released on every terminal path (client disable, opt-out, server-disabled config, failed config fetch with nothing persisted, stale config after a failed refresh, script load failure, missing loader, adblocked recorder, dispose), while pending states (pre-config, AWAITING_CONFIG, LAZY_LOADING) keep buffering; each teardown path has a test that fails if its call is removed.

arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: all six of these paths now tear the hook down and release the buffer (the DISABLED branch does it whenever the off is definitive, i.e. client-disabled/opted-out or a server decision exists, so the pre-config pending window keeps buffering as intended), plus dispose and the no-loader case; each path has a dedicated test.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: start() now consumes the buffer (stopping further buffering) immediately after the live listener registers and before any _checkOverride runs, so an override-triggered $snapshot cannot enter it; the replay of the consumed events still happens at the end of start() so activation behaves as if the events arrived just after start. The eager hook also skips $snapshot outright since it is recorder output with heavy payloads.

packages/browser/src/__tests__/extensions/replay/sessionRecording-onRemoteConfig.test.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 4ce1344b: the file now sets posthog.sessionRecording = sessionRecording explicitly, and there are dedicated tests for teardown on the disabled paths (remote-config false, client disable, opt-out, failed fetch, load failure, missing loader, adblocker) and for a new chunk running against an older core without the buffer API (no crash, no replay).

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -108,9 +119,40 @@ export class SessionRecording implements Extension {
...
         this.stopRecording()
     }
 
+    /** called by the lazy-loaded recorder on start, so pre-start events can be replayed through trigger matching */
+    public consumeEventsCapturedBeforeRecorderStarted(): CaptureResult[] {
+        const events = this._eventsCapturedBeforeRecorderStarted
+        this._stopBufferingPreStartEvents()
+        // a rotation or reset during the chunk load must not let a previous session's events activate this one
+        const sessionId = this._instance.sessionManager?.checkAndGetSessionAndWindowId(true)?.sessionId
+        return events.filter((event) => event.properties?.$session_id === sessionId)
```
marandaneto: blocking: `eventCaptured` is emitted after `before_send`, which may legally remove or replace `$session_id`. In that configuration this filter drops a matching initial `$pageview` captured during lazy loading, while the same event captured after the live listener is installed still activates the trigger. Please store the session ID alongside each event when buffering it, then filter using that immutable metadata rather than `event.properties.$session_id`.

marandaneto (approved):
left a last comment otherwise lgtm
@PostHog/team-replay might want to double check this

packages/browser/src/extensions/replay/session-recording.ts
arnohillen: <small>(created by claude code)</small>

Fixed in 62fc8149: the session id is now stamped from the sessionManager at buffer time and the filter compares that stamp, with a regression test for a before_send that strips $session_id.

arnohillen:
Resolved the merge conflict with main (147feda): union of both sides' additions to `terser-mangled-names.json` (`_eventsCapturedBeforeRecorderStarted` + `_eventsDroppedWhileIdle`); no source changes. Verified locally: typecheck, `check-mangled-property-consistency` (31/24/6 OK), full browser unit suite (127 suites, 5527 passed).

Answers to the incident-risk check's matched classes, diff vs main:

- Class 1 (version skew): the only cross-boundary addition is one new method on the eager extension, called via optional chaining from the chunk; both skew directions are covered by dedicated tests and the compat suite. No persisted-state shape changes (buffer is in-memory only).
- Class 2 (rotation/idle/flush): the buffer feeds trigger matching only, never ships snapshot data; it is capped at 100, filtered to the current session id, and released on every terminal path. A zero-interaction tab still ships zero recordings; the change only lets a trigger event captured during chunk load activate recording, same as if it fired after start (billing note in the description). The volume-invariant scenario tests from main pass on the merged tree.
- Class 8 (masking/consent): no masking or serialization paths touched; opt-out and `disable_session_recording` tear the buffer down before anything is consumed (tested), and `$snapshot` events are excluded from the buffer.

Remaining gate: re-review from @TueHaulund (all changes-requested threads addressed in 4ce1344/62fc814 and resolved; marandaneto approved after).

<small>(analysis and merge by Claude Fable 5, on Arno's behalf)</small>

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -108,9 +120,41 @@ export class SessionRecording implements Extension {
...
+        // a rotation or reset during the chunk load must not let a previous session's events activate this one
+        const sessionId = this._instance.sessionManager?.checkAndGetSessionAndWindowId(true)?.sessionId
+        return buffered.filter((entry) => entry.sessionId === sessionId).map((entry) => entry.event)
+    }
+
+    private _startBufferingPreStartEvents(): void {
+        if (this._removePreStartEventBufferHook || this._sessionRecordingDisposed) {
+            return
+        }
+        this._removePreStartEventBufferHook = this._instance.on?.('eventCaptured', (event) => {
```
TueHaulund, line 140: This hook is registered in the constructor, so it outlives the extension unless `dispose()` runs. `_replaceExtension` and the opt-out path in posthog-core both drop the instance without calling it, so the hook stays subscribed for the life of the page, holding the dead recorder and up to 100 buffered events with it. Opting back in then leaves two hooks running. This is the gap from my earlier review. The reply says it's covered, but I don't think it is.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1303,6 +1312,14 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         if (this.status === ACTIVE) {
             this._reportStarted(startReason || 'recording_initialized')
         }
+
+        // events captured while this chunk was loading never reached the live listener above, so a trigger
+        // on the initial $pageview could otherwise never match; replay them now that the recorder is running
+        if (eventTriggerCallback) {
```
TueHaulund, line 1327: The replayed events reach `_activateTrigger('event')`, which calls `_releaseHoldAndFlush()`. That hold is what stops a tab nobody touched from shipping a billable recording, per its own comment about prefetch and background loads. A `$pageview` fires at init on every load, so a customer with a `$pageview` trigger now releases the hold on prefetched and background tabs too. Every buffered event happened before the recorder started, so none of them are interaction evidence. Could the catch-up path match and scope triggers without releasing the hold, or release only when `_isIdle === false`? The incident check matched the rotation and idle class here, which is the same class as the July 2026 idle fix that billed zero-interaction recordings for 11 days.

TueHaulund (changes requested):

packages/browser/src/extensions/replay/session-recording.ts
arnohillen: Fixed in db1e962b. Worth reporting what I found reproducing it, because the mechanism turned out not to be the one described.

The capture hook itself does get released on both paths: `startIfEnabledOrStop`'s disabled branch calls `_stopBufferingPreStartEvents()` whenever `clientDisabled || serverDecided`, and both drop sites are consent-gated, so by the time `_replaceExtension` or the opt-out path runs the buffer hook is already gone. Instrumenting `_internalEventEmitter` across three consent flips, the `eventCaptured` subscriber count stays flat at 1 with and without a fix, so the "up to 100 buffered events" and "two hooks running" parts do not reproduce.

What does leak is the `visibilitychange` listener, which `stopRecording()` never removed. Same probe, counting document listeners across three flips: 4 added, 0 removed. So the conclusion holds (the instance is dropped without `dispose()` and outlives the page), the thing left behind is just the DOM listener rather than the capture hook, and it predates this PR.

`Extension` now has an optional `dispose()`, `_replaceExtension` calls it on the instance it evicts, and the cookieless opt-out path calls it instead of `stopRecording()`. The probe now reads 4 added / 3 removed. Regression test in `consent.test.ts` asserts `added - removed === 1` after three flips and reads 4 without the fix.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
arnohillen: Agreed, and fixed in db1e962b. A buffered `$pageview` is not interaction evidence, and releasing on it would bill prefetched and background tabs.

Took the second suggestion rather than the first, since scoping without releasing is the split that is wanted: the replay is wrapped in `_isReplayingPreStartEvents`, and `_activateTrigger` skips only the `_releaseHoldAndFlush()` call while that flag is set. The trigger still matches, still persists activation, still scopes the session. `_flushBuffer` independently honours `_holdFlushUntilInteraction`, so nothing ships until a real interaction arrives, and when one does the existing path releases as usual and the recording is still playable from t=0. A live trigger match after the recorder has started is untouched and still releases.

Gating on `_isIdle === false` would not have worked: at that point in `start()` it is still `'unknown'`, so it would have suppressed the release permanently rather than deferring it.

Three tests cover it: a replayed pre-load event leaves the hold on and ships nothing, a real interaction afterwards releases and ships, and a live post-start match still releases. The stale "Note on billing" paragraph in the description has been rewritten.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1025,7 +1029,13 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
         // 'any' a URL trigger may already have satisfied the combined status, making this
         // activation a no-op — the hold release must still happen.
-        if (triggerType === 'event') {
+        //
+        // Replayed pre-start events are the exception: they prove the trigger matched, so the
+        // session is still scoped and activated below, but they were captured before the recorder
+        // existed and so are not interaction evidence. Releasing on them would ship a billable
+        // recording for a prefetched or background tab whose only event is the initial $pageview.
+        // A later real interaction releases the hold through the usual path.
+        if (triggerType === 'event' && !this._isReplayingPreStartEvents) {
```
TueHaulund, line 1038: 🤖 *Agent-drafted, reviewed by Tue.*

This check only covers V1. `_onTriggerActivated` at `:983` still releases the hold no matter what, and that's the path V2 uses — the strategy gets `_onTriggerActivated` at `:1187`, and its listener is the same callback we replay at `:1331`.

So with `version: 2` and trigger groups, a replayed `$pageview` still releases the hold on a tab nobody touched. Same problem as last time, other code path.

Better to put the check inside `_releaseHoldAndFlush()` so both paths are covered, and drop it from here. The `override*` callers aren't affected, they never run during the replay.

Also worth a test with `version: 2` — the new hold tests all run on V1, which is how this got missed.

marandaneto:
draft until comments and conflicts are addressed

PR #4433, opened 2026-08-05
fix(mcp): deliver the session handle on errors and capability gaps

gesh:
**PR 3 of 3** in the `_mcp_instructions` stack. Stacked on #4431. Two narrow cases where the session handle was withheld, both splitting one conversation across sessions.

Context: [MCP 2026-07-28](https://modelcontextprotocol.io/specification/2026-07-28/changelog) removed protocol-level sessions, so correlation depends entirely on the agent receiving and echoing an explicit handle. Anywhere we withhold it, the conversation fragments.

## 1. Errored results

The prompt-back was suppressed when `isError` was true:

```ts
if (resultObj.isError === true) return false   // removed
```

A tool that fails on the *first* call is exactly when the agent needs the handle — without it the retry starts a fresh conversation, so the failure and its fix land in different sessions. The only requirement for the text block is an array to append to.

## 2. `get_more_tools`

The virtual `reportMissing` tool was excluded from injection, so `$mcp_missing_capability` carried no handle and fell back to the transport session:

```
$mcp_tool_call            session=ses_4ee7c718…   ← the work
$mcp_missing_capability   session=ses_019fd1ec…   ← the gap it hit, orphaned
```

A capability gap only means something next to the work that hit it. Now both share a session. Its bespoke `context` argument is untouched.

One supporting fix: the virtual tool is pushed *after* the listing is cached, so it had no ownership entry at all — its calls silently resolved no handle. Now cached alongside the push. `injectedMissingCapabilityTool` became dead once the skip flag went away and is removed.

## 3. The two options are now independent

`resolveConversationId` used to take the missing-capability tool name and return no handle for it, which made *stitching* behaviour depend on *which tool* was called. It is now `resolveConversationId(enabled, args)`.

| `enableConversationId` | `reportMissing` | handle on tools | `get_more_tools` exists |
|---|---|:--:|:--:|
| false | false | ✗ | ✗ |
| false | true | ✗ | ✓ |
| true | false | ✓ | ✗ |
| true | true | ✓ | ✓ |

`enableConversationId` owns session stitching; `reportMissing` owns whether the virtual tool exists. When both are on they compose — the virtual tool stitches like any other.

`isMissingCapabilityTool` survives in one place only, and it is not about sessions: it stops us stripping `get_more_tools`' own `context` argument, which has different semantics from the injected one.

## Tests

475 → **485**. Two behaviour tests, four asserting the off state (errored results, `get_more_tools`, capability-gap events, and an agent-supplied `conversation_id` being ignored), and a four-row independence matrix.

Verified they bite: removing the `enableConversationId` gate fails **5 of 485**.

Five existing tests asserted the old behaviour and were updated — they encode the decisions being reversed here, so the diff is the point rather than incidental churn.

- [x] @posthog/mcp

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility

---

## Review follow-ups

- **The session assertion is labelled as a forward guard**, since on this branch alone both events share the in-memory session id whether or not the handle is wired — it only becomes load-bearing once #4428 lands.
- **Added a test that does bite here**: two different handles in one connection produce two different conversations. Anything falling back to the transport or in-memory session reports one id for both, so only the handle can separate them.
- The `products/mcp_analytics` note — sessions filter to `event = '$mcp_tool_call'`, so capability gaps still won't show alongside calls — is recorded as a backend follow-up.

### DIFF
--- a/.changeset/mcp-conversation-id-edge-cases.md
+++ b/.changeset/mcp-conversation-id-edge-cases.md
@@ -1,0 +1,8 @@
+---
+'@posthog/mcp': patch
+---
+
+Deliver the `conversation_id` session handle on errored tool results, and inject it into
+the virtual `get_more_tools` tool. A first call that fails no longer sends the agent's
+retry into a different conversation, and a reported capability gap now groups with the
+work that hit it.
--- a/packages/mcp/src/__tests__/conversation-id.test.ts
+++ b/packages/mcp/src/__tests__/conversation-id.test.ts
@@ -564,3 +564,50 @@ describe('conversation_id edge cases', () => {
     })
   })
 })
+
+describe('enableConversationId and reportMissing are independent', () => {
+  let server: HighLevelMCPServerLike
+  let client: any
+  let cleanup: () => Promise<void>
+
+  beforeEach(async () => {
+    resetTodos()
+    const setup = await setupTestServerAndClient()
+    server = setup.server
+    client = setup.client
+    cleanup = setup.cleanup
+  })
+
+  afterEach(async () => {
+    await cleanup()
+  })
+
+  /** [enableConversationId, reportMissing] → [handle on a real tool, virtual tool exists] */
+  const matrix: [boolean, boolean, boolean, boolean][] = [
+    [false, false, false, false],
+    [false, true, false, true],
+    [true, false, true, false],
+    [true, true, true, true],
+  ]
+
+  it.each(matrix)(
+    'enableConversationId=%s reportMissing=%s → handle=%s virtualTool=%s',
+    async (enableConversationId, reportMissing, expectHandle, expectVirtual) => {
+      instrument(server, fakePostHog(), { enableConversationId, reportMissing })
+
+      const { tools } = await client.request({ method: 'tools/list' }, ListToolsResultSchema)
+      const realTool = tools.find((t: any) => t.name === 'add_todo')
+      const virtual = tools.find((t: any) => t.name === 'get_more_tools')
+
+      // One option decides stitching, the other decides whether the tool exists.
+      expect(!!(realTool.inputSchema as any).properties.conversation_id).toBe(expectHandle)
+      expect(!!virtual).toBe(expectVirtual)
+
+      // When both are on they compose: the virtual tool stitches like any other.
+      if (expectVirtual) {
+        expect(!!(virtual.inputSchema as any).properties.conversation_id).toBe(expectHandle)
+        expect((virtual.inputSchema as any).properties.context).toBeDefined()
+      }
+    }
+  )
+})
--- a/packages/mcp/src/__tests__/conversation-id.test.ts
+++ b/packages/mcp/src/__tests__/conversation-id.test.ts
@@ -124,13 +124,15 @@ describe('conversation-id', () => {
   })
 
   describe('addConversationIdToTools', () => {
-    it('skips the get_more_tools tool', () => {
+    it('injects into get_more_tools too, so capability gaps join their session', () => {
       const tools = [
         { name: 'get_more_tools', description: 'report missing' },
         { name: 'other_tool', description: 'fine' },
       ]
       const result = addConversationIdToTools(tools)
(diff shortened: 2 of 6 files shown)


gesh:
Reviewed and ran this: 485 pass at this tip, 496 merged with #4428, `tsc --noEmit` clean, prettier clean. This one looks correct to me and both reversals are well argued — delivering the handle on a failed first call is clearly right, and decoupling `enableConversationId` from `reportMissing` removes a genuinely confusing coupling. I checked the five updated tests and they're real inversions with the assertions flipped rather than coverage quietly dropped, which is the right way to do it. The ownership-cache fix for the virtual tool is a good catch — I confirmed it's cached before `addConversationIdToTools` runs, so `get_more_tools` keeps its own `context` (ownership `false`, not stripped) while we own its `conversation_id`.

One note on what a test proves rather than what it does:

```ts
expect(missing?.sessionId).toBe(toolCall?.sessionId)
```

This passes on this branch even without the handle being wired to `$session_id` at all — both events fall back to the same in-memory session id on a single server instance over `InMemoryTransport`, so it can't distinguish "grouped by the handle" from "grouped because nothing rotated". It only becomes load-bearing once #4428 lands. Asserting on `missing?.conversationId === handle` (which the test already does, and which is the thing this PR actually controls) is the part carrying weight; the session assertion is worth either a comment saying it's a forward guard, or strengthening so it would fail if the handle were ignored — e.g. two different handles in the same connection producing two different session ids.

Unrelated to the diff, flagging since this PR is what makes `get_more_tools` part of the session story: `$mcp_missing_capability` now carries `$mcp_conversation_id`, but `products/mcp_analytics` groups sessions off `$session_id` and filters to `event = '$mcp_tool_call'`, so capability gaps still won't appear alongside the calls in the sessions UI even once they share a session. Backend-side follow-up, not something to fix here.

gesh:
Fair — the session assertion passes for the wrong reason on this branch.

Kept it, labelled as a forward guard, and added one that bites here: two different handles in one connection producing two different conversations. Anything falling back to the transport or in-memory session reports one id for both, so only the handle can separate them.

`products/mcp_analytics` filtering sessions to `$mcp_tool_call` is recorded as a backend follow-up — agreed it's not this PR.

lucasheriques (approved):
the isError reasoning (a failure and its retry belong in the same session) is the most convincing single argument in the stack, and the 2x2 independence matrix test is great. approving.

**`conversation-id.test.ts:384` — question:** flipping the on-error test removed the only coverage of the `minted && !delivered` clear branch, which is still reachable (minted handle, result with no content array, no declared output schema). if the delivered accounting regresses nothing fails now. worth one test for that cell?

**`conversation-id.test.ts:506` — nit:** the comment here (and in `output-instructions.test.ts:306`) says an arbitrary string is treated as no handle at all, but `extractConversationId` takes any non-empty string. that validation only exists in #4428. worth fixing the comments so nobody relies on shape validation this stack doesn't have.

gesh:
Both valid, both fixed (`b92d5f1`).

**`conversation-id.test.ts:384` — you were right that the cell went uncovered.** I checked how thoroughly before writing anything: deleting the clear branch outright left **all 493 tests passing**. So it wasn't thin coverage, it was none.

Added a test for it. The cell needs both channels to miss at once — no declared output schema so no mirror, and a result with no `content` array so no prompt-back — at which point the handle exists only in our event, describing a conversation the agent was never told about.

One note on how it's written: my first attempt passed for the wrong reason. Asserting `toolCall?.conversationId` is undefined is equally true when no event was captured at all, and replacing the handler after `instrument()` did exactly that. It now replaces the underlying handler first so our wrapper still runs, and asserts the event exists before asserting the id is absent. Verified it fails with the branch removed.

**`conversation-id.test.ts:506` and `output-instructions.test.ts:306` — fixed.** Those comments were mine and they were wrong: `extractConversationId` takes any non-empty string here, and the shape check is #4428's. They now say which branch adds it, so nobody reads them and relies on validation this stack doesn't have.

PR #4435, opened 2026-08-05
fix(replay): adopt stylesheets that arrive before the shadow root is attached

arnohillen:
## Problem

Reported through a customer support ticket: a Stencil web component styled exclusively via `shadowRoot.adoptedStyleSheets` renders completely unstyled ("mobile mode") in session replay, while users saw the correctly styled component.

The recording is fine: the `AdoptedStyleSheet` event is captured with all rules. The replayer drops it. When a full snapshot races the component's hydration, the host element is serialized before `attachShadow` runs, and the recorder emits the `AdoptedStyleSheet` event (via its intentional `setTimeout(0)` deferral) before the mutation that attaches the shadow children. On replay, `applyAdoptedStyleSheet` then:

1. skips constructing the `CSSStyleSheet` entirely, because `hostWindow` is only derived when the host already has a shadow root, and
2. never recovers, because the retry loop only retries adoption (a `styleMirror` lookup), never construction, and the recorder sends rules only once per `styleId`.

Result: shadow root present, `adoptedStyleSheets.length === 0`, all failures silent. This affects any site styling shadow DOM components with constructed stylesheets (Stencil, Lit). Upstream rrweb has the same bug.

## Changes

Three changes in `packages/rrweb/rrweb/src/replay/index.ts`:

- Derive the host window from `ownerDocument` for element hosts, so the stylesheet is constructed and registered in `styleMirror` even when the shadow root is not attached yet.
- Extend the retry condition so it also retries while adoption was impossible (element host without a shadow root), instead of only when sheets were missing from the mirror. The existing 10-retry backoff is unchanged.
- Track adoptions that could not happen yet in a `pendingAdoptedStyleSheets` map and complete them deterministically at the moment the shadow root is attached (in `applyMutation`, or after the virtual dom flush). This covers what the wall-clock retries cannot: a paused, buffering, or background-throttled player outlasting the ~4.5s retry budget, or the shadow-attaching mutation simply arriving later than that in replay time (raised by Greptile, and it fails in a real browser). The map is cleared wherever `styleMirror` is reset.

Two regression tests replay the exact stream shape from the affected recording (full snapshot with a non-shadow-host element, `AdoptedStyleSheet` at +10ms, shadow-attaching mutation at +20ms): one plays straight through, one pauses between the two events until the retry budget is exhausted and then resumes. Both fail without their respective fix (`adoptedStyleSheets.length` is 0) and pass with it; the existing adopted-stylesheet tests (which only exercise the mutation-first ordering, why this was never caught) still pass, as do `replayer.test.ts` and `fast-forward.test.ts` in full (62 tests).

Backwards compatible: no API changes; for streams where the shadow root exists when the event applies (every case that worked before), behavior is identical.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Ran `pnpm changeset` to generate a changeset file (authored manually: `.changeset/replay-adopted-stylesheet-shadow-root-race.md`, versions only `posthog-js` per rrweb convention)

## 🤖 Agent context

- Root cause investigated and fixed with PostHog Code (Claude) starting from a customer support ticket. The mechanism was confirmed before fixing: a puppeteer repro of the recorded event ordering showed `adoptedStyleSheets.length === 0` with the UA default link color, while the identical stream with the two incremental events swapped adopted all rules, proving pure order sensitivity.
- Alternatives ruled out during investigation: sandboxed-iframe `CSSStyleSheet` construction failure, `styleMirror.reset` on rebuild, node id remapping across snapshots, CSS parse errors, virtual-dom queue loss (the PostHog player runs `useVirtualDom: false`), and a real media-query match.
- Considered also fixing the partial-adoption assignment (an incomplete retry overwrites previously adopted sheets) but left it: it is upstream behavior, semantically "set the full list", and not needed for this bug.

---

### DIFF
--- a/.changeset/replay-adopted-stylesheet-shadow-root-race.md
+++ b/.changeset/replay-adopted-stylesheet-shadow-root-race.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): stop dropping adopted stylesheets that arrive before the host's shadow root is attached. When the recorder's full snapshot races a web component's hydration, the AdoptedStyleSheet event can be recorded before the mutation that attaches the host's shadow root. The replayer silently dropped those styles for the rest of the page view, so components styled via `shadowRoot.adoptedStyleSheets` (Stencil, Lit) rendered completely unstyled. The replayer now constructs the stylesheet even when the shadow root does not exist yet and keeps retrying adoption until it is attached.
--- a/packages/rrweb/rrweb/src/replay/index.ts
+++ b/packages/rrweb/rrweb/src/replay/index.ts
@@ -157,6 +157,10 @@ export class Replayer {
   // was attached, keyed by node id; adopted when the shadow root appears.
   private pendingAdoptedStyleSheets: Map<number, number[]> = new Map();
 
+  // Latest adoption per host, keyed by node id; a wall-clock retry armed by
+  // an older AdoptedStyleSheet event must not overwrite a newer one.
+  private adoptedStyleSheetTokens: Map<number, object> = new Map();
+
   // Used to track video & audio elements, and keep them in sync with general playback.
   private mediaManager: MediaManager;
 
@@ -380,6 +384,7 @@ export class Replayer {
       this.mirror.reset();
       this.styleMirror.reset();
       this.pendingAdoptedStyleSheets.clear();
+      this.adoptedStyleSheetTokens.clear();
       this.mediaManager.reset();
       this.lastScrollMap.clear();
     };
@@ -669,6 +674,7 @@ export class Replayer {
     this.mirror.reset();
     this.styleMirror.reset();
     this.pendingAdoptedStyleSheets.clear();
+    this.adoptedStyleSheetTokens.clear();
     this.mediaManager.reset();
     this.resetCache();
 
@@ -932,6 +938,7 @@ export class Replayer {
           this.mediaManager.reset();
           this.styleMirror.reset();
           this.pendingAdoptedStyleSheets.clear();
+          this.adoptedStyleSheetTokens.clear();
           this.rebuildFullSnapshot(event, isSync);
           // 'instant' so the offset is not animated when the page sets scroll-behavior: smooth
           this.iframe.contentWindow?.scrollTo({
@@ -2326,6 +2333,9 @@ export class Replayer {
   private applyAdoptedStyleSheet(data: adoptedStyleSheetData) {
     const targetHost = this.mirror.getNode(data.id);
     if (!targetHost) return;
+    // supersede retries still pending from an older event for this host
+    const token = {};
+    this.adoptedStyleSheetTokens.set(data.id, token);
     // Create StyleSheet objects which will be adopted after.
     data.styles?.forEach((style) => {
       let newStyleSheet: CSSStyleSheet | null = null;
@@ -2361,6 +2371,8 @@ export class Replayer {
     const MAX_RETRY_TIME = 10;
     let count = 0;
     const adoptStyleSheets = (targetHost: Node, styleIds: number[]) => {
+      // a newer AdoptedStyleSheet event for this host supersedes this one
+      if (this.adoptedStyleSheetTokens.get(data.id) !== token) return;
       const stylesToAdopt = styleIds
         .map((styleId) => this.styleMirror.getStyle(styleId))
         .filter((style) => style !== null) as CSSStyleSheet[];
--- a/packages/rrweb/rrweb/src/replay/index.ts
+++ b/packages/rrweb/rrweb/src/replay/index.ts
@@ -153,6 +153,10 @@ export class Replayer {
   // Used to track StyleSheetObjects adopted on multiple document hosts.
   private styleMirror: StyleSheetMirror = new StyleSheetMirror();
 
+  // Hosts whose AdoptedStyleSheet event was applied before their shadow root
+  // was attached, keyed by node id; adopted when the shadow root appears.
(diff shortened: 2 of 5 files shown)


arnohillen:
[arno's agent] Valid catch, addressed in f775fb7c2. The wall-clock retries could indeed be outlasted by a paused, buffering, or background-throttled player (and their ~4.5s budget also capped how late the shadow-attaching mutation may arrive in replay time). Adoptions that cannot happen yet are now tracked in a `pendingAdoptedStyleSheets` map and completed deterministically when `applyMutation` (or the virtual dom flush) attaches the shadow root; the retry loop remains as a backstop for live-mode out-of-order sheets. Added a regression test that pauses between the two events for 5s (past the retry budget) and resumes: it fails without the map and passes with it.

packages/rrweb/rrweb/src/replay/index.ts
```diff
@@ -2334,18 +2364,31 @@ export class Replayer {
...
+      else this.pendingAdoptedStyleSheets.set(data.id, styleIds);
 
       /**
        * In the live mode where events are transferred over network without strict order guarantee, some newer events are applied before some old events and adopted stylesheets may haven't been created.
-       * This retry mechanism can help resolve this situation.
+       * The same applies to recorded streams when this event was emitted before the mutation that attaches the host's shadow root.
+       * This retry mechanism can help resolve these situations.
        */
-      if (stylesToAdopt.length !== styleIds.length && count < MAX_RETRY_TIME) {
+      if (
```
marandaneto, line 2400: **[P1] Cancel stale retries when a newer adoption event arrives.** Each event starts an independent backoff while the shadow root is absent. If the same host receives two `AdoptedStyleSheet` events before attachment, an older timer can run after the newer event and assign stale `styleIds` (and overwrite the pending map), leaving the replay with the obsolete stylesheet list. Since the map now handles missing roots deterministically, avoid retrying solely for `!adopted`, or generation-guard/cancel retries per host.

packages/rrweb/rrweb/src/replay/index.ts
```diff
@@ -153,6 +153,14 @@ export class Replayer {
   // Used to track StyleSheetObjects adopted on multiple document hosts.
   private styleMirror: StyleSheetMirror = new StyleSheetMirror();
 
+  // Hosts whose AdoptedStyleSheet event was applied before their shadow root
+  // was attached, keyed by node id; adopted when the shadow root appears.
+  private pendingAdoptedStyleSheets: Map<number, number[]> = new Map();
```
dustinbyrne, line 158: i looked into whether or not style sheet ordering was applied correctly (yes, for our usage)

though, agent noted that rrweb has something called `liveMode` which is a configuration option on this `Replayer` class. if that's on, we'd need to consider timestamp sorting instead of relying on event ordering.

we're not using `liveMode`, so this is non-blocking and more of a consideration - i don't have a sense of whether or not it's something we'd use in the future

.changeset/replay-adopted-stylesheet-shadow-root-race.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): stop dropping adopted stylesheets that arrive before the host's shadow root is attached. When the recorder's full snapshot races a web component's hydration, the AdoptedStyleSheet event can be recorded before the mutation that attaches the host's shadow root. The replayer silently dropped those styles for the rest of the page view, so components styled via `shadowRoot.adoptedStyleSheets` (Stencil, Lit) rendered completely unstyled. The replayer now constructs the stylesheet even when the shadow root does not exist yet and keeps retrying adoption until it is attached.
```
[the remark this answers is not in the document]
dustinbyrne: don't do this

dustinbyrne (approved):
approving to unblock

packages/rrweb/rrweb/src/replay/index.ts
arnohillen: [arno's agent] Fixed in 9e35875: each applyAdoptedStyleSheet call now sets a per-host token that pending retries check before running, so a newer event supersedes older retry chains (regression test: "does not let a stale adoption retry overwrite a newer stylesheet list").

packages/rrweb/rrweb/src/replay/index.ts
```diff
@@ -2331,21 +2371,36 @@ export class Replayer {
...
-       * This retry mechanism can help resolve this situation.
+       * The same applies to recorded streams when this event was emitted before the mutation that attaches the host's shadow root.
+       * This retry mechanism can help resolve these situations.
        */
-      if (stylesToAdopt.length !== styleIds.length && count < MAX_RETRY_TIME) {
+      if (
+        (!adopted || stylesToAdopt.length !== styleIds.length) &&
+        count < MAX_RETRY_TIME
+      ) {
         setTimeout(
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 5f714ca: the retry is now scheduled via this.addTimeout so destroy() cancels pending retries.

packages/rrweb/rrweb/src/replay/index.ts
arnohillen: [arno's agent] Makes sense; this PR keeps the existing last-applied-event-wins semantics, so liveMode ordering is unchanged. If we ever turn liveMode on we'd need timestamp-aware adoption on top.

marandaneto (approved):

TueHaulund (approved):

PR #4436, opened 2026-08-05
feat(surveys): add optional intro screen shown before the first question

jakesciotto:
## Problem

Surveys can show a confirmation ("thank you") screen after the last question, but there is no equivalent screen before the first one — popover surveys drop respondents straight into question 1 with no framing. The only workaround is a throwaway first question, which pollutes response data.

Requested in PostHog/posthog#74064 (see the issue for full context and cross-repo sequencing). This is the critical-path piece: surveys are rendered by the SDKs, so the main-app config UI is inert until this ships.

## Why

Give survey authors a way to set context before question 1 without recording a phantom question, mirroring the existing confirmation message.

## Changes

An optional intro screen, off by default, shown before question 0 — the leading mirror of the trailing confirmation message:

- New `SurveyAppearance` fields in `@posthog/core`: `displayIntroScreen`, `introScreenHeader`, `introScreenDescription`, `introScreenDescriptionContentType`, `introScreenButtonText`; matching `SurveyTranslation` fields, applied by the shared translation helper.
- Browser: new `IntroScreen` component rendered as a sibling branch to `Questions`/`ConfirmationMessage` in `SurveyPopup`. Advancing flips local state only — no event, no response, no localStorage write, and no effect on completion/partial-response accounting. The intro is skipped when a survey is resumed with in-progress answers (including URL prefill), and the confirmation branch always wins for completed surveys. Dismissing (X) from the intro fires the normal `survey dismissed` event.
- Preview bundle: `renderSurveysPreview` accepts `previewPageIndex: -1` for the intro page, exported as `INTRO_SCREEN_PREVIEW_INDEX` (also usable by the main app as a capability check). `Questions` defensively ignores negative preview indices.
- React Native parity: `IntroMessage` component + `SurveyModal` wiring + theme defaults.

Tested: unit suites pass (`test:unit:surveys` 513 tests, core 16, RN SurveyModal 9), typecheck + lint + full `pnpm build` (including mangled-property consistency checks) pass locally. The new Playwright spec (`playwright/mocked/surveys/intro-screen.spec.ts`) was written to mirror `thank-you-message.spec.ts` but has **not** been executed locally — please run `pnpm playwright:surveys` in CI/locally.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [ ] ~~posthog-ios / posthog-android / posthog-flutter parity~~ — tracked as follow-ups in PostHog/posthog#74064

### If releasing new changes

- [x] Changeset file added (`.changeset/survey-intro-screen.md`, minor bump)

## 🤖 Agent context

- Authored with PostHog Code (Claude) from PostHog/posthog#74064, which specifies the design (appearance-field model mirroring the confirmation message).
- Key decision: the intro is a **boolean-gated sibling branch** in `SurveyPopup` with a `-1` preview sentinel, rather than shifting question indices by +1 — questions keep their natural indices and branching logic is untouched. `getNextSurveyStep` is never called with the intro index.
- Enter advances the intro (non-preview only); Escape is deliberately not bound, since on the intro it would dismiss the whole survey rather than close a completed one.
- Companion main-app PR (editor UI, backend fields, version gating) follows; the main app's SDK version-gating entry needs the release version of this PR.

---


packages/browser/src/extensions/surveys/components/IntroScreen.tsx
```diff
@@ -0,0 +1,73 @@
...
+        // explicit action via the cancel button. In preview mode the parent owns navigation, and a
+        // window-level listener would hijack Enter presses in the surrounding editor UI.
+        if (isPreviewMode) {
+            return
+        }
+        const handleKeyDown = (event: KeyboardEvent) => {
+            if (event.key === 'Enter') {
+                event.preventDefault()
+                onStart()
+            }
```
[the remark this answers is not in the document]
jakesciotto: Fixed in 37841eb — removed the window-level key handling from IntroScreen entirely. A global Enter listener can't reliably tell a host-page keypress from a survey interaction (shadow-DOM retargeting makes target checks brittle), and the advance button is already natively keyboard-operable when focused. The unit test now asserts the opposite: a window-level Enter does not skip the intro.

marandaneto:
@PostHog/team-surveys since its a new feat

dustinbyrne:
i'd be concerned about an additional click affecting response rates, what do you think @lucasheriques

lucasheriques:
i think if it's optional, i don't mind it. @jakesciotto did this come because of a use case you need? or came from a customer too? can you also ship the PRs for the monorepo, as well as other mobile SDKs to maintain feature parity too?

jakesciotto:
@lucasheriques a customer brought it to me (beyond pricing) and it made reasonable sense to me when i went through it myself, i am happy to roll up the rest of the surfaces for feature parity

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1334,6 +1342,19 @@ export function SurveyPopup({
...
+    /**
+     * The intro screen is a leading page, the mirror of the trailing confirmation message. It is
+     * skipped whenever the survey already has answers in progress (resumed session or URL
+     * prefill), and dismissing it only flips local state — no event, no response, no effect on
+     * completion or partial-response accounting. The confirmation check above always wins so a
+     * completed survey never shows the intro.
+     */
+    const shouldShowIntroScreen = isPreviewMode
+        ? previewPageIndex === INTRO_SCREEN_PREVIEW_INDEX
+        : !!survey.appearance?.displayIntroScreen && !introScreenDismissed && !hasInProgressState
```
lucasheriques: `displayIntroScreen: true` with a blank header and description renders an empty box with a "Get started" button in it. `IntroScreen` guards both fields, so there's nothing left to draw.

The thank-you screen dodges this by defaulting its header to `'Thank you!'`. The intro has no floor like that, and the SDK has to survive whatever the API hands it:

```suggestion
    const hasIntroContent = !!(survey.appearance?.introScreenHeader || survey.appearance?.introScreenDescription)
    const shouldShowIntroScreen = isPreviewMode
        ? previewPageIndex === INTRO_SCREEN_PREVIEW_INDEX
        : !!survey.appearance?.displayIntroScreen &&
          hasIntroContent &&
          !introScreenDismissed &&
          !hasInProgressState
```

With a test on it.

lucasheriques (commented):
Good call on the shape. Keeping the intro as a sibling branch instead of shifting question indices by +1 means branching, `visitedIndices` and partial-response accounting all stay untouched.

One change before I approve, inline below.

Two calls from our side so you're not guessing:

`survey shown` stays as-is. It fires while the intro is on screen, so shown -> sent will drop for any survey that enables this. Correct behaviour, and we'll note it in the editor.

For the editor PR: at `previewPageIndex: -1` this renders the intro whether or not `displayIntroScreen` is on, mirroring how `questions.length` renders the confirmation regardless of `displayThankYouMessage`. The editor needs to stop sending `-1` when the toggle is off.

packages/core/src/surveys/translations.ts
```diff
@@ -219,6 +225,18 @@ export function applySurveyTranslation<
           translated.appearance.thankYouMessageCloseButtonText = translation.thankYouMessageCloseButtonText
           hasTranslation = true
         }
+        if (!isUndefined(translation.introScreenHeader)) {
```
lucasheriques: We'll keep adding appearance copy here, so I don't want to leave it at eight copies of this block plus a matching eight-term boolean in `hasAppearanceTranslation` plus an eight-field type. Three edit sites per field, and nothing fails if you miss one.

A field list collapses all three:

```ts
const TRANSLATABLE_APPEARANCE_FIELDS = [
  'thankYouMessageHeader',
  'thankYouMessageDescription',
  'thankYouMessageCloseButtonText',
  'introScreenHeader',
  'introScreenDescription',
  'introScreenButtonText',
  'submitButtonText',
  'backButtonText',
] as const
```

`hasAppearanceTranslation` becomes a `.some()`, the apply block a loop. Follow-up is fine, but I want it done.

packages/browser/src/extensions/surveys.tsx
jakesciotto: Fixed in fa52b66 — exactly the suggested gate (`hasIntroContent` on header/description), with a unit test asserting a copy-less intro skips straight to question 1. Also propagated the same skip to the React Native modal here, and to the iOS (92558a7), Android (7017850), and Flutter (2229a6b) PRs so all renderers agree.

packages/core/src/surveys/translations.ts
jakesciotto: Done in fa52b66 rather than as a follow-up: `TRANSLATABLE_APPEARANCE_FIELDS` now drives `hasAppearanceTranslation` (a `.some()`), the apply loop, and `TranslatableSurveyAppearance` (mapped type), so adding a copy field is one list entry. Existing translation tests unchanged and green (16/16).

lucasheriques (approved):

jakesciotto:
@lucasheriques all three points are in:

- **Inline change**: done in fa52b66 — the `hasIntroContent` gate with a test, and the same skip propagated to the RN modal here plus the iOS/Android/Flutter PRs so every renderer agrees.
- **`survey shown` stays as-is**: no change made. The editor now says so where the intro is configured — the panel notes the survey counts as shown while the intro is on screen, so response rates include people who never start the questions ([PostHog/posthog@ef4f6a9](https://github.com/PostHog/posthog/commit/ef4f6a9e64bb5851280b322ddbcfb7ca0e961872) on PostHog/posthog#78445).
- **Editor must not send `-1` when the toggle is off**: same commit — a shared `clampPreviewPageIndex` falls back to question 0 in the form preview, sidebar preview, and appearance modal (the wizard and hosted canvas already clamped), with unit tests.

lucasheriques:
@jakesciotto there's no trunk merge here btw, only on the monorepo. you can merge once CI is unblocked (github on 🔥)

PR #4440, opened 2026-08-05
chore(ci): add replay incident risk registry and advisory PR check

ksvat:
## Problem

Changes to posthog-js and the rrweb fork have repeatedly caused severe incidents: fleet-wide recording loss (Mar 2026), idle-tab over-recording (Aug 2026), fetch-wrapper regressions that broke customer sites (Jan 2026, May 2026), silent replay corruption (May 2026), and more. Most repeated through a small set of code-level patterns, and nothing warns an engineer that a diff resembles a past incident.

## Changes

- `.agents/skills/replay-incident-risk/INCIDENTS.md`: a registry of 7 incident classes with mechanisms and review questions. Internal impact numbers and customer names are omitted.
- `.agents/skills/replay-incident-risk/SKILL.md`: an agent/reviewer skill that cross-references a diff against the registry.
- `.agents/skills/replay-incident-risk/check.mjs` + `risk-map.json`: a dependency-free checker. Run locally with `node .agents/skills/replay-incident-risk/check.mjs [base-ref]`. Always exits 0.
- `.github/workflows/replay-incident-risk.yml`: runs the checker on PRs and maintains one advisory comment when the diff matches an incident class. Modeled on `bundled-size.yaml` (same pinned actions, fork-PR skip).

> [!NOTE]
> The check is advisory only. It flags resemblance to past incidents, it never fails the build.

This PR flags itself (it adds a workflow file, which matches the delivery-channel class). A stacked test PR exercises the recorder-path classes.

### Libraries affected

None, no package code changes and no version bump.

## Checklist

- [x] Tests for new code (checker validated against recent recorder commits)

## 🤖 Agent context

I (actually Claude) built this from a Slack sweep of past replay/SDK incidents. Skills invoked: writing-user-facing-copy, writing-pr-descriptions.

### DIFF
--- a/.agents/skills/replay-incident-risk/INCIDENTS.md
+++ b/.agents/skills/replay-incident-risk/INCIDENTS.md
@@ -1,0 +1,112 @@
+# Replay and SDK incident pattern registry
+
+A catalog of past incidents caused by changes to posthog-js and the rrweb fork, organized by the code-level pattern that caused them. Used by the `replay-incident-risk` skill and the `Replay incident risk` CI check to flag diffs that resemble a past incident.
+
+Each entry describes the mechanism, not the blast radius. Internal impact numbers, customer names, and postmortem contents are deliberately omitted; PostHog employees can find the detail through the linked incident channels and the internal post-mortems repo.
+
+---
+
+## Class 1: Version skew across the lazy-load boundary
+
+**The single most repeated cause.** The recorder, surveys, web-vitals, and tracing-headers extensions are lazy-loaded from the CDN. With `strict_script_versioning: false` (the default for most teams), the `?v=` query param is only a cache-buster: the CDN serves the **latest published bundle to every SDK version, including npm-pinned ones** (`external-scripts-loader.ts`). Publishing an extension change is a fleet-wide deploy with no rollout control.
+
+The canonical failure: new extension code assumes something only newer cores provide (a function, a config field, a persisted value), or changes a cross-boundary function signature. Old cores load the new bundle and break.
+
+Known incidents in this class:
+
+- **Recorder rejected persisted config from old cores (Mar 2026, INC-749).** The new recorder added a staleness check on persisted remote config keyed on a `cache_timestamp` field old cores never wrote. `timestamp ?? 0` made every saved config infinitely stale, so the recorder deleted it and never started. Only reproduced when the recorder started from persisted config before the live config fetch returned, a race that instant-mock CI never hit. Millions of recordings were lost over a weekend, unrecoverably. Reverted in [#3213](https://github.com/PostHog/posthog-js/pull/3213) (v1.360.0); introduced by [#3191](https://github.com/PostHog/posthog-js/pull/3191).
+- **Surveys stopped rendering for all older cores (Jul 2025).** A change added a second argument to `generateSurveys()`; old cores loading the new unversioned `surveys.js` passed `undefined`, treated as `false`. Fixed in [#2074](https://github.com/PostHog/posthog-js/pull/2074) by treating `undefined` as the old behavior.
+- **Surveys code called a function old cores don't install (Oct 2025).** Produced identical exceptions captured at very high volume in customer apps, inflating their error-tracking usage. Reverted, patched in v1.270.1 ([#2355](https://github.com/PostHog/posthog-js/pull/2355) was the trigger). Public postmortem: [post-mortems repo](https://github.com/PostHog/post-mortems).
+- **Web vitals capture broke for six days (Jan 2026, INC-695).** A non-backward-compatible release broke `$web_vitals` capture fleet-wide; no per-event-type volume alerting existed, so it ran for six days. Fixed in [#2973](https://github.com/PostHog/posthog-js/pull/2973).
+
+**Review questions for any diff touching `entrypoints/` bundles or `extensions/`:**
+
+1. Does this change a function signature, argument list, or return shape that is called across the core/extension boundary?
+2. Does new extension code call anything that might not exist in a core released a year ago?
+3. Does new code read persisted state (localStorage, cookies, persistence keys) that old cores wrote in a different shape? Absent fields must be treated as neutral, never as fatal or stale.
+4. Does the compat test suite (old core + new lazy bundles, added in [#3003](https://github.com/PostHog/posthog-js/pull/3003)) cover this path, including a cold start from persisted config with a slow remote-config response?
+
+## Class 2: Session rotation, idle handling, and flush semantics
+
+Changing when sessions rotate, how idle state is tracked, or when buffers flush directly changes **how many recordings ship**, and every shipped recording bills. There is a long history of session-rotation changes being painful in exactly this way.
+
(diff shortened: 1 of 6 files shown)


marandaneto:
whats about also expanding codewoners to all replay files? https://github.com/PostHog/posthog-js/blob/d0933e1bcb600957b82e37d72dab895ceed16e27/.github/CODEOWNERS#L4-L5

.agents/skills/replay-incident-risk/check.mjs
```diff
@@ -0,0 +1,127 @@
...
+    for (const candidate of [baseArg, 'origin/main', 'main'].filter(Boolean)) {
+        try {
+            git('rev-parse', '--verify', '--quiet', `${candidate}^{commit}`)
+            return candidate
+        } catch {
+            /* try next */
+        }
+    }
+    console.error('Could not resolve a base ref. Pass one explicitly, e.g. `check.mjs origin/main`.')
+    process.exit(2)
```
TueHaulund: This is meant to be advisory, but it can go red three ways:

1. This `process.exit(2)` contradicts the "Always exits 0" comment at the top. It surfaces because line 37 in the workflow is missing the `|| true` that line 38 has.
2. Any throw exits non-zero. Most likely one is `maxBuffer` at line 21 on a big diff.
3. `pathHits` prints with no limit at lines 113-116, while `contentHits` stops at 8 on line 119. Touch a few hundred files under `packages/browser/src/` and the comment goes over GitHub's 65536 char limit, the API 422s, and the job fails.

Cap the path list like `contentHits` is capped, and make the script always exit 0.

.github/workflows/replay-incident-risk.yml
```diff
@@ -0,0 +1,86 @@
...
+      - name: Setup Node.js
+        uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020 # v7.0.0
+        with:
+          node-version: 24
+          package-manager-cache: false
+
+      - name: Run incident-pattern check
+        id: check
+        run: |
+          BASE="origin/${{ github.event.pull_request.base.ref }}"
```
TueHaulund: Nit: `base.ref` goes straight into the shell script. Not really exploitable since it's the base branch and needs push access, but `env: BASE_REF: ${{ ... }}` then `BASE="origin/$BASE_REF"` costs nothing.

Same idea for the fixed `RISK_EOF` delimiter on lines 59-63, since the body includes lines from the PR diff. A random delimiter (`DELIM="RISK_$(openssl rand -hex 16)"`) is the usual habit.

.agents/skills/replay-incident-risk/risk-map.json
```diff
@@ -0,0 +1,159 @@
...
+    },
+    {
+      "id": "start-conditions-config",
+      "title": "Recording start conditions and remote config semantics",
+      "anchor": "class-6-recording-start-conditions-and-remote-config-semantics",
+      "paths": [
+        "^packages/browser/src/extensions/replay/external/triggerMatching\\.ts$",
+        "^packages/browser/src/extensions/replay/external/config\\.ts$",
+        "^packages/browser/src/extensions/replay/external/denylist\\.ts$",
+        "^packages/browser/src/extensions/replay/external/sampling",
```
TueHaulund: This one matches nothing. There's no `sampling` file under `packages/browser/src/extensions/replay/external/`, it's at `packages/browser/src/extensions/sampling.ts`.

The other 28 path patterns all resolve, so this is the only miss, but it's the failure mode the registry itself warns about: a dead pattern just silently never fires. Might be worth a cheap test that runs `git ls-files` once and asserts every `paths` entry has at least one match, plus that every `anchor` resolves to a heading in INCIDENTS.md.

.agents/skills/replay-incident-risk/risk-map.json
```diff
@@ -0,0 +1,159 @@
...
+        "^packages/browser/src/sessionid\\.ts$",
+        "^packages/browser/src/extensions/replay/session-recording\\.ts$",
+        "^packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder\\.ts$",
+        "^packages/browser/src/extensions/replay/external/recording-strategies\\.ts$",
+        "^packages/browser/src/extensions/replay/external/flushed-size-tracker\\.ts$",
+        "^packages/browser/src/extensions/replay/external/mutation-throttler\\.ts$"
+      ],
+      "contentPatterns": [
+        "_isIdle",
+        "isIdle",
```
TueHaulund, line 32: `isIdle` already covers `_isIdle` on the line above, so both fire and you get two hits for the same line.

More generally, content patterns are matched against every in-scope file regardless of which class's paths they belong to, so `setInterval`, `innerHTML`, `password`, `XMLHttpRequest` and `setAttribute\(` are going to fire on a lot of PRs that have nothing to do with those classes. A bot that flags most PRs stops getting read, including on the real Class 3 hit.

Cheapest fix is giving each class an explicit `contentScope` matching its own `paths`, the way `replay-reconstruction` and `delivery-channel` already do. Worth running this over the last ~50 merged PRs before turning it on to see the actual hit rate.

TueHaulund (commented):
Nice registry. Grouping by mechanism instead of by incident makes it actually reusable, and the review questions are concrete enough to answer by walking the code.

Main thing to fix: the check is documented as advisory but has three ways to fail the build. Plus a dead path pattern and some noise concerns. Comments inline.

.agents/skills/replay-incident-risk/INCIDENTS.md
```diff
@@ -0,0 +1,142 @@
...
+
+A catalog of past incidents caused by changes to posthog-js and the rrweb fork, organized by the code-level pattern that caused them. Used by the `replay-incident-risk` skill and the `Replay incident risk` CI check to flag diffs that resemble a past incident.
+
+Each entry describes the mechanism, not the blast radius. Internal impact numbers, customer names, and postmortem contents are deliberately omitted; PostHog employees can find the detail through the linked incident channels and the internal post-mortems repo.
+
+---
+
+## Class 1: Version skew across the lazy-load boundary
+
+**The single most repeated cause.** The recorder, surveys, web-vitals, and tracing-headers extensions are lazy-loaded from the CDN. With `strict_script_versioning: false` (the default for most teams), the `?v=` query param is only a cache-buster: the CDN serves the **latest published bundle to every SDK version, including npm-pinned ones** (`external-scripts-loader.ts`). Publishing an extension change is a fleet-wide deploy with no rollout control.
```
dustinbyrne, line 11: I think we can default `strict_script_versioning` to true now, but I'm still weary of edge cases. It's a scary change because we don't have a great way of measuring the outcome. I'm a little uncomfortable with that.

We can likely get away with attempting to load the versioned script and falling back to the old behavior if it fails. The downside is that any failure cases would incur longer load times. I'd _assume_ that the first load of the versioned asset would work for >99% of cases, but I have no hard data to back that up.

I'll let you know what ends up coming of this.

arnohillen (approved):

ksvat:
* **#4440** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4440?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4440?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ksvat:
### Merge activity

* **Aug 7, 9:01 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/4440).

PR #4442, opened 2026-08-05
chore(replay): add recording volume invariant tests

ksvat:
## Problem

The recorder's idle and rotation tests assert the mechanics of the current fix (held epochs, releases). No test asserts the billing-level outcome: how many recordings a tab ships over a long timeline. A future change could ship recordings through a different mechanism and pass every existing test. That is the shape of the Aug 2026 over-recording incident, where idle tabs shipped one billed recording per rotation.

## Changes

- Adds a `recording volume invariants` block to `lazy-sessionrecording.test.ts` with five scenario tests over multi-day fake-timer timelines.
- Idle rotation is exercised through both real paths: the recorder's own readOnly per-emit check (rotates at the 24 hour cap), and external rotations from the session manager (the incident path, one rotation per idle timeout).
- Scenarios: purely idle tab ships nothing over three days; externally-rotated idle tab ships nothing over three days; one interaction ships exactly one session; an interaction after a day of held rotations ships one session, not the backlog; a continuously active tab (activity inside the 30 minute timeout) rotates at the 24 hour cap and no shipped session spans longer.
- Each idle helper asserts its rotations actually happened, so a silent no-op loop cannot pass.
- The tests count `$session_id` values in `$snapshot` captures, not internal recorder state, so they survive refactors of the hold mechanism.

Mutation verification: disabling the `holdNextEpoch` decisions fails four tests; disabling `sessionPastMaximumLength` in `sessionid.ts` fails the organic-idle and active-tab tests. Every test fails under at least one mutation. All 331 tests in the file pass in under 1 second.

Companion to #4440, which documents these invariants in the incident registry.

### Libraries affected

None, test-only change, no version bump.

## Checklist

- [x] Tests for new code

## 🤖 Agent context

I (actually Claude) reworked these after review. Skills invoked: writing-tests, writing-pr-descriptions.

### DIFF
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -2053,6 +2053,79 @@ describe('Lazy SessionRecording', () => {
                 })
             })
 
+            describe('recording volume invariants', () => {
+                // Billing-level invariants over long simulated timelines. The mechanics of
+                // holding rotation-born sessions are covered above; these assert the outcome
+                // that matters regardless of mechanism: how many recordings a tab ships.
+
+                function shippedSessionIds(): Set<string> {
+                    return new Set(
+                        (posthog.capture as Mock).mock.calls
+                            .filter(([eventName]) => eventName === '$snapshot')
+                            .map(([, properties]) => properties.$session_id)
+                    )
+                }
+
+                function emitInactiveWithoutAssertion(activityTimestamp: number): void {
+                    _emit({
+                        event: 123,
+                        type: INCREMENTAL_SNAPSHOT_EVENT_TYPE,
+                        data: { source: 0, adds: [], attributes: [], removes: [], texts: [] },
+                        timestamp: activityTimestamp,
+                    })
+                }
+
+                function runIdleRotations(fromTimestamp: number, days: number): void {
+                    let timestamp = fromTimestamp
+                    const endTimestamp = fromTimestamp + days * 24 * 60 * 60 * 1000
+                    let rotationCount = 0
+                    while (timestamp < endTimestamp) {
+                        timestamp += sessionManager['_sessionTimeoutMs'] + 1000
+                        rotationCount++
+                        sessionIdGeneratorMock.mockImplementation(() => `volume-rotation-${rotationCount}`)
+                        jest.setSystemTime(new Date(timestamp))
+                        sessionManager.checkAndGetSessionAndWindowId(false, timestamp)
+                        emitInactiveWithoutAssertion(timestamp + 10)
+                        jest.advanceTimersByTime(RECORDING_BUFFER_TIMEOUT)
+                    }
+                }
+
+                beforeEach(() => {
+                    jest.useFakeTimers().setSystemTime(new Date(startingTimestamp))
+                })
+
+                it('a tab with zero user interaction ships zero recordings across three days of rotations', () => {
+                    runIdleRotations(startingTimestamp, 3)
+
+                    expect(shippedSessionIds()).toEqual(new Set())
+                })
+
+                it('a single interaction ships exactly one session and later idle rotations add none', () => {
+                    emitActiveEvent(startingTimestamp + 100)
+                    jest.advanceTimersByTime(RECORDING_BUFFER_TIMEOUT)
+                    expect(shippedSessionIds()).toEqual(new Set([sessionId]))
+
+                    runIdleRotations(startingTimestamp + 100, 2)
+
+                    expect(shippedSessionIds()).toEqual(new Set([sessionId]))
+                })
+
+                it('no flushed batch spans more than the 24 hour session age cap', () => {
+                    emitActiveEvent(startingTimestamp + 100)
+                    runIdleRotations(startingTimestamp + 100, 2)
+
+                    const dayInMillis = 24 * 60 * 60 * 1000
+                    const snapshotCalls = (posthog.capture as Mock).mock.calls.filter(
+                        ([eventName]) => eventName === '$snapshot'
+                    )
+                    expect(snapshotCalls.length).toBeGreaterThan(0)
+                    for (const [, properties] of snapshotCalls) {
+                        const timestamps = properties.$snapshot_data.map((event: eventWithTime) => event.timestamp)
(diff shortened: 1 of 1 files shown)


packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -2053,6 +2053,79 @@ describe('Lazy SessionRecording', () => {
...
+                    runIdleRotations(startingTimestamp + 100, 2)
+
+                    const dayInMillis = 24 * 60 * 60 * 1000
+                    const snapshotCalls = (posthog.capture as Mock).mock.calls.filter(
+                        ([eventName]) => eventName === '$snapshot'
+                    )
+                    expect(snapshotCalls.length).toBeGreaterThan(0)
+                    for (const [, properties] of snapshotCalls) {
+                        const timestamps = properties.$snapshot_data.map((event: eventWithTime) => event.timestamp)
+                        expect(Math.max(...timestamps) - Math.min(...timestamps)).toBeLessThanOrEqual(dayInMillis)
```
marandaneto: This assertion does not actually exercise the 24-hour session-age invariant. `runIdleRotations` forces a non-read-only session check every session timeout, so these sessions rotate through the inactivity path, while this loop only measures the timestamp span within each individual capture batch (normally flushed every ~2 seconds). In practice the captured batch can just be the initial interaction, making the assertion trivially pass. Could we instead drive non-interactive/read-only events past the 24-hour cap, assert that the session ID rotates, then release the held epoch with an interaction and aggregate event timestamps by `$session_id` across all capture calls?

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -2053,6 +2053,79 @@ describe('Lazy SessionRecording', () => {
...
+
+                function runIdleRotations(fromTimestamp: number, days: number): void {
+                    let timestamp = fromTimestamp
+                    const endTimestamp = fromTimestamp + days * 24 * 60 * 60 * 1000
+                    let rotationCount = 0
+                    while (timestamp < endTimestamp) {
+                        timestamp += sessionManager['_sessionTimeoutMs'] + 1000
+                        rotationCount++
+                        sessionIdGeneratorMock.mockImplementation(() => `volume-rotation-${rotationCount}`)
+                        jest.setSystemTime(new Date(timestamp))
```
TueHaulund, line 2110: This rotates the session by hand, so the recorder never gets tested.

The recorder already calls `checkAndGetSessionAndWindowId` on every emit (lazy-loaded-session-recorder.ts L2220-2223). That call is what stops an idle tab from building up a multi-day recording. Since the helper rotates first, you could delete that call from the source and all three tests still pass.

Easiest fix: remove this line and let `emitInactiveWithoutAssertion` cause the rotation on its own.

Also `readOnly` should be `true` here. The recorder passes `!isUserInteraction`, and these are inactive events. With `false` you bump `_sessionActivityTimestamp`, so this simulates a user active every 30 minutes rather than an idle tab.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -2053,6 +2053,79 @@ describe('Lazy SessionRecording', () => {
...
+                    }
+                }
+
+                beforeEach(() => {
+                    jest.useFakeTimers().setSystemTime(new Date(startingTimestamp))
+                })
+
+                it('a tab with zero user interaction ships zero recordings across three days of rotations', () => {
+                    runIdleRotations(startingTimestamp, 3)
+
```
TueHaulund, line 2132: An empty set doesn't prove much. You'd also get it if nothing was captured at all, if `_emit` threw, or if the loop never ran. Nothing checks that the ~144 rotations happened.

Return `rotationCount` from `runIdleRotations` and assert on it, or add `expect(sessionIdGeneratorMock).toHaveBeenCalledTimes(rotationCount)`. Same for L2109.

Bonus: the loop count comes from `session_idle_timeout_seconds`. If that default ever changes, this catches it instead of quietly running many more iterations.

packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
```diff
@@ -2053,6 +2053,79 @@ describe('Lazy SessionRecording', () => {
...
+                it('a single interaction ships exactly one session and later idle rotations add none', () => {
+                    emitActiveEvent(startingTimestamp + 100)
+                    jest.advanceTimersByTime(RECORDING_BUFFER_TIMEOUT)
+                    expect(shippedSessionIds()).toEqual(new Set([sessionId]))
+
+                    runIdleRotations(startingTimestamp + 100, 2)
+
+                    expect(shippedSessionIds()).toEqual(new Set([sessionId]))
+                })
+
```
TueHaulund, line 2147: This test can't fail. Only one batch ever ships, the one holding the single active event. Everything after the first rotation is held by `_holdFlushUntilInteraction` and never captured, and `splitBuffer` only splits on size (7 MB), never on time.

So you're checking a span of a few hundred milliseconds against a 24 hour limit. Remove the age cap from the source and this still passes.

To make it real, send active events periodically for over 24 hours so `sessionPastMaximumLength` actually kicks in (the test at L1682 shows the setup) and tighten the limit. If that's more work than it's worth, I'd just delete this test.

TueHaulund (commented):
Worth adding: an idle tab that gets an interaction after several rotations should ship one session, not the whole held backlog.

That's the case most likely to break, since it's where `_holdFlushUntilInteraction` releases. Should be about two lines with the helper you already have.

arnohillen (approved):

ksvat:
* **#4442** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4442?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4442?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ksvat:
### Merge activity

* **Aug 7, 9:05 PM UTC**: @ksvat merged this pull request with [Graphite](https://app.graphite.com/github/pr/PostHog/posthog-js/4442).

PR #4443, opened 2026-08-06
fix(replay): close the stylesheet budget review findings shipped in #4320

arnohillen:
# Close the review findings the stylesheet budget merged without

## Problem

[#4320](https://github.com/PostHog/posthog-js/pull/4320) shipped the stylesheet inlining budget with three findings from the [customer review](https://github.com/PostHog/posthog-js/pull/4320#issuecomment-5147864308) unresolved: the 10,000-rule default was baked into the recorder chunk while its off-switch only existed in newer array.js bundles, so during version skew the behaviour change was live and the documented escape hatch was silently inert; deferral moved whole-sheet stringification off the snapshot but kept it as one long task, so a single large sheet still froze the main thread for up to a second, just later; and deferred CSS could be lost rather than delayed (cancelled at `stop()` or navigation with nothing flushed), while the new telemetry could not see either problem and the async delivery path had no checked-in tests.

## Changes

- **Budget default moved from rrweb into posthog-js.** The recorder chunk loads from an unversioned path, so a cached or npm-pinned array.js silently picked up the new default while its config allowlist dropped the user's `inlineStylesheetBudgetRules` override, including `0` to disable. The default (still 10,000) now lives in the session recording options object inside the same artifact that reads the user's override, so the budget only turns on where the off-switch works, and `rrweb.record()` itself is unbounded unless the caller opts in.
- **Intra-sheet chunking.** A resumable cursor (`createStylesheetTextCursor`) stringifies deferred sheets 200 rules per slice, re-checks the `requestIdleCallback` deadline between slices (spaced `setTimeout` ticks where rIC is unavailable), and emits each sheet's `_cssText` mutation atomically when its last slice completes. Output is byte-identical to the single pass, `@import` chains included, and no single task scales with sheet size any more.
- **Teardown flush plus SDK pagehide.** The deferred queue is flushed synchronously (bounded by a safety cap) on `stop()` and on `pagehide`; the SDK registers its own later `pagehide` listener that re-drains the compression queue and flushes the buffer so those final mutations actually ship. Docs and changeset now state the residual fidelity risk precisely instead of claiming "no CSS is dropped".
- **Telemetry gaps and charge split.** The cost window now wraps the whole synchronous `takeFullSnapshot` task (buffer unlock drain and document `adoptedStyleSheets` included, no longer misattributed to the mutation-batch timer). New session-cumulative counters cover the deferred path: `$sdk_debug_replay_deferred_stylesheets` (now cumulative, not read off the slowest snapshot) plus `_failed`, `_abandoned`, `$sdk_debug_replay_deferred_stylesheet_ms`, and `_slowest_slice_ms`. Never-deferrable rules (CSSOM-only `<style>`, `adoptedStyleSheets`) no longer charge the budget, and their bucket is reported separately (`$sdk_debug_replay_slowest_full_snapshot_css_rules_non_deferrable`) so CSSOM-dominated pages can tell which bucket they are in.
- **Regression tests for the async delivery machinery.** End-to-end tests drive a stubbed `requestIdleCallback` through a real `record()` session and assert deferred links keep `href` in the FullSnapshot, `_cssText` mutations carry exactly those mirror ids, a second snapshot before idle re-defers under the new ids and emits nothing for the old ones, slices never exceed 200 rules, and the multi-level `@import` chain from the review (A imports B imports C holding the bulk) is counted in full at the gate.

## Backwards compatibility

- Snippet and npm users keep the existing 10,000-rule default: it is applied by posthog-js when it starts the recorder, and `inlineStylesheetBudgetRules` (including `0` to disable) now works regardless of array.js version because the default and the allowlist live in the same artifact.
- Direct `rrweb.record()` consumers regain the pre-#4320 unbounded inlining unless they pass `inlineStylesheetBudgetRules` explicitly, which makes the rrweb-side change purely additive.
- No wire-format changes: deferred sheets still arrive as ordinary attribute mutations carrying `_cssText`, the same shape the replayer already rebuilds from, and the FullSnapshot representation of a deferred link (`rel`/`href` retained) is unchanged.

## Test strategy

- rrweb: 485 passed, 3 skipped (41 files), including a new 18-test `deferred-stylesheet-inlining` suite (slicing, cancellation, flush caps, pagehide, end-to-end id matching against a real `record()` session), a new `stylesheet-budget` suite pinning the unbounded default and explicit opt-in, and an extended `stylesheet-manager` suite (resumable cursor tasks, masking of emitted `_cssText`, failure counting).
- rrweb-snapshot: 249 passed, 1 skipped, including 20 `snapshot-cost` tests (charge split, cumulative deferred stats, mutation-drain attribution, multi-level `@import` chain counting) and new `createStylesheetTextCursor` tests proving slice output is byte-identical to the single pass, cycles included.
- browser (jest): 363 passed across `lazy-sessionrecording`, `lazy-sessionrecording-compression`, and `sessionRecording-onRemoteConfig`, covering the 10k default passthrough, `0` disabling the budget, raised budgets, cumulative deferred telemetry properties, and the pagehide drain-and-flush ordering.
- `tsc` clean for both rrweb packages; package-local eslint clean on every file this branch touches.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

---


packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -912,8 +1021,10 @@ function record<T = eventWithTime>(
       );
     }
     return () => {
-      cancelDeferredStylesheetInlining?.();
-      cancelDeferredStylesheetInlining = undefined;
+      // finish the deferred CSS while the emit path is still wired up, so the
+      // sheets this recording deferred don't silently vanish with it
+      deferredStylesheetInlining?.flush();
```
marandaneto: **[P1] Flush deferred CSS before the SDK's final drain**

This emits deferred `_cssText` mutations when rrweb stops, but `LazyLoadedSessionRecording.stop()` calls `_flushBuffer()` and `_clearBuffer()` before `_teardown()` reaches `_stopRrweb()`. These mutations therefore arrive after the final flush and are then left unscheduled or discarded by `_invalidateCompressionQueue()`. Please stop rrweb before the SDK's final drain/flush, or explicitly drain and flush again after this handler runs (including the async compression stop path).

packages/rrweb/rrweb-snapshot/src/utils.ts
```diff
@@ -215,6 +215,215 @@ export function isCSSStyleRule(rule: CSSRule): rule is CSSStyleRule {
...
+      // guard true import cycles, which the single-pass recursion has no
+      // terminating path for; the rule falls back to its `@import url(...)` form
+      const cyclic = imported && stack.some((f) => f.sheet === imported);
+      if (imported && !cyclic && openFrame(imported, rule)) {
+        // descend; frame.index advances when the child frame completes
+        return;
+      }
+      resolveImport(frame, rule, null);
+    } else {
+      let ruleStringified = rule.cssText;
```
marandaneto: **[P1] Grouping rules bypass the per-slice limit**

A `CSSMediaRule`, `CSSSupportsRule`, or other grouping rule is charged as one cursor step here, but reading its `cssText` synchronously serializes its entire nested rule tree. One grouping rule with thousands of children can therefore recreate the long main-thread freeze this slicing is intended to prevent. Please make grouping-rule children part of the resumable traversal rather than treating the whole subtree as one rule.

packages/rrweb/rrweb/src/record/stylesheet-manager.ts
```diff
@@ -58,30 +76,78 @@ export class StylesheetManager {
...
+      advance: (maxRules: number) => {
+        // deferred work runs outside the snapshot's tracking window, so the
+        // snapshot timers miss it; measure each slice (emit included) here
+        const startedAt = nowMs();
+        try {
+          if (!readyCursor.advance(maxRules)) {
+            return false;
+          }
+          const cssText = readyCursor.text();
+          if (!linkEl.isConnected) {
```
marandaneto, line 168: **[P2] Verify the stylesheet did not change while slicing**

The cursor was created from the `linkEl.sheet` instance present when this task began, but this completion check only verifies that the link is still connected. If an SPA changes the same link's `href` while slicing, this can emit CSS from the old sheet after the href mutation and override the new stylesheet in replay. Please retain the captured sheet and verify `linkEl.sheet` is still that instance before emitting, otherwise discard or restart the task.

packages/rrweb/rrweb-snapshot/src/snapshot-cost.ts
```diff
@@ -251,5 +348,8 @@ export function resetSnapshotCostState(): void {
   lastCost = null;
   stylesheetBudgetRules = null;
   deferredStylesheetLinks = [];
+  deferredLinkCount = 0;
+  nonDeferrableDepth = 0;
   mutationCost = { slowestBatchMs: 0 };
+  deferredStylesheetStats = emptyDeferredStats();
```
marandaneto, line 449: **[P2] Reset deferred stylesheet telemetry after the old recorder stops**

The session-change path currently calls `resetSnapshotCostState()` before `_restartForSessionIdChange()` stops the previous rrweb recorder. The old recorder's teardown flush can then record deferred duration, failures, or abandonment into these newly reset counters, attributing them to the new session. Please move the reset to after the old recorder has stopped and before the new recorder starts.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1680,13 +1691,19 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+            // the FullSnapshot event is emitted partway through rrweb's takeFullSnapshot,
+            // whose cost window only closes after the post-snapshot buffer drain and
+            // adoptedStyleSheets work; the sync read below picks up any snapshot we have
+            // not seen yet, and the microtask (which runs as soon as the synchronous
+            // snapshot task finishes, before any newer snapshot can replace the globals)
+            // picks up the one in progress right now
             this._trackFullSnapshotCost()
+            Promise.resolve()
+                .then(() => this._trackFullSnapshotCost())
+                .catch(() => {})
```
marandaneto: **[P2] Do not silently swallow snapshot telemetry errors**

This empty catch converts any programming error from `_trackFullSnapshotCost()` into success, making telemetry failures invisible. Please either let the error surface or handle it explicitly with a stable error signal/logging path rather than silently degrading the telemetry.

marandaneto (commented):
Review findings from the deferred stylesheet follow-up pass.

marandaneto:
moving to draft until comments are resolved

packages/rrweb/rrweb-snapshot/src/utils.ts
```diff
@@ -215,6 +215,215 @@ export function isCSSStyleRule(rule: CSSRule): rule is CSSStyleRule {
...
+      }
+      countStylesheetRules(rules);
+      let sheetHref = s.href;
+      if (!sheetHref && s.ownerNode) {
+        // an inline <style> element
+        sheetHref = s.ownerNode.baseURI;
+      }
+      stack.push({
+        sheet: s,
+        rules,
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 7319fefba: the cursor now snapshots the rule list (Array.from) when it opens a sheet or grouping frame, so slicing iterates a stable array and insertRule/deleteRule between slices can no longer skip, duplicate, or throw. Post-snapshot CSSOM mutations are intentionally not reflected (same semantics as the previous single synchronous pass at defer time) and are recorded by the StyleSheetRule observer, so replay stays correct. Regression test mutates the list between slices and failed before the fix.

packages/rrweb/rrweb/src/record/index.ts
arnohillen: [arno's agent] Fixed in 706708847, SDK-side as suggested: LazyLoadedSessionRecording.stop() now stops the recording producers first, so rrweb's teardown flush lands the deferred _cssText mutations in the buffer (or compression queue) before the drain check and the final _flushBuffer/_clearBuffer. The async compression stop path reads the queue promise after producers stop, so both paths ship the CSS. New SDK test asserts a deferred sheet pending at stop() reaches posthog.capture; it failed before the fix. No rrweb change was needed (rrweb already flushes before its own handler teardown).

packages/rrweb/rrweb-snapshot/src/utils.ts
arnohillen: [arno's agent] Fixed in 7319fefba: @media/@supports rules now push a resumable child frame like @import frames; children are stringified rule-by-rule against the slice budget and the frame completion assembles prelude { children }. Byte-identical to the single pass in the checked-in harness. @keyframes/@layer/@container keep the one-shot read (their preludes are not exactly rebuildable); a child failure fails the whole sheet, matching single-pass throw semantics. Tests: a 5000-rule @media sheet asserts bounded slices plus byte identity, and a giant @media whose own cssText getter throws proves the whole-subtree read never happens.

packages/rrweb/rrweb/src/record/stylesheet-manager.ts
arnohillen: [arno's agent] Fixed in 706708847: the task retains the captured CSSStyleSheet instance and verifies linkEl.sheet is still that instance before emitting; on mismatch the emit is discarded and counted via recordDeferredStylesheetFailure, since the new sheet is handled by its own mutation flow. Test swaps the sheet mid-slicing and asserts no mutation plus failedCount 1; it failed before the fix.

packages/rrweb/rrweb-snapshot/src/snapshot-cost.ts
arnohillen: [arno's agent] Fixed in f33b9b54e: the per-session cost resets (slowest-snapshot record, dedup marker, resetSnapshotCostState) moved from _onSessionIdCallback into _restartForSessionIdChange, between stop() and start(). The old recorder's teardown flush runs synchronously inside stop() on both stop paths, so its deferred duration/failures/abandonment are recorded before the reset and attributed to the old session. Test asserts the ordering (rrweb stop, then reset, then the new record() call) and failed before the fix.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
arnohillen: [arno's agent] Fixed in f33b9b54e: both the sync call and the microtask now go through _safeTrackFullSnapshotCost, which catches, logs once per recorder via the file's existing logger.error pattern, and keeps recording alive; the empty catch is gone (the previously unguarded sync call could also throw out of onRRwebEmit, now handled too). Test makes getLastSnapshotCost throw and asserts exactly one error log across two FullSnapshots with both events still buffered.

.changeset/stylesheet-budget-default-in-sdk.md
```diff
@@ -0,0 +1,12 @@
+---
+'posthog-js': patch
+'@posthog/types': patch
+---
```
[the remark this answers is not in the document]
dustinbyrne: invalid according to AGENTS.md

packages/rrweb/rrweb-snapshot/src/utils.ts
```diff
@@ -215,6 +215,337 @@ export function isCSSStyleRule(rule: CSSRule): rule is CSSStyleRule {
...
+      stack.push({
+        kind: 'sheet',
+        sheet: s,
+        // Snapshot the live CSSRuleList: application insertRule()/deleteRule()
+        // calls between idle slices would shift the numeric indexes the cursor
+        // resumes by, skipping or duplicating rules. Post-snapshot CSSOM
+        // mutations are therefore not reflected here - the same semantics as a
+        // single synchronous pass taken at defer time - and they are recorded
+        // separately by the StyleSheetRule observer, so replay still applies them.
+        rules: Array.from(rules),
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] The interleaving described here is benign: a capture taken after the mutation includes the mutated rules in the emitted _cssText, so replay converges on the post-mutation state (the earlier StyleSheetRule event just double-applies against the href-loaded sheet before the _cssText lands). The real exposure is the opposite order, a mutation landing after the cursor's capture, which the multi-slice window widened; fixed in 3cc5ab0 by abandoning the stale capture (the retained href serves the sheet and the rule event applies on top), covered by unit and end-to-end tests.

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -912,8 +1021,10 @@ function record<T = eventWithTime>(
...
+      deferredStylesheetInlining = undefined;
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 3cc5ab0: flush() now contains a throwing emit/mask callback, counts what was left as abandoned, and lets the stop()/pagehide teardown continue (regression test included).

packages/rrweb/rrweb/src/record/index.ts
```diff
@@ -167,19 +219,48 @@ function inlineDeferredStylesheets(
...
+        for (let i = index; i < links.length; i++) {
+          if (links[i] !== null) {
+            abandoned += 1;
+          }
+        }
+        recordDeferredStylesheetsAbandoned(abandoned);
+        task = null;
+      }
+      onDone();
+    },
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 3cc5ab0: flush() reserves one slice of budget per sheet still queued, so an oversized head sheet is abandoned once only the reservations remain and the cheap tail still ships (the [11k, 50, 50] case now salvages both small sheets; test added).

packages/rrweb/rrweb-snapshot/src/utils.ts
```diff
@@ -215,6 +215,337 @@ export function isCSSStyleRule(rule: CSSRule): rule is CSSStyleRule {
...
+  }
+  if (
+    typeof grouping.conditionText === 'string' &&
+    !('containerName' in grouping) &&
+    !isCSSStyleRule(rule)
+  ) {
+    return `@supports ${grouping.conditionText} {`;
+  }
+  return null;
+}
```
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 3cc5ab0: getGroupingRulePrelude now opens resumable frames for @layer blocks (named and anonymous) and @container rules too; @keyframes and nested style rules keep the one-shot read. Covered by new cursor tests.

packages/rrweb/rrweb-snapshot/src/utils.ts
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in f41bf70af. Investigated the observer first: getIdAndStyleId resolves sheet.ownerNode via the mirror (or the adopted-sheet styleMirror), and an @import'ed sheet has neither, so no StyleSheetRule event is ever emitted for imported sheets; capture-at-defer-time is the correct source of truth there and cannot overwrite anything. The real hazard was the root sheet (a StyleSheetRule event recorded between defer and emit replays before the later _cssText and would be overwritten), and the fix closes both: the cursor now snapshots the whole @import chain's rule lists eagerly at creation (defer time), and the insertRule/deleteRule/setProperty/removeProperty proxies call StylesheetManager.onCssomSheetMutation, which cancels any pending deferral whose captured root sheet matches; the link keeps its href fallback and the cancellation is counted. End-to-end tests cover insertRule and setProperty mid-deferral (no _cssText emitted, failure counted, teardown flush does not resurrect it); they failed before the fix.

packages/rrweb/rrweb/src/record/index.ts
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 3f5a7db73: deferredStylesheetInlining.flush() is wrapped in try/catch at both call sites (stop closure and pagehide listener), so a plain Error from user emit/maskAttributeFn can no longer abort teardown (callSafely was unsuitable since it re-throws non-SecurityErrors). Tests: a throwing emit during the stop-time flush leaves stop() non-throwing with full teardown completed (observers disconnected, recording false, suspension listeners removed), and a throwing emit during pagehide is contained. Both failed before the fix.

packages/rrweb/rrweb/src/record/index.ts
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in 3f5a7db73 with the skip-oversized policy: the cursor exposes remainingRules() (a lower bound that tightens as frames open) and flush() estimates each sheet's remaining slices before every slice; if it exceeds the remaining budget the sheet is abandoned immediately (counted, href fallback kept) and the next sheet proceeds, so cheap sheets are always salvaged. The per-slice re-check self-corrects underestimates once group/import frames open. Tests: [oversized 12k, 30, 30] emits both small sheets and counts one abandoned with zero budget spent on the oversized one; a sheet hiding 15k rules inside one @media is caught after its first slice.

packages/rrweb/rrweb-snapshot/src/utils.ts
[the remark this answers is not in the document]
arnohillen: [arno's agent] Fixed in cf2664584: @layer blocks, @container (rebuilt from containerName/containerQuery since the name is not reliably in conditionText), @keyframes, and natively-nested style rules now open resumable child frames like @media/@supports; the declaration block of a nested style rule rides in the prelude exactly once. Documented one-shot fallbacks remain only for names needing CSS escaping, @container without containerQuery support, @page, and prelude-less types (@scope, @starting-style). Tests use the throwing-cssText-getter mocks to prove no one-shot subtree read for every new type plus a real jsdom @keyframes byte-comparison; the Tailwind-style giant @layer case is covered by a bounded-slices assertion. Fail-before-fix proven for six of the new tests.

.changeset/stylesheet-budget-default-in-sdk.md
```diff
@@ -0,0 +1,12 @@
...
...optedStyleSheets`) no longer charge the inlining budget, since deferring `<link>` sheets buys those pages no freeze reduction. The telemetry now covers the gaps the review called out: the full-snapshot duration window wraps the whole synchronous `takeFullSnapshot` task (buffer unlock drain and document `adoptedStyleSheets` stringification included, and no longer misattributed to the mutation-batch timer), `$sdk_debug_replay_deferred_stylesheets` is cumulative for the session instead of read off the slowest snapshot, and new properties report the never-deferrable rule bucket (`$sdk_debug_replay_slowest_full_snapshot_css_rules_non_deferrable`) plus deferred idle-time stringification cost (`$sdk_debug_replay_deferred_stylesheet_ms` / `_slowest_slice_ms`). The duration gauges are also guarded against tab-suspension artifacts: a sample whose measurement window straddles a `visibilitychange`/`freeze`/`resume` boundary, or whose synchronous window exceeds a 60-second plausibility cap (a suspended renderer's wall clock spans the sleep, producing readings of minutes or more), is discarded rather than recorded, with the discard count exposed as `$sdk_debug_replay_discarded_duration_samples`.
```
marandaneto: i'd reduce this to make it digestible, its a wall of text

marandaneto (approved):

.changeset/stylesheet-budget-default-in-sdk.md
arnohillen: [arno's agent] Trimmed to a short bullet list in 07612a7.

PR #4445, opened 2026-08-06
fix(nuxt): support Nuxt 5 and Nitro 3 plugin boundaries

onmax:
## Problem

`@posthog/nuxt` loads its browser plugin during SSR, where importing `posthog-js` can evaluate `window` before hydration. Its Nitro plugin also depends on Nitro 2 imports, while Nuxt 5 uses Nitro 3.

## Changes

- register the browser plugin as client-only
- select a Nitro 2 adapter for Nuxt 3/4 and a Nitro 3 adapter for Nuxt 5
- keep the PostHog server behavior shared between both adapters
- return the PostHog flush promise from Nitro error hooks so request and context-free errors are delivered
- add regression coverage for plugin registration, request and context-free errors, shutdown behavior, and SSR

This preserves the current Nuxt 3/4 boundary instead of replacing its imports with Nitro 3 imports globally. Nuxt 5 and Nitro 3 are still pre-stable, so the consumer test pins the exact nightly and beta versions, packs the built module, installs it into a fresh project, builds it, and runs the generated server.

## Verification

- `pnpm --filter @posthog/nuxt lint`
- `pnpm --filter @posthog/nuxt build`
- `pnpm --filter @posthog/nuxt test:unit`, including the packed Nuxt 5/Nitro 3 context-free error fixture

### Related ecosystem work

- [nuxt/scripts#844](https://github.com/nuxt/scripts/pull/844) uses separate Nitro 2 and 3 runtime paths
- [nuxt-open-fetch#141](https://github.com/Norbiros/nuxt-open-fetch/pull/141) updates a Nuxt module for Nitro 3 runtime imports
- [nuxt-site-config#101](https://github.com/harlan-zw/nuxt-site-config/pull/101) updates another module for the same Nitro 3 boundary

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Codex with GPT-5.6 Sol researched the Nuxt and Nitro compatibility boundary, implemented the adapters, and ran an independent review.

I am now testing this in my own project using Nuxt 5 nightly + pnpm patch (adding pkg.pr.new in this repo would be nice ;) )


packages/nuxt/src/runtime/nitro-plugin-v2.ts
```diff
@@ -0,0 +1,13 @@
+import { defineNitroPlugin, useRuntimeConfig } from 'nitropack/runtime'
+import { setupPostHogNitroPlugin } from './nitro-plugin'
+
+export default defineNitroPlugin((nitroApp) => {
+  setupPostHogNitroPlugin({
+    useRuntimeConfig,
+    onError: (handler) =>
```
marandaneto: [P1] The required `Lint packages` check is currently failing: ESLint reports 11 errors across both adapters, `nitro-plugin.ts`, and `nitro-plugin.test.mjs` (arrow parentheses, trailing commas, member delimiter, and the duplicate regex alternative). Please apply the package lint fixes and rerun `pnpm --filter @posthog/nuxt lint`.

packages/nuxt/tests/nitro-plugin.test.mjs
```diff
@@ -0,0 +1,125 @@
...
+  { $process_person_profile: false, path: '/api/test', method: 'POST' },
+])
+
+await handlers.close()
+assert.deepEqual(calls[3], ['shutdown'])
+
+function loadAdapter(filename, defineName) {
+  const adapterSource = readFileSync(new URL(`../src/runtime/${filename}`, import.meta.url), 'utf8')
+  const executableAdapter = adapterSource
+    .replace(/^import .*$/gm, '')
```
marandaneto, line 99: [P2] Removing every import and injecting identity stubs means this test cannot validate the real Nitro 2/3 imports, callback contracts, package resolution, or generated module output—the compatibility boundary changed by this PR. Please add a packed consumer or build fixture against actual Nuxt 5/Nitro 3 dependencies.

marandaneto:
thanks @onmax left a few comments

onmax:
Thanks for the review - I addressed both points in [a34b8fe83](https://github.com/PostHog/posthog-js/commit/a34b8fe83).

marandaneto:
@PostHog/team-error-tracking wanna double check? looks good to me, i dont know much about nuxt tho

marandaneto (approved):

packages/nuxt/src/runtime/nitro-plugin-v3.ts
```diff
@@ -0,0 +1,14 @@
...
+import { setupPostHogNitroPlugin } from './nitro-plugin'
+
+export default definePlugin((nitroApp) => {
+  setupPostHogNitroPlugin({
+    useRuntimeConfig,
+    onError: handler =>
+      nitroApp.hooks.hook('error', (error, { event }) =>
+        handler(error, event ? { path: new URL(event.req.url).pathname, method: event.req.method } : undefined),
+      ),
+    onClose: handler => nitroApp.hooks.hook('close', handler),
```
cat-ph, line 20: I don't think I was able to get the`hooks.callHook("close")` to show up in the generated server here, I was trying to test all combos too just to make sure it works, the TL;DR was:

Image

I think this loses the queued events on shutdown, do you know if there's another lifecycle hook or a different way to do this for nitro v3? 🤔

cat-ph (commented):
thanks @onmax, looks great! just one question for the nuxt5/nitro3 shutdown from me, I'm not *too* familiar with nuxt but I noticed that wasn't working

packages/nuxt/src/runtime/nitro-plugin-v3.ts
onmax: I will take a look and setup a test scenario for this case. Thanks for the catch!

onmax: I think i fix it! Please let me know :)

cat-ph: looks better @onmax! but I think we still have an issue for errors without a request context.

Nitro already owns the request lifetime: In Nitro 3, captureError calls the error hook, keeps the promise returned by that hook, and when an event exists, passes it to event.req.waitUntil ([Nitro 3 source](https://github.com/nitrojs/nitro/blob/df5799e75654c50016863278fc5a036d57648af8/src/build/virtual/app.ts#L39-L61)). Hookable’s callHook includes promises returned by hook handlers ([Hookable source](https://github.com/unjs/hookable/blob/b77477c027039362ee0ec4f39b8998c4f1b21707/src/utils.ts#L74-L90)). Nitro 2 does the equivalent with callHookParallel and event.waitUntil ([Nitro 2 source](https://github.com/nitrojs/nitro/blob/62ac3065ef99da6a23716f205abd3e8a982ba9ad/src/runtime/app.ts#L46-L61)).


so I think we shouldn’t need to pass the raw request waitUntil into the PostHog handler. Returning the flush promise from the error hook is enough for Nitro to attach it to the request lifetime.
The current implementation only flushes inside request?.waitUntil ([current code](https://github.com/PostHog/posthog-js/blob/e76597a9d9ad52c56e1d889db185f3b96a22052d/packages/nuxt/src/runtime/nitro-plugin.ts#L33-L47)). However, Nitro also sends unhandledRejection and uncaughtException through captureError without an event ([Nitro 3 source](https://github.com/nitrojs/nitro/blob/df5799e75654c50016863278fc5a036d57648af8/src/runtime/internal/error/hooks.ts#L3-L10)), so those errors currently get queued but not flushed.


The close hook also doesn’t cover that gap in the exact Nitro 3 beta pinned by the test. Its Node entry creates the srvx server and installs the process error handlers, but never calls the Nitro close hook ([Nitro Node entry](https://github.com/nitrojs/nitro/blob/df5799e75654c50016863278fc5a036d57648af8/src/presets/node/runtime/node-server.ts#L19-L47)). srvx handles SIGINT/SIGTERM by closing its own server ([srvx shutdown handler](https://github.com/h3js/srvx/blob/19efb135b7f10d5e5a7bab22760280827488acdc/src/_plugins.ts#L17-L82)) and waits only for promises registered with its own waitUntil ([srvx server lifetime](https://github.com/h3js/srvx/blob/19efb135b7f10d5e5a7bab22760280827488acdc/src/adapters/node.ts#L189-L203)). By comparison, Nitro 2 explicitly calls nitroApp.hooks.callHook("close") during graceful shutdown ([Nitro 2 shutdown](https://github.com/nitrojs/nitro/blob/62ac3065ef99da6a23716f205abd3e8a982ba9ad/src/runtime/shutdown.ts#L16-L47)).


I reproduced this with the packed Nuxt 5/Nitro 3 fixture: a background unhandledRejection, followed by SIGTERM, produced zero PostHog batches. Returning client.flush() from the hook delivered the exception.

cat-ph: tl;dr I think the simplest is to do:

```js
client.captureException(error, uuidv7(), props)
return client.flush().catch(() => {})
```

and get rid of the request waitUntil. and maybe also extending the test with a context-free error case, but this should work in all cases, just double checked too!

onmax: Thanks @cat-ph,

you were right about the request-context gap. I applied the simpler lifecycle ownership you suggested in 346a62ac: the shared error handler now returns `client.flush()` and no longer receives raw `waitUntil`.

I also extended the packed Nuxt 5/Nitro 3 fixture with a context-free `unhandledRejection`

cat-ph: thanks @onmax, looks great! 💙 (& sorry for the delay)

cat-ph (approved):

PR #4446, opened 2026-08-06
fix: emit deployment environment under both semconv spellings

jonmcwest:
## Problem

The SDK emitted the deployment environment only as `deployment.environment.name`, the OpenTelemetry semantic-convention 1.27 key. Backends, saved views, and alerts still reading the older `deployment.environment` key stopped matching environment values, so environment filtering silently returned nothing for them.

## Changes

- Logs and metrics resource attributes now carry the environment under **both** `deployment.environment.name` and the pre-1.27 `deployment.environment`.
- Config resolution (browser logs, core metrics) accepts the environment from either spelling in `resourceAttributes`, preferring the current key when both are present.
- Doc comments and unit tests updated to cover both spellings.

The old key is kept for backwards compatibility and can be dropped in a future major.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored by the PostHog Slack app (Claude Opus 4.8) from a Slack thread. The environment attribute was the only faceted resource attribute the OTel semantic conventions have ever renamed, so the fix is scoped to it rather than a general aliasing layer. Emitting both keys (rather than switching resolution to read-time only) keeps already-ingested data queryable on either spelling. Verified with the core logs/metrics unit specs (102 passing) and the browser `logs-defaults` config resolution tests (13 passing).

### DIFF
--- a/.changeset/dual-emit-deployment-environment.md
+++ b/.changeset/dual-emit-deployment-environment.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+---
+
+Emit the deployment environment under both `deployment.environment.name` (semconv 1.27+) and the older `deployment.environment` for logs and metrics, and accept either spelling from `resourceAttributes`. Keeps environment filtering working for backends still reading the pre-1.27 key.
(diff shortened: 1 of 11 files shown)


jonmcwest:
Closing — not needed. The read side ([PostHog/posthog#78739](https://github.com/PostHog/posthog/pull/78739)) resolves the environment resource attribute across spellings (`deployment.environment.name`, `deployment.environment`, `env`), and the SDK already emits `deployment.environment`, so the facet rail / MCP / API surface it without the SDK needing to double-emit.

PR #4464, opened 2026-08-07
fix(mcp): install and type-check cleanly on a v2-only project

gesh:
## What

> Stacked on #4462

A project built on `@modelcontextprotocol/server` (v2) could not adopt `@posthog/mcp` cleanly. Two halves of the same problem, and the second only becomes reachable once the first is fixed, so they ship together.

**1. The v1 SDK was a required peer.** Installing this package therefore dragged the entire v1 SDK into a v2-only project — one it will never load — and dependency-tree tooling reported it as missing whenever it was absent. Both majors are now declared and both are optional, which is what the code has always assumed: shipped code imports no `@modelcontextprotocol/*` package (pinned by `sdk-import-boundary.test.ts`) and detects server shapes structurally.

```jsonc
"peerDependencies": {
  "@modelcontextprotocol/sdk": ">=1.26.0",     // v1
  "@modelcontextprotocol/server": ">=2.0.0",   // v2
  "posthog-node": "^5.0.0"
},
"peerDependenciesMeta": {
  "@modelcontextprotocol/sdk": { "optional": true },
  "@modelcontextprotocol/server": { "optional": true }
}
```

**2. The published types still imported the v1 SDK.** Once the peer is optional, a consumer legitimately has no v1 SDK on disk — but `dist/types.d.ts` and `dist/extensions/tools.d.ts` imported `CallToolResult` / `ListToolsResult` from `@modelcontextprotocol/sdk/types.js`. That is a `TS2307` for anyone type-checking without `skipLibCheck`: an install that runs fine and will not compile. Shipping only the first half would have traded a noisy install for a confusing compile error, so the MCP wire shapes are now declared structurally in `types.ts` as well.

The shapes we *read* are deliberately open-ended, so a value typed by either SDK major assigns to them. What we *return* stays precise — that is the direction that would break callers silently — so `getMoreToolsResult()` can still be handed straight back to the SDK.

v2 also joins `devDependencies` at `~2.0.0`, matching the existing `~1.29.0` on v1. pnpm auto-installs the optional peer into the workspace regardless, so this adds nothing to the tree; it declares *which* version this repo develops against. Patch-only on purpose for both majors: this package duck-types SDK internals (`_requestHandlers`, `_registeredTools`, `_serverInfo`) that carry no semver guarantee, so adopting a new minor should be a reviewed commit rather than install-time drift.

## What was measured

**Install**, into a project with `@modelcontextprotocol/server@2.0.0` + `posthog-node@5`, from a packed tarball carrying the old manifest and then the new one:

```
before   added 87 packages    including @modelcontextprotocol/sdk@1.30.0, pulled in as a required peer
after    added  1 package     @modelcontextprotocol/sdk → UNMET OPTIONAL DEPENDENCY, correctly absent
```

**Type-check**, same project, `skipLibCheck: false`, `moduleResolution: nodenext`:

```
before   dist/types.d.ts(1,37):            error TS2307: Cannot find module '@modelcontextprotocol/sdk/types.js'
         dist/extensions/tools.d.ts(1,59): error TS2307: Cannot find module '@modelcontextprotocol/sdk/types.js'
after    no errors from @posthog/mcp
```

**Tests.** 564 unit tests green, lint and build clean. `report-missing.test.ts` now asserts at compile time that `getMoreToolsResult()` is still assignable to the SDK's `CallToolResult`, so a structural type drifting away from the SDK's fails the build rather than a consumer's. `sdk-import-boundary.test.ts` gains the stronger rule this implies — no `@modelcontextprotocol` reference in shipped source *at all*, not merely no runtime one — and its scanner self-check moves to a fixture, since `src` no longer contains the type-only import it used to look for.

**Behaviour.** Types are erased, so no runtime change is expected; the testbed confirms none. All four `v1` matrix rows all-green, `v2` rows identical to the parent branch, late-handler probe 9/9, `nest v1` 15/16 · 15/16, `nest v2` 24/33 · 24/33 — every number unchanged from `#4462`.

```
                              calls errors schema session client protocol warnings header alive
 v1  stateful   high             ✓      ✓      ✓      ✓       ✓       ✓        ✓       ·     ·
 v1  stateful   low              ✓      ✓      ✓      ✓       ✓       ✓        ✓       ·     ·
 v1  stateless  high             ✓      ✓      ✓      ✓       ✓       ✓        ✓       ·     ·
 v1  stateless  low              ✓      ✓      ✓      ✓       ✓       ✓        ✓       ·     ·
 v2  high  2025  conv=off        ✓      ✓      ✓      ·       ✗       ✗        ✓       ✓     ✓
 v2  high  2025  conv=on         ✓      ✓      ✓      ✓       ✗       ✗        ✓       ✓     ✓
 v2  high  2026  conv=off        ✓      ✓      ✓      ·       ✓       ✗        ✓       ✓     ✓
 v2  high  2026  conv=on         ✓      ✓      ✓      ✓       ✓       ✗        ✓       ✓     ✓
 v2  low   2025  conv=off        ✓      ✓      ✓      ·       ✗       ✗        ✓       ✓     ✓
 v2  low   2025  conv=on         ✓      ✓      ✓      ✗       ✗       ✗        ✓       ✓     ✓
 v2  low   2026  conv=off        ✓      ✓      ✓      ·       ✓       ✗        ✓       ✓     ✓
 v2  low   2026  conv=on         ✓      ✓      ✓      ✗       ✓       ✗        ✓       ✓     ✓
```

`protocol` is red here because the fix for it is the next branch in the stack, not because anything regressed.

## Noticed, not fixed here

The same v2-only type-check surfaces `posthog-node/dist/extensions/express.d.ts` referencing `express`, which is optional there too — the identical bug in a different package. Left alone rather than smuggled into this stack.

### DIFF
--- a/packages/mcp/package.json
+++ b/packages/mcp/package.json
@@ -47,6 +47,7 @@
     "@babel/preset-env": "catalog:",
     "@babel/preset-typescript": "catalog:",
     "@modelcontextprotocol/sdk": "~1.29.0",
+    "@modelcontextprotocol/server": "~2.0.0",
     "@posthog-tooling/tsconfig-base": "workspace:*",
     "@rslib/core": "catalog:",
     "@types/jest": "catalog:",
@@ -57,7 +58,16 @@
   },
   "peerDependencies": {
     "@modelcontextprotocol/sdk": ">=1.26.0",
+    "@modelcontextprotocol/server": ">=2.0.0",
     "posthog-node": "^5.0.0"
+  },
+  "peerDependenciesMeta": {
+    "@modelcontextprotocol/sdk": {
+      "optional": true
+    },
+    "@modelcontextprotocol/server": {
+      "optional": true
+    }
   },
   "keywords": [
     "posthog",
--- a/packages/mcp/src/__tests__/report-missing.test.ts
+++ b/packages/mcp/src/__tests__/report-missing.test.ts
@@ -1,4 +1,5 @@
 import { type CallToolResult, CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
+import type { CallToolResult as V2CallToolResult } from '@modelcontextprotocol/server'
 import { z } from 'zod'
 import { instrument } from '../index'
 import { DEFAULT_CONTEXT_PARAMETER_DESCRIPTION } from '../extensions/constants'
@@ -313,8 +314,16 @@ describe('reportMissing (get_more_tools virtual tool)', () => {
  * the exported surface; internal descriptors are read by us, never handed on.
  */
 describe('results we hand back stay assignable to the SDK types', () => {
-  it('getMoreToolsResult() satisfies CallToolResult', () => {
+  it('getMoreToolsResult() satisfies v1 CallToolResult', () => {
     const result: CallToolResult = getMoreToolsResult()
+    expect(result.content).toHaveLength(1)
+  })
+
+  // Both majors, because either can drift away from our structural type — and v2
+  // is the likelier of the two, being the newer schema. Available only because
+  // this package now carries v2 as a devDependency.
+  it('getMoreToolsResult() satisfies v2 CallToolResult', () => {
+    const result: V2CallToolResult = getMoreToolsResult()
     expect(result.content).toHaveLength(1)
   })
 })
--- a/packages/mcp/src/__tests__/report-missing.test.ts
+++ b/packages/mcp/src/__tests__/report-missing.test.ts
@@ -1,9 +1,10 @@
-import { CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
+import { type CallToolResult, CallToolResultSchema, ListToolsResultSchema } from '@modelcontextprotocol/sdk/types.js'
 import { z } from 'zod'
 import { instrument } from '../index'
 import { DEFAULT_CONTEXT_PARAMETER_DESCRIPTION } from '../extensions/constants'
 import { MCPAnalyticsEventType } from '../extensions/event-types'
 import { getServerTrackingData } from '../extensions/internal'
+import { getMoreToolsResult } from '../index'
 import { EventCapture, fakePostHog } from './test-utils'
 import { resetTodos, setupTestServerAndClient } from './test-utils/client-server-factory'
 
@@ -299,5 +300,21 @@ describe('reportMissing (get_more_tools virtual tool)', () => {
 
       await capture.stop()
     })
+  })
+})
+
+/**
+ * Shipped code declares the MCP wire shapes structurally instead of importing
+ * them, because both SDK majors are optional peers and a type import of one
+ * would break `tsc` for a consumer that installed the other. The risk that
+ * creates is the *other* direction: a host passing what we return straight back
+ * to the SDK. This is a compile-time assertion — if the type we return drifts
+ * away from the SDK's, this file stops compiling, which is the point. It covers
+ * the exported surface; internal descriptors are read by us, never handed on.
+ */
+describe('results we hand back stay assignable to the SDK types', () => {
+  it('getMoreToolsResult() satisfies CallToolResult', () => {
+    const result: CallToolResult = getMoreToolsResult()
+    expect(result.content).toHaveLength(1)
   })
 })
--- a/.changeset/mcp-v2-optional-peers.md
+++ b/.changeset/mcp-v2-optional-peers.md
@@ -2,8 +2,10 @@
 '@posthog/mcp': patch
 ---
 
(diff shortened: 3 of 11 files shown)


packages/mcp/package.json
```diff
@@ -57,8 +57,17 @@
   },
   "peerDependencies": {
     "@modelcontextprotocol/sdk": ">=1.26.0",
+    "@modelcontextprotocol/server": ">=2.0.0",
     "posthog-node": "^5.0.0"
   },
+  "peerDependenciesMeta": {
+    "@modelcontextprotocol/sdk": {
+      "optional": true
+    },
```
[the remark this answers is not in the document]
gesh: Valid, and a real break introduced by this PR rather than a theoretical one — thanks. Reproduced it first against a project with only `@modelcontextprotocol/server@2.0.0` + `posthog-node@5` installed, `skipLibCheck: false`, `moduleResolution: nodenext`:

```
dist/types.d.ts(1,37):            error TS2307: Cannot find module '@modelcontextprotocol/sdk/types.js'
dist/extensions/tools.d.ts(1,59): error TS2307: Cannot find module '@modelcontextprotocol/sdk/types.js'
```

Fixed in 499031e by declaring the MCP wire shapes structurally in `types.ts` instead of importing `CallToolResult` / `ListToolsResult`. Same check after the change: no errors from `@posthog/mcp`.

Two details worth a reviewer's eye, because the naive version of this fix is subtly wrong:

- The shapes we **read** are open-ended, so a value typed by either SDK major assigns to them. What we **return** is typed precisely — that is the direction that breaks callers silently, since a loose return type stops `getMoreToolsResult()` being assignable to the SDK's `CallToolResult` and breaks v1 hosts returning it straight from a tool callback. `report-missing.test.ts` now asserts that assignability at compile time, so drift fails our build rather than someone else's.
- `sdk-import-boundary.test.ts` already pinned "no runtime reference to `@modelcontextprotocol/*`". It now pins the stronger rule this implies: no reference **at all**, not even type-only, since a type-only import survives into the published `.d.ts`.

Out of scope but surfaced by the same check: `posthog-node/dist/extensions/express.d.ts` references `express`, an optional peer there too — identical bug, different package, left alone rather than pulled into this stack.

packages/mcp/src/__tests__/report-missing.test.ts
```diff
@@ -301,3 +302,19 @@ describe('reportMissing (get_more_tools virtual tool)', () => {
...
+ * Shipped code declares the MCP wire shapes structurally instead of importing
+ * them, because both SDK majors are optional peers and a type import of one
+ * would break `tsc` for a consumer that installed the other. The risk that
+ * creates is the *other* direction: a host passing what we return straight back
+ * to the SDK. This is a compile-time assertion — if the type we return drifts
+ * away from the SDK's, this file stops compiling, which is the point. It covers
+ * the exported surface; internal descriptors are read by us, never handed on.
+ */
+describe('results we hand back stay assignable to the SDK types', () => {
+  it('getMoreToolsResult() satisfies CallToolResult', () => {
```
lucasheriques: Non-blocking: this pins assignability to v1's `CallToolResult` only. Now that `@modelcontextprotocol/server` is a devDependency, the sibling assertion against v2's tool-result type (whatever `registerTool` expects the callback to return) is available for free - and v2 is the direction likelier to drift, being the newer schema. Would make the compile-time guarantee cover both majors the package claims to serve.

lucasheriques (approved):
Optional peers matching what the code already assumed (no runtime import of either SDK), and the structural types are loose on what we read / precise on what we return, which is the right asymmetry. The install and type-check before/after measurements make this easy to trust. One non-blocking suggestion inline.

Approving to unblock.

packages/mcp/src/__tests__/report-missing.test.ts
gesh: Good call — done in 391015c4. `getMoreToolsResult()` is now asserted assignable to both majors' `CallToolResult`, using the v2 devDependency this PR adds.

One thing I found while verifying it, worth knowing: **neither assertion is enforced by CI.** I loosened the return type to `content: unknown[]` and all 610 tests still passed, and the build stayed green — jest compiles through babel and strips types unchecked, and `rslib build` typechecks `src` for declarations but excludes `__tests__`. Running `tsc` over the file directly does catch both:

```
report-missing.test.ts(318,11): error TS2322: Type 'CompatibleTextToolResult' is not assignable …  ← v1
report-missing.test.ts(326,11): error TS2322: …                                                    ← v2
```

So the guarantee is real but only under `tsc`; my PR description overstated it as failing the build, and I have corrected that.

Not adding a `typecheck` script here — it would surface any pre-existing type error jest and rslib have been ignoring, which could turn this PR red for unrelated reasons. Worth doing on its own evidence; happy to open it separately.

PR #4467, opened 2026-08-07
fix(mcp): keep the conversation prompt-back out of the captured error

gesh:
## What

> Stacked on #4466

With `enableConversationId` on, a `[SERVER]: Reuse conversation_id=…` block is appended to a tool result so the agent echoes the handle back on later calls. The captured error was read from that **already-appended** result, so a failed call reported:

```
$mcp_error_message = "intentional failure [SERVER]: Reuse conversation_id=019fdc0c-631c-7656-a0d7-3c268527d5ee
                      on every subsequent tool call in this conversation. …"
```

A fresh uuid inside the message on every failure, so one recurring failure becomes a new error group each time it occurs — the exact signal error grouping exists to give you, destroyed by the feature meant to correlate it.

The error is now read from the result **as the tool produced it**, before the handle is written in. What the caller receives does not change: the agent still gets the prompt-back on a failed call, which is when it matters most — otherwise the retry opens a new conversation and the failure and its fix land in different sessions. Both halves are asserted, because the tempting wrong fix is to withhold the handle.

v2-specific in practice: on v1 the throw is stashed by the tool-callback wrapper and captured from there, so the appended result is never read. v2 flattens the throw into an `isError` result before our wrapper sees it, which makes that result the only description of the failure available.

## The harness could not see this, so I fixed the harness

The matrix already asserts `$mcp_error_message === 'intentional failure'` — and it was **green before this change**. Its fixture drives four tool calls, fails on the **last** one, and echoes the handle it received from the first. So on the failing call the handle is already minted, no prompt-back is appended to that result, and there is nothing to pollute. The column was green because the fixture never reached the code, which in a grid is indistinguishable from green because the code is right.

`dual-era/probe-first-call-error.mjs` covers the ordering it misses: the **first** call of a conversation failing — a bad argument, an expired token, a cold dependency. Run against the parent branch's build it reproduces the bug verbatim, and against this one:

```
parent branch    ✗ $mcp_error_message is clean   "intentional failure [SERVER]: Reuse conversation_id=019f…"
this branch      ✓ error captured
                 ✓ $mcp_error_message is clean   "intentional failure"
                 ✓ handle still delivered to the caller
```

It is wired into `run-all.mjs` and runs on every pass from now on.

## Matrix

No cell moved, and the grid is not where the evidence lives — see above.

```
                            calls  errors  schema  session  client  protocol  warnings  header  alive
──────────────────────────────────────────────────────────────────────────────────────────────────────
 v1  stateful   high          ✓      ✓       ✓        ✓       ✓        ✓         ✓        ·       ·
 v1  stateful   low           ✓      ✓       ✓        ✓       ✓        ✓         ✓        ·       ·
 v1  stateless  high          ✓      ✓       ✓        ✓       ✓        ✓         ✓        ·       ·
 v1  stateless  low           ✓      ✓       ✓        ✓       ✓        ✓         ✓        ·       ·
──────────────────────────────────────────────────────────────────────────────────────────────────────
 v2  high  2025  conv=off     ✓      ✓       ✓        ·       ✗        ✓         ✓        ✓       ✓
 v2  high  2025  conv=on      ✓      ✓       ✓        ✓       ✗        ✓         ✓        ✓       ✓
 v2  high  2026  conv=off     ✓      ✓       ✓        ·       ✓        ✓         ✓        ✓       ✓
 v2  high  2026  conv=on      ✓      ✓       ✓        ✓       ✓        ✓         ✓        ✓       ✓
 v2  low   2025  conv=off     ✓      ✓       ✓        ·       ✗        ✓         ✓        ✓       ✓
 v2  low   2025  conv=on      ✓      ✓       ✓        ✗       ✗        ✓         ✓        ✓       ✓
 v2  low   2026  conv=off     ✓      ✓       ✓        ·       ✓        ✓         ✓        ✓       ✓
 v2  low   2026  conv=on      ✓      ✓       ✓        ✗       ✓        ✓         ✓        ✓       ✓

 late-handler probe                                        9/9  (unchanged)
 first-call-error probe                                    3/3  (new — 2/3 on the parent branch)
 nest v1   instrument(server) / instrument(server.server)   15/16 · 15/16  (unchanged)
 nest v2   instrument(server) / instrument(server.server)   25/33 · 25/33  (unchanged)
```

`error message is clean` stays red on both Nest rows for an unrelated reason this PR cannot touch: NestJS's `RpcExceptionsHandler` replaces the thrown error with `"Internal server error"` before it reaches the MCP layer, so every distinct failure on a mcp-nest server groups together regardless. Worth a docs note to those users; not our bug.

587 unit tests green, lint and build clean. Revert-checked: backing the fix out fails exactly one unit test, plus the new probe.

### DIFF
--- a/packages/mcp/docs/adr/0009-mint-session-header-on-legacy-era-only.md
+++ b/packages/mcp/docs/adr/0009-mint-session-header-on-legacy-era-only.md
@@ -15,8 +15,9 @@ Dropping the mechanism entirely was considered. It was rejected because it is th
 
 Minting is gated on the revision **the request declares**, resolved through the same per-request chain as identity (ADR-0008): the version an `initialize` body asks for, else envelope → `_meta` → `MCP-Protocol-Version` header.
 
-- Any revision **at or after** `2026-07-28` is treated as sessionless. Revisions sort as ISO dates, so a future one is modern by default — sessions were removed, not re-added.
-- An **unknown** version counts as legacy. A client that declares nothing is on a transport that has always carried a session header, and taking it away would be the regression this guard exists to prevent.
+- Any revision **at or after** `2026-07-28` is treated as sessionless. Revisions sort as ISO dates, so a future one is modern by default — sessions were removed, not re-added. The spec's rolling `draft` counts as modern for the same reason.
+- Only **date-shaped** tokens are compared as dates. The comparison is lexicographic, so without that guard any junk string starting above `'2'` would sort as modern and lose the header — the opposite of the rule below.
+- An **unknown** version counts as legacy, including unknown because it was malformed. A client that declares nothing, or declares nonsense, is one we should assume still wants the header it has always had; taking it away is the regression this guard exists to prevent.
 - The gate lives in the session layer, not in the handler, so the policy is stated once.
 
 ## Consequences
--- a/packages/mcp/src/__tests__/session-id.test.ts
+++ b/packages/mcp/src/__tests__/session-id.test.ts
@@ -381,6 +381,22 @@ describe('isModernEraRequest', () => {
     expect(call({ method: 'tools/call', params: { name: 'echo' } }, viaEnvelope)).toBe(true)
   })
 
+  it.each([
+    ['junk that sorts above the revision', 'garbage'],
+    ['a truncated token', '2026'],
+    ['an empty-ish token', '   '],
+  ])('treats %s as legacy rather than comparing it as a date', (_label, protocolVersion) => {
+    // Lexicographically 'garbage' > '2026-07-28', so an unguarded comparison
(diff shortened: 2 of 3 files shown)


lucasheriques (approved):
Reading the error off the result as the tool produced it, while still delivering the prompt-back to the caller, is the right split - and the writeup on why the matrix was green for the wrong reason (fixture fails on the last call, so the handle predates the failure) plus the new first-call-error probe is exactly the kind of harness honesty that makes these grids trustworthy.

One CI note: the red `semgrep-package-managers` check was *cancelled* (concurrency), not failed - and this PR touches no package-manager files. Re-run it to green before merge.

Approving to unblock.

PR #4473, opened 2026-08-07
fix(ai): declare gemini cache reporting as inclusive on generations

fivestarspicy:
## Problem

Cache-heavy Gemini generations can be priced wrong in AI observability, because the SDK never tells ingestion how Gemini counts cache tokens.

Gemini counts `cachedContentTokenCount` inside `promptTokenCount`, so cache reads are a subset of input, not a separate pool. The SDK reports both numbers but declares nothing, so ingestion infers the accounting model from the token counts.

That inference is unreliable for Gemini. Under explicit context caching the two counts come from separate measurements: the cache object is counted when it is created, and the prompt is counted per request. They describe the same tokens but can disagree by a few percent, which puts the cache pool just above the input total and makes the counts look like two separate pools. Ingestion then bills the full input at the prompt rate and the cache again at the cache rate, so a nearly fully cached prompt bills several times over.

This is the JS counterpart of PostHog/posthog-python#860.

## Changes

- `TokenUsage` gains `cacheReportingExclusive?: boolean`, left undefined when a provider's accounting model is not known.
- The Gemini wrapper sets it to false on generations that report cache reads, on both the streaming and non-streaming paths.
- `captureAiGeneration` maps it onto `$ai_cache_reporting_exclusive`.

The property mapping checks `!== undefined` rather than truthiness. Every neighbouring field in `additionalTokenValues` uses a truthiness guard, which is correct for counts but would silently drop `false`, the only value this field ever carries.

The flag is set only when cache reads are present, so generations without caching do not carry an extra property.

### Libraries affected

- [x] @posthog/ai

## Checklist

- [x] Tests for new code

- Ran `npx jest` in `packages/ai`. 764 passed. Two failures in `tests/openai-agents-resolution.test.ts` are unrelated and fail the same way on a clean checkout.
- Added two cases to `tests/gemini.test.ts`. The first asserts `$ai_cache_reporting_exclusive` is `false` when cached tokens are present, which fails if the flag stops being set or if the property mapping reverts to a truthiness guard. The second asserts the property is absent when nothing was cached, which guards against sending it on every generation.
- Ran `npx tsc --noEmit`. The only errors are missing `posthog-node` declarations, caused by declaration-file generation failing in this sandbox. No errors in the changed files.
- Did not run a live Gemini call.

## 🤖 Agent context

I (the PostHog Slack app, running Claude) traced this from a Slack thread on Gemini cost reconciliation, and a PostHog engineer directed the fix. I did not set an assignee because I could not verify their GitHub handle in session.

The Python change landed first and needed the flag threaded through three separate tagging sites. The JS side has one central property mapper, so this diff is smaller, and the only real trap was the truthiness guard described above.

**Public artifact:** no customer or session material reached this PR. The token values in the new tests are invented.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C087XQ7K9K7/p1786121664214699?thread_ts=1786121664.214699&cid=C087XQ7K9K7)*

### DIFF
--- a/.changeset/gemini-cache-reporting-exclusive.md
+++ b/.changeset/gemini-cache-reporting-exclusive.md
@@ -1,0 +1,5 @@
+---
+'@posthog/ai': patch
+---
+
+fix(gemini): declare Gemini's cache accounting model on generations with cache reads, so ingestion prices cached tokens from `$ai_cache_reporting_exclusive` instead of inferring it from the token counts
--- a/packages/ai/src/captureAiGeneration.ts
+++ b/packages/ai/src/captureAiGeneration.ts
@@ -165,6 +165,11 @@ export const captureAiGeneration = async (client: PostHog, options: CaptureAiGen
       ...(usage.reasoningTokens ? { $ai_reasoning_tokens: usage.reasoningTokens } : {}),
       ...(usage.cacheReadInputTokens ? { $ai_cache_read_input_tokens: usage.cacheReadInputTokens } : {}),
       ...(usage.cacheCreationInputTokens ? { $ai_cache_creation_input_tokens: usage.cacheCreationInputTokens } : {}),
+      // Checked against undefined rather than truthiness, because false is the meaningful
+      // value here and a truthiness guard would drop it.
+      ...(usage.cacheReportingExclusive !== undefined
+        ? { $ai_cache_reporting_exclusive: usage.cacheReportingExclusive }
+        : {}),
       ...(usage.webSearchCount ? { $ai_web_search_count: usage.webSearchCount } : {}),
       ...(usage.rawUsage ? { $ai_usage: usage.rawUsage } : {}),
     }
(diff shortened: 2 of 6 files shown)


fivestarspicy:
Added in [cf73d3e](https://github.com/PostHog/posthog-js/commit/cf73d3e663cb928007786c6e9a12ac1d7d06e095).

The derived flag is now suppressed whenever the caller overrides any token count through `posthogProperties`. Reused the existing passthrough detection, extracted as `hasTokenOverrides`, which `getTokensSource` now shares so the two can't drift apart.

A caller who knows their own accounting model can still pass `$ai_cache_reporting_exclusive` explicitly, and that value wins via the later `options.properties` spread.

Two tests added: the flag is absent when token counts are overridden, and an explicitly-passed flag survives.

Worth noting the same hazard exists in [posthog-python#860](https://github.com/PostHog/posthog-python/pull/860), where `posthog_properties` merges over the SDK tags the same way.

Radu-Raicea (approved):

PR #4477, opened 2026-08-09
fix(replay): don't open a recording that holds only idle markers

TueHaulund:
## Problem

Small follow-up to the idle-tab recording incident fixed in #4407.

An idle tab rotates its session, and the lifecycle markers for that rotation (`sessionIdle`, `$session_id_change`) land in the new session's empty buffer. Flushing them opens a recording with no Meta and no FullSnapshot, so it plays back as nothing while still counting as a recording.

#4407 stopped these rotations shipping a full payload. The markers were left shipping on their own, which this tidies up.

## Changes

- `_flushBuffer` will not open a recording when every event in the buffer is a custom-event marker and the session has no FullSnapshot yet.
- The markers are kept rather than dropped, so they ship with the content that follows. Past the buffer size cap they are cleared instead.
- `$session_starting` and `$session_ending` are exempt. They exist only in the replay stream and session linking needs them.
- The gate sits in `_flushBuffer`, so it covers the idle, scheduled and unload flushes as one rule.
- Seven existing tests emitted a lone synthetic custom event. Real rrweb emits Meta and FullSnapshot first, so each now stages a full snapshot and keeps its original assertion.
- This guards the flush, not the rotation that leaves the buffer empty.
- No visual change.

## Testing

Three cases in the `recording volume invariants` block: markers alone open no recording, linking markers still do, and held markers ship with the content that follows. Two of the three fail without this change. Full replay suite passes locally, 764 tests.

Not run: Playwright, and the repo typecheck, which fails on pre-existing errors in surveys and rrweb unrelated to this diff.

### Libraries affected

- [x] posthog-js (web)

## 🤖 Agent context

I (actually Claude) wrote this while looking into replay volume after the idle-tab fix. The diagnosis came from reading production recording payloads. Skills invoked: `writing-pr-descriptions`, `writing-tests`, `writing-code-comments`.

### DIFF
--- a/.changeset/drop-marker-only-recordings.md
+++ b/.changeset/drop-marker-only-recordings.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): don't open a recording that holds only idle lifecycle markers
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -366,6 +366,7 @@ describe('Lazy SessionRecording', () => {
                         },
                     })
                 )
+                sessionRecording.onRRwebEmit(createFullSnapshot({ timestamp: Date.now() }))
                 const snapshot = createCustomSnapshot({ timestamp: Date.now() })
                 sessionRecording.onRRwebEmit(snapshot)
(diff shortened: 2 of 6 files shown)


marandaneto (approved):
Unrelated docs reference updates? Otherwise lgtm

fasyy612 (approved):
nice 👌

PR #4494, opened 2026-08-11
fix(types): accept Segment Analytics SDK types

### LINKED ISSUE #1683: Type mismatch when using @segment/analytics-next
When using the latest `@segment/analytics-next`, passing `window.analytics` to segment in `posthog-js` triggers type errors. Our segment config currently expects the older Segment interface `SegmentAnalytics`. We should update `posthog-js` to accept the latest type `AnalyticsSnippet`.

Internal ticket: https://posthoghelp.zendesk.com/agent/tickets/23120

marandaneto:
## Problem

Passing an `AnalyticsSnippet` from recent `@segment/analytics-next` versions to `posthog.init({ segment })` fails TypeScript validation. Segment user IDs can be `null`, and its `register` method returns a context instead of `void`.

Fixes #1683.

## Changes

- Accept nullable Segment user IDs in the config-facing integration type without changing the exported `SegmentUser` contract.
- Accept the `register` signatures used by current and older Segment SDK versions while preserving the legacy `Promise<void>` result contract.
- Use the broader config-facing user type inside the browser integration.
- Add compile-time coverage for constrained Segment plugins, synchronous snippets, asynchronous browser instances, and previous PostHog Segment contracts.
- Add patch changesets for `posthog-js` and `@posthog/types`.

The compatibility check also passed against published `@segment/analytics-next` versions 1.8.0, 1.20.0, and 1.84.1.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. The public type remains structural, so Segment does not become a dependency of `@posthog/types`. The exported `SegmentUser` and legacy `Promise<void>` result remain unchanged while the config-facing shape accepts current Segment types. Compatibility was checked against the oldest published Analytics Next release, an intermediate release, and the current release. Pi autoreview reported no actionable findings.

### DIFF
--- a/packages/types/src/__tests__/segment.spec.ts
+++ b/packages/types/src/__tests__/segment.spec.ts
@@ -53,8 +53,8 @@ describe('SegmentAnalytics', () => {
             id: () => undefined,
         }
         const legacyUserId: string | undefined = legacyUser.id()
-        const getLegacyRegisterResult = (analytics: SegmentAnalytics): Promise<void> =>
-            analytics.register({} as SegmentPlugin)
+        const getLegacyRegisterResult = (analytics: SegmentAnalytics, integration: SegmentPlugin): Promise<void> =>
+            analytics.register(integration)
 
         expect(legacyUserId).toBeUndefined()
         expect(getLegacyRegisterResult).toEqual(expect.any(Function))
--- a/packages/types/src/__tests__/segment.spec.ts
+++ b/packages/types/src/__tests__/segment.spec.ts
@@ -1,0 +1,58 @@
+import * as ts from 'typescript'
+
+import type { SegmentAnalytics, SegmentPlugin } from '../segment'
+
+type SegmentId = string | null | undefined
+
+type AnalyticsNextUser = {
+    anonymousId(id?: SegmentId): SegmentId
+    id(id?: SegmentId): SegmentId
+}
+
+// Relevant public shape shared by @segment/analytics-next 1.8.0 through 1.84.1.
+type AnalyticsNextSnippet = {
+    user(): AnalyticsNextUser
+    register(...plugins: unknown[]): Promise<unknown>
+}
+
+type AnalyticsNextBrowser = {
+    user(): Promise<AnalyticsNextUser>
+    register(...plugins: unknown[]): Promise<unknown>
+}
+
+// Preserve the narrower contract accepted by previous posthog-js versions.
+type LegacySegmentAnalytics = {
+    user(): {
+        anonymousId(): string | undefined
+        id(): string | undefined
+    }
+    register(integration: SegmentPlugin): Promise<void>
+}
+
+const acceptsSegmentAnalytics = (analytics: SegmentAnalytics): void => {
+    void analytics
+}
+
+describe('SegmentAnalytics', () => {
+    it('accepts current and legacy Segment SDK shapes', () => {
+        acceptsSegmentAnalytics({} as AnalyticsNextSnippet)
+        acceptsSegmentAnalytics({} as AnalyticsNextBrowser)
+        acceptsSegmentAnalytics({} as LegacySegmentAnalytics)
+
+        const program = ts.createProgram([__filename], {
+            module: ts.ModuleKind.CommonJS,
+            moduleResolution: ts.ModuleResolutionKind.Node10,
+            noEmit: true,
+            skipLibCheck: true,
+            strict: true,
+            target: ts.ScriptTarget.ES2022,
+            types: ['jest', 'node'],
+        })
+        const errors = ts
+            .getPreEmitDiagnostics(program)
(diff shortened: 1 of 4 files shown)


packages/types/src/__tests__/segment.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+
+const acceptsSegmentAnalytics = (analytics: SegmentAnalytics): void => {
+    void analytics
+}
+
+describe('SegmentAnalytics', () => {
+    it('accepts current and legacy Segment SDK shapes', () => {
+        acceptsSegmentAnalytics({} as AnalyticsNextSnippet)
+        acceptsSegmentAnalytics({} as AnalyticsNextBrowser)
+        acceptsSegmentAnalytics({} as LegacySegmentAnalytics)
```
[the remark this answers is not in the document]
marandaneto: Fixed in a3e3011ee. The legacy result assertion now accepts a typed SegmentPlugin parameter instead of casting an empty object.

packages/types/src/segment.ts
```diff
@@ -10,12 +10,21 @@ export type SegmentUser = {
     id(): string | undefined
 }
 
+type SegmentAnalyticsUser = {
+    anonymousId(): string | null | undefined
+    id(): string | null | undefined
+}
+
```
dustinbyrne, line 17: given we're just widening to include `null`, i don't believe we _need_ the additional `SegmentAnalyticsUser` type - `SegmentUser` could include null on its properties

packages/types/src/segment.ts
```diff
@@ -10,12 +10,21 @@ export type SegmentUser = {
...
  */
 export type SegmentAnalytics = {
-    user: () => SegmentUser | Promise<SegmentUser>
-    register: (integration: SegmentPlugin) => Promise<void>
+    user: () => SegmentAnalyticsUser | Promise<SegmentAnalyticsUser>
+    /**
+     * Segment SDK versions use different plugin and resolved value types. `any` keeps current SDK types assignable
+     * without breaking legacy consumers that expect `Promise<void>`.
+     */
+    register: (integration: any) => Promise<any>
```
dustinbyrne, line 27: nit: consider `Promise<unknown>` instead of any

dustinbyrne (approved):

packages/types/src/segment.ts
marandaneto: I kept these types separate because SegmentUser is exported from @posthog/types and re-exported from posthog-js. Adding null to its return values breaks existing assignments to string | undefined. SegmentAnalyticsUser accepts Segment’s nullable IDs at the config boundary while preserving that public contract, and the regression test checks both shapes.

packages/types/src/segment.ts
marandaneto: I considered Promise<unknown>, but SegmentAnalytics.register previously returned Promise<void>. Changing it to Promise<unknown> breaks existing consumers that assign the result to Promise<void>. Promise<any> accepts Segment’s Promise<Context> while preserving that legacy assignment. The comment and regression test document this intentional compatibility tradeoff.

marandaneto: Promise Promise Promise Promise :D

marandaneto: > The difference is what TypeScript permits after calling or awaiting the promise.
> 
>  ### Promise (unknown)
> 
>  Safe but restrictive:
> 
>  ```ts
>    const result = await analytics.register(plugin)
>    // result is unknown
>    result.foo // TypeScript error
>  ```
> 
>  Both current and older Segment implementations can satisfy it:
> 
>  ```ts
>    Promise<Context> -> Promise<unknown> // valid
>    Promise<void>    -> Promise<unknown> // valid
>  ```
> 
>  However, it changes the public return contract:
> 
>  ```ts
>    const result: Promise<void> = analytics.register(plugin)
>    // Error: Promise<unknown> is not assignable to Promise<void>
>  ```
> 
>  ### Promise (any)
> 
>  Permissive:
> 
>  ```ts
>    const result = await analytics.register(plugin)
>    // result is any
>    result.foo // allowed without checking
>  ```
> 
>  It accepts current Segment’s Promise<Context> and preserves the previous consumer assignment:
> 
>  ```ts
>    const result: Promise<void> = analytics.register(plugin) // still valid
>  ```
> 
>  So unknown provides better type safety, but any preserves both the new Segment compatibility and the old
>  PostHog public contract. That backward compatibility is why this PR uses Promise<any>.

PR #4496, opened 2026-08-11
feat(browser): keep cross-subdomain cookie state in sync

### LINKED ISSUE #1399: Cross-domain behavior of localstorage+cookie
Hello,

I am somewhat confused about how `localstorage+cookie` works across different subdomains. The current implementation merges the values stored in the cookie with those stored in local storage, preferring the values in local storage:

https://github.com/PostHog/posthog-js/blob/34204f4aff107162bdb259d2a7c855d67bf0260c/src/storage.ts#L262

But this creates a very undesirable behavior when the user navigates between different subdomains. Take the following sequence of events:

1. User visits `www.example.com`, is assigned a random distinct ID, which is stored in a cookie (for `.example.com`) and in local storage as well (for `www.example.com`)
2. User then navigates to `support.example.com`, where the distinct ID is loaded from the cookie (for `.example.com`) which is also saved to local storage there (for `support.example.com`).
3. The user logs in on `support.example.com`, identify is called, and now has an authenticated distinct ID, which overwrites the value in the cookie (for `.example.com`) and in local storage (for `support.example.com`)
4. The user navigates back to `www.example.com`, where the authenticated distinct ID is loaded from the cookie (for `.example.com`), but is overwritten by the anonymous distinct ID in local storage (for `www.example.com`). Afterwards, the anonymous distinct ID also overwrites the value in the cookie (for `.example.com`).
5. (If the user then visits a third subdomain, the authenticated distinct ID is already lost.)

In most use cases, I think that if the user is identified, that information should spread across subdomains. Of course one could keep using the cookie-only storage, which solves the issue, but makes all requests to the server larger and can cause problems if the cookie grows too large (which is why I gather `localstorage+cookie` was made the default persistence option).

One way to fix this would be to prefer the values in the cookie, effectively reversing the merge order. Additional logic may be added to avoid merging values between different authenticated users. A conservative approach would be:

- If the distinct ID in the cookie and local storage is the same, merge without further checks
- If only one of them is authenticated, prefer that value, and discard the contents of the other
- Otherwise (if they are for different, authenticated users), do not merge

A more aggressive approach would be that always prefers the value in the cookie:

- If the distinct ID in the cookie and local storage is the same, merge without further checks
- Otherwise, discard value in local storage and use the one from the cookie

Is the current behavior by design? Would it be possible to change it / make it configurable, or should we simply roll our own PersistentStore to work around it?

marandaneto:
## Problem

Closes #1399.

Customers who use `localStorage+cookie` across sibling subdomains can keep stale identity and session data in an already-open tab. A shared cookie may contain a newer `identify()` or `reset()` result, but the tab can continue capturing with its per-origin localStorage state and can write that stale state back to the cookie.

The existing `__preview_cookie_wins_on_conflict` option only changes the merge order during persistence load. It does not keep already-open tabs in sync.

## Changes

- Add `cookieWinsOnConflict` as the supported configuration option and keep `__preview_cookie_wins_on_conflict` as a deprecated alias. The supported option wins when both are set.
- Enable `cookieWinsOnConflict` for the `2026-08-29` defaults snapshot. It remains disabled for earlier defaults.
- Check the shared cookie before event properties are calculated and before persistence writes. This lets an open sibling subdomain adopt newer identity and session state without replacing local-only properties.
- Flush `identify()` and `reset()` persistence immediately when the option is enabled so another subdomain does not observe the old cookie during the persistence debounce window.
- Reload feature flags after adopting a different distinct ID and notify session listeners only when the session ID changes.
- Preserve the shared cookie if `set_config` enables cookie precedence while also rebuilding persistence storage.
- Add unit coverage and a Playwright test that runs the flow across `a.example.test` and `b.example.test` in Chromium, Firefox, and WebKit.

The fast `array.js` comparison against `origin/main` measured an increase of 1.62 KiB minified, 0.35 KiB gzip, and 0.24 KiB Brotli.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi coding agent implemented the change in a dedicated worktree. The bundled autoreview found that a combined precedence and storage configuration change could clear the shared cookie. The implementation now adopts the cookie before storage migration, with regression coverage for that case. A second autoreview completed without actionable findings.


marandaneto:
## Issue #1399 coverage

| Reported scenario | Addressed? |
|---|---|
| Stale subdomain localStorage overrides the newer shared cookie after reload | Yes — cookie precedence is reversed when `cookieWinsOnConflict` is enabled. |
| Already-open tab continues capturing with stale in-memory identity | Yes — cookie state is reconciled before event properties are calculated. |
| Stale tab writes its old identity back into the cookie | Yes for the reported sequential flow — reconciliation occurs before persistence writes. |
| Sibling `identify()` is not adopted | Yes. |
| Sibling `reset()` leaves the other tab using the previous user | Yes — removals and the new anonymous identity propagate. |
| `$user_state`/person-processing state remains stale | Yes — built-in cookie-backed properties are reconciled. |
| Session ID remains stale | Yes — session adoption is included. |
| Feature flags remain evaluated for the old identity | Yes — identity adoption triggers a flag reload. |
| Existing preview option needs a supported replacement | Yes — `cookieWinsOnConflict` replaces the deprecated alias. |

The cross-subdomain Playwright test directly models the two-tab `a.example.test` / `b.example.test` flow and covers both `identify()` and `reset()`.

packages/browser/src/posthog-persistence.ts
```diff
@@ -421,17 +548,27 @@ export class PostHogPersistence {
...
+            this.syncCookieProperties()
+        }
+
         if (this._splitStorage) {
             this._writeNowSplit()
             return
         }
 
         this._writeEntry(this._storage, this._name, this.props, MAIN_STORAGE_SLOT)
+        this._rememberCurrentCookieProperties()
```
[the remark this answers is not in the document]
marandaneto: Fixed in the latest branch. Post-write bookkeeping now records the exact snapshot that was attempted and verifies the shared cookie before marking it observed. Reconciliation also reads and applies one stable main-cookie snapshot, so a sibling write landing after our write remains unseen and is adopted on the next sync. Added regression coverage for an update landing between the storage write and fingerprint bookkeeping.

packages/browser/src/posthog-core.ts
```diff
@@ -2724,97 +2737,120 @@ export class PostHog implements PostHogInterface {
...
+            } else if (shouldTransitionToIdentified && (userPropertiesToSet || userPropertiesToSetOnce)) {
+                this.reloadFeatureFlags()
+            }
+        } finally {
+            if (cookieSyncSuppressionStarted) {
+                this.persistence._endCookieSyncSuppression()
             }
-        } else if (shouldTransitionToIdentified && (userPropertiesToSet || userPropertiesToSetOnce)) {
-            this.reloadFeatureFlags()
         }
```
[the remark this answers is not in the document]
marandaneto: Good catch. Fixed in 0eaf87fad: identify now tracks successful completion and releases cookie synchronization suppression with publish=false when the transaction throws. Added regression coverage for a persistence exception during identify.

packages/browser/src/posthog-core.ts
```diff
@@ -1646,6 +1649,16 @@ export class PostHog implements PostHogInterface {
             return eventProperties
         }
 
+        // Cookies do not emit cross-origin storage events. Reconcile before
+        // reading any event properties so already-open sibling subdomains pick
+        // up identify/reset changes, including for replay snapshot events.
+        this.persistence.syncCookieProperties()
+        if (this.persistence.consumeCookieIdentityChange()) {
+            this._cachedPersonProperties = null
+            this.reloadFeatureFlags()
```
dustinbyrne: [P1] Clear reset-scoped flag targeting state after adopting a sibling reset

An identified → anonymous cookie transition currently updates the identity and starts an async flag reload, but local-only `$stored_person_properties` / `$stored_group_properties` and the cached flag result survive. I reproduced the first event using the new anonymous `distinct_id` with the previous user's `$feature/...` values, and the following `/flags` request sending the previous user's targeting properties under that anonymous ID.

Could reconciliation expose the transition type and perform reset-specific flag/person cleanup before event enrichment and reload? A two-subdomain Playwright regression should assert both the first event properties and the following `/flags` body.

.changeset/fresh-cookies-cross-subdomains.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/types': patch
+---
+
+Add `cookieWinsOnConflict` to keep shared cross-subdomain identity and session state ahead of stale per-origin localStorage, deprecate `__preview_cookie_wins_on_conflict`, and enable the new behavior for the `2026-08-29` defaults.
```
dustinbyrne, line 6: +4kB for this change, which isn't _huge_, but it's also not nothing for something that's fixing an edge case. i chatted with my agent about this and i'll have it post some recommendations, take them as you will

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
         return isNumber(value) && value > 0 ? value : 0
     }
 
+    private _rememberCurrentCookieProperties(props?: Properties): void {
```
dustinbyrne, line 227: Could we remove this post-write verification path and let the next synchronization observe our own cookie write? `_lastSeenCookiePropertiesFingerprint` is only assigned/reset and never read, and the no-argument branch of this method has no callers. The exact-mirror/race bookkeeping adds a fair amount of code for a narrow interleaving without closing the underlying cookie check-then-write race.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
...
+        let cookieProperties: Properties
+        try {
+            cookieProperties = JSON.parse(cookieValue) || {}
+        } catch {
+            return false
+        }
+        // If the main cookie changed while its metadata was read, retry on the
+        // next synchronization rather than applying the old values while marking
+        // the new snapshot as observed.
+        if ((cookieStore._get(this._name) || undefined) !== cookieValue) {
```
dustinbyrne, line 308: Can we drop this second read? It guards one narrow race while metadata is read, but a sibling can still write immediately after this check, so it does not establish atomicity. Applying the snapshot we read and accepting ordinary last-writer-wins semantics seems sufficient; a newer main-cookie snapshot can be picked up at the next synchronization point.

packages/browser/src/storage.ts
```diff
@@ -273,14 +335,23 @@ export const createLocalPlusCookieStore = (
...
                 if (preferCookieOnConflict) {
+                    // Built-in keys are stable and authoritative for legacy cookies too.
+                    // Separate metadata adds only custom keys understood by the writer.
+                    const authoritativeCookieProperties = [
+                        ...COOKIE_PERSISTED_PROPERTIES,
+                        ...getCookiePersistedPropertiesFromMetadata(name, cookieValue),
+                    ]
                     // Defensive: skip null / empty-string cookie values so a malformed
                     // legacy cookie cannot wipe out valid localStorage data.
                     const safeCookieProperties: Properties = {}
```
dustinbyrne, line 402: Could we let the cookie-authoritative branch merge the SDK-written cookie directly rather than carrying special null/empty-value recovery? The writer already omits undefined, null, and empty-string values. Retaining JSON parse failure handling seems sufficient; this otherwise adds malformed-storage semantics to the production path.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
...
+            clearTimeout(this._pendingSaveTimer)
+            this._pendingSaveTimer = undefined
+        }
+        // Force the complete local snapshot through without reconciling a
+        // stale sibling write observed during this authoritative transaction.
+        delete this._slotState[MAIN_STORAGE_SLOT]
+        this._writeNow(true)
+    }
+
+    _endCookieSyncSuppression(publish: boolean = true): void {
```
dustinbyrne, line 443: Could this always publish the final snapshot and release suppression, rather than supporting `publish=false` and pending-timer cancellation? The alternate path only protects identity operations that throw midway (for example from a customer hook or device-ID callback). Removing it would also remove the completion flags in both `identify()` and `reset()` while preserving the normal identity flow.

packages/browser/src/posthog-core.ts
```diff
@@ -2724,97 +2737,124 @@ export class PostHog implements PostHogInterface {
...
+                    { $set: userPropertiesToSet || {}, $set_once: userPropertiesToSetOnce || {} },
+                    false
+                )
 
-            // Update current user properties
-            this.setPersonPropertiesForFlags(
-                { $set: userPropertiesToSet || {}, $set_once: userPropertiesToSetOnce || {} },
-                false
-            )
+                if (this.config.cookieWinsOnConflict) {
```
dustinbyrne, line 2813: Do we need to publish here in addition to publishing when suppression ends? Publishing once when `identify()` completes preserves the sequential cross-subdomain guarantee. This early publication only lets a sibling observe the new identity while this identify/capture is still in progress, and the same duplicate path exists in the same-ID transition below.

packages/browser/src/posthog-persistence.ts
```diff
@@ -770,33 +1016,86 @@ export class PostHogPersistence {
...
-        this.set_secure(config['secure_cookie'])
+        const persistenceModeChanged = config.persistence !== oldConfig.persistence
+        const cookiePersistedPropertiesChanged = !isArrayContentsEqual(
+            config.cookie_persisted_properties || [],
+            oldConfig.cookie_persisted_properties || []
+        )
+        const persistenceChanged = persistenceModeChanged || cookiePersistedPropertiesChanged
+        const cookiePrecedenceChanged = config.cookieWinsOnConflict !== oldConfig.cookieWinsOnConflict
+
+        const disabled = config['disable_persistence'] || !!isDisabled
```
dustinbyrne, line 1165: Could `cookieWinsOnConflict` remain initialization-time configuration? The old/new-policy reconciliation and suppression around persistence re-enablement, cookie-option changes, and combined storage migrations adds substantial complexity beyond the core cross-subdomain behavior. Keeping the previous init-time constraint would let this retain the simpler storage migration path.

packages/browser/src/storage.ts
```diff
@@ -254,14 +254,76 @@ const COOKIE_PERSISTED_PROPERTIES = [
...
+    ;[...COOKIE_PERSISTED_PROPERTIES, ...customCookieProperties].forEach((key) => {
+        const persistedValue = value[key]
+        if (!isUndefined(persistedValue) && !isNull(persistedValue) && persistedValue !== '') {
+            cookiePersistedProperties[key] = persistedValue
+        }
+    })
+    return cookiePersistedProperties
+}
+
+const getCookieSnapshotFingerprint = (cookieValue: string): string => {
```
dustinbyrne, line 321: This sidecar protocol is a substantial amount of machinery. Could we use the locally configured `cookie_persisted_properties` list as authoritative and require consistent configuration across sibling deployments? That preserves normal custom-property propagation/removal while dropping mixed-version/config rollout protection, hashing, ordered publication, readback verification, and fallback behavior. If mixed-fleet support is a required contract, we should make that explicit; otherwise this looks like a good simplification boundary.

lricoy:
Wow this is a great one! I will try it locally first before reviewing.

packages/browser/src/sessionid.ts
```diff
@@ -402,7 +402,13 @@ export class SessionIdManager {
         }
         const timestamp = _timestamp || new Date().getTime()
 
+        const managedSessionIdBeforeCookieSync = this._sessionId
+        this._persistence.syncCookieProperties?.()
```
dustinbyrne: [P1] Avoid synchronously scanning cookies for every replay event

`checkAndGetSessionAndWindowId()` is also called for each accepted rrweb event. With this option enabled, every call reaches `document.cookie`, while ordinary captures already reconcile in core and then repeat the read here. In steady state I observed two cookie reads per ordinary event; on replay the work scales with accepted rrweb event volume.

Could we separate forced capture/write reconciliation from a throttled session/replay check, or pass an internal “already reconciled” signal? Please add deterministic read-count coverage plus a test that sibling session rotation is still adopted within the chosen bound.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
...
+        } catch {}
+        if (!cookieValue || cookieValue === this._lastSeenMainCookieValue) {
+            // Ignore sidecar-only changes. Metadata is meaningful only together
+            // with a new main snapshot, and treating it as a sibling identity
+            // update could roll back a local write when the main mirror failed.
+            return false
+        }
+        const cookieFingerprint = getCookiePropertiesFingerprint(this._name, cookieValue)
+
+        let cookieProperties: Properties
```
dustinbyrne, line 299: [P2] Reject non-object JSON cookie roots before merging

`JSON.parse()` can successfully return a string, number, boolean, or array. The subsequent `in` checks assume an object; I reproduced `posthog.init()` throwing `TypeError: Cannot use 'in' operator ... in abc` with a cookie containing the valid JSON string `"abc"`.

Could we treat non-array object roots as the only valid persistence shape and ignore other roots? A no-throw test covering primitives and arrays would preserve the SDK boundary without adding per-property recovery logic.

dustinbyrne (commented):
overall direction makes sense to me

some bot comments in here because it looks like the PostHog code review failed. it looks like some commits are still coming in so i'll circle back on this

packages/browser/src/posthog-core.ts
marandaneto: Fixed in 91271ade8 and 5bcb270e5. Cookie-driven identity transitions now invalidate person/group targeting and all cached flag state before enriching the first event or reloading flags, including direct identified-user-to-identified-user transitions. The cross-subdomain Playwright test now verifies that the first anonymous event has no prior-user feature values and the next /flags payload uses the new ID without prior person/group targeting properties.

.changeset/fresh-cookies-cross-subdomains.md
marandaneto: Thanks — I reviewed the complexity and bundle impact. The additional machinery is intentional because the fix has to cover already-open tabs, identify/reset transaction ordering, mixed-version lazy extensions, and session adoption without exposing metadata as event properties. I did simplify the hot replay path in 91271ade8 by avoiding duplicate reads and bounding direct session checks. The final branch review is clean.

packages/browser/src/posthog-persistence.ts
marandaneto: I kept the exact-write verification because localStorage+cookie reports the durable localStorage write as successful even when its cookie mirror fails. Advancing the observed snapshot without verification can hide either a failed mirror or a sibling write. The field is part of that coherent-snapshot bookkeeping, while _lastSeenMainCookieValue provides the fast comparison. Regression coverage pins both failed mirrors and the post-write sibling race.

packages/browser/src/posthog-persistence.ts
marandaneto: I kept the second read deliberately. It does not claim atomicity; it prevents applying snapshot A while recording concurrently written snapshot B as observed, which otherwise creates a stable missed-update state. If the value changes again after the check, the next synchronization still sees it because only A was recorded. This is the smallest retry boundary that keeps the observed marker coherent.

packages/browser/src/storage.ts
marandaneto: I kept the null/empty recovery for compatibility with malformed and older cookies. The writer omits these values today, but cookies survive SDK upgrades and are writable by sibling origins. Allowing malformed values to erase valid local identity state during initialization is avoidable. 91271ade8 additionally rejects non-object roots and unsafe property names while preserving valid falsy values.

packages/browser/src/posthog-persistence.ts
marandaneto: I kept publish=false because identify/reset can execute customer callbacks and device-ID hooks that throw. Publishing an incomplete identity transaction would expose a half-reset or half-identify snapshot to sibling origins. The failure path now always releases suppression, cancels pending publication, and leaves the last complete cookie snapshot authoritative; normal successful operations still publish exactly once at transaction completion.

packages/browser/src/posthog-core.ts
marandaneto: The publication before capture is intentional: identify captures $identify using the completed identity snapshot, and sibling tabs must not observe the old identity after that event has been emitted. The suppression-end publication is a guarded finalization fallback; the explicit publication makes the identify event and cookie state sequentially consistent. Tests cover both normal and exceptional identify paths.

packages/browser/src/posthog-persistence.ts
marandaneto: I kept set_config support because persistence mode, persistence enablement, cookie options, and this supported config option are all mutable today. Ignoring a runtime policy change would leave the in-memory/storage precedence dependent on initialization history. The migration is suppression-guarded so the current complete state is published once rather than adopting a sibling midway.

packages/browser/src/storage.ts
marandaneto: Mixed-version/config support is a required contract here: browser extensions are served unversioned and can run against older cached cores, while sibling subdomains commonly deploy independently. A local-only list cannot distinguish an authoritative removal from an older writer that never knew the custom key. The snapshot-bound sidecar preserves that distinction without leaking metadata into events; built-in identity/session fields still propagate when the sidecar cannot persist.

packages/browser/src/sessionid.ts
marandaneto: Fixed in 91271ade8. Ordinary captures now tell SessionIdManager that persistence was already reconciled, eliminating the duplicate cookie read. Direct/replay session checks are bounded to a 1-second reconciliation interval. Deterministic unit coverage asserts the read count and that a sibling rotation is adopted at the bound.

packages/browser/src/posthog-persistence.ts
marandaneto: Fixed in 91271ade8. Cookie persistence now accepts only non-array object roots and copies only safe own keys. Primitive and array roots are ignored without throwing; tests cover strings, numbers, booleans, arrays, and unsafe property names.

packages/browser/src/posthog-core.ts
```diff
@@ -234,6 +235,7 @@ const defaultsThatVaryByConfig = (
     split_storage: !!(defaults && defaults >= '2026-05-30'),
     detect_google_search_app: !!(defaults && defaults >= '2026-05-30'),
     disable_capture_url_hashes: !!(defaults && defaults >= '2026-06-25'),
+    cookieWinsOnConflict: !!(defaults && defaults >= '2026-08-29'),
```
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. Explicit defaults: "unset" now keeps cookieWinsOnConflict disabled, with regression coverage. I limited the change to this newly introduced default rather than expanding scope to the pre-existing defaults comparisons.

packages/browser/playwright/compat-skips.ts
```diff
@@ -14,6 +14,11 @@ import { satisfies } from 'compare-versions'
 
  */
 export const compatSkips: { range: string; test: string; reason: string }[] = [
+    {
+        range: '<1.415.5',
+        test: 'already-open sibling subdomains adopt identify and reset cookie changes',
+        reason: 'live cross-subdomain cookie reconciliation is added by this change',
+    },
```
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. The changeset is now minor, so the compat skip covers versions below 1.417.0, including the currently published 1.416.x releases that do not contain this feature.

packages/browser/src/posthog-core.ts
```diff
@@ -1646,6 +1649,16 @@ export class PostHog implements PostHogInterface {
...
+        // Cookies do not emit cross-origin storage events. Reconcile before
+        // reading any event properties so already-open sibling subdomains pick
+        // up identify/reset changes, including for replay snapshot events.
+        this.persistence.syncCookieProperties()
+        if (this.persistence.consumeCookieIdentityChange()) {
+            this._cachedPersonProperties = null
+            this.reloadFeatureFlags()
+            this.unregister(FLAG_CALL_REPORTED)
+        }
+
```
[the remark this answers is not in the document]
marandaneto: I investigated this and kept identity cleanup active for read-only calculations. syncCookieProperties has already adopted the new identity, so deferring cleanup would let a replay snapshot use the new distinct ID with the prior user’s cached flag/person context. The read-only contract prevents event/session advancement, but externally caused identity reconciliation must remain internally coherent. Duplicate session cookie reads are now avoided and direct replay checks are throttled in 91271ade8.

packages/browser/src/sessionid.ts
```diff
@@ -402,7 +402,13 @@ export class SessionIdManager {
...
         let [, sessionId, startTimestamp] = this._getSessionId()
+        const cookieSessionAdoption =
+            !isNull(managedSessionIdBeforeCookieSync) &&
+            !isUndefined(managedSessionIdBeforeCookieSync) &&
+            sessionId !== managedSessionIdBeforeCookieSync
```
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. The new-session branch now explicitly resets crossTabAdoption to false, while the existing-session branch still reports actual sibling adoption. Regression coverage pins both reset-generated sessions and adopted sessions.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
         return isNumber(value) && value > 0 ? value : 0
     }
 
+    private _rememberCurrentCookieProperties(props?: Properties): void {
+        if (!this._config.cookieWinsOnConflict || this._config.persistence.toLowerCase() !== 'localstorage+cookie') {
```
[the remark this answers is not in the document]
marandaneto: The camelCase name is intentional and is part of this change’s requested public API contract: cookieWinsOnConflict. It also follows recent supported options such as rewriteRequestPath and addTracingHeaders. Renaming it now would break the agreed API and the documentation/reference surface; the deprecated preview snake_case option remains mapped for compatibility.

.changeset/fresh-cookies-cross-subdomains.md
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. Both posthog-js and @posthog/types now receive minor changesets, matching the cited additive-config precedent.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +201,198 @@ export class PostHogPersistence {
...
+        cookiePersistedProperties.forEach((key) => {
+            if (
+                authoritativeCookieProperties.indexOf(key) !== -1 &&
+                !(key in cookieProperties) &&
+                !invalidCookieProperties[key]
+            ) {
+                delete nextProps[key]
+            }
+        })
+        this.props = extend(nextProps, cookieProperties)
```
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. Cookie-derived JSON is validated as a non-array object and copied through a safe own-key filter that drops __proto__, constructor, and prototype before any merge. Initialization and live reconciliation share the sanitizer, with regression coverage for prototype integrity.

packages/browser/src/posthog-core.ts
```diff
@@ -1646,6 +1649,16 @@ export class PostHog implements PostHogInterface {
...
 
+        // Cookies do not emit cross-origin storage events. Reconcile before
+        // reading any event properties so already-open sibling subdomains pick
+        // up identify/reset changes, including for replay snapshot events.
+        this.persistence.syncCookieProperties()
+        if (this.persistence.consumeCookieIdentityChange()) {
+            this._cachedPersonProperties = null
+            this.reloadFeatureFlags()
+            this.unregister(FLAG_CALL_REPORTED)
+        }
```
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. Both sibling identity-transition tests now seed _cachedPersonProperties and assert that it is cleared; the identified-user transition also asserts person/group targeting cleanup.

packages/browser/src/sessionid.ts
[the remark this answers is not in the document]
marandaneto: Fixed in 91271ade8. A null manager session after reset is now distinct from undefined cold start, so adopting a sibling non-null session notifies listeners. A regression test preserves the existing window ID and asserts crossTabAdoption plus the callback.

marandaneto:
> overall direction makes sense to me
> 
> some bot comments in here because it looks like the PostHog code review failed. it looks like some commits are still coming in so i'll circle back on this

addressed the comments
yeah i was fighting CI and my autoreview skill

arnohillen (approved):
[arno's agent] Reviewed at head 77d3c372: change verified against head code, CI classified, no blocking findings.

arnohillen:
[arno's agent] A pending cookie-identity change consumed mid-identify wipes the person properties that identify call just set for flags, since the adoption path rebuilds flag context from the cookie state.

arnohillen:
[arno's agent] Sibling reset adoption keeps the previous user's $groups, so the new identity inherits group associations the reset should have cleared.

arnohillen:
[arno's agent] Old-SDK sibling cookie writes delete falsy built-ins like $session_recording_is_sampled: false because the merge treats false as absent, flipping sampled-out users back in.

marandaneto:
@arnohillen Addressed all three agent findings in signed commits `728dd5bb3` and `92723326b`:

- [Pending identity cleanup during `identify()`](https://github.com/PostHog/posthog-js/pull/4496#issuecomment-5286474643): pending sibling identity changes are now consumed and cleaned before the explicit identify properties are applied, so the identify capture cannot wipe the new flag context.
- [Sibling reset retaining `$groups`](https://github.com/PostHog/posthog-js/pull/4496#issuecomment-5286474826): adopting an anonymous/reset identity now removes `$groups`, including during initial cookie reconciliation. The two-subdomain Playwright test asserts that the first post-reset event has no previous groups and the next `/flags` request has no prior group targeting properties.
- [Legacy writers omitting falsy built-ins](https://github.com/PostHog/posthog-js/pull/4496#issuecomment-5286475009): current writers now publish snapshot-bound metadata even without custom cookie keys. A missing/stale marker identifies a legacy writer, so omitted local `false`/`0` built-ins are preserved; a matching current marker keeps omissions authoritative.

I also mirrored live identity-bound cache cleanup during initial cookie merge after autoreview identified the equivalent reopened-tab path. Focused unit tests (609 passed, 1 skipped), Chromium cross-subdomain Playwright (3 passed), typecheck, ESLint, build, and exact-HEAD autoreview all pass.

arnohillen:

[arno's agent] The Unit tests job fails at this head: persistence-key-policy.test.ts flags posthog-core.ts:1642 because identityBoundProperties conditionally pushes the raw '$groups' literal, which the checker cannot resolve to a persistence key constant (run 31776496341, 1 failed of 5478).

arnohillen:

[arno's agent] The remaining Unit tests failure at 51e440f1 (lazy-sessionrecording pagehide flush, run 31791176297) also failed identically on the unrelated branch refactor/browser-autocapture-extension-lifecycle (run 31738063539) and passed on this PR's previous head, so it looks like a pre-existing flake; rerun triggered.

packages/browser/src/posthog-persistence.ts
```diff
@@ -279,6 +512,54 @@ export class PostHogPersistence {
         if (this._splitStorage) {
             this._loadGroupEntries()
         }
+
+        if (reconcileCookieIdentity && entry) {
+            const previousDistinctId = localEntryBeforeMerge?.[DISTINCT_ID]
+            const previousUserState = localEntryBeforeMerge?.[USER_STATE] ?? USER_STATE_ANONYMOUS
+            const nextDistinctId = entry[DISTINCT_ID]
+            const nextUserState = entry[USER_STATE] ?? USER_STATE_ANONYMOUS
+            if (nextDistinctId !== previousDistinctId || nextUserState !== previousUserState) {
```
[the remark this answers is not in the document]
marandaneto: Fixed in `9e9afdf9f` and `26b15d688`: startup identity adoption is marked pending, and the pending cleanup marker survives a background persistence write plus reload. Core clears session persistence before the first event under the adopted anonymous identity. Regression coverage verifies both startup detection and reload behavior.

packages/browser/src/posthog-persistence.ts
```diff
@@ -174,6 +202,200 @@ export class PostHogPersistence {
...
+                delete cookieProperties[key]
+            }
+        })
+        if (isEmptyObject(cookieProperties)) {
+            return false
+        }
+
+        const previousDistinctId = this.props[DISTINCT_ID]
+        const previousUserState = this.props[USER_STATE]
+        const nextProps = extend({}, this.props)
```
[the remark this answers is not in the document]
marandaneto: Good catch. Fixed in `9e9afdf9f` and hardened through `36c985db5`: sibling resets now clear stale event-visible local properties during both live and startup reconciliation, while preserving fields explicitly carried by the new cookie and mutations made after reconciliation. Regression coverage includes unit and two-subdomain browser assertions.

marandaneto:
@dustinbyrne @lricoy can you pls double check?

arnohillen (approved):

Re-approving at 36c985db5: all three earlier findings are fixed with dedicated unit and playwright coverage (pending sibling identity consumed before explicit person properties in identify, sibling reset clears $groups in live/startup/split-storage paths, legacy falsy built-ins preserved via the metadata sidecar), CI fully green.

arnohillen:

Leftover editing fragment in `_processCookieIdentityChange`: the comment line "properties before this event is assembled under the anonymous ID." (posthog-core.ts:1637) duplicates the sentence directly above it and reads as a dangling clause.

marandaneto:
@arnohillen Fixed the dangling duplicate comment in signed commit `25d1a0ced`. Thanks for spotting it.

arnohillen (approved):

Re-approving at 25d1a0ce: delta is a main merge (release churn only) plus the removal of the duplicated comment line flagged earlier; no source logic changed, CI fully green.

lricoy (approved):
LGTM! Re-reviewed after the ping

turnipdabeets (approved):
Post-merge review of the full production-source diff (persistence, storage, core, session manager, config/types). Solid work — the pieces that make this safe are all in place:

- Sync points cover every read/write edge: before event-property calculation (including replay), before persistence writes (closing the debounce race), on register/unregister/group, and a bounded 1s check in the session manager so replay events don't scan `document.cookie` per event.
- The suppression window makes `identify()`/`reset()` atomic against sibling writes, with try/finally releasing suppression and refusing to publish a partial identity when a customer hook throws.
- Cookie parsing filters `__proto__`/`constructor`/`prototype`, validates `$user_state` against the enum, and drops null/empty values, so a malformed or hostile cookie can't wipe or pollute local state.
- The metadata sidecar's fingerprint binding to the exact main-cookie snapshot (with the built-in-only fallback when the sidecar can't land) handles mixed-version and legacy writers conservatively.
- Default only flips on for the `2026-08-29` defaults snapshot with the `__preview_` alias kept, so existing setups are untouched.

The identity-bound-local-properties cleanup (flags, person props, `$user_id`, alias) on adopted identity changes matches what a local `identify()`/`reset()` would have done, which is exactly the invariant this needed.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4497, opened 2026-08-11
fix(surveys): stop Chrome renderer crash (grey tab) on survey close

hpouillot:
## Problem

Closing an in-app survey could crash the Chrome tab — the whole page went grey ("Aw, Snap") — on heavy SPAs. A JS exception cannot do that; this is a **renderer (tab) crash**, and the trigger was the survey close path.

## Root cause

The close path in `usePopupVisibility` wrapped the survey container's DOM removal inside the View Transitions API:

```js
const transition = document.startViewTransition(() => {
  surveyContainerRef?.current?.remove()
})
transition.finished.then(() => { ... })
```

Removing the element **inside** the transition callback left the transition's captured snapshot pointing at a removed node. On complex pages this triggered a Chromium renderer crash (a known class of view-transition bug — Chromium landed a related fix in Oct 2025, `b83917a`, crashtest `replace-html.html`). The result was a flat grey crashed tab, not a thrown exception — which is why it looked like "the page crashed" rather than a normal survey bug.

## Fix

- The transition callback now only **animates a fade-out** (`opacity: 0`). React tears the container down via `setIsPopupVisible(false)` + `removeSurveyFromFocus` once the transition settles, so the snapshot never points at a removed node.
- **Guard against overlapping transitions** (a second close while one is animating — e.g. Enter + button click), which is itself a renderer crash risk.
- **Always settle the popup state** when `transition.finished` resolves OR rejects (skipped/interrupted transition, e.g. tab backgrounded / reduced motion), so the survey can never be left visible with a stale reference.
- **Fall back to a plain synchronous remove** if `startViewTransition` throws (e.g. during unload).

## Tests

The view-transition close path was previously untested — jsdom has no `startViewTransition`, so only the synchronous fallback path ever ran. Added a `usePopupVisibility view-transition close path` suite with a controllable `startViewTransition` mock:

- container is **not** removed inside the transition callback (only faded), and the popup hides + `removeSurveyFromFocus` runs once the transition settles
- an **interrupted** transition (`finished` rejects) still hides the popup — no stale visible state
- repeated close calls do **not** start a second transition
- `startViewTransition` throwing falls back to a plain remove
- no-API path still tears down synchronously (unchanged behaviour)

All 236 survey tests pass; prettier + eslint clean.

## Notes

- This is a real-world crash reported against the PostHog app page itself (a heavy SPA), reproduced on close.
- Behaviour is unchanged for browsers without the View Transitions API and in jsdom.
- Includes a changeset (`posthog-js: patch`).

### DIFF
--- a/.changeset/survey-view-transition-crash.md
+++ b/.changeset/survey-view-transition-crash.md
@@ -1,0 +1,9 @@
+---
+'posthog-js': patch
+---
+
+Fix a Chrome renderer crash (grey "Aw, Snap" tab) that could occur when closing an in-app survey.
+
+The survey close path wrapped the survey container's DOM removal in `document.startViewTransition`. Removing the element inside the transition callback left the captured snapshot pointing at a removed node, which on heavy SPAs triggered a Chromium renderer crash and took down the whole tab.
+
+The close path now only animates a fade-out inside the transition and lets React tear the container down once the transition settles. It also guards against overlapping transitions (a second close while one is animating) and always settles the popup state if the transition is skipped or interrupted, so the survey can never be left visible with a stale reference.
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -262,6 +262,199 @@ describe('usePopupVisibility', () => {
     })
 })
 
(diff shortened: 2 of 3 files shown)


pauldambra (approved):
nice

Issue #4449, opened 2026-08-06
@posthog/mcp: instrument() silently no-ops on MCP TypeScript SDK v2 servers (string setRequestHandler registrations are never wrapped)


OrbitingBucket:
### Environment

- `@posthog/mcp` 0.10.7
- `@modelcontextprotocol/server` 2.0.0 (the new modular SDK v2, stable)
- Server framework: `@rekog/mcp-nest` 2.0.0 (NestJS microservice strategy; stateless streamable HTTP — a fresh `McpServer` per request, handlers bound after `serverMutator`)
- Node 26

### Symptom

`instrument(server, posthog, options)` produces **zero telemetry** on SDK v2 servers, silently. Because `instrument()` catches internally and returns a no-op handle, nothing throws — the integration looks healthy while `$mcp_*` events stop entirely. For a team relying on MCP analytics for adoption reporting, this failure mode is invisible until someone notices flat dashboards.

### Three composition defects (in order encountered)

1. **High-level compat check requires the removed `.tool()` method.** `isCompatibleServerType` demands `typeof server.tool === "function"` for any object with a `.server` property. SDK v2's `McpServer` only has `registerTool()`, so the high-level entry throws (internally). The low-level `server.server` passes every low-level check (`setRequestHandler`, `_requestHandlers` Map, `getClientVersion`, `_serverInfo`) — so the gap is narrow.

2. **The `setRequestHandler` patch only recognizes Zod-schema registrations.** `patchRequestHandlers` derives the handler name via `getObjectShape(requestSchema)` + `getLiteralValue(shape.method)`. SDK v2 (and frameworks built on it) register handlers with **string method names**: `server.setRequestHandler("tools/call", handler)`. Those fall through unwrapped — so any handler registered *after* `instrument()` (the normal order for per-request/stateless servers, where instrumentation runs in a server-mutator hook before handlers are bound) silently **replaces** the analytics wrapper. Result: no `$mcp_tool_call`, no `$mcp_tools_list`, no intent injection.

3. **Context/intent ownership is only learned in the tools/list wrapper.** `cacheToolAnalyticsParameterOwnership` runs inside the wrapped `tools/list`. A stateless server instance serves exactly one request, so a `tools/call` never shares an instance with a `tools/list` → `analyticsOwnsContext` is always false → the `context` argument is never captured as `$mcp_intent` (and never stripped from the downstream request), even when wrapping works.

### Validated fix (small)

[...]

PR #4502, opened 2026-08-12
fix(mcp): capture $mcp_intent when ownership cannot be resolved

gesh:
## The issue

`$mcp_intent` is never captured on a server that builds a **fresh instance per HTTP request** — `createMcpHandler`, or `@rekog/mcp-nest` with `statelessMode: true`. `context` is on by default, so this is the default path on those topologies.

The trigger is instance lifetime, not statelessness: a server that is stateless at the transport (`sessionIdGenerator: undefined`) but keeps one long-lived server object is unaffected. And it is **not** specific to the 2026-07-28 revision — SDK v1 with a per-request instance is broken in exactly the same way; v2 only makes that topology the norm, since it is the only one that can serve the new revision.

## Where it comes from

We inject a `context` parameter while serving `tools/list`, and cache "we own this parameter" on the server instance. When the next request builds a new instance, that instance never served a listing — ownership read `false`, and the agent's answer was discarded.

The defect is in the type, not the lookup: a boolean collapsed *"the host declares it"* and *"we cannot tell"* into the same value, so the answer that is correct for **stripping** was also applied to **capture**, where it is wrong.

## The fix

Ownership becomes three-valued, and the two questions it answers are decided separately:

| | Capture as `$mcp_intent` | Strip before the tool runs |
|---|:--:|:--:|
| Ours | yes | yes |
| Host's | no | no |
| **Unknown** | **yes** ← the fix | no |

Reading fails open, stripping fails closed. Deleting an argument the host declared costs them the call; reading one costs at worst a mislabelled property in their own project, capped by the existing truncation and escapable with `context: false` or `beforeSend`.

**Stripping behaviour is unchanged in every state**, so no host tool starts losing an argument. Injection is untouched.

ADR-0011 records the rule and closes ADR-0004's open follow-up, which proposed a process-scoped ownership cache — measured to hand one tenant's arguments to another, so it is corrected rather than left standing.

## Test coverage

Unit: 636/636, lint clean. Acceptance harness (`run-all.mjs --local`):

| | Before | After |
|---|:--:|:--:|
| `nest v1` · both paths | 15/16 | **16/16** |
| `nest v2` · both paths | 25/33 | **35/37** |
| dual-era matrix | — | unchanged, all `v1` rows green |
| late-handler · first-call-error probes | 9/9 · 3/3 | 9/9 · 3/3 |

- `nest v1` is `@rekog/mcp-nest` 1.9.x on **SDK v1** with `statelessMode: true`, and `$mcp_intent` was the single assertion keeping it off a perfect score. 16/16 is the proof this was never a v2-only defect. (The dual-era matrix has no intent column, so its green `v1` rows never covered this.)
- The `nest v2` denominator moved 33 → 37 in the same run: the harness `identify()` fixture used to read `extra.requestInfo` the v1 way and fail *on purpose*, standing in as the ADR-0006 assertion. It now uses the exported `getRequestHeaders()`, and that invariant is asserted directly instead, adding two assertions per era.
- Still red, not ours: `error message is clean` — NestJS's `RpcExceptionsHandler` replaces the thrown error before it reaches the MCP layer.

Closes the ownership half of #4449.


packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -109,6 +109,22 @@ export async function captureToolCall(params: TraceToolCallParams): Promise<unkn
     parameterOwnership,
     resolvedEventType === MCPAnalyticsEventType.mcpMissingCapability
   )
+  // Reading the `context` argument and removing it are different questions, and
+  // only one of them is dangerous.
+  //
+  // Removing an argument the *application* declared costs the customer their
+  // call, so the strip below stays gated on positive ownership. Reading it costs
+  // at worst a mislabelled property in the customer's own project — and refusing
+  // to read it costs the intent of every call on a stateless server, where no
```
marandaneto: suggestion: Narrow the stateless-server explanation

This implies every stateless server loses ownership, but a transport-stateless server using one long-lived server object still caches ownership after `tools/list`. Could this refer specifically to servers that create a fresh instance per request, and align the README’s “any stateless topology” wording too?

gesh: Applied in d0c0474. Narrowed both: the code comment now says "a server that builds a fresh instance per request", and the README drops "any stateless topology" for "`@rekog/mcp-nest` in its stateless mode" — plus an explicit line that a transport-stateless server keeping one long-lived server object learns ownership from its first `tools/list` and keeps it.

marandaneto (approved):

PR #4503, opened 2026-08-12
fix(dead-clicks): treat visibility and focus changes as liveness signals, not dead-click evidence

pauldambra:
## Problem

The dead-click detector was flagging a large share of ordinary clicks as `$dead_click`. The dominant source was the visibility branch: it measured the gap between a click and the most recent `visibilitychange` with `Math.abs(...)` and, once that gap exceeded a threshold, timed the click out as **dead**. Because the handler only recorded the tab becoming *visible*, any click in a session where the tab had ever been backgrounded looked like it had a multi-second "response" and was flagged — the reported delays were far too large to be real reactions to the click.

That logic is also backwards: a visibility or focus change near a click is evidence the click **did something** (it woke/focused the tab, opened a new tab, or opened a new window/popup), not evidence that it was dead. And the liveness detection was incomplete — a click that opens a new window/popup often leaves the tab visible, so its only trace is the current window losing focus, which wasn't observed at all.

**Why:** these false positives make the dead-click signal noisy for every customer and drown out genuinely dead clicks. Raised from an investigation in a Slack thread.

## Changes

- **Visibility/focus are now liveness-only — they can only ever *suppress* a dead click, never cause one.** The visibility signal no longer feeds the dead-marking path.
- **Broader, correct liveness detection:** visibility changes are recorded in *both* directions (a click that opens a new tab sends the current tab to `hidden`), and a new window `focus`/`blur` observer covers the new-window/popup case where the tab stays visible.
- **Wake-up window:** a click within a 1s interaction window (a dedicated constant, wide enough for a real "tab back, then click the page" gesture) of any visibility/focus change is treated as the click that woke/focused the tab and suppressed — rather than the previous 100ms, which was too tight to catch real wake-up clicks.
- A genuinely dead click — no mutation, scroll, selection, visibility, or focus change — is still caught by the absolute-timeout backstop.
- `$dead_click_visibility_changed_timeout` stays in the payload (always `false`) so the event shape is unchanged; a new `$dead_click_focus_changed_delay_ms` is emitted for observability (`focusChangedDelayMs` added to the candidate type).
- Added an overview comment documenting every gate / suppression / timeout signal, and unit tests for the wake-up, new-window, and stale-change cases.

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored by the PostHog Slack app after a maintainer challenged an earlier position-based explanation of a dead click. Confirmed dead-click detection is purely activity-based, queried `$dead_click` events across the affected pages, and found the visibility branch dominating the volume with implausibly large delays. Through review feedback the fix evolved from "only count post-click visibility changes" to the correct model: visibility **and** focus are liveness signals (a click opening a new tab/window may only surface as a visibility or focus change), so they suppress but never mark a click dead, with a wake-up window sized for a real gesture. Verified locally: dead-click + swipe unit suites pass, typecheck/lint/prettier clean; the mangled-property-names list was regenerated for the new private members.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C03PB072FMJ/p1786270706603799)*


pauldambra:
Follow-up tracked in #4507: treat an in-flight `fetch`/`XHR` started shortly after a click as a liveness signal, to catch async-action buttons whose visible response lands after the mutation window. Scoped as a separate PR because it involves network instrumentation (best done by reusing the recorder's existing hook rather than re-patching globals).

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -43,6 +43,12 @@ function deadEventName(candidate: DeadClickCandidate): '$dead_click' | '$dead_sw
 // root — so a swipe over them can never be judged dead
 const UNOBSERVABLE_SURFACE_SELECTOR = 'canvas,video,audio,embed,object'
 
+// a click within this window of the tab (re)gaining visibility is treated as the click that
+// woke/focused the tab and suppressed rather than flagged dead. wider than the other thresholds
+// because a real "tab back, then click the page" gesture has a human-scale gap (refocus, move the
+// mouse, click); still short enough that a genuinely dead click a while after refocusing is caught.
+const VISIBILITY_CHANGE_SUPPRESSION_MS = 1000
```
pauldambra: > [!NOTE]

**[router]** ⚪ NIT

`VISIBILITY_CHANGE_SUPPRESSION_MS` is a hard-coded module constant, unlike `scroll_threshold_ms` / `selection_change_threshold_ms` / `mutation_threshold_ms` which are configurable via `DeadClicksAutoCaptureConfig`. Not a bug — the suppress-only signals (visibility, focus) share this one window by design. Worth noting only if a future consumer needs to tune it (e.g. SPA apps with unusual tab-switch UX).

pauldambra:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers (qa-team, paul-reviewer, xp-reviewer, security-audit as warranted)

## Verdict: ✅ APPROVE <sub>(round 2 @ 11fb2626)</sub>

Round 2 reviewed the no-behavior simplify refactor (constant rename + `absDelay` helper extraction). Verified behavior-preserving and consistent; no findings. The underlying dead-clicks change remains correct.

### Key findings

None this round.

### Convergence

None — single cheap-first pass; router judged the change low-danger and did not delegate.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | LOW danger / HIGH confidence. Confirmed `absDelay(clickTimestamp, lastSeenAt)` is character-equivalent to the two original inline `lastSeenAt ? Math.abs(...) : undefined` expressions, the `LIVENESS_SUPPRESSION_MS` rename has no lingering old references in source, and no logic branches changed. No delegation warranted. |

<details>
<summary>Previous rounds (1)</summary>

round 1 @ b240793c — ✅ APPROVE: one NIT (hard-coded `VISIBILITY_CHANGE_SUPPRESSION_MS`), addressed by the round-2 rename/refactor; NIT thread resolved.

</details>

---

packages/browser/src/entrypoints/dead-clicks-autocapture.ts
```diff
@@ -216,9 +258,21 @@ class LazyLoadedDeadClicksAutocapture implements LazyLoadedDeadClicksAutocapture
     }
 
     private _onVisibilityChange = (): void => {
-        if (document?.visibilityState === 'visible') {
-            this._lastVisibilityChange = Date.now()
-        }
+        // record both directions: a tab going _hidden_ right after a click (the click opened a new
+        // tab) is as much a liveness signal as it becoming visible (the click that woke the tab)
+        this._lastVisibilityChange = Date.now()
```
marandaneto: blocking: Later transitions can erase the relevant liveness signal

Only the latest visibility/focus timestamp is retained. If this click hides or blurs the page and `_checkClicks` is suspended, the later visible/focus event overwrites the click-correlated transition; after more than one second, the absolute timeout can incorrectly capture it as dead. Please record these signals on each queued candidate when the event fires, and cover delayed hide→show and blur→focus sequences.

marandaneto (approved):

PR #4504, opened 2026-08-12
fix(mcp): resolve identify on tools/list

gesh:
`tools/list` was the only instrumented request path that never invoked the `identify` callback. It read whatever identity an earlier request had cached on the server instance, and inherited it when one was there.

`initialize` and `tools/call` each resolve `identify` themselves, so on a long-lived server the listing picked up the identity the handshake had already cached and looked correct. Where the next request builds a new instance, that cache is always empty — so every `$mcp_tools_list` was attributed to the session id and stamped `$process_person_profile: false`, while the surrounding tool calls were attributed to the user.

The callback is now awaited before the listing is served, as on the other two paths.

## Testing

- New regression test in `identify.test.ts`: a bare `tools/list` with no preceding tool call, so nothing has cached an identity. Fails on `main`.
- Full package suite passes. `concurrent-attribution.test.ts` still holds — its cross-session isolation case returns `null` for `tools/list`, so it is unaffected.
- Verified against a per-request NestJS + MCP SDK v2 server: `$mcp_tools_list` now carries the user's distinct id, and the `announcedAtInitialize` guard keeps it from publishing a second `$identify` on a token session.

### DIFF
--- a/packages/mcp/src/extensions/instrumentation.ts
+++ b/packages/mcp/src/extensions/instrumentation.ts
@@ -518,20 +518,26 @@ export async function handleListToolsRequest(
   // Snapshot before identify, metadata resolution, or the list handler can yield
   // to a concurrent request using the same instrumented server.
   const sessionInfo = getSessionInfo(server, data, sessionId)
-  // `getSessionInfo` only surfaces an identity some earlier request already
-  // cached on this instance. On a per-request instance that cache is always
-  // empty, so without resolving `identify` here too, `tools/list` would be the
-  // one request path that never attributes to a person.
-  const identity = data ? await handleIdentify(server, data, sessionId, request, sessionInfo, extra) : undefined
-  const requestAttribution = withIdentity(sessionInfo, identity)
   const event: McpEvent = {
     sessionId,
     parameters: buildCapturedMcpParameters(request),
     eventType: MCPAnalyticsEventType.mcpToolsList,
     timestamp: startTime,
   }
+  // Stamp before the identify await below. The last link of the identity chain is
+  // the server's own `getClientVersion()`, which a concurrent `initialize` can
+  // replace while a slow identify callback is in flight — and `captureEvent`
+  // prefers a stamped value over the `sessionInfo` snapshot, so a late stamp
+  // would win with the wrong client's name.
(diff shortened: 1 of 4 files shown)


packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -515,9 +515,15 @@ export async function handleListToolsRequest(
...
-  // concurrent request using the same instrumented server.
-  const requestAttribution = getSessionInfo(server, data, sessionId)
+  // Snapshot before identify, metadata resolution, or the list handler can yield
+  // to a concurrent request using the same instrumented server.
+  const sessionInfo = getSessionInfo(server, data, sessionId)
+  // `getSessionInfo` only surfaces an identity some earlier request already
+  // cached on this instance. On a per-request instance that cache is always
+  // empty, so without resolving `identify` here too, `tools/list` would be the
+  // one request path that never attributes to a person.
+  const identity = data ? await handleIdentify(server, data, sessionId, request, sessionInfo, extra) : undefined
```
marandaneto: blocking: Stamp client metadata before awaiting identification

`handleIdentify` can yield before `stampClientIdentity` runs. If another client initializes on this server meanwhile, `getClientVersion()` may switch, causing this request's `$mcp_tools_list` event to be stamped with the other client's metadata because capture prefers stamped values over the earlier `sessionInfo` snapshot. Please build and stamp the event before this await while continuing to use the identity-enriched `requestAttribution`, and add an interleaving regression test.

marandaneto (commented):
Inline review finding.

packages/mcp/src/extensions/instrumentation.ts
gesh: Good catch — fixed in d1337f49a: the event is built and stamped before the identify await, with `requestAttribution` still computed after. Added an interleaving test in `concurrent-attribution.test.ts` (a second handshake lands mid-callback); it fails without the reorder. Note the same window exists on `prepareToolCallEvent` and `handleInitializeRequest` — leaving those for a separate PR since `$mcp_tool_call` volume deserves its own review.

marandaneto (approved):

PR #4509, opened 2026-08-12
fix(replay): take a full snapshot on idle wake when mutations were dropped

ksvat:
## Problem

- A replay can show duplicated, overlapping DOM after an idle pause in the recording, while the live page was fine.
- While recording is idle, rrweb keeps observing the DOM, but the recorder drops every emitted event at the idle gate - including the mutations that create new nodes.
- On wake, later mutations reference those never-emitted nodes. The player cannot find them, skips the removes, and grafts the new subtree next to the stale one.
- The wake path only schedules the next periodic full snapshot, up to `full_snapshot_interval_millis` (default 5 minutes) away. If the window goes quiet before the timer fires, the corruption is permanent for that window.
- #1411 (Sept 2024) introduced this: it replaced the immediate snapshot on idle-return with `_scheduleFullSnapshot()`. An existing unit test still carries a comment promising the immediate snapshot while asserting its absence.

Analysis of an affected recording export confirmed the mechanism: the pre-idle full snapshot topped out at node id 3068, and the first post-wake mutations removed and re-parented nodes 3069-3444 that no emitted event ever created.

## Changes

- On return from idle, the recorder now takes an immediate full snapshot when mutations were dropped during the idle period, then schedules the periodic timer as before.
- A counter tracks mutation events discarded by the idle gate. A clean idle (nothing dropped) skips the extra snapshot, so parked tabs pay no new cost.
- This mirrors the URL-unblock resume path, which already takes an immediate snapshot before rescheduling.

> [!NOTE]
> The extra full snapshot fires at the first interaction after an idle period in which the DOM changed. On very large DOMs this is a main-thread task, but it replaces a corrupted replay, and only fires when the mirror is actually stale.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Testing

- New unit test replays the failing sequence (idle, dropped mutation, wake) and asserts the full snapshot enters the buffer before the wake event.
- A second unit test asserts a clean idle takes no immediate snapshot on wake.
- The existing 5-minute idle test now asserts the full snapshot its comment always promised.
- New Playwright test drives a real browser through an idle DOM swap and asserts the wake snapshot precedes post-wake incrementals and contains the idle-era DOM. Verified locally on Chromium in both directions: it passes with this fix, and fails on the missing wake snapshot against main's recorder bundle.

## 🤖 Agent context

- Claude Code (Opus 4.8) drove the investigation and fix under Kim's direction, starting from a customer replay-corruption report.
- Diagnosis came from parsing an exported recording: post-wake mutations referenced node ids that no emitted event created, isolating the idle drop gate as the cause. `git log -S` traced the regression to #1411.
- Skills invoked: writing-pr-descriptions. Unit tests were run locally (770/770 replay suite green); the Playwright spec is new and unverified locally.


ksvat:
* **#4509** <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4509?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/PostHog/posthog-js/4509?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

ksvat:
## Incident-pattern review (replay-incident-risk)

The pre-push check matched this diff against two past incident classes. Judgment pass on each:

### Class 1: version skew across the lazy-load boundary

**Verdict: not exposed.**

1. No cross-boundary signature changes. The new field and all three call sites are private to `LazyLoadedSessionRecording`, inside the recorder bundle.
2. No new calls into the core. `takeFullSnapshot` is called on the rrweb record instance bundled with the same lazy file.
3. No persisted state. The counter is in-memory, initialized to 0, reset on idle entry and on wake. A core of any age produces the same behavior.
4. The compat suite (`playwright.config.compat.ts`) runs the whole mocked directory, so the new idle-wake spec also runs against the previous core with the new recorder bundle.

The unversioned-bundle fact still applies: merging publishes this to every SDK version at once. That argues for the small diff shape (12 lines, additive, no rotation or flush semantics touched).

### Class 2: session rotation, idle handling, and flush semantics

**Verdict: billing-neutral by construction; walked the INC-975 scenarios:**

1. **Idle background tab over 24h: zero change.** The snapshot fires only on `returningFromIdle`, which requires an `ACTIVE_SOURCES` incremental - a real user interaction. A parked tab never wakes, so it ships exactly what it ships today: nothing new.
2. **URL triggers / `recordCrossOriginIframes`: no amplification.** URL triggers activate recording without interaction, but they cannot produce the idle-to-active transition; only user interaction can. Worst case is one extra full snapshot per idle-wake cycle, and a cycle needs 5+ minutes of inactivity followed by interaction.
3. **Oldest event in a flushed buffer: unchanged.** No flush timing or buffering changes. The wake snapshot carries a current timestamp; idle-era events stay dropped.
4. **Volume invariants hold:** zero-interaction tabs still ship zero recordings; the change creates no sessions, no rotations, and no recordings - it adds one event to a recording that is already shipping. Recording *count* metrics should not move at all after release; only bytes per recording can grow, and only for sessions that idle with DOM churn and then resume.

### Residual risk

- The snapshot runs on the main thread at the first interaction after wake. It is gated to sessions where mutations were actually dropped, but on very large DOMs it is a perceptible task. This trades a jank spike for a permanently corrupted replay.
- Post-release: the recordings week-over-week anomaly alert should stay flat. If recording counts move after this ships, that is unexpected and worth investigating immediately.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -2298,6 +2306,10 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
                 this._releaseHoldAndFlush()
             }
             if (returningFromIdle) {
+                if (this._mutationsDroppedWhileIdle > 0) {
```
TueHaulund: This takes a full snapshot even when the session is sampled out, and the buffer gets thrown away later anyway. So on 10% sampling we'd serialize the whole DOM on the main thread every idle wake, for 90% of sessions, right as the user comes back. `_onTriggerActivated` already checks `['sampled', 'active'].includes(this.status)`, same check here?

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -2298,6 +2306,10 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
+                    this._mutationsDroppedWhileIdle = 0
```
TueHaulund: `_tryTakeFullSnapshot()` can return false and queue a retry, and that retry can expire. The counter is already back to 0 by then, so the snapshot never happens and the recording stays broken. Only reset it when the call returns true?

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -1770,6 +1774,9 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
         // When in an idle state we keep recording but don't capture the events,
         // we don't want to return early if idle is 'unknown'
         if (this._isIdle === true && !isAllowedWhenIdle(event)) {
+            if (event.type === EventType.IncrementalSnapshot && event.data.source === IncrementalSource.Mutation) {
```
TueHaulund: Idle drops more than mutations. Stylesheet changes get dropped too, so the replay can come back styled wrong with nothing to heal it. Counting any dropped incremental would cover that. Skipping canvas is right though, a full snapshot can't fix canvas anyway.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -2298,6 +2306,10 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
                 this._releaseHoldAndFlush()
```
TueHaulund, line 2326: `_releaseHoldAndFlush` takes its own full snapshot if the held buffer overflowed, so we could end up taking two in the same tick. Clearing the counter there would avoid the second one.

TueHaulund (commented):
Mechanism looks right. I checked the two things it leans on: the snapshot does land in the buffer before the wake event, and rotation can't mis-fire it. Five notes, the first is the one I'd fix before merge.

packages/browser/playwright/mocked/session-recording/session-recording-idle-timeout.spec.ts
```diff
@@ -298,6 +298,101 @@ test.describe('Session recording - idle timeout behavior', () => {
...
+
+        // user returns: recording must re-sync the player with a fresh full snapshot
+        await page.waitingForNetworkCausedBy({
+            urlPatternsToWaitFor: ['**/ses/*'],
+            action: async () => {
+                await page.locator('[data-cy-input]').type('back from idle')
+            },
+        })
+        // let the post-wake buffer flush ship everything
+        await page.waitForTimeout(2500)
```
TueHaulund: This is the most likely thing to flake on a slow runner, `expect.poll` would be steadier. The 800/200/200 sleeps above should stay, they're crossing the idle threshold on purpose.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
marandaneto: **blocking: Track dropped full snapshots too**\n\nA periodic `FullSnapshot` can itself cross the idle threshold: `_updateWindowAndSessionIds` marks the recorder idle, then this branch drops the snapshot after rrweb has reset its node IDs. Because only `Mutation` increments the counter, wake-up skips the healing snapshot and later incrementals can reference IDs the player never received. Treat a dropped `FullSnapshot` as requiring a healing snapshot, and cover this timer-driven sequence with a regression test.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in 1c064fe34 — the wake heal now requires `['sampled', 'active'].includes(this.status)`, same shape as `_onTriggerActivated`. Covered by a unit test that pins status to 'buffering' and asserts no snapshot.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in 1c064fe34, with a stronger shape than the return-value check: the counter now clears only when a FullSnapshot event actually passes the idle gate in `onRRwebEmit`. A failed or expired take never clears it, so the next wake retries. Unit test covers the throw-once-then-retry sequence.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in 1c064fe34 — the gate now counts any dropped FullSnapshot, Meta, or non-canvas incremental (`isMirrorDesyncingWhenDropped`). Canvas stays excluded. Parameterized test covers mutation / stylesheet / full snapshot (heal) vs canvas / plugin (no heal).

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in 1c064fe34 — falls out of the clear-on-emitted-FullSnapshot design: `_releaseHoldAndFlush` runs before the `returningFromIdle` check and its snapshot emits synchronously, so it clears the counter before this branch reads it. No second snapshot in the same tick.

packages/browser/playwright/mocked/session-recording/session-recording-idle-timeout.spec.ts
ksvat: Done in 1c064fe34 — the fixed 2.5s wait is now an `expect.poll` that waits for the wake marker plus a type-2 snapshot to ship (10s ceiling). The threshold-crossing sleeps are unchanged. Re-verified on Chromium locally.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in 1c064fe34 — the gate now counts a dropped FullSnapshot (and Meta) as requiring a heal, and the parameterized regression test includes exactly this sequence: a full snapshot event crosses the idle threshold, gets declared idle and dropped, and the wake takes a healing snapshot.

One precision on the mechanism: `mirror.reset()` doesn't run in `takeFullSnapshot` (only on record start/stop), so node ids survive the drop. The desync comes from `stylesheetManager.reset()` plus mutations processed while the buffers are locked during serialization — they can reference state only the dropped snapshot describes. Same conclusion, so counting dropped snapshots is the right call either way.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
```diff
@@ -2298,6 +2321,11 @@ export class LazyLoadedSessionRecording implements LazyLoadedSessionRecordingInt
...
             }
             if (returningFromIdle) {
+                // a snapshot from _releaseHoldAndFlush above has already cleared the
+                // counter by the time this check runs, so we never take two in one tick
+                if (this._eventsDroppedWhileIdle > 0 && ['sampled', 'active'].includes(this.status)) {
```
marandaneto: **blocking: Trigger-buffered recordings can still ship stale replay data**\n\n`buffering` can mean a recording trigger is still pending, not just that sampling discarded the session. Skipping the wake heal here leaves post-idle incrementals in the buffer against the pre-idle DOM; when a later trigger activates, `_activateTrigger()` flushes that buffer before any scheduled full snapshot. Include trigger-pending buffering in the heal (or snapshot before activation flush), and cover idle → wake → trigger activation ordering.

marandaneto (approved):
left a last comment, approving to unblock
@PostHog/team-replay might want to double check

arnohillen:

throttleMutations runs before the idle gate, so an attribute-only mutation fully swallowed by the rate limiter while idle never increments _eventsDroppedWhileIdle and a wake after pure throttled-attribute churn skips the heal, leaving stale attributes until the next periodic snapshot.

arnohillen:

session_idle_threshold_ms has no lower clamp, so a low-threshold config on a page with background mutations can take the healing full snapshot on every interaction burst; the only bound on the heal rate is the idle threshold itself.

packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts
ksvat: Done in cbd07c65e — the wake heal now runs when `_strategy?.hasPendingTriggers(this.sessionId)` is true, alongside sampled/active, mirroring the trigger-pending branch in `_fullSnapshotIntervalMillis`. It also composes with the existing `_clearBufferBeforeMostRecentMeta` on full snapshots with pending triggers, so the heal both fixes the stream and truncates the held buffer to post-snapshot state.

Test covers both buffering cases: pending trigger heals (and asserts the snapshot lands in the held buffer before the wake event, which is what activation flush ships), no pending trigger skips.

ksvat:
Replies to the two top-level findings:

**Throttler-swallowed mutations while idle** — real gap, fixed in cbd07c65e. The swallow point (`throttleMutations` returning nothing) sits before the idle gate, so those events never reached the counter. When awake this self-corrects - attribute mutations carry current values, so a later update restores the state once the bucket refills. While idle that later update is dropped too, so a fully-swallowed event now increments the drop counter directly at the swallow point. Unit test covers swallow-while-idle then wake heals.

**No lower clamp on `session_idle_threshold_ms`** - accurate observation, leaving it out of this PR deliberately. The threshold bounds all idle machinery (sessionIdle events, idle flushes), not just this heal, so a clamp is a behavior change for anyone intentionally running a low threshold and belongs in its own change. Two things bound the cost meanwhile: the heal fires at most once per idle-wake cycle, and only when the DOM actually changed while idle - a low threshold on a quiet page heals nothing. Happy to open a follow-up for the clamp if the team wants it.

arnohillen (approved):

Re-reviewed cbd07c65: the trigger-pending heal (bufferCanShip now includes hasPendingTriggers, composing with the pre-existing buffer truncation so activation ships [Meta, FullSnapshot, wake, ...]) and the throttler-swallow counter increment are both correct end-to-end, with tests that pin the heal, the no-trigger skip, the buffer ordering, and the swallow count; CI fully green on the new head.

arnohillen:

Heads-up: #4432 inserts "_eventsCapturedBeforeRecorderStarted" at the same position in terser-mangled-names.json (directly after "_events", both hunks are @@ -188,6 +188,7 @@), so whichever PR merges second gets a one-line conflict there and needs a trivial rebase.

PR #4511, opened 2026-08-12
feat(dead-clicks): share the third-party telemetry host deny list with session replay

pauldambra:
## What

Moves the third-party analytics / error / session-replay vendor host deny list (Sentry, LogRocket, Clarity, GA, Datadog, Segment, Amplitude, Mixpanel, Hotjar, FullStory, …) into a shared `@posthog/browser-common/utils/network-deny-list` module, and imports it from **both** session replay's network capture (`payloadHostDenyList`) and dead-click autocapture — a single source of truth.

Dead-click's network liveness signal now excludes those beacons in addition to PostHog's own ingestion hosts. They fire in the background near clicks and would otherwise stand in as a click's "response", wrongly suppressing genuinely dead clicks. Matching is host-suffix (`endsWith`), mirroring the recorder so vendor subdomains (e.g. `o1.ingest.sentry.io`) match; a cheap substring pre-check keeps the hot path allocation-free. Dead-clicks no longer imports the recorder config module, so its bundle stays lean.

**Why:** direct follow-up to reviewer guidance on #4508 — reuse the recorder's curated deny list rather than only skipping PostHog's own hosts, and keep it in one place so adding a vendor benefits both features.

> Stacked on #4508 (base is its branch); GitHub shows only this commit's diff and it will retarget when #4508 merges. Split out only because the PR-head branch of #4508 is currently locked for direct writes.

## Sanity

- [x] Recorder config + denylist suites still green (identical values after extraction)
- [x] Dead-click + swipe unit suites green (added a Sentry-subdomain exclusion test)
- [x] typecheck / lint / prettier clean; mangled-property-names regenerated

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C03PB072FMJ/p1786270706603799)*


pauldambra:
Folding this into #4508 per maintainer request — the shared `network-deny-list` change is being committed directly onto #4508's branch, so this stacked PR is redundant. (It was also making `posthog/dead-click-network-liveness` a base branch, which blocked committing the fold.)

pauldambra:
Closing — folding into #4508 rather than keeping a stacked PR. The fold commit itself is temporarily blocked this session (the signing tool cached `posthog/dead-click-network-liveness` as a base branch while this PR was based on it); its base has been moved off, so a fresh run can commit the fold directly onto #4508. The change is preserved on this branch (commit edec5490).

PR #4520, opened 2026-08-14
chore(browser): stop publishing dist source maps to npm

### LINKED ISSUE #3205: Remove source maps from distribution
## posthog is blowing up our app size
`posthog-js` is at 33 MB by far the heaviest dependency we bundle, almost 40% larger than Typescript (24 mb).
This seems to be almost exclusively because if the bundled source maps, which aren't useful for the consumer of the library.

An easy way to verify is npm's [code exploration](https://www.npmjs.com/package/posthog-js/v/1.358.1?activeTab=code)

Image

Strip source maps from `/dist` folder before uploading to npm

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

veksa:
## Problem

Closes #3205.

`posthog-js` unpacks to ~41 MB, and 28 MB of that is `dist/*.js.map` — 41 bundles' worth of maps with their sources inlined. Nothing loads them at runtime, but every install pays for them: CI caches, Docker images that ship production `node_modules`, disk on every machine with the SDK in a lockfile.

To answer the question on the issue: no, they don't reach anyone's bundle output. That's exactly why this is cheap to fix and still worth fixing.

## Changes

One line in `files`: `"!dist/*.map"`.

The maps are still built and the CDN still gets them — the `upload-s3` job reads `packages/browser/dist/*.js.map` from its own build, never from the npm tarball. So the `x_google_ignoreList` console attribution from #4299 is untouched for snippet users, and `check-sourcemap-ignore-list.js` still guards it.

Left alone on purpose: the 2.4 MB of `lib/**/*.js.map` (negation in `files` doesn't reach files pulled in by a directory match, so `"!lib/**/*.map"` is a no-op on the pnpm we pin — better fixed via `sourceMap` in the `tsc` config), and `lib/*` itself, which is 6 MB of unreachable output but has real deep importers like `posthog-js/lib/src/constants`.

## Testing

Packed the real `posthog-js@1.416.1` file tree both ways with the pinned pnpm, so the globs ran against actual filenames:

| | before | after |
|---|---|---|
| unpacked | 40.8 MB | 12.4 MB |
| tarball | 10.9 MB | 3.3 MB |
| files | 766 | 725 |

Diffed the file lists: exactly 41 files dropped, all `dist/*.js.map`, nothing else. All 41 bundles, all 154 `.d.ts`, `react/dist/**` and the subpath shims still ship.

The bundles keep their `//# sourceMappingURL` comments. esbuild — what Vite prebundles deps with — resolves a missing map silently. A `source-map-loader` walking `node_modules` is the one setup likely to warn.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [ ] Tests for new code

### If releasing new changes

## 🤖 Agent context

Claude Code. I asked why the package was so heavy; it traced the build and release pipeline and established the split this rests on — the tarball maps are dead weight, the CDN ones are load-bearing.

Two things it tried that I dropped: negating the `lib/` maps (silently a no-op on the pinned pnpm, several spellings checked) and removing `lib/*` outright (breaks deep imports). Numbers above come from packing the real published tree, not estimates.

### DIFF
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -53,6 +53,7 @@
     "files": [
         "lib/*",
         "dist/*",
+        "!dist/*.map",
         "react/dist/**",
         "react/package.json",
         "react/surveys/package.json",
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -17,6 +17,7 @@
         "build": "tsc -b && rollup -c",
         "bundle-size:array": "node scripts/compare-array-bundle-size.mjs",
         "postbuild": "node scripts/strip-lib-package-json.js && node scripts/check-mangled-property-consistency.js && node scripts/check-sourcemap-ignore-list.js",
+        "prepack": "node scripts/strip-sourcemap-sources-content.js",
         "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz",
         "lint": "eslint src && eslint playwright",
         "lint:fix": "eslint src --fix && eslint playwright --fix",
@@ -53,7 +54,6 @@
     "files": [
         "lib/*",
         "dist/*",
-        "!dist/*.map",
         "react/dist/**",
         "react/package.json",
         "react/surveys/package.json",
--- a/packages/browser/scripts/strip-sourcemap-sources-content.js
+++ b/packages/browser/scripts/strip-sourcemap-sources-content.js
@@ -33,44 +33,60 @@ function findMaps(dir) {
     })
 }
 
-const distDir = path.join(PACKAGE_ROOT, 'dist')
-if (!fs.existsSync(distDir)) {
-    console.error('FAIL: dist/ is missing — run `pnpm build` before packing')
-    process.exit(1)
-}
-
-// Every directory in `files` that carries maps. `react/dist` is npm-only (the S3 upload takes
-// `dist/*.js{,.map}` and nothing else), so it's safe to strip here too.
-const maps = [distDir, path.join(PACKAGE_ROOT, 'lib'), path.join(PACKAGE_ROOT, 'react', 'dist')].flatMap(findMaps)
-
-if (maps.length === 0) {
-    console.warn('warn: no source maps found in dist/, lib/ or react/dist/ — nothing to strip')
-    process.exit(0)
-}
-
-let stripped = 0
-let bytesBefore = 0
-let bytesAfter = 0
-
-for (const mapPath of maps) {
-    const raw = fs.readFileSync(mapPath, 'utf8')
-    const map = JSON.parse(raw)
-    bytesBefore += Buffer.byteLength(raw)
-
-    if (!map.sourcesContent) {
-        bytesAfter += Buffer.byteLength(raw)
-        continue
+function stripSourceMapSourcesContent(packageRoot) {
+    const distDir = path.join(packageRoot, 'dist')
+    if (!fs.existsSync(distDir)) {
+        throw new Error('dist/ is missing — run `pnpm build` before packing')
     }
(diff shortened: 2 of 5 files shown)


marandaneto:
@PostHog/team-error-tracking do we need the source maps as part of the npm packages? (see my pr comments, likely side effects)

marandaneto:
thanks @veksa 
checking this out

packages/browser/package.json
```diff
@@ -53,6 +53,7 @@
     "files": [
         "lib/*",
         "dist/*",
+        "!dist/*.map",
```
marandaneto: **blocking: Preserve downstream source-map chaining**

Downstream bundlers consume these maps when generating application source maps. Without `module.js.map`, an esbuild reproduction mapped SDK frames only to minified `dist/module.js` instead of the 118 original sources, degrading customer debugging and error symbolication. Please retain usable maps (for example, without `sourcesContent`) or provide and validate an equivalent source-map chain with error-tracking owners.

marandaneto: **blocking: Avoid dangling `sourceMappingURL` references**

All 41 shipped `dist/*.js` files still contain `//# sourceMappingURL=*.js.map`, so this exclusion leaves references to absent files. Webpack with `source-map-loader` emits an `ENOENT` warning, which can fail builds that promote warnings to errors. Please either publish the maps or strip these comments from npm-packaged JavaScript while leaving CDN artifacts unchanged.

veksa: Fair point. How about keeping the maps but dropping sourcesContent? Still ~4x smaller and the chain stays intact.

veksa: Dropping sourcesContent covers this too - maps stay, so the comments aren't dangling.

ablaszkiewicz:
@marandaneto we don't. Dropping sources context as @veksa suggested sounds reasonable

veksa:
Reworked in e86cee3 - maps stay, only `sourcesContent` goes. Chain intact, no dangling refs, and 40.8 MB -> 17.4 MB unpacked. Stripped at `prepack`, so the CDN artifacts keep their sources.

packages/browser/scripts/strip-sourcemap-sources-content.js
```diff
@@ -0,0 +1,76 @@
...
+    process.exit(1)
+}
+
+// Every directory in `files` that carries maps. `react/dist` is npm-only (the S3 upload takes
+// `dist/*.js{,.map}` and nothing else), so it's safe to strip here too.
+const maps = [distDir, path.join(PACKAGE_ROOT, 'lib'), path.join(PACKAGE_ROOT, 'react', 'dist')].flatMap(findMaps)
+
+if (maps.length === 0) {
+    console.warn('warn: no source maps found in dist/, lib/ or react/dist/ — nothing to strip')
+    process.exit(0)
```
marandaneto: **blocking: Fail when source maps are absent**

If `dist/` exists but a partial or stale build contains no maps, this exits successfully and can publish JavaScript with dangling `sourceMappingURL` references. Please fail `prepack` with a non-zero exit instead.

marandaneto: i pushed a change

marandaneto (approved):

PR #4521, opened 2026-08-14
fix: normalize SDK timestamps to UTC

marandaneto:
## Problem

Capture timestamp overrides are not serialized consistently across the JavaScript SDKs. Offset-bearing timestamps can reach the wire without a canonical UTC representation, which makes equivalent instants look different across browser, Node.js, and React Native capture paths.

## Changes

Valid capture timestamp overrides are now serialized as the equivalent UTC ISO string ending in `Z`. The browser, shared core, Node.js Capture V1, and React Native paths keep caller-defined event properties unchanged. Node.js also normalizes timezone-less strings from the configured local timezone and preserves fractional-second precision beyond milliseconds.

Regression coverage verifies the following behavior:

- Browser capture converts `2023-11-15T03:43:20.000+05:30` to `2023-11-14T22:13:20.000Z` while preserving the same offset-bearing value in a caller property.
- Shared core capture converts `2021-01-02T03:04:05.000+05:30` to `2021-01-01T21:34:05.000Z`.
- Node.js capture converts Date and string offsets to UTC, converts a timezone-less string under `America/Los_Angeles`, and preserves six-digit fractional precision.
- React Native capture converts `2024-01-02T12:04:05.000+09:00` to `2024-01-02T03:04:05.000Z` while preserving caller properties.

Focused Jest coverage is present in the browser wire-envelope test, shared core capture test, Node.js transform and capture tests, and React Native event snapshot test.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [x] @posthog/browser-common

The changeset also includes `@posthog/core` and `@posthog/mcp`, which are changed by the shared timestamp behavior and public timestamp documentation but are not listed above.

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi was used to prepare the existing changes, run focused validation, and run the bundled autoreview. No shareable session link is available. The review identified that parsing an RFC 3339 timestamp through `Date` could truncate fractional seconds beyond milliseconds. The accepted finding was fixed with focused regression coverage before the branch was pushed.

### DIFF
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,12 +1,7 @@
 ---
 'posthog-js': patch
-'@posthog/browser-common': patch
-'@posthog/convex': patch
 '@posthog/core': patch
-'@posthog/mcp': patch
 'posthog-node': patch
-'posthog-react-native': patch
-'@posthog/types': patch
 ---
 
-Normalize capture timestamp overrides to equivalent UTC ISO strings across the browser, Node.js, and React Native SDKs.
+Normalize capture timestamp overrides to equivalent UTC ISO strings in the browser and Node.js SDKs and shared core.
--- a/.changeset/calm-clocks-align.md
+++ b/.changeset/calm-clocks-align.md
@@ -1,0 +1,12 @@
+---
+'posthog-js': patch
+'@posthog/browser-common': patch
+'@posthog/convex': patch
+'@posthog/core': patch
+'@posthog/mcp': patch
+'posthog-node': patch
+'posthog-react-native': patch
+'@posthog/types': patch
+---
+
+Normalize capture timestamp overrides to equivalent UTC ISO strings across the browser, Node.js, and React Native SDKs.
--- a/packages/browser-common/src/client.ts
+++ b/packages/browser-common/src/client.ts
@@ -34,7 +34,7 @@ export interface CapturedEventInfo {
 
 /** Per-call capture overrides, mirroring the client's public capture options. */
 export interface CaptureOptions {
-    /** Override the event timestamp sent to PostHog. */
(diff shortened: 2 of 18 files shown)


arnohillen (approved):

[arno's agent] Verified at 5da15b36: every path converts to the UTC instant via toISOString (V8 fraction truncation checked against the fraction-graft regex), naive strings on capture-v1 interpret as process-local then convert, and tests assert offset inputs equal their UTC instants across browser, core, node, and RN.

arnohillen:

[arno's agent] normalizeMessage only converts Date instances, so an untyped caller passing an offset string timestamp on the default v0/core path still reaches the wire un-normalized, unlike the capture-v1 transform which normalizes strings.

arnohillen:

[arno's agent] Converting the Date in normalizeMessage means on('capture') listeners and persisted queue entries now see timestamp as an ISO string where a caller-passed Date was previously delivered as a Date object.

marandaneto:
Addressed both timestamp concerns in 9256e7102:

- The default v0/core transport now normalizes parseable string timestamp overrides to their equivalent UTC instant at serialization time, preserving sub-millisecond precision.
- Timestamp normalization is deferred until the wire payload is built, so `on("capture")` listeners and queued messages retain a caller-provided `Date` object.

Added regression coverage for both behaviors; core tests, lint, build, and the Capture V1 transform tests pass.

PR #4523, opened 2026-08-14
fix(replay): recover from observer initialization failures

### LINKED ISSUE #4522: Session replay: uncaught TypeError on stop() when observe() returned undefined after a swallowed observer-init error — recorder is then permanently broken for the session
## Bug description

The lazy-loaded session recorder can crash with an uncaught `TypeError: t is not a function`
whenever recording is stopped (forced idle reset, session-id rotation, `stopSessionRecording()`,
etc.), and once it happens the recorder is unrecoverable for the rest of the page lifetime:
every subsequent `stop()`/restart re-throws, and replay data is silently lost until reload.

We see this in production (reported via Sentry). The root cause is an interaction between
`LazyLoadedSessionRecording`'s rrweb `errorHandler` and how the bundled rrweb `record()`
collects teardown handlers — a non-function ends up in rrweb's `handlers` array.

## Root cause walk-through

**1.** `LazyLoadedSessionRecording` registers an `errorHandler` that swallows internal rrweb
errors (`src/extensions/replay/external/lazy-loaded-session-recorder.ts`):

```ts
errorHandler: (error, context) => {
    if (context !== 'rrweb') {
        return false
    }
    if (!this._hasLoggedRecorderCallbackError) {
        this._hasLoggedRecorderCallbackError = true
        logger.error('rrweb internal error - recording will continue but may be incomplete', error)
    }
    return true
},
```

**2.** In the bundled rrweb (`rrweb/record/dist/rrweb-record.js` inside `dist/lazy-recorder.js`),
`callbackWrapper` returns `undefined` when the wrapped function throws and the error handler
returns `true`:

```js
const rrwebWrapped = ((...rest) => {
  try {
    return cb(...rest);
  } catch (error) {
    if (errorHandler && errorHandler(error, context) === true) {
      return;                    // <-- wrapped call now returns undefined
    }
    throw error;
  }
});
```

**3.** `record()` pushes the return value of `observe()` (= `callbackWrapper(initObservers)(...)`)
into its `handlers` array **without checking it's a function** — both for the top document and
for iframes:

```js
const init = () => {
  takeFullSnapshot();
  handlers.push(observe(document));          // <-- can be undefined
  ...
};

const loadListener = (iframeEl) => {
  try {
    const iframeId = mirror.getId(iframeEl);
    const cleanup = observe(iframeEl.contentDocument);
    handlers.push(cleanup);                  // <-- can be undefined
    ...
```

So if *anything* throws during `initObservers` (e.g. one of the later observer inits, after the
core observers are already attached), the error is swallowed by design, recording continues and
looks healthy — but `handlers` now contains `undefined`.

**4.** On stop, `callSafely` only tolerates `SecurityError` and calls the handler unconditionally:

```js
return () => {
  ...
  handlers.forEach((h) => callSafely(h));   // callSafely(undefined)
  ...
};

function callSafely(fn) {
  try {
    fn();                                    // TypeError: fn is not a function
  } catch (error) {
    if (!(error instanceof DOMException && error.name === "SecurityError")) {
      throw error;
    }
  }
}
```

The `TypeError` escapes `_stopRrweb()` → `_stopRecordingProducers()` → `stop()` uncaught.

**5.** Because the throw aborts the stop function mid-way, none of the remaining teardown runs
(`processedNodeManager.destroy()`, `iframeManager.destroy()`, `mirror.reset()`, …) and the
`undefined` stays in `handlers`. Every later `stop()` — idle reset, session rotation
(`_restartForSessionIdChange`), manual stop — throws again. In our Sentry data we see the same
session crash first on the forced-idle-reset path and again two minutes later on the
session-id-rotation path (via the heatmaps interval → `checkAndGetSessionAndWindowId`).

## Observed stack (Sentry, production)

```
TypeError: t is not a function
  at ue (callSafely)                          rrweb-record.js (bundled in lazy-recorder)
  at Array.forEach → handlers.forEach((h) => callSafely(h))
  at Wt (record stop fn)
  at LazyLoadedSessionRecording._stopRecordingProducers   this._stopRrweb?.()
  at LazyLoadedSessionRecording.stop
  at this._onSessionIdleResetForcedListener               this.stop()
```

## Suggested fix

Either (ideally both):
- In `record()`: only push cleanups that are functions (`isFunction(cleanup) && handlers.push(cleanup)`)
  in `init()` and in the iframe `loadListener`.
- In `callSafely`: guard with `typeof fn === 'function'` before invoking.

That would let teardown complete and keep the recorder restartable even when an observer init
failed and was intentionally swallowed.

## Minimal reproduction of the invariant

```js
// what happens internally when any observer init throws while an errorHandler is registered:
const wrapped = callbackWrapper(() => { throw new Error('observer init failed') });
handlers.push(wrapped());                    // pushes undefined — no warning
// later:
handlers.forEach((h) => callSafely(h));      // TypeError: h is not a function
```

## Environment

- posthog-js `1.415.7` (npm), recorder script `lazy-recorder` served from PostHog CDN;
  verified the same code is present in `1.417.0` `dist/lazy-recorder.js`
- Init options: `session_recording: { recordCrossOriginIframes: true }`
- Seen on Chrome 150 / Windows, but the bug is browser-independent

marandaneto:
## Problem

When an rrweb observer throws during initialization, the session recorder's error handler can intentionally contain the error. The wrapped initialization then returns `undefined`, which rrweb stores as a teardown handler. Stopping recording later calls that value and throws before teardown finishes. Earlier observers remain attached, and session rotation or a manual restart can leave the recorder broken or produce duplicate events.

Fixes #4522.

## Changes

- Clean up observers that initialized successfully before an initialization error is passed to the configured error handler.
- Only store teardown handlers that are functions for the main document and iframe paths.
- Add real-browser coverage for main-document and iframe initialization failures, listener cleanup, and recorder restarts.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi was used to reproduce the failure, implement the fix, and run the test and build suites. The autoreview skill reviewed commit `29d447590` and reported no actionable findings. The implementation cleans observers created before the failure and enforces the teardown-handler invariant at the two insertion sites, which prevents leaked listeners after a restart.

### DIFF
--- a/.changeset/calm-owls-restart.md
+++ b/.changeset/calm-owls-restart.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Prevent swallowed rrweb observer initialization errors from breaking session replay teardown and subsequent recorder restarts.
--- a/packages/rrweb/rrweb/src/record/index.ts
+++ b/packages/rrweb/rrweb/src/record/index.ts
@@ -1009,15 +1009,17 @@ function record<T = eventWithTime>(
       try {
         const iframeId = mirror.getId(iframeEl);
         const cleanup = observe(iframeEl.contentDocument!);
-        handlers.push(cleanup);
-        // Accumulate cleanups across iframe navigations.
-        if (iframeId !== -1) {
-          let bucket = iframeObserverCleanups.get(iframeId);
-          if (!bucket) {
-            bucket = new Set();
-            iframeObserverCleanups.set(iframeId, bucket);
+        if (typeof cleanup === 'function') {
+          handlers.push(cleanup);
+          // Accumulate cleanups across iframe navigations.
+          if (iframeId !== -1) {
+            let bucket = iframeObserverCleanups.get(iframeId);
+            if (!bucket) {
+              bucket = new Set();
+              iframeObserverCleanups.set(iframeId, bucket);
+            }
+            bucket.add(cleanup);
           }
-          bucket.add(cleanup);
         }
       } catch (error) {
         // TODO: handle internal error
@@ -1072,7 +1074,8 @@ function record<T = eventWithTime>(
 
     const init = () => {
       takeFullSnapshot();
-      handlers.push(observe(document));
+      const cleanup = observe(document);
+      if (typeof cleanup === 'function') handlers.push(cleanup);
       handlers.push(on('fullscreenchange', emitFullscreenChange));
       handlers.push(on('webkitfullscreenchange', emitFullscreenChange));
       handlers.push(on('mozfullscreenchange', emitFullscreenChange));
--- a/packages/rrweb/rrweb/src/record/observer.ts
+++ b/packages/rrweb/rrweb/src/record/observer.ts
@@ -83,31 +83,43 @@ export function initMutationObserver(
   rootEl: Node,
 ): { observer: MutationObserver; buffer: MutationBuffer } {
   const mutationBuffer = new MutationBuffer();
-  mutationBuffers.push(mutationBuffer);
   // see mutation.ts for details
   mutationBuffer.init(options);
-  const observer = new (mutationObserverCtor() as new (
-    callback: MutationCallback,
-  ) => MutationObserver)(
-    callbackWrapper(mutationBuffer.processMutations.bind(mutationBuffer)),
-  );
-  const mutationObserverInit: MutationObserverInit = {
-    attributes: true,
-    attributeOldValue: true,
-    characterData: true,
-    characterDataOldValue: true,
-    childList: true,
-    subtree: true,
-  };
-  // Delegate attribute filtering to the native MutationObserver: unlisted
-  // attributes never fire the callback, so they cost no recording CPU.
-  // An empty array would mean "observe no attributes at all", which is never
-  // what a caller wants and could come from bad config, so treat it as unset.
-  if (options.attributeFilter && options.attributeFilter.length > 0) {
-    mutationObserverInit.attributeFilter = options.attributeFilter;
+  let observer: MutationObserver | undefined;
+  try {
+    observer = new (mutationObserverCtor() as new (
+      callback: MutationCallback,
+    ) => MutationObserver)(
+      callbackWrapper(mutationBuffer.processMutations.bind(mutationBuffer)),
+    );
+    const mutationObserverInit: MutationObserverInit = {
+      attributes: true,
+      attributeOldValue: true,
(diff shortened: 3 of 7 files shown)


arnohillen (approved):

[arno's agent] Verified at 29d44759: the catch-path cleanup (including the MutationBuffer splice) closes the leaked-observer/double-emission bug from #4522, both storage sites guard non-function returns, and the 4-engine tests prove single emission after restart and clean iframe detach.

arnohillen:

[arno's agent] After this change a single failing observer init tears down every observer for the document, so the session records only periodic full snapshots until the next restart, whereas before the observers initialized ahead of the failure kept emitting; worth confirming the all-or-nothing tradeoff is intended.

arnohillen:

[arno's agent] If initMutationObserver throws after mutationBuffers.push (observer.ts line 86), the local mutationBuffer in initObservers is never assigned, so the catch-path cleanup cannot splice the orphaned buffer out of the global array and it survives restarts.

arnohillen:

[arno's agent] Observer init failures still surface only as a single console logger.error on the customer'\''s page, so a document that lost all observers this way is invisible in PostHog telemetry and unmarked in the recording itself.

marandaneto:
@arnohillen Thanks for the review — addressing the three follow-ups:

1. The all-or-nothing teardown is intentional. Once initialization fails, rrweb cannot return a valid aggregate teardown for the partially initialized set. Leaving those observers active would leak them and cause duplicate emissions after a restart, so retaining periodic full snapshots is safer than retaining unmanaged incremental observers.
2. Good catch on the mutation buffer. Fixed in ef831de51: `initMutationObserver()` now publishes the buffer only after native observation succeeds and releases the acquired resources on failure. The new regression test verifies that a failed setup leaves `mutationBuffers` empty.
3. Agreed that recorder failure telemetry/recording markers would improve observability. That behavior predates this fix and requires a separate decision about the telemetry/event contract, so I have kept it out of this targeted teardown/restart patch.

Issue #4525, opened 2026-08-14
perf(react-native): useFeatureFlags / useFeatureFlag re-read the full flag store on every render


crhallen:
## Bug description

`useFeatureFlags` and `useFeatureFlag` in `posthog-react-native` initialize state with an **eager** `useState(...)` argument:

https://github.com/PostHog/posthog-js/blob/master/packages/react-native/src/hooks/useFeatureFlags.ts#L9-L11

```ts
const [featureFlags, setFeatureFlags] = useState<PostHogFlagsResponse['featureFlags'] | undefined>(
  posthog?.getFeatureFlags()
)
```

https://github.com/PostHog/posthog-js/blob/master/packages/react-native/src/hooks/useFeatureFlag.ts#L9

```ts
const [featureFlag, setFeatureFlag] = useState(posthog?.getFeatureFlag(flag))
```

React only *uses* that initial value on mount, but it still **evaluates the argument on every render**. After mount, that work is thrown away.

`getFeatureFlags()` / `getFeatureFlag()` are not cheap. They go through `@posthog/core` `getFeatureFlagDetails()` → `normalizeFlagsResponse()`, which rebuilds the entire flag map (`Object.entries` → `map` → `filter` → `Object.fromEntries`, plus `JSON.parse` on payloads). That result is not memoized.

So every re-render of a component that calls these hooks pays a full store renormalize, even when flags have not changed and `onFeatureFlags` has not fired.

`@posthog/react` already avoids this. For example `useFeatureFlagEnabled` uses a lazy initializer:

```ts
const [featureEnabled, setFeatureEnabled] = useState(() => client.isFeatureEnabled(flag))
```

RN should do the same. `useEffect` / `onFeatureFlags` can stay as-is so reload / reset / identity behaviour does not change.

## How to reproduce

1. Create a PostHog RN client and a child that calls `useFeatureFlags(posthog)` (or `useFeatureFlag('some-flag', posthog)`).
2. Spy or breakpoint `PostHog.prototype.getFeatureFlags` / `getFeatureFlag`.
3. Re-render the **parent** (state bump, no flag reload).

Each parent re-render calls `getFeatureFlag(s)` again. After changing the initializer to `useState(() => posthog?.getFeatureFlags())` (and the same for `getFeatureFlag`), parent re-renders do not call it.

Minimal sketch:

```tsx
import { useState } from 'react'
import { Button, Text, View } from 'react-native'
[...]

PR #4528, opened 2026-08-14
fix(react-native): lazily initialize feature flag result

luke-belton:
## Problem

The merged fix in #4526 made `useFeatureFlags` and `useFeatureFlag` use lazy state initializers. A broader audit found the sibling `useFeatureFlagResult` hook still eagerly calls `getFeatureFlagResult(flag)` on every render, repeatedly normalizing cached flag details even though React discards the value after mount.

Follow-up to #4526 and #4525. Supersedes #4527, whose branch was rewritten after #4526 merged.

## Changes

- Lazily initialize `useFeatureFlagResult`
- Add a regression test proving unrelated rerenders do not reread the flag result
- Add a patch changeset for `posthog-react-native`

## Validation

- 13 focused hook tests pass
- Full `posthog-react-native` lint passes
- Dependency-aware `posthog-react-native` Turbo build passes

## Agent context

This PR is based on merged `main` and intentionally contains only the remaining sibling hook discovered in the audit.

---

### DIFF
--- a/.changeset/lazy-rn-feature-flag-hooks.md
+++ b/.changeset/lazy-rn-feature-flag-hooks.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+Avoid rereading cached feature flag results on unrelated React Native component rerenders.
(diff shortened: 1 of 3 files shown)


marandaneto:
oh that makes sense, thanks @luke-belton

marandaneto (approved):

luke-belton:
@marandaneto sorry for the confusing order of PRs - posthog was working in the background and got its wired crossed and you were much quicker getting things merged/closed! 😓

PR #4529, opened 2026-08-14
feat(react-native): add Android screenshot mask alignment config

dustinbyrne:
## Problem

PostHog Android PR [#676](https://github.com/PostHog/posthog-android/pull/676) adds an opt-in screenshot mask-alignment verification mode, but React Native has no supported way to configure it. Android React Native apps therefore remain on the legacy redraw guard even after the native capability is available.

This is Android-specific. iOS behavior is unchanged.

## Changes

- Add `sessionReplayConfig.verifyScreenshotMaskAlignment`, documented as Android-only, disabled by default, and requiring `@posthog/react-native-plugin`.
- Forward the option through both React Native native-plugin initialization paths.
- Apply it to `PostHogSessionReplayConfig` in `@posthog/react-native-plugin` with a safe `false` fallback.
- Use the current PostHog Android 3.60.7 dependency, which includes Android PR #676.
- Add JS envelope coverage and Android bridge tests, and run the Android plugin tests in native CI.
- Add minor changesets for `posthog-react-native` and `@posthog/react-native-plugin`.

Companion Flutter PR: [posthog-flutter#533](https://github.com/PostHog/posthog-flutter/pull/533).

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Validation

- Focused React Native Jest tests and snapshots.
- React Native and native-plugin lint/build.
- Android plugin unit tests through the native example Gradle project.
- Generated API reference reproducibility and npm package contents.
- Changeset status, rebase topology, and `git diff --check`.

## 🤖 Agent context

Implemented with Pi worker agents and independently reviewed with Pi's reviewer. The human selected the cross-SDK scope and the public option name. The implementation keeps the option explicitly Android-only and leaves iOS unchanged.


dustinbyrne:
### Replay incident-risk review

**Matched class:** Release and delivery channel integrity (Class 7)

**Same failure mode:** No.

- The workflow remains a `pull_request` workflow; this change does not introduce `pull_request_target` or alter checkout behavior.
- It only adds `:posthog-react-native-plugin:testDebugUnitTest` to the existing Android build command so the new native regression tests execute in CI.
- It does not change permissions, credentials, trusted publishing, package lifecycle scripts, release jobs, CDN paths, CORS, or cache behavior.
- `.github` remains covered by the repository's existing CODEOWNERS policy.

No additional incident-specific code or browser test is required for this workflow-only test invocation.

.changeset/quick-dingos-align.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/react-native-plugin': patch
+'posthog-react-native': patch
```
turnipdabeets: [question] The equivalent iOS-only experimental option shipped as a minor (`posthog-react-native` 4.45.0, #3552) — was patch deliberate here? A new public config field reads as minor to me.

```suggestion
'@posthog/react-native-plugin': minor
'posthog-react-native': minor
```

.changeset/quick-dingos-align.md
```diff
@@ -0,0 +1,6 @@
...
+---
+
+Add an Android session replay option to verify screenshot mask alignment during capture.
```
turnipdabeets: [nit] Could we name the option and the floor so this is actionable straight from the changelog — `sessionReplayConfig.verifyScreenshotMaskAlignment`, Android only, requires `@posthog/react-native-plugin`?

packages/react-native/src/types.ts
```diff
@@ -185,6 +185,17 @@ export type PostHogSessionReplayConfig = {
    * @default true
    */
   captureNetworkTelemetry?: boolean
+  /**
+   * Verify that masks remain aligned while capturing session replay screenshots.
+   * Android only.
```
turnipdabeets: [suggestion] The native property is still `@PostHogExperimental` in the 3.58.3 we pin, and the option is inert on `posthog-react-native-session-replay` (it pins posthog-android 3.40.2 and reads a fixed key list). `screenshotModeBackgroundCapture` below already carries "Experimental support" — worth matching:

```suggestion
   * Verify that masks remain aligned while capturing session replay screenshots.
   * Android only. Experimental. Requires `@posthog/react-native-plugin`.
```

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
```diff
@@ -158,6 +159,7 @@ class PosthogReactNativePluginModule(
               sessionReplayConfig.maskAllImages = maskAllImages
               sessionReplayConfig.maskAllTextInputs = maskAllTextInputs
               sessionReplayConfig.sampleRate = getDoubleOrNull(sdkReplayConfig, "sampleRate")
+              applyVerifyScreenshotMaskAlignment(sessionReplayConfig, sdkReplayConfig)
```
turnipdabeets: [suggestion] `getBoolean(map, key, default)` already does the same `runCatching` + `hasKey` + default dance the six lines above use, and JS always sends a concrete boolean so the `isNull` check can't fire:

```suggestion
              sessionReplayConfig.verifyScreenshotMaskAlignment =
                getBoolean(sdkReplayConfig, "verifyScreenshotMaskAlignment", false)
```

That makes the top-level helper redundant — two readers of the same shape will drift once one of them gets a fix.

packages/react-native-plugin/android/build.gradle
```diff
@@ -85,6 +85,8 @@ dependencies {
   //noinspection GradleDynamicVersion
   implementation "com.facebook.react:react-native:+"
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
-  implementation "com.posthog:posthog-android:3.58.0"
+  implementation "com.posthog:posthog-android:3.58.3"
+
+  testImplementation "junit:junit:4.13.2"
```
turnipdabeets, line 90: [nit] `files` in package.json includes `android` and excludes only `__tests__`, so `android/src/test/**` ships to npm and any consumer running `./gradlew test` compiles it. An `"!android/src/test"` entry would keep it out.

.changeset/quick-dingos-align.md
dustinbyrne: Updated the changeset to name `sessionReplayConfig.verifyScreenshotMaskAlignment`, call out the Android scope, and state that it requires `@posthog/react-native-plugin`.

packages/react-native-plugin/android/src/main/java/com/posthogreactnativeplugin/PosthogReactNativePluginModule.kt
dustinbyrne: Updated this to use the existing `getBoolean` reader and removed the duplicate parsing logic. The shared reader is now module-internal so the focused native bridge tests can still cover true and omitted/default-false values.

.changeset/quick-dingos-align.md
dustinbyrne: Agreed — switched both packages to minor, matching the prior public session replay config addition in #3552.

packages/react-native/src/types.ts
dustinbyrne: The React Native API is intentionally shipping without the experimental label. I added `Requires @posthog/react-native-plugin` so the supported integration boundary is explicit and users do not expect the legacy session replay plugin to consume it.

packages/react-native-plugin/android/build.gradle
dustinbyrne: Added `!android/src/test` to the package files list. Verified with `npm pack --dry-run --json --ignore-scripts` that no Android test files are included.

marandaneto (approved):

PR #4531, opened 2026-08-14
refactor(browser): migrate surveys to shared extension lifecycle

dustinbyrne:
## Problem

Surveys currently manage initialization, request state, listeners, trigger definitions, and cleanup independently of the shared browser extension lifecycle. This makes deferred initialization and disposal harder to reason about, and refreshed survey definitions can leave stale trigger state or timers behind.

## Changes

- Add a browser-specific surveys adapter and migrate surveys to the shared extension lifecycle.
- Preserve the callback-based `getSurveys` API, synchronous cache hits, stale-while-revalidate behavior, request deduplication, failure backoff, and pre-extension-setup access.
- Use the request promise directly as the in-flight survey request while adapting the legacy browser transport at the boundary.
- Replace event, cancellation, and action trigger definitions on refresh, including refresh-to-empty cases.
- Dispose survey listeners and popup timers cleanly and guard against late asynchronous responses.
- Add lifecycle, deferred initialization, request, trigger replacement, timer cleanup, and compatibility coverage.

### Validation

- Browser Jest suite: 5,439 tests passed.
- Browser-common Jest suite: 81 tests passed.
- ESLint, TypeScript, persistence-key policy, `git diff --check`, and the full `posthog-js` build passed.
- Post-rebase focused suite: 141 tests passed.
- Fast `array.js` comparison against the autocapture lifecycle branch: +1.17% minified, +0.90% gzip, and +0.74% Brotli.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent and validated with Jest, ESLint, TypeScript, the repository build, and the array bundle-size comparison. The implementation was kept compatible with the existing callback API and lazy survey bundle boundary; the browser-only transport adaptation was chosen instead of expanding the shared client API. A fresh-context review identified dropped-request and Promise rejection hazards, which were fixed before submission.


dustinbyrne:
Recreating this PR from a PostHog repository branch so the stack does not use fork-based heads.

PR #4533, opened 2026-08-14
refactor(browser): migrate logs to shared extension lifecycle

dustinbyrne:
## Problem

Browser logs still use the browser-v1 extension lifecycle, including direct remote-config dispatch and lifecycle resources that are not owned by the shared extension runtime. This leaves logs inconsistent with the autocapture and surveys lifecycle migrations.

This draft is stacked on #4592; review only this PR's diff.

## Changes

- Enroll built-in logs through the shared `Extension`/`Client` lifecycle while preserving the PostHog constructor and `initialize()` contract for custom `__extensionClasses.logs` implementations.
- Subscribe to remote config through `BrowserClientAdapter` and prevent duplicate lazy-script loads during synchronous config replay.
- Dispose remote-config and reconnect listeners, core flush timers, late deferred setup, and console wrappers while retaining queued records for the shutdown `sendBeacon` flush.
- Pass the shared `Client` from current cores to lazy logs bundles, expose live `Client.canCapture`, and resolve the installed logs extension through typed stable-name lookup. Historical core-backed loaders continue passing `PostHog`; the independently built bundle handles them with an isolated compatibility shim for their consistently emitted `le` console-capture method, while current cores use the stable `captureConsoleLog` ABI.
- Add focused lifecycle, cross-version compatibility, deferred-init, lazy-load, and console cleanup coverage.

Validation:
- 127 browser Jest suites passed (5,450 tests passed, 8 skipped; 40 snapshots)
- Follow-up focused suites passed: 82 browser-common tests and 221 browser tests.
- End-to-end logs version-skew smoke tests passed with the current core and NPM `posthog-js@1.392.0` `array.js`, the first release using core-backed console capture; each test loaded the current local `logs.js` and asserted the expected current or historical console-capture method received the record.
- Browser TypeScript, ESLint, Prettier, `git diff --check`, production build, declaration bundling, property-mangling checks, and source-map checks passed

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with Pi and reviewed by fresh-context builtin reviewer agents. The human supplied the autocapture and surveys lifecycle precedents and requested a narrowly scoped logs migration. We discussed lifecycle compatibility and cleanup edge cases during the work; the human has been good to the agent.


packages/browser/src/entrypoints/logs.ts
```diff
@@ -350,13 +353,48 @@ const originalConsoleMethod = (method: any): any => {
...
+
+    // `_captureConsoleLog` was introduced in 1.392.0 and mangled to `le` in every
+    // published core-backed release before `captureConsoleLog` became a stable ABI.
+    // Those cores pass PostHog here because their lazy loader predates the shared Client boundary.
+    // Keep the historical generated name isolated to this compatibility shim; using
+    // `captureLog` instead would change the service name, scope, queue, and rate limits.
+    const historicalCaptureConsoleLog = (logs as unknown as HistoricalLogs).le
+    if (isFunction(historicalCaptureConsoleLog)) {
+        historicalCaptureConsoleLog.call(logs, options)
+    }
```
turnipdabeets: Moving console capture onto a stable unmangled name is the right fix — but `_captureConsoleLog`'s generated name isn't `le` across every published release. Pulling `dist/array.js` from unpkg and matching the method whose body is `this.<x>().captureLog(t)`: it's `le` in 1.392.0–1.410.10, `he` in 1.411.0–1.418.4, and `ui` in 1.418.5–1.418.10 (current npm `latest`). So from 1.411.0 on, `logs.le` is undefined and the record is silently dropped — confirmed with a failing test that drives the wrapper against each name: `le` captures 1, `he` and `ui` capture 0. That's narrower than the previous revision, which reached those hosts through `captureLog`. I think we need a fallback so the record still lands somewhere:

```suggestion
    const historicalCaptureConsoleLog = (logs as unknown as HistoricalLogs).le
    if (isFunction(historicalCaptureConsoleLog)) {
        historicalCaptureConsoleLog.call(logs, options)
        return
    }
    // The generated name changed after 1.410.10 (`he`, then `ui`), so on most published
    // hosts the probe above misses. Fall back to the stable public API rather than dropping
    // the record — service name, scope, queue and rate limits differ, but the log lands.
    logs.captureLog(options)
```

(Probing `he`/`ui` by name instead looked tempting, but those are just generated identifiers — on another release they may belong to a different method on `logs`, so a name table risks calling the wrong thing.)

packages/browser/src/entrypoints/logs.ts
```diff
@@ -371,7 +409,9 @@ const initializeLogs = (posthog: PostHog) => {
                 // reopen the capture path while the outer invocation is still running.
                 let acquiredGuard = false
                 try {
-                    if (args.length > 0 && !isCapturingLog && posthog.is_capturing()) {
+                    const activeHost = currentHost
+                    const logs = activeHost ? getCapturingLogs(activeHost) : undefined
```
turnipdabeets: [nit] `getCapturingLogs` runs `is_capturing()` plus a registry lookup before we check `args.length > 0 && !isCapturingLog`, so it also runs on zero-arg console calls and on re-entrant frames we're about to skip. Moving it inside the `if` restores the short-circuit we had before.

packages/browser/src/posthog-logs.ts
```diff
@@ -160,12 +170,50 @@ export class PostHogLogs implements Extension {
...
+        const subscription = client.onRemoteConfig((result) => {
+            replayedEnabledConfig = result.ok && result.config.logs?.captureConsoleLogs === true
+            this.onRemoteConfig(result)
+        })
+        if (this._disposed) {
+            subscription.dispose()
+            return
+        }
+        this._remoteConfigSubscription = subscription
+        if (!replayedEnabledConfig) {
```
turnipdabeets, line 188: [suggestion] `posthog-surveys.ts` calls `loadIfEnabled()` unconditionally right after subscribing. The flag here only suppresses a retry when `loadExternalDependency` fails *synchronously* during the replay — the one case `_isLoading` can't cover, since the callback already cleared it. Worth a one-line comment saying that, otherwise the next reader will wonder why logs needs a flag surveys doesn't.

packages/browser-common/src/client.ts
```diff
@@ -106,6 +106,8 @@ export interface Client {
     readonly groups: DeepReadonly<Record<string, string>>
     /** The current session, created on first read if needed. */
     readonly session: SessionContext
+    /** Whether the host currently permits data capture. */
+    readonly canCapture: boolean
```
turnipdabeets, line 110: [question] `canCapture` is required rather than optional, so anyone implementing `Client` outside this repo gets a compile error on upgrade (extensions that only consume it are fine). The package is 0.5.0 and described as internal, so `patch` is probably right — just want to confirm that's deliberate rather than something to soften with `canCapture?` for a release.

.changeset/browser-logs-extension-lifecycle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/browser-common': patch
+'posthog-js': patch
+---
+
+Migrate browser logs to the shared extension lifecycle and expose host capture permission to extensions.
```
turnipdabeets: [suggestion] This is all mechanism — the outcome someone scanning the changelog cares about is that console autocapture starts working again on pages whose loaded bundle predates this release.

```suggestion
Fix console log autocapture silently failing on pages whose loaded `posthog-js` bundle predates this release, and expose host capture permission to extensions.
```

packages/browser/src/entrypoints/logs.ts
dustinbyrne: funny i stopped checking at 1.410.0. i'm just going to roll them all in. i've fixed the name mangling in this PR so it won't be an issue moving forward.

packages/browser/src/entrypoints/logs.ts
```diff
@@ -413,8 +436,18 @@ const initializeLogs = (posthog: PostHog) => {
         // Flatten an existing marker if another console plugin already wrapped this method.
         ;(wrapped as any).__rrweb_original__ = originalConsoleMethod(originalConsoleLog)
         assignableWindow.console[level] = wrapped
+        restoreConsoleMethods.push(() => {
+            if (assignableWindow.console[level] === wrapped) {
+                assignableWindow.console[level] = originalConsoleLog
+            }
```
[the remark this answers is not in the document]
dustinbyrne: this is also the behavior on main - i've opted not to change it yet

packages/browser/src/entrypoints/logs.ts
```diff
@@ -350,13 +353,74 @@ const originalConsoleMethod = (method: any): any => {
...
+
+    const minor = Number(match[1])
+    const patch = Number(match[2])
+    if (minor < 392 || minor > 418) {
+        return undefined
+    }
+    if (minor === 410) {
+        return patch <= 4 ? 'le' : patch <= 10 ? 'de' : undefined
+    }
+    if (minor === 418) {
```
turnipdabeets: The table is right for the range it covers — I checked it against all 118 published `dist/array.js` in [1.392.0, 1.418.10] and every version routes to the name it actually generated.

But it's already stale. 1.418.11 and 1.418.12 shipped on 2026-08-24 and the name churned again:

```js
// 1.418.10  ui(t){this.vv().captureLog(t)}
// 1.418.11  ci(t){this.gv().captureLog(t)}
```

`historicalCaptureConsoleLogName` returns `undefined` above 1.418.10, so both drop every console log. Driving the shipped `initializeLogs` against a host declaring each version, carrying the name it really generates:

```
✓ 1.418.10 (ui) captured
✕ 1.418.11 (ci) → 0 calls
✕ 1.418.12 (ci) → 0 calls
```

1.418.12 is npm `latest`, so right now the fix misses the core it most needs to reach. That's five names in 27 minor versions on boundaries with no pattern to them — could we refresh the table and add a guard, rather than catch it by hand each release?

```ts
it('covers the version this bundle ships as', () => {
    expect(historicalCaptureConsoleLogName(Config.LIB_VERSION)).toBeDefined()
})
```

The `1.418.11` and `1.419.0` cases in `logs.test.ts` assert the drop today, so they'd move with it.

packages/browser/playwright/mocked/logs-version-skew-smoke.spec.ts
```diff
@@ -0,0 +1,75 @@
...
+        return 'de'
+    }
+    if ((minor >= 411 && minor < 418) || (minor === 418 && patch <= 3)) {
+        return 'he'
+    }
+    if (minor === 418 && patch <= 10) {
+        return 'ui'
+    }
+    return undefined
+}
```
turnipdabeets: [suggestion] Second copy of the table, and the two already disagree: this regex has no `$` anchor and doesn't pin the major, so `1.410.5-canary` resolves to `de` and `1.418.10-invalid` to `ui` here, while production returns `undefined` for both. Those are the exact strings `logs.test.ts` asserts get dropped. Could we export `historicalCaptureConsoleLogName` from a shared module and import it here, so there's one table?

packages/browser/src/entrypoints/logs.ts
```diff
@@ -350,13 +353,74 @@ const originalConsoleMethod = (method: any): any => {
...
+        return patch <= 3 ? 'he' : patch <= 10 ? 'ui' : undefined
```
ioannisj: I'd want this fixed before merge: the `minor === 418` branch stops at `ui`, but there are six generated names in the wild, not four.

I probed the published bundles from both sides, the core's `dist/array.js` (the method whose body calls `.captureLog`) and the lazy `dist/logs.js` (the `.logs)||s.<name>(` call site), and they agree:

- `le` 1.392.0 to 1.410.4
- `de` 1.410.5 to 1.410.10
- `he` 1.411.0 to 1.418.3
- `ui` 1.418.4 to 1.418.10
- `ci` 1.418.11 to 1.418.14
- `vi` 1.418.15 to 1.418.16

1.418.16 is current npm `latest`, so on any core from 1.418.11 on the lookup returns `undefined`, `captureConsoleLogForHost` returns without capturing, and the record is dropped silently. That's the failure the changeset says it fixes, on the versions carrying most of the live traffic.

`patch <= 3 ? 'he' : patch <= 10 ? 'ui' : patch <= 14 ? 'ci' : 'vi'` covers it, though `vi` is current so the last boundary wants re-deriving against npm right before this merges. Same change in `historicalCaptureMethod` in the playwright helper, plus `['1.418.11','ci']` and `['1.418.15','vi']` in the jest table and dropping `'1.418.11'` from the unsupported list.

The part that worries me more than the missing rows: nothing pins this table to the published bundles. The jest table and the playwright helper both hand-encode the same assumption, so they agree with each other while disagreeing with npm, and a regression here would still pass. Worth deriving it from a committed fixture that a script regenerates?

packages/browser/src/posthog-core.ts
```diff
@@ -972,7 +972,13 @@ export class PostHog implements PostHogInterface {
 
     private _enrollExtension(extension: Extension | BrowserCommonExtension, initTasks: Array<() => void>): void {
         if (this._isSharedExtension(extension)) {
-            initTasks.push(() => void this._getBrowserClientAdapter().add(extension).catch(__NOOP))
+            initTasks.push(
+                () =>
+                    void this._getBrowserClientAdapter()
+                        .add(extension)
+                        .catch(() => extension.dispose?.())
```
ioannisj, line 979: [suggestion] If that `dispose()` throws, the trailing `.catch(__NOOP)` swallows it with nothing logged, unlike `ExtensionRuntime._disposeExtension` which try/catches and logs every dispose failure.

packages/browser/src/posthog-logs.ts
```diff
@@ -187,14 +231,18 @@ export class PostHogLogs implements Extension {
...
     captureLog(options: CaptureLogOptions): void {
-        this._getCore().captureLog(options)
+        if (!this._disposed) {
+            this._getCore().captureLog(options)
+        }
     }
 
     // Console auto-capture (the lazy `logs` chunk) routes here so its records run
     // through the shared core pipeline and carry `service.name: posthog-browser-logs`.
     /** @internal */
```
ioannisj, line 241: [suggestion] should this go now?

packages/browser/src/entrypoints/logs.ts
ioannisj: This should probably be updated to latest release 1.418.16?

ioannisj (approved):

packages/browser/src/entrypoints/logs.ts
dustinbyrne: 😭

dustinbyrne: Updated in e1a0033e5 through current main (1.419.2): `ci` covers 1.418.11–1.418.14 and `vi` covers 1.418.15–1.419.2. Unit boundary cases and the version-skew Playwright observer/helper were updated; the compatibility smoke test passes against npm latest 1.419.0.

packages/browser/src/posthog-logs.ts
dustinbyrne: Keeping this `@internal`. `captureConsoleLog` is a stable ABI between our independently built core and lazy logs bundles, but it is not intended to become public SDK API.

packages/browser/src/posthog-core.ts
dustinbyrne: Fixed in e1a0033e5. Cleanup failures now use the same `Failed to dispose browser extension` logging as `ExtensionRuntime`, with focused coverage for a synchronous disposal error after enrollment rejects.

PR #4536, opened 2026-08-16
fix(core): detect LinkedIn, Twitter, TikTok, WeChat, Line in-app browsers; narrow Pinterest bot filter

waterWang:
`detectBrowser` in user-agent-utils.ts matched only Facebook iOS (`FBIOS`)
and the Google Search App (`GSA/`). All other in-app browsers fell through
to the platform webview (Chrome or Mobile Safari), making in-app traffic
impossible to segment, exclude from experiments, or even count.

This commit adds detection for:
- Facebook Android (`FBAN`/`FB_IAB`)
- Instagram
- LinkedIn (`LinkedInApp` / `[LinkedInApp]`)
- Twitter/X (`Twitter for`)
- TikTok (`musical_ly`)
- WeChat (`MicroMessenger/`)
- Line (`Line/`)

Also narrows the `pinterest` bot blocklist entry to `pinterestbot` so that
Pinterest's in-app browser (`[Pinterest/iOS]`) is not incorrectly dropped
as bot traffic.

Refs: PostHog/posthog#83413

### DIFF
--- a/packages/core/src/utils/bot-detection.ts
+++ b/packages/core/src/utils/bot-detection.ts
@@ -28,7 +28,7 @@ export const DEFAULT_BLOCKED_UA_STRS = [
   'msnbot',
   'nessus',
   'petalbot',
-  'pinterest',
+  'pinterestbot',
   'prerender',
   'rogerbot',
   'screaming frog',
--- a/packages/core/src/utils/user-agent-utils.ts
+++ b/packages/core/src/utils/user-agent-utils.ts
@@ -58,6 +58,11 @@ const PALE_MOON = 'Pale Moon'
 const WATERFOX = 'Waterfox'
 const BRAVE = 'Brave'
 const GOOGLE_SEARCH_APP = 'Google Search App'
+const LINKEDIN = 'LinkedIn'
+const TWITTER = 'Twitter'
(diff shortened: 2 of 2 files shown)


dustinbyrne (commented):
hey @waterWang, thanks for putting this together! after looking at this PR and the associated issue i've added some thoughts [here](https://github.com/PostHog/posthog/issues/83413#issuecomment-5318323948) if you'd like to take a look.

as is, this change isn't safe to merge because it changes the semantics of how we record `$browser` for all sites that upgrade. some users may be actively tracking the existing properties and this can constitute a breaking change for them.

we typically solve for this by [changing configuration defaults and gating by date](https://github.com/PostHog/posthog-js/blob/b2a445a32e685ec810edc0abb2eb05edbdaabdef/packages/browser/src/posthog-core.ts#L202), making this an opt-in change. we'd need that at minimum, but i'd also like to see where the discussion in the issue goes before we move forward with any concrete changes here.

open to any thoughts you might have on the matter!

dustinbyrne:
@waterWang thanks again for getting the ball rolling on this. given the discussion in https://github.com/PostHog/posthog/issues/83413, we will close this for now

PR #4543, opened 2026-08-17
fix(replay): force preserveDrawingBuffer before the recorder chunk loads

ksvat:
## Problem

Canvas recording can be enabled and working, and a WebGL canvas still replays blank.

A WebGL context created with `preserveDrawingBuffer: false` — the spec default, and what
most renderers ask for — lets the browser discard the drawn pixels as soon as the frame has
been composited. Replay reads pixels back out of the canvas some time after the page drew
into them, so the captured frames come back empty.

rrweb already forces that attribute on inside its `getContext` patch, but the patch is
installed by `CanvasManager`, which only exists once the lazily loaded recorder bundle has
arrived. Context attributes are fixed at creation time and cannot be changed afterwards, so
any renderer that boots while the page is loading — WebGL editors, map and 3D canvases,
WASM-backed engines — has already created a context we can never capture. The existing
`context.clear(COLOR_BUFFER_BIT)` fallback does not reliably rescue those pixels, and for a
renderer that only redraws on demand it can blank pixels that were still valid.

This came out of a support thread where an app embedding a third-party WebGL editor had
canvas capture switched on and still saw an empty canvas in every recording.

## Changes

Run the same `preserveDrawingBuffer` forcing synchronously during `posthog.init()`, so a
context created during page load is capture-ready before the recorder exists.

It is deliberately narrow — it forces the one attribute, only for WebGL context types, and
only when canvas recording is already known to be on: either declared client side via
`session_recording.captureCanvas.recordCanvas`, or read from a remote config persisted on an
earlier page load. Projects that turn canvas recording on purely in project settings
therefore get the fix from their users' next page load onwards.

Note for anyone tracking this: the patch has to live in `array.js`, not the recorder extension —
the entire point is to run before the extension has loaded. So this reaches users when they upgrade
the core bundle; a cached older `array.js` picking up fresh extensions will not get it. That is what
the backward-compatibility job flags, and why the new test is skipped below the shipping version.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

---
*Created with [PostHog Desktop](https://posthog.com/desktop?ref=pr)*


packages/browser/src/extensions/replay/preserve-drawing-buffer.ts
```diff
@@ -0,0 +1,49 @@
...
+        return
+    }
+    patched = true
+
+    const originalGetContext = canvasPrototype.getContext
+    canvasPrototype.getContext = function (this: HTMLCanvasElement, contextType: string, ...args: any[]) {
+        try {
+            if (WEBGL_CONTEXT_NAMES.indexOf(contextType) !== -1) {
+                if (isObject(args[0])) {
+                    args[0].preserveDrawingBuffer = true
```
TueHaulund, line 38: This changes an object the page gave us. If the page reuses that object for other canvases, we have changed it for them too. And if the object is frozen, this line throws, the `catch` hides it, and the canvas gets created without the fix.

Copying instead of editing avoids both:

```js
args[0] = isObject(args[0]) ? { ...args[0], preserveDrawingBuffer: true } : { preserveDrawingBuffer: true }
```

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -102,9 +103,34 @@ export class SessionRecording implements Extension {
...
+     * is on.
+     *
+     * Deliberately not gated on `_isRecordingEnabled`: that waits on a persisted remote config, which
+     * a first-ever page load does not have yet, and someone who asked for canvas recording in their
+     * own config should not have to wait a page load to get it. Turning canvas recording on purely in
+     * project settings still works, just from the second page load onwards - the first load persists
+     * the remote config, and `onRemoteConfig` covers any canvas created after it arrives.
+     */
+    private _preserveCanvasDrawingBuffers() {
+        if (!window || this._config.disable_session_recording || this._instance.consent.isOptedOut()) {
```
TueHaulund, line 123: This returns early when the visitor has not opted in yet, which is every load on a site with a cookie banner.

When they click accept, `opt_in_capturing()` starts recording but never calls this method again. It only replays remote config in the cookieless branch. So the patch never runs on that page, and canvas capture stays broken even though recording is now on. That is a common setup in the EU.

Calling `_preserveCanvasDrawingBuffers()` from `startIfEnabledOrStop()` would handle init, remote config, and opt-in in one place.

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -102,9 +103,34 @@ export class SessionRecording implements Extension {
     }
 
     initialize() {
+        this._preserveCanvasDrawingBuffers()
```
marandaneto, line 106: blocking: Deferred extension initialization misses the critical window

With `__preview_deferred_init_extensions: true`, this runs from a timer, but `_loaded()` and the customer's `loaded` callback run synchronously first. A WebGL context created there still precedes the patch, defeating the intended ordering fix. Please install this critical patch outside extension deferral and cover this ordering in a browser test.

marandaneto, line 106: blocking: Dynamically starting recording does not install the patch

If initialization used `disable_session_recording: true`, this call is skipped. A later `startSessionRecording()` only reaches `startIfEnabledOrStop()` through `set_config()`, so a canvas created immediately after that public call remains unpatched before the recorder chunk loads. Please install the patch synchronously on the recording-start/config-change path too.

packages/browser/src/extensions/replay/session-recording.ts
```diff
@@ -102,9 +103,34 @@ export class SessionRecording implements Extension {
...
+     * the remote config, and `onRemoteConfig` covers any canvas created after it arrives.
+     */
+    private _preserveCanvasDrawingBuffers() {
+        if (!window || this._config.disable_session_recording || this._instance.consent.isOptedOut()) {
+            return
+        }
+
+        const clientSide = this._config.session_recording?.captureCanvas?.recordCanvas
+        const serverSide = this._instance.get_property(SESSION_RECORDING_REMOTE_CONFIG)?.canvasRecording?.enabled
+        if (clientSide ?? serverSide) {
```
marandaneto, line 129: blocking: Known-disabled projects still have WebGL behavior changed

When persisted session-recording config explicitly has `enabled: false`, a local `captureCanvas.recordCanvas: true` still wins here and patches WebGL even though the recorder cannot start. That imposes the `preserveDrawingBuffer` memory/performance cost on a known-disabled project. Please distinguish missing remote config from explicit disablement.

packages/browser/src/extensions/replay/preserve-drawing-buffer.ts
```diff
@@ -0,0 +1,49 @@
...
+ *
+ * Patching during `posthog.init()` closes that window. It is deliberately narrow: it only forces the
+ * one attribute, and only when canvas recording is already known to be on.
+ */
+export function forcePreserveDrawingBuffer(): void {
+    const canvasPrototype = window?.HTMLCanvasElement?.prototype
+    if (patched || !canvasPrototype?.getContext) {
+        return
+    }
+    patched = true
```
marandaneto, line 31: blocking: The global patch is never removed

`patched` only transitions to true, and the original `getContext` is not retained for restoration. After stop, remote disablement, opt-out, or `shutdown()`, newly created contexts therefore keep forcing `preserveDrawingBuffer` for the page lifetime. Please add multi-instance-safe, reference-counted cleanup that restores the original when no recorder requires it.

marandaneto:
draft until reviews are addressed so its clear from our review queue

PR #4549, opened 2026-08-18
fix(error-tracking): recognise Firefox and Safari extension frames

ablaszkiewicz:
## Problem

An issue on posthog.com (`TypeError: Cannot call a class as a function`, Safari-only, single frame with filename `webkit-masked-url://hidden/`) is reported as if it were the site's own code. Two separate gaps let it through.

**1. Extension filtering is Chrome-only.** `_isExtensionException` tested only `chrome-extension://`, so Firefox (`moz-extension://`) and Safari extension frames were captured and reported against the site.

Worth noting: `safari-extension:` / `safari-web-extension:` are *already* synthesised by `extractSafariExtensionDetails` in the stack parser specifically to mark these frames — the drop check just never looked at them. Including them here is slightly beyond "add Firefox", but leaving them out would ship a fix that still misses the browser in the original report. Happy to split them out if reviewers would rather keep this minimal.

**2. Every browser frame is hardcoded `in_app: true`.** `createFrame` marked all frames as the site's own, so even non-app code pulled issues into the site's stack.

## What this does not do

`webkit-masked-url://` is **not** treated as an extension and is **not** dropped. Safari uses that placeholder for any script it will not attribute to the page — extension content scripts are the common case, but `blob:`, `eval`'d and injected code are masked identically, and the filename alone cannot distinguish them.

That matters because `_isExtensionException` uses `.some()`: adding the masked prefix to the drop list would discard the *entire* exception whenever one masked frame appears anywhere in the stack, including a genuine first-party bug where an extension has monkey-patched `fetch` mid-stack. Silently dropping real customer errors is a worse failure than the noise it removes.

So masked frames are marked `in_app: false` instead — the exception is still captured, it just stops counting as the site's own code.

## Note on the original diagnosis

The report that prompted this claimed the throwing bundle must be third-party because "posthog-js is compiled by TypeScript, which never emits `_classCallCheck`". That premise is wrong: the browser bundle runs through `@rollup/plugin-babel` with `@babel/preset-env` (`rollup.config.mjs:79-99`), and **both** targets downlevel `class` to ES5. Running the repo's exact config:

| `loose` | class → ES5 | emits `_classCallCheck` |
|---|---|---|
| `true` (current) | yes | no |
| `false` | yes | **yes** |

We don't emit that string today only because of `loose: true` on line 81 — one flag away from flipping. The "not ours" conclusion still holds (vendored rrweb, `lib/` and preact are all clean), just not for the stated reason. Flagging it so nobody re-derives the same false premise later.

## Testing

- New `packages/core/src/error-tracking/parsers/base.spec.ts` covers in-app, masked, and undefined-filename frames.
- Existing extension test in `posthog-exceptions.test.ts` becomes an `it.each` across all four schemes, plus a negative case ensuring a page URL that merely *contains* `moz-extension://` is still captured.
- Both changes mutation-tested: reverting each source change fails exactly the intended tests (1 and 3 respectively) and nothing else.
- `jest` (48 core + 31 browser), `eslint` and `prettier` clean. The one `tsc` error (`moduleResolution=node10` deprecation) is pre-existing and reproduces on a clean tree.

`filename` is typed `string` but arrives `undefined` at runtime from non-participating regex groups, so the new check is optional-chained — an unknown filename keeps the previous `in_app: true`.

### DIFF
--- a/.changeset/extension-frames-firefox-safari.md
+++ b/.changeset/extension-frames-firefox-safari.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+---
+
+Recognise Firefox and Safari extension frames when filtering extension exceptions, and stop counting Safari's masked `webkit-masked-url://` frames as in-app code.
--- a/packages/browser/src/__tests__/posthog-exceptions.test.ts
+++ b/packages/browser/src/__tests__/posthog-exceptions.test.ts
@@ -223,8 +223,13 @@ describe('PostHogExceptions', () => {
         })
 
         describe('Extension exceptions', () => {
-            it('does not capture exceptions with frames from extensions by default', () => {
-                const frame = { filename: 'chrome-extension://', platform: 'javascript:web' }
+            it.each([
+                ['chrome', 'chrome-extension://abc/content.js'],
+                ['firefox', 'moz-extension://abc/content.js'],
+                ['safari', 'safari-extension:abc/content.js'],
+                ['safari web', 'safari-web-extension:abc/content.js'],
+            ])('does not capture exceptions with frames from %s extensions by default', (_browser, filename) => {
+                const frame = { filename, platform: 'javascript:web' }
                 const exception = { stacktrace: { frames: [frame], type: 'raw' } }
                 exceptions.sendExceptionEvent({ $exception_list: [exception] })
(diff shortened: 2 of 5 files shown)


cat-ph (approved):
🚀

PR #4550, opened 2026-08-18
fix(ci): support pnpm 10 in PostHog upgrade workflow

marandaneto:
## Problem

The PostHog upgrade workflow checks out the `PostHog/posthog` repository, which uses pnpm 10. The recent upgrade to `pnpm/setup@v2` made the workflow fail during setup because that action only supports pnpm 11 or newer.

Failed run: https://github.com/PostHog/posthog-js/actions/runs/32130685267/job/95713128259

## Changes

Use `pnpm/action-setup@v6.0.10` for this workflow. This is the action recommended for pnpm 10. The action remains pinned to a full commit SHA and continues to read the pnpm version from the checked-out repository.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [ ] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi inspected the failed GitHub Actions job and made the workflow-only change. The workflow was validated with `actionlint`, with ShellCheck disabled because the file has a pre-existing style warning in an unrelated step. `git diff --check` also passed.

### DIFF
--- a/.github/workflows/posthog-upgrade.yml
+++ b/.github/workflows/posthog-upgrade.yml
@@ -49,10 +49,10 @@ jobs:
                   node-version: 24
 
             - name: Setup pnpm
-              uses: pnpm/setup@84cb39b217b10273981911c288cd62326dc7c6d2 # v2.0.2
+              uses: pnpm/action-setup@0977fd99725f1db4007ccb2928dbb4e90d06cc86 # v6.0.10
               with:
                   cache: false
-                  install: false
+                  run_install: false
 
             - name: Wait for package to become available on npm
               shell: bash


dustinbyrne (approved):

arnohillen (approved):

Verified the pin: `0977fd99725f1db4007ccb2928dbb4e90d06cc86` dereferences to the `pnpm/action-setup` v6.0.10 tag, and the failed job log confirms the cause verbatim ("this action only installs pnpm v11 or newer... To install older pnpm, use the pnpm/action-setup action instead") against PostHog/posthog's pinned `pnpm@10.29.3`.

arnohillen:

`.github/workflows/release.yml:653` has the identical break (the Build toolbar job checks out `PostHog/posthog`, which pins `pnpm@10.29.3`, then runs `pnpm/setup@v2.0.2`), so the next release will fail the same way unless it gets the same swap.

PR #4552, opened 2026-08-18
fix(ci): gate npm publishing on posthog-js S3 release

dustinbyrne:
## Problem

A browser SDK release can currently publish npm packages before its S3/CDN artifacts finish. If the toolbar or regional upload fails, npm can contain `posthog-js` or coordinated packages from a release whose browser assets are incomplete.

This is the base PR in stack #4554. The manual recovery workflow is stacked in #4553.

## Changes

- Check whether `posthog-js` has a new version before starting browser CDN work.
- Require a successful SDK build and complete US+EU S3 upload before any npm package from the release can publish.
- Keep `posthog-js` in a separate npm job while applying the same S3 gate to the general package matrix.
- Let releases without a new `posthog-js` version proceed without browser S3 work.
- Resolve one trusted toolbar source commit for both regions and use the pnpm 10-compatible setup action.
- Select regional AWS roles without falsey fallback and remove the obsolete npm-only partial-release notification.

This is CI-only and does not require a package changeset.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [ ] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi implemented the human-directed release ordering and branch split. Builtin reviewer agents checked the workflow DAG, failure/no-browser paths, regional source consistency, and AWS role selection. Validation used Actionlint with ShellCheck, Prettier, and `git diff --check`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -242,11 +242,13 @@ jobs:
 
     publish:
         name: Publish packages
-        needs: [version-bump, notify-approval-needed]
+        needs: [version-bump, notify-approval-needed, gate-posthog-js-publish]
         runs-on: ubuntu-latest
-        # Use `always()` to ensure the job runs even if the check-release-label job fails
-        # but still depend on it to be able to use `needs.notify-approval-needed.outputs.slack_ts`
-        if: always() && needs.version-bump.outputs.commit-hash != ''
+        # Use `always()` to evaluate the S3 gate even if an earlier dependency failed.
+        if: >-
+            always() &&
+            needs.version-bump.outputs.commit-hash != '' &&
+            needs.gate-posthog-js-publish.result == 'success'
         permissions:
             contents: write
             id-token: write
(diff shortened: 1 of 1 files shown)


arnohillen (approved):
<small><em>(created by PostHog Desktop)</em></small>

Traced every path through the new DAG and the gate fails closed in all of them, including an empty or unexpected `is-new-version` hitting the `*` case; `needs.upload-s3.result` aggregates the region matrix so a single-region S3 failure still blocks npm, and there is no cycle. Also confirmed the `steps: &publish-package-steps` anchor pair is valid, since GitHub Actions added YAML anchor support in September 2025.

PR #4555, opened 2026-08-18
fix(surveys): don't snapshot the whole page when closing a survey

HaynesPostHog:
## Problem

Closing an in-app survey popup could crash the Chrome tab — the grey "Aw, Snap!" page with `Error code: 5` (a renderer out-of-memory kill). It was reproducible on a heavy page such as a large dashboard, and took the whole tab down rather than just dismissing the survey.

Root cause: the close path animated the fade-out with `document.startViewTransition`, which snapshots the **entire page viewport**. The survey applied no `view-transition-name` scoping, so nothing limited that snapshot to the popup itself. On a heavy host page, capturing the whole-page snapshot could exhaust renderer memory and kill the tab.

[#4497](https://github.com/PostHog/posthog-js/pull/4497) fixed a *related* crash on this path (the transition's captured snapshot pointing at an already-removed node), but it kept the document-level view transition — so the whole-page snapshot cost, which is what OOMs on heavy pages, was still there. The crash was still reproducing on a build that already contained that fix.

## Changes

The survey renders into an isolated shadow root, so it never needed a document-level transition in the first place.

- Fade the popup out with a plain CSS opacity transition scoped to the survey's own container, then let React unmount it once the fade has run. No page snapshot, same fade-out UX.
- Use a timer rather than a `transitionend` listener. `transitionend` never fires for a cancelled or zero-duration transition (backgrounded tab, reduced motion), so a timer has to back it up regardless — and settling a few ms later than strictly necessary is imperceptible for an element that is already fully transparent. Dropping the listener also removes the double-settle guard and descendant-event filtering it would otherwise need.
- Keep the existing guard against a second close while one is animating.

Net **-66 lines**, and the replaced code path is simpler than what it replaces.

Tests cover the fade not removing the container, the unmount after the fade, repeated closes tearing down exactly once, and an explicit assertion that `document.startViewTransition` is never called — so this can't silently regress.

### Libraries affected

- [x] posthog-js (web)

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Investigated from a report that closing a survey popup crashed the tab on a dashboard page. Tools used: PostHog MCP (`execute-sql`, `surveys-get-all`) and the `debugging-surveys` skill, plus local git/grep/jest.

Decisions along the way:

- The `debugging-surveys` playbook steers most survey reports toward eligibility gates. Ruled that out early — the symptom was a renderer crash on dismiss, not a survey failing to show — and treated it as a rendering bug instead.
- Confirmed this was **not** a stale-SDK problem before writing any code: the crashing session ran `posthog-js` 1.417.1, that version is what the app serves fleet-wide, and 1.417.1 already contains [#4497](https://github.com/PostHog/posthog-js/pull/4497) (verified with `git merge-base --is-ancestor`). That ruled out "just upgrade" and established the earlier fix was incomplete rather than un-deployed.
- First implementation kept a `transitionend` listener with a fallback timer, a settle mutex, descendant filtering, and a forced reflow. On self-review that was strictly worse: the fallback timer must exist anyway, so the listener bought ~50ms on an already-invisible element while adding an `eslint-disable`, a forward-declared variable, and a lint violation against the repo's `addEventListener` helper rule. Cut it to the timer alone and dropped the reflow (unnecessary — the element has been painted at opacity 1 for a while).
- Considered dropping the close animation entirely, which would be the smallest possible diff, but that regresses the fade-out UX that this code path has had; kept the fade.

Verified: all 518 survey unit tests pass, plus typecheck and lint clean on the changed files. **Not** manually verified in a browser against a real heavy dashboard — the crash needs a real Chromium renderer under memory pressure, which I could not reproduce in this environment. Worth a manual check before release.

---
*Created with [PostHog](https://posthog.com?ref=pr) from a [Slack thread](https://posthog.slack.com/archives/C0ACRAMJUAG/p1787074502515409?thread_ts=1787074502.515409&cid=C0ACRAMJUAG)*

### DIFF
--- a/.changeset/survey-close-viewport-snapshot-crash.md
+++ b/.changeset/survey-close-viewport-snapshot-crash.md
@@ -1,0 +1,9 @@
+---
+'posthog-js': patch
+---
+
+Fix a Chrome renderer crash (grey "Aw, Snap" tab, "Error code: 5") that could still occur when closing an in-app survey on a heavy page such as a large dashboard.
+
+Closing a survey animated the fade-out with `document.startViewTransition`, which snapshots the entire page viewport. The survey applied no `view-transition-name` scoping, so on a heavy host page capturing that whole-page snapshot could exhaust renderer memory and crash the tab. A previous fix addressed a related crash (a snapshot pointing at a removed node) but left the document-level transition — and its whole-page snapshot cost — in place.
+
+The survey renders in an isolated shadow root, so it never needed a document-level transition. The close now fades the popup out with a plain CSS opacity transition scoped to the survey's own container, then unmounts it once the fade has run. No whole-page snapshot, no crash, same fade-out UX.
--- a/packages/browser/src/__tests__/extensions/surveys.test.ts
+++ b/packages/browser/src/__tests__/extensions/surveys.test.ts
@@ -262,7 +262,7 @@ describe('usePopupVisibility', () => {
     })
 })
 
-describe('usePopupVisibility view-transition close path', () => {
+describe('usePopupVisibility close animation path', () => {
     const mockSurvey: Survey = {
         id: 'testSurvey1',
         name: 'Test survey 1',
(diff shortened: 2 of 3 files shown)


thmsobrmlr (approved):
I can repro this pretty consistently on https://us.posthog.com/login?next=/project/2/dashboard/250270. (1) log out (2) navigate to the dashboard and log in (3) close the survey -> aw! snap.

Tested the animation on main and this PR with the vite playground, and all looks good to me.

PR #4556, opened 2026-08-18
feat(ai): add LangChain v1 agent middleware

### LINKED ISSUE #2907: Implement PostHog AI analytics middleware for LangChainJS v1.x

The PostHog AI Analytics product already ships a [LangChainJS v0.x callback](https://posthog.com/docs/llm-analytics/installation/langchain).

However, since LangChainJS v1.x, the callback pattern has evolved into the [middleware pattern](https://docs.langchain.com/oss/javascript/langchain/middleware/custom), with slight changes into how the logic works:

- Hooks for the agent as a whole (`beforeAgent` and `afterAgent`)
- Hooks for all model calls (`beforeModel` and `afterModel`)
- Wrapping tool calls and model calls with `wrapModelCall` and `wrapToolCall`.

I'd love for PostHog to ship an updated middleware for LangChain 1.x so that I can integrate it with a LangChain 1.x agent.

Implementing my own middleware, but having a first-party one would be great.

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

gouveags:
## Problem

`@posthog/ai` supports LangChain through its callback handler, but it does not expose a middleware-native integration for LangChain v1 agents. Users adopting the v1 middleware API therefore have no first-party middleware for capturing agent traces, model generations, and tool spans.

Closes #2907.

## Changes

- Add `createPostHogMiddleware` at `@posthog/ai/langchain/middleware`.
- Accept the same optional custom Zod state schema used by createAgent, preserving application state in trace input and output.
- Reuse the existing callback handler's provider, token, privacy, and event normalization.
- Keep `@posthog/ai/langchain` callback-only, so existing users do not need the optional `langchain` runtime.
- Cover real agent tool loops, failures, privacy, concurrency, explicit trace IDs, and callback-only module loading.

```ts
import { createAgent } from 'langchain'
import { createPostHogMiddleware } from '@posthog/ai/langchain/middleware'

const agent = createAgent({
  model,
  tools,
  middleware: [createPostHogMiddleware({ client: posthog })],
})
```

Use either the callback or the middleware, not both, to avoid duplicate events. On a terminal agent failure, LangChain does not invoke `afterAgent`; the failed model or tool event is still captured without a dangling `$ai_parent_id`, but no root trace is emitted.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [x] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Gabriel directed the contribution. I used Codex for implementation, test design, and independent review. The middleware lives in a separate subpath so callback-only users do not load the optional `langchain` runtime, while event normalization continues to use the existing callback handler. A public session link is unavailable in this environment.


gouveags:
@greptileai review

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,197 @@
...
+          'LangChain Agent',
+          { posthogStartTime: state._posthogStartTime }
+        )
+        callback.handleChainEnd(withoutPostHogState(state), runId)
+      })
+    },
+
+    wrapModelCall: async (request, handler) => {
+      const runId = uuidv7()
+      const parentRunId = getRunId(request.state)
```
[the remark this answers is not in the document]
gouveags: Fixed in d1698088. Middleware child events now preserve the generated agent root span ID as `$ai_parent_id`, while the configured `traceId` remains the shared `$ai_trace_id`. The concurrent explicit-trace regression verifies both root span IDs stay unique and both children attach to those roots.

gouveags:
@greptileai review

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,203 @@
...
+ *
+ * LangChain only invokes `afterAgent` for completed runs. A terminal agent
+ * failure still captures the failed model or tool call, but not a root trace.
+ */
+export const createPostHogMiddleware = (options: PostHogLangChainMiddlewareOptions) => {
+  const callback = new LangChainMiddlewareCallbackHandler(options)
+
+  return createMiddleware({
+    name: 'PostHogMiddleware',
+    stateSchema: postHogStateSchema,
```
marandaneto: blocking: Preserve custom agent state in traces — LangChain projects lifecycle-hook state through this middleware schema, which only declares the `_posthog*` fields plus built-ins. With a custom `createAgent({ stateSchema })`, fields such as tenant or workflow context remain in the agent result but are silently absent from `$ai_input_state` and `$ai_output_state`; a custom-state probe reproduced the omission.

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    wrapModelCall: async (request, handler) => {
+      const runId = uuidv7()
+      const parentRunId = getRunId(request.state)
+      const messages = [request.systemMessage, ...request.messages].filter(
+        (message): message is BaseMessage => message !== undefined
+      )
+      const invocationParams = {
+        ...request.modelSettings,
+        tools: request.tools,
```
marandaneto: blocking: Normalize tools before capture — `request.tools` contains `DynamicStructuredTool` instances, so standard `tool(...)` values serialize in `$ai_tools` as a generic `{"type":"not_implemented","id":["langchain","tools","DynamicStructuredTool"]}`, losing the actual name, description, and schema. Normalize these instances to capture-ready definitions before passing them to the callback handler.

marandaneto (commented):

marandaneto:
thanks @gouveags 
pinging @PostHog/team-ai-observability to double check this

gouveags:
I'll have a proper fix for the blocking review comments in a bit

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,203 @@
...
+
+      safely(() =>
+        callback.handleChatModelStart(
+          serializeModel(request.model),
+          [messages],
+          runId,
+          parentRunId,
+          { invocation_params: invocationParams },
+          undefined,
+          getModelMetadata(request.model, request.modelSettings)
```
carlos-marchal-ph, line 202: `request.modelSettings` is undefined for agent model calls, LangChain only populates it when another middleware adds one, and `BaseChatModel.getLsParams` reads `options.stop` straight off its argument. So this throws, the catch swallows it, and no `$ai_generation` from the middleware ever carries `$ai_model` or `$ai_provider`, which also means costs cannot be computed. The same field feeds `invocation_params` above, and since `modelSettings` is bind-time settings such as headers rather than the model's invocation parameters, `$ai_model_parameters` comes out empty too.

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,203 @@
...
+          { posthogStartTime: state._posthogStartTime }
+        )
+        callback.handleChainEnd(withoutPostHogState(state), runId)
+      })
+    },
+
+    wrapModelCall: async (request, handler) => {
+      const runId = uuidv7()
+      const parentRunId = getRunId(request.state)
+      const messages = [request.systemMessage, ...request.messages].filter(
```
carlos-marchal-ph: `request.systemMessage` is never undefined, it defaults to an empty `SystemMessage`, and LangChain drops it before the model call when the text is empty. Agents with no `systemPrompt` will therefore show a leading empty system message in `$ai_input` that was never actually sent.

packages/ai/src/langchain/middleware/index.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    beforeAgent: (state) => {
+      return {
+        _posthogRunId: uuidv7(),
+        _posthogStartTime: Date.now(),
+        _posthogInput: withoutPostHogState(state),
+      }
+    },
+
+    afterAgent: (state) => {
```
carlos-marchal-ph, line 166: On a terminal failure this never runs, so the model and tool events keep an `$ai_parent_id` pointing at an agent span nobody emits, and the failed run has no root carrying the input state, latency or error flag. Could the root be emitted from the `wrapModelCall` and `wrapToolCall` catch paths, or does an outer retry middleware make that unsafe?

packages/ai/src/langchain/callbacks.ts
```diff
@@ -98,11 +100,15 @@ export class LangChainCallbackHandler extends BaseCallbackHandler {
     tags?: string[],
     metadata?: Record<string, unknown>,
     _runType?: string,
-    runName?: string
+    runName?: string,
+    extra?: Record<string, unknown>
```
carlos-marchal-ph, line 104: This ninth argument is not part of LangChain's `handleChainStart` contract, and it widens a published class's public method to serve one internal caller. Could the middleware subclass override `handleChainStart` and fix up `startTime` after calling `super`, with `runs` made `protected`?

packages/ai/tests/langchain-middleware.test.ts
```diff
@@ -0,0 +1,248 @@
...
+  afterEach(() => {
+    jest.restoreAllMocks()
+  })
+
+  it('captures an agent trace with model and tool children', async () => {
+    const middleware = createPostHogMiddleware({
+      client: mockPostHogClient,
+      distinctId: 'user-id',
+      properties: { environment: 'test' },
+    }) as any
```
carlos-marchal-ph, line 67: Six of the eight cases cast to `any` and drive the hooks directly with a hand-built state, so they never exercise how LangChain builds the request or prunes state, and the two `createAgent` tests only assert event names. That is why both the pruned custom state and the missing model metadata go unnoticed here; could the lifecycle cases run through `createAgent` and assert properties instead?

packages/ai/src/langchain/middleware/index.ts
gouveags: Good catch on the dangling parent. An outer retry does make root finalization from these catch paths unsafe depending on middleware order. LangChain wrap hooks may call their handler more than once, and the first middleware is outermost: https://docs.langchain.com/oss/javascript/langchain/middleware/custom#execution-order

If PostHog is inside retry middleware, it observes an attempt that can still recover, so emitting the agent root there would falsely mark the whole run as failed. The safest behavior I see is to keep the failed child event `$ai_trace_id` but omit `$ai_parent_id` when no root is emitted, then cover PostHog both inside and outside model and tool retries. A complete failed root would require documenting PostHog as outermost and deduplicating finalization across parallel failures. Does that tradeoff match what you had in mind?

packages/ai/src/langchain/callbacks.ts
gouveags: Thanks for raising this. I checked the supported `@langchain/core@1.2.8` interface and runtime before changing it. `BaseCallbackHandler.handleChainStart` does declare `extra` as the ninth parameter, and `CallbackManager.handleChainStart` forwards that value to handlers. The current reference shows the same contract: https://reference.langchain.com/javascript/langchain-core/callbacks/base/BaseCallbackHandler/handleChainStart

Since this keeps the internal run storage private, I think the current approach is safer than making `runs` protected. Would you be okay with keeping it and adding a focused regression test for the propagated start time?

packages/ai/src/langchain/middleware/index.ts
gouveags: Fixed in `960b983`. `createPostHogMiddleware` now accepts the same custom Zod `stateSchema` used by `createAgent` and merges it with the private PostHog fields. A real-agent regression test verifies `tenantId` and `workflow` are present in both `$ai_input_state` and `$ai_output_state`, while `_posthog*` remains private.

packages/ai/src/langchain/middleware/index.ts
gouveags: Fixed in `960b983`. Model-call tools are now normalized with LangChain's `convertToOpenAITool` before capture. The real tool-loop test asserts the captured name, description, and JSON schema instead of the `DynamicStructuredTool` placeholder.

packages/ai/src/langchain/middleware/index.ts
gouveags: Fixed in `960b983`. Metadata extraction now passes `{}` when `modelSettings` is absent, and model parameters come from `invocationParams` with bind-time settings as fallback. The real-agent test uses a `BaseChatModel` path and asserts model, provider, and temperature.

packages/ai/src/langchain/middleware/index.ts
gouveags: Fixed in `960b983`. Captured messages now use the same `systemMessage.text === ''` condition as LangChain. Real-agent regressions cover both the default empty message being omitted and a non-empty system prompt being retained.

packages/ai/tests/langchain-middleware.test.ts
gouveags: Addressed in `960b983`. I added five property-level `createAgent` regressions covering custom state, model/provider/parameters, real tool schemas, and empty/non-empty system prompts. I kept the direct hook tests for exact return/error identity, concurrency, privacy, and tool-error behavior, where controlled hook-level coverage is still useful.

packages/ai/src/langchain/callbacks.ts
carlos-marchal-ph: Yep sounds good!

packages/ai/src/langchain/middleware/index.ts
carlos-marchal-ph: Yep just skipping the property is what I meant

gouveags: Fixed in `86cb23c`. Thrown model and tool failures keep their `$ai_trace_id` and event type, but now omit `$ai_parent_id` when LangChain will not emit the root. The tests cover terminal model and tool failures plus a recovered retry.

packages/ai/src/langchain/callbacks.ts
gouveags: Added the focused start-time latency regression in `86cb23c`. Thanks!

gouveags:
@carlos-marchal-ph and @marandaneto thanks for the reviews! Especially Carlos for being so fast!

All comments are now resolved. Let me know if something else pops up!

gouveags:
I checked the Python SDK and decided to open an [issue](https://github.com/PostHog/posthog-python/issues/901) there too. If you guys can check it, that would be nice.

marandaneto (approved):

marandaneto:
@PostHog/team-ai-observability if you're happy, we can merge this
thanks @gouveags

PR #4560, opened 2026-08-19
fix(replay): guard lazy recorder startup after teardown

skvark:
## Summary

- ignore lazy recorder callbacks when recording is no longer allowed or the session manager has changed
- dispose and detach recorders before cookieless identity changes, discarding pending producer and compression work
- add regression coverage for manager replacement, shared consent changes, and synchronous or deferred teardown events

## Why

Session recording downloads its recorder code asynchronously. A user can opt out
while that download is still in progress. In `cookieless_mode: 'on_reject'`,
opt-out removes session recording and destroys its session manager.

When the delayed download callback runs, it can still try to create a recorder.
Without a session manager, that constructor throws:

`[SessionRecording] must be started with a valid sessionManager.`

A quick opt-out followed by opt-in creates a second case: the old callback can see
the new session manager and start a detached recorder beside the replacement.
This change ties each callback to the manager that started it, rechecks whether
recording is still allowed, and disposes the old recorder before either identity
transition. Final disposal also invalidates pending compression and clears events
emitted while rrweb stops, so detached work cannot flush under a replacement identity.

GitHits helped find and verify this bug.[^investigation]

[^investigation]: **GitHits helped find and verify this bug.** Starting from the reported runtime error, GitHits' version-pinned source search resolved `posthog-js@1.418.1` to commit `16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd`. It showed that replay eligibility is checked before the asynchronous recorder download ([load path](https://github.com/PostHog/posthog-js/blob/16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd/packages/browser/src/extensions/replay/session-recording.ts#L153-L179)), but the completion callback initializes the recorder without rechecking consent or `sessionManager` ([callback](https://github.com/PostHog/posthog-js/blob/16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd/packages/browser/src/extensions/replay/session-recording.ts#L336-L355)). During cookieless opt-out, PostHog removes replay and then destroys and clears the manager ([teardown](https://github.com/PostHog/posthog-js/blob/16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd/packages/browser/src/posthog-core.ts#L4324-L4335)); a delayed callback can therefore construct the lazy recorder after teardown, whose constructor immediately reads the manager ([constructor](https://github.com/PostHog/posthog-js/blob/16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L607-L616)) and throws the reported error when it is missing ([guard](https://github.com/PostHog/posthog-js/blob/16b9d0a0a4da0eb6aad314bf8e330fc555d8e8dd/packages/browser/src/extensions/replay/external/lazy-loaded-session-recorder.ts#L575-L581)). The earlier fix in [#3544](https://github.com/PostHog/posthog-js/pull/3544) covered late recorder events after teardown, but not this pending download callback.

## Validation

- 420 focused replay and cookieless unit tests passed across 4 suites
- 1 replay snapshot passed
- browser TypeScript typecheck passed
- targeted ESLint and diff checks passed

### DIFF
--- a/.changeset/gentle-otters-record.md
+++ b/.changeset/gentle-otters-record.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Prevent stale session recorders from initializing after consent or session teardown, and dispose of recorders when switching to cookieless mode.
--- a/packages/browser/src/__tests__/cookieless.test.ts
+++ b/packages/browser/src/__tests__/cookieless.test.ts
@@ -314,7 +314,10 @@ describe('cookieless', () => {
             const { posthog, beforeSendMock } = await setup({
                 cookieless_mode: 'on_reject',
             })
+            const detachedSessionRecording = posthog.sessionRecording!
+            const detachedRemoteConfig = jest.spyOn(detachedSessionRecording, 'onRemoteConfig')
             posthog.opt_out_capturing()
+            expect(posthog['_extensions']).not.toContain(detachedSessionRecording)
             posthog.register({ test: 'test' })
             posthog.capture(eventName, eventProperties)
             expect(beforeSendMock).toBeCalledTimes(2)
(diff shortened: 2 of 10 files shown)


marandaneto (approved):

marandaneto:
@PostHog/team-replay lgtm, pls double check

TueHaulund (approved):

skvark:
Rebased onto the latest `main` and tightened teardown handling so delayed recorder loads, pending compression, and events emitted while rrweb stops cannot survive a cookieless identity transition.

Validation:

- 420 focused replay and cookieless tests passed
- browser TypeScript typecheck passed
- targeted ESLint and diff checks passed

PR #4563, opened 2026-08-19
feat(webpack): add event release mode with native debug ids

ablaszkiewicz:
In this PR we add support for the new releases mode to rely on webpack generated debug ids.

It needs this PR first https://github.com/PostHog/posthog/pull/85307 - it has been tested in that PR.

### DIFF
--- a/packages/webpack-plugin/src/index.spec.ts
+++ b/packages/webpack-plugin/src/index.spec.ts
@@ -31,6 +31,7 @@ const config: ResolvedPluginConfig = {
     sourcemaps: {
         enabled: true,
         deleteAfterUpload: true,
+        releaseMode: 'symbol-set',
     },
 }
 
@@ -52,6 +53,18 @@ async function exists(filePath: string): Promise<boolean> {
     } catch {
         return false
     }
+}
+
+function createCompiler(version: string | undefined): {
+    compiler: webpack.Compiler
+    sourceMapDevToolPlugin: jest.Mock
+} {
+    const sourceMapDevToolPlugin = jest.fn().mockImplementation(() => ({ apply: jest.fn() }))
+    const compiler = {
+        webpack: { SourceMapDevToolPlugin: sourceMapDevToolPlugin, version },
+        hooks: { done: { tapAsync: jest.fn() } },
+    } as unknown as webpack.Compiler
+    return { compiler, sourceMapDevToolPlugin }
 }
 
 describe('PosthogWebpackPlugin', () => {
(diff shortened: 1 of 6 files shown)


ablaszkiewicz:
Verified the nextjs-config path end to end against a local stack (next 15.5, webpack build): both compilations upload per build, event mode injects release ids into client and server chunks, a rebuild of unchanged code dedupes all 19 chunks as already present, and a server-captured exception came back with $release_id and frames resolved to the original app sources. One finding: Next's vendored webpack reports no `compiler.webpack.version`, so the native debug ids gate always declines there and the CLI's content-addressed ids carry event mode - graceful fallback, working as designed, now documented in the example.

packages/webpack-plugin/src/index.spec.ts
```diff
@@ -54,6 +55,18 @@ async function exists(filePath: string): Promise<boolean> {
     }
 }
 
+function createCompiler(version: string | undefined): {
+    compiler: webpack.Compiler
+    sourceMapDevToolPlugin: jest.Mock
+} {
+    const sourceMapDevToolPlugin = jest.fn().mockImplementation(() => ({ apply: jest.fn() }))
```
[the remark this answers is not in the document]
ablaszkiewicz: A "complete" webpack.Compiler mock is dozens of properties the test never touches. Ignoring

packages/webpack-plugin/src/index.ts
```diff
@@ -22,11 +37,18 @@ export class PosthogWebpackPlugin {
...
+            // own, so one id identifies the chunk across the whole toolchain. On webpacks
+            // without the option the CLI falls back to content-derived ids, which are equally
+            // stable — just not shared with other tooling.
+            const eventReleaseMode = this.resolvedConfig.sourcemaps.releaseMode === 'event'
             new compiler.webpack.SourceMapDevToolPlugin({
                 filename: '[file].map',
                 noSources: false,
                 moduleFilenameTemplate: '[resource-path]',
                 append: this.resolvedConfig.sourcemaps.deleteAfterUpload ? false : undefined,
+                ...(eventReleaseMode && webpackSupportsDebugIds(compiler.webpack.version) ? { debugIds: true } : {}),
```
marandaneto, line 51: blocking: Bundled CLI does not adopt webpack debug IDs

`@posthog/cli` is still pinned to `~0.13.0`. That release recognizes `//# chunkId=` in JavaScript while treating the sourcemap's existing `debugId` as an ID, so enabling webpack's native `debugIds` can make the CLI inject a different PostHog chunk ID into the JavaScript but skip adjusting the sourcemap. The artifacts then disagree and mappings are not shifted for the prepended runtime snippet, breaking symbolication. Please gate this on a CLI release that adopts `//# debugId` consistently in both artifacts (with an integration regression test), or keep native `debugIds` disabled.

marandaneto (approved):
approving to unblock, see comments

packages/webpack-plugin/src/index.ts
ablaszkiewicz: yes - waiting for linked PR to be merged to bump the version

ablaszkiewicz: bumped

PR #4569, opened 2026-08-19
fix(node): return undefined instead of null variant in getFeatureFlagResult

v11t:
## Problem

The remote evaluation path in `_getFeatureFlagResult` copies `flagDetail.variant` straight from the parsed `/flags` response, where missing variants are serialized as `null`. `FeatureFlagResult.variant` is typed `string | undefined`, so consumers that serialize or validate the result get a value the type says cannot exist. We hit this caching the result with a zod schema written against the type: every cached entry failed validation. The override and local evaluation paths already normalize to `undefined`.

## Changes

- Normalize `null` variant to `undefined` in the remote evaluation path
- Regression test (fails without the fix: `null` does not `toEqual` `undefined`)

### DIFF
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -1347,7 +1347,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
           result = {
             key,
             enabled: flagDetail.enabled,
-            variant: flagDetail.variant,
+            // The flags API serializes missing variants as null
+            variant: flagDetail.variant ?? undefined,
             payload: parsedPayload,
           }
         }
--- a/.changeset/node-null-variant.md
+++ b/.changeset/node-null-variant.md
@@ -1,0 +1,5 @@
+---
+'posthog-node': patch
+---
+
+fix getFeatureFlagResult returning null variant instead of undefined when remotely evaluated
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -355,6 +355,7 @@ describe('Lazy SessionRecording', () => {
     })
 
     afterEach(() => {
+        sessionRecording.stopRecording()
         // @ts-expect-error this is a test, it's safe to write to location like this
         window!.location = originalLocation
     })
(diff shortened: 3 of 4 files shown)


marandaneto:
thanks @v11t

marandaneto (approved):

packages/node/src/client.ts
```diff
@@ -1347,7 +1347,8 @@ export abstract class PostHogBackendClient extends PostHogCoreStateless implemen
           result = {
             key,
             enabled: flagDetail.enabled,
-            variant: flagDetail.variant,
+            // The flags API serializes missing variants as null
+            variant: flagDetail.variant ?? undefined,
```
marandaneto, line 1351: SDK API is value | undefined
should the SDK API change to value | undefined | null or is this just a SDK bug?
Do we expect 'null' to be a valid variant value?

v11t: SDK bug imo?
The /flags API serializes a missing variant as null, every other path in client.ts already normalizes it (`typeof value === 'string' ? value : undefined`, e.g. overrides and local eval); only this remote path leaked the raw value. A variant is always a string key, so null only ever means no variant and the API can stay `string | undefined`.

v11t: originally (https://github.com/PostHog/posthog/blob/83ff51ff795d/rust/feature-flags/src/api/types.rs#L430-L443) - 

```rust
pub variant: Option<String>,
```

-> no skip attribute, so serde serializes `None` as `"variant": null` (unlike the 
```rust
#[serde(skip_serializing_if = "std::ops::Not::not")]
pub failed: bool,
```
just below it)

sooo technically you can have undefined there directly via

```rust
 #[serde(skip_serializing_if = "Option::is_none")]
 ```
 
 ? 
 
 but that will affect [needs fact checking, maybe everyone is already treating a missing key same way as null] all /flags consumers (all sdks and any direct callers), so arguably a larger blast radius

marandaneto: i asked the flags team to be sure, but i agree with the changes

JamesPatrickGill: After a quick sanity check, this all looks good from a feature flags perspective, thank you for the contribution 🙏

haacked: I agree. LGTM!

PR #4573, opened 2026-08-19
feat(react-native): record automatic exception steps

jakesciotto:
Refs PostHog/posthog#43989, the public request for Sentry-style breadcrumbs. That issue asks for them in the browser SDK, so this does not close it.

## Problem

`addExceptionStep` shipped in `posthog-react-native` 4.50.0 (#3861), and it does what the docs promise. But every step is manual, so an exception carries a timeline only where somebody remembered to add a call. Bugsnag and Sentry leave breadcrumbs for navigation, network requests, taps and lifecycle on their own, so a crash arrives with a trail even in code nobody instrumented.

No PostHog SDK records a step automatically today. I checked `posthog-js`, `posthog-ios`, `posthog-android` and `posthog-flutter`: `addExceptionStep` is a public API in each, and nothing calls it internally.

This came from a React Native customer who asked for the Bugsnag Breadcrumbs equivalent, found the manual API, and said the manual calls feel like the part they would forget.

## What this does

Set `errorTracking.exceptionSteps.automatic` to record a step for a signal the SDK already observes:

```ts
const posthog = new PostHog('<token>', {
  errorTracking: {
    exceptionSteps: {
      automatic: true, // or { navigation: true, taps: true, lifecycle: true }
    },
  },
})
```

| Signal | Source event | Step |
| --- | --- | --- |
| `navigation` | `$screen` | `Screen: Cart`, `$type: navigation` |
| `taps` | `$autocapture` (touch) | `Tap: CheckoutButton`, `$type: tap` |
| `lifecycle` | `Application Opened` and the other four | `Application Backgrounded`, `$type: lifecycle` |

Every signal is off by default. Each step adds bytes to every captured exception, so this is opt-in, which also matches `errorTracking.autocapture`.

## How it works

The recorder hangs off the existing `processBeforeEnqueue` chokepoint in `posthog-rn.ts`, which already sees `$screen`, `$autocapture` and the lifecycle events. Nothing new is patched or wrapped. The mapping itself is a pure function, so it is testable without the SDK.

Automatic steps share the byte-bounded buffer and the native forwarding that manual steps use, so a native crash carries the same timeline.

## Decisions worth a look

1. **`$type` on the wire.** The error tracking UI already reads `$type` and `$level` per step (`frontend/src/lib/components/Errors/exceptionStepsValidation.ts`) and renders `$type` as the secondary line in the session timeline, but no SDK emits either one. This adds both to `EXCEPTION_STEP_INTERNAL_FIELDS` in `@posthog/core` to match that contract. Neither becomes reserved, so a caller can still categorise a manual step by hand.
2. **A manual step stays untyped.** Stamping `$type: 'manual'` would put a new label next to every existing step for every current user, so only automatic steps carry `$type`.
3. **A dropped event leaves no step.** A caller drops an event in `before_send` to keep data out of PostHog. Resurrecting it inside an exception payload would defeat that, so only a surviving event records a step.
4. **A tap step carries neither `$el_text` nor coordinates.** Element text is user-visible copy that can hold personal data, and the label from `tag_name` is either a `ph-label` or a component display name.
5. **The config type and resolver live in `@posthog/core`**, not in the RN package, so iOS, Android and web can adopt the same shape and the same `$type` vocabulary. Only React Native reads it today.

## Not in this PR

- **Network request steps.** Bugsnag's `request` breadcrumb is the notable remaining gap. It needs request interception, and the RN SDK patches only `fetch`, not `XMLHttpRequest`, so an `axios` default would go unrecorded. That deserves its own change.
- **Console steps.** Console output already has two homes in this SDK: `errorTracking.autocapture.console` and the Logs product. A third path would be redundant.
- **Docs.** `posthog.com/docs/error-tracking/capture` needs an automatic-steps section. Happy to follow with that PR once the config shape is settled here.

## Testing

- 41 tests across the exception step specs, 626 across the whole React Native package, all passing.
- New: `exception-steps-automatic.spec.ts` covers the mapping and the buffer wiring; `exception-steps-automatic-capture.spec.ts` drives a live client through `ready()` so the assertions run through the real enqueue path rather than a mocked one.
- `automatic-steps.ts` is at 100% statement coverage.
- Core: `resolveAutomaticExceptionStepsConfig` and the `$type`/`$level` strip behavior are covered; the full core suite passes (845).
- Three failures in the `browser` package (`persistence-key-policy`, `module.test.ts`, and a playwright spec run under jest) reproduce on a clean `main` with these changes stashed, so they are unrelated.
- `errorTracking.exceptionSteps.automatic: true` is wired into `example-expo-53` for manual verification: switch tabs, tap a button, background the app, then capture an exception and read the timeline.

### DIFF
--- a/.changeset/automatic-exception-steps-react-native.md
+++ b/.changeset/automatic-exception-steps-react-native.md
@@ -6,3 +6,5 @@
 Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
 
 Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
+
+`reset()` now clears the exception-step buffer, for automatic and manual steps alike. Exception steps are user-session state, so on a shared device the previous user's screen names and tap labels must not reach the next user's exception. The buffer still survives a capture, so every exception in one session carries the same steps.
--- a/.changeset/automatic-exception-steps-react-native.md
+++ b/.changeset/automatic-exception-steps-react-native.md
@@ -3,7 +3,7 @@
 'posthog-react-native': minor
 ---
 
-Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
+Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap or React Native Web click (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
 
 Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
 
--- a/.changeset/automatic-exception-steps-react-native.md
+++ b/.changeset/automatic-exception-steps-react-native.md
@@ -1,0 +1,8 @@
+---
+'@posthog/core': minor
+'posthog-react-native': minor
+---
+
+Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
+
+Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
--- a/.changeset/automatic-exception-steps-react-native.md
+++ b/.changeset/automatic-exception-steps-react-native.md
@@ -7,4 +7,4 @@ Add automatic exception steps to React Native, so a captured exception carries a
 
 Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
 
-`reset()` now clears the exception-step buffer, for automatic and manual steps alike. Exception steps are user-session state, so on a shared device the previous user's screen names and tap labels must not reach the next user's exception. The buffer still survives a capture, so every exception in one session carries the same steps.
+The buffer keeps its existing lifetime. A capture keeps the steps, and `reset()` keeps them too, because steps scope to the app session rather than to the user session.
--- a/.changeset/automatic-exception-steps-react-native.md
+++ b/.changeset/automatic-exception-steps-react-native.md
@@ -3,8 +3,8 @@
 'posthog-react-native': minor
 ---
 
-Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap or React Native Web click (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
+Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap or React Native Web click (`taps`), an app lifecycle transition (`lifecycle`), or an identity change (`identity`). Every signal stays off by default, because each step adds bytes to every captured exception.
 
 Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
 
-The buffer keeps its existing lifetime. A capture keeps the steps, and `reset()` keeps them too, because steps scope to the app session rather than to the user session.
+The buffer keeps its existing lifetime. A capture keeps the steps, and `reset()` keeps them too, because steps scope to the app session rather than to the user session. The `identity` signal marks that boundary instead of erasing it: it records a `User changed` step at `reset()`, so the previous user's steps do not read as the next user's. The step carries no distinct ID.
--- a/packages/core/src/error-tracking/exception-steps.ts
+++ b/packages/core/src/error-tracking/exception-steps.ts
@@ -1,14 +1,33 @@
 import { isArray, isNumber, isObject, isString, safeJsonStringify } from '@/utils'
 
 export const EXCEPTION_STEP_INTERNAL_FIELDS = {
+  TYPE: '$type',
   MESSAGE: '$message',
+  LEVEL: '$level',
   TIMESTAMP: '$timestamp',
 } as const
 
+/**
+ * Only `$message` and `$timestamp` are reserved: the SDK owns their canonical values, so a caller
+ * cannot spoof them. `$type` and `$level` stay writable by callers who want to categorise their own
+ * steps, and the SDK sets `$type` itself only on automatic steps.
+ */
 const RESERVED_EXCEPTION_STEP_KEYS = new Set<string>([
   EXCEPTION_STEP_INTERNAL_FIELDS.MESSAGE,
   EXCEPTION_STEP_INTERNAL_FIELDS.TIMESTAMP,
 ])
+
+/**
+ * `$type` values the SDK sets on automatic steps. Every SDK uses these exact strings, so the error
+ * tracking timeline labels an automatic step the same way on web and on mobile.
+ */
+export const EXCEPTION_STEP_TYPES = {
+  NAVIGATION: 'navigation',
+  TAP: 'tap',
+  LIFECYCLE: 'lifecycle',
+} as const
+
+export type ExceptionStepType = (typeof EXCEPTION_STEP_TYPES)[keyof typeof EXCEPTION_STEP_TYPES]
 
 export type ExceptionStep = {
   [EXCEPTION_STEP_INTERNAL_FIELDS.MESSAGE]: string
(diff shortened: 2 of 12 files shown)


packages/react-native/src/error-tracking/index.ts
```diff
@@ -135,6 +148,44 @@ export class ErrorTracking {
...
+    }
+
+    try {
+      const step = buildAutomaticExceptionStep(this._automaticStepsConfig, event, properties)
+      if (!step) {
+        return
+      }
+
+      const stepProperties = { [CoreErrorTracking.EXCEPTION_STEP_INTERNAL_FIELDS.TYPE]: step.type }
+      this._exceptionStepsBuffer.add({
```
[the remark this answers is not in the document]
jakesciotto: Confirmed, and fixed in the JS layer in b676545d. Thanks.

`clearExceptionSteps()` only ran in `_shutdown()`, so the buffer did survive `posthog.reset()`. `reset()` now clears it, for automatic and manual steps alike. The clear runs outside the wrapped `super.reset()`, so a disabled or not-yet-initialized client still drops the steps. Two tests cover it: one asserts an exception after `reset()` carries no steps, and one asserts recording resumes afterwards.

Two notes for the reviewers on the rest of it.

**This reverses a deliberate line of the previous contract.** The `clearExceptionSteps()` docstring said the buffer clears on SDK close and "not on capture or identity changes" (#3861). I updated that comment. I think privacy wins here, and the repo already applies the same rule twice: `reset()` clears `SessionReplayEventTriggerActivatedSession` so "one user's activation" cannot leak into the next, and `PostHogSDK.reset()` in posthog-ios clears `_lastScreenName`, which is the same kind of label. @ioannisj, say so if you chose the old behavior for a reason I am missing.

**The native half is not fixed, and I cannot fix it from this PR.** The bridge in `@posthog/react-native-plugin` exposes `addExceptionStep` and no clear, and `PostHogSDK.reset()` in posthog-ios does not clear its own `exceptionStepsBuffer`. So after `reset()` the JS buffer is empty while the native buffer still holds the previous user's steps, and a native crash would carry them. Closing that needs a clear on the native buffer in posthog-ios and posthog-android, plus a bridge method to call it at the identity boundary. Happy to file that as a follow-up issue across the three repos.

Worth flagging separately: `posthog-js` does not clear on `reset()` either. Its window is much smaller, because it clears the buffer after every successful capture, but the boundary is the same. Out of scope here.

packages/node/references/posthog-node-references-latest.json
```diff
@@ -1727,6 +1727,28 @@
       ],
       "path": "src/types.ts"
     },
+    {
+      "id": "AutomaticExceptionStepsConfig",
```
hpouillot: Can we specialize the react native config for this as I don't think it applies to node ?

packages/core/src/error-tracking/exception-steps.ts
```diff
@@ -1,15 +1,34 @@
 import { isArray, isNumber, isObject, isString, safeJsonStringify } from '@/utils'
 
 export const EXCEPTION_STEP_INTERNAL_FIELDS = {
+  TYPE: '$type',
   MESSAGE: '$message',
+  LEVEL: '$level',
```
hpouillot: do we use level anywhere ?

.changeset/automatic-exception-steps-react-native.md
```diff
@@ -0,0 +1,10 @@
+---
+'@posthog/core': minor
+'posthog-react-native': minor
+---
+
+Add automatic exception steps to React Native, so a captured exception carries a timeline even where nobody called `addExceptionStep`. Set `errorTracking.exceptionSteps.automatic` to `true`, or to an object, to record a step for a screen change (`navigation`), an autocaptured tap (`taps`), or an app lifecycle transition (`lifecycle`). Every signal stays off by default, because each step adds bytes to every captured exception.
+
+Automatic steps carry `$type`, which the error tracking timeline already renders, and they share the byte-bounded buffer and the native forwarding that manual steps use. A manual step stays untyped. An event that `before_send` dropped leaves no step.
+
+`reset()` now clears the exception-step buffer, for automatic and manual steps alike. Exception steps are user-session state, so on a shared device the previous user's screen names and tap labels must not reach the next user's exception. The buffer still survives a capture, so every exception in one session carries the same steps.
```
ioannisj: This leaves RN the odd one out from our SDKs? iirc, web and other mobile SDKs don't clear exception steps on reset. 

This was previously discussed [here](https://posthog.slack.com/archives/C07AA937K9A/p1781516724482669?thread_ts=1781508756.725049&cid=C07AA937K9A). 

For exception tracking I think steps should be scoped to app session and not user session. The whole idea isn't understanding user behavior, it's understanding app behavior, and user changing is part of a normal app flow. Losing the steps around a login or logout flow is losing exactly the context you'd want when something breaks there?

cc @hpouillot wdyt?

ioannisj (commented):
Left a small comment on reset() behavior. Looks good to me otherwise, i'll leave the final approval for error tracking team since this is ultimately a product decision

packages/node/references/posthog-node-references-latest.json
jakesciotto: Agreed, done in 4042b7de. `AutomaticExceptionStepsOptions`, the resolver and the `$type` vocabulary now live in `packages/react-native/src/error-tracking/automatic-steps.ts`. `navigation`, `taps` and `lifecycle` are signals React Native observes, so they had no business in a type `posthog-node` re-exports.

This file is back to what main has. Core now adds one field name and nothing else.

packages/core/src/error-tracking/exception-steps.ts
jakesciotto: No, nothing writes it. I added `$level` only to mirror the field the error tracking timeline already reads, and this PR never sets it. I removed it in 4042b7de.

Core now adds only `TYPE: '$type'`, which automatic steps do set. Neither field is reserved, so a caller can still categorise a manual step by hand.

.changeset/automatic-exception-steps-react-native.md
jakesciotto: You are right, and I reverted it in 4042b7de. RN keeps steps across `reset()` again, so it matches web, iOS, Android and Flutter. I added a test that locks the behavior in, plus a comment on `reset()` so the next person does not clear it by accident.

Your app-behavior argument is the one that convinced me, and my privacy argument turned out weaker than I first thought. Automatic steps carry no `$el_text` and no touch coordinates, only structural labels like `Screen: Cart` and `Tap: CheckoutButton`, so what crossed the boundary was app shape rather than user content.

For context on why I touched it at all: the review bot flagged the shared-device case, and automatic steps do change the volume, because the buffer now fills on every screen change and tap instead of only where an app calls `addExceptionStep`. That is a difference in degree, not a reason to break consistency with four SDKs.

Two notes from the thread you linked, for whoever makes the call.

The identity-change question reads as open rather than settled. @hpouillot said "I think I would scope it to the session not the device", then "so ideally we reset the buffer when user changes", then "but I don't mind releasing a first version without". You said "I don't have a strong opinion tbh" and made the app-level argument, and he answered "Yes ok make sense". So the current behavior looks like the v1 choice, and hugues leaned the other way at the time.

There is also a middle option in that thread, and it is hugues's: "we could also create an automatic exception step 'user changed'?" You deferred it partly because you did not want to "mix up completely manual + semi-automatic api at the sdk level just yet". This PR adds exactly that semi-automatic API, so that objection has lapsed. An identity step at `reset()` keeps the login and logout context you want, and it marks the boundary so nobody reads user A's steps as user B's.

I am happy to add that step here, or to leave it out and keep this PR to automatic steps only. Your call, and hugues's.

packages/react-native/src/posthog-rn.ts
```diff
@@ -757,6 +757,11 @@ export class PostHog extends PostHogCore {
    * @public
    */
   reset(propertiesToKeep?: PostHogPersistedProperty[]): void {
+    // NOTE: reset() deliberately keeps the exception-step buffer. Steps scope to the app session, not
```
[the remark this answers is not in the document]
jakesciotto: Tracking, but not changing it. You flagged this on the first commit, I did clear the buffer in `reset()`, and @ioannisj asked for that revert on review. Exception steps scope to the app session across every PostHog SDK, so clearing here would make React Native the only one that drops the steps explaining a failure in a login or logout flow.

The exposure is narrower than it reads. An automatic step carries no `$el_text` and no touch coordinates, only a structural label such as `Screen: Cart` or `Tap: CheckoutButton`, so what crosses the boundary is app shape rather than user content. A manual step can carry more, and the docs already tell callers to keep personal data out of step payloads.

The open follow-up is hugues's idea from the original design thread: record a "user changed" step at `reset()` instead of clearing. That keeps the login and logout context and marks the identity boundary in the timeline. That decision sits with the error tracking team, in the thread above.

.changeset/automatic-exception-steps-react-native.md
ioannisj: I agree that a "user changed" step would be valuable here as well, esp now that we're adding automatic exception steps

jakesciotto:
Rebased onto `main` and fixed a regression the rebase surfaced.

**The conflict was the example app.** `examples/example-expo-53` was removed on `main` and replaced by `example-expo-57`, and this branch wires the demo config into two files under the old path. Git's rename detection resolves it locally and carries the wiring into `example-expo-57`; GitHub's merge doesn't detect the rename across that many files, which is why the PR read as conflicting. Nothing was lost — the demo config now sits in `examples/example-expo-57/posthog.ts`.

**The regression is worth a look, because it made the feature dead on web.** `buildTapStep` matched only `$event_type: 'touch'`, which was every interaction autocapture emitted when I wrote this. #4643 then added click autocapture for React Native Web, emitting `$event_type: 'click'`. So on RN Web a click recorded no step at all, and `taps` would have shipped silently doing nothing there.

Both types now map to the `tap` step type — they are one signal to a caller who enabled `taps` — and only the message verb tells them apart, so a mouse click reads `Click: CheckoutButton` rather than `Tap:`. Two tests cover the click path, and one covers an `$autocapture` event with no `$event_type`.

Full React Native suite passes (712), core passes (1037), lint and typecheck clean, references regenerated.

---

**Still open: whether the "user changed" step lands here or follows.** That is the only thing left, and it is a scope call rather than a code question, so I have not guessed at it.

Both open threads point at the same missing piece. @ioannisj asked to keep steps across `reset()` so a login or logout flow keeps its context, which I agree with and reverted to. The review bot's remaining thread is open on the flip side of exactly that: after `reset()` the timeline gives no sign the identity changed, so a reader can attribute the previous user's steps to the next one. A "user changed" step — @hpouillot's own suggestion from the original design thread — resolves both at once. It keeps the steps, and it marks the boundary so nobody misreads them.

Concretely I would add it as a fourth opt-in signal, `identity`, recording a step with `$type: 'identity'` and the message `User changed` at `reset()`. No distinct ID in the payload, since carrying one would reintroduce the concern the step exists to answer.

Happy either way — say the word and I add it here, or I ship this as automatic-signals-only and open a follow-up. This is otherwise ready.

.changeset/automatic-exception-steps-react-native.md
jakesciotto: Following up on this one, since it is the last thing holding the PR.

Your "I agree that a 'user changed' step would be valuable here as well" reads to me like a yes, but I do not want to assume it means *in this PR*. Shape I would add: a fourth opt-in signal `identity`, one step with `$type: 'identity'` and the message `User changed`, recorded at `reset()`, carrying no distinct ID.

That also closes the bot's open thread, which is the same question from the other side — it flagged that after `reset()` nothing in the timeline says the identity changed, so the previous user's steps read as the next user's. Keeping the steps and marking the boundary answers both.

In or out of this PR? Either is fine by me. @hpouillot, it was your idea originally, so shout if the shape is wrong.

jakesciotto: taking the quiet as no objection, so i'm adding the identity step here rather than splitting it out. it's small enough that it's easier to judge in code than in prose, and it's one revert if you'd rather this pr stayed narrow.

shape: a fourth opt-in signal `identity`, off unless enabled like the other three. on `reset()` it records one step, `$type: 'identity'`, message `User changed`, no distinct id and no properties, emitted before `super.reset()` so it sits at the boundary rather than after it.

net effect: the buffer still survives `reset()`, so rn stays consistent with web, ios, android and flutter, and the timeline now marks where the identity changed instead of leaving the reader to infer it.

@hpouillot this was your idea in the original thread, so shout if the shape is wrong. otherwise i'll push it and this pr is done.

packages/react-native/src/posthog-rn.ts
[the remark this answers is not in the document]
jakesciotto: this will be closed by the identity step rather than by clearing the buffer. your concern was that after `reset()` nothing marks the identity change, so the previous user's steps read as the next user's. an explicit `User changed` step answers that while keeping the login and logout context that clearing would destroy. discussion is in the changeset thread above.

packages/node/references/posthog-node-references-latest.json
jakesciotto: addressed in 4042b7d, the rn config is specialised out of the node references. fine to resolve?

packages/core/src/error-tracking/exception-steps.ts
jakesciotto: addressed in 4042b7d, `$level` is removed. fine to resolve?

.changeset/automatic-exception-steps-react-native.md
jakesciotto: pushed in 57fe205. `identity` is opt-in like the other three, `automatic: true` turns it on with them. the existing reset test now asserts the boundary marker rather than a bare buffer survival, and there is a second test covering reset with identity off.

jakesciotto:
Bumping this one before the stale bot gets it.

Everything from the review round is addressed as of `57fe205`: `identity` is opt-in like the other three signals, `$level` is removed, the RN config is specialised out of the node references, and the reset test asserts the `User changed` boundary marker rather than bare buffer survival. CI is green.

@hpouillot @ioannisj — I think your threads are answered, but happy to keep talking if any of it still reads wrong to you.

@ablaszkiewicz @cat-ph — a review whenever you have a spare moment would be much appreciated. No rush, and shout if you'd rather someone else take it.

Removing the stale label so this stays open in the meantime.

PR #4578, opened 2026-08-20
fix: drop events when beforeSend hooks throw

marandaneto:
## Problem

When a `beforeSend` or `before_send` hook throws, the browser, core, Node, and Convex clients could continue with the previous event value or rely on an outer rejected-promise handler. This could send an event that the hook was expected to inspect or redact, or log only a bare stack trace.

## Changes

- Catch errors from each before-send hook and log them through the SDK logger.
- Stop the remaining hook chain and drop the event when a hook throws.
- Keep passing each successful hook's returned event to the next hook.
- Add regression tests that verify thrown hooks do not enqueue, schedule, or send events.
- Add direct release notes for Node, React Native, Next, and Nuxt because they expose or route through the affected clients.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [x] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

`@posthog/core` also receives a patch release and is not listed in the checklist above.

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

The Pi coding agent used the PR and autoreview skills. The implementation follows the shared fail-closed contract: a thrown hook is contained and logged, later hooks do not run, and the event is not queued or sent. Focused tests and production builds were run for Core, Browser, Convex, and Node. Reviewer feedback prompted the Node coverage and expanded release notes. The final isolated autoreview found no actionable issues.

### DIFF
--- a/packages/core/src/posthog-core.ts
+++ b/packages/core/src/posthog-core.ts
@@ -1739,6 +1739,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         }
       } catch (e) {
         this._logger.error(`Error in before_send function for event '${captureEvent.event}':`, e)
+        return null
       }
     }
 
--- a/.changeset/fail-closed-before-send.md
+++ b/.changeset/fail-closed-before-send.md
@@ -1,0 +1,7 @@
+---
+'@posthog/core': patch
+'posthog-js': patch
+'@posthog/convex': patch
+---
+
+Drop events when a before-send hook throws instead of sending the unmodified event.
--- a/.changeset/fail-closed-before-send.md
+++ b/.changeset/fail-closed-before-send.md
@@ -2,6 +2,10 @@
 '@posthog/core': patch
 'posthog-js': patch
 '@posthog/convex': patch
+'posthog-node': patch
+'posthog-react-native': patch
+'@posthog/next': patch
+'@posthog/nuxt': patch
 ---
 
 Drop events when a before-send hook throws instead of sending the unmodified event.
--- a/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
+++ b/packages/browser/src/__tests__/posthog-core.beforeSend.test.ts
@@ -120,20 +120,20 @@ describe('posthog core - before send', () => {
     it('can take an array of fns', () => {
         const posthog = posthogWith({
             before_send: [
-                (cr) => {
-                    cr.properties = { ...cr.properties, edited_one: true }
-                    return cr
-                },
+                (cr) => ({ ...cr, properties: { ...cr.properties, edited_one: true } }),
                 (cr) => {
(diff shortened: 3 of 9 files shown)


packages/core/src/posthog-core.ts
```diff
@@ -1739,6 +1739,7 @@ export abstract class PostHogCore extends PostHogCoreStateless {
         }
       } catch (e) {
         this._logger.error(`Error in before_send function for event '${captureEvent.event}':`, e)
+        return null
```
turnipdabeets, line 1742: [suggestion] `packages/node/src/client.ts` has its own `_runBeforeSend` with no try/catch and didn't get this treatment.

It does fail closed today, but only because `_prepareEventMessage` is `async` and `_sendPreparedEvent`'s trailing `.catch` swallows the rejection before enqueue. Two consequences: a user hitting this gets a bare stack trace via `console.error` rather than `Error in before_send function for event '…'` through `_logger`, and nothing pins the behavior if that call ever moves out of the promise chain.

Could we add the same catch plus a regression test there? That would be 5 of the 6 `_runBeforeSend` copies aligned — `core/logs` and `core/metrics` already were. Happy for it to be a follow-up since node isn't actually leaking events.

turnipdabeets (approved):
Approving. Catching per-hook rather than around the whole loop is the right call — it keeps the existing null/empty-properties logging intact instead of collapsing every outcome into one handler.

[question] Before merge, should `@posthog/react-native` be ticked under affected libraries? It extends `PostHogCore`, so it picks this change up via the `@posthog/core` patch, but its changelog will only read "Updated dependencies" — RN users with a throwing hook get a behavior change with nothing pointing at it. Same question for `@posthog/next` / `@posthog/nuxt` if they route capture through core. Flagging it pre-merge since release notes aren't fixable after the fact.

One suggestion inline about `packages/node`, fine as a follow-up.

packages/core/src/posthog-core.ts
marandaneto: Good catch! Fixed in 7d7d9c25b — Node now catches and logs each thrown before_send hook, stops the chain, and drops the event. The regression test verifies no later hook or transport runs.

marandaneto:
@turnipdabeets I also expanded the changeset in 7d7d9c25b to include `posthog-node`, `posthog-react-native`, `@posthog/next`, and `@posthog/nuxt`, so each affected public package gets a direct release note.

PR #4579, opened 2026-08-20
feat(node): distributed tracing spans

turnipdabeets:
## Problem

Developers instrumenting a Node service with PostHog have no way to record spans. They can send events, logs and metrics, but nothing that shows *where time went* in a request or how work fans out across services. Pointing an OpenTelemetry SDK at PostHog works, but it means adding an OTel dependency and wiring the join to PostHog identity by hand — so traces end up disconnected from the person and session they belong to.

Part of a stacked series adding first-class tracing to the JS SDKs (Q3 Goal 4). Scoped to be releasable on its own, one commit, on top of the shared OTLP attribute encoder from #4708 (merged).

## Changes

Adds `startSpan`, `withSpan` and `getActiveSpan` to `posthog-node`, behind a new `traces` client option. Spans are encoded as OpenTelemetry-shaped OTLP JSON and POSTed to `/i/v1/traces` — **without** an OpenTelemetry dependency.

```ts
const posthog = new PostHog('phc_...', { traces: { serviceName: 'checkout-api' } })

await posthog.withSpan('POST /checkout', { parent: req.headers.traceparent }, async (span) => {
    span.setAttribute('plan', user.plan)
    return processOrder()
})
```

- **Off until configured.** No `traces` option means no spans. Every span API still returns a working (inert) handle, so calling code never branches on whether tracing is on.
- **Identity join.** Spans started inside a request context carry `posthogDistinctId` and `sessionId`, taken from `withContext` or the existing Express/NestJS middleware. This is what makes a trace reachable from a person or session.
- **W3C trace context.** `parent` accepts an inbound `traceparent` string to continue a remote trace; `span.traceparent()` gives you the header to propagate onward. `tracestate` is preserved opaquely.
- **Nesting across `await`** on Node, via `AsyncLocalStorage` injected at the Node entrypoint. Core stays runtime-agnostic (no `node:async_hooks`), so the edge build and future browser/RN hosts work off the same engine.
- **`flush()` drains spans**, concurrently with the event queue. Serverless handlers call `flush()`, not `shutdown()`, so leaving spans on their own timer would silently lose them for the most common `posthog-node` deployment. A span ending also refreshes the `waitUntil` cycle, so a handler that only traces still holds its invocation open.
- **Bounded live spans.** `maxLiveSpans` (default 10000) caps how many spans may be open at once and `maxSpanAgeMs` (default one hour) stops accounting for one that stays open longer, so code that starts spans and never ends them cannot grow the SDK's bookkeeping without limit. At the cap `startSpan` returns an inert handle; an evicted span is never exported. Both drops go through the existing span-drop warning.
- **Bounded retries.** A batch the endpoint keeps refusing is retried with exponential backoff capped at 30s, then dropped after 8 consecutive failures, so a stuck batch cannot pin the queue while fresher spans are refused at the cap.

### Reviewer notes

- **Edge runtime** — the edge build restores the active span when a callback *returns its promise*, not when it settles, so spans started after an `await` there begin a new trace. Pass `parent` explicitly to nest them. This is the documented browser limitation too.
- **`IPostHog` gains three required members** (`startSpan`, `withSpan`, `getActiveSpan`). Consumers using it as a type annotation are unaffected; anyone *implementing* it (hand-written test doubles, DI wrappers) will get a compile error. Same shape as `metrics` in #4117, which shipped under a `minor` bump.
- **JSON, not protobuf — and this is conformant, not a divergence.** The spec makes protobuf canonical only "where the encoder adds no meaningful dependency or binary-size cost". A protobuf encoder is a real dependency for a zero-dependency package, so the condition does not fire and the SHOULD-send-JSON branch applies. It also matches what the shipped logs and metrics pipelines already send.
- **A 200 is not proof of correct configuration** — the ingestion endpoint accepts a well-formed but unknown project key with a 200 and drops the spans.
- **The shared OTLP attribute encoder moved to #4708**, which has merged. Spans reuse it rather than adding a third copy. Span batches likewise go through `_sendOtlpBatch` (from #4623) rather than a third copy of the retry policy — the bearer-auth branch is reintroduced there for traces.
- **Docs** — the Node library page section is drafted in [PostHog/posthog.com#19837](https://github.com/PostHog/posthog.com/pull/19837), held as a draft until this PR merges and ships. It documents both `flush()` and `shutdown()` as span drains.
- **Live spans are bounded without retaining spans.** The obvious way to implement the spec's `Live span bounds` requirement is a registry of live span objects, which would keep every leaked handle alive for the length of the age bound. Instead the engine keeps a `Map` of span id to monotonic start — ids and numbers, never the span — so a handle the caller drops is still collected like any other object, and the count bound can be generous because a slot costs tens of bytes rather than a whole span. `packages/core/src/traces/live-spans.spec.ts` keeps its `WeakRef` probe, now as the guard that stops a later change from turning this into a registry of spans. Eviction is lazy at `startSpan`, and sweeps *before* reading the bound, so a process that has leaked its way to the cap recovers on the first call after the leaks age out rather than losing tracing for the rest of its life.
- **`Retry-After` is not honoured** — the backoff is purely exponential, capped at 30s. This is a shared gap with the logs and metrics senders rather than something this PR introduces, and the service does not emit 429 yet. Tracked separately.
- **`beforeSpanSend` and the per-span caps ship here.** They were written as #4584 and have since been folded into this branch, so this PR's public `traces` surface is `serviceName`, `serviceVersion`, `environment`, `resourceAttributes`, `flushIntervalMs`, `maxExportBatchSize`, `maxQueueSize`, `beforeSpanSend`, `maxAttributesPerSpan`, `maxEventsPerSpan`, `maxAttributeValueLength`, `maxLiveSpans` and `maxSpanAgeMs`. The per-event attribute cap stays internal.
- **Deliberately deferred** to later PRs in this stack: OpenTelemetry soft-detect, browser host, and log/exception correlation.

### Verification

End-to-end against a real project (381971), twice: first a scripted run, then a **real Express app** using `setupExpressRequestContext` with two routes, a genuine HTTP hop between them, and an error path. Spans read back out of the product assembled correctly — `POST /checkout` root with `db.query` and `http.post payments` beneath it, `GET /payments/charge` continued across the hop via `traceparent`, and a separate trace for the thrown route carrying `status_code: 2`. `posthogDistinctId` / `sessionId` came from the `X-POSTHOG-*` request headers on every in-context span, with no manual plumbing.

Also exercised against production ingestion: an oversized batch really does 413 (`11.5 MB → 413`, halved to `5.75 MB → 200` twice, all spans queryable afterwards), and 50k spans push through at ~310k/sec with 204 KB average batches — the first real evidence that the 512-span default sits under the body cap. With the endpoint down, 200k spans cost exactly **one send per flush pass** rather than one per span — the transport still retries within a send, so a pass can make several HTTP attempts.

`packages/core` 1305 pass (62 suites), `packages/node` 1000 pass (35 suites) — identical under the edge runtime environment — lint clean, public API references regenerated.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

Backwards compatibility caveat: all runtime surface is additive, but `IPostHog` gains three required members — a compile-time break for implementors only, matching the `metrics` precedent. Bundle: the traces module is core-resident and the browser is not wired up in this PR.

### If releasing new changes

## 🤖 Agent context

Built with Claude Code, directed by @turnipdabeets, from the merged `traces` capability spec in `sdk-specs`.

Decisions worth flagging for review:

- **Top-level methods over a `posthog.traces.*` namespace** — matches how the OTel API reads and keeps the common call short.
- **Vertical slice over a horizontal one.** An earlier plan split encoding, transport and API into separate PRs; that was rejected because each merged PR is releasable, and a released half-feature isn't useful to anyone. Each PR in this stack is a feature a developer could adopt.
- **`OtlpSpanKeyValue` is an alias** of the logs/metrics `OtlpKeyValue` rather than a separate declaration — one encoder produces all three payloads. The span-flavoured name stays so the span types read as span types.
- **Wire types live in `@posthog/types` only**, matching how the logs and metrics wire types are already organised.

Put through five rounds of independent fresh-context review, each reviewer required to demonstrate a finding by running it rather than by reading. Bugs they caught, all fixed here and each now pinned by a test that fails when the fix is reverted:

- a background flush trigger per span end, stacking an unbounded number of concurrent drain loops on a busy service (measured: 193 outstanding, 14× CPU degradation, linear heap growth)
- the depth-based flush cancelling its own retry backoff, so an outage cost ~1,500 requests instead of one
- span names, event names, `status.message` and `traceState` skipping the sanitiser attributes go through, so one lone surrogate would 400 an entire 512-span batch
- the retry budget charging fresh spans to a batch's already-spent budget
- a throwing accessor in `resourceAttributes` rethrowing on every flush, exporting nothing, forever
- the `waitUntil` serverless path having no traces coverage at all, and a span-only handler never registering with it

Several of those were introduced by earlier rounds' own fixes, which is why the loop ran as long as it did.


packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,548 @@
...
+          this._queue.splice(0, size)
+          remaining -= size
+          removed += size
+          this._headBatchFailures = 0
+          continue
+        }
+
+        const outcome = await this._instance._sendTracesBatch(
+          buildOtlpTracesPayload(spans, resourceAttributes, scopeName, scopeVersion, this._logger)
+        )
```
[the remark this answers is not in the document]
turnipdabeets: Fixed in the latest push — good catch, this was real.

`_flushInner` now discards the queue when `isDisabled` or `optedOut` is set, before anything reaches `_sendTracesBatch`. Previously consent was only checked at `startSpan` and at `_onSpanEnd`, so a span queued while consent was granted and flushed after `optOut()` would still export — carrying `posthogDistinctId` and `sessionId` with it. `_sendOtlpBatch` checks `disabled` but not `optedOut`, so nothing downstream caught it either.

Two regression tests added in `packages/core/src/traces/index.spec.ts` (`consent withdrawn after a span is queued`), verified to fail when the guard is removed.

marandaneto:
@PostHog/team-apm i think it makes sense for you all to give an initial look to make sure this works well within the product/ui/ingestion

packages/core/src/traces/span.ts
```diff
@@ -0,0 +1,250 @@
...
+  setStatus(): this {
+    return this
+  }
+  recordException(): this {
+    return this
+  }
+  updateName(): this {
+    return this
+  }
+  traceparent(): string | null {
```
jonmcwest, line 438: **The SDK drops trace context when tracing is off. This severs distributed traces at services that do not have tracing enabled yet.**

`startSpan` returns the shared `NOOP_SPAN` before it parses `parent` (`packages/core/src/traces/index.ts:107`, `packages/node/src/client.ts:683` and `:723`). `NoopSpan.traceparent()` returns `null`. Thus a service in the middle of a traced chain, with `traces` unconfigured, forwards nothing. The propagation example in the `withSpan` docstring (`getActiveSpan()?.traceparent()`) sends no header there. Downstream services then start fresh traces.

This diverges from the OTel spec, which is normative here. With no SDK installed, "the API MUST return a non-recording `Span` with the `SpanContext` in the parent `Context`" (trace/api.md, "Behavior of the API in the absence of an installed SDK"). The propagated context then "will be propagated through to any child span and ultimately also `Inject`". The spec makes this the one exception to no-op behavior, exactly so that a pass-through service keeps the chain intact.

The `NoopSpan` comment gives this reason: `null` makes sure the SDK cannot propagate "an id that was never recorded". Pass-through does not do that. When the SDK forwards the inbound header unchanged, it propagates the span id of the upstream caller, and upstream recorded that id. The SDK invents no ids.

Suggestion: when `parent` is a valid traceparent string but tracing is off, return a per-call inert span. Make its `traceparent()` echo the inbound header, the tracestate, and the original flags. Activate it in `withSpan` so that `getActiveSpan()` works. The shared `NOOP_SPAN` can stay for the no-parent case.

If you keep the current behavior on purpose, add a test that passes `parent` on the disabled path, and add a docs note. Today no test covers this case.

This does not block an experimental MVP. A later move to pass-through is compatible with the documented `traceparent ? ... : {}` guard. Decide before the API becomes stable.

packages/core/src/traces/traceparent.ts
```diff
@@ -0,0 +1,82 @@
...
+    return undefined
+  }
+  return { traceId, spanId }
+}
+
+/**
+ * Builds the `traceparent` header value for a span. The sampled flag is always
+ * set, because a span we exported is by definition recorded.
+ */
+export function formatTraceparent(traceId: string, spanId: string): string {
```
jonmcwest: **The SDK drops the inbound sampled flag and always sends `01` outbound.**

The spec position first: this is legal. W3C Trace Context lists "Update sampled" as a permitted mutation when parent-id changes, and it changes here. The spec also says a definitive recording decision "SHOULD be reflected in the sampled flag". PostHog records everything, so `01` is truthful. OTel with a bare `AlwaysOnSampler` does the same thing.

The cost is interop with parent-based samplers from other vendors. Service A samples a trace out (`00`). A PostHog-traced service B continues it and propagates `01`. A downstream OTel service C with the default `ParentBased` sampler then records a trace that its own head sampler rejected. Its backend gets paid-for fragments with no root.

Suggestion, as a follow-up and not a blocker: use the OTel `RECORD_ONLY` shape for this case. Store the inbound flag on `RemoteSpanContext`. Record and export the span as now, but propagate the inbound byte in `traceparent()` and set OTLP `flags` from it. That keeps the header and the wire consistent. First make sure that ingestion does not treat `flags=0` spans differently.

At minimum, document the current behavior. It will surprise users who integrate with head-sampled OTel fleets.

packages/core/src/traces/span.ts
```diff
@@ -0,0 +1,250 @@
...
+
+// Typed as `Span`, not `NoopSpan`: the class's methods take no parameters, so the
+// concrete type would reject calls the interface allows.
+export const NOOP_SPAN: Span = /* @__PURE__ */ new NoopSpan()
+
+/**
+ * Extracts the OTel `exception.type` / `exception.message` pair from whatever was
+ * thrown. Anything can be thrown in JS, so non-Errors are described by type.
+ */
+export function describeError(error: unknown): { type: string; message: string } {
```
jonmcwest: `recordException` and the `withSpan` error path record only `exception.type` and `exception.message`. Do you drop `error.stack` on purpose? OTel semconv marks `exception.stacktrace` as Recommended, and OTel JS captures it in `recordException`. People usually debug from the stack.

There is a case for a delay. No `beforeSpanSend` hook or attribute cap exists yet (both deferred to #4584), so users could not scrub or bound stacks. A deep async stack is a few KB per span. Also, if a later release adds stacks silently, it changes what leaves customer servers. Such a change must be visible either way.

Suggestion: if this is a deferral, say so in the `recordException` doc in `packages/types/src/traces.ts` and in the PR body. Land stack capture with #4584, so that the caps and the scrub hook ship first. If it is not deliberate, `describeError` can read `error.stack` behind the existing try/catch today.

packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,612 @@
...
+          remaining -= size
+          removed += size
+          // Ramp back toward the configured max after a 413 shrink.
+          if (this._maxExportBatchSize < this._config.maxExportBatchSize) {
+            this._maxExportBatchSize++
+          }
+          continue
+        }
+
+        if (outcome.kind === 'too-large') {
```
jonmcwest, line 832: A span with one multi-MB attribute takes an expensive path here. The 413 halving retries the same head batch. Thus the SDK uploads the oversized body up to 11 times before it isolates and drops the span. To drain a full 512-span queue around it takes 35 to 47 POSTs (simulated against this exact loop). The sticky shrink then leaves `_maxExportBatchSize` near 22, and the +1 ramp needs about 480 successful batches to recover.

The OTel default is no better on the value itself. `AttributeValueLengthLimit` defaults to Infinity, and the OTel exporter drops the whole batch on failure. This path drops only the poison span, which is the right call. But #4584 adds count caps (128 attributes, 128 events) and no value-length cap, so a single huge attribute still lands here after it merges.

Ask: add a client-side size check before the POST. `_sendOtlpBatch` already serializes the payload, and the client knows the ~2MB server cap (see the comment in `traces-defaults.ts`). Treat a single span whose serialized size exceeds the cap as too-large, without a send. That removes every redundant upload and most of the halving churn. It can ship as a small follow-up.

packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,612 @@
...
+/** `instanceof` and property access both throw on a hostile proxy; `startSpan` must not. */
+function isOwnSpan(value: unknown): value is PostHogSpan {
+  try {
+    return value instanceof PostHogSpan
+  } catch {
+    return false
+  }
+}
+
+function looksLikeSpan(value: unknown): boolean {
```
jonmcwest, line 54: Nit: an OTel span passed as `parent` has no `traceparent` method. It fails `looksLikeSpan`, and the SDK drops it silently with "Ignoring an unusable span parent". The new span then parents to the active PostHog span, or to a new root. That breaks the trace the caller expected. Inside `withSpan` this is wrong parentage, not just a fresh trace. A follow-up could duck-type `spanContext()` and adopt `traceId`/`spanId` after `isValidTraceId`/`isValidSpanId`. That fits the deferred soft-detect PR. The inline comment "Not a span at all" is also inaccurate for this case.

packages/core/src/traces/otlp.ts
```diff
@@ -0,0 +1,171 @@
...
+
+export function buildOtlpSpan(record: SpanRecord, logger?: Logger): OtlpSpan {
+  const span: OtlpSpan = {
+    traceId: record.traceId,
+    spanId: record.spanId,
+    name: wireString(record.name),
+    kind: spanKindToOtlp(record.kind),
+    startTimeUnixNano: msToUnixNanoString(record.startTime),
+    endTimeUnixNano: msToUnixNanoString(record.endTime),
+    flags: TRACE_FLAGS_SAMPLED,
```
jonmcwest: Optional nit: `flags` is always 1, so bits 8-9 (parent-remoteness, `0x100`/`0x200`) stay unset and the row records "unknown". OTel SDKs set these bits, and this SDK knows the answer: a traceparent-string parent is remote, a handle parent is local. The server stores `flags` verbatim but reads nothing from it today, so this is optional. If you set the bits now, you avoid a gap that no backfill can repair, in case service-entry detection ever uses them.

packages/core/src/traces/otlp.ts
```diff
@@ -0,0 +1,171 @@
...
+  }
+  if (record.parentSpanId) {
+    span.parentSpanId = record.parentSpanId
+  }
+  if (record.traceState) {
+    span.traceState = wireString(record.traceState)
+  }
+  const attributes = toOtlpKeyValueList(record.attributes, logger)
```
jonmcwest, line 123: Nit: an empty attribute key passes `assignUserAttributes` and `encodeKeyValueList` unexamined and ships as `{ "key": "" }`. The OTel spec requires non-empty keys. The server stores the key verbatim, so it surfaces as a nameless attribute in filters. Skip empty keys with a debug log in `encodeKeyValueList` (`packages/core/src/utils/otlp-any-value.ts`, outside this diff). The same fix covers logs and metrics, because all three signals share this encoder.

jonmcwest (commented):
I could not review the full diff line by line, so I compared the design and the trace-context behavior against the OTel specification and OTel JS, and validated each comment against the code (at 5f7b2ff, re-checked at b07c9eb).

The implementation matches or exceeds the reference behavior on clocks, id generation, batching defaults, 413 handling, backpressure, and hostile input. No comment below blocks the MVP. Comments 1 and 2 are trace-context design questions, and comment 1 deserves an explicit decision before the API stabilizes. The rest are follow-ups and nits.

packages/core/src/traces/span.ts
turnipdabeets: Addressed in bf5e7af75 — inert spans now pass the inbound `traceparent` through, and `withSpan` activates them so `getActiveSpan()` works.

packages/core/src/traces/traceparent.ts
turnipdabeets: Addressed in a78b6c3b5, in #4728. Confirmed ingestion stores `flags` verbatim and nothing reads it, so the sampled-out byte is safe to propagate.

packages/core/src/traces/span.ts
turnipdabeets: Addressed in e6b77d308, in #4584 — stacks land there, after the caps and the scrub hook.

packages/core/src/traces/index.ts
turnipdabeets: Half addressed: f587e482d in #4584 adds `maxAttributeValueLength` (8192), which bounds the value that causes this. The pre-send size check still needs a number — the cap in the code comment says ~2 MB but this PR measured 5.75 MB accepted and 11.5 MB rejected, so I don't want to guess it. Tracking separately.

packages/core/src/traces/index.ts
turnipdabeets: Comment fixed in bf5e7af75 — it now says a span from another tracer exposes `spanContext()` instead. The duck-typing itself stays with the soft-detect PR.

packages/core/src/traces/otlp.ts
turnipdabeets: Addressed in a78b6c3b5, in #4728 — `0x100` always, `0x200` for a header parent.

packages/core/src/traces/otlp.ts
turnipdabeets: Addressed in 1e8a15144, in #4727 — fixed in the shared encoder, so logs and metrics get it too. Off main rather than this stack.

packages/core/src/traces/index.ts
turnipdabeets: Second half now in 1f1c33f0d, in #4726 — found the cap in the monorepo (`MAX_REQUEST_BODY_SIZE_BYTES`, 2 MB, applied to the compressed body), so a batch over it is reported too-large without being sent.

packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,618 @@
...
+        if (!spans.length) {
+          // Nothing survived encoding; drop the batch rather than re-encoding it forever.
+          this._queue.splice(0, size)
+          remaining -= size
+          removed += size
+          this._headBatchFailures = 0
+          continue
+        }
+
+        const outcome = await this._instance._sendTracesBatch(
```
dustinbyrne, line 810: **[P1] Recheck consent before each batch send.** The guard above only protects the start of the drain. If consent changes while an earlier batch is in flight, the loop can begin exporting the next identity-linked batch afterward. I reproduced this through the public Node API during retry recovery: A was in flight, `disable()` completed, then B started a third trace request. Please recheck at each loop iteration and discard the remaining queue, with a regression covering retry → queued second batch → opt-out during the first in-flight retry.

dustinbyrne (approved):
Approved. The tracing-team feedback is addressed or appropriately isolated in follow-up PRs, and the implementation, focused tests, published build, edge separation, and release scope look good. I left one consent-handling follow-up inline.

Human-driven, agent-assisted review.

packages/core/src/traces/index.ts
dustinbyrne: fwiw, this is documented in sdk-specs, but i don't 100% agree with the necessity here because we're talking about very narrow windows and edge cases / data collected while consent was given

lately agents have a strong tendency to introduce a lot of complexity just to cover cases like this, and i'm not convinced that it's worth incurring the complexity of every 0.001% edge case

turnipdabeets: Fixed in 8997a166b, in #4584 (stacked on this branch, so it ships in the same release — leaving this PR alone since it's already approved).

`_flushInner` now re-checks `isDisabled`/`optedOut` at the top of every loop iteration, not just before the drain, and discards the rest of the queue if either has flipped. It's the existing discard block moved into `_discardQueueIfConsentWithdrawn()` and called twice, so no new machinery. Two regression tests: a backlog built up during an outage, then `optOut()` (and separately `disable()`) landing while the retried head batch is in flight — both verified to fail with the loop-body check removed.

On your reproduction, one correction: `disable()` mid-drain was already safe before this. `_sendOtlpBatch` short-circuits on `this.disabled` and returns `{ kind: 'fatal' }`, which the loop treats as non-retriable and splices away, so no HTTP request leaves — what you saw was the loop calling `_sendTracesBatch`, not a POST going out. The genuine hole was `optOut()`: `_sendOtlpBatch` checks `disabled` but not `optedOut`, so batches behind the in-flight one really did export with `posthogDistinctId` and `sessionId` on them.

And I take your point on the complexity ceiling — agreed in general. Took this one because it's the same guard called from a second place rather than a new mechanism, and because it's identity data.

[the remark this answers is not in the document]
turnipdabeets: Fixed in 8997a166b, in #4584 (stacked on this branch, so it ships in the same release).

`_flushInner` re-checks `isDisabled`/`optedOut` at the top of every loop iteration now and discards the remaining queue if either has flipped, rather than only gating the start of the drain. Two regression tests cover it: a backlog built during an outage, then `optOut()` (and separately `disable()`) landing while the retried head batch is in flight.

One correction to the report: `disable()` was already covered. `_sendOtlpBatch` short-circuits on `disabled` and returns `{ kind: 'fatal' }`, so no request goes out. The real gap was `optOut()`, which nothing downstream checked.

arnohillen:
`packages/core/src/traces/traceparent.ts:41` rejects only version `ff` and the regex at `:10` allows a trailing `(-.*)?`, so a version-`00` traceparent longer than the 55 characters W3C requires is accepted and the trace continues where a conformant peer would restart it.

packages/core/src/traces/index.ts
turnipdabeets: Closing the loop on this — it shipped in #4726, but two things you should know, because one of them means your repro no longer reproduces.

**The 2 MB you cited isn't the server's cap.** The `traces-defaults.ts` comment you pointed at said "the server's 2 MB body cap", and that comment was wrong — it came from the fallback in `capture-logs/src/config.rs`, not from what's deployed. `charts` `shared/capture-logs/common.yaml:195` sets `MAX_REQUEST_BODY_SIZE_BYTES: "10485760"`, so the real limit is **10 MiB**, with no lower cap at the ingress. I've fixed that comment and the matching one in `logs/types.ts`.

So a 3 MB span is simply accepted today — no 413, no halving, nothing to save. The first cut of the check used 2 MB and would have *dropped* records the endpoint accepts: measured, one 3 MiB log record among two went from 2 of 2 delivered to 1 of 2, and a 25k-series metrics window was discarded whole with zero requests, since metrics has no shrink path. @dustinbyrne and @ioannisj caught it. It's 10 MiB now.

**Your mechanism is intact, just at a higher threshold.** Re-measured with one 12 MiB record among 511 healthy ones:

| | requests | uploaded |
|---|---|---|
| traces, reactive only | 42 | 120.3 MiB |
| traces, with the check | 32 | 0.1 MiB |
| logs, reactive only | 16 | 72.0 MiB |
| logs, with the check | 10 | 0.0 MiB |

Your sticky-shrink number reproduces exactly, incidentally: `maxExportBatchSize` goes 512 → 33 and needs **479** healthy batches to recover.

**On the root cause you flagged** — "#4584 adds count caps and no value-length cap" — that's since been addressed there: `maxAttributeValueLength`, default 8192, applied to span attributes, event attributes, status messages, resource attributes and nested strings.

But it still doesn't bound a span, which I think is worth your eye: `addEvent` caps the event count and truncates each value, and nothing caps how many attributes an event carries. Worst case is 128 events × unbounded × 8 KB. Even adding OTel's `maxAttributesPerEvent: 128` leaves 128 × 128 × 8192 ≈ 128 MiB. I've filed that on #4584 ([here](https://github.com/PostHog/posthog-js/pull/4584#discussion_r3936881584)).

Which is the argument for keeping what you asked for: count caps bound a span's shape, only a byte measurement bounds its size, and the batch boundary is the one place that's affordable — 3% of send cost on node, zero allocation.

packages/core/src/traces/span.ts
```diff
@@ -0,0 +1,739 @@
...
+  } catch {
+    return {}
+  }
+}
+
+/**
+ * The handle to return when a span cannot be recorded: a pass-through when the
+ * caller supplied a usable `parent` header, the shared no-op otherwise.
+ */
+export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }): Span {
```
jonmcwest: **Follow-up: preserve the implicit pass-through parent.**

A child without an explicit `parent` ignores an active `PassThroughSpan`: tracing off loses the context; tracing on can create an unrelated trace.

OTel's no-op rule requires preserving implicit parent context and recommends reusing an already non-recording parent. Reusing the handle on the tracing-on path is a PostHog fallback choice, not that rule's requirement. Cover both paths, including across `await`. Non-blocking for the MVP; settle before API stability.

Sources: [OTel propagation](https://github.com/open-telemetry/opentelemetry-specification/blob/c09e58480b03dfef67a2fe81b1ce2f462daf5b4c/specification/trace/api.md#behavior-of-the-api-in-the-absence-of-an-installed-sdk) · [implementation](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/traces/span.ts#L494-L506).

*Agent-assisted review.*

packages/core/src/traces/index.ts
jonmcwest: **Release dependency: include the pre-send size check.**

Thanks, the re-measurement resolves my earlier concern. Please hold the first traces release until [#4726](https://github.com/PostHog/posthog-js/pull/4726) is included: this branch only reacts to a 413 after sending an oversized payload. Merge order is not the concern; releasing between the merges is.

That follow-up also corrects the config comment: the string-length cap does not bound an entire span. No additional change needed here.

Sources: [OTLP request limits](https://github.com/open-telemetry/opentelemetry-proto/blob/09f8394ecb171e889029fcd0036a3ab856c9b801/docs/specification.md#otlphttp-request) · [pre-send check](https://github.com/PostHog/posthog-js/blob/030c294947679328cbc026a545188ac89efc3a78/packages/core/src/posthog-core-stateless.ts#L1738-L1751).

*Agent-assisted review.*

packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,937 @@
...
+  /**
+   * Starts a span without making it active. Always returns a handle — an inert
+   * one when tracing cannot run — so calling code never branches.
+   */
+  startSpan(name: string, options?: StartSpanOptions): Span {
+    if (this._instance.isDisabled || this._instance.optedOut) {
+      return inertSpan(options)
+    }
+
+    const explicitParent = options?.parent
```
jonmcwest: **Correction: foreign OTel parents are optional interoperability work, not a docs bug.**

`parent` accepts this SDK's `Span | string`. A foreign handle with `traceparent()` takes the inert path as documented; an OTel `spanContext()`-only object is outside that type.

I withdraw the documentation-fix request. OTel's tracing API takes a `Context` parent, not a raw `Span`, so accepting foreign span objects directly is a PostHog API choice. No merge action needed here.

Sources: [current parent contract](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/types/src/traces.ts#L78-L88) · [OTel span creation](https://github.com/open-telemetry/opentelemetry-specification/blob/c09e58480b03dfef67a2fe81b1ce2f462daf5b4c/specification/trace/api.md#span-creation) · [foreign-handle handling](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/traces/index.ts#L197-L209).

*Agent-assisted review.*

packages/core/src/traces/span.ts
```diff
@@ -0,0 +1,739 @@
...
+      // throwing getter costs its own key — then write each through the cap.
+      const safe: SpanAttributes = assignUserAttributes({}, attributes)
+      for (const key of Object.keys(safe)) {
+        this._writeAttribute(key, safe[key])
+      }
+    }
+    return this
+  }
+
+  addEvent(name: string, attributes?: SpanAttributes, timestamp?: SpanTimeInput): this {
```
jonmcwest, line 184: **Covered by the follow-up; no extra change here.**

OTel makes the per-event attribute limit optional, with a default of 128 when provided. This is not a missing mandatory limit.

On this branch, the encoder drops excess event attributes without a per-event dropped count. [#4792](https://github.com/PostHog/posthog-js/pull/4792) adds the cap and counter. Keeping this note to connect the concern to that follow-up. Non-blocking.

Sources: [OTel span limits](https://github.com/open-telemetry/opentelemetry-specification/blob/c09e58480b03dfef67a2fe81b1ce2f462daf5b4c/specification/trace/sdk.md#span-limits) · [event handling](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/traces/span.ts#L175-L193).

*Agent-assisted review.*

packages/core/src/index.ts
```diff
@@ -87,6 +87,24 @@ export type {
   Metrics,
   MetricsConfig,
 } from './metrics/types'
+export { PostHogTraces } from './traces'
```
jonmcwest, line 90: **Nit: mark cross-package plumbing as `@internal`.**

Please tag `PostHogTraces`, `SyncSpanContextManager`, `inertSpan`, `runWithActiveSpan`, `resolveTracesConfig`, `ResolvedTracesConfig`, and `TraceSdkContext`. They are exported from the published core package but are not intended as stable public API.

Use the existing cross-package JSDoc wording. Leave `SpanContextManager` public because it appears in the protected initialization hook. Non-blocking.

Sources: [core exports](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/index.ts#L90-L94) · [existing internal convention](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/utils/otlp-resource.ts#L86-L92).

*Agent-assisted review.*

packages/core/src/traces/index.ts
```diff
@@ -0,0 +1,937 @@
...
+    if (explicitParent && typeof explicitParent !== 'string' && !isOwnSpan(explicitParent)) {
+      if (looksLikeSpan(explicitParent)) {
+        // Inert like its parent, never an orphan with invented ids — but a
+        // pass-through parent's inbound context carries to the child rather than
+        // the trace ending here.
+        this._logger.debug('Span parent is not a span from this SDK; returning an inert span')
```
jonmcwest, line 226: **Optional API convenience: accept a single-element header array.**

Node's `headersDistinct.traceparent` supplies an array, but our public `parent` type is `Span | string`. OTel's first-value getter rule concerns carrier extraction; it does not require our parent argument to accept arrays.

Consider unwrapping exactly one element while keeping the multi-element fallback. Also correct the nearby comment: Node's ordinary headers join duplicates into a string. Non-blocking; not an OTel conformance defect.

Sources: [Node header arrays](https://nodejs.org/api/http.html#messageheadersdistinct) · [OTel getter contract](https://github.com/open-telemetry/opentelemetry-specification/blob/c09e58480b03dfef67a2fe81b1ce2f462daf5b4c/specification/context/api-propagators.md#get).

*Agent-assisted review.*

packages/core/src/traces/traceparent.ts
```diff
@@ -0,0 +1,135 @@
...
+ * Validates an incoming `tracestate` far enough to know it is safe to echo back.
+ * An invalid one is discarded without invalidating its traceparent, so a
+ * malformed vendor entry never costs us the trace continuation.
+ */
+export function sanitizeTracestate(value: unknown): string | undefined {
+  if (typeof value !== 'string') {
+    return undefined
+  }
+  const trimmed = value.trim()
+  if (!trimmed || trimmed.length > TRACESTATE_MAX_LENGTH) {
```
jonmcwest: **Correction: distinguish tracestate size from validity.**

W3C recommends propagating at least 512 characters; that is not a validity ceiling. For an otherwise valid header exceeding our size limit, truncate whole entries, preferably removing entries over 128 characters first, then from the right.

An incoming list with more than 32 members violates the grammar, so my earlier request to salvage that case was too broad. Narrow the test change to valid oversized headers. Non-blocking.

Sources: [W3C tracestate limits](https://www.w3.org/TR/trace-context-2/#tracestate-limits) · [current validation](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/traces/traceparent.ts#L114-L133) · [W3C member-count grammar](https://www.w3.org/TR/trace-context-2/#list).

*Agent-assisted review.*

jonmcwest:
**Follow-up: preserve array positions for null values.**

OTel requires preserving null positions in arrays and permits empty-string substitution when the exporter cannot emit null. The current encoder skips null/undefined slots, shifting later elements and misaligning paired arrays.

Given the receiver limitation documented here, emit `{ stringValue: '' }` for those slots and update the paired-array test. This addresses the null-position requirement; it is not a claim about other encoding choices. Non-blocking release follow-up.

Sources: [OTel AnyValue](https://github.com/open-telemetry/opentelemetry-specification/blob/c09e58480b03dfef67a2fe81b1ce2f462daf5b4c/specification/common/README.md#anyvalue) · [array encoding](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/utils/otlp-any-value.ts#L174-L184).

*Agent-assisted review.*

jonmcwest:
**Nit: distinguish send calls from HTTP attempts.**

The PR body's "exactly 1 request" counts `_sendTracesBatch` calls. The transport still retries, so one flush pass can make four HTTP attempts with the default retry count. The outage fix remains valid.

Please say "one send per flush pass" instead. Optionally cover the real transport in the outage test; the current mock hides retries. Non-blocking.

Sources: [SDK retry defaults](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/posthog-core-stateless.ts#L361-L365) · [transport retry loop](https://github.com/PostHog/posthog-js/blob/6fc61fadbc951c73f7df929304cf37e3b4b2fb40/packages/core/src/utils/index.ts#L80-L86).

*Agent-assisted review.*

packages/core/src/traces/span.ts
turnipdabeets: Done in `2581b66ef`. Both paths: `inertSpan` takes the active handle as the implicit parent, and `_resolveParent` adopts an active `PassThroughSpan`. Across-`await` is covered by a node test in `c8f9a66ea`, where `AsyncLocalStorage` actually applies. Taken your point that the tracing-on half is our fallback choice rather than the OTel rule.

packages/core/src/traces/index.ts
turnipdabeets: Agreed — #4726 goes out in the first traces release, not between the merges.

packages/core/src/traces/index.ts
turnipdabeets: Thanks, reverted in `c8f9a66ea`. I had already rewritten the `parent` docs against your earlier note; they are back to the original wording. I kept one test pinning what an untyped `spanContext()`-only object does today, since it is cheap coverage either way.

packages/core/src/traces/span.ts
turnipdabeets: Agreed — landed in #4792 with the cap and the per-event counter.

packages/core/src/index.ts
turnipdabeets: Done in `40278277f`. All seven tagged at the declaration rather than the re-export, `SpanContextManager` left public. `pnpm generate-references` is unchanged by it.

packages/core/src/traces/index.ts
turnipdabeets: Done in `2581b66ef`. Exactly one element is unwrapped, longer arrays are still ignored, and the nearby comment about `req.headers` is corrected.

packages/core/src/traces/traceparent.ts
turnipdabeets: Rebuilt to your revised version in `c8f9a66ea`. My first pass had >32 members trimming to 32 — that is back to a rejection, since it is a grammar violation rather than a size problem. A valid over-long header now drops members above 128 characters first, then from the right. Tests narrowed to valid oversized input.

turnipdabeets:
@jonmcwest — on the [send-calls vs HTTP-attempts nit](https://github.com/PostHog/posthog-js/pull/4579#issuecomment-5585537974): PR body updated. It now says "one send per flush pass" and names the transport retry, so the number is not read as an HTTP count.

turnipdabeets:
@jonmcwest — on [preserving array positions for null values](https://github.com/PostHog/posthog-js/pull/4579#issuecomment-5585537731). Not taken in this release, and the reason is cross-SDK rather than cost.

The receiver limitation you cite is real and I confirmed it: `patch_otel_json` only rewrites an empty object under a key named `value` or `body`, so an empty `{}` sitting inside `arrayValue.values` is not rescued and hits [opentelemetry-rust#1253](https://github.com/open-telemetry/opentelemetry-rust/issues/1253). `{ "stringValue": "" }` is the right substitution.

What stopped me is that this is not a posthog-js choice. All three SDKs drop nullish array elements identically today — `compactMap` in `PostHog/Logs/PostHogLogsOTLP.swift`, `mapNotNull` in `PostHogLogsOTLP.kt` — and the encoder here is shared by traces, logs and metrics, both of which have shipped. Changing one SDK would put the same attribute on the wire in two different shapes depending on which SDK sent it.

The contract is also silent on this, rather than agreeing with either behaviour. Both capabilities say only that "a `null` or `undefined` value SHALL cause the entire key to be omitted", which is key-level; neither says anything about elements inside an `arrayValue`. So writing a spec line now would be inventing the answer rather than recording it.

Suggested order, if you agree: settle it in `sdk-specs` first, then change js, ios and android together. Happy to open that proposal — I did not want to land a wire-format change across three SDKs on the back of a review comment.

jonmcwest (changes requested):
Agent-assisted final pass at `18976ec0cfe0b6efab33eb8088d06f0102d3e9c0`.

**Please preserve mandatory resource attribution when the encoder budget is exhausted.**

A valid resource attribute can consume the shared traversal budget before `service.name` and `telemetry.sdk.*` are encoded. User attributes are inserted first by the resource builder, and `buildOtlpTracesPayload()` passes the whole bag through that budget at `packages/core/src/traces/otlp.ts:181`.

Repro configuration:
```ts
traces: {
    serviceName: 'checkout-api',
    resourceAttributes: {
        payload: Array.from({ length: 10 }, () => Array(1000).fill(1)),
    },
}
```
Reproduced through the built public Node client on both this head and #4726's head, with fetch intercepted locally (no live ingestion). After `startSpan('checkout').end(); await posthog.flush()`, the request's resource keys are `["os.name", "os.version", "payload"]`: the explicitly configured service name and SDK attribution are missing. This violates our traces contract's “The SDK SHALL always send `service.name`”; ingestion reads service attribution from that attribute.

Please retain the traversal safety bound but reserve capacity for required attribution, and add an export-level regression with a budget-exhausting user attribute. Reordering object keys alone needs care because integer-like user keys enumerate first.

Holding approval for this. The first tracing release should still include #4726.

turnipdabeets:
@jonmcwest — on [resource attribution under an exhausted budget](https://github.com/PostHog/posthog-js/pull/4579#pullrequestreview-5168346760): done in `4896cf4e0`, with the test extended in `daeefd630`.

Reproduced first through the built Node client with your config: the resource went out as `["os.name","os.version","payload"]`.

The keys the resource builder sets itself (`service.name`, `deployment.environment`, `service.version`, `telemetry.sdk.*`) are now encoded on a traversal budget of their own, after the user's attributes (`toOtlpResourceKeyValueList` in `utils/otlp-resource.ts`). User attributes keep the full bound. They are split out by key rather than reordered, so an integer-like key cannot jump ahead of them, and SDK keys still win on collision.

Logs and metrics share this encoder and builder and have the same bug on main, so all three envelopes use it; there is a patch changeset for those. A differential run over 15k random resource configs gives byte-identical output except for key order on a user/SDK collision, and the budget-exhaustion and >1,000-key cases this fixes.

Tests: an export-level one through a traces flush, and one across the logs, metrics and traces payload builders for both a named key and `"0"`, with all five SDK keys set. Both fail on the previous head. The expected keys are read from the builder, so a key added there without being reserved fails the test.

`os.name`/`os.version` still share the user budget, since node passes them in as resource attributes. The contract makes only `service.name` a SHALL, so I left them; happy to reserve them too if you want.

jonmcwest (commented):
Agent-assisted final re-review at `daeefd6300142ecae1e6fdcd126b30329881fea1`.

The resource-attribution blocker is resolved. My unchanged reproduction now passes through the built public Node client: the large resource value no longer removes `service.name`, `telemetry.sdk.name`, or `telemetry.sdk.version`. I also checked integer-key coverage and the shared logs/metrics/traces encoding change.

Local verification: core 1,678 passed (one skipped), Node 996 passed, and the original public export reproduction passed. No new confirmed code blocker in the final delta.

I am withholding the approval stamp until CI is cleared: the minimum-TypeScript matrix failed during dependency installation (`@posthog/types@^1.410.1` could not be resolved), and the Firefox ingestion tests also failed. I have not established the cause of the Firefox failures. These CI failures should not be confused with the resource fix, which passed the targeted checks. The first tracing release should still include #4726.

turnipdabeets:
@jonmcwest — thanks for re-running it. On the two CI failures at `daeefd630`: the minimum-TS matrix installs the published `posthog-js@1.429.3`, which needed `@posthog/types@1.410.1` before npm had it (only 1.410.0 was published at the time; 1.410.1 is out now), so it was a release race rather than this diff. #4726 passed that same matrix a few minutes later. #4726 is now squash-merged into this branch (`397d4d8e0`), and CI on that head is green: all 61 checks, including Firefox and all 15 minimum-TS jobs.

jonmcwest (approved):

PR #4580, opened 2026-08-20
feat(node): flush() drains queued spans

turnipdabeets:
## Problem

Stacked on #4579 — review that first.

In a serverless handler you call `flush()`, not `shutdown()`: the container is reused across invocations, so tearing the client down would throw away the connection pool and the flag cache. But `flush()` drained only the event queue, so ended spans sat in the span queue until the flush timer happened to fire or the container was frozen.

That makes the most common `posthog-node` deployment shape the one that silently loses spans. Ending a span and calling `flush()` looks like it worked.

## Changes

`flush()` now drains the span queue alongside the event queue.

- **Concurrent, not sequential** — a handler waits for one round trip rather than two.
- **Never rejects on span export failure.** Spans that could not be sent stay queued for the next flush. `flush()` is a method callers already treat as safe, and adding a rejection path to it would be a worse regression than the one being fixed.
- Applies to the debounced `waitUntil` path too, not just direct calls, so it holds on Vercel/Cloudflare-style runtimes.
- No change when `traces` is unconfigured: `flush()` returns the event-queue promise unchanged.

This is a **contract change** to an existing shipped method, which is why it is its own PR with its own changeset and release note rather than being folded into #4579.

**How much later does `flush()` resolve?** Nothing changes without `traces` configured. With it, an empty span queue returns immediately, and otherwise it's one POST overlapping the events POST — so usually no measurable difference. A slow endpoint is bounded the same way events already are (10s timeout, 3 retries per batch).

### Reviewer notes

- The test that asserted the old behaviour (`flushes on its own interval, not with the events pipeline`) encoded the contract being changed here, so it was rewritten rather than deleted — the "own interval" half is still asserted separately.
- `shutdown()` is unchanged; it already drained spans.
- Originally planned as PR 7 of the tracing stack. Moved to the front because the failure mode is silent span loss for serverless users, which is the wrong thing to leave sitting behind five other PRs.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

Deliberately unchecked: this changes the timing contract of `flush()` for anyone with `traces` configured — it resolves later than before. No signature or type changes, and no effect when tracing is off.

### If releasing new changes

## 🤖 Agent context

Built with Claude Code, directed by @turnipdabeets.

This started as PR 7 in the tracing plan. While reviewing #4579's description, @turnipdabeets questioned why it sat that late in the stack; the deferral rationale ("the only outright contract change") justified giving it its own PR but not its position behind five others, given the failure mode is silent data loss for the most common serverless shape. Re-ordered to stack directly on the MVP.

The one design call worth review: swallowing span-export failures instead of surfacing them. The alternative — letting `flush()` reject — matches how event flushing behaves, but adds a new rejection path to a method most callers do not wrap in a `try`. Chose the conservative option, matching what `shutdown()` already does.


turnipdabeets:
Folded into #4579.

`flush()` draining spans only became observable once tracing existed, so splitting it out framed it as a contract change to a shipped method when it is really part of completing the feature. Keeping them separate risked a release cut between the two that silently dropped spans for every serverless user — the exact failure this PR existed to prevent.

The code and its test moved to #4579 unchanged; the separate changeset is gone, since `Add distributed tracing to posthog-node` now covers it.

Issue #4570, opened 2026-08-19
Extract the three duplicated OTLP senders; logs drops batches on 5xx where metrics and traces retry


turnipdabeets:
`PostHogCoreStateless` has three near-identical OTLP senders — `_sendLogsBatch`, `_sendMetricsBatch`, and `_sendTracesBatch`. Each is ~50 lines and they differ only in the endpoint path and auth style. Nothing keeps them in sync, and they have already drifted in a way that loses data.

## The drift

All three classify a send failure into the same tagged outcome, but not by the same rule:

| sender | retry-later when |
|---|---|
| `_sendLogsBatch` | `err instanceof PostHogFetchNetworkError` |
| `_sendMetricsBatch` | `isPostHogFetchRetryableError(err)` |
| `_sendTracesBatch` | `isPostHogFetchRetryableError(err)` |

`isPostHogFetchRetryableError` returns true for HTTP **408, 429, and ≥500**, *or* a network error. `PostHogFetchNetworkError` is network-only.

So once `fetchWithRetry` exhausts its internal retries against a struggling ingestion endpoint:

- **logs** classifies the 429/5xx as `fatal` and **drops the batch**
- **metrics** and **traces** classify it as `retry-later` and keep the batch queued for the next flush

Logs is the outlier — the two senders written later both use the broader predicate, and `_sendMetricsBatch` carries a comment explaining the choice. This looks like drift rather than a deliberate decision, but either way the behaviour isn't documented as intentional anywhere.

## Why extracting helps

The bodies are otherwise identical: disabled check, `JSON.stringify`, gzip unless `disableCompression`, custom headers, `fetchWithRetry` with a `retryCheck` that excludes 413, then the same four-way outcome mapping.

A single private `_sendOtlpBatch({ path, auth, payload })` returning `SendBatchOutcome` collapses ~150 lines to ~15 plus three thin wrappers, and makes the retry policy one decision instead of three copies that can silently disagree.

## Notes

- Fixing the logs classification is a behaviour change to a shipped pipeline (batches that are currently dropped would start being retried), so it deserves its own verification rather than riding along silently in a refactor.
- Purely internal — no public API surface involved.

turnipdabeets:
The encoder-side counterpart to this is #4571 — the logs `toOtlpAnyValue` emits int64 as a JSON number, which is why the traces encoder was written separately instead of sharing it.

PR #4583, opened 2026-08-20
fix(logs): encode OTLP attributes in the canonical spec shapes

### LINKED ISSUE #4571: Logs OTLP encoder emits int64 as a JSON number instead of a string
`toOtlpAnyValue` in `packages/core/src/logs/logs-utils.ts` encodes integers as a JSON number:

```ts
if (Number.isInteger(value)) {
  return { intValue: value }
}
```

OTLP/HTTP rides the proto3 JSON mapping, which requires int64 fields to be **strings**. The canonical form is `{ "intValue": "42" }`, not `{ "intValue": 42 }`.

## Consequences

1. **Precision loss above 2^53.** A JS integer larger than `Number.MAX_SAFE_INTEGER` serializes to a JSON number that cannot round-trip — the receiver reads a different value than the caller set.
2. **No int64 range guard.** An integer outside `[-2^63, 2^63-1]` is emitted as `intValue` regardless. A value the service can't fit into a signed 64-bit field risks rejecting the request, which for a batched endpoint takes every other record in the same request with it.
3. **Strict parsers.** Any consumer validating against the proto3 JSON mapping is entitled to reject a numeric `intValue`.

The same function also flattens objects to a JSON string rather than using OTLP's `kvlistValue`, which is called out in an existing comment as a simplicity tradeoff — that part is deliberate and not what this issue is about.

## Reference implementation

The traces encoder (`packages/core/src/traces/otlp.ts`) already does this correctly and can be used as the model:

- integer → `{ intValue: String(value) }`
- integer outside the int64 range → `{ stringValue: <exact decimal string> }` plus a debug warning, computed via `BigInt` so the exact value survives (`String(2**64)` gives `"18446744073709552000"`, not the number the caller holds)
- non-finite float → `{ stringValue: "NaN" | "Infinity" | "-Infinity" }` (logs already handles this correctly)

This is why the traces encoder was written separately rather than reusing the logs one — sharing would have meant adopting the non-conforming integer encoding. Once logs is fixed, the two encoders are worth merging; the sender-side duplication is tracked in #4570.

## Metrics is affected too

`packages/core/src/metrics/index.ts` imports `toOtlpKeyValueList` from `../logs/logs-utils` to encode data-point attributes, so metric attributes inherit the same numeric `intValue`. An existing test asserts the current shape:

```ts
expect(attrs).toContainEqual({ key: 'retries', value: { intValue: 2 } })
```

Metric *values* are unaffected — those use `asDouble`, which is a proto3 `double` and correctly a JSON number.

## Notes

- Changes the wire format of two shipped pipelines, so it wants verification against ingestion rather than a blind fix — and the assertion above has to change with it.

turnipdabeets:
## Problem

Fixes #4571.

`toOtlpAnyValue` encodes log and metric attributes into OTLP JSON. Several ordinary values encode into shapes PostHog's ingestion refuses — and a refusal is not scoped to the offending value. The request 400s and **every record in it is lost** (up to 100 logs). On `/i/v1/metrics` the same parse error is swallowed server-side: HTTP 200, and the metric never appears.

Verified against ingestion on project 381971:

| attribute value | what `main` sends | result |
| --- | --- | --- |
| integer ≥ 2^63, or `-(2**63)` | `{"intValue": 9223372036854775808}` | **400**, whole batch lost |
| a function or a symbol | `{}` — `JSON.stringify` returns `undefined` | **400**, whole batch lost |
| `[1, , 3]` | `null` as an array element | **400**, whole batch lost |
| `'ok\ud83d'` — a truncated emoji | a lone surrogate escape | **400**, whole batch lost |
| any of the above via `/i/v1/metrics` | same | **200**, metric silently dropped |

Separately, the encoding violates the canonical SDK contract in `sdk-specs` (`logs/spec.md`, "Attribute value encoding"): an integer must be a stringified int64, and an object must be a `kvlistValue`. iOS and Android already conform; posthog-js was the only SDK that didn't.

## Changes

- Integers encode as `{"intValue": "<decimal>"}`; outside int64, as the exact decimal in a `stringValue` with a debug warning — both per spec.
- Objects encode as a nested `kvlistValue` instead of a JSON string, per spec.
- Values that cannot be represented degrade to a marker rather than breaking the request: `[Circular]`, `[Truncated]` (20 levels, 1,000 items per container, 10,000 nodes), `[Unserializable]` (a property whose getter throws), `[Function]`.
- Strings and keys are surrogate-sanitised. Array holes and `null`s are dropped, matching iOS (`compactMap`) and Android (`mapNotNull`). `Date` encodes as ISO, `toJSON` is honoured, and an attribute named `__proto__` survives.
- `captureLog` no longer throws into caller code for any input, including a throwing getter or 25,000-level nesting — `JSON.stringify` threw too, but inside a `catch` this rewrite removed.
- `OtlpAnyValue.intValue` is now `string | number`. The SDK only emits strings; `number` stays valid because React Native rehydrates already-encoded records persisted by an older version and re-sends them unchanged.

### Verification

Every case below was POSTed to `/i/v1/logs` and read back out of ClickHouse.

| attribute | stored |
| --- | --- |
| `{ a: 1, nested: { deep: true } }` | `{"a":1,"nested":{"deep":true}}` — identical to the JSON-string form `main` sends |
| `2 ** 63` / `-(2 ** 63)` | `9223372036854775808` / `-9223372036854775808` |
| `'user said ok\ud83d'` | `user said ok�` |
| `{ ok: 1, get bad() { throw } }` | `{"bad":"[Unserializable]","ok":1}` |
| `{ handler: fn, retries: 2 }` | `{"handler":"[Function]","retries":2}` |
| 30-level nesting / 2^20-node graph | 20 levels then `[Truncated]` / capped at 496 KB |

**Object attributes store byte-identically** (same row, both forms, `length 40`, equal), so the change is invisible to anything querying them. Two stored values do move, both toward correctness: an integer above 2^53 stores its exact value rather than a rounded one, and `Date` loses the quotes `JSON.stringify` wrapped it in. Object attributes are ~2.4× larger on the wire, which brings the 413 batch-shrink path closer for object-heavy records.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for backwards compatibility of any changes (no breaking changes at runtime) — both wire forms are accepted and store identically. One intentional type-level break: widening `OtlpAnyValue.intValue` to `string | number` affects code that *reads* the field under strict TS. Disclosed in the changeset; no in-repo consumer reads it.
- [x] Took care not to unnecessarily increase the bundle size — **+1.54 kB per browser bundle** (`array.js` +0.6%), the recursive walker and its guards. The size bot's "+27 kB" is that same delta counted across 18 artifacts. Nothing here is optional at runtime: without the walker objects stay spec-violating, and without the guards ordinary values take down whole batches.

### If releasing new changes

The changeset-hygiene bot flags `posthog-js`, `posthog-node` and `posthog-react-native` as declared without modified source files. That is deliberate: the fix lives in `@posthog/core`, and `updateInternalDependencies: "patch"` against `workspace:^` ranges means a patch bump of core does not pull dependents into the release — without naming them, the three SDKs that ship this encoder would never cut a version carrying the fix. Precedent: `.changeset/metrics-node-wiring.md`.

## 🤖 Agent context

Claude Code (Opus 5), directed by @turnipdabeets, who asked whether the issue was worth fixing before any code was written, then ran four independent review cycles over it.

- The scope came from reading the receiver (`rust/capture-logs`, `opentelemetry-proto` 0.29's `deserialize_from_value`) rather than the OTLP spec alone. That reversed two of the issue's three stated consequences — a numeric `intValue` is accepted by every conformant parser, and the sub-2^53 precision argument doesn't hold — and turned the missing range guard into a confirmed batch-killer, alongside four more the issue never mentioned.
- Every claim about what ingestion accepts was tested against a live project rather than inferred; two reviewer findings ("stored data changes", "`@posthog/types` needs a minor bump") were rejected on that evidence and on `validate-versioning.yml`, which fails a `fix:` PR carrying a `minor` changeset.
- The author made two calls the agent did not: follow the traces spec's `stringValue` rule for out-of-int64 integers rather than propose amending it, and close the `kvlistValue` gap here rather than track it separately.
- Follow-ups, not in this PR: the unmerged traces encoder (`packages/core/src/traces/otlp.ts`) has two of these batch-killers — `String(-(2**63))` and `{}` for nullish array elements — and #4570's encoder merge is mechanical once it lands. `Error`, `Map` and `Set` still encode as an empty `kvlistValue` (accepted by ingestion, but their contents are lost), and the markers and budgets here are SDK-specific behaviour no spec covers yet.

### DIFF
--- a/packages/core/src/logs/logs-utils.ts
+++ b/packages/core/src/logs/logs-utils.ts
@@ -12,7 +12,16 @@ import type {
 import type { Logger } from '../types'
 import type { LogSdkContext, ResolvedPostHogLogsConfig } from './types'
 import { isArray, isBoolean, isNull, isNullish, isUndefined } from '../utils'
-import { sanitizeString } from '../utils/json-utils'
+import {
+  CIRCULAR_VALUE,
+  FUNCTION_VALUE,
+  MAX_JSON_SAFE_VALUE_DEPTH,
+  MAX_JSON_SAFE_VALUE_ITEMS,
+  MAX_JSON_SAFE_VALUE_NODES,
+  sanitizeString,
+  TRUNCATED_VALUE,
+  UNSERIALIZABLE_VALUE,
+} from '../utils/json-utils'
 
 // ============================================================================
 // Severity mapping
@@ -44,17 +53,6 @@ export function getOtlpSeverityNumber(level: LogSeverityLevel): number {
 // 2^63 — one past int64 max.
 const INT64_RANGE_LIMIT = 9223372036854775808
 
-// Matches `toJsonSafeValue`: depth bounds stack use, the counts stop a shared
-// object graph expanding into millions of key-values on the way to the wire.
-const MAX_DEPTH = 20
-const MAX_ITEMS = 1_000
-const MAX_NODES = 10_000
-
-const CIRCULAR_MARKER = '[Circular]'
-const TRUNCATED_MARKER = '[Truncated]'
-const UNSERIALIZABLE_MARKER = '[Unserializable]'
-const FUNCTION_MARKER = '[Function]'
-
 const propertyIsEnumerable = Object.prototype.propertyIsEnumerable
 
 interface EncodeState {
@@ -64,7 +62,7 @@ interface EncodeState {
 }
 
 function newState(): EncodeState {
-  return { ancestors: new WeakSet(), remainingNodes: MAX_NODES }
+  return { ancestors: new WeakSet(), remainingNodes: MAX_JSON_SAFE_VALUE_NODES }
 }
 
 export function toOtlpAnyValue(value: LogAttributeValue, logger?: Logger): OtlpAnyValue {
@@ -73,7 +71,7 @@ export function toOtlpAnyValue(value: LogAttributeValue, logger?: Logger): OtlpA
   } catch {
     // Runs inside `captureLog` and the metrics flush: an error escaping here
     // surfaces in the caller's own code.
-    return { stringValue: UNSERIALIZABLE_MARKER }
+    return { stringValue: UNSERIALIZABLE_VALUE }
   }
 }
 
@@ -92,7 +90,7 @@ function encodeAnyValue(
   depth: number
 ): OtlpAnyValue {
   if (state.remainingNodes <= 0) {
-    return { stringValue: TRUNCATED_MARKER }
+    return { stringValue: TRUNCATED_VALUE }
   }
   state.remainingNodes--
 
@@ -132,17 +130,17 @@ function encodeAnyValue(
   }
   // `String(value)` would put a function's source text on the wire.
   if (typeof value === 'function') {
-    return { stringValue: FUNCTION_MARKER }
+    return { stringValue: FUNCTION_VALUE }
   }
   if (typeof value === 'symbol') {
     return { stringValue: String(value) }
   }
   if (typeof value === 'object' && value !== null) {
     if (state.ancestors.has(value)) {
-      return { stringValue: CIRCULAR_MARKER }
+      return { stringValue: CIRCULAR_VALUE }
     }
-    if (depth >= MAX_DEPTH) {
-      return { stringValue: TRUNCATED_MARKER }
+    if (depth >= MAX_JSON_SAFE_VALUE_DEPTH) {
+      return { stringValue: TRUNCATED_VALUE }
     }
     if (value instanceof Date) {
       const time = value.getTime()
@@ -188,7 +186,7 @@ function encodeArrayValues(
   depth: number
 ): OtlpAnyValue[] {
   const result: OtlpAnyValue[] = []
-  const itemCount = Math.min(values.length, MAX_ITEMS)
+  const itemCount = Math.min(values.length, MAX_JSON_SAFE_VALUE_ITEMS)
   let index = 0
   for (; index < itemCount && state.remainingNodes > 0; index++) {
     try {
@@ -200,11 +198,11 @@ function encodeArrayValues(
       }
       result.push(encodeAnyValue(element as LogAttributeValue, logger, state, depth))
     } catch {
-      result.push({ stringValue: UNSERIALIZABLE_MARKER })
+      result.push({ stringValue: UNSERIALIZABLE_VALUE })
     }
   }
   if (values.length > index) {
-    result.push({ stringValue: TRUNCATED_MARKER })
+    result.push({ stringValue: TRUNCATED_VALUE })
   }
   return result
 }
@@ -222,7 +220,7 @@ function encodeKeyValueList(
     if (!propertyIsEnumerable.call(attrs, key)) {
       continue
     }
-    if (result.length >= MAX_ITEMS || state.remainingNodes <= 0) {
+    if (result.length >= MAX_JSON_SAFE_VALUE_ITEMS || state.remainingNodes <= 0) {
       // Reported rather than written into the attributes: a synthetic key would
       // land in the user's own namespace and could collide with a real one.
       logger?.debug('Attributes truncated: the value exceeds the OTLP encoder budget')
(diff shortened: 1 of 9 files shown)


packages/core/src/logs/logs-utils.ts
```diff
@@ -146,18 +306,39 @@ export function buildOtlpLogRecord(options: CaptureLogOptions, sdkContext: LogSd
...
   const record: OtlpLogRecord = {
     timeUnixNano: now,
     observedTimeUnixNano: now,
     severityNumber,
     severityText,
-    body: { stringValue: options.body },
-    attributes: toOtlpKeyValueList(mergedAttributes),
+    // `body` is only typed as a string: an untyped caller passing a number
+    // emits a non-string stringValue, which the server refuses.
+    body: { stringValue: sanitizeString(String(options.body)) },
```
ioannisj: Should we wrap the `String(options.body)` coercion here in the same `[Unserializable]` fallback the attributes get? It's the one conversion path the never-throw hardening didn't reach: `captureLog` calls `buildOtlpLogRecord` with no try/catch, so a body whose `toString` throws propagates straight out of `captureLog`, and `beforeSend` can hand back any body.

As it stands a throwing-`toString` regression would still pass, the "coerces a non-string body" test only covers a well-behaved number.

packages/core/src/logs/logs-utils.ts
```diff
@@ -39,49 +41,203 @@ export function getOtlpSeverityNumber(level: LogSeverityLevel): number {
 // OTLP AnyValue conversion
 // ============================================================================
 
-export function toOtlpAnyValue(value: LogAttributeValue): OtlpAnyValue {
+// 2^63 — one past int64 max.
+const INT64_RANGE_LIMIT = 9223372036854775808
+
+// Matches `toJsonSafeValue`: depth bounds stack use, the counts stop a shared
+// object graph expanding into millions of key-values on the way to the wire.
+const MAX_DEPTH = 20
```
ioannisj: [suggestion] Worth importing `MAX_DEPTH`/`MAX_ITEMS`/`MAX_NODES` from `json-utils.ts` instead of re-declaring them?

ioannisj (approved):

packages/core/src/logs/logs-utils.ts
turnipdabeets: Good catch — fixed in f57a2da0a. `String()` throws on both a throwing `toString` and `Object.create(null)`, and `buildOtlpLogRecord` is called straight from `captureLog`, so it escaped the hardening. Same `[Unserializable]` fallback now, with a test covering both shapes.

packages/core/src/logs/logs-utils.ts
turnipdabeets: [suggestion] Done in f57a2da0a — the budgets and the four markers now come from `json-utils`, so the "matches `toJsonSafeValue`" comment is enforced rather than aspirational.

One consequence worth your read: `core/src/index.ts` does `export * from './utils'`, so this puts those constants — and `sanitizeString` — on `@posthog/core`'s public surface. Happy to move them to an internal module if we'd rather not commit to them.

turnipdabeets: Went with the narrower version in b55deaee6: `utils/index.ts` now re-exports `toJsonSafeValue` by name instead of the whole module, so the budgets, markers and `sanitizeString` are shared with the encoder but stay out of `@posthog/core`'s public API — the published surface is identical to `main`.

PR #4584, opened 2026-08-20
feat(node): beforeSpanSend hook and per-span limits

turnipdabeets:
## Problem

Two gaps in what a finished span may carry, which only make sense together:

**Nothing scrubs a span.** Spans carry whatever attributes the application puts on them, and a service that instruments its HTTP layer ends up with headers, query strings and user fields on them. There was no way to strip any of it before it left the process, and no way to drop a span you don't want exported at all (health checks, noisy internal polling).

**Nothing bounded a span.** A loop calling `span.addEvent()` per iteration, or an instrumentation layer copying every request header onto a span, grows it without limit until the ingestion endpoint rejects the payload as too large — at which point the 413 path shrinks the batch to a single span, that span is still too big, and it is **dropped entirely**. You lose the whole span rather than the excess.

Implements the `Gating and beforeSpanSend` and `Span limits` requirements of the traces capability spec.

## Changes

```ts
new PostHog('phc_...', {
  traces: {
    serviceName: 'checkout-api',
    maxAttributesPerSpan: 128,     // default
    maxEventsPerSpan: 128,         // default
    maxAttributeValueLength: 8192, // default
    beforeSpanSend: (span) => {
      if (span.attributes['http.route'] === '/health') return null
      delete span.attributes['http.request.header.authorization']
      return span
    },
  },
})
```

**`beforeSpanSend`** runs on every finished span before it is queued.

- **Return `null` to drop the span.** Nothing is enqueued and no error surfaces.
- **Plain values, not the wire format.** The hook reads `userId: 42`, not `{ intValue: "42" }`. It runs after auto-context attributes are attached, so `posthogDistinctId` and `sessionId` are visible and scrubbable.
- **Identity fields are read-only.** `traceId`, `spanId` and `parentSpanId` are typed `readonly`, and an assignment that slips past the type system is reverted with a debug warning — rewriting ids after children have already shipped corrupts parentage.
- **A throwing hook drops the span (fail-closed).** The hook is the documented scrubbing point, so a broken scrubber must not leak an unscrubbed record. This matches `PostHogLogs._runBeforeSend`, which also catches and drops — the spec describes the logs hook as fail-open, but this SDK's implementation is not.
- **Arrays run left to right**, and the first hook returning `null` stops the chain. Non-callable entries are ignored rather than called.
- **Synchronous.** A returned promise is not awaited; the span is dropped and the type says so.

**Per-span caps** at OpenTelemetry's defaults.

- **Earliest wins.** The first 128 attributes and events are kept; later ones are dropped silently.
- **The counts ship with the span** as `droppedAttributesCount` / `droppedEventsCount`, so a truncated span is visibly truncated rather than quietly wrong. Both are omitted when nothing was dropped.
- **SDK-attached keys are exempt and never evicted.** `posthogDistinctId`, `sessionId`, `url.full`, `screen.name` and `app.state` do not count toward the cap, so a span at the limit still links back to its person and session. The exemption is from the *count* only — a hook can still delete them, which is the point of a scrubbing hook.
- **Overwriting an existing key always succeeds.** The cap counts distinct user keys, not writes.
- **The event cap is absolute.** An `exception` event the SDK records spends an ordinary slot like any other, so a span that fills its events and then throws keeps its `error` status but not the exception detail. `droppedEventsCount` reports the loss. See *The exception reserve, and why it isn't here* below.

**A per-value length bound**, `maxAttributeValueLength`, default 8192.

- The count caps alone don't bound a span: one multi-MB value takes it past the body limit on its own, and the too-large path then loses that span whole. OpenTelemetry leaves its equivalent unlimited; the spec requires a finite default.
- The bound reaches **every string the value contains**, including strings nested inside arrays and objects — a value the caller nested is no smaller than one they didn't.
- It applies to span attributes, event attributes, **status messages and resource attributes** alike.

**`exception.stacktrace` on spans.** `recordException` and a throwing `withSpan` callback now attach the stack alongside `exception.type` and `exception.message`. Stacks ship by default and carry your server's file paths — the changeset says so, and `beforeSpanSend` can remove them.

**The caps re-apply after `beforeSpanSend`**, which is why these ship together.

## Why one PR

These were opened as #4584 and #4586, two branches off the same base. The spec requires *"The caps SHALL be re-applied after `beforeSpanSend`"* — and neither branch can satisfy that alone, because neither has both halves in its tree. Split, the best either could do was leave a note asking whichever merged second to remember.

The failure that requirement prevents is real and not theoretical: the hook writes to the plain record, not through the span's guarded writer, so a hook that *enriches* a span — adding a customer tier, a region, a git SHA, the second most obvious use after scrubbing — pushes it straight back past the cap, with no trim and no dropped count. It then 413s and the span is lost whole, which is exactly what the cap exists to prevent.

They also conflicted in six files, so there was never a version of this where the two could be reviewed independently.

### Reviewer notes

- **The re-apply lives in `applySpanLimits`**, called at the end of the hook chain. Earliest-set entries win, matching the span-side rule, SDK-attached keys stay exempt, and the counts add to whatever the span itself already dropped. A hook that only *removes* attributes cannot invent a dropped count.
- **`PostHogSpan`'s end callback now passes the auto-attached keys** alongside the record. This is internal — the public `SpanRecord` is deliberately unchanged, so the exempt-key set never becomes API surface.
- **Enforcement otherwise lives in `PostHogSpan`, not at encode time**, so an over-cap span never occupies memory in the first place — the point is to bound growth, not to trim on the way out.
- **Gate order** is spec-mandated: SDK disabled → opted out → `beforeSpanSend`. Only then does the queue-depth check run, so a dropped span never counts against the queue.
- Auto-context keys are identified by what the SDK attached at span start, so a user who explicitly sets `posthogDistinctId` themselves gets the exempt treatment for that key.
- **The post-hook pass works on a copy of the record.** Every write it makes would throw on a frozen one, and being fail-closed that would silently drop every span a defensive `Object.freeze({ ...span })` scrubber returned.
- **The value walk carries the encoder's own budget** — an ancestors `WeakSet` and a 10,000-node cap — not just a depth limit. Depth alone is not a bound when siblings share a subtree.
- **Span events are re-sanitized after the hook too**, not just `name` / `startTime` / `endTime`. A hook that appends or rebuilds `events` bypasses the checks `addEvent` applies, so an event with no timestamp encoded as `timeUnixNano: "NaN000NaN"` and one scaled by `1e6` encoded as a 25-digit value. Neither throws in `_encodeBatch`, so the span shipped, the server rejected the request, and `_flushInner`'s fatal branch spliced the whole batch — up to `maxExportBatchSize` unrelated spans lost.

### What review changed

Two fresh-context reviewers went at the branch, one required to demonstrate every finding by running it. Each fix below is pinned by a test that fails when it is reverted.

- **The value walk could hang the event loop.** `truncateAttributeValue` had a depth cap and nothing else, so a value whose children point back at their siblings cost `fanout ** depth` visits: two self-references measured 254 ms of synchronous work inside the caller's own `setAttribute`, three did not finish in two minutes. It now carries the same `ancestors` WeakSet and node budget as `encodeAnyValue`, which it was supposed to mirror. This was a regression against the base, where the raw value went straight to the guarded encoder.
- **A `toJSON` bypassed the bound entirely.** The walk read own keys; the encoder prefers `toJSON`. A dayjs value, a Decimal, or an ORM document could put a megabyte on the wire. The walk now resolves `toJSON` the way the encoder does.
- **A frozen record dropped 100% of spans**, debug-only. The post-hook pass now copies first. The previous test *asserted* the drop; it now asserts the export.
- **The constructor walked the prototype chain** (`for...in`), so a polluted `Object.prototype` key became an attribute of every span, defeating the encoder's own `propertyIsEnumerable` guard.
- **A nested `__proto__` key was lost** and swapped the copy's prototype — the copy used assignment where the rest of the codebase uses `defineProperty` for exactly this.
- **Resource attributes were never length-bounded**, though the spec applies the bound to "span attributes, event attributes, and resource attributes alike" — and resource attributes ride on every batch, not on one span.
- **Status messages were unbounded**, including the one `recordException` set from the same string it had just truncated for the event.
- **A hook writing an unknown `status.code`** encoded as an empty status object, silently losing an error the span really had. An invented `droppedAttributesCount` reached the wire uncoerced.
- Each key was read twice per walk, running a getter a second time.

### The exception reserve, and why it isn't here

Review found that a span already at `maxEventsPerSpan` which then throws keeps its `error` status but loses the `exception` event, stack included. This PR carried a fix for a while: four slots reserved past the cap, on the reading that the spec caps **"user-supplied"** content and an exception the SDK attaches on your behalf isn't user-supplied.

It has been removed, and the cap is now absolute.

Two rounds of review went into the reserve and it produced two bugs of its own — first a reserve that was really a smaller exclusive cap (twenty `recordException` calls on an empty span kept four), then eligibility granted by event *name*, so a caller's own `addEvent('exception')` could put 132 events past a documented 128. Fixing the second properly meant an internal provenance marker on the event record, which worked but added a third thing to carry.

Set against that: nobody can show the case occurs. It needs a single span holding 128 events *and* then throwing. Traces has not shipped, so there is no usage to appeal to.

The decisive asymmetry is that **adding the reserve later is purely additive** — no caller breaks when a cap grows — while shipping it now and removing it later is a behaviour change. And the case is measurable in production without it: a span that fills up and throws still exports `status: error` and a non-zero `droppedEventsCount`, so `status = error AND droppedEventsCount > 0` finds exactly this population once traces is live. If it turns out to be real, the reserve comes back with numbers behind it.

Removing it deleted 95 lines and one whole class of duplication: the rule no longer has two independent implementations in the span writer and the post-hook re-apply, so they cannot drift. It also means `maxEventsPerSpan` matches the spec's stated number exactly, with no `sdk-specs` change needed to ship.

Span and event *names* remain unbounded. That is 4579's path and unchanged here.

Known residuals, each rated NIT and deferred deliberately:

- **A hook can still mutate a queued span through a nested value it kept a reference to.** The walk returns the same reference when nothing needed shortening, so a hook that holds a nested object and edits it after returning reaches the wire. This is the window `setAttribute` has always had rather than something the hook introduces, but the "edit in place" documentation invites the assumption that returning ends it.
- **An array whose own `slice` throws escapes the length bound.**

- **Integer-like attribute keys break "earliest-set wins" in the post-hook re-apply.** `Object.keys` front-loads array-index keys, so a hook adding an attribute named `"0"` jumps the queue and can evict a real one. Fixing it properly means tracking insertion order in a `Map` rather than an object, which is a data-structure change for a case that needs an attribute literally named `"0"`.
- **A span that fills its events and then throws loses the exception detail.** Deliberate, as above: the cap is absolute and the loss is reported through `droppedEventsCount`. Raising `maxEventsPerSpan` is the workaround for a span that both records many events and can fail.

### Second review round

Two more fresh-context reviewers went over the post-fix branch, one required to demonstrate every finding by running it. Both independently found the same two blocking defects, and both are now fixed and pinned by tests:

- **The exception reserve was a smaller exclusive cap, not a reserve.** Fixed at the time, then removed entirely along with the reserve — see above.
- **A multi-megabyte string could still escape `maxAttributeValueLength`.** The traversal budget was charged before the string leaf was bounded, and the walk had no per-container item cap where the encoder stops at 1,000. So `{ rows: [20k ints], html: 2MB string }` spent the whole 10,000-node budget on elements the encoder never emits and shipped the sibling string whole — 2,000,000 characters against a bound of 8192, on default config with no hook. Strings are now bounded before the budget is consulted, only containers are charged, and arrays are walked to the encoder's own item cap. The same repro now yields 8192 characters and a 27 KB body.

Also fixed from that round:

- **A hook mutating `status.code` in place defeated the fallback**, because the snapshot held a reference to the object the hook was editing — and in-place mutation is the style this PR's own doc example uses. The same bug applied to the dropped counters. Both are now snapshotted by value before the chain runs.
- **The post-hook copy is now an explicit field-by-field rebuild** rather than a spread. A spread copies own properties only, so a hook returning a class instance that exposes `events` through a prototype getter lost every span to the fail-closed branch. Naming the fields also means nothing a hook attached beyond them can reach the wire, and the SDK's own dropped counters are taken from the span rather than from the hook's return value.
- **A hole or a non-object in `events`** cost the whole span; it now costs that entry.
- **`_writeAttribute` bounded values before checking the cap**, spending ~354 ms of a 389 ms `setAttributes` call walking values it was about to drop. The cap is checked first.
- **`toJSON` ran twice**, and a value that answered differently the second time escaped the bound entirely. It now resolves once and the resolved form is what is stored.

### Third review round

Two more independent reviewers: one hunting bugs, one looking only for accidental damage to already-shipped code.

One more blocking defect, fixed and pinned:

- **A single throwing nested getter disabled `maxAttributeValueLength` entirely.** The walk read child properties inside the *parent's* `try`, so one throw abandoned the whole value and the catch returned it untruncated. On default config, `{ body: <5 MB string>, get lazyRelation() { throw } }` shipped a **5,000,717-byte payload**; the same value without the getter shipped 8,843. That is a lazy ORM relation or a disposed resource sitting next to a large field — the shape `assignUserAttributes` already guards at depth 0, but only at depth 0. Per-key reads are now guarded individually: a throwing accessor costs its own key, which gets the encoder's `[Unserializable]` marker, and its siblings are still bounded. Re-measured: **8,979 bytes**.
  The test covering this asserted the value was *left alone*, so the bug was pinned as intended behaviour. It now asserts the sibling is bounded.

Also fixed:

- **A `null` in `events` cost the whole span**, not one event: the sanitising pass ran before the guard meant to catch it. Each event is now read behind its own guard.
- **`attributes` was not repaired the way `events` was**, so a hook setting it to `null` dropped the span.
- **A `toJSON` resolving to `undefined`** consumed a cap slot while encoding to nothing, and an invalid `Date` vanished where it previously shipped `"Invalid Date"`. A nullish result now keeps the original value.
- **A non-string `status.message` from a hook** bypassed the length bound and was stringified at full length by the encoder.

**No regressions to shipped code.** Measured, not assumed: the `posthog-js` browser bundle is **+0 bytes** (un-minified bundles byte-identical; none of the traces symbols appear in either), `packages/core/src/utils` is untouched, `describeError`'s new `stack` field reaches nothing outside traces (it is on no barrel), neither `SpanRecord` nor `BeforeSpanSendFn` collides with an existing export, and there is no import cycle. Full suites across 24 packages: ~13,370 passing, 0 failures. One `@posthog/nuxt` fixture fails identically on the base commit — upstream nuxt-nightly breakage, unrelated.

### Fifth review round

A pass over the branch after merging #4579's latest base into it. One conflict, in `traces/index.ts`'s `./span` import: this branch's `applySpanLimits` and `truncateAttributes` kept, `NOOP_SPAN` dropped, because the base replaced that call site with `inertSpan(options)` so a pass-through parent's inbound context carries to the child. Everything else auto-merged.

`positiveInteger` now falls back to the default for a fractional knob rather than flooring it, so that residual is gone: `maxAttributesPerSpan: 1.5` resolves to `128`, not `1`.

The span-limits changeset still promised "a small reserve for exception events" after the reserve was removed, which would have shipped a wrong entry in the published changelog for three packages. Corrected, and it now names the dropped counters, which are how a caller sees a truncated span.

Re-verified on the merged tree: `packages/core` 1530 pass / 1 skipped (64 suites), `packages/node` 995 pass (35 suites) plus the edge-runtime config, `@posthog/types` 3 pass, oxfmt and oxlint clean, public API references regenerate with no drift. Run as a real service against a local ingestion endpoint, reading the decoded OTLP off the wire — 45 assertions covering the hook (drop, scrub, forge, throw, frozen return, `async` return, prototype-getter return), the caps and their dropped counts, the join-key exemption under a real `withContext`, the value bound through nesting / `toJSON` / a cyclic value / a throwing sibling getter, bounded status messages and resource attributes, and an inbound distributed `traceparent` reaching the hook with its real ids.

### Fourth review round

Two reviewers on the ground earlier rounds had not covered: lifecycle and concurrency, and the thing run as a real application.

Three fixes, all of them consequences of round three's own repairs:

- **A repaired record could export as junk instead of being dropped.** Forcing `attributes` to `{}` when it was not an object meant an `async` hook — whose Promise is truthy and reads `undefined` for every field — exported a span named `unknown` carrying *no attributes at all*: no `posthogDistinctId`, no `sessionId`, joinable to nobody, and silently. Before that repair it threw and was a warned drop. A return value missing either collection is now dropped and counted, which is the fail-closed rule the rest of the hook contract already follows. This also covers `attributes: null` and an array, which would have encoded as `{ "0": … }`.
- **Keeping the original value when `toJSON` resolved nullish let the walk bypass a redaction.** The value fell through to the own-key walk, which produced a plain copy the encoder no longer recognises as self-describing — so `class Redacted { toJSON() { return null } }` put its internals on the wire at `maxAttributeValueLength: 10` and redacted correctly at 100. A self-describing value is now left whole.
- **Coercing a hook's `status.message` could throw and cost the span.** A `{ toString() { throw } }` message exported nothing, where the encoder downstream only marks the field. The coercion is guarded.

**Ran as a real Express service against production ingestion** (`cycle4-svc-1788369331`, project 381971) with `setupExpressRequestContext`, a scrubbing hook, nested spans, a genuine HTTP hop and an error route — 17 spans across 4 traces read back out of the product. Parentage held across the hop on one trace id; `posthogDistinctId` and `sessionId` on 17/17 spans including downstream ones; the scrubbed header absent from all of them; zero `/health` spans from five requests; error spans at `status_code: 2`. The hook wrote `traceId = '0'.repeat(32)` on every span and every exported id is real — identity immutability demonstrated in production, not just in a unit test.

**Spec conformance walked scenario by scenario**: 24 checks over `Span limits`, `Gating and beforeSpanSend`, `Configuration knobs` and the exception-recording scenarios — 0 failures, with `context manager form` not applicable (Python-only).

### Cost

Measured against the base over 200k spans, three repeats, medians: a plain span goes 478k/s → 403k/s (2.09 → 2.48 µs), a span with 20 attributes 249k/s → 151k/s (4.02 → 6.62 µs), and a configured hook costs about the same again. No heap regression.

Most of the attribute-heavy cost is not the new work. It was attributed rather than guessed: neutering truncation, the cap accounting, and the ancestors set each moved it by ~5k/s, while swapping the null-prototype attribute store for `{}` moved it to 195k/s — V8 puts null-prototype objects in dictionary mode. That store is the `__proto__`-pollution guard, so this is a deliberate trade, kept: the worst case is 6.6 µs per span, 0.66% of a core for a service producing a thousand spans a second.

### Verification

`packages/core` 1461 pass (63 suites), `packages/node` 1034 pass (36 suites), oxfmt and oxlint clean, `@posthog/types` / `@posthog/core` / `posthog-node` build clean, public API references regenerate with no drift.

Checked end-to-end against a real project (381971) with the caps set low so the behaviour is visible — attributes 5, events 3, values 64 — and read back out of the product, not just off the wire:

- `GET /health` dropped by the hook: absent from the payload and from the product.
- `http.request.header.authorization` deleted by a hook that then froze what it returned: gone, span still exported, `enriched.tier` present.
- The hook forged `traceId` to all zeros on every span; the exported ids are the real ones.
- `overflow`: kept `attr-0`..`attr-4` plus both join keys, `droppedAttributesCount: 16` (15 user attributes plus the hook's own enrichment, trimmed by the post-hook re-apply), `droppedEventsCount: 7`.
- `long-values`: a 5000-character string, the same string nested inside an array, a `toJSON` returning 5000 characters, a doubly self-referencing object and a 5000-character status message — all bounded to 64, nothing hung.
- A 5000-character `resourceAttributes` value bounded to 64 on the batch resource.
- `retries`, eight `recordException` calls with the event cap at three: three kept, `droppedEventsCount: 5`.
- `throwing`, on a span that added ten events before it threw with the event cap at three: `step-0`..`step-2` kept, `droppedEventsCount: 8`, `status_code: 2`, original error propagated unchanged. The `exception` event does not fit, which is the documented behaviour of an absolute cap.

  These two figures are recomputed against the implementation, not re-observed: the product read-back described in this section was run before the reserve was removed, so its `retries` and `throwing` spans reflect the reserve's behaviour. Everything else in the list is unaffected by that change and stands as recorded.
- Both counters omitted on the spans that dropped nothing.

### Carried over from #4579's review

**Consent is now re-checked between span batches.** A drain sends one batch per loop iteration and gates consent only before the first, so a user opting out while a batch was in flight left the batches behind it exporting `posthogDistinctId` and `sessionId` afterwards. `_flushInner` now re-checks `isDisabled` / `optedOut` at the top of every iteration and discards the rest of the queue. It is the existing discard block moved into one method and called from a second place, not a new mechanism. Two regression tests: a backlog built during an outage, then `optOut()` — and separately `disable()` — landing while the retried head batch is in flight, both verified to fail with the loop-body check removed.

Fixed here rather than in #4579 because that PR is approved and this one is stacked on it, so both ship in the same release. `disable()` was already covered before this: `_sendOtlpBatch` short-circuits on `disabled` and returns `{ kind: 'fatal' }`, which the loop splices away without a request. The genuine hole was `optOut()`, which nothing downstream checked.

No changeset: the behaviour it corrects is inside the traces feature #4579 introduces, which has not shipped.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

All surface is additive on top of #4579, which has not shipped, so nothing here can break released code.

### If releasing new changes

**Docs**: [PostHog/posthog.com#19837](https://github.com/PostHog/posthog.com/pull/19837) covers this surface too — a *Scrubbing and dropping spans* section, a *Span limits* section, and the new configuration rows. It stays a draft until this and #4579 ship, and documents them as one release, since this PR merges into #4579's branch.

## 🤖 Agent context

Built with Claude Code, directed by @turnipdabeets, from the merged `traces` capability spec in `sdk-specs`. The merge of #4586 into this PR, and the cap re-apply, came out of a review pass that checked the stack against that spec and found the requirement no single PR could satisfy.


packages/node/src/traces-defaults.ts
```diff
@@ -53,6 +62,18 @@ function withUsableIdentityKeys(attributes: TracesConfig['resourceAttributes']):
...
+/**
+ * Keeps only the callable hooks. Anything else is dropped rather than called: an
+ * untyped caller passing the wrong shape would otherwise have every span dropped
+ * by a hook that throws, leaving tracing silently off.
+ */
+function resolveBeforeSpanSend(beforeSpanSend: TracesConfig['beforeSpanSend']): BeforeSpanSendFn[] {
+  if (!beforeSpanSend) {
+    return []
+  }
+  return [beforeSpanSend].flat().filter((hook): hook is BeforeSpanSendFn => typeof hook === 'function')
```
marandaneto: blocking: Invalid scrubber configuration fails open — Filtering non-functions turns a malformed `beforeSpanSend` configuration into an empty hook chain, so spans, including attributes the caller expected to redact, are exported without an error or warning. A focused resolver test confirmed that invalid configuration returns normally with no hooks rather than failing fast.

packages/core/src/traces/span.ts
```diff
@@ -259,22 +502,169 @@ export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }):
...
+  // A string is a leaf and costs no traversal, so it is bounded before the
+  // budget is consulted. Charging it would let one large collection spend the
+  // budget and leave every string after it on the wire at full length.
+  if (typeof value === 'string') {
+    return truncateString(value, maxLength)
+  }
+  if (value === null || typeof value !== 'object') {
+    return value
+  }
+  if (state.remainingNodes <= 0 || depth >= MAX_JSON_SAFE_VALUE_DEPTH || state.ancestors.has(value)) {
```
marandaneto: blocking: Cyclic values reintroduce unbounded strings — Returning the original object for an ancestor leaves a raw reference in the copied parent; the OTLP encoder descends into that reference once before detecting the cycle and emits its long strings despite `maxAttributeValueLength`. A focused test encoded `abcdefgh` from a self-reference with a limit of four.

packages/core/src/traces/span.ts
```diff
@@ -259,22 +502,169 @@ export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }):
...
+      // Only the items the encoder will emit are walked; it stops at the same
+      // cap, so bounding the rest is work spent on values that never ship.
+      const walked = Math.min(value.length, MAX_JSON_SAFE_VALUE_ITEMS)
+      let boundedItems: SpanAttributeValue[] | undefined
+      for (let index = 0; index < walked; index++) {
+        const item = value[index]
+        const boundedItem = truncateValue(item, maxLength, state, depth + 1)
+        if (boundedItem !== item) {
+          // Copied lazily, so an array that needed nothing allocates nothing.
+          boundedItems = boundedItems ?? value.slice()
```
marandaneto: blocking: Array copying can discard all truncation — Although only the first 1,000 entries are walked, `value.slice()` reads the entire array. A throwing accessor beyond the encoder cap reaches the outer catch, returns the original array, and lets oversized entries in the encoded range ship untruncated; a focused test observed an eight-character first entry with a four-character limit.

packages/core/src/traces/span.ts
```diff
@@ -259,22 +502,169 @@ export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }):
...
+          boundedItems[index] = boundedItem
+        }
+      }
+      return boundedItems ?? value
+    }
+    const bounded: SpanAttributes = {}
+    for (const key of Object.keys(value)) {
```
marandaneto, line 645: blocking: Object traversal ignores the encoder item cap — This loop enumerates and reads every property even though OTLP emits at most 1,000, adding unbounded synchronous getter work to `setAttribute()`. A focused test attached 20,000 enumerable getters and observed all 20,000 being read.

packages/core/src/traces/span.ts
```diff
@@ -259,22 +502,169 @@ export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }):
...
+      // defineProperty, not assignment: a nested `__proto__` key would otherwise
+      // swap the copy's prototype and vanish.
+      Object.defineProperty(bounded, key, {
+        value: boundedItem,
+        enumerable: true,
+        writable: true,
+        configurable: true,
+      })
+    }
+    return changed ? bounded : value
```
marandaneto: blocking: Mutable nested attributes bypass the value limit — Returning the original object when the initial walk finds nothing to truncate leaves caller-owned mutable state attached to the span. Without a hook there is no end-time limit pass, so a nested string changed after `setAttribute()` ships at full length; a focused test mutated `ok` to `abcdefgh` and encoded all eight characters with a limit of four.

marandaneto (commented):

packages/core/src/traces/span.ts
turnipdabeets: Confirmed and fixed in 308c686c6.

Reproduced exactly as described: `{ body: 'abcdefgh', self: <itself> }` with a limit of 4 encoded as

```json
{"key":"self","value":{"kvlistValue":{"values":[{"key":"body","value":{"stringValue":"abcdefgh"}}, ...
```

The mechanism is that the walk hands back the raw ancestor, and the parent it lands in is a *copy* — so the encoder's `ancestors` set, which tracks its own path, doesn't recognise the reference and descends one full level before detecting the cycle on the next hop.

The fix returns `CIRCULAR_VALUE`, the same marker the encoder produces, rather than the value. It now encodes as `{"key":"self","value":{"stringValue":"[Circular]"}}` with `body` at `abcd`.

I split the condition while I was there, because the other two arms of it are not the same situation: `remainingNodes` and `depth` are genuinely safe to return raw, since the encoder stops at the same depth and charges *every* value including strings, so its budget is always spent no later than this walk's. Only the cycle case escaped it. Regression test in `span.spec.ts`; it fails against the previous code.

packages/core/src/traces/span.ts
turnipdabeets: Confirmed and fixed in 308c686c6.

Reproduced with a 1,200-entry array whose index 1,100 is a throwing getter: the first entry came back as `abcdefgh` with a limit of 4. `slice()` reads the whole array, the getter throws, and the outer catch returns the value unbounded — so the throw doesn't just cost its own item, it costs every item the encoder *will* emit.

Rather than guarding the `slice()`, the walked items are now accumulated into a fresh array, so nothing outside the encoder's range is ever read:

```ts
const boundedItems: SpanAttributeValue[] = []
for (let index = 0; index < walked; index++) { ... }
if (value.length > walked) {
  boundedItems.length = value.length
}
```

Carrying the original length matters — the encoder emits its `TRUNCATED_VALUE` marker off `values.length > index`, so trimming to `walked` would silently drop the "there was more" signal. Holes past the cap read as `undefined`, which the encoder skips exactly as it skipped the holes `slice()` used to preserve.

One thing beyond what you flagged: I gave each element its own try/catch. A throwing accessor *inside* the walked range had the same effect as one past the cap, and the object branch and the encoder both already handle per-item throws this way. First entry is now `abcd`. Regression test in `span.spec.ts`.

packages/core/src/traces/span.ts
turnipdabeets: Confirmed and fixed in 308c686c6. 20,000 enumerable getters, all 20,000 read; it is now 1,000.

The counting needed care, so worth spelling out. The encoder's cap is on what it *emits*, not what it walks: `encodeKeyValueList` skips a nullish value with `continue`, without charging `result.length`. So a naive "stop after 1,000 keys" would be wrong — a bag whose first 1,000 values are all `undefined` legitimately emits key 1,001, and truncating only the first 1,000 would let that one ship at full length. The walk now counts the same way the encoder does:

```ts
if (key && !isNullish(boundedItem)) {
  emittable++
}
```

so it stops where the encoder's output stops, not where its iteration stops. Keys past that point aren't copied, which is fine because the encoder cannot emit them either.

Empty keys are excluded from the count for the same reason — the encoder drops them without charging a slot.

Worth noting this is a bound on work, not a leak: everything past the cap was being read but was never going to ship. Regression test asserts the read count in `span.spec.ts`.

packages/core/src/traces/span.ts
turnipdabeets: Confirmed and fixed in 308c686c6.

Reproduced: `setAttribute('payload', { nested })` where `nested = { body: 'ok' }` with a limit of 4, then mutating `nested.body = 'abcdefgh'` after the call — all eight characters encoded. The copy held the caller's object (`out.nested === nested` was `true`).

The cause is `return changed ? bounded : value`, and it's broader than the top level: any nested object that needed no truncation was referenced rather than copied, so the mutation didn't even have to be at the root. As you say, without a hook there's no end-time pass to catch it.

Bounded values are now always copied — `return bounded` for objects, the accumulated array for arrays — so the span owns its own snapshot, which is the semantics OTel gives `setAttribute`. That drops the "an array that needed nothing allocates nothing" optimisation, deliberately: the allocation is bounded by the encoder's own item cap, and the alternative is the span shipping whatever the caller's object happens to hold at flush time. The `changed` flag is gone with it.

Test mutates after `setAttribute` and asserts `ok` still ships; it fails against the previous code.

packages/node/src/traces-defaults.ts
turnipdabeets: Confirmed the behaviour, fixed the silence, but pushed back on failing fast — 308c686c6.

Verified both shapes: `beforeSpanSend: 'redact'` resolves to zero hooks, and `[fn, null, 'oops']` resolves to one, in both cases with nothing logged. You're right that this is the worst kind of fail-open, because `beforeSpanSend` is *the* redaction point — a config that filters nothing exports exactly the values it was added to remove.

It now warns, naming the count so a partially-valid array isn't mistaken for a working one:

```
beforeSpanSend: ignoring 1 of 2 entries that are not functions. Spans export without them, so whatever they were redacting is not redacted.
```

I didn't make it fail fast. `resolveTracesConfig` runs from the `PostHog` constructor, so throwing takes down the application at startup over a misconfigured *optional* hook — and PostHog SDKs are expected never to throw into the host. Dropping every span (calling a non-function) is worse still, which is what the existing filter avoids. So: keep the filter, keep tracing up, make the loss loud. If you'd rather this were `error` than `warn`, easy change.

This required threading the logger through `resolveTracesConfig` (optional third parameter, passed `this._logger` from `client.ts`) — it had no way to say anything before. Two tests: one asserting the warning, one asserting a fully valid config stays quiet.

packages/core/src/traces/span.ts
```diff
@@ -275,22 +556,206 @@ export function inertSpan(options?: { parent?: unknown; tracestate?: unknown }):
...
+    }
+    // The representation the value defines for itself is what the encoder puts
+    // on the wire, so it is what has to be bounded — a `toJSON` returning a
+    // megabyte of text is invisible to a walk over the object's own keys.
+    const resolved = resolveToJson(value)
+    if (resolved.selfDescribed) {
+      // Resolving to nothing is the value's answer. Walking its keys anyway
+      // would build a plain object the encoder no longer treats as
+      // self-describing, putting the internals of a redacted value on the wire.
+      return isNullish(resolved.value) ? value : truncateValue(resolved.value, maxLength, state, depth + 1)
```
dustinbyrne: **Blocking: stateful `toJSON()` can bypass the configured value limit.**

When the first call returns `null` or `undefined`, this returns the original self-describing object. OTLP encoding later invokes `toJSON()` again, so a stateful serializer can return an oversized value on that second call.

With a limit of 4, I reproduced a serializer returning `null` first and a 100-character string second: it was called twice and all 100 characters were encoded.

Please materialize a stable, wire-equivalent representation from the first resolution so the encoder cannot invoke the original object again, and add a regression that runs through `buildOtlpSpan`.

packages/core/src/traces/index.ts
```diff
@@ -49,6 +66,41 @@ function looksLikeSpan(value: unknown): boolean {
...
+ * Whether a `beforeSpanSend` return value still carries the two collections the
+ * rest of the pipeline reads. An array is rejected for `attributes`: it would
+ * encode as `{ "0": ... }` rather than fail.
+ */
+function isSpanRecordShape(record: SpanRecord): boolean {
+  return (
+    !!record.attributes &&
+    typeof record.attributes === 'object' &&
+    !Array.isArray(record.attributes) &&
+    Array.isArray(record.events)
```
dustinbyrne: **Blocking: incomplete hook results are treated as valid span records.**

This check only establishes that `attributes` is object-like and `events` is an array. For example:

```ts
beforeSpanSend: () => ({ attributes: {}, events: [] } as any)
```

passes validation and is exported as an `unknown` span with fallback timing and preserved internal identity instead of being dropped.

Please construct the public `SpanRecord` projection before invoking the first hook, validate its required public fields after replacements, and attach internal propagation/bookkeeping fields only after the hook chain.

packages/core/src/traces/config.ts
```diff
@@ -75,6 +115,10 @@ export function resolveTracesConfig(
     serviceVersion: (resourceAttributes?.['service.version'] as string | undefined) ?? config?.serviceVersion,
     environment: (resourceAttributes?.['deployment.environment'] as string | undefined) ?? config?.environment,
     resourceAttributes,
+    beforeSpanSend: resolveBeforeSpanSend(config?.beforeSpanSend, logger),
+    maxAttributesPerSpan: positiveInteger(config?.maxAttributesPerSpan, DEFAULT_MAX_ATTRIBUTES_PER_SPAN),
+    maxEventsPerSpan: positiveInteger(config?.maxEventsPerSpan, DEFAULT_MAX_EVENTS_PER_SPAN),
+    maxAttributeValueLength: positiveInteger(config?.maxAttributeValueLength, DEFAULT_MAX_ATTRIBUTE_VALUE_LENGTH),
```
dustinbyrne, line 127: **Blocking: fractional options do not use the documented defaults.**

`Math.floor()` makes any finite positive fraction valid even though these options require positive integers. I reproduced:

- `maxAttributesPerSpan: 1.5` → `1`
- `maxEventsPerSpan: 2.5` → `2`
- `maxAttributeValueLength: 3.5` → `3`

The tracing specification says a numeric knob that is not a positive integer must use its documented default. Please reject fractions with `Number.isInteger()`—using a strict helper just for the new limits if legacy batch coercion must remain—and add fractional test cases.

packages/core/src/traces/index.ts
```diff
@@ -408,6 +496,183 @@ export class PostHogTraces {
...
+    // is not what gets exported. A frozen return refuses every write, and the
+    // span must survive that.
+    try {
+      restoreField(hooked, 'traceId', original.traceId)
+      restoreField(hooked, 'spanId', original.spanId)
+      restoreField(hooked, 'parentSpanId', original.parentSpanId)
+      // A hook that rebuilds the record instead of spreading it would otherwise
+      // drop tracestate, which is not part of the record the hook is handed.
+      restoreField(hooked, 'traceState', original.traceState)
+    } catch {
```
dustinbyrne, line 695: **Non-blocking: a later hook can observe identity forged by an earlier frozen result.**

If an earlier hook returns a frozen replacement with changed or omitted identity, these restoration writes fail and are swallowed. Final export uses the snapshot, but the next hook still receives the incorrect identity, despite identity being documented as immutable across the chain.

Passing a fresh corrected public projection to each subsequent hook would preserve the intended contract without mutating hook-owned objects.

packages/types/src/traces.ts
```diff
@@ -239,6 +278,68 @@ export interface TracesConfig {
...
+     * @default 128
+     */
+    maxAttributesPerSpan?: number
+
+    /**
+     * Maximum events on a single span. On overflow the earliest events are kept
+     * and later ones are dropped, with the number dropped reported on the
+     * exported span.
+     *
+     * Once the cap is spent, up to four more `exception` events are still
```
dustinbyrne: **Contract decision: should `maxEventsPerSpan` be an absolute maximum?**

The canonical tracing specification says a span carries at most the configured number of events, but this documents and implements four additional slots. Because eligibility is based only on the name `exception`, caller-created `addEvent('exception')` events receive the reserve too.

Please align the implementation and specification before release. If only SDK-generated `recordException()` events should receive priority, they need an internal distinction rather than the public event name.

dustinbyrne (dismissed):
Requesting changes for the three confirmed correctness issues called out inline. Each is reachable through the public tracing API and was reproduced against `8997a166`:

- stateful `toJSON()` can bypass `maxAttributeValueLength`;
- incomplete hook return values can be exported as malformed spans;
- fractional numeric options are floored instead of using their documented defaults.

The broader ownership looks sound: live spans enforce limits on write, limits are reapplied after hooks, queue/retry/consent remain in `PostHogTraces`, and cross-signal flushing remains in `PostHogBackendClient`.

I also called out one non-blocking chained-hook identity edge and one contract decision concerning the exception-event reserve.

Changeset provenance looks correct: the hook, limit, and stack-trace changesets belong to this PR; the other tracing changesets are inherited from #4579. The branch is currently stacked cleanly and does not need a rebase at the reviewed commit.

<sub>Agent-assisted review provenance: Dustin requested an independent review plus changeset and stacking checks without seeding suspected code defects. We discussed the resulting findings and architecture during the review. Dustin has been good to the agent.</sub>

ioannisj (approved):
Approving to unblock one more review cycle but https://github.com/PostHog/posthog-js/pull/4584#discussion_r3931037776 is my finding as well, so worth addressing

packages/core/src/traces/span.ts
turnipdabeets: Confirmed and fixed in 39fb86109, with the end-to-end regression in 3f95aeb3c.

Reproduced exactly as described: a serializer returning `null` first and `'x'.repeat(100)` second, with a limit of 4, was called twice and encoded all 100 characters.

The cause is that a nullish resolution returned the *original object*, and the encoder probes `toJSON` again on whatever it is handed. That return was deliberate — walking the object's own keys instead would hand the encoder a plain object it no longer recognises as self-describing, putting the internals of a value that redacts itself on the wire — so the fix had to keep that property while leaving nothing to re-probe.

It now stores the string the encoder builds from the same nullish result: `String(resolved.value)`, so `"null"` or `"undefined"`. That is byte-identical to what shipped before on every non-stateful value, which I checked in all three positions:

- `class Redacted { toJSON() { return null } }` → `{"stringValue":"null"}`, unchanged, secret absent
- `{ toJSON: () => undefined }` → `{"stringValue":"undefined"}`, unchanged
- an invalid `Date` → `"Invalid Date"`, unchanged (it never reaches this branch; the `instanceof Date` arm above returns first)

The marker is not itself truncated, matching `[Circular]` and `[Unserializable]` — it is nine characters at most, and trimming it to `unde` at a low limit would only make it unreadable.

Two tests: one asserting the resolution count and the stored value, and the one you asked for running through `buildOtlpSpan`, which is where the second call would have come from. Both fail against the previous code — the end-to-end one with `expected 2 to be 1`.

Two existing tests pinned the old return and now assert the new one; the redaction test asserts the secret is absent from what the span kept rather than asserting the object's identity, which is what it was actually there to protect.

packages/core/src/traces/index.ts
turnipdabeets: Confirmed and fixed in 28883b011.

Reproduced with your exact hook. It exported:

```json
{"traceId":"1b511b1d…","spanId":"6cfbe79b…","name":"unknown","startTimeUnixNano":"…","flags":257}
```

— no attributes at all, so no `posthogDistinctId` and no `sessionId`, joinable to nobody, and silently.

`isSpanRecordShape` now requires every field the public `SpanRecord` declares as required — `name`, `kind`, `startTime`, `endTime` alongside the two collections — and a return value missing any of them is a counted drop, which is the fail-closed rule the rest of the hook contract already follows.

One deliberate difference from what you suggested. I did not restructure into a pre-hook projection: the rebuild already takes identity, propagation and the dropped counters from the snapshot rather than from the hook return value, so the internal fields were never the leak here — the shape check was simply too narrow. Adding the missing validation was the whole gap, and it keeps the diff to one predicate.

The check reads **presence, not usability**. A field that is there holding the wrong type is a hook editing a real record badly, and the sanitising below (`sanitizeName`, `toEpochMs`, the status-code check) is what answers that. A field that is absent means the hook returned something that was never a span record, and those fallbacks would dress it up as one. That line keeps every legitimate hook working, including one that sets `startTime` to a `Date`.

Five new table cases (your example, plus each required field omitted individually), one asserting the drop is warned rather than lost, and one asserting an ordinary scrub still exports with its attributes intact. All six fail with the predicate reverted.

packages/core/src/traces/config.ts
turnipdabeets: Confirmed and fixed in b4f08b739. Reproduced all three of your cases: `1.5` → `1`, `2.5` → `2`, `3.5` → `3`.

I went further than the narrow fix you offered as a fallback, and it is worth flagging because it reaches beyond this PR's own knobs.

`positiveInteger` now rejects any non-integer rather than flooring it, so **all eight** traces options resolve the same way — the three limits here plus `flushIntervalMs`, `maxExportBatchSize`, `maxQueueSize`, `maxLiveSpans` and `maxSpanAgeMs`, which are #4579's. I did not add a strict helper for the new limits only:

- Nothing here has shipped. #4579 has not released, so there is no legacy coercion to preserve — the "if legacy batch coercion must remain" branch of your suggestion does not apply yet, and this is the last moment it is free.
- Two coercion rules in one resolver is a trap. The next knob added picks whichever helper it happens to be next to, and no comment survives to explain which was intended.
- The spec rule is about numeric knobs generally, not about these three.

This does change one of #4579's documented expectations: a test asserting `maxExportBatchSize: 10.9` floors to `10` now asserts it takes the default of `512`. That is the same defect, and I would rather it not ship. Calling it out explicitly since it is outside the surface you were reviewing — happy to split it back out if you or @ioannisj would rather #4579's knobs keep flooring.

Fractional cases added to the per-span cap table, the `flushIntervalMs` table, and a new one for the live-span bounds; the batch-size test rewritten. Five fail with `Math.floor` restored.

packages/core/src/traces/index.ts
turnipdabeets: Confirmed and fixed in da5384f4b, even though you rated it non-blocking — it was cheap and the contract is documented.

Reproduced: with a first hook returning `Object.freeze({ ...span, traceId: '0'.repeat(32), spanId: 'f'.repeat(16) })`, the second hook read

```
{ traceId: "00000000000000000000000000000000", spanId: "ffffffffffffffff" }
```

while the export used the real ids, exactly as you described.

`_keepSpanIdentity` now hands the next hook a corrected view when the restoring writes are refused. Built from the record's own descriptors rather than a spread, with only the four identity descriptors replaced:

```ts
const descriptors = Object.getOwnPropertyDescriptors(hooked)
for (const field of ['traceId', 'spanId', 'parentSpanId', 'traceState'] as const) {
  descriptors[field] = { value: original[field], enumerable: true, writable: true, configurable: true }
}
return Object.create(Object.getPrototypeOf(hooked), descriptors)
```

That was the reason not to take the obvious copy: a spread reads own properties only, so a hook returning a class instance exposing `events` through a prototype getter would lose the span to the fail-closed branch — the same trap the post-hook rebuild already had to avoid. Going through the descriptors keeps the prototype (so `instanceof` and prototype getters still answer) and keeps `Object.keys` reading what it read before, while making the identity fields writable where the original was frozen.

It runs only on the failure path, so an unfrozen record is untouched. The descriptor read is itself guarded — a hostile descriptor costs the next hook a correct id, not the span, since the export reads the snapshot regardless.

Two tests: one asserting the second hook sees the exported ids, one asserting it still reads the previous hook's `name`, `attributes` and key set. The first fails with the fix reverted.

packages/types/src/traces.ts
turnipdabeets: Taking this one as: keep the reserve, and fix the spec rather than the code. Reasoning below, and I checked the thing that would have changed the answer.

**The spec has a gap here rather than a rule this violates.** *Span limits* caps "**user-supplied** per-span content" — `user-supplied` is bolded in the spec's own text — and then explicitly exempts SDK-managed auto-context keys from the attribute cap, on the grounds that they are the product's join keys. It never writes down the events half: there is no sentence about content the SDK attaches on the caller's behalf, and no rule for what a span at the event cap does when it then throws. Meanwhile *Scoped span helpers and exception recording* says `recordException` SHALL attach an `exception` event, with no "unless the span is full" carve-out. So the two requirements are in tension with each other, not just with this implementation.

The behaviour the reserve protects is the one worth keeping: a span that fills its events and then throws is marked `error` with no type, message or stack — the span you most want to read, gutted exactly when it matters.

**On whether 132 is a problem on the wire — it isn't, and I checked rather than assumed.** The ingestion service (`rust/capture-logs`) validates one limit, `MAX_REQUEST_BODY_SIZE_BYTES = 2097152`. There is no per-span event count validation anywhere in it, and its Avro schema carries `droppedEventsCount` documented as "Number of events that were dropped due to limits" — the server's model is that the SDK caps and reports, not that the server enforces a count.

Measured, at the default `maxAttributeValueLength` of 8192:

| | size |
|---|---|
| 128 events, one bounded attribute each | 1038 KB |
| the same span plus eight `recordException` calls (4 kept) | 1051 KB |
| **cost of the reserve** | **13 KB, 0.61% of the 2 MB cap** |

And the event cap is not what bounds a span anyway — event *attributes* are capped in value length but not in count, so 128 events carrying ten bounded attributes each is already **10.06 MB**, five times the cap, at exactly 128 events. For the reserve to be what pushes a span over, the span would have to land within 13 KB of the limit by coincidence. It cannot introduce the failure mode; a span that big is lost at 128 too.

**On keying the reserve by name.** You are right that it is the wrong key in principle. I looked at making it internal and decided against it, because it does not actually buy the guarantee: the caps are re-applied after `beforeSpanSend`, and at that point events are plain `{ name, timestamp, attributes }` objects on the public `SpanRecord`. There is nowhere to carry an SDK-generated marker that a hook rebuilding the `events` array would not drop — and a rebuilding hook is exactly the case in the PR's own docs. So the mechanism would hold on the span side, silently degrade to name-matching after a hook, and still need the same spec sentence, since 132 > 128 either way. That is cost without the property.

So the trade-off stands as documented, and it is the same one the attribute exemption already makes and documents.

I am drafting the `sdk-specs` change now — a sentence in *Span limits* exempting SDK-attached exception events from the event cap, bounded, plus a scenario for a full span that throws — so "align before release" has something concrete to land against rather than staying an open question here. Will link it on this thread.

turnipdabeets: Reversing course on this after Anna pushed back, and correcting something I got wrong above.

**The correction.** I said an internal marker would "silently degrade to name-matching after a hook". That is wrong. Keyed on a marker, an event that arrives without one falls back to an *ordinary* slot — it never claims the reserve. So the degradation is fail-safe, not silent over-admission, which was the argument I built the rest of the paragraph on. With that corrected, your original suggestion is simply the better design, and it is now implemented in b8425c808.

**What changed.** SDK-recorded exception events carry a symbol-keyed marker on the internal event record. Verified rather than assumed:

- invisible to `Object.keys`, `for...in` and `JSON.stringify`, so `beforeSpanSend` cannot see, forge or delete it
- absent from the encoded span — asserted against `buildOtlpSpan` output
- carried through an object spread, which is what the post-hook re-sanitising pass uses, so the two enforcement points reach the same verdict
- `SpanEventRecord` is core-internal and in no public export, so this adds nothing to the published surface

The public `SpanRecord` is untouched. This follows the split the code already documents on the internal record: *"a field added here rather than to the public record is a field `beforeSpanSend` cannot see, and so cannot corrupt."*

**It is a net deletion.** `eventNameBound()` existed only so the literal name `exception` survived truncation at a small `maxAttributeValueLength` — a short bound would otherwise have disabled the reserve. With eligibility on provenance the name decides nothing, so the function, its export, its import and both call sites are gone, and the JSDoc no longer has to warn that a caller's own `exception` event gets the reserve too. That warning is now false, which is the point.

**The residual, pinned rather than discovered.** A hook that rebuilds its events into fresh objects field by field drops the marker, and those exceptions fall back to ordinary slots. `.filter()` and `.map(e => ({ ...e }))` both keep it; only explicit field-by-field reconstruction loses it. There are four tests covering the boundary — in-place edit keeps the reserve, spread keeps it, field-by-field rebuild loses it, and an `exception` event a hook *appends* never gets it.

Six new tests in total, and the two that pinned name-based eligibility now assert the opposite.

**Spec.** Drafted, and it no longer needs the "MAY match by name" escape hatch your comment was aimed at — it now requires provenance and forbids deciding on the name. I checked the implementation against each scenario before writing them down, which caught one of my own scenarios being wrong (`droppedEventsCount` was 1 where the implementation correctly gives 4). Will link the `sdk-specs` PR here once it is up.

turnipdabeets: Sorry for noise, I’m going to drop this because its an over optimization we dont know we even need. We can revisit if we need more headroom for exceptions when that time comes. For now I think this is fine.

turnipdabeets: Final answer on this thread, and it lands on your side of it: **the reserve is gone.** `maxEventsPerSpan` is now an absolute cap, in 0bb43aa55.

Anna pushed back on whether the complexity was worth it, and once I costed the removal honestly it wasn't:

- Two rounds of review went into the reserve and it produced two defects of its own — first a reserve that was really a smaller exclusive cap, then the name-based eligibility you flagged here.
- Nobody can show the case occurs. It needs one span holding 128 events *and* then throwing. Traces has not shipped, so there is no usage to appeal to, and I was defending it on a story rather than on numbers.
- **Adding it back later is purely additive.** No caller breaks when a cap grows; shipping it now and removing it later would be a behaviour change. The cheap direction was to defer.
- The case stays measurable without it. A span that fills up and throws still exports `status: error` with a non-zero `droppedEventsCount`, so `status = error AND droppedEventsCount > 0` finds exactly this population once traces is live. If it turns out to be real it comes back with evidence behind it.

Removing it deleted 95 lines and the duplication you would have been right to flag next: the rule no longer has two independent implementations in the span writer and the post-hook re-apply, so they cannot drift. `eventNameBound`, the provenance marker and the reserve constant are all gone; `applySpanLimits` is a plain cap and `addEvent` is a counter check.

**On aligning the spec.** No `sdk-specs` change is needed to ship this, which was the substance of your ask. *Span limits* says at most 128 events per span and the implementation now says exactly that.

One thing I'd rather name than quietly leave: the spec is still ambiguous underneath. *Scoped span helpers and exception recording* says `recordException` SHALL attach an `exception` event, with no carve-out for a full span, so on a span at the cap the two requirements still cannot both hold. The difference is that node now resolves it toward the number the spec states explicitly, which is the conservative reading and needs nobody's permission — rather than extending past a stated limit, which did. It is worth one clarifying sentence in the spec before the next SDK ports traces, but it no longer blocks this PR. Happy to file that separately if you think it's worth doing now.

Public JSDoc on `maxEventsPerSpan` now documents the cap as absolute and points at raising it for spans that record many events and can also fail. Six tests changed to assert the drop rather than the reserve.

Thanks for pushing on this one — the version that ships is smaller and more honest than the one you first reviewed.

packages/core/src/traces/span.ts
```diff
@@ -92,27 +124,72 @@ export class PostHogSpan implements Span {
...
+      }
+      this._userEventCount++
       this._events.push({
-        name: sanitizeName(name, 'Span event name', this._logger),
+        name: sanitizeName(name, 'Span event name', this._maxAttributeValueLength, this._logger),
         timestamp: resolveSuppliedTime(timestamp, this._now(), 'event timestamp', this._logger),
         // Copied so a caller reusing one object across events can't mutate a recorded one.
-        ...(attributes && { attributes: assignUserAttributes({}, attributes) }),
+        ...(attributes && {
+          attributes: truncateAttributes(assignUserAttributes({}, attributes), this._maxAttributeValueLength),
```
turnipdabeets: `addEvent` caps how many events a span may hold (`_userEventCount >= this._maxEvents`) and truncates each event attribute's *value* to `maxAttributeValueLength`, but nothing caps how many attributes an event carries. `truncateAttributes` walks every key and truncates values in place; it never drops one.

So the per-span bound is `maxEventsPerSpan` × **unbounded** × `maxAttributeValueLength` — 128 × n × 8192. `config.ts:18` says the caps keep "a span at the attribute cap under 1 MB, comfortably inside the 2 MB body cap", which holds for span attributes (`_maxAttributes`, enforced at `span.ts:147`) but not for event attributes.

Why it matters beyond tidiness: this is the hole @jonmcwest identified in the #4579 review — a single span big enough that the 413 path shrinks the batch to one span, that span is still too big, and it is dropped whole. That's the case this PR's limits exist to close. Measured on the current `feat/traces-node-mvp` behaviour, one 12 MiB span among 511 healthy ones costs 42 requests and 120.3 MiB uploaded, drops `maxExportBatchSize` from 512 to 33, and needs 479 healthy batches to ramp back to 512.

Suggest a `maxAttributesPerEvent` (OTel's `AttributeCountPerEventLimit`, default 128) applied in `addEvent` alongside the existing value truncation, with the dropped count reported the way `droppedAttributesCount` already is.

Flagging it because the decision on whether posthog-js keeps its pre-send body-size check (#4726) depends on whether this PR actually bounds a span. As it stands it doesn't, so the check is still doing real work.

turnipdabeets: Correcting my own suggestion above: a `maxAttributesPerEvent` of 128 is right for OTel parity, but it does **not** bound a span under the body limit, and I implied it would.

128 events × 128 attributes × 8192 bytes ≈ **128 MiB**, plus ~1 MiB of span attributes — still an order of magnitude over the limit, before JSON key and quoting overhead. Bounding a span under 10 MiB through per-event counts alone would need roughly 10 attributes per event, which is too tight to be useful.

Count caps limit a span's *shape*; only a byte budget limits its *size*. If the goal is the guarantee `config.ts:18` already claims ("keeps a span at the attribute cap under 1 MB"), that wants a per-span serialized budget rather than another count.

Doesn't change the suggestion — the uncapped event attribute count is still a real hole worth closing — only what it buys. And it means the pre-send body check in #4726 stays load-bearing either way, so nothing there is blocked on this.

PR #4586, opened 2026-08-20
feat(node): cap attributes and events per span

turnipdabeets:
## Problem

Stacked on #4579 — review that first.

Nothing bounded how much a single span could carry. A loop that calls `span.addEvent()` per iteration, or an instrumentation layer that copies every request header onto a span, grows it without limit until the ingestion endpoint rejects the payload as too large — at which point the 413 path shrinks the batch to a single span, that span is still too big, and it is **dropped entirely with a warning**. You lose the whole span rather than the excess.

Implements the `Span limits` requirement of the traces capability spec.

## Changes

Caps user-supplied content per span, at OpenTelemetry's defaults.

```ts
traces: {
  serviceName: 'checkout-api',
  maxAttributesPerSpan: 128, // default
  maxEventsPerSpan: 128,     // default
}
```

- **Earliest wins.** The first 128 attributes and events are kept; later ones are dropped silently.
- **The counts ship with the span** as `droppedAttributesCount` / `droppedEventsCount`, so a truncated span is visibly truncated rather than quietly wrong. Both are omitted when nothing was dropped.
- **SDK-attached keys are exempt and never evicted.** `posthogDistinctId`, `sessionId`, `url.full`, `screen.name` and `app.state` do not count toward the cap, so a span at the limit still links back to its person and session — those are the join keys, and losing them to a noisy attribute loop would defeat the point of the span.
- **Overwriting an existing key always succeeds.** The cap counts distinct user keys, not writes, so `setAttribute('plan', …)` twice on a span at the cap updates rather than drops.
- Attributes passed to `startSpan` are capped the same way as ones set later.

### Reviewer notes

- **The two wire counters were deliberately removed from the public types in #4579** — they were declared but nothing populated them, which is documenting what does not exist. This PR is the code that fills them, so they return alongside it. That ordering was the point.
- **The spec also requires the caps to re-apply after `beforeSpanSend`**, because a hook can add attributes. `beforeSpanSend` lives in #4584, which is a sibling branch off the same base — neither PR can implement that half alone. Whichever merges second should wire it up; I have not tried to coordinate it inside either branch, since that would make each depend on the other.
- Enforcement lives in `PostHogSpan`, not at encode time, so an over-cap span never occupies memory in the first place — the point is to bound growth, not to trim on the way out.
- Auto-context keys are identified by what the SDK attached at span start, so a user who explicitly sets `posthogDistinctId` themselves gets the exempt treatment for that key. That seemed better than special-casing a hardcoded list.

### Verification

Six core tests covering both spec scenarios plus the boundary cases (overwrite-at-cap, start-time attributes, counters omitted when unused), and two node tests exercising the real client including the 128 default. `packages/core` 1003 pass, `packages/node` 951 pass, typecheck clean, lint clean, public API references unchanged.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

Additive from a consumer's perspective: two new optional config keys and two optional wire fields. `SpanInit` and `ResolvedTracesConfig` gain required members, but both ship for the first time in #4579 and neither is constructible by a published consumer.

Behaviour does change for a span that exceeds 128 user attributes or events — it is now truncated and marked rather than sent whole. Given the alternative today is the whole span being dropped by the endpoint, truncation is the more useful outcome, and the caps are configurable for anyone who disagrees.

### If releasing new changes

## 🤖 Agent context

Built with Claude Code, directed by @turnipdabeets, against the `Span limits` requirement in the traces capability spec.

Two decisions worth a look:

- **Cap at write time rather than at encode time.** Trimming during encoding would satisfy the spec's wording just as well and touch less code, but it lets a pathological span sit in memory at full size until flush. Bounding growth at the source is the behaviour the requirement is actually protecting against.
- **Exemption is by "what the SDK attached", not a hardcoded key list.** It stays correct automatically if auto-context ever gains a key, at the cost of also exempting a key the user happens to set with the same name as one the SDK attached. That felt like the right trade, but it is the kind of thing worth a second opinion.

### DIFF
--- a/.changeset/node-span-limits.md
+++ b/.changeset/node-span-limits.md
@@ -1,0 +1,7 @@
+---
+'posthog-node': minor
+'@posthog/core': minor
+'@posthog/types': minor
+---
+
+Cap spans at 128 user attributes and 128 events, configurable with `traces.maxAttributesPerSpan` and `traces.maxEventsPerSpan`.
--- a/packages/core/src/traces/index.spec.ts
+++ b/packages/core/src/traces/index.spec.ts
@@ -18,6 +18,8 @@ const resolveForTest = (partial?: Partial<ResolvedTracesConfig>): ResolvedTraces
   flushIntervalMs: 5000,
   maxExportBatchSize: 512,
   maxQueueSize: 2048,
+  maxAttributesPerSpan: 128,
+  maxEventsPerSpan: 128,
   ...partial,
 })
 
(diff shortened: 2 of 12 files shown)


turnipdabeets:
Folded into #4584.

The spec requires `The caps SHALL be re-applied after beforeSpanSend runs` — and neither branch could satisfy that alone, since neither had both halves in its tree. Split, the best either could do was leave a note asking whichever merged second to remember, which is not tracked work.

The failure is not theoretical: the hook writes to the plain record rather than through the span's guarded writer, so a hook that enriches a span pushes it back past the cap with no trim and no dropped count, and it is then rejected as too large and lost whole.

The two branches also conflicted in six files, so they were never independently reviewable in practice.

Everything here moved to #4584 unchanged, plus the re-apply and four tests for it. The changeset is unchanged and still ships as its own entry.

PR #4588, opened 2026-08-20
fix(replay): keep backdated idle lifecycle markers out of rotation-born sessions

clr182:
## Problem

A tab suspended for a long time emits nothing while backgrounded, so idle is only detected on wake — by which time an app-level capture may already have rotated the session. Two lifecycle events then landed in the rotation-born session with timestamps from before it existed, dragging the recording's start back by the whole idle gap. The player reports that prefix as unplayable: *"The first 1h 31m of this recording can't be shown — the initial snapshot of the screen arrived late."* Diagnosed from a customer recording (support ticket #68226) where a single backdated `sessionIdle` marker started the session 1h31m before its first full snapshot.

## Changes

- `sessionIdle` markers now pin their target session/window ids in the payload at emit time (same pattern as `$session_ending`) and skip the nested session check, so the backdated marker is captured into the old epoch's buffer before any rotation adoption. Falls back to the old routing when the payload carries no ids.
- `$session_starting` is no longer restamped to the old session's `lastActivityTimestamp` — it ships to the *new* session, so it keeps its emit-time stamp (the rotation moment). Only `$session_ending` keeps the backdating.
- New regression test simulates the wake sequence (confirmed active → 97min suspension → wake rotation) and asserts the marker ships with the old session and nothing in the new epoch predates the rotation; fails without the fix.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Authored with Claude Code (Fable 5) while investigating support ticket #68226. The root cause was pinned down by analysing the affected recording's raw export: the only content in the "unplayable" prefix was one `sessionIdle` marker stamped `lastActivity + threshold`, in a session whose UUIDv7 creation time matched the first full snapshot. The regression test's timestamp invariant then surfaced the second vector (`$session_starting` backdating), which was fixed the same way rather than clamped, since the backdating is only correct for the session that actually went idle.

### DIFF
--- a/.changeset/idle-marker-session-attribution.md
+++ b/.changeset/idle-marker-session-attribution.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+fix(replay): attribute the backdated sessionIdle marker to the session that went idle, so a rotation-born session's recording no longer starts hours before its first snapshot
--- a/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
+++ b/packages/browser/src/__tests__/extensions/replay/lazy-sessionrecording.test.ts
@@ -1270,28 +1270,47 @@ describe('Lazy SessionRecording', () => {
                 expect(bufferedEvent.data.tag).toBe(eventTag)
             })
 
-            it.each(['$session_ending', '$session_starting'])(
-                'corrects timestamp for %s events when idle',
-                (eventTag: string) => {
-                    const lastActivityTime = startingTimestamp + 100
-                    const eventRecordedTime = startingTimestamp + 5000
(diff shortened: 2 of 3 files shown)


ksvat (approved):
ran the incident skill locally, all good

PR #4591, opened 2026-08-20
docs: document posthog-js release flow

dustinbyrne:
## Problem

`RELEASING.md` still described the old failure mode where npm could publish before the browser S3/CDN release failed. The release workflow now gates all coordinated npm publishing on successful US and EU browser uploads, and the protected recovery path was undocumented.

## Changes

- Document the normal changeset, browser S3/CDN, npm, GitHub release, and downstream-upgrade sequence.
- Explain that browser changes must upload successfully to both regions before any npm publishing, while releases without a browser change skip S3.
- Document when and how to use manual S3 recovery, including its inputs, source trust checks, approval boundaries, regional non-atomicity, overwrite behavior, and verification steps.
- Include the resolved SDK package version in the recovery approval summary and Slack message so reviewers can identify an intentional source-version/target-version mismatch.
- Render the recovery request with Slack Block Kit and the same primary **Approve/Reject on GitHub** button used by normal release approvals.

This updates internal release tooling and documentation and does not require a package changeset.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## Validation

- Prettier and `git diff --check`
- Actionlint syntax validation for both release workflows
- Actionlint with ShellCheck for the changed recovery workflow
- Workflow source-version output and Slack Block Kit payload assertions
- Verified recovered `posthog-js@1.418.2` representative assets in US and EU
- Compared both regions' cache-busted `array.js` responses with the recovery workflow artifact

## 🤖 Agent context

Pi updated the release documentation and approval message based on the implemented workflows. Fresh builtin reviewers checked the operational instructions against the workflow and release tooling, confirmed earlier ambiguity in mismatch and verification guidance was addressed, and reviewed the final Block Kit payload with no actionable findings.


dustinbyrne:
Incident-risk review (Class 7: release and delivery integrity): this is a path-only match, not the prior failure mode. The workflow change only carries the already-validated `source_version` into the approval summary and Slack message. It does not add `pull_request_target`, change any checkout ref, add lifecycle scripts, expose credentials, alter OIDC/protected environments, or change CDN paths, CORS, cache headers, or served bundle bytes. Existing CODEOWNERS and release authorization boundaries remain unchanged.

RELEASING.md
```diff
@@ -1,34 +1,145 @@
...
+| ----------------------- | ------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `target_version`        | empty   | Required S3 destination version, for example `1.418.2`.                                                                                                                                                                                      |
+| `source_sha`            | empty   | Full `PostHog/posthog-js` commit to build. When empty, resolves the `posthog-js@<target_version>` tag. A version that never received a tag requires its full version-bump commit SHA. The commit must be current `main` or an ancestor.      |
+| `toolbar_sha`           | empty   | Full `PostHog/posthog` commit used to build the toolbar. When empty, uses current `master`. The commit must be current `master` or an ancestor. Prefer the toolbar commit recorded in the original failed run when reproducing that release. |
```
turnipdabeets: [suggestion] Could we say where that commit is recorded? The original run prints it in the `Resolve toolbar source` job summary — without that pointer the instruction isn't followable. Worth noting alongside it that normal releases build the toolbar from `PostHog/posthog@master` HEAD at release time rather than a commit pinned to the SDK version, since that's what makes reproducing a past release need this input at all.

RELEASING.md
```diff
@@ -1,34 +1,145 @@
...
+
+If any of that work fails, no package from the coordinated npm release is published. If `posthog-js` did not change, the S3 jobs are skipped and npm publishing proceeds normally.
+
+A stable browser release uploads:
+
+- immutable versioned assets under `/static/<version>/`;
+- mutable major-version aliases under `/static/<major>/`; and
+- top-level compatibility aliases under `/static/`.
+
+Prerelease versions receive only immutable versioned assets. The workflow does not purge CDN caches, so mutable aliases can continue serving cached bytes until their normal TTL expires.
```
turnipdabeets: [suggestion] Could we put the actual numbers here? `upload-posthog-js-s3.ts` sets `public, max-age=300` on the mutable aliases and `public, max-age=31536000, immutable` on the versioned assets, so "normal TTL" is 5 minutes. That turns "wait for the TTL" into an answerable "when do users get the fix?" — same for the `cache TTL` mention further down under Verifying recovery.

RELEASING.md
```diff
@@ -1,34 +1,145 @@
...
+
+PostHoggers can join [`#alerts-posthog-js`](https://posthog.slack.com/archives/C07HTMN9X47) for release workflow failure notifications.
+
+## Manual S3 recovery
+
+S3 recovery is an exceptional path for repairing a failed or incomplete browser release. If the failure was transient and the original run is still safe to resume, first retry its failed jobs. Use recovery when the original run cannot be completed safely, for example when one regional upload succeeded before the other failed or an existing release is missing immutable assets.
+
+Recovery always rebuilds and uploads the browser SDK and toolbar. It is not an npm-only, alias-only, artifact-promotion, rollback, or S3 deletion mechanism.
```
turnipdabeets: [question] This rules recovery out as a rollback mechanism, which is right, but leaves open what we do instead when a bad version is already live. Is the answer ship a new patch + `npm deprecate`, with alias rollback done by recovering the *previous* version with `update_latest_aliases=true` and `force_overwrite=true`? If that's sanctioned a line saying so would help; if it isn't, worth saying that explicitly so nobody improvises it during an incident.

turnipdabeets (approved):
LGMT, just a few suggestions

RELEASING.md
dustinbyrne: Addressed in afe9dc1e2. The `toolbar_sha` docs now point to the original run’s **Resolve toolbar source** job summary and explain that normal releases resolve `PostHog/posthog@master` at release time instead of pinning toolbar source to the SDK version.

RELEASING.md
dustinbyrne: Addressed in afe9dc1e2. The release docs now give both cache policies explicitly (`max-age=300` for mutable aliases and `max-age=31536000, immutable` for versioned assets), and the recovery verification step says to allow up to five minutes for alias caches.

RELEASING.md
dustinbyrne: Addressed in afe9dc1e2. The sanctioned response is to ship a corrected patch and deprecate the bad npm version if needed. Repointing mutable aliases to an earlier release through recovery is explicitly not sanctioned.

PR #4592, opened 2026-08-20
feat(browser-common): expose installed extensions by name

dustinbyrne:
## Problem

Independently loaded browser extensions need a stable way to resolve other installed extensions without depending on private `PostHog` properties or restoring the removed provider/capability registry. Surveys specifically need to resolve feature flags when a newly deployed `surveys.js` is loaded by either a new core or an older cached `array.js`.

This PR is stacked on #4532; review only this PR's diff.

## Changes

- Add phantom-typed stable string tokens and `Client.getExtension(...)` overloads to `@posthog/browser-common`.
- Use the existing extension runtime name map for lookup, including during synchronous setup, and remove entries after failed setup or disposal.
- Delegate lookup through `BrowserClientAdapter` and expose it on `PostHog` for independently compiled lazy bundles.
- Give autocapture, feature flags, and surveys internal tokens typed to their concrete extension classes; the classes continue to implement the shared `Extension` lifecycle contract.
- Resolve survey feature flags through the installed extension first, with a fallback to `posthog.featureFlags` for older cached cores.
- Document stable-name lookup and add runtime, adapter, inference, lifecycle, survey, and declaration coverage.

### Validation

- Focused browser Jest suite: 164 tests passed.
- Browser-common Jest suite: 82 tests passed.
- The new `surveys.js` passed 9 Chromium Playwright survey tests with `array.js` from NPM `posthog-js@1.350.0`.
- Browser and browser-common TypeScript, ESLint, Prettier, `git diff --check`, the full `posthog-js` build, declaration bundling, property-mangling checks, and source-map checks passed.
- Bundle comparison against the surveys parent: `array.js` +30 gzip bytes; `surveys.js` +43 gzip bytes.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [x] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent and reviewed by fresh-context builtin reviewer agents. The human directed the lifecycle constraints, requested compatibility with `posthog-js@1.350.0`, chose concrete extension types over parallel API interfaces, and kept built-in tokens internal rather than adding new public exports. The human has been good to the agent.

### DIFF
--- a/.changeset/browser-common-extension-lookup.md
+++ b/.changeset/browser-common-extension-lookup.md
@@ -1,0 +1,6 @@
+---
+'@posthog/browser-common': patch
+'posthog-js': patch
+---
+
+Add typed stable-name lookup for installed browser extensions and use it when independently loaded survey code resolves feature flags.
--- a/.changeset/browser-common-extension-lookup.md
+++ b/.changeset/browser-common-extension-lookup.md
@@ -1,5 +1,5 @@
 ---
-'@posthog/browser-common': patch
+'@posthog/browser-common': minor
 'posthog-js': patch
 ---
 
--- a/packages/browser-common/.agents/skills/develop-extension/SKILL.md
+++ b/packages/browser-common/.agents/skills/develop-extension/SKILL.md
@@ -56,6 +56,21 @@ export class MyExtension implements Extension {
 - `dispose()` is optional, synchronous, idempotent, and best-effort.
 - Static SDK configuration belongs in explicit constructor options, not in `Client`.
 
+When other extensions need its controls, export an implementation-free typed token from a small contract module:
+
+```ts
+import type { Extension, ExtensionToken } from '@posthog/browser-common'
+
+export interface MyExtensionApi extends Extension {
+    stop(): void
+}
+
+export const MyExtensionToken = 'myExtension' as ExtensionToken<MyExtensionApi>
+```
+
+The token string must match the implementation's `name`. Keep the interface and token separate from the implementation
+so consumers do not pull its code into their bundles.
+
 ## Client capabilities
 
 | Need                          | Use                                                        |
@@ -65,6 +80,7 @@ export class MyExtension implements Extension {
 | record an event               | `await client.capture(event, properties?, options?)`       |
 | add properties to every event | `client.registerDynamicEventProperties(() => ({ … }))`     |
 | react to finalized events     | `client.onEvent(({ event, properties }) => …)`             |
+| access an installed extension | `client.getExtension(MyExtensionToken)`                    |
 | call a PostHog endpoint       | `await client.sendRequest(path, init?)`                    |
 | react to server config        | `client.onRemoteConfig(…)`                                 |
 | persist small state           | `client.kv`                                                |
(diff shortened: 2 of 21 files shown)


dustinbyrne:
Review follow-up:

- **Replay incident-risk:** This PR does not change recording start conditions, remote-config endpoints or response shapes, authorized-domain checks, or proxy-dependent inputs. The only `posthog-featureflags.ts` changes type the existing `name` value through a stable string token (still exactly `featureFlags`) and explicitly implement the existing `Extension` contract, so recording fail-open/fail-closed and reverse-proxy behavior are unchanged.
- **Greptile missing-extension sentinel:** Keeping `undefined` is intentional. Lookup is backed by `Map.get`, the existing `@posthog/browser` registry also returns `undefined`, and shared client documentation/tests consistently define an absent or disposed extension as `undefined`.
- **Greptile feature-flags fixture:** Fixed in 317ed998c. The test now constructs a real `PostHogFeatureFlags` with a concrete config source and spies on its methods instead of casting a partial object.
- **Compatibility bot:** Its `captureException(Object) with name and message` failure is outside this PR’s touched extension path; this PR changes no exception code. Separately, the new `surveys.js` passed all 9 selected core-display and feature-flag eligibility tests with NPM `array.js` from posthog-js@1.350.0.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -150,6 +152,11 @@ export class SurveyManager {
         this._surveyInFocus = null
     }
 
+    private get _featureFlags(): PostHogFeatureFlags | undefined {
```
turnipdabeets, line 155: `_featureFlags` matches the mangle regex, so it needs to land in `terser-mangled-names.json` — the `Write mangled property names` job failed on 321ab211 asking for exactly `+ "_featureFlags"`, and neither file has changed since. I think we need `pnpm write-mangled-property-names` committed here. (CI hasn't run on the last two commits either, so a re-run is the real green signal.)

.changeset/browser-common-extension-lookup.md
```diff
@@ -0,0 +1,6 @@
+---
+'@posthog/browser-common': patch
```
turnipdabeets: [suggestion] `getExtension` is a required member on the exported `Client`, plus a new exported `ExtensionToken` — #4300 and #4316 both went minor for comparable `Client` additions.

```suggestion
'@posthog/browser-common': minor
```

packages/browser-common/.agents/skills/develop-extension/SKILL.md
```diff
@@ -56,6 +56,21 @@ export class MyExtension implements Extension {
...
+import type { Extension, ExtensionToken } from '@posthog/browser-common'
+
+export interface MyExtensionApi extends Extension {
+    stop(): void
+}
+
+export const MyExtensionToken = 'myExtension' as ExtensionToken<MyExtensionApi>
+```
+
+The token string must match the implementation's `name`. Keep the interface and token separate from the implementation
```
turnipdabeets, line 73: [nit] The built-ins avoid drift by assigning from the token (`readonly name = AutocaptureExtension`) — worth showing that here, since a drifted name returns `undefined` silently and reads as "not installed".

packages/browser-common/src/client.ts
```diff
@@ -111,6 +113,11 @@ export interface Client {
     /** Registers a synchronous producer of properties merged into every captured event. */
     registerDynamicEventProperties(producer: () => Record<string, unknown>): Disposable
 
+    /** Returns the extension registered under a typed stable name, or `undefined` when it is not installed. */
+    getExtension<T extends Extension>(token: ExtensionToken<T>): T | undefined
+    /** Returns the extension registered under a stable name, or `undefined` when it is not installed. */
+    getExtension<T extends Extension = Extension>(name: string): T | undefined
```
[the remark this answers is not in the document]
dustinbyrne: ??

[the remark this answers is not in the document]
dustinbyrne: bad bot

ioannisj (approved):

PR #4601, opened 2026-08-22
fix(rollup): upload Vite 8 client sourcemaps

### LINKED ISSUE #4596: @posthog/rollup-plugin 1.4.9+ silently skips Vite 8 client sourcemaps after Oxc strips chunkId comments
## Bug description

`@posthog/rollup-plugin` 1.4.9 introduced a regression for Vite 8 client builds using the default Oxc minifier.

PR #4512 moved chunk-ID injection into `renderChunk`. It injects both the executable `_posthogChunkIds` IIFE and a trailing `//# chunkId=...` comment.

Vite 8 minifies client output after that hook. The executable IIFE survives, but Oxc removes the trailing comment. During `writeBundle`, the plugin only includes chunks where `determineChunkIdFromSource(chunk.code)` finds that comment. The detector intentionally ignores the IIFE.

Consequently every client chunk is excluded, `filePaths` remains empty, and the build succeeds with:

```text
No chunks found, skipping sourcemap processing for this stage. Your build may be multi-stage and this stage may not be relevant
```

No client source maps are uploaded. SSR maps can still upload because Vite disables minification for server builds by default, making a multi-stage build appear partially successful.

This remains present on current `main` (`6ccc78e`):

- [Marker injection](https://github.com/PostHog/posthog-js/blob/6ccc78e028c9e375efff2560d2f422c86ab23d85/packages/rollup-plugin/src/index.ts#L50-L58)
- [Comment-only check in `writeBundle`](https://github.com/PostHog/posthog-js/blob/6ccc78e028c9e375efff2560d2f422c86ab23d85/packages/rollup-plugin/src/index.ts#L152-L183)
- [Detector intentionally ignores the IIFE](https://github.com/PostHog/posthog-js/blob/6ccc78e028c9e375efff2560d2f422c86ab23d85/packages/plugin-utils/src/chunk-ids.ts#L68-L95)
- [Vite client/server minification defaults](https://github.com/vitejs/vite/blob/v8.0.16/packages/vite/src/node/build.ts#L445)
- [Vite output comment and minification settings](https://github.com/vitejs/vite/blob/v8.0.16/packages/vite/src/node/build.ts#L747-L773)

## How to reproduce

1. Build a Vite 8.0.16 client app with `@posthog/rollup-plugin@1.5.0`, source maps enabled, and the default `build.minify`.
2. Inspect the generated code across hooks. Instrumenting a representative build produced:

```text
                   JS chunks   runtime IDs   chunkId comments
renderChunk              468           468                468
generateBundle           468           468                  0
writeBundle              468           468                  0
```

3. Observe `No chunks found` and no client symbol-set uploads.
4. Exact-pinning `@posthog/rollup-plugin` to 1.4.4 restores client processing. The regression starts in 1.4.9, which contains #4512.

Environment:

- Node 24.14.0
- Yarn 4.18.0
- Vite 8.0.16
- Rolldown 1.0.3
- `@posthog/rollup-plugin` 1.5.0

## Expected behavior

The plugin should identify and upload the final client chunks even when output minification removes comments added during `renderChunk`.

An integration test that performs a real minified Vite 8 build would catch this. The current tests invoke plugin hooks directly and do not include Vite's output-minification stage.

## Related sub-libraries

- [x] `@posthog/rollup-plugin`

## Workaround

Pin an exact pre-regression version. Version 1.4.4 is production-verified; 1.4.8 is the final release before #4512.

marandaneto:
## Problem

Vite 8 uses Oxc to minify client output after the Rollup `renderChunk` hook. Oxc removes the `//# chunkId=` comment added by the PostHog plugin but keeps the runtime chunk ID code. The plugin then skips every client chunk in `writeBundle`, so the build succeeds without uploading client source maps.

Closes #4596

## Changes

The plugin now remembers each chunk ID during `renderChunk` and restores a missing chunk ID comment in an early `generateBundle` hook after output minification. It uses Rollup's preliminary filename and the tracked ID to avoid adding comments to prebuilt or unrelated chunks.

This also adds focused unit tests and a functional test that runs a minified Vite 8 client build with Oxc and verifies that the client chunk is passed to `posthog-cli sourcemap upload`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [x] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi implemented the fix and ran the autoreview skill. The implementation restores the existing CLI-facing comment after minification instead of parsing arbitrary runtime code or uploading every JavaScript chunk. This keeps the existing prebuilt chunk exclusion and does not require a coordinated `posthog-cli` protocol change.

### DIFF
--- a/.changeset/fix-vite8-client-sourcemaps.md
+++ b/.changeset/fix-vite8-client-sourcemaps.md
@@ -1,0 +1,5 @@
+---
+'@posthog/rollup-plugin': patch
+---
+
+Restore chunk ID comments after Vite 8 output minification so client source maps are uploaded.
--- a/packages/rollup-plugin/package.json
+++ b/packages/rollup-plugin/package.json
@@ -17,9 +17,10 @@
     "scripts": {
         "build": "rslib build",
         "clean": "rimraf dist",
-        "lint": "eslint src",
-        "lint:fix": "eslint src --fix",
+        "lint": "eslint src 'test/*.mjs'",
+        "lint:fix": "eslint src 'test/*.mjs' --fix",
         "test:unit": "jest --passWithNoTests",
+        "test:functional": "node --test test/*.test.mjs",
         "dev": "rslib build --watch",
         "package": "pnpm pack --out $PACKAGE_DEST/%s.tgz"
     },
@@ -41,6 +42,7 @@
         "@rslib/core": "catalog:",
         "@types/jest": "catalog:",
         "jest": "catalog:",
-        "rollup": "~4.53.2"
+        "rollup": "~4.53.2",
+        "vite": "^8.0.16"
     }
 }
--- a/packages/rollup-plugin/src/index.spec.ts
+++ b/packages/rollup-plugin/src/index.spec.ts
@@ -34,6 +34,10 @@ type TestPlugin = {
         order: 'post'
         handler: (code: string, chunk: { fileName: string }) => RenderChunkResult
     }
+    generateBundle: {
+        order: 'pre'
+        handler: (options: OutputOptions, bundle: Record<string, unknown>) => void
+    }
     writeBundle: {
         sequential: true
         handler: (options: OutputOptions, bundle: Record<string, unknown>) => Promise<void>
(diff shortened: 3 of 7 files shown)


marandaneto:
caused by https://github.com/PostHog/posthog-js/pull/4512
worth testing @ablaszkiewicz

marandaneto:
Validated the reported reproduction against the exact Vite 8.0.16 version with Node 24, source maps enabled, and the default Oxc minifier. I used a local `posthog-cli` stub to capture whether the upload command was invoked.

- `origin/main` (`9b2a1b18d`): the runtime `_posthogChunkIds` code survived, the `//# chunkId=` comment was removed, the plugin logged `No chunks found`, and no client upload was invoked.
- This branch (`cf1f27570`): the runtime chunk ID and restored `//# chunkId=` comment were both present, and `posthog-cli sourcemap upload --stdin` was invoked with the emitted client chunk.

This confirms the same Vite 8 client build fails on `main` and uploads the client source map with this fix.

ablaszkiewicz (approved):
Thanks for this @marandaneto 🙏 . I ran my own tests against legacy/releaseless event modes and confirmed this PR fixed it

Issue #4574, opened 2026-08-19
`posthog-react-native/expo`: fix iOS dSYM build ordering and EAS versions


Joezo:
## Bug description

i think the `posthog-react-native/expo` plugin has two bugs around iOS symbol uploads

### the dSYM upload can run before the app dSYM is ready

i hit this during an EAS archive:

```text
ExampleApp.app.dSYM/Contents/Resources/DWARF/ExampleApp: The file was not recognized as a valid object file
```

the generated PostHog Xcode phase doesn't have an input dependency on the app's dSYM executable, so Xcode can run it before the dSYM has finished being generated

adding this as an input path fixed the build:

```text
$(DWARF_DSYM_FOLDER_PATH)/$(DWARF_DSYM_FILE_NAME)/Contents/Resources/DWARF/$(EXECUTABLE_NAME)
```

this is also the input path in PostHog's manual iOS setup docs, so i'm wondering if the Expo plugin has just missed adding it?

### the symbols can be uploaded against the wrong version

once the build worked, the symbols uploaded successfully, but they ended up under the wrong release in PostHog

the actual archive was version `2.10.0 (154)`, but the Hermes source map and native symbols were attached to:

```text
com.example.app@1.0+1
```

rather than:

```text
com.example.app@2.10.0+154
```

it looks like the upload scripts use `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`. With Expo prebuild these can still be the generated Xcode defaults, `1.0` and `1`, even though EAS remote versioning has written the real version and build number into the app's Info.plist

so the upload works, but the symbols are attached to a release that doesn't match the shipped app

## How to reproduce

1. set up an Expo app with `posthog-react-native` and enable iOS source map/dSYM uploads through the `posthog-react-native/expo` plugin
2. set `appVersionSource` to `remote` in `eas.json` and use `autoIncrement: true` for the iOS build profile
3. run an EAS iOS archive build with the PostHog upload credentials configured
4. the build can fail in the PostHog dSYM phase because the app dSYM isn't a valid object file yet
5. if it completes, check the release in PostHog. The symbols can be uploaded under the generated Xcode version, `1.0+1`, instead of the version in the archive

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-js-lite (web lite)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
[...]

PR #4602, opened 2026-08-22
fix(react-native): use EAS versions for iOS source maps

marandaneto:
## Problem

Expo and EAS can write the shipped iOS version and build number to the source Info.plist without updating `MARKETING_VERSION` or `CURRENT_PROJECT_VERSION`. PostHog then uploads Hermes source maps under a different release from the one reported by the native SDK.

This is the React Native part of #4574. Native dSYM readiness and native release version handling shipped in posthog-ios 3.69.10 through [PostHog/posthog-ios#776](https://github.com/PostHog/posthog-ios/pull/776). The React Native plugin now requires that version through [PostHog/posthog-js#4624](https://github.com/PostHog/posthog-js/pull/4624).

## Changes

- Read `CFBundleShortVersionString` and `CFBundleVersion` from the source `INFOPLIST_FILE` when uploading Hermes source maps.
- Resolve literal, custom, and compound Xcode build-setting references from the plist.
- Keep the existing Xcode version settings when plist values are missing, unresolved, or C-preprocessed.
- Pass posthog-cli arguments as Bash arrays so release values and file paths retain their argument boundaries.
- Require `@posthog/react-native-plugin` 2.4.2 or later when the optional native plugin is installed.
- Update the Expo 57 example to use the same minimum plugin version.
- Add regression tests for version resolution and fallback behavior.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi implemented and tested the change. Native dSYM handling lives in the shared posthog-ios upload script so native iOS and React Native builds use the same behavior. Shared Info.plist handling in posthog-cli is tracked in [PostHog/posthog#87775](https://github.com/PostHog/posthog/issues/87775).


packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -127,16 +127,51 @@ fi
 # mimics how the file is defined in node_modules/react-native/scripts/react-native-xcode.sh (PACKAGER_SOURCEMAP_FILE)
 SOURCEMAP_PACKAGER_FILE="$CONFIGURATION_BUILD_DIR/$SOURCEMAP_NAME"
 
-# Pass release info from Xcode build settings when available
+# Expo EAS remote versioning writes the release version to the source Info.plist
```
turnipdabeets: [suggestion] The runtime reads these two keys straight off the built bundle (`PostHogContext.swift` -> `Bundle.main.infoDictionary`), so the contract this needs to hold is "match what the SDK reports", not "match the build settings". Could the comment say that instead? It's the part that stays true, and it makes the custom-substitution case (a plist holding `$(APP_VERSION)`, which greptile flagged on #4618) a visible gap rather than something this and posthog-ios#776 get identically wrong.

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -127,16 +127,51 @@ fi
...
+resolve_posthog_ios_release_info
+
 CLI_RELEASE_ARGS=""
 if [ -n "${PRODUCT_BUNDLE_IDENTIFIER}" ]; then
   CLI_RELEASE_ARGS="$CLI_RELEASE_ARGS --release-name $PRODUCT_BUNDLE_IDENTIFIER"
 fi
-if [ -n "${MARKETING_VERSION}" ]; then
-  CLI_RELEASE_ARGS="$CLI_RELEASE_ARGS --release-version $MARKETING_VERSION"
+if [ -n "${POSTHOG_RELEASE_VERSION}" ]; then
+  CLI_RELEASE_ARGS="$CLI_RELEASE_ARGS --release-version $POSTHOG_RELEASE_VERSION"
```
turnipdabeets: [nit] `$POSTHOG_RELEASE_VERSION` gets word-split when `CLI_RELEASE_ARGS` is interpolated into the `/bin/sh -c` calls below — same exposure `MARKETING_VERSION` already had, but the value now comes from arbitrary plist content rather than a build setting. Version strings are constrained enough in practice that I wouldn't block on it, just flagging.

.changeset/fix-expo-ios-symbol-upload.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+Use EAS-managed iOS versions from the source Info.plist when uploading Hermes source maps.
```
turnipdabeets: [suggestion] Until posthog-ios#776 ships and people bump the pod, Hermes source maps will land on `2.10.0+154` while native dSYMs still land on `1.0+1` — better than today, but a confusing split if you're looking at one release. Worth naming the posthog-ios version floor here so it isn't a surprise?

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -127,16 +127,51 @@ fi
 # mimics how the file is defined in node_modules/react-native/scripts/react-native-xcode.sh (PACKAGER_SOURCEMAP_FILE)
 SOURCEMAP_PACKAGER_FILE="$CONFIGURATION_BUILD_DIR/$SOURCEMAP_NAME"
 
-# Pass release info from Xcode build settings when available
+# Expo EAS remote versioning writes the release version to the source Info.plist
+# without necessarily updating MARKETING_VERSION or CURRENT_PROJECT_VERSION.
+resolve_posthog_ios_release_info() {
```
turnipdabeets: [question] This resolver now exists here and in posthog-ios#776's `upload-symbols.sh`, with no shared source. Would `--info-plist` on posthog-cli be the real home for it? It already owns these semantics (`--build`'s help text says "CFBundleVersion on iOS"), and then both scripts just pass a path instead of each keeping their own copy. Not for this PR, but worth an issue if you agree.

packages/react-native/tooling/posthog-xcode.sh
turnipdabeets: [suggestion] The runtime reads these two keys straight off the built bundle (`PostHogContext.swift` -> `Bundle.main.infoDictionary`), so the contract this needs to hold is "match what the SDK reports", not "match the build settings". Could the comment say that instead? It's the part that stays true, and it makes the custom-substitution case (a plist holding `$(APP_VERSION)`, which greptile flagged on #4618) a visible gap rather than something this and posthog-ios#776 get identically wrong.

packages/react-native/tooling/posthog-xcode.sh
turnipdabeets: [nit] `$POSTHOG_RELEASE_VERSION` gets word-split when `CLI_RELEASE_ARGS` is interpolated into the `/bin/sh -c` calls below — same exposure `MARKETING_VERSION` already had, but the value now comes from arbitrary plist content rather than a build setting. Version strings are constrained enough in practice that I wouldn't block on it, just flagging.

.changeset/fix-expo-ios-symbol-upload.md
turnipdabeets: [suggestion] Native dSYMs get their version from `upload-symbols.sh` in the installed pod, so an app with `PostHog` pinned at 3.69.2 in its lockfile keeps uploading dSYMs under `1.0+1` while these source maps move to the real version — a confusing split to look at on one release. Could the changeset name the posthog-ios floor that carries the matching fix (PostHog/posthog-ios#776)?

packages/react-native/tooling/posthog-xcode.sh
turnipdabeets, line 132: [question] This resolver now exists here and in posthog-ios#776's `upload-symbols.sh`, with no shared source. Would `--info-plist` on posthog-cli be the real home for it? It already owns these semantics (`--build`'s help text says "CFBundleVersion on iOS"), and then both scripts just pass a path instead of each keeping their own copy. Not for this PR, but worth an issue if you agree.

turnipdabeets (approved):

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Fixed in 933e62eb0. The comment now states that source maps must match the values reported by the native runtime, and the resolver supports custom and compound Xcode build settings.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Agreed. I opened PostHog/posthog#87775 to track an `--info-plist` option in posthog-cli so both wrappers can eventually share these semantics.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Fixed in 933e62eb0. Release, upload, and path arguments now use Bash arrays so values retain their argument boundaries.

.changeset/fix-expo-ios-symbol-upload.md
marandaneto: Updated in 933e62eb0. The changeset now names posthog-ios 3.69.10 as the native dSYM attribution floor and references PostHog/posthog-ios#776.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Updated in 933e62eb0. The comment now describes matching the values reported by the native runtime, and the resolver handles literal, custom, and compound build-setting values from Info.plist.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Fixed in ca90402e0. Release and upload options are now Bash arrays, and the CLI is invoked directly with quoted array expansion so values and paths retain their argument boundaries.

.changeset/fix-expo-ios-symbol-upload.md
marandaneto: Updated in ca90402e0. The changeset now states that matching native dSYM attribution requires posthog-ios 3.69.10 or later and links PostHog/posthog-ios#776.

packages/react-native/tooling/posthog-xcode.sh
marandaneto: Agreed. Centralizing this contract in posthog-cli would remove the duplicate shell implementations. I opened PostHog/posthog#87775 to track an `--info-plist` option and the precedence rules for explicit release arguments.

PR #4605, opened 2026-08-22
feat(react-native): let surveys cap how far their text scales, per text role

safaiyeh:
React Native survey text currently scales without a ceiling, and host apps cannot apply a limit through the surrounding views. This adds app-level `maxFontSizeMultiplier` configuration, either for the whole survey or by text role:

```tsx
<PostHogSurveyProvider
  defaultSurveyAppearance={{
    maxFontSizeMultiplier: { question: 1.5, description: 1.8, ratingNumber: 1.2 },
  }}
>
  {children}
</PostHogSurveyProvider>
```

A number applies to all roles. Omitted roles retain the existing uncapped behavior. `0` explicitly means no maximum; use `1` or greater for a cap. Other values below `1` are unsupported by React Native and may be ignored or rejected, depending on platform and version. Fixed-size icon glyph fallbacks disable font scaling.

The package exports a React Native `SurveyAppearance` that extends the core contract, preserving every existing appearance key, along with `SurveyAppearanceTheme` and `SurveyTextRole`. The surveys team approved app-level configuration with per-role limits; no companion backend or editor PR is required.

Rebased onto main, adapted the regression tests to Vitest, and regenerated `posthog-react-native-references-latest.json`. All three PR commits are signed and Verified. Includes a minor changeset.

Validation on Node 24:

- React Native lint and formatting checks pass.
- React Native tests: **897 passed across 61 suites**, including 11 font-scaling regression tests.
- `pnpm generate-references` passes, including the React Native build and dependency builds.
- `pnpm check:public-api` passes.


marandaneto:
thanks @safaiyeh 
i'll defer to @PostHog/team-surveys since this is a SDK configuration only

The shared SurveyAppearance type and PostHog survey API do not include this field, so it cannot currently be configured from the PostHog survey editor/API

Moving to draft until we hear from @PostHog/team-surveys

marandaneto (commented):
Inline review findings from the validated review.

packages/react-native/src/surveys/PostHogSurveyProvider.tsx
```diff
@@ -73,8 +73,12 @@ export type PostHogSurveyProviderProps = {
 
   /**
    * The default appearance for surveys when not specified in PostHog.
+   *
+   * Accepts the React Native-only appearance fields as well (e.g.
+   * `maxFontSizeMultiplier`) — they are merged into the same theme object the
+   * survey components read, and PostHog never sends them down.
    */
-  defaultSurveyAppearance?: SurveyAppearance
+  defaultSurveyAppearance?: Partial<SurveyAppearanceTheme>
```
marandaneto: blocking: Preserve previously accepted appearance keys

Changing this prop from `SurveyAppearance` to `Partial<SurveyAppearanceTheme>` drops `widgetSelector`, `widgetType`, `widgetColor`, `widgetLabel`, and `shuffleQuestions` from accepted object-literal keys, so existing TypeScript callers using them fail excess-property checks. Please extend the existing public appearance contract with the React Native-only field rather than narrowing it.

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -121,12 +121,85 @@ export const defaultDescriptionOpacity = 0.8
...
+   * is what the OS asks for, but not always what a fixed-size survey card can
+   * hold - at the largest accessibility sizes an unbounded 18pt headline renders
+   * roughly one word per line.
+   *
+   * React Native only inherits this prop through nested `Text`, so a host app
+   * cannot apply it from the outside - it has to come in here.
+   *
+   * @default undefined (no ceiling)
+   */
+  maxFontSizeMultiplier?: number | Partial<Record<SurveyTextRole, number>>
```
marandaneto, line 109: blocking: Expose the new option through a public appearance type

The package root still exports only `SurveyAppearance` from `@posthog/core`, which lacks this field, while neither `SurveyAppearanceTheme` nor the referenced `SurveyTextRole` is publicly exported. Consumers therefore cannot type reusable appearance configuration containing `maxFontSizeMultiplier`, and API Extractor reports forgotten exports. Please export a public React Native appearance type and its referenced role type, and use it for public props.

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -121,12 +121,85 @@ export const defaultDescriptionOpacity = 0.8
...
 export type SurveyAppearanceTheme = Omit<
   Required<SurveyAppearance>,
   'widgetSelector' | 'widgetType' | 'widgetColor' | 'widgetLabel' | 'shuffleQuestions' | 'textColor' | 'inputTextColor'
 > & {
   textColor?: string
   inputTextColor?: string
+  /**
+   * Caps how far survey text may grow under the OS text-size setting, as a
+   * multiple of its base size - React Native's `maxFontSizeMultiplier`, applied
+   * to every `Text` and `TextInput` the survey renders.
```
marandaneto: blocking: Apply the survey-wide cap to fallback icon text

When `react-native-svg` or its native view managers are unavailable, `icons.tsx` renders glyph fallbacks with React Native `Text`, but those nodes never receive this cap. A numeric whole-survey limit therefore does not cover every `Text`, and the glyphs can scale and clip inside their fixed dimensions. Pass the relevant cap to those fallbacks or disable font scaling for icon-only glyphs.

packages/react-native/src/surveys/PostHogSurveyProvider.tsx
safaiyeh: Fixed in 61ed76d2: `defaultSurveyAppearance` now uses the public React Native `SurveyAppearance`, which extends the core contract and preserves all existing keys. New location: packages/react-native/src/surveys/PostHogSurveyProvider.tsx:86.

packages/react-native/src/surveys/surveys-utils.ts
safaiyeh: Fixed in 61ed76d2: the package root now exports the extended `SurveyAppearance`, `SurveyAppearanceTheme`, and `SurveyTextRole`, and public provider props use the extended type. New location: packages/react-native/src/surveys/index.ts:6.

packages/react-native/src/surveys/surveys-utils.ts
safaiyeh: Fixed in 61ed76d2: the shared fallback-icon renderer now disables font scaling, keeping every glyph inside its fixed dimensions. New location: packages/react-native/src/surveys/icons.tsx:41.

safaiyeh:
Thanks for the review @marandaneto I've addressed ur feed back and rebased to main

marandaneto:
> Thanks for the review @marandaneto I've addressed ur feed back and rebased to main

thanks, i pinged @PostHog/team-surveys again

lucasheriques (approved):
The SDK API looks good to me, including the justification for per-role limits. Before merging, please add a companion PR in the [posthog main repo](https://github.com/PostHog/posthog/) so these settings can also be saved through the survey API and configured in the survey editor, with matching defaults and value semantics. 

Since otherwise we're shipping this changes for nothing as users won't have a way to edit them and they will always be undefined

turnipdabeets:
@safaiyeh We talked it over with the surveys team, and we're happy to ship this as app-level config, per-role limits included. No companion posthog PR needed. 

Three small things before we can merge:

1. Rebase onto main again. 
2. Regenerate the API references. Run pnpm generate-references and commit the updated posthog-react-native-references-latest.json, otherwise the public-API check in CI fails.
3. Sign your commits so they show as Verified. If you set up commit signing and turn on git config commit.gpgsign true before rebasing, the rebase will sign them all.

marandaneto:
moved to draft until [this](https://github.com/PostHog/posthog-js/pull/4605#issuecomment-5636878130) is resolved

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -49,13 +49,99 @@ export const defaultRatingLabelOpacity = 0.7
...
+   * inside a rating button, and a host app that already caps its own text by
+   * role will want to match those ceilings here.
+   *
+   * ```ts
+   * maxFontSizeMultiplier: 1.6
+   * // or
+   * maxFontSizeMultiplier: { question: 1.5, description: 1.8, ratingNumber: 1.2 }
+   * ```
+   *
+   * Roles left out of the object are uncapped, as they are today.
```
turnipdabeets, line 97: [suggestion] React Native ignores anything below 1 on both iOS and Android, so `0.8` gives no cap rather than smaller text. Could we say so here?
```suggestion
   * Roles left out of the object are uncapped, as they are today. Values below 1
   * (including `0`) are ignored by React Native, so they leave text uncapped too.
```

safaiyeh: Clarified in 947bafee: `0` means no maximum, and caps use `1` or greater. Other values below `1` do not shrink text and are unsupported. I used “ignored or rejected” because the React Native 0.69 Android implementation in the package’s development dependencies throws for nonzero values below `1`. The generated API reference includes this wording too.

safaiyeh:
Addressed the remaining requests in 947bafee:

- Rebased onto main (`f998ee9c`) and preserved the updated survey shuffling and placeholder behavior.
- Regenerated and committed `posthog-react-native-references-latest.json`; `pnpm check:public-api` passes.
- All three rebased commits are signed by GitHub and show Verified.
- Clarified the supported font-cap values and migrated the regression tests to main’s Vitest setup.

Validation on Node 24: React Native lint and formatting pass; all **897 tests in 61 suites** pass; `pnpm generate-references` succeeds, including the package build. The PR description now reflects the final public types and checks.

PR #4612, opened 2026-08-23
fix(rrweb): harden recording correctness

marandaneto:
## Problem

Several narrow rrweb recording paths could serialize incorrect state or change page behavior. These included failed CSSOM operations, shadow DOM serialization, cross-origin iframe ID sentinels, blocked canvas setters, deep DOM traversal, and canvas hydration in RRDOM.

## Changes

- Emit stylesheet replacement and declaration changes only after the native operation succeeds.
- Queue subsequent CSSOM rule and declaration events behind asynchronous replacements so invocation order is preserved without recording rejected replacements.
- Bound stalled replacement queues, release later mutations after five seconds, and capture the final stylesheet if a replacement settles late.
- Preserve `CSSStyleSheet.replace()` promise identity, rejection behavior, and patched non-promise return values for application callers.
- Serialize shadow roots with their owner document without observing shadow content beneath blocked hosts.
- Preserve cross-origin iframe sentinel and zero IDs.
- Avoid recording blocked or unsupported canvas setters.
- Replace recursive traversal where deeply nested DOM could overflow the stack.
- Hydrate RRDOM canvas content during diffing.
- Add focused recorder and RRDOM regression coverage.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi coding agent implemented and validated these fixes from a Slopo duplication report in a local session with no public transcript. The scope was kept to recorder correctness seams. Replay architecture, lazy loading, session rotation, and network wrappers were intentionally left unchanged.


packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -752,21 +756,31 @@ function initStyleSheetObserver(
...
-              id,
-              styleId,
-              replace: text,
-            });
-          }
-          return target.apply(thisArg, argumentsList);
+          const result = target.apply(thisArg, argumentsList);
+
+          return result.then((stylesheet) => {
+            try {
```
TueHaulund, line 917: Emitting inside the promise flips the event order. Anything the page does right after `replace()` gets recorded first, so `sheet.replace(css); sheet.insertRule(x)` replays as insertRule then replace, and the replace wipes the rule. The promise waits on `@import` fetching, so this isn't just a microtask gap.

`onCssomSheetMutation` gets deferred too, which brings back the race the other call sites in this PR were moved to fix.

You can check up front whether this will reject: `replace()` only fails on a non-constructed sheet or one already replacing. So `if (thisArg.ownerNode) return result;` at the top, and leave the emit where it was. A test with `replace()` followed by `insertRule` would catch this, nothing covers it now.

TueHaulund, line 917: This assumes the native return value, but `replace` is only feature-detected at L748. A polyfill or a page that patched `replace` can return something without `.then`, and then this throws a `TypeError` out of the customer's own call. `if (!result || typeof result.then !== 'function') return result;` would cover it.

Returning `result.then(...)` also hands back a different promise than the native one. Attaching the side effect and still returning `result` avoids that.

ioannisj (commented):
Looks good, will wait for @TueHaulund to approve this after those threads are resolved

packages/rrweb/rrweb/src/record/observer.ts
marandaneto: Good catch — fixed in 5ae71b22c. Replacement events now hold their invocation-order slot until the original promise settles, and subsequent CSSOM rule events queue behind them. Fulfilled replacements flush before later rules; rejected replacements are discarded. Added real-browser coverage for replace() followed immediately by insertRule(), consecutive replacements, and rejection.

marandaneto: Fixed in 5ae71b22c. The wrapper now guards non-thenable return values and always returns the exact original value/promise. Settlement handlers are attached only for bookkeeping, with both fulfillment and rejection handled; coverage includes promise identity, non-promise patched returns, and a proxied implementation that rejects.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -654,6 +654,63 @@ function initStyleSheetObserver(
...
+    queue.push(mutation);
+    let completed = false;
+    return (success: boolean) => {
+      if (completed) return;
+      completed = true;
+      mutation.ready = true;
+      if (!success) mutation.emit = undefined;
+      flushStyleSheetMutations(sheet);
+    };
+  };
```
TueHaulund: This is the one I would want changed before merge: a `replace()` whose promise never settles permanently silences CSSOM recording for that sheet.

Once `queuePendingStyleSheetMutation` has created the queue, `queueStyleSheetMutation` stops emitting directly (L685-689) and appends instead, and `flushStyleSheetMutations` only drains `while (queue[0]?.ready)` (L669). If the head's `completeMutation` is never called there is no timeout, no cap, and no abandonment path, so every later `insertRule` / `deleteRule` / `replaceSync` on that sheet is dropped for the rest of the session and the array grows for as long as the page keeps mutating the sheet.

Not hypothetical: the guard at L822-826 accepts any thenable, and this PR's own tests monkeypatch `CSSStyleSheet.prototype.replace` to return arbitrary promises. A host-page polyfill or instrumentation that returns a promise resolving late (or never) hits it. The failure is silent event loss, which is the worst shape for a recorder bug.

Bounding the wait and dropping the barrier when it expires would be enough:

```suggestion
  const queuePendingStyleSheetMutation = (
    sheet: CSSStyleSheet,
    emit: () => void,
  ) => {
    let queue = queuedStyleSheetMutations.get(sheet);
    if (!queue) {
      queue = [];
      queuedStyleSheetMutations.set(sheet, queue);
    }
    const mutation: QueuedStyleSheetMutation = { emit, ready: false };
    queue.push(mutation);
    let completed = false;
    const complete = (success: boolean) => {
      if (completed) return;
      completed = true;
      mutation.ready = true;
      if (!success) mutation.emit = undefined;
      flushStyleSheetMutations(sheet);
    };
    // A host-patched replace() may return a thenable that never settles; don't
    // let that barrier drop every later mutation on this sheet.
    setTimeout(() => complete(false), 5000);
    return complete;
  };
```

Worth a test too: a `replace()` left pending, asserting a later `insertRule` still gets recorded.

Related, same design: teardown restores the prototypes but abandons any pending queue, so mutations sitting behind an unsettled `replace` at stop time are lost. Probably acceptable, but it is the same root cause.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -752,21 +817,48 @@ function initStyleSheetObserver(
...
-            mirror,
-            stylesheetManager.styleMirror,
-          );
+          if (
+            thisArg.ownerNode ||
+            !result ||
+            typeof result.then !== 'function'
+          ) {
+            return result;
+          }
```
TueHaulund: These two property reads sit just outside the `try` at L830, whose catch comment says "Recorder errors must not affect the native promise result". An exotic `thisArg` with a throwing `ownerNode` or `then` getter turns the page's `sheet.replace()` into a synchronous throw where it previously returned a promise.

Worth noting `callbackWrapper` gives no cover here: in `error-handler.ts` it returns `cb` completely unwrapped when no `errorHandler` is registered, which is the default, and rethrows otherwise unless the handler returns `true`. So the `try` below is the only protection, and these reads are outside it.

Moving both inside the `try` is the whole fix.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -654,6 +654,63 @@ function initStyleSheetObserver(
...
+  };
+  const queueStyleSheetMutation = (
+    sheet: CSSStyleSheet,
+    emit: () => void,
+  ) => {
+    const queue = queuedStyleSheetMutations.get(sheet);
+    if (!queue) {
+      emit();
+      return;
+    }
```
TueHaulund: Asymmetry worth closing: `flushStyleSheetMutations` wraps `mutation?.emit?.()` in `try/catch` (L671-675) with the comment "Recorder errors must not affect native CSSOM operations", but this path, which is the common one, calls `emit()` bare.

It matters more now than it used to. Because `target.apply` runs first, the native operation has already succeeded by the time `emit()` throws, so the page sees a spurious throw from an operation that actually completed. Same `try/catch` as the flush path fixes it.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -752,21 +817,48 @@ function initStyleSheetObserver(
...
+                }),
+            );
+            try {
+              void result.then(
+                () => completeMutation(true),
+                () => completeMutation(false),
+              );
+            } catch {
+              completeMutation(false);
+            }
```
TueHaulund: `result` is the same promise object handed back to the caller at L861, which your own "preserves the promise returned by a patched stylesheet replacement" test asserts. Attaching a rejection handler to it marks it handled, so a failing `sheet.replace()` that the page never awaits stops surfacing as an `unhandledrejection`.

That is the recorder quietly changing the host page's error reporting, which is the kind of thing this file otherwise works hard to avoid. Either re-raise asynchronously from the rejection handler, or leave it and say so in a comment so it reads as a decision.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -1047,9 +1155,11 @@ function initStyleDeclarationObserver(
       ) => {
         const [property, value, priority] = argumentsList;
 
+        const result = target.apply(thisArg, argumentsList);
```
TueHaulund: `setProperty` / `removeProperty` emit `styleDeclarationCb` directly and never touch the new queue, so the ordering guarantee only half exists.

While a `replace()` is pending on a sheet, a declaration change on that same sheet is emitted ahead of the queued `replace`. Replay then applies the declaration and the `replace` wipes it, which inverts what happened live. This code already resolves `thisArg.parentRule?.parentStyleSheet` two lines down, so it has the key it would need to go through `queueStyleSheetMutation`, falling back to direct emit when there is no sheet.

Fine as a follow-up if you would rather keep this PR to the `replace` path, but then it is worth a comment saying the two are deliberately not ordered against each other.

packages/rrweb/rrweb/src/record/observers/canvas/webgl.ts
```diff
@@ -72,14 +72,17 @@ function patchGLPrototype(
...
-          // TODO: this could potentially also be an OffscreenCanvas as well as HTMLCanvasElement
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          cb(this.canvas as HTMLCanvasElement, {
-            type,
-            property: prop,
-            args: [v],
-            setter: true,
-          });
+          if (
+            'tagName' in this.canvas &&
```
TueHaulund: `in` throws a `TypeError` on `null` / `undefined`, and this runs inside a setter on the host page's hot path. Cheap guard:

```suggestion
            !!this.canvas &&
            'tagName' in this.canvas &&
```

Separately, this check also stops recording setter mutations for `OffscreenCanvas` contexts, which the previous code cast to `HTMLCanvasElement` and let through. It is consistent with the patched-function path at L54, so I think it is intended, but it is a real behavior change and the changeset doesn't mention it.

While you're here: this `catch` branch is the path taken for every accessor property on the context prototypes (`fillStyle`, `strokeStyle`, `globalAlpha`, `lineWidth`, `font`, and so on), since reading those off the prototype throws `Illegal invocation`. Those setters fire per draw call, and `isBlocked(..., checkAncestors = true)` does a `closest()` walk every time. The masking fix is worth the cost, but a per-canvas `WeakMap` cache invalidated per emit batch would get the walk off the hot path. Same applies to `2d.ts:74`.

packages/rrweb/rrweb/test/record.test.ts
```diff
@@ -610,18 +610,225 @@ describe('record', function (this: ISuite) {
...
+
+  it('does not emit when a proxied stylesheet replacement rejects', async () => {
+    const result = await ctx.page.evaluate(async () => {
+      let patchedResult: Promise<CSSStyleSheet> | undefined;
+      const rejectReplacement = () => {
+        patchedResult = Promise.reject(
+          new DOMException('Replacement failed', 'NotAllowedError'),
+        );
+        return patchedResult;
+      };
```
TueHaulund: This and the assignment further down both replace `CSSStyleSheet.prototype.replace` without restoring it. If the suite ever shares a page context these leak into later tests as an invisible ordering dependency, which is a miserable thing to debug later. Capture the original and restore in a `finally`, or patch inside a fresh page.

Also `await ctx.page.waitForTimeout(20)` at L660: the surrounding tests use `waitForRAF(ctx.page)`, and a fixed 20ms is a flake vector on a loaded CI runner.

packages/rrweb/rrweb/test/record/canvas-setter-blocking.test.ts
```diff
@@ -0,0 +1,120 @@
+/**
+ * @vitest-environment jsdom
+ */
+import { describe, expect, it, vi } from 'vitest'
```
TueHaulund: This file and `cross-origin-iframe-mirror.test.ts` use 4-space indentation and no semicolons, while everything else under `packages/rrweb/rrweb/` is prettier-formatted at 2 spaces with semicolons. Probably a format check failure in CI rather than anything you need to think about.

TueHaulund (commented):
.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -643,9 +644,121 @@ function getIdAndStyleId(
...
+  ) => {
+    if (!active) return () => undefined;
+    let queue = queuedMutations.get(sheet);
+    if (!queue) {
+      queue = [];
+      queuedMutations.set(sheet, queue);
+    }
+    const mutation: QueuedStyleSheetMutation = { emit, ready: false };
+    queue.push(mutation);
+    let expired = false;
```
TueHaulund, line 712: A `replace()` promise that never settles permanently silences CSSOM recording for that sheet. Once the queue exists, `queueStyleSheetMutation` stops emitting directly (L685-689) and `flushStyleSheetMutations` only drains `while (queue[0]?.ready)`. No timeout, no cap, so every later `insertRule` / `deleteRule` / `replaceSync` on that sheet is dropped for the rest of the session.

Not hypothetical: the guard at L822-826 accepts any thenable, and the new tests monkeypatch `replace` to return arbitrary promises. A host-page polyfill doing the same hits it.

A `setTimeout(() => complete(false), 5000)` alongside the returned callback would bound it.

packages/rrweb/rrweb/src/record/observer.ts
```diff
@@ -643,9 +644,121 @@ function getIdAndStyleId(
...
+    }
+    if (queue.length === 0) queuedMutations.delete(sheet);
+  };
+  const queueMutation = (
+    sheet: CSSStyleSheet | null | undefined,
+    emit: () => void,
+  ) => {
+    if (!sheet) {
+      safeEmit(emit);
+      return;
```
TueHaulund, line 689: `flushStyleSheetMutations` wraps `emit()` in try/catch (L671-675), but this path, which is the common one, calls it bare.

Worth closing because `callbackWrapper` gives no cover here: in `error-handler.ts` it returns `cb` completely unwrapped when no `errorHandler` is registered, which is the default. So a throwing `styleSheetRuleCb` propagates into the page's `insertRule` call, and since `target.apply` now runs first, the page sees a throw from an operation that already succeeded.

packages/rrweb/rrweb/src/record/observers/canvas/webgl.ts
```diff
@@ -72,14 +73,18 @@ function patchGLPrototype(
...
         set(v) {
-          // TODO: this could potentially also be an OffscreenCanvas as well as HTMLCanvasElement
-          // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-          cb(this.canvas as HTMLCanvasElement, {
-            type,
-            property: prop,
-            args: [v],
-            setter: true,
-          });
+          if (
```
TueHaulund, line 76: This also stops recording setter mutations for `OffscreenCanvas` contexts, which the previous code cast to `HTMLCanvasElement` and let through. Consistent with the patched-function path at L54 so I assume it's intended, but it's a real behavior change and the changeset doesn't mention it.

Minor: `in` throws a `TypeError` on a nullish `this.canvas`, from inside a setter on the host page's hot path. `!!this.canvas &&` in front is cheap insurance.

TueHaulund (commented):
`hasShadowRoot(currentN)` is a great catch, and replay compatibility holds: the new `index` field is additive, old recordings with `index: undefined` are unchanged. Notes are all in the new stylesheet queue.

packages/rrweb/rrweb/src/record/mutation.ts
```diff
@@ -400,9 +400,12 @@ export default class MutationBuffer {
               currentN as HTMLLinkElement,
             );
           }
-          if (hasShadowRoot(n)) {
+          if (hasShadowRoot(currentN)) {
```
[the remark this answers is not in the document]
marandaneto: Good catch — fixed in f71d0fc90. Shadow-root observers are no longer registered for blocked hosts or hosts beneath blocked ancestors. Added a real-browser regression test that appends a private marker after serialization and verifies it never enters recorded events.

packages/rrweb/rrweb/src/record/observers/canvas/webgl.ts
marandaneto: Fixed in ea18aaffa. Both WebGL paths now guard nullish canvases, the changeset calls out unsupported canvases, and coverage includes a detached/null canvas alongside blocked and offscreen canvases.

packages/rrweb/rrweb/src/record/observer.ts
marandaneto: Fixed in ea18aaffa. Pending barriers expire after five seconds so later CSSOM mutations keep recording, timers and queues are cleared on teardown, and a replacement that eventually succeeds emits a fresh serialization of the final stylesheet. Added deterministic coverage for never-settling and late-settling replacements.

packages/rrweb/rrweb/src/record/observer.ts
marandaneto: Fixed in ea18aaffa. Immediate and queued emissions now share safeEmit(), preventing recorder callback failures from escaping after a successful native CSSOM operation. Added real-browser coverage with a throwing emit callback.

TueHaulund (approved):

PR #4614, opened 2026-08-23
fix(node): send gzip bytes instead of a Blob request body

iamomiid:
## Problem

#4423 moved Node compression to `node:zlib`, which removed the `Blob.slice().arrayBuffer()` reads in `validateNativeGzip`. The compressed payload is still wrapped in a `Blob` though, and that Blob is what gets handed to `fetch` as the request body — so one Blob is still read per request.

On Node 24.16 and later every Blob read leaks a native `BlobReader` that stays registered on the Node `Environment` for the life of the process (nodejs/node#63574). That lines up with the numbers in #4423: RSS improved from 145 MB to 111 MB but kept climbing instead of flattening.

Measuring retained `heapUsed` after a forced GC, on published `posthog-node` 5.50.0. Module-level client, `flushAt: 1`, `flushInterval: 0`, `capture()` + `await flush()` in a loop against a local sink:

| Node | Config | heap per capture |
| --- | --- | --- |
| 24.17.0 | default | **2.29 KB, linear across 8,000 captures** |
| 24.17.0 | `disableCompression: true` | ~0, flat |
| 20.11.1 | default | ~0, flat |

A heap snapshot after 4,200 captures has 4,201 live `BlobReader`, 4,203 `Blob` and 4,204 `ReadableStream` instances one per capture. The strong retainer path ends in Node itself, not in SDK code:

```
(GC root) -> Node / Environment -> Node / PrincipalRealm
          -> Node / BaseObjectList -> native: Node / Blob::Reader -> BlobReader
```

With this change the same run is flat at ~10.2 MB across 8,000 captures, and the snapshot has zero live `BlobReader`s. 

## Changes

`fetch` takes a `Uint8Array` body directly, so the `Blob` wrapper isn't needed.

- `gzip.node.ts` returns the `Uint8Array` instead of wrapping it in a `Blob`.
- Widen the body type through the places it flows: `compressPayload` in core and in the Node entrypoint, `PostHogFetchOptions.body`, and the Capture V1 `compress` hook and local `body`.
- `fetchWithRetry` computes `reqByteLength` from either a `Blob` or a string; it now handles `Uint8Array` too, in both the `try` and the `catch`.
- The Capture V1 wiring tests asserted `toBeInstanceOf(Blob)` on the request body, so those assertions and the test's `readCompressedBody` helper now expect bytes.
- Added a case to `gzip.node.spec.ts` covering the success path — it asserts the result is a `Uint8Array`, not a `Blob`, and gunzips back to the input. The existing spec mocked `node:zlib` to always fail, so the mock is now toggleable to keep both cases in one file.

The edge entrypoint is untouched and still uses `CompressionStream`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

This started from a memory leak on my own Next.js site, which turned out to be a different SDK entirely. Chasing it down meant building a harness that forces GC over the V8 inspector before sampling `heapUsed`, so the numbers are retention rather than GC lag, that's why they disagree with the RSS/`heapTotal` figures in #4423, which plateau visually but keep climbing underneath.

Claude Code did the bisecting, heap-snapshot analysis and the patch under my direction. The `Uint8Array<ArrayBuffer>` detail and the `fetchWithRetry` byte-length branch only surfaced from type-checking the real monorepo, not from reading the source.

Verified locally: `pnpm build`, `pnpm test:unit` (882 in core, 921 in node), `pnpm test:functional`, `test:edge-compat`, and lint. `posthog-js` and `posthog-react-native` build too, since they also consume `PostHogFetchOptions`. The leak measurement used the freshly built fork `dist` overlaid on a 5.50.0 install, and a sink that gunzips and parses the body accepted 10/10 requests with `Content-Encoding: gzip` intact, so the wire format is unchanged.

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -299,12 +299,14 @@ export enum PostHogPersistedProperty {
   DeviceId = 'device_id', // only used by posthog-react-native
 }
 
+export type PostHogFetchBodyBytes = Uint8Array & { buffer: ArrayBuffer }
+
 export type PostHogFetchOptions = {
   method: 'GET' | 'POST' | 'PUT' | 'PATCH'
   mode?: 'no-cors'
   credentials?: 'omit'
   headers: { [key: string]: string }
-  body?: string | Blob | Uint8Array<ArrayBuffer>
+  body?: string | Blob | PostHogFetchBodyBytes
   signal?: AbortSignal
 }
 
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -304,7 +304,7 @@ export type PostHogFetchOptions = {
   mode?: 'no-cors'
   credentials?: 'omit'
   headers: { [key: string]: string }
-  body?: string | Blob
+  body?: string | Blob | Uint8Array<ArrayBuffer>
   signal?: AbortSignal
 }
 
--- a/.changeset/node-gzip-uint8array-body.md
+++ b/.changeset/node-gzip-uint8array-body.md
@@ -1,0 +1,6 @@
+---
+'@posthog/core': patch
+'posthog-node': patch
+---
+
+The Node SDK now sends the raw gzip bytes as the request body instead of wrapping them in a `Blob`. On Node 24.16 and later, reading a `Blob` request body leaks a native `BlobReader` that is never released, so a service calling `capture()` and `flush()` once per request grew by roughly 2.3 KB of heap per event and never gave it back. This completes the work in #4423: switching to `node:zlib` removed the compression-time Blob reads, but the body itself was still a Blob and still got read once per request. Compression behaviour, headers and the wire format are unchanged, and the edge build keeps using `CompressionStream`.
--- a/packages/core/src/posthog-core-stateless.ts
+++ b/packages/core/src/posthog-core-stateless.ts
@@ -1447,7 +1447,7 @@ export abstract class PostHogCoreStateless {
    * Compresses an outgoing payload. Runtime-specific clients can override this
    * to avoid using the Web Streams compression implementation.
    */
-  protected compressPayload(payload: string): Promise<Blob | null> {
+  protected compressPayload(payload: string): Promise<Blob | Uint8Array<ArrayBuffer> | null> {
     return gzipCompress(payload, this.isDebug)
   }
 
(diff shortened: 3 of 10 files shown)


packages/core/src/types.ts
```diff
@@ -304,7 +304,7 @@ export type PostHogFetchOptions = {
   mode?: 'no-cors'
   credentials?: 'omit'
   headers: { [key: string]: string }
-  body?: string | Blob
+  body?: string | Blob | Uint8Array<ArrayBuffer>
```
marandaneto: blocking: Avoid requiring generic typed arrays in public declarations — `Uint8Array<ArrayBuffer>` requires newer generic typed-array declarations; TypeScript 4.7.4, which is still used by `examples/example-node` and represented in the repository's minimum-version matrix, reports `TS2315: Type 'Uint8Array' is not generic`. Because this type and the changed `PostHog` method signatures are emitted in public declarations, Node consumers compiling without `skipLibCheck` will fail immediately after upgrading.

marandaneto (commented):

marandaneto:
thanks @iamomiid left a comment

packages/core/src/types.ts
marandaneto: suggestion: Use the backward-compatible, non-generic `Uint8Array` type in all seven changed signatures (for example, `body?: string | Blob | Uint8Array`) while keeping `return new Uint8Array(compressed)` unchanged. This still sends raw gzip bytes and avoids the `BlobReader` leak without raising the consumer TypeScript minimum; ideally, add a `posthog-node` consumer check using TypeScript 4.7 without `skipLibCheck` as regression coverage.

iamomiid: @marandaneto I forgot about older versions of TS. Also, plain Uint8Array alone doesn't hold. Recent `lib.dom` types BodyInit as accepting only ArrayBuffer-backed views, so `PostHogFetchOptions` stops being assignable to `RequestInit` and the build breaks where the SDK passes its options to the global fetch.

Rather than casting, I had to create a new type `PostHogFetchBodyBytes`.

Parses on TS 4.7, still satisfies `BodyInit`, and `new Uint8Array(compressed)` assigns to it with no cast and no copy. 

On the regression check: a `posthog-node` consumer on TS 4.7 without `skipLibCheck` already fails on published 5.50.0 - `TS1383` in core's `export type *` and missing `express` types. I get those two against this branch and nothing else, no `TS2315`. So a no-`skipLibCheck` job would go red on `main` today. Happy to add one once those are fixed, or a cheaper check that fails if generic typed arrays appear in published `.d.ts`.

It's your call.

marandaneto: nice, thanks

marandaneto (approved):

PR #4615, opened 2026-08-23
Fix toolbar loading in no-external builds

### LINKED ISSUE #3950: posthog-js: toolbar does not load with `module.full.no-external` variant
## Bug description

When using the `posthog-js/dist/module.full.no-external` import for posthog-js, the toolbar cannot load as its code is not bundled in the JS library.

## How to reproduce

1. Setup Posthog client using `posthog-js/dist/module.full.no-external`
2. Go to Posthog and launch the toolbar -> does not load
3. Copy the launching code from posthog in the console -> toolbar does not load

## Related sub-libraries

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin

vaibhavmashal:
Fixes #3950

### DIFF
--- a/packages/browser/src/entrypoints/module.full.no-external.es.ts
+++ b/packages/browser/src/entrypoints/module.full.no-external.es.ts
@@ -1,5 +1,6 @@
 import './all-external-dependencies'
 import './lazy-recorder'
+import './external-scripts-loader'
 import posthog from './module.no-external.es'
 export * from './module.no-external.es'
 export default posthog
--- a/.changeset/fix-toolbar-no-external.md
+++ b/.changeset/fix-toolbar-no-external.md
@@ -1,0 +1,5 @@
+---
+"posthog-js": patch
+---
+
+Fix toolbar loading in no-external builds
--- a/packages/browser/src/entrypoints/array.full.no-external.ts
+++ b/packages/browser/src/entrypoints/array.full.no-external.ts
@@ -1,4 +1,5 @@
 // Same as loader-globals.ts except includes all additional extension loaders
 import './all-external-dependencies'
 import './lazy-recorder'
+import './external-scripts-loader'
 import './array.no-external'


packages/browser/src/entrypoints/module.full.no-external.es.ts
```diff
@@ -1,5 +1,6 @@
 import './all-external-dependencies'
 import './lazy-recorder'
+import './external-scripts-loader'
```
marandaneto, line 3: blocking: Preserve the no-external contract — importing the general script loader enables runtime script injection in an entrypoint that explicitly promises no loaders, and it makes the existing `module.test.ts` no-loader assertion fail after building. This regresses CSP/offline/browser-extension consumers; toolbar support needs an explicit opt-in entrypoint or another contract-level design. The same issue applies to `array.full.no-external.ts`.

marandaneto (commented):

marandaneto:
@dustinbyrne something to think off due to extensions
thanks @vaibhavmashal left a few comments

dustinbyrne:
hey @vaibhavmashal, thanks for the pull request!

> @dustinbyrne something to think off due to extensions

`Toolbar` is already a tree-shakable extension, but the issue is that the toolbar assets aren't bundled into any entrypoint that we ship here. the `Toolbar` extension _requires_ `external-scripts-loader` for this reason. however, the `*.no-external` bundle variants explicitly do not allow loading external scripts, so the toolbar is incompatible.

given that, we have two options:
1. allow the toolbar to be the one exception to `no-external` and load the external script. some users might consider this to be unexpected. we'd also need to guarantee that this doesn't open the door for loading other external scripts given current behavior.
2. include the toolbar assets into the `no-external` bundle. this stays true to the `no-external` entrypoint, but bloats the bundle size with a (mostly) non-user facing asset.

each option has its upsides and downsides, so it's not clear to me that either is the better option. @vaibhavmashal is this something that you're affected by?

vaibhavmashal:
Since both options have their trade-offs (Option 1 breaks the strict no-external CSP contract, and Option 2 bloats the bundle size), and @marandaneto rightly pointed out that importing the general script loader breaks the module test assertions for no-loader builds, I'll close this PR for now to avoid forcing a controversial architectural change. Happy to reopen once the maintainer team aligns on a contract-level design for toolbar support in no-external builds!

Issue #710, opened 2023-06-22
Uncaught RangeError: invalid field value


dynalz:
I'm having an issue on this page in particular (live):
https://lp.ricardomatos.me/e2_imo4all_o2f1-masterclass_1

For some reason I cannot capture anything, as it just throws an error
Error to reproduce in console:
`posthog.capture('input-capture', { $set: {foo: "bar"} })`

Traceback as follows:
```
uuidv7.ts:59 Uncaught RangeError: invalid field value
    at Function.value (uuidv7.ts:59:19)
    at e.value (uuidv7.ts:181:21)
    at e.value (uuidv7.ts:137:28)
    at F (uuidv7.ts:244:92)
    at se (uuidv7.ts:241:37)
    at e.value (posthog-core.ts:851:19)
    at e.value (posthog-core.ts:520:18)
    at posthog-core.ts:2107:18
    at L (utils.ts:129:26)
    at posthog-core.ts:2106:5
```

I managed to debug and trace it back to here:
![image](https://github.com/PostHog/posthog-js/assets/12519753/db3d08f6-006d-4d3c-a753-012cbaf42bb7)

neilkakkar:
(cc: @pauldambra )

pauldambra:
Copying random code off the internet finally back-fired 🤣 Thanks for the heads-up @neilkakkar and thanks for the report @dynalz . This is a super clear bug report thanks for taking the time 🙌

pauldambra:
If I visit your page with adblockers on then I can run PostHog without seeing the error. 

In your screenshot, the value of the first argument should be a number but appears to be a Date object. I _think_ one of the other trackers on the page is over-riding `Date.now()` which is very cheeky of them 🤣 Or maybe Chrome is being super helpful 

<img width="404" alt="Screenshot 2023-06-23 at 12 51 48" src="https://github.com/PostHog/posthog-js/assets/984817/e5b8262a-34c4-4cf7-bcea-f3d6ae539ed4">

My chrome (at least) provides a number as expected.

So I guess there are two things here

1) Is there a safer way for us to get the current time
2) Should we really throw in this method? If we send an event to PostHog without a UUID we generate one on the backend. So we lose re-triablity but you don't lose your data 🤔

pauldambra:
My whole network at home blocks ad-trackers so this was hard to investigate 🤣 

I hooked my phone up for remote debugging since that device has a VPN. 

[...]

PR #4616, opened 2026-08-24
fix(core): clamp uuidv7 RNG output to a valid uint32 so a nonconformant Math.random cannot crash event capture

shahidrogers:
## Problem

The vendored UUIDv7 generator in `@posthog/core` builds its random fields from a `Math.random()`-based `nextUint32()` (Web Crypto is deliberately disabled for React Native compatibility). If the environment's `Math.random()` ever returns a nonconformant value — **≥ 1.0 or NaN** — `nextUint32()` produces a value outside `[0, 2^32)`, and `UUID.fromFieldsV7` throws `RangeError: invalid field value`.

This is not hypothetical. On React Native Android, Hermes implements `Math.random` with C++ `std::uniform_real_distribution` ([facebook/hermes#1169](https://github.com/facebook/hermes/issues/1169)), which is [documented](https://en.cppreference.com/w/cpp/numeric/random/uniform_real_distribution) to occasionally return its upper bound (1.0) in most standard-library implementations. In our production fintech app (posthog-react-native 4.61.4 / @posthog/core 1.46.7) affected devices hit the throw inside `PromiseQueue.add → uuidv7()` during the SDK's **internal event-queue flush** — a code path the host app cannot wrap in a try/catch — producing a fatal startup crash loop ~1.5 s after every cold start (Crashlytics: repeated `RangeError: invalid field value` with the `fromFieldsV7 → generateOrAbortCore → generateOrResetCore → generate → uuidv7 → add → addPendingPromise → flushInternal → flush` JS stack). Reinstalling the app does not help; those users are hard-locked out. Related older report: #710.

## Changes

- `packages/core/src/vendor/uuidv7.ts`: clamp `getDefaultRandom().nextUint32()` with `>>> 0` (ToUint32). A misbehaving `Math.random` now degrades entropy for that one id (NaN → 0, ≥ 2^32 wraps) instead of throwing. The timestamp bits are untouched and every downstream field (`randA ≤ 0xfff`, `randBHi < 2^30`, `randBLo ≤ 0xffffffff`, counter ≤ MAX_COUNTER) is mathematically in range, so generated ids remain spec-valid UUIDv7. Conformant engines are unaffected: for in-range inputs `>>> 0` is an identity.
- `packages/core/src/vendor/uuidv7.spec.ts`: new spec asserting `V7Generator.generate()` returns a well-formed v7 UUID when `Math.random()` is stubbed to `1.0`, `1.5`, and `NaN` (all three throw today), plus a sanity check with the real `Math.random`.
- Changeset included (`@posthog/core`: patch).

### Libraries affected

The changed package is `@posthog/core`, so every SDK consuming it picks the fix up via the changesets `updateInternalDependencies` flow. The consumer where the crash was observed:

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for the impact of any changes across different platforms (identity for conformant `Math.random`; only nonconforming values change behavior, from throw to wrap)
- [x] Took care not to unnecessarily increase the bundle size (one `>>> 0`)

### If releasing new changes

- [x] Changeset file added (`.changeset/fix-uuidv7-clamp-nextuint32.md` — authored manually, same format as `pnpm changeset` output)

## 🤖 Agent context

- Authored with Claude Code during a production Crashlytics investigation directed by @shahidrogers (who previously attempted the broader #4064, closed by the stale bot).
- Decisions: an earlier approach sanitized the timestamp/inputs across the generator (#4064); this PR is deliberately narrower — since 1.46.x already validates `unixTsMs` in `generateOrAbortCore` with a distinct error message, production stacks prove the failing fields are RNG-derived, so only `nextUint32` needs guarding.
- Verified by running the patched generator (tsx) with `Math.random` stubbed to 1.0 / 1.5 / NaN — all previously threw, all now produce regex-valid v7 UUIDs. Full monorepo test suite not run locally; relying on CI.

### DIFF
--- a/.changeset/fix-uuidv7-clamp-nextuint32.md
+++ b/.changeset/fix-uuidv7-clamp-nextuint32.md
@@ -1,0 +1,5 @@
+---
+'@posthog/core': patch
+---
+
+Stop crashing when the environment's `Math.random()` misbehaves. The vendored UUIDv7 generator builds its random fields from a `Math.random()`-based `nextUint32()`, and a nonconformant implementation that returns a value of 1 or greater, or NaN, pushed those fields out of range, so `fromFieldsV7` threw `RangeError: invalid field value` on every event captured. On React Native this is not hypothetical: Hermes implements `Math.random` with C++ `std::uniform_real_distribution`, which is documented to occasionally return its upper bound, and affected Android devices crash-looped on startup during the SDK's internal event-queue flush — a path applications cannot wrap in a try/catch. `nextUint32()` now clamps its result to a valid unsigned 32-bit integer (`>>> 0`), so a bad random value degrades UUID entropy for that id instead of taking the app down; the timestamp bits are untouched and generated ids remain spec-valid UUIDv7.
--- a/.changeset/fix-uuidv7-clamp-nextuint32.md
+++ b/.changeset/fix-uuidv7-clamp-nextuint32.md
@@ -1,5 +1,6 @@
 ---
 '@posthog/core': patch
+'posthog-react-native': patch
 ---
 
 Stop crashing when the environment's `Math.random()` misbehaves. The vendored UUIDv7 generator builds its random fields from a `Math.random()`-based `nextUint32()`, and a nonconformant implementation that returns a value of 1 or greater, or NaN, pushed those fields out of range, so `fromFieldsV7` threw `RangeError: invalid field value` on every event captured. On React Native this is not hypothetical: Hermes implements `Math.random` with C++ `std::uniform_real_distribution`, which is documented to occasionally return its upper bound, and affected Android devices crash-looped on startup during the SDK's internal event-queue flush — a path applications cannot wrap in a try/catch. `nextUint32()` now clamps its result to a valid unsigned 32-bit integer (`>>> 0`), so a bad random value degrades UUID entropy for that id instead of taking the app down; the timestamp bits are untouched and generated ids remain spec-valid UUIDv7.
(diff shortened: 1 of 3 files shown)


marandaneto:
thanks @shahidrogers and good catch

.changeset/fix-uuidv7-clamp-nextuint32.md
```diff
@@ -0,0 +1,5 @@
+---
+'@posthog/core': patch
```
marandaneto, line 2: blocking: This changeset only releases `@posthog/core`, so consumers can publish without requiring the fixed core version. Please also add patch entries for every package whose runtime directly uses this UUIDv7 generator, inherits the core capture path, or ships the affected `posthog-node`: `@posthog/ai`, `@posthog/convex`, `@posthog/mcp`, `@posthog/next`, `@posthog/nuxt`, `posthog-js-lite`, `posthog-node`, and `posthog-react-native`.

marandaneto (approved):

marandaneto:
this lib is vendored so https://github.com/LiosK/uuidv7/pull/17

marandaneto:
@shahidrogers 

you need to sign your commits and force push so we can merge your pr, thanks again

shahidrogers:
Thanks @marandaneto for the speedy review and merge! Confirmed the mask is in @posthog/core 1.48.12 onwards — we'll drop our local patch when we bump.

PR #4617, opened 2026-08-24
feat(react-native): add event release mode to the build tooling

ablaszkiewicz:
## Changes

Adds event release mode flag and its propagation to ios/android tooling. Sibling PR in which everything was tested - https://github.com/PostHog/posthog/pull/87660


packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -262,6 +328,15 @@ export function buildDsymUploadShellScript(includeSource = false, skipOnConflict
...
 
+  if (releaseMode === 'event') {
+    lines.push(
+      '# Upload dSYMs without binding them to a release, so each crash resolves its own from the',
+      '# app version and namespace the SDK sends. posthog-ios versions whose upload-symbols.sh does',
+      '# not read this variable ignore it and keep binding the dSYMs.',
+      'export POSTHOG_NO_RELEASE_BIND=1'
+    )
+  }
+
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

ablaszkiewicz:
This needs to wait for the new CLI - https://github.com/PostHog/posthog/pull/87660

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -282,7 +317,7 @@ fi
 
 # Execute posthog cli clone
 set +x +e
-CLI_CLONE_OUTPUT=$("$PH_CLI_PATH" hermes clone --minified-map-path "$SOURCEMAP_PACKAGER_FILE" --composed-map-path "$SOURCEMAP_FILE" "${CLI_RELEASE_ARGS[@]}" 2>&1)
+CLI_CLONE_OUTPUT=$("$PH_CLI_PATH" hermes clone --minified-map-path "$SOURCEMAP_PACKAGER_FILE" --composed-map-path "$SOURCEMAP_FILE" "${CLI_RELEASE_ARGS[@]}" "${POSTHOG_RELEASE_MODE_ARGS[@]}" 2>&1)
```
turnipdabeets: `hermes clone` and `hermes upload` have no `--release-mode` argument in any released posthog-cli — I checked `@posthog/cli@0.15.0` (current `latest`) and `PostHog/posthog@master` (0.15.1, no `release_mode` field in `cli/src/sourcemaps/hermes/{clone,upload}.rs`). That support is still open in PostHog/posthog#87660. Ran this script at 4c95e09 against the real 0.15.0 binary with a stub `react-native-xcode.sh`:

```
POSTHOG_RELEASE_MODE=event      -> exit 2, "error: unexpected argument '--release-mode' found"
POSTHOG_RELEASE_MODE=symbol-set -> exit 1 (credentials only, my sandbox has none)
```

so `releaseMode: 'event'` fails the Xcode build right here at `exit $CLONE_EXIT_CODE`. The tests stay green because `runWrapper`'s stub CLI accepts any argument. I think we need to land #87660 first — or drop `POSTHOG_RELEASE_MODE_ARGS` altogether, since #87660 declares the arg as `env = "POSTHOG_RELEASE_MODE"` and this script already exports that, so the env var alone reaches a new CLI and an old one just ignores it.

packages/react-native/tooling/posthog.gradle
```diff
@@ -97,6 +97,20 @@ plugins.withId('com.android.application') {
...
+                        // uploaded source maps; `event` uploads them release-independent and lets
+                        // each event resolve its own release from the $app_namespace /
+                        // $app_version / $app_build the SDK already sends, which the release
+                        // coordinates below are built from. Passed only outside the default mode,
+                        // so a symbol-set build keeps working against a posthog-cli predating the
+                        // flag.
+                        def posthogReleaseMode = resolvePostHogReleaseMode(project)
+                        def posthogReleaseModeArgs = []
+                        if (posthogReleaseMode != null && posthogReleaseMode != "symbol-set") {
+                            posthogReleaseModeArgs.addAll(["--release-mode", posthogReleaseMode])
```
turnipdabeets: Same cause as the iOS side — `hermes clone`/`hermes upload` don't accept `--release-mode` in any released posthog-cli, and `execOps.exec` sets no `ignoreExitValue`, so an `event` build throws on exit 2. (I confirmed the CLI rejection directly; I didn't run a Gradle build.) I think this can go away too if we lean on the `POSTHOG_RELEASE_MODE` environment already set on both exec blocks below.

.changeset/react-native-event-release-mode.md
```diff
@@ -0,0 +1,5 @@
+---
+'posthog-react-native': minor
+---
+
+Add experimental event release mode to React Native builds. Set `releaseMode: 'event'` on the `posthog-react-native/expo` config plugin (or export `POSTHOG_RELEASE_MODE=event`, or set `posthog.releaseMode=event` in `android/gradle.properties`) and the build uploads its Hermes source maps, iOS dSYMs and Android R8 mappings without binding them to a release. Each exception then resolves its own release from the `$app_namespace` / `$app_version` / `$app_build` the SDK already sends, instead of inheriting the release of the symbols its frames resolved against. Use it when two releases can ship identical JavaScript or identical native code: symbol ids are derived from content, so the default `symbol-set` mode makes both releases report whichever one uploaded first. An unrecognized mode fails the build rather than falling back. Needs posthog-cli 0.15.0 or newer for the Hermes upload.
```
turnipdabeets: [suggestion] "Needs posthog-cli 0.15.0 or newer for the Hermes upload" — 0.15.0 is the current `latest` and rejects `--release-mode` on both `hermes clone` and `hermes upload`. Could we hold this sentence until PostHog/posthog#87660 ships and put the real version in? As written it points people at a version that fails their build.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -481,6 +600,29 @@ type PostHogPluginProps = {
...
+   * JavaScript or identical native code: symbol ids are derived from content, so in `symbol-set`
+   * mode both releases report whichever one uploaded first.
+   *
+   * Reaches every upload hook: iOS as `POSTHOG_RELEASE_MODE` in the bundle build phase (plus
+   * `POSTHOG_NO_RELEASE_BIND` in the dSYM phase when `uploadNativeSymbols` is on), Android as a
+   * `posthog.releaseMode` entry in android/gradle.properties.
+   *
+   * Requires posthog-cli >= 0.15.0 for the Hermes upload and >= 0.12.0 for the Android mapping
+   * upload. The dSYM half needs posthog-cli >= 0.10.0 and a posthog-ios whose
+   * `upload-symbols.sh` reads `POSTHOG_NO_RELEASE_BIND`; older ones ignore it and keep binding.
```
turnipdabeets: [suggestion] Two of these floors look off. `proguard upload --release-mode` first works at 0.13.0, not 0.12.0 — 0.12.0 gives "unexpected argument" (checked across clean installs of 0.12.0 / 0.13.0 / 0.14.1 / 0.15.0). And `POSTHOG_NO_RELEASE_BIND` first appears in posthog-ios 3.69.1, which is a lot easier to check against than "a posthog-ios whose `upload-symbols.sh` reads it". The Hermes number needs whatever version ships PostHog/posthog#87660.

packages/react-native/tooling/posthog-xcode.sh
```diff
@@ -33,12 +33,21 @@ print_command_error() {
...
+POSTHOG_RELEASE_MODE_VALUE="${POSTHOG_RELEASE_MODE:-}"
 while [ "$#" -gt 0 ]; do
   case "$1" in
     --posthog-skip-on-conflict)
       POSTHOG_SKIP_ON_CONFLICT_ENABLED=1
       shift
       ;;
+    --posthog-release-mode)
+      POSTHOG_RELEASE_MODE_VALUE="$2"
+      shift 2
```
turnipdabeets: [nit] Nothing the plugin generates passes `--posthog-release-mode` — `buildBundlePhaseExports` only emits `export` lines — so this is an escape hatch for hand-written build phases. With `set -e` on, `shift 2` with no value aborts the wrapper before bundling and prints nothing (`+ shift 2` then `exit=1`). Worth a `[ "$#" -ge 2 ]` guard with a message, or dropping the flag and leaving the env var as the one hatch.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -14,6 +14,33 @@ const resolvePostHogReactNativePackageJsonPath =
...
+/**
+ * How the release a build belongs to gets associated with the exceptions it reports.
+ * `symbol-set` stamps the release onto the uploaded source maps, dSYMs and mappings, and an
+ * exception inherits the release of the symbols its frames resolved against. `event` uploads them
+ * release-independent and lets each event resolve its own release from the app version and
+ * namespace the SDK already sends.
+ */
+export type PostHogReleaseMode = 'symbol-set' | 'event'
+
+const POSTHOG_RELEASE_MODES: PostHogReleaseMode[] = ['symbol-set', 'event']
```
turnipdabeets: [suggestion] This list now lives in four places — here, the `case` in `buildDsymUploadShellScript`, `symbol-set|event)` in posthog-xcode.sh, and `["symbol-set", "event"]` in posthog.gradle — and nothing fails when they drift, so a third CLI mode gets rejected at prebuild by whichever copy was missed. `posthog-xcode-parse.spec.ts` already reads the shell source off disk; a test asserting the `.sh` and `.gradle` lists match `POSTHOG_RELEASE_MODES` would be cheap here.

packages/react-native/tooling/posthog-xcode.sh
turnipdabeets, line 367: [question] Assuming this waits for PostHog/posthog#87660, the flag will exist by the time it ships — but we don't pin posthog-cli, we use whatever is on the box. Someone still on 0.15.0 who sets `releaseMode: 'event'` gets `error: unexpected argument '--release-mode' found` and exit 2 right here, with nothing pointing at "upgrade posthog-cli" (ran this script at 4c95e09 against the real 0.15.0 binary to check). Since #87660 declares the arg as `env = "POSTHOG_RELEASE_MODE"` and we already export that above, dropping `POSTHOG_RELEASE_MODE_ARGS` would leave an old CLI quietly on symbol-set instead — or we keep the flag and add a version check with a real message. Which trade do we want?

packages/react-native/tooling/posthog.gradle
turnipdabeets, line 111: [nit] Same trade as the iOS side: `execOps.exec` sets no `ignoreExitValue`, so a user on a posthog-cli predating PostHog/posthog#87660 gets a raw clap error and a failed build here rather than a message telling them to upgrade. Whatever we settle on for the shell script probably wants mirroring here.

.changeset/react-native-event-release-mode.md
turnipdabeets: [suggestion] This will need the version that actually carries the flag once PostHog/posthog#87660 ships — 0.15.0 is today's `latest` and rejects `--release-mode` on both `hermes clone` and `hermes upload`, so as written the release notes point people at a build failure. Worth a TODO on this line so it doesn't go out with 0.15.0 in it.

packages/react-native/src/tooling/expoconfig.ts
turnipdabeets: [suggestion] Two of these floors are off independently of #87660. `proguard upload --release-mode` first works at 0.13.0, not 0.12.0 — 0.12.0 gives "unexpected argument" (checked across clean installs of 0.12.0 / 0.13.0 / 0.14.1 / 0.15.0). And `POSTHOG_NO_RELEASE_BIND` first appears in posthog-ios 3.69.1, which is a lot easier to check against than "a posthog-ios whose `upload-symbols.sh` reads it". The Hermes number will need updating to whatever #87660 ships as.

packages/react-native/tooling/posthog-xcode.sh
turnipdabeets, line 50: [nit] Nothing the plugin generates passes `--posthog-release-mode` — `buildBundlePhaseExports` only emits `export` lines — so this is an escape hatch for hand-written build phases. With `set -e` on, `shift 2` with no value aborts the wrapper before bundling and prints nothing (`+ shift 2` then `exit=1`). Worth a `[ "$#" -ge 2 ]` guard with a message, or dropping the flag and leaving the env var as the one hatch.

packages/react-native/src/tooling/expoconfig.ts
turnipdabeets: [suggestion] This list now lives in four places — here, the `case` in `buildDsymUploadShellScript`, `symbol-set|event)` in posthog-xcode.sh, and `["symbol-set", "event"]` in posthog.gradle — and nothing fails when they drift, so a third CLI mode gets rejected at prebuild by whichever copy was missed. `posthog-xcode-parse.spec.ts` already reads the shell source off disk; a test asserting the `.sh` and `.gradle` lists match `POSTHOG_RELEASE_MODES` would be cheap here.

packages/react-native/tooling/posthog-xcode.sh
ablaszkiewicz: Added a version check

packages/react-native/tooling/posthog.gradle
ablaszkiewicz: Mirrored

.changeset/react-native-event-release-mode.md
ablaszkiewicz: I replaced 0.15.0 with TODO:PLACEHOLDER to be triple sure I don't forget. Will leave this comment unresolved also until I merge that other PR

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: Confirmed both and fixed - you were right

packages/react-native/tooling/posthog-xcode.sh
ablaszkiewicz: Good catch. Reproduced that. Added guard with a message

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: Added a test

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -296,24 +386,33 @@ function decodePbxShellScript(stored: string): string {
...
+  includeSource = false,
+  skipOnConflict = false,
+  releaseMode?: PostHogReleaseMode
+): void {
   const existing = xcodeProject.pbxItemByComment(POSTHOG_DSYM_BUILD_PHASE_NAME, 'PBXShellScriptBuildPhase')
   if (existing) {
     const generatedVariants = [false, true].flatMap((source) =>
-      [false, true].map((skip) => buildDsymUploadShellScript(source, skip))
+      [false, true].flatMap((skip) =>
+        [undefined, ...POSTHOG_RELEASE_MODES].map((mode) => buildDsymUploadShellScript(source, skip, mode))
```
marandaneto: blocking: Preserve migration support for existing dSYM phases — `generatedVariants` now contains only scripts produced by the new implementation. A project generated by an earlier SDK has the old script without the release-mode block, so it no longer matches and is left unchanged when event mode is enabled; Hermes maps become release-independent while dSYMs remain release-bound. Include the previously generated variants in this ownership check.

packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -405,9 +505,28 @@ export function updateDotenvFileGradleProperties(
...
+  properties: GradlePropertiesItem[],
+  releaseMode?: PostHogReleaseMode
+): GradlePropertiesItem[] {
+  const rest = properties.filter(
+    (item) => !(item.type === 'property' && item.key === POSTHOG_RELEASE_MODE_GRADLE_PROPERTY)
+  )
+  if (!releaseMode) {
+    return rest
+  }
+  rest.push({ type: 'property', key: POSTHOG_RELEASE_MODE_GRADLE_PROPERTY, value: releaseMode })
```
marandaneto, line 578: blocking: Inject a release-mode-aware Android plugin — This writes `posthog.releaseMode`, but the Expo plugin still injects `posthog-android-gradle-plugin` 1.4.0, which ignores this property; support starts in 1.5.0. With native-symbol upload enabled, event mode therefore leaves the R8 mapping release-bound while Hermes maps are unbound.

marandaneto (commented):

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: You were right, thanks. The 4.63 text matched none of the new variants, because every variant now carries the runtime `case` block. Fixed in 2330985844127b2afa2ba53d571fa5a2b47817d3: the generator is split into a shared skeleton plus the release-mode block, and `buildLegacyDsymUploadShellScript` (the skeleton with no block, which is the exact pre-release-mode text) is now part of the ownership list. Added a test that feeds the verbatim 4.63 script and expects it refreshed to event mode.

It did not show up in testing because the example app always prebuilds with `--clean`, so no old phase ever existed.

packages/react-native/src/tooling/expoconfig.ts
ablaszkiewicz: Right, thanks. Testing ran against a locally built plugin, so this did not show up. Fixed in 2330985844127b2afa2ba53d571fa5a2b47817d3: the plugin now injects 1.5.1 (1.5.0 is the first version that reads `posthog.releaseMode`), and the prop docs and changeset state the floor. Also documented that a project whose `android/build.gradle` already carries the classpath line keeps its version, since the injection is skipped when the line exists, so it has to be bumped by hand or with a clean prebuild.

.changeset/react-native-event-release-mode.md
ablaszkiewicz: Bumped!

ablaszkiewicz:
@PostHog/team-client-libraries this is now ready to merge. I've merged that other PR and bumped version here

turnipdabeets (approved):

PR #4618, opened 2026-08-24
fix(react-native): key the release on Info.plist, not the build settings

ablaszkiewicz:
## Problem

An Expo app's release in PostHog does not describe the app that shipped. A build whose app reports version `1.0.0` creates its release as `1.0`.

The SDK reports `$app_version` and `$app_build` from the app's `Info.plist`. `posthog-xcode.sh` passed Xcode's `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION` instead. Those are only the usual source for those plist keys. Expo writes literal versions into `Info.plist` and leaves the build settings at the Xcode template default of `1.0`, so the two disagree.

Seen on a real simulator build of a stock Expo app:

```
Info.plist   CFBundleShortVersionString = 1.0.0    ← what the SDK reports
pbxproj      MARKETING_VERSION          = 1.0      ← what the release was created from
```

The release row landed as `com.posthog.example.rnexpo@1.0+1`.

A bare React Native app writes `$(MARKETING_VERSION)` into its plist instead, so it never diverged and is unaffected.

## Changes

- An iOS build now keys its release on the values in the app's `Info.plist`.
- The build setting stays the fallback, used when the plist holds an unexpanded reference such as `$(MARKETING_VERSION)` or has no value at all.
- The bundle identifier still comes from `PRODUCT_BUNDLE_IDENTIFIER`. `Info.plist` references it rather than repeating it, and the SDK reports the resolved value.

> [!NOTE]
> An existing Expo project on the default release mode will start creating release rows under its real version. Releases already created under the old version are untouched, so expect a new row rather than a changed one.

The test file gains a helper that runs the wrapper against a posthog-cli stub and records its arguments. That helper is new surface, not a change to existing tests.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

Found by building the `react-native-expo` example in PostHog/error-tracking-examples end to end on the iOS simulator, not by reading the code. Before this change the build created `com.posthog.example.rnexpo@1.0+1` while the app on screen reported `1.0.0`. After it, the same build creates `@1.0.0+1`.

Three new cases, each running the real script against a posthog-cli stub:

- `keys the release on Info.plist rather than the build settings` — the regression itself, a literal plist version losing to `MARKETING_VERSION`.
- `falls back to the build settings when Info.plist only references them` — a bare React Native app breaking, which is the larger installed base.
- `falls back to the build settings when there is no Info.plist at all` — the same, for a project that sets no `INFOPLIST_FILE`.

Not checked: an Android build, which does not use this script, and a physical iOS device.

## 🤖 Agent context

Authored with Claude Code (Claude Opus 5). Skills invoked: `/writing-tests`, `/writing-code-comments`, `/writing-pr-descriptions`.

Split out of PostHog/posthog-js#4617, which stacks on this branch. That PR adds experimental event release mode, where this bug stops being cosmetic: the server rebuilds the release key from the event's app metadata, matches nothing, and every exception reports no release. The fix stands on its own, so it is here rather than buried in the feature.

Nothing in this PR draws on non-public material.

### DIFF
--- a/.changeset/react-native-release-coordinates.md
+++ b/.changeset/react-native-release-coordinates.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+Key the release an iOS build uploads on the app's Info.plist rather than on Xcode's `MARKETING_VERSION` and `CURRENT_PROJECT_VERSION`. The SDK reports `$app_version` and `$app_build` from Info.plist, and Expo writes literal versions there while leaving the build settings at the Xcode template default of `1.0`. A build whose app reported `1.0.0` therefore created its release as `1.0`, so the release did not describe the app that shipped. Releases created by an Expo build now carry the app's real version, which means a project on the default release mode will start creating release rows under that version instead. A project whose Info.plist references the build settings, as a bare React Native app does, is unaffected.
--- a/packages/react-native/test/posthog-xcode-parse.spec.ts
+++ b/packages/react-native/test/posthog-xcode-parse.spec.ts
@@ -1,4 +1,4 @@
-import { execFileSync, execSync } from 'child_process'
+import { execFileSync, execSync, spawnSync } from 'child_process'
 import * as fs from 'fs'
 import * as os from 'os'
 import * as path from 'path'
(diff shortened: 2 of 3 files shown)


marandaneto:
dupe https://github.com/PostHog/posthog-js/pull/4602 ?

ablaszkiewicz:
@marandaneto yeah that's duplicated. Closing mine!

PR #4626, opened 2026-08-24
fix(surveys): re-translate displayed surveys when user language changes

### LINKED ISSUE #4174: Re-translate displayed survey on language change

nandinitiw:
Reopening #4195 as a fresh PR — the original was auto-closed by the stale bot after a week of no activity, not for cause. lucasheriques's last review was positive ("the re-translate path holds up") and all requested changes were already addressed; the GitHub reopen API is failing with an opaque validation error, so opening fresh from the same branch instead. Same commits, same diff.

## What

When a survey popover is on screen, it was never re-translated if the user's language changed mid-session. This happened in two scenarios:
1. `posthog.identify()` / `setPersonPropertiesForFlags()` sets a new `language` person property
2. The browser fires a `languagechange` event (OS/browser language switch)

The TODO comment in `survey-translations.ts` already called this out as a known gap.

Also fixes a gap this surfaced: `$survey_questions[].question` and `$survey_language` were read at capture time, not answer time — if the language changed mid-survey, submitted events showed the *new* language's question text even for questions the user answered under the *old* language.

## How

In `SurveyManager`:
- Register a `window.languagechange` listener in the constructor
- Subscribe to `posthog.onFeatureFlags()` (which fires after `identify()` reloads flags)
- Both call `_onLanguageChange()`, which:
  - No-ops if no survey is in focus, or if the survey hasn't actually rendered yet (gates against a delayed survey — `surveyPopupDelaySeconds` — being forced on screen early by a mid-delay reload, skipping the eligibility recheck the delay timeout does)
  - Re-detects the resolved language
  - No-ops if the language is unchanged (idempotent, safe to call on every flag reload)
  - Otherwise updates `_currentLanguage` and re-renders the popover via `renderPopover()`
- Listeners are cleaned up on `destroy()`
- `_currentLanguage` is cleared when a survey is dismissed

**`questionSnapshots`**: `onNextButtonClick` now snapshots the currently-displayed question text into `questionSnapshots: Record<string, string>` (question id → text) before advancing, and persists it to `InProgressSurveyState` in localStorage. `buildSurveyResponseProperties` in `@posthog/core` now accepts this optional param and prefers snapshot text over the current (possibly re-translated) survey question text, so `$survey_questions` reflects what the user actually saw when they answered. `$survey_language` on `dismissed` now similarly prefers the in-progress language over the current display-time language.

The iOS reference implementation in [posthog-ios #686](https://github.com/PostHog/posthog-ios/pull/686) follows the same pattern.

## Tests

Unit tests covering:
- Re-renders when language differs
- Does not re-render when language is unchanged
- No-ops when no survey is in focus
- Does not re-render a survey that's in focus but not yet rendered (pending delay)
- Clears `_currentLanguage` on dismiss
- Removes `languagechange` listener on `destroy()`
- `questionSnapshots` recorded and used in place of current question text
- A real end-to-end render test: renders a popup via the unmocked `handlePopoverSurvey`, types an answer, fires a real `languagechange` event, and asserts both that the rendered question text updates and that the typed answer survives the re-render

Closes #4174


packages/browser/src/__tests__/extensions/surveys.test.ts
```diff
@@ -496,6 +497,14 @@ describe('SurveyManager', () => {
         surveyManager = new SurveyManager(mockPostHog)
     })
 
+    afterEach(() => {
+        // SurveyManager attaches a 'languagechange' window listener on construction (see
+        // surveys.tsx). Without destroying it, listeners from earlier tests' instances stay
+        // attached and fire on later tests' window.dispatchEvent(new Event('languagechange')),
+        // hitting a stale mockPostHog whose mocks may no longer exist.
+        surveyManager.destroy()
```
turnipdabeets: Confirmed with a run on this head: `jest src/__tests__/extensions/surveys.test.ts` fails with `TypeError: this._posthog.get_property is not a function` inside `_onLanguageChange`, while the same test passes on its own via `-t "updates the rendered question text"`. It passed at the pre-merge head (92/92), so main's newer tests shifted the ordering. This `afterEach` only destroys whatever is in `surveyManager` at that moment — the nested `describe`s reassign it (`new SurveyManager(mockPH)`) and orphan the outer instance, whose `languagechange` listener stays attached. I think we need to track every instance and destroy each one here.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -141,11 +141,63 @@ export class SurveyManager {
...
+
+        this._languageChangeListener = () => this._onLanguageChange()
+        addEventListener(window, 'languagechange', this._languageChangeListener)
+
+        // Re-translate when identify() or setPersonPropertiesForFlags() reloads flags,
+        // which may have updated the 'language' person property.
+        this._unsubscribeFeatureFlags = posthog.onFeatureFlags(() => this._onLanguageChange())
+    }
+
+    public destroy(): void {
```
turnipdabeets: `destroy()` has no production caller — teardown goes through `dispose()`, which `posthog-surveys.ts` calls as `this._surveyManager?.dispose?.()`. `dispose()` landed in #4532 after this branch was written, so as it stands the `languagechange` listener and the `onFeatureFlags` subscription both survive a `posthog.surveys.dispose()`. I think we need to move this cleanup into `dispose()`.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -141,11 +141,63 @@ export class SurveyManager {
...
+            this._languageChangeListener = null
+        }
+        if (this._unsubscribeFeatureFlags) {
+            this._unsubscribeFeatureFlags()
+            this._unsubscribeFeatureFlags = null
+        }
+    }
+
+    private _onLanguageChange(): void {
+        if (isNull(this._surveyInFocus) || !this._surveyIsRendered) {
```
turnipdabeets: [question] `_handleWidget` never calls `_addSurveyToFocus`, so `_surveyInFocus` stays null for `feedback_button` surveys and this returns on the first line — and the render below builds a `SurveyPopup` rather than a `FeedbackWidget`. #4174 mentions widgets too; is popover-only the intended scope here?

.changeset/survey-retranslate-on-language-change.md
```diff
@@ -0,0 +1,6 @@
+---
+'posthog-js': patch
+'@posthog/core': patch
+---
+
+fix(surveys): re-translate displayed surveys when the user's language changes mid-session — handles both browser `languagechange` events and `posthog.identify()` calls that update the `language` person property. Also snapshots each question's displayed text and language at answer time (`questionSnapshots`), so `$survey_questions` and `$survey_language` reflect what the user actually saw rather than the language active when the event was sent.
```
turnipdabeets: [suggestion] "displayed surveys" reads as every survey type, but widgets aren't covered (see the `_surveyInFocus` gate). Could we name popovers explicitly so the changelog doesn't over-promise?

```suggestion
Re-translate popover surveys when the display language changes while the survey is on screen, either from a browser `languagechange` event or from `identify()` updating the `language` person property. In-progress answers are preserved. `$survey_questions[].question` and `$survey_language` on `survey sent` / `survey dismissed` now report the text and language the user saw when they answered, not the language active when the event fired. Feedback-button (widget) surveys are unchanged.
```

packages/core/src/surveys/events.ts
```diff
@@ -45,7 +46,7 @@ export function buildSurveyResponseProperties(
   return {
     $survey_questions: survey.questions.map((question: SurveyQuestionForResponses) => ({
       id: question.id,
-      question: question.question,
+      question: (question.id && questionSnapshots?.[question.id]) ?? question.question,
```
turnipdabeets: [nit] `??` only catches null/undefined, so a question with `id: ''` gives `'' ?? question.question` → `''`. Undefined ids fall through fine. `questionSnapshots?.[question.id ?? ''] ?? question.question` would cover it, if empty ids are reachable.

turnipdabeets (commented):
Thanks for reopening this and for the earlier rounds. Confirmed the diff here is identical to #4195.

A few things need a push before we can merge:

1. Need to address merge conflict
2. The end-to-end test you added now fails on this head. It passes in isolation but fails in the full file — main added tests to that describe and the ordering shifted.
3. #4532 also introduced SurveyManager.dispose(), which is what posthog-surveys.ts actually calls on teardown — destroy() has no production caller anymore. Could you move the listener/subscription cleanup into dispose() so the languagechange listener and onFeatureFlags subscription don't leak on reset?
4. All your commits need to be verified and signed to merge the PR.

I'll help run CI once the above is pushed.

nandinitiw:
Thanks for the detailed review! Addressed the first three:

1. **Merge conflict**: rebased on latest main and resolved. Main had reworked the delayed-render path into `renderAfterDelay`/`renderIfStillEligible` helpers (for delay-resumption-across-navigations) since this diff was last synced, so `_surveyIsRendered = true` now lives in the new `renderAfterDelay` function.

2. **The end-to-end test failing in the full file**: traced this down — several *other* `SurveyManager` instances constructed elsewhere in the same test file were never disposed, so their `languagechange` window listeners stayed attached and fired on my test's real `dispatchEvent` call. Found and fixed three separate leaks: (a) several nested describes reassign the shared `surveyManager` variable without disposing the previous instance — added a `createSurveyManager()` helper that tracks every constructed instance so cleanup covers all of them; (b) a `jest.clearAllMocks()` in the "timeout management" describe left a `jest.spyOn(global, 'clearTimeout')` spy installed rather than restored, which went stale once fake timers tore down — switched to `jest.restoreAllMocks()`; (c) `generateSurveys()`'s own test never disposed the manager it returns — now does, in a `finally`.

3. **`dispose()` vs `destroy()`**: moved the `languagechange` listener removal and `onFeatureFlags` unsubscribe into `dispose()` and removed `destroy()` entirely, since it has no production caller.

4. **Signed commits**: I don't have commit signing configured on this machine yet — need to sort that out on my end before these can be verified. Will follow up once that's set up.

Full survey test suite (268 tests across `surveys.test.ts`, `extensions/surveys.test.ts`, `survey-popup.test.tsx`, `posthog-surveys.test.ts`) passes, along with lint and prettier.

nandinitiw:
All commits on this branch are now signed and verified (SSH signing key registered with GitHub) — re-signed the full commit history and force-pushed. Should show as Verified now.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -144,11 +144,52 @@ export class SurveyManager {
...
+        render(
+            <SurveyPopup
+                posthog={this._posthog}
+                survey={translatedSurvey}
+                removeSurveyFromFocus={this._removeSurveyFromFocus}
+                surveyLanguage={newLanguage}
+                style={this._surveyPopupProps?.style}
+                properties={this._surveyPopupProps?.properties}
+                isSurveyCompleted={this._surveyPopupProps?.isSurveyCompleted}
+            />,
```
turnipdabeets, line 206: `_onLanguageChange` re-renders from `translatedSurvey`, but `handlePopoverSurvey` renders the *override-merged* survey — the one with `options.position` / `options.selector` folded into `appearance`. So a survey shown via `posthog.displaySurvey(id, { displayType: 'popover', position: 'top_left' })` snaps back to its configured position on the first language change. Confirmed against this head:

```
before: top: 0px; left: 0px; transform: translate(30px, 30px);
after : right: 30px;
```

`skipShownEvent` is dropped the same way — harmless today only because `usePopupVisibility`'s effect is `[]`-gated and preact patches in place rather than remounting.

I think this needs the overrides re-applied to the freshly translated survey, not just the stored props spread back in — translating from the raw survey is right (so switching to a language with no translation still falls back to the original), but the merge has to happen again after it. What worked when I tried it: pull the `appearance` merge out of `handlePopoverSurvey` into a small `_applyDisplayOverrides(survey, options)`, keep `options` alongside the popup props, and call it here on `translatedSurvey`. Widening `_surveyPopupProps` to the full `SurveyPopupProps` and spreading it then picks up `skipShownEvent` for free.  With that change applied, surveys.test.ts, extensions/surveys.test.ts, survey-popup.test.tsx and posthog-surveys.test.ts all still pass.

packages/core/src/surveys/events.ts
turnipdabeets: [nit] `''` isn't nullish, so a question with an empty-string id would report an empty `question` here instead of falling back to `question.question`. Not reachable with server-assigned ids, but the precedence is easy to sidestep:

```suggestion
      question: questionSnapshots?.[question.id ?? ''] ?? question.question,
```

turnipdabeets (commented):
Thanks for the updated! One last comment and nit.

nandinitiw:
Thanks for the thorough follow-up! Addressed everything:

**Position/selector overrides lost on language change** — confirmed and fixed. Extracted the override-merge logic out of `handlePopoverSurvey` into `_applyDisplayOverrides(survey, options)`, stored `options` alongside `_surveyPopupProps`, and `_onLanguageChange` now re-derives the survey from the fresh translation (so falling back to the original when there's no translation for the new language still works) and re-applies the same overrides before rendering. Widened `_surveyPopupProps` to the full `SurveyPopupProps` and spread it in the re-render, so `skipShownEvent` comes along for free rather than needing its own explicit carry-through.

Added a regression test: renders via `handlePopoverSurvey` with a `position` override, fires a real `languagechange`, and asserts the position style survives — verified it fails without the fix (snaps to the default `right: 30px` instead of staying at `top: 0; left: 0`) and passes with it.

**`events.ts` empty-string id nit** — applied your suggested fix exactly (`questionSnapshots?.[question.id ?? ''] ?? question.question`), and added tests for it plus the previously-untested snapshot-present/snapshot-absent cases.

**Changeset wording** — used your suggested text verbatim, naming popovers explicitly and calling out that widget surveys are unchanged.

**Widget scope question** — yes, popover-only is the intended scope for this PR; #4174 is tracked as popover behavior and widget re-translation would need its own `_addSurveyToFocus` wiring in `_handleWidget`, which felt like a separate change. Happy to file a follow-up issue for that if useful.

Full survey suite (269 tests) + the new `events.spec.ts` tests + lint + prettier all pass. Mangled property names regenerated for the two new private members (`_applyDisplayOverrides`, `_displayOptions`).

packages/browser/src/extensions/surveys.tsx
```diff
@@ -144,11 +144,73 @@ export class SurveyManager {
...
+        this._languageChangeListener = () => this._onLanguageChange()
+        addEventListener(window, 'languagechange', this._languageChangeListener)
+
+        // Re-translate when identify() or setPersonPropertiesForFlags() reloads flags,
+        // which may have updated the 'language' person property.
+        this._unsubscribeFeatureFlags = posthog.onFeatureFlags(() => this._onLanguageChange())
+    }
+
+    private _onLanguageChange(): void {
+        if (isNull(this._surveyInFocus) || !this._surveyIsRendered) {
```
turnipdabeets: A language flip while a delayed survey is still counting down gets lost, so the survey pops up in the old language. This returns before line 183 updates `_currentLanguage`, so nothing records that the flip happened.

Confirmed: 5s delay, flipped `en`->`fr` mid-delay, advanced timers, got English. posthog-ios fixed the same case in [#686](https://github.com/PostHog/posthog-ios/pull/686) ([`dcd38baa3`](https://github.com/PostHog/posthog-ios/pull/686/commits/dcd38baa3bfd1da3a8294955d1c105f9d6a0a57a)).

I think we want to update `_currentLanguage` here and skip only the `render()`.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1565,6 +1638,10 @@ export function Questions({
             (index) => index >= 0 && index < survey.questions.length
         )
     })
+    const [questionSnapshots, setQuestionSnapshots] = useState<Record<string, string>>(() => {
+        const inProgressSurveyData = getInProgressSurveyState(survey)
+        return inProgressSurveyData?.questionSnapshots ?? {}
+    })
     const surveyQuestions = useMemo(
         () => getDisplayOrderQuestions(survey, initialInProgressState),
         [survey, initialInProgressState]
```
turnipdabeets: On a shuffled survey, a language change swaps the question the user is looking at and drops whatever they'd typed. The re-render hands `survey` a new identity, so this memo re-runs and reshuffles.

Confirmed: 8-12 of 12 trials showed a different question after the flip; 0/12 on `main`.

I think we need to key the order on `survey.id` and map those ids onto the current questions — text re-translates, order holds.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1616,6 +1702,7 @@ export function Questions({
                 questionOrder: getQuestionOrder(surveyQuestions),
                 visitedIndices: newVisitedIndices,
                 surveyLanguage,
+                questionSnapshots: newSnapshots,
```
turnipdabeets, line 1760: Pressing Back wipes the snapshots we save here. `onBackButtonClick`'s `setInProgressSurveyState` call omits the field, and the helper replaces the record rather than merging.

Confirmed: `{"a":"Q-a-EN"}` before Back, `undefined` after. A dismiss after that falls back to current-language text.

I think adding `questionSnapshots` to that call covers it.

packages/browser/src/extensions/surveys/surveys-extension-utils.tsx
```diff
@@ -491,9 +494,13 @@ export const dismissedSurveyEvent = (
     }
 
     const inProgressSurvey = getInProgressSurveyState(survey)
+    // Prefer the language snapshotted when the user last answered (answer-time language).
+    // Only fall back to the current display language when no in-progress state exists
+    // (i.e. the survey was dismissed without answering any question).
+    const effectiveLanguage = inProgressSurvey?.surveyLanguage ?? surveyLanguage
```
turnipdabeets: [suggestion] Answering in the default text and then switching to French reports `$survey_language: "fr"`. The stored language is `null` when no translation matched, and `??` treats that as absent.

Checking whether the record exists, rather than its value, would match the comment above.

packages/browser/src/extensions/surveys.tsx
```diff
@@ -931,6 +997,10 @@ export class SurveyManager {
         }
         this._clearSurveyTimeout(survey.id)
         this._surveyInFocus = null
+        this._currentLanguage = null
+        this._surveyIsRendered = false
+        this._surveyPopupProps = null
+        this._displayOptions = undefined
```
turnipdabeets, line 1038: [nit] `cancelSurvey` clears `_surveyInFocus` but leaves these four set. I couldn't turn it into a real bug (`usePopupVisibility` gates on the delay independently), but one nullable `_activeRender` cleared alongside focus would make it correct by construction.

packages/browser/src/__tests__/surveys.test.ts
```diff
@@ -1847,4 +1848,102 @@ describe('surveys', () => {
...
+            name: 'Lang Survey',
+            type: SurveyType.Popover,
+            questions: [{ type: SurveyQuestionType.Open, question: 'Hello?', id: 'q1', description: '' }],
+            appearance: null,
+            conditions: null,
+            start_date: '2024-01-01T00:00:00Z',
+            end_date: null,
+            current_iteration: null,
+            current_iteration_start_date: null,
+            translations: [
```
turnipdabeets: [nit] This `translations: [...]` array shape isn't in the `Survey` type — per-question translations are `Record<string, SurveyQuestionTranslation>` (`packages/core/src/types.ts:745`). These tests mock `_translateSurveyForRendering`, so they'd pass against a broken translator too.

turnipdabeets (commented):
A few more connects and a merge conflict

nandinitiw:
Rebased on latest main (conflict resolved) and fixed all four bugs found in this review round:

**1. Mid-delay language flip was lost** — confirmed and fixed. `_onLanguageChange` returned before updating `_currentLanguage` when the survey was still counting down its popup delay, so when the delay elapsed the popup rendered in whatever language was active when the delay *started*, not the current one. Now `_currentLanguage` updates regardless of render state, and `renderAfterDelay` re-translates from the original survey when the delay actually elapses rather than reusing what was captured at `handlePopoverSurvey` call time. This matches the approach in [posthog-ios#686](https://github.com/PostHog/posthog-ios/pull/686/commits/dcd38baa3bfd1da3a8294955d1c105f9d6a0a57a).

**2. Shuffled survey reshuffled on every language change** — confirmed and fixed. The display-order `useMemo` was keyed on the `survey` object, which gets a new identity on every re-translation. Now the order is computed once per `survey.id` (question ids only) and mapped onto whatever the current (possibly re-translated) question objects are — text updates, order doesn't. Worth noting for the test: with only 2 questions, `reverseIfUnshuffled` makes the "shuffled" result deterministic (there's only one non-identity permutation), which is why the bug wasn't visible in a 2-question test — used 4 questions × 10 trials with fresh survey ids to make the regression test reliable.

**3. Back button wiped `questionSnapshots`** — confirmed and fixed with the change you suggested (added the field to `onBackButtonClick`'s `setInProgressSurveyState` call).

**4. `$survey_language` on dismiss could report the wrong language** — confirmed and fixed with the check you suggested (record presence, not value).

Also fixed the test fixture nit — moved `translations` onto the question object in the correct `Record<string, SurveyQuestionTranslation>` shape.

**On the `cancelSurvey` nit**: rather than the full `_activeRender` refactor, I went with the more contained fix of also clearing `_currentLanguage`/`_surveyIsRendered`/`_surveyPopupProps`/`_displayOptions` there, matching what `_removeSurveyFromFocus` already does. Happy to do the fuller refactor if you'd still prefer it.

Added a regression test for each of the four bugs, each verified locally to fail against the pre-fix code and pass against the fix. Full survey suite (294 tests across 8 files) + lint + prettier all pass. All commits signed and verified.

turnipdabeets (approved):
Approving + #4195 already has an approval from @ioannisj and 👀 from @lucasheriques

packages/browser/src/extensions/surveys.tsx
```diff
@@ -1573,10 +1673,32 @@ export function Questions({
...
+            // Some questions lack an id (older data) — there's no reliable way to map a stable
+            // order onto fresh question objects, so fall back to recomputing fresh each time.
+            return getDisplayOrderQuestions(survey, initialInProgressState)
+        }
+        const byId = new Map(survey.questions.map((question) => [question.id, question]))
+        const mapped = questionOrderIds
+            .map((id) => byId.get(id))
+            .filter((question): question is SurveyQuestion => !!question)
+        return mapped.length === questionOrderIds.length ? mapped : survey.questions
+    }, [questionOrderIds, survey.questions, survey, initialInProgressState])
```
turnipdabeets, line 1703: CI caught this one, and we all missed it in the earlier passes. When two questions share an id — the E2E survey spreads `npsRatingQuestion` twice — the Map keeps only the last one, so question 1 renders as the scale-5 question: `.ratings-number` expected 11, got 5 on all three browsers.

Since `applySurveyTranslation` maps `survey.questions` 1:1, we can carry the order as positions rather than ids — nothing to collide, and the `!questionOrderIds` branch can go away too (it recomputes, and so reshuffles, on every language change for questions without ids). Something like this does it — `question-types.spec.ts` passes and the 85 mocked survey specs stay green:

```suggestion
    // reshuffle — the order on every language change. Compute the order once per survey.id and
    // carry it as positions in survey.questions: applySurveyTranslation rewrites question text
    // in place, never the number of questions or their positions, so the same positions applied
    // to the re-translated questions give the same order with updated text.
    const questionOrder = useMemo(() => {
        const positions = getDisplayOrderQuestions(survey, initialInProgressState).map((question) =>
            survey.questions.indexOf(question)
        )
        // getDisplayOrderQuestions only ever returns questions drawn from survey.questions, so a
        // missing position means an unexpected shape; fall back to the configured order.
        return positions.some((position) => position < 0) ? null : positions
    }, [survey.id, initialInProgressState])
    const surveyQuestions = useMemo(
        () => questionOrder?.map((position) => survey.questions[position]) ?? survey.questions,
        [questionOrder, survey.questions]
    )
```

nandinitiw:
Fixed the failing Playwright E2E test (\`question-types.spec.ts\` — \"rating questions that are on the 10 scale start at 0\").

**Root cause:** the shuffle-reshuffling fix from the previous round mapped stable question ids to freshly-translated question objects via a \`Map\` keyed by \`question.id\`. That test fixture uses two questions sharing the same id (\`nps_rating_1\`), so the Map collapsed them into one entry and the survey rendered the last question twice instead of both distinct ones.

**Fix:** detect duplicate ids in \`questionOrderIds\` and fall back to recomputing the display order directly in that case, the same fallback path already used when questions are missing ids entirely.

Verified: full unit suite (294 tests) passes, the specific failing E2E test now passes, and the full \`playwright/mocked/surveys/\` suite (85 tests) passes cleanly.

PR #4628, opened 2026-08-24
feat(replay): capture clipboard interactions

robbie-c:
## Summary

- capture cut, copy, and paste actions as rrweb custom `$clipboard` events
- record only the action type, text length, and rrweb node ID, never clipboard contents
- respect replay block selectors and `.ph-no-capture` before reading clipboard metadata
- support Shadow DOM and same-origin iframe documents through an rrweb observer plugin
- count clipboard activity in replay idle and fresh-session handling
- add a `posthog-js` patch changeset

## Privacy

Clipboard text is not recorded. Each event contains the clipboard action, the string length, and the rrweb node ID when one is available. Blocked elements do not produce clipboard events.

## Test plan

- [x] 378 focused replay unit tests pass
- [x] Clipboard unit coverage includes cut, copy, paste, blocked elements, invalid clipboard data, Shadow DOM, iframes, listener cleanup, idle handling, and fresh-session behavior
- [x] Playwright compatibility test passes with published `posthog-js@1.418.13` and the new recorder bundle
- [x] Lint, Prettier, and browser source type checks pass
- [x] Recorder bundle builds complete and emit output

## Build note

The full Rollup build emitted all bundles and declarations, but the process stayed alive after the final output in two runs. Focused lazy-recorder and recorder builds exited successfully.

### DIFF
--- a/.changeset/tidy-cats-clip.md
+++ b/.changeset/tidy-cats-clip.md
@@ -1,0 +1,5 @@
+---
+'posthog-js': patch
+---
+
+Record cut, copy, and paste interactions in session replays without storing clipboard text.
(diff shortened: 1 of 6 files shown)


robbie-c:
Closing this in favour of using an autocapture approach

PR #4631, opened 2026-08-24
chore(react-native): require @posthog/react-native-plugin 2.4.3

turnipdabeets:
Draft — **do not merge until `@posthog/react-native-plugin@2.4.3` is published**, since the floor points at a version that does not exist yet.

Follow-up to #4630. That PR bumps the plugin's `com.posthog:posthog-android` pin to 3.60.7 for the manual session replay fix, and will release the plugin as 2.4.3.

`posthog-react-native` declares the plugin as a peer with a hand-maintained floor (`packages/react-native/package.json`), currently `>= 2.4.2`. Nothing cascades it — it is a literal range, not `workspace:`, so changesets will not raise it. Left alone, an app can stay on 2.4.2 and never pick up the Android fix.

Raises the floor to `>= 2.4.3`.

Spotted by @ioannisj.

### DIFF
--- a/.changeset/rn-plugin-floor-2-4-3.md
+++ b/.changeset/rn-plugin-floor-2-4-3.md
@@ -1,0 +1,5 @@
+---
+'posthog-react-native': patch
+---
+
+Fix session replay started with `startRecording()` capturing nothing on Android by requiring `@posthog/react-native-plugin` 2.4.3 or newer.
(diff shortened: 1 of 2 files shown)


turnipdabeets:
Intentional — the changeset is right, and this warning can't be satisfied for a PR of this shape.

`check-changeset-coverage.mjs` skips `/package.json` when mapping changed files to affected packages (`ignoreSuffixes`, line 36), and a peer-range bump lives entirely in `package.json`. So any dependency-floor PR declares a package the check can't see a change in.

`posthog-react-native`'s published `peerDependencies` range is what's changing here, so it does need a release for users to pick up the new floor — hence the patch changeset.

ioannisj (approved):

PR #4633, opened 2026-08-24
feat(mcp): opt-in captureModel — agent-self-reported model id on MCP events

DanielVisca:
## Problem

MCP server owners want to know which LLM is driving their server so they can tell whether tool-call quality degrades on specific models (raised by a customer via the MCP analytics team). The MCP wire deliberately carries no model identity — we validated with a raw-logging MCP server that traffic from the same client is byte-identical across model and effort settings, and no spec revision (through 2026-07-28) changes that. The one place the information exists is the agent itself: harnesses inject the model id into the agent's system prompt, and in testing agents restated it exactly (5/5 exact matches, including full dated model ids) via an injected tool parameter.

## Changes

Adds an opt-in `captureModel` option to `@posthog/mcp`, reusing the `context` → `$mcp_intent` injection machinery:

- Injects a required `llm_model` string parameter into every tool at `tools/list`; strips it before the handler runs.
- Captures it as `$mcp_llm_model` with `$mcp_llm_model_source = "self_reported"` — honestly labeled as unverified, matching how the MCP spec treats `clientInfo`.
- Ownership-safe: a tool that already declares `llm_model` keeps it — never stolen, never captured.
- An honest `"unknown"` from the agent is dropped rather than recorded; value is truncated like other metadata fields.
- Reasoning effort is deliberately **not** collected: it never crosses the wire and agents can't reliably self-report it (validated — the guess was wrong under a forced effort setting), so it would only be noise.

### Libraries affected

- [x] @posthog/mcp

## Checklist

- [x] Tests for new code

### If releasing new changes

- [x] Changeset file included (`.changeset/model-capture-option.md`)

## 🤖 Agent context

- Authored with Claude (PostHog Desktop task). TDD: tests written and confirmed red before implementation; all 665 unit tests, eslint, and tsc pass.
- Validated the design empirically before writing code: a raw-logging HTTP MCP server + real Claude Code client confirmed no model/effort on the wire, and that an injected self-report parameter returns exact model ids; production `$mcp_intent` fill rates across major harnesses support cross-harness viability of injected parameters.
- Chose opt-in (unlike `context`, which is on by default) because the parameter adds tokens to every tool call and the value is self-reported.

---
*Created with [PostHog Desktop](https://posthog.com/desktop?ref=pr)*


marandaneto (commented):

packages/mcp/src/extensions/instrumentation.ts
```diff
@@ -646,6 +661,9 @@ async function getTracedToolsList(
     if (data && isContextEnabled(data.options.context)) {
       tools = addContextParameterToTools(tools, getContextDescription(data.options.context), data.logger)
     }
+    if (data && isCaptureModelEnabled(data.options.captureModel)) {
+      tools = addModelParameterToTools(tools, getModelDescription(data.options.captureModel), data.logger)
```
marandaneto: blocking: Inject the model parameter into the virtual tool — model injection runs before `reportMissing` appends `get_more_tools`, so with both options enabled the advertised virtual tool lacks `llm_model` and its missing-capability calls cannot self-report the model, contrary to the documented every-tool behavior.

packages/mcp/src/extensions/model-parameters.ts
```diff
@@ -0,0 +1,144 @@
...
+  return typeof captureModel === 'object' && captureModel !== null ? captureModel.description : undefined
+}
+
+/**
+ * Adds an `llm_model` parameter to a tool's JSON Schema. Called AFTER the MCP
+ * SDK has converted Zod schemas to JSON Schema, so only JSON Schema needs
+ * handling. Skip rules match `addContextParameterToTool`: a tool that already
+ * declares `llm_model` owns it, and complex schemas can't safely gain keys.
+ */
+export function addModelParameterToTool<TTool extends ModelInjectableTool>(
```
marandaneto, line 48: suggestion: Reuse the existing parameter-injection implementation — this function and its batch wrapper duplicate almost all of `addContextParameterToTool`; a shared analytics-parameter helper would keep schema ownership, cloning, and required-field behavior from drifting.

DanielVisca:
Both review comments and both ReviewHog findings were valid. Fixed in a960d92.

**Virtual tool missing `llm_model`** (blocking). Model injection ran before `reportMissing` appended `get_more_tools`, so the virtual tool's advertised schema never carried the parameter. Ownership was still cached for it, meaning the SDK would strip and capture an `llm_model` argument it had never asked for, and missing capability reports were the one call type structurally unable to self report a model. Injection now runs after the append, mirroring how `conversation_id` already behaved. Ownership is still resolved from the un-injected descriptor, and `context` stays excluded because the virtual tool declares its own and that argument is the report itself.

**Duplicated injection logic** (suggestion). `addModelParameterToTool` was a near verbatim copy of `addContextParameterToTool`. Both now delegate to a shared `addAnalyticsParameterToTool` in `analytics-parameters.ts`, which already owned the ownership predicates, so schema cloning, `additionalProperties` handling, and required field behaviour cannot drift between the two parameters. Warning strings are unchanged.

**Undocumented ownership limit.** Worth flagging that `captureModel` degrades differently from `context`: context still records when ownership is unresolved, this captures nothing at all. I confirmed the high level path is unaffected, since `instrument-highlevel.ts` builds ownership per request from `_registeredTools`, so the limit is specific to a low level `Server` on fresh instance per request hosts. Documented in the README beside the existing `$mcp_intent` section, and the changeset no longer promises unconditional stripping and capture.

**Stale reference docs.** Added the `$mcp_tool_call` extras, the two property rows for `$mcp_llm_model` and `$mcp_llm_model_source`, and the `captureModel` options row in `docs/ARCHITECTURE.md`, following the existing `ConversationId` / `enableConversationId` precedent.

On verification: the package suites were initially failing to load in my environment because the workspace deps were unbuilt, so I took a stashed baseline first and confirmed the failures were identical with and without my changes before fixing the environment. Full suite is 46 suites and 667 tests passing, lint clean. I mutation tested the new regression test rather than trusting it: reverting only the reorder fails the schema assertion, restoring it passes. The second new test covers the missing capability capture path but passes either way, since capture keys off cached ownership rather than the advertised schema, so the schema assertion is the actual guard.

One deliberate omission: both markdown files were already non conforming to Prettier at HEAD and the package lint script only covers `src` and `harness`, so I left them alone rather than bury these fixes in an unrelated reformat.

packages/mcp/src/extensions/instrument-highlevel.ts
```diff
@@ -189,6 +190,7 @@ function addTracingToToolCallbackInternal(
     const cleanedArgs = stripOwnedAnalyticsArguments(args, {
       context: isContextEnabled(options?.context) && analyticsOwnsParameter(inputSchema, 'context'),
       conversationId: options?.enableConversationId === true && analyticsOwnsParameter(inputSchema, 'conversation_id'),
+      llmModel: isCaptureModelEnabled(options?.captureModel) && analyticsOwnsParameter(inputSchema, 'llm_model'),
```
dustinbyrne, line 194: **Question:** Does `instrument(McpServer)` support applications that create a new high-level server instance for each request?

The README appears to say that it does. Conceptually, an application can use this pattern:

```ts
async function handleRequest(request) {
  const server = createAndInstrumentServer()
  return handleWithServer(server, request)
}
```

A `tools/list` request creates Instance A. The later `get_more_tools` call creates Instance B. The two instances do not share the ownership map.

The virtual-tool branch returns before the registry-based `parameterOwnership` logic. Also, `isToolAdvertised()` does not fill the ownership cache. This appears to make `ownership.llmModel` false on Instance B. The SDK would then neither capture nor remove `llm_model`.

If this lifecycle is supported, should this branch pass ownership from the virtual-tool descriptor? A test with separate list and call instances could verify the behavior.

This was the only potential gap found during our agent-assisted human review.

dustinbyrne (commented):
Agent-led review, human-reviewed before posting.

packages/mcp/src/extensions/instrument-highlevel.ts
DanielVisca: Confirmed, and reproduced with the test you suggested: separate advertising and calling instances, `$mcp_llm_model` came back undefined. Fixed in 6c10c1b.

That branch only runs when `isToolAdvertised()` returns false, so the descriptor is the SDK's own `getReportMissingToolDescriptor()`. Ownership is knowable statically there and we simply weren't consulting it. The fail-closed rule for registered tools exists because an unresolved `llm_model` may be the application's own argument, which cannot be the case when the application has no tool by that name. So the fix resolves ownership from the descriptor rather than relaxing the policy.

The low-level adapter had the same gap, so both now go through one shared helper. Each adapter has a test with separate list and call instances, and I checked they fail without the fix.

Scoped to `llm_model` on purpose. `conversation_id` still resolves from the cache: resolving it statically starts minting handles and appending prompt-back blocks on instances that mint none today, which is a session anchoring change (ADR-0004) rather than this gap, and it broke the existing fresh-instance test. Worth its own PR. `context` is already forced off for this event type and `outputInstructions` reads the cache directly, so `llm_model` is the only value that moves.

Also corrected the README line that prompted the question. It credited the live registry, which never covered the virtual tool.

lucasheriques (approved):

PR #4640, opened 2026-08-25
feat: capture JSON-LD as replay custom events

robbie-c:
## Summary

- Add opt-in JSON-LD capture with `session_recording.captureJsonLd`.
- Emit sanitized data as rrweb custom events with the `$json_ld` tag.
- Keep the payload as a JSON-LD object, root array, or root `@graph`. DOM position is not part of the event.
- Keep all script nodes out of replay snapshots when the option is enabled.
- Load the parser and observer only with the lazy session recorder.

This is an alternative to #4625. The option is off by default.

## Whitelist

The whitelist accepts the 124 entity types that Google names in its current Search structured-data feature guides and examples. It also accepts 43 common Schema.org types outside that list. These additions cover common pages, actions, services, catalogs, organizations, local businesses, events, media, and product subtypes. Examples include `WebSite`, `FAQPage`, `SearchAction`, `Service`, `OfferCatalog`, `School`, `CafeOrCoffeeShop`, `MusicEvent`, `PodcastEpisode`, and `ProductModel`.

The sanitizer accepts an `@type` array or a full `https://schema.org/Type` value. It removes unknown type values and drops an entity if no allowed type remains. A root `@graph` can contain supported and unsupported entities. The sanitizer keeps only rebuilt supported entities and drops the complete graph if none remain.

Every captured entity can contain `@id`. Types use these property rules:

- `Action` and named action subtypes: `actionStatus`
- `AggregateOffer`: price range, currency, count, availability, and nested offers
- `AggregateRating`, `EmployerAggregateRating`, and `Rating`: rating values and counts
- `Brand`: `name`
- `BreadcrumbList` and `ItemList`: list metadata and rebuilt `ListItem` values
- `CreativeWork` and named subtypes: language, format, date, accessibility, education, rating, publisher, and potential action fields
- `Event` and named subtypes: date, status, attendance, capacity, accessibility, rating, and offer fields
- `ListItem`: position and one rebuilt supported entity
- `Offer`: price, currency, validity, availability, condition, and seller fields
- `OfferCatalog`: name and rebuilt supported entities
- `Organization` and named subtypes: name, date, nonprofit, rating, and brand fields
- `Place` and named subtypes: access, capacity, accessibility, and rating fields
- `Product` and named subtypes: identifiers, model, category, physical attributes, dates, brand, manufacturer, offers, and rating fields
- `Service`: name, type, category, provider, service area, offers, and rating fields
- `Person` and PII-heavy support types such as `ContactPoint` and `PostalAddress`: no properties other than `@id`
- Other named Google support types: no properties other than `@type` and `@id`

Leaf values must be a string, number, boolean, null, or an array that contains only those values. Nested values must match an allowed entity type.

## Privacy and lifecycle

- Parse the source string, build a new object from own allowlisted properties, and serialize only the new object.
- Drop invalid JSON, invalid Schema.org contexts, unsupported entity types, oversized blobs, and values inside text masks or blocked elements.
- Do not place JSON-LD in the recorder retry queue. Capture starts only after rrweb is ready and the current URL block state is known.
- Suppress JSON-LD seen on a blocked URL. It stays suppressed after navigation until the source changes on an allowed URL.
- Restore current JSON-LD after idle recovery and after a trigger-pending full snapshot truncates the buffer.
- Limit one input to 100 KB, one sanitized output to 20 KB, and total accepted JSON-LD to 100 KB per recorder lifetime.

`@id` is kept without changes by design. We need a separate follow-up PR to define and add `@id` scrubbing before we enable this option for customers.

## Incident review

- The change does not add a new core-to-recorder method, persisted field, or cross-bundle signature.
- The change does not modify session rotation, flush, or buffer ownership.
- The privacy checks run before the event is accepted. The browser test also checks the raw recording bytes for private markers.

## Bundle size

- Main `array.js`: `+0.00%` minified, `-0.01%` gzip, `+0.04%` Brotli against `origin/main`.
- Production `posthog-recorder.js`: `+8,718` minified bytes and `+3,470` gzip bytes.
- The 124-type expansion accounts for `+2,263` minified bytes and `+1,027` gzip bytes of the recorder total.
- The 43 common types, nested list paths, and root `@graph` support add another `+1,017` minified bytes and `+453` gzip bytes.

## Tests

- 405 focused browser unit tests, including the recorder snapshot and the complete type tables.
- Browser TypeScript check.
- Focused ESLint and Prettier checks.
- PostHog config snapshot test.
- Three focused Chromium tests, including the raw replay-byte privacy test for a root graph.
- Compatibility probe with the published `posthog-js@1.418.16` core and this recorder bundle.

Firefox and WebKit were not available in the local Playwright installation. CI will run the configured browser coverage.


packages/browser/src/extensions/replay/external/json-ld.ts
```diff
@@ -0,0 +1,436 @@
...
+    ['AggregateRating EmployerAggregateRating Rating'.split(' '), EMPTY_ENTITY_RULES],
+]
+const TYPES_WITHOUT_PROPERTIES =
+    'AlignmentObject BedDetails Certification ContactPoint CreditCard DefinedRegion EducationalOccupationalCredential EntryPoint GeoCoordinates GeoShape InteractionCounter JobPosting LocationFeatureSpecification MathSolver MemberProgram MemberProgramTier MerchantReturnPolicy MerchantReturnPolicySeasonalOverride MonetaryAmount NutritionInformation OccupationalExperienceRequirements OfferShippingDetails OpeningHoursSpecification PeopleAudience PostalAddress PriceSpecification PropertyValue QuantitativeValue ServicePeriod ShippingConditions ShippingDeliveryTime ShippingRateSettings ShippingService SpeakableSpecification Thing UnitPriceSpecification'.split(
+        ' '
+    )
+
+export const JSON_LD_EVENT_TAG = '$json_ld'
+
+function isObject(value: unknown): value is Record<string, unknown> {
```
marandaneto: nit: Reuse the shared object type guard — this duplicates `isObject` already exported by `@posthog/core`, while subtly accepting arrays as records. Importing the shared implementation removes this one-off abstraction and uses the repository's established semantics.

packages/types/src/posthog-config.ts
```diff
@@ -686,6 +686,18 @@ export interface SessionRecordingOptions {
      */
     slimDOMOptions?: true | Partial<SlimDOMOptions> | 'all'
 
+    /**
+     * Captures sanitized Schema.org JSON-LD as session replay custom events.
+     * JSON-LD inside a text mask or blocked element is never captured.
+     * The recorder keeps `@id` values without changes.
+     * The event tag is `$json_ld`. The payload is a JSON-LD object or array.
+     * The recorder removes all script nodes from snapshots when this option is enabled.
+     * Supported types are Action, AggregateOffer, AggregateRating, Brand, CreativeWork, Offer, Organization, Person, Place, and Product.
```
marandaneto: suggestion: Document the complete capture surface — this lists only ten supported types, but the sanitizer also captures `Event`, `ItemList`, `OfferCatalog`, `Service`, and over a hundred named or inherited types. Because this is a privacy-sensitive capture contract, please describe all categories or link to an authoritative allowlist.

marandaneto (commented):

marandaneto (approved):

PR #1885, opened 2025-04-09
fix: new survey position for feedback button

lucasheriques:
## Changes

necessary to fix [this issue](https://github.com/PostHog/posthog/issues/29773). Will also require a change on the main app.

Verified locally positions are now working as expected:

image

PR #4641, opened 2026-08-25
fix(surveys): respect NextToTrigger position for tab widget surveys

### LINKED ISSUE #31267: Embedded tab feedback button survey does not respect the position
### Bug Description

Very similar to https://github.com/PostHog/posthog/issues/29773

The fix will be closely the same - use the new "next to button/tab" position, and only make the survey next to the embedded tab if that's the selected value.

Here's the current state:

Image

(Left is selected)

Image

But the survey is still shown next to the button.

[More internal context here on Slack](https://posthog.slack.com/archives/C07QD3LT8U9/p1740996931280049).

### Debug info

```shell
```

mayuriphad:
## Problem

Fixes PostHog/posthog#31267.

Similar to #1885 / posthog/posthog#31034 (which fixed this for the *selector* widget type), the **tab** widget type ignores the survey's configured position when it's set to "next to trigger" - the popup always renders vertically centered next to the tab, regardless of the position setting.

## Root cause

`FeedbackWidget`'s mount effect unconditionally sets:

\`\`\`ts
if (survey.appearance?.widgetType === 'tab') {
    setStyleOverrides({ top: '50%', bottom: 'auto' })
}
\`\`\`

This `styleOverrides` object is spread *last* onto the popup's `style` in `SurveyPopup` (after `getPopoverPosition(...)`), so it always wins - even when `survey.appearance?.position === SurveyPosition.NextToTrigger`.

The selector widget type doesn't have this problem: its click listener already checks for `NextToTrigger` and computes a position from the click target's bounding rect via `getNextToTriggerPosition`, dispatched through the `ph:show_survey_widget` custom event.

## Fix

Mirror that behavior for the tab widget:
- Added a ref to the tab `<button>`.
- In `toggleSurvey`, if `widgetType === 'tab'` and `position === NextToTrigger`, compute the position relative to the tab button via the existing `getNextToTriggerPosition` helper.
- The mount effect's hardcoded `{ top: '50%', bottom: 'auto' }` now only applies when position is *not* `NextToTrigger`, so it no longer clobbers the computed position.

## Test plan

- Added `positions survey next to the tab button when position is NextToTrigger` to `feedback-widget.test.tsx`, asserting the popup gets `position: fixed` with a computed (non-`50%`) `top`, using a mocked `getBoundingClientRect` on the tab button.
- **Note:** I wasn't able to get a full `pnpm build` + `jest` run working end-to-end in my environment (workspace package build chain / husky prepare script friction unrelated to this change), so I wasn't able to execute the suite locally. I traced the logic carefully by hand against the existing selector-widget code path it mirrors. Happy to fix up anything CI flags.

### DIFF
--- a/packages/browser/src/__tests__/extensions/surveys/feedback-widget.test.tsx
+++ b/packages/browser/src/__tests__/extensions/surveys/feedback-widget.test.tsx
@@ -1,7 +1,7 @@
 import '@testing-library/jest-dom'
 import { act, cleanup, fireEvent, render, screen, waitFor } from '@testing-library/preact'
 import { FeedbackWidget } from '../../../extensions/surveys'
-import { Survey, SurveyQuestionType, SurveyType, SurveyWidgetType } from '../../../posthog-surveys-types'
+import { Survey, SurveyPosition, SurveyQuestionType, SurveyType, SurveyWidgetType } from '../../../posthog-surveys-types'
 import { createMockPostHog } from '../../helpers/posthog-instance'
 import { PostHogFeatureFlags } from '../../../posthog-featureflags'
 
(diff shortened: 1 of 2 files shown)


mayuriphad:
Closing — this PR was opened in error by an unattended automation run, without authorization. Apologies for the noise.

PR #4643, opened 2026-08-25
feat(react-native): autocapture touches and clicks on React Native Web

### LINKED ISSUE #2167: `expo-router` touch events support
### Description

https://docs.expo.dev/router/installation/#quick-start
https://github.com/expo/expo/tree/main/packages/expo-router
right now we support `@react-navigation/native` and `react-native-navigation` but not yet `expo-router` which is getting more popular day by day
https://results.stateofreactnative.com/navigation/
Screenshot 2024-02-01 at 14 50 52

ioannisj:
## Problem

Touch autocapture produced no `$autocapture` events on React Native Web, including expo-router apps running on web. `autocaptureFromTouchEvent` returned early whenever `e._targetInst` was missing, and react-dom's synthetic events never set it, so every touch was dropped silently. The fiber was reachable the whole time, just under a randomised `__reactFiber$*` key on `e.target` rather than on the event.

Native was never affected. Both `ReactFabric-prod.js` and `ReactNativeRenderer-prod.js` set `_targetInst` in the `SyntheticEvent` constructor and their own dispatch reads it back, so old architecture and Fabric already worked.

Browsers also only fire `touchend` for touch input, so mouse and trackpad users on web produced nothing at all, even with the fallback in place.

Closes #2167.

## Changes

Falls back to walking up from `e.target` to the nearest node carrying a React fiber key (`__reactFiber$*`, or the legacy `__reactInternalInstance$*`) when `_targetInst` is absent. The element walk itself is untouched, since the fiber shape is identical either way.

Wraps the capture path in try/catch so a throw can no longer escape into the host app's event dispatch, and warns once rather than on every interaction.

Registers a capture-phase `click` listener on `document` on web when `captureTouches` is on, emitted with `$event_type: 'click'`. Both the capture phase and the document level are needed: react-native-web's `Pressable` calls `stopPropagation`, and its `Modal` portals content outside the provider's subtree. `onTouchEndCapture` is no longer wired on web, so a tap that fires both `touchend` and a synthesised `click` is still captured once.

Skips react-native-web's own `LocaleProvider` wrapper only where RNW puts it, at the head of the chain on the fallback path, so an app component of the same name keeps its place in `elements`.

Tests cover both paths against the same fixture: the `_targetInst` path, the fiber fallback, the legacy key, the ancestor walk, a target whose property access throws, `ph-no-capture` through the fallback, and the click listener's phase, portal, re-render and unmount behaviour.

## Release info

Sub-libraries affected

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code
- [x] Accounted for the impact of any changes across the different platforms
- [x] Took care not to unnecessarily increase bundle size

### If releasing new changes

## 🤖 Agent context

**DRI:** @ioannisj


packages/react-native/src/PostHogProvider.tsx
```diff
@@ -173,11 +174,38 @@ export const PostHogProvider = ({
...
+  useEffect(() => {
+    // The package targets ESNext without the DOM lib, so reach the document off the global.
+    const doc = (globalThis as any)?.document
+    if (!isWeb() || !captureTouches || !doc?.addEventListener) {
+      return
+    }
+    const handler = (e: any): void => {
+      autocaptureFromTouchEvent({ target: e.target, nativeEvent: e }, posthog, optionsRef.current, 'click')
+    }
+    doc.addEventListener('click', handler, true)
```
turnipdabeets: [question] Two sibling providers sharing one client each register their own listener, so one click in one subtree enqueues two `$autocapture` events (measured in jsdom: expected 1, got 2 — native gives 1, since `onTouchEndCapture` is subtree-scoped). The autocapture spec has a scenario for exactly this — "Repeated setup does not install duplicate autocapture observers", *exactly one event should be enqueued for that interaction* ([openspec/specs/autocapture/spec.md](https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/autocapture/spec.md)). Worth refcounting the listener per client, or is one-provider-per-app the assumption we're happy with?

packages/react-native/src/PostHogProvider.tsx
```diff
@@ -173,11 +174,38 @@ export const PostHogProvider = ({
...
+    optionsRef.current = autocaptureOptions
+  }, [autocaptureOptions])
+
+  useEffect(() => {
+    // The package targets ESNext without the DOM lib, so reach the document off the global.
+    const doc = (globalThis as any)?.document
+    if (!isWeb() || !captureTouches || !doc?.addEventListener) {
+      return
+    }
+    const handler = (e: any): void => {
```
turnipdabeets: [question] Document scope means clicks *outside* the provider's subtree get captured too — a sibling button under a named app component comes back as `elements: ["OutsideAppComponent"], $event_type: "click"`, where native captures nothing. Probably the right trade for Modal portals, but posthog.com currently says "touch events for children of `PostHogProvider` are tracked" — can we update that line in the same batch? (Portals work either way; React keeps the fiber `.return` chain intact through them.)

packages/react-native/src/autocapture.tsx
```diff
@@ -61,7 +61,58 @@ export const defaultPostHogLabelProp = 'ph-label'
 
 const captureAttributePrefix = 'data-ph-capture-attribute-'
 
-export const autocaptureFromTouchEvent = (e: any, posthog: PostHog, options: PostHogAutocaptureOptions = {}): void => {
+// react-native-web internals; skipped only where RNW puts them, so a same-named app component is kept.
+const frameworkInternalLabels = ['LocaleProvider']
```
turnipdabeets, line 69: [nit] Could we note the RNW version this was verified against? I checked 0.20.0 — `createElement/index.js` only wraps in `LocaleProvider` when `domProps.dir` is set, so a nested `<Text>` inside a text ancestor never gets one. That conditional is easy to lose track of on the next RNW bump.

packages/react-native/src/types.ts
```diff
@@ -14,7 +14,10 @@ export type PostHogNavigationRef = {
 
 export type PostHogAutocaptureOptions = {
   /**
-   * Enable autocapture of touch events
+   * Enable autocapture of touch events.
+   *
+   * On React Native Web this also captures `click` events from mouse and trackpad input,
+   * emitted with `$event_type: 'click'`, since browsers fire `touchend` only for touch input.
```
turnipdabeets: [suggestion] `click` also fires for keyboard activation and programmatic `.click()`, not just pointers — worth saying so, since those produce autocapture events with no touch behind them.

```suggestion
   * On React Native Web this also captures `click` events — mouse, trackpad, keyboard
   * activation and programmatic clicks — emitted with `$event_type: 'click'`, since
   * browsers fire `touchend` only for touch input.
```

packages/react-native/src/autocapture.tsx
```diff
@@ -61,7 +61,58 @@ export const defaultPostHogLabelProp = 'ph-label'
...
+  posthog: PostHog,
+  options: PostHogAutocaptureOptions = {},
+  eventType: 'touch' | 'click' = 'touch'
+): void => {
+  try {
+    captureFromEvent(e, posthog, options, eventType)
+  } catch (error) {
+    if (!warnedCaptureFailure) {
+      warnedCaptureFailure = true
+      console.warn('PostHog autocapture: capturing the interaction threw:', error)
```
turnipdabeets, line 129: [nit] Not debug-gated, so this fires once in production for everyone. Fine given warn-once — just noting a systematic failure then goes fully silent, which is the case the comment above is worried about.

turnipdabeets (commented):
One thing with no line to hang off: #2167's thread mixes web and native asks — AdamDorwart's repro was `example_expo_rn_web`, but a few commenters are on native expo-router. Worth confirming native touch capture works there before `Closes #2167` retires the whole issue.

ioannisj:
> One thing with no line to hang off: #2167's thread mixes web and native asks — AdamDorwart's repro was `example_expo_rn_web`, but a few commenters are on native expo-router. Worth confirming native touch capture works there before `Closes #2167` retires the whole issue.

Native was never broken, and I've confirmed it (Android but iOS should be the same). Ran `examples/example-expo-57` (expo-router, `newArchEnabled=true`, so Fabric) on an Android emulator against this branch's build, and a tap gives `$autocapture` with `$event_type: "touch"` and a full 20-entry `$elements` chain. The chain only fills in when `_targetInst` is present, so native takes the original path and never touches the new fallback.

packages/react-native/src/PostHogProvider.tsx
ioannisj: Good catch on the spec, I had this filed as a preference rather than a conformance gap.

Fixed in 8797151bc. The document listener is now refcounted per client in a module-level WeakMap, so however many providers mount against one client exactly one listener is installed, and a single interaction enqueues one $autocapture. Two tests came with it: two sibling providers sharing a client produce one event, and the listener only detaches when the last provider unmounts. Both fail against the previous commit, so they're real regression guards rather than passing by accident.

Worth being explicit that this fixes the duplication and not the scope. Clicks outside the provider's subtree are still captured, and with two different clients each still gets its own document listener, so the cross-client misrouting greptile raised is untouched. Picking that up separately since it needs a fiber ancestry check rather than a counter.

packages/react-native/src/autocapture.tsx
ioannisj: Noted in d12d76800. I checked 0.21.2 as well and the condition is unchanged, createElement only wraps in LocaleProvider when domProps.dir is set, so the comment now records both versions and says to recheck on a bump.

packages/react-native/src/types.ts
ioannisj: Applied verbatim in d12d76800. You're right that keyboard activation and programmatic .click() both land here with no pointer behind them, and the old wording denied it.

packages/react-native/src/autocapture.tsx
ioannisj: Leaving this one as is. Bare console.warn is already the pattern across the package (storage.ts:92, 113, 128, 161 and PostHogProvider.tsx:137), so gating just this call site would make it the odd one out.

Routing it through posthog.logger isn't a free swap either. That ships a record to the logs product over the network on every failure, which is heavier than a console line and uncomfortably close to a loop when the thing failing is capture itself.

Agreed on the substance though, warn-once means a systematic failure surfaces once and then goes quiet for the life of the bundle. Worth revisiting if we ever grow an internal SDK-error channel that isn't the logs product.

dustinbyrne:
## Pull request risk assessment: CLEAR

I checked commit `d12d7680095d11258a13a4519d3c58da0a6366eb` against `main`.

| Check | Result | Evidence |
| --- | --- | --- |
| The pull request fixes a real issue. | CLEAR | The base Expo web sandbox emitted no `$autocapture` event for the affected click. The same action on this commit emitted one `$autocapture` event with `$event_type: "click"` and an element chain. The fallback in `packages/react-native/src/autocapture.tsx` handles the missing `_targetInst`. |
| All non-bot authors are internal contributors. | CLEAR | The pull request author and all eight commit authors are `ioannisj`. There are no co-authors. GitHub confirms PostHog organization membership. |
| The production change uses only private or internal logic. | CLEAR | The implementation is behind the existing provider and autocapture option. The changed autocapture helper is not part of a package entry point. |
| The public API does not change. | CLEAR | Package exports, provider props, option types, and signatures are unchanged. `captureTouches` remains an optional boolean. The public API check passed. |
| Default data capture does not change. | CLEAR | `captureTouches` remains disabled by default. Identical base and head sandboxes without an `autocapture` prop emitted only the manual test event and no `$autocapture` event. |
| The example application works without an observed regression. | CLEAR | An isolated Expo 57 web application used React Native 0.86.2, React Native Web 0.21.2, and Chromium. The affected click and a normal manual capture succeeded. Two sibling providers that shared one client emitted one event per click. No browser error occurred. |
| The changeset has the correct scope and release type. | CLEAR | `.changeset/brave-moons-shake.md` names only `posthog-react-native`, uses `minor` for the compatible feature, and describes this change. |
| The change is not a major breaking change. | CLEAR | Supported entry points and configuration shapes remain compatible. No consumer migration is required. |

All checks are clear. This pull request meets the automated low-risk rules.

### Review points

None.

dustinbyrne (approved):
Risk assessment: CLEAR.

Details: https://github.com/PostHog/posthog-js/pull/4643#issuecomment-5414837640

packages/react-native/src/PostHogProvider.tsx
ioannisj: Scoped in 16b0fca17. The listener stays on the document so Modal still works, but capture now walks the fiber `.return` chain and drops anything whose ancestry doesn't pass through a mounted provider's root node. SO `OutsideAppComponent` case captures nothing now, and a click in one provider's subtree no longer reaches a second provider's client. This was verified in the browser.

Tbh I did weigh leaving it to a follow-up, multiple roots with distinct clients is rare I would think, but once the wide behaviour ships people depend on it and narrowing later becomes the breaking change

packages/react-native/src/autocapture.tsx
```diff
@@ -70,13 +123,15 @@ export const autocaptureFromTouchEvent = (e: any, posthog: PostHog, options: Pos
     propsToCapture = ['style', 'testID', 'accessibilityLabel', customLabelProp, 'children'],
   } = options
 
-  if (!e._targetInst) {
+  const nativeInst = e._targetInst
+  const targetInst: Element | undefined = nativeInst || getFallbackTargetInstance(e)
```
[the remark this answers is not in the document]
ioannisj: Confirmed, and it isn't web-only. The `ph-no-capture` check at autocapture.tsx:170 sits inside the walk that stops at `elements.length < maxElementsCaptured` (line 161), so the same truncation drops the opt-out on the native touch path too. This PR widens the exposure rather than introducing it.

Splitting it into its own PR since it's a native bug as much as a web one. The fix is the same shape as the ownership walk added here, scan the full `.return` chain for the opt-out before the truncating element walk, so the cap governs what gets emitted and never what gets decided.

Issue #4637, opened 2026-08-25
Declare React peer dependencies for posthog-js/react


tifandotme:
## Bug description

The published `posthog-js/react` entry point requires React at runtime, but the package does not declare React or React DOM peer dependencies.

## How to reproduce

1. Install `posthog-js@1.379.0`, `react@19.2.7`, and `react-dom@19.2.7`.
2. Configure Bun with:

   ```toml
   [install]
   linker = "isolated"
   globalStore = true
   ```

3. Run `bun install`.
4. Import `posthog-js/react` from a test or application module.

The import fails with:

```text
Cannot find module 'react'
```

The stack points to:

```text
posthog-js/react/dist/umd/index.js
```

Please declare the React requirements for this entry point, without making React mandatory for users who do not use the React integration.

PR #4645, opened 2026-08-25
fix(react): drop JSX namespace import from public type declarations

turnipdabeets:
## Problem

Follow-up to #4638, which is now merged.

Reviewing that PR, @marandaneto asked how we know `>=16.8.0` is the minimum React version we support. The `react` floor checked out — the built bundle only touches `createContext`, `useRef`, `useMemo`, `useEffect`, `useContext`, `useState`, `useCallback`, `Children`, `React.Component`, `React.createElement`, `React.Fragment`, `React.isValidElement`, and hooks are the binding constraint at exactly 16.8.0.

The **`@types/react`** floor did not. Our published declarations do:

```ts
import React__default, { JSX, FunctionComponent, ErrorInfo } from 'react'
```

`JSX` only became an exported member of the `react` types module in `@types/react@18.2.6`. Bisected against the published tarball:

```
@types/react@18.2.5 -> error TS2305: Module '"react"' has no exported member 'JSX'.
@types/react@18.2.6 -> OK
```

So anyone on older React types checking our `.d.ts` with `skipLibCheck: false` hit `TS2305`, while both packages advertise `>=16.8.0`. Raising the declared range to match is not an option — it makes `npm install` fail with `ERESOLVE` for React 17 projects. The fix is to make the low number true.

## Changes

Return `ReactElement<any, any>` instead of `JSX.Element` from the six components that annotated it, and give the two `PostHogProvider`s explicit return types so TypeScript stops *inferring* `React.JSX.Element` (`React.JSX` has the same 18.2.6 floor).

`ReactElement<any, any>` is the definition of `JSX.Element`, so the public types are unchanged. It is also available from `@types/react@16.8.0`, so **no peer range changes** — `packages/browser/package.json`, `packages/react/package.json`, and `pnpm-lock.yaml` are untouched.

## Validation

Rebuilt the package, confirmed no `JSX` remains in any emitted `.d.ts`, then swapped the rebuilt declarations into the published `posthog-js@1.379.0` tarball and typechecked a consumer under the layout from #4637 (`linker = "isolated"`, `globalStore = true`, `skipLibCheck: false`):

| `@types/react` | before | after |
| --- | --- | --- |
| 16.8.0 | `TS2305` | ✅ clean |
| 16.9.0 | `TS2305` | ✅ clean |
| 17.0.2 | `TS2305` | ✅ clean |
| 18.3.1 | ✅ clean | ✅ clean |
| 19.2.7 | ✅ clean | ✅ clean |

The declared `>=16.8.0` floor now holds exactly.

Backwards compatibility was checked separately, since a return type appears in the public API. Six consumer patterns — JSX usage, direct call assigned to `JSX.Element`, `ReturnType` round-trip in both directions, a `React.ReactNode` slot, and a `React.FC` slot — compile identically against the old and new declarations.

`@types/react` 18.0.0–18.2.5 report `TS2307: Cannot find module 'scheduler/tracing'`, which comes from `@types/react`'s own `index.d.ts` (the known `@types/react` ↔ `@types/scheduler` drift), reproduces on `main`, and is unaffected by this change.

Also: 105 unit tests pass, eslint and prettier clean, `pnpm install --frozen-lockfile` passes.

Thanks @dustinbyrne for catching that bare `ReactElement` is not the same type as `JSX.Element` and that the explicit form needs no version bump — the first revision of this PR used bare `ReactElement` and raised `@types/react` to `>=16.9.0`, which was an artifact of that choice rather than a real constraint.

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [x] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

`ReactElement<any, any>` is the definition of `JSX.Element`, so the public types are structurally identical — this widens the range of `@types/react` that can consume them rather than narrowing it, and no peer range moves. Type annotations are erased at compile time, so there is no runtime or bundle size impact.

### If releasing new changes

## 🤖 Agent context

Follow-up split out of the #4638 review rather than widening that packaging PR. The `@types/react` floors were verified by bisecting real `@types/react` installs against the published tarball, not read from release notes.


packages/react/src/components/internal/VisibilityAndClickTracker.tsx
```diff
@@ -20,7 +20,7 @@ export function VisibilityAndClickTracker({
     onClick?: MouseEventHandler<HTMLDivElement>
     trackView: boolean
     options?: IntersectionObserverInit
-}): JSX.Element {
+}): ReactElement {
```
dustinbyrne: > `JSX.Element` is declared as `ReactElement<any, any>`, so this is the same type structurally — no consumer-visible behavior change.

`JSX.Element` is the same type as `ReactElement<any, any>`, but not necessarily `ReactElement`. this is potentially a return type change, are we sure it's backwards compatible?

packages/react/package.json
```diff
@@ -34,7 +34,7 @@
         "slim"
     ],
     "peerDependencies": {
-        "@types/react": ">=16.8.0",
+        "@types/react": ">=16.9.0",
         "posthog-js": ">=1.257.2",
         "react": ">=16.8.0"
```
dustinbyrne: `ReactElement<any, any>` is available in 16.8.0, do we need to bump?

if so, `react`/`@types/react` version desync here

turnipdabeets: good call, updated

packages/react/src/components/internal/VisibilityAndClickTracker.tsx
turnipdabeets: updated to `ReactElement<any, any>`

dustinbyrne (approved):

PR #4647, opened 2026-08-25
fix(react-native): order native symbol upload after dSYM generation

### LINKED ISSUE #4646: uploadNativeSymbols: main app dSYM never ready on EAS Build, archive always fails
With `uploadNativeSymbols: true`, every EAS iOS archive fails in the
"Upload PostHog Debug Symbols" phase:

    ▸ Executing MyApp » Upload PostHog Debug Symbols
    ❌ error: Main app dSYM was not ready after 60 seconds:
       .../ArchiveIntermediates/MyApp/BuildProductsPath/Release-iphoneos/MyApp.app.dSYM/Contents/Resources/DWARF/MyApp
    ▸ ** ARCHIVE FAILED **

It is not flaky — it fails identically every run. The app compiles, links and
signs fine; only the symbol-upload script brings the build down.

In the Xcode log the embedded app extension's dSYM is generated
(`Generating debug MyApp/MyExtension » MyExtension.appex.dSYM`), but no
equivalent line appears for the main app before the script runs.

`upload-symbols.sh` polls for the dSYM rather than declaring it as an input,
with this comment:

> Declaring the dSYM as an input can create dependency cycles for apps with
> embedded extensions, so wait until the dSYM belongs to the current executable instead.

The project does have an embedded extension, so it is presumably the case that
comment is guarding against — but on EAS the poll never succeeds and the
60s timeout is a guaranteed failure rather than a slow path.

sentry-react-native hit the same symptom (main app dSYM missing, Expo plugin,
EAS) in getsentry/sentry-react-native#5288 and fixed it by adding `inputPaths`
to their build phase.

**Env:** posthog-react-native 4.64.2, posthog-cli 0.15.1, Expo SDK 57,
EAS image `macos-tahoe-26.5-xcode-26.6`, one embedded app extension.

**Ask:** either declare the dSYM as an input where it is safe to, or give this
phase a supported way not to fail the archive. Right now the only workaround is
`uploadNativeSymbols: false`, which means no native symbolication at all.

Caveat on the diagnosis: the timeout fires if the dSYM is missing, if its UUIDs
do not match the executable, or if it is still open for writing. The error does
not distinguish, so "never generated" is a reading of the log rather than confirmed.

marandaneto:
## Problem

Expo apps with `uploadNativeSymbols` can start the PostHog upload phase before Xcode produces the main app dSYM. In apps with an embedded extension, the upload script can then wait for 60 seconds and fail every archive because later build work is blocked behind the running phase.

Fixes #4646.

## Changes

- Add the documented main app DWARF path to the upload phase's Xcode input files.
- Keep plugin-generated upload phases last, after app extension embedding, while preserving custom input paths and customized scripts.
- Update the Expo 57 example so the PostHog plugin and generated Xcode phase run last and include the documented input path.
- Add regression coverage for new and existing build phases, input preservation, extension ordering, and customized phases.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Investigated and implemented with Pi. A minimal Xcode 26.6 project with an embedded app extension reproduced the timeout when the phase ran too early and the dependency cycle when only the dSYM input was added. Putting the phase after extension embedding and declaring the dSYM input built successfully. Pi autoreview found one incorrect use of the `xcode` API, which was fixed and reviewed again with no remaining findings.

### DIFF
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -314,7 +314,7 @@ export function addDsymUploadBuildPhase(xcodeProject: any, includeSource = false
         new Set([...(Array.isArray(existing.inputPaths) ? existing.inputPaths : []), POSTHOG_DSYM_INPUT_PATH])
       )
 
-      const buildPhases = xcodeProject.getFirstTarget?.().target?.buildPhases
+      const buildPhases = xcodeProject.getFirstTarget?.().firstTarget?.buildPhases
       if (Array.isArray(buildPhases)) {
         const phaseIndex = buildPhases.findIndex((phase: any) => phase.comment === POSTHOG_DSYM_BUILD_PHASE_NAME)
         if (phaseIndex !== -1 && phaseIndex !== buildPhases.length - 1) {
--- a/packages/react-native/src/tooling/expoconfig.ts
+++ b/packages/react-native/src/tooling/expoconfig.ts
@@ -2,7 +2,7 @@
 // Copyright (c) 2017 Sentry
 // Licensed under the MIT License: https://github.com/getsentry/sentry-react-native/blob/main/LICENSE.md
 
-const { withAppBuildGradle, withGradleProperties, withProjectBuildGradle, withXcodeProject } =
+const { withAppBuildGradle, withBaseMod, withGradleProperties, withProjectBuildGradle, withXcodeProject } =
   require('@expo/config-plugins')
 
 // com.posthog.android uploads R8 mapping files and injects a matching map-id so native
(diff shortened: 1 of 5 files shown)


packages/react-native/src/tooling/expoconfig.ts
```diff
@@ -307,11 +310,23 @@ export function addDsymUploadBuildPhase(xcodeProject: any, includeSource = false
...
+        if (phaseIndex !== -1 && phaseIndex !== buildPhases.length - 1) {
+          buildPhases.push(...buildPhases.splice(phaseIndex, 1))
+        }
+      }
     }
     return
   }
 
   xcodeProject.addBuildPhase([], 'PBXShellScriptBuildPhase', POSTHOG_DSYM_BUILD_PHASE_NAME, null, {
+    inputPaths: [POSTHOG_DSYM_INPUT_PATH],
```
turnipdabeets: The move-to-last splice only runs on the existing-phase branch — a fresh `addBuildPhase` appends wherever this mod happens to run, so on a clean prebuild (what EAS does) the ordering comes entirely from plugin order in `app.json`. That's why the example needed `posthog-react-native/expo` moved last. Ran this branch against a real pbxproj with `xcode@3.0.1`:

```
1. after posthog mod:       [..., "[CP] Copy Pods Resources", "Upload PostHog Debug Symbols"]
2. later plugin embeds ext: [..., "Upload PostHog Debug Symbols", "Embed Foundation Extensions"]  -> not last, input already declared
3. 2nd run of posthog mod:  [..., "Embed Foundation Extensions", "Upload PostHog Debug Symbols"]  -> last
```

So anyone whose app-extension plugin is listed after ours gets the same non-last + dSYM-input combination the description reproduces as a dependency cycle — a build break where today they'd only risk the 60s timeout. I think we need to hoist the splice out of the `existing` branch and run it after the `addBuildPhase` call too, so the ordering doesn't depend on `app.json`.

marandaneto: Good catch! Fixed in 224cec61e. The reorder now runs for both existing and newly added phases, and a post-processing Xcode mod runs after the rest of Expo’s Xcode mod chain so later extension plugins cannot leave the upload phase non-last. I also added a clean-prebuild/plugin-order regression test.

marandaneto:
I also validated this against a disposable real Xcode project using commit `224cec61e`, Xcode 26.6, and `xcode@3.0.1`. The project has an iOS app target plus an embedded notification-service extension and exercises the actual PostHog Expo plugin implementation.

| Scenario | `addBuildPhase` calls | Result |
|---|---:|---|
| Fresh add | 1 | Added the upload phase after `Embed App Extensions`; build succeeded |
| Existing generated phase | 0 | Reused a phase deliberately placed before extension embedding, moved it last; build succeeded |
| Full Expo mod chain | 1 | A simulated extension plugin displaced the freshly added phase, then PostHog's final Xcode wrapper restored it to last; build succeeded |

The existing-phase transition was:

```text
Before second run:
Sources → Frameworks → Resources → Upload PostHog Debug Symbols → Embed App Extensions

After second run:
Sources → Frameworks → Resources → Embed App Extensions → Upload PostHog Debug Symbols
```

The full Expo mod-chain scenario produced the same final order after the simulated later plugin displaced the phase.

All three builds verified that:

- the documented dSYM input path is present;
- the app dSYM binary was generated;
- the notification extension was embedded;
- `GenerateDSYMFile` ran before `PhaseScriptExecution Upload PostHog Debug Symbols`;
- `xcodebuild` completed with `** BUILD SUCCEEDED **`.

The upload script emitted the expected “not found” warning because the minimal project intentionally has neither CocoaPods nor a SwiftPM checkout of `posthog-ios`.

turnipdabeets (approved):

PR #4650, opened 2026-08-26
fix(react-native): fix survey position and Expo route warnings

marandaneto:
## Problem

React Native surveys can receive CSS-style positions such as `bottom-right` from the survey API. The SDK only recognizes canonical values such as `right`, so it warns and places the survey in the bottom center instead.

The Expo 57 sample also keeps its PostHog client in `app/posthog.tsx`. Expo Router treats that helper module as a route and warns that it has no default component export.

## Changes

- Normalize CSS-style survey positions before resolving the modal alignment. `bottom-right` now maps to the canonical `right` position.
- Fall back safely when remote configuration contains a non-string survey position.
- Add tests for CSS-style position names and malformed runtime values.
- Move the Expo sample's PostHog client outside the `app` directory, update its imports, and correct the README path.
- Add a patch changeset for `posthog-react-native`.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi implemented and reviewed the changes. XcodeBuildMCP and adb were used to run the Expo sample on iOS and Android simulators. The tests used the same `bottom-right` survey payload returned by the sample project's API. Both target warnings were absent from the simulator and Metro logs.


packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -92,8 +98,9 @@ export function resolveSurveyAlignment(position: string | undefined): {
 } {
   let resolvedPosition: SurveyPosition = defaultSurveyAppearance.position
   if (position) {
-    if (KNOWN_SURVEY_POSITIONS.has(position)) {
-      resolvedPosition = position as SurveyPosition
+    const normalizedPosition = normalizeSurveyPosition(position)
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 6ae9db2b9. The resolver now normalizes only string values and falls back safely for malformed remote positions. I also added a regression test for a non-string value.

marandaneto:
Fixed the stale Expo helper path noted by Greptile in 6ae9db2b9. The README now points to the root-level `posthog.ts` file.

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -82,6 +82,12 @@ export type SurveyFlexAlign = 'flex-start' | 'center' | 'flex-end'
 const KNOWN_SURVEY_POSITIONS: ReadonlySet<string> = new Set(Object.values(SurveyPosition))
 const warnedUnknownPositions = new Set<string>()
 
+// Some survey API clients use CSS-style names such as `bottom-right`, while
+// SurveyPosition represents the bottom row as `left`, `center`, and `right`.
+function normalizeSurveyPosition(position: string): string {
```
ioannisj, line 87: [suggestion] Web doesn't do any of this today: `getPopoverPosition` in `packages/browser/src/extensions/surveys/surveys-extension-utils.tsx:809` switches on the raw `appearance.position`, so a hyphenated value falls through to the default. That's harmless for `bottom-right` since web's default is `Right` anyway, but it only lines up by accident: `top-left` lands top-left on RN and bottom-right on web. 

Should this normalizer live in `packages/core` next to the `SurveyPosition` map (`packages/core/src/types.ts:663`) rather than in the RN package?

If the API really is sending these, fixing it server side and dropping the client normalizer is probably the cleaner call as well?

packages/react-native/src/surveys/surveys-utils.ts
```diff
@@ -92,8 +98,9 @@ export function resolveSurveyAlignment(position: string | undefined): {
 } {
   let resolvedPosition: SurveyPosition = defaultSurveyAppearance.position
   if (position) {
-    if (KNOWN_SURVEY_POSITIONS.has(position)) {
-      resolvedPosition = position as SurveyPosition
+    const normalizedPosition = typeof position === 'string' ? normalizeSurveyPosition(position) : ''
+    if (KNOWN_SURVEY_POSITIONS.has(normalizedPosition)) {
+      resolvedPosition = normalizedPosition as SurveyPosition
     } else if (!warnedUnknownPositions.has(position)) {
```
ioannisj, line 104: [nit] The lookup uses `normalizedPosition` now but the dedup set still keys on the raw `position`, so `foo-bar` and `foo_bar` each warn separately even though they normalize to the same thing. Keying the set on `normalizedPosition` closes it.

ioannisj (approved):
Should this be fixed upstream if it's sent by our API? That's the main question, otherwise lg

ioannisj:
Ah didn't see this was auto-merged, my bad

packages/react-native/src/surveys/surveys-utils.ts
marandaneto: Yes, the API should emit canonical values. The payload I reproduced came from the flags API for an existing survey with `bottom-right`, so the RN compatibility path still needs to cover persisted and cached legacy records after server-side validation is fixed. I kept this RN-local because the observed regression was RN falling back to center. Moving it into browser would broaden behavior for aliases such as `top-left` without evidence that the API emits them. We can share the normalizer in core if we decide to support every CSS-style alias across SDKs.

packages/react-native/src/surveys/surveys-utils.ts
marandaneto: Good catch! Fixed in b202e1817 and opened follow-up PR #4654. Warning deduplication now uses the normalized position, with regression coverage for equivalent hyphenated and underscored unknown values.

marandaneto: Follow-up PR: https://github.com/PostHog/posthog-js/pull/4654

Issue #49, opened 2020-06-19
CORS policy issue


fizerkhan:
I am getting following error during posthog metrics send.

```
Access to XMLHttpRequest at 'https://posthog.mysite-ops.com/decide/?ip=1&_=1592583273915' 
from origin 'https://app.mysite.com' has been blocked by CORS policy: The value of the 
'Access-Control-Allow-Credentials' header in the response is 'true, true' which must be 'true' 
when the request's credentials mode is 'include'. The credentials mode of requests initiated by 
the XMLHttpRequest is controlled by the withCredentials attribute.
````

How do I resolve it?

mariusandra:
I think this is now solved. Feel free to reopen if that's not the case.

PR #4651, opened 2026-08-26
fix(flags): align presence operator semantics

marandaneto:
## Problem

[sdk-specs PR #49](https://github.com/PostHog/sdk-specs/pull/49) defines omitted properties as inconclusive because SDK property maps contain partial context. The shared local evaluator treated an omitted key as matching `is_not_set`. That could produce a local result that disagrees with stored PostHog properties and remote evaluation.

## Changes

- Make an omitted key throw `InconclusiveMatchError` for `is_not_set`, matching existing `is_set` behavior.
- Keep own present keys conclusive, including null, undefined, false, zero, empty strings, and empty collections; inherited object properties do not count as map entries.
- Preserve remote fallback when an inconclusive presence condition precedes an `early_exit` rollout exclusion.
- Add shared Core, Node, and Convex regression coverage.

Validated with:

- `pnpm --filter @posthog/core exec jest --runInBand src/__tests__/featureFlagLocalEvaluation.spec.ts`
- `pnpm --filter posthog-node exec jest --runInBand src/__tests__/feature-flags.spec.ts`
- ESLint and Prettier checks for the changed Core and Node files
- `git diff --check`
- Autoreview on commit `4c709e522` with no actionable findings

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [x] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [x] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Pi coding agent worker subagents implemented the shared matcher change and its package-level tests. The bundled autoreview tool reviewed the final committed diff against `origin/main`.


packages/core/src/featureFlagLocalEvaluation.ts
```diff
@@ -219,7 +219,7 @@ export function matchFeatureFlagProperty(
     case 'is_not':
       return !computeExactMatch(value, overrideValue)
     case 'is_set':
-      return key in propertyValues
+      return hasProperty
```
ioannisj: [nit] `hasProperty` can only be `true` by the time we reach this case, the `!hasProperty` branch at line 190 already threw. `return true` says that outright, as written it reads like presence is still an open question here.

```suggestion
      return true
```

ioannisj (approved):

packages/core/src/featureFlagLocalEvaluation.ts
marandaneto: Agreed — simplified the already-established `is_set` result to `return true` in 4c709e522.

PR #4652, opened 2026-08-26
feat(error-tracking): standardize exception event metadata

hpouillot:
## Problem

Exception events emitted by different PostHog SDKs use inconsistent severity, source, mechanism, linkage, and metadata ownership semantics. This makes cross-platform grouping and processing less predictable.

## Changes

- Emit canonical `$exception_level` and `$exception_source` values at capture boundaries.
- Normalize mechanisms, synthetic provenance, and deterministic cause linkage across browser, Node.js, React Native, and Sentry integrations.
- Prevent generic properties from overriding SDK- and processor-owned exception metadata while preserving custom fingerprints.
- Add coverage for canonical metadata and a release changeset.

Canonical contract: https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/exception-event-metadata/spec.md

### Libraries affected

- [ ] All of them
- [x] posthog-js (web)
- [x] posthog-node
- [x] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with pi using the archived OpenSpec contract as the source of truth. The implementation keeps custom fingerprints caller-owned while reserving canonical metadata for SDK capture boundaries and downstream processors.

### DIFF
--- a/packages/browser/src/extensions/sentry-integration.ts
+++ b/packages/browser/src/extensions/sentry-integration.ts
@@ -95,9 +95,23 @@ export function createEventProcessor(
 
         const exceptions: _SentryException[] = event.exception?.values || []
 
-        const exceptionList = exceptions.map((exception) => {
+        const exceptionList = exceptions.slice(0, 50).map((exception, index) => {
+            const mechanism: Record<string, any> = {
+                ...(exception.mechanism || {}),
+                exception_id: index,
+            }
+            if (index === 0) {
+                delete mechanism.parent_id
+                delete mechanism.source
+            } else {
+                mechanism.type = 'chained'
+                mechanism.source = 'cause'
+                mechanism.parent_id = index - 1
+                delete mechanism.handled
+            }
             return {
                 ...exception,
+                mechanism,
                 stacktrace: exception.stacktrace
                     ? {
                           ...exception.stacktrace,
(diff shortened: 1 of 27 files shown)


marandaneto:
This change introduces backwards-incompatible public schema/API/contract changes: Generic captureException property bags can no longer override reserved $exception_* metadata, and emitted mechanism/handled/linkage semantics change across SDKs

If thats intentional thats ok

packages/browser/src/extensions/sentry-integration.ts
```diff
@@ -95,9 +95,23 @@ export function createEventProcessor(
...
+            }
+            if (index === 0) {
+                delete mechanism.parent_id
+                delete mechanism.source
+            } else {
+                mechanism.type = 'chained'
+                mechanism.source = 'cause'
+                mechanism.parent_id = index - 1
+                delete mechanism.handled
+            }
```
marandaneto, line 111: Sentry defines exception.values in oldest-to-newest order, but this mapping treats index 0 as the outermost exception and makes each later entry its child. For [root cause, outer error], the generated linkage is therefore reversed; with more than 50 entries, slice(0, 50) can also discard the outermost exception entirely. 

The sorting i believe its intentional right, but i think we should not cap it to 50

marandaneto (approved):
left comments but unblocking

PR #4655, opened 2026-08-26
feat(browser): add August 2026 config defaults

robbie-c:
## Problem

JSON-LD replay capture requires a separate opt-in setting. Customers who select the next config defaults should receive the sanitized replay metadata.

## Changes

- Add the 2026-08-30 config defaults after the existing 2026-08-29 defaults.
- Enable session_recording.captureJsonLd for the new defaults.
- Keep JSON-LD capture disabled for earlier defaults and explicit unset.
- Preserve partial replay configuration and an explicit false override.
- Keep autocapture.capture_copied_text opt-in.
- Update the public types, API reference, tests, and changeset.

This change only affects customers who select defaults: 2026-08-30 or a later version. JSON-LD still uses the existing sanitizer. It still preserves @id values, which needs a separate scrubbing follow-up.

Fast array.js comparison against origin/main: +0.02 KiB gzip (+0.02%).

### Libraries affected

- [x] posthog-js (web)
- [x] @posthog/types

## Checklist

- [x] Tests for new code
- [x] Accounted for the impact across supported platforms
- [x] Preserved legacy defaults and explicit customer overrides
- [x] Checked the gzip bundle-size change
- [x] Added a changeset file

## Validation

- 636 focused browser tests passed.
- All @posthog/types tests passed.
- The browser TypeScript build passed.
- ESLint passed.
- API reference generation passed with existing warnings.
- Replay incident-risk check found no matching incident pattern.

The separate browser test-typecheck command could not find the dompurify and dotenv type stubs in this worktree. The normal browser TypeScript build passed.

## 🤖 Agent context

Codex desktop implemented this user-directed change. The existing 2026-08-29 defaults already belong to another feature, so this PR uses 2026-08-30 as a separate boundary. Clipboard capture remains opt-in because copy and cut events include selected text. Paste events do not read pasted text.


packages/browser/src/posthog-core.ts
```diff
@@ -248,7 +255,7 @@ export const defaultConfig = (defaults?: ConfigDefaults): PostHogConfig => ({
     ui_host: null,
     asset_host: null,
     token: '',
-    autocapture: true,
+    autocapture: defaults && defaults !== 'unset' && defaults >= '2026-08-30' ? { capture_copied_text: true } : true,
```
marandaneto: blocking: Honor `mask_all_text` before enabling clipboard capture by default — with `defaults: '2026-08-30'`, projects that already set `mask_all_text: true` now register copy/cut capture, but the clipboard path still writes `window.getSelection()` to `$selected_content` regardless of `maskAllText`. This can send arbitrary selected page text despite the privacy setting, so clipboard content capture needs to respect that setting before becoming a default.

marandaneto (commented):

packages/browser/src/posthog-core.ts
```diff
@@ -740,6 +747,9 @@ export class PostHog implements PostHogInterface {
         if (isObject(baseConfig.rageclick) && isObject(userConfig.rageclick)) {
             mergedConfig.rageclick = extend({}, baseConfig.rageclick, userConfig.rageclick)
         }
+        if (isObject(baseConfig.autocapture) && isObject(userConfig.autocapture)) {
```
marandaneto: blocking: Preserve the date-gated option when `autocapture` is explicitly `true` — for `{ defaults: '2026-08-30', autocapture: true }`, the shallow merge replaces the base `{ capture_copied_text: true }` object with the boolean, and this object-only merge does not restore it. Clipboard handlers therefore remain disabled even though the new defaults were selected; boolean `true` should preserve the date-gated sub-options while `false` remains a full opt-out.

packages/browser/src/posthog-core.ts
marandaneto: Existing ph-no-capture/sensitive-element checks reduce exposure, but the selected text is global and, importantly, mask_all_text: true does not currently suppress $selected_content.
 That makes enabling this option by default a concrete PII risk unless masking behavior is fixed or clipboard events capture metadata only.

packages/types/src/posthog-config.ts
```diff
@@ -1723,6 +1726,7 @@ export interface PostHogConfig {
      * - `'2026-05-30'`: Defaults from '2026-01-30' plus `persistence_save_debounce_ms` defaults to `250`, `split_storage` and `detect_google_search_app` default to `true`, and rageclick defaults also exclude stepper controls and text-selection surfaces
      * - `'2026-06-25'`: Defaults from '2026-05-30' plus `session_recording.streamNetworkBody` defaults to `true` (streams network bodies to enforce the payload size limit)
      * - `'2026-08-29'`: Defaults from '2026-06-25' plus `cookieWinsOnConflict` defaults to `true` (the shared cross-subdomain cookie wins over stale per-origin localStorage)
+     * - `'2026-08-30'`: Defaults from '2026-08-29' plus `session_recording.captureJsonLd` and `autocapture.capture_copied_text` default to `true`
```
marandaneto: i think its safer to keep capture_copied_text disabled by default due to PII risk

robbie-c: Agreed, I didn't mean to include this, I just didn't describe what I wanted to the robot accurately enough

packages/browser/src/posthog-core.ts
robbie-c: capture_copied_text is no longer enabled by default

packages/browser/src/posthog-core.ts
robbie-c: capture_copied_text is no longer enabled by default

marandaneto (approved):

packages/browser/src/posthog-core.ts
```diff
@@ -200,6 +200,26 @@ const PRIMARY_INSTANCE_NAME = 'posthog'
...
+    }
+    if (defaults >= '2025-11-30') {
+        sessionRecording.strictMinimumDuration = true
+    }
+    if (defaults >= '2026-05-30') {
+        sessionRecording.canvasCapture = { resolutionScale: 0.6 }
+    }
+    if (defaults >= '2026-06-25') {
+        sessionRecording.streamNetworkBody = true
+    }
```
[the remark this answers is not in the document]
robbie-c: fixed

PR #4656, opened 2026-08-26
feat(next): capture route templates in pageviews

### LINKED ISSUE #3785: @posthog/next: allow capturing route template

I want to capture pageviews as /foo/${id}, not /foo/5

Something like what DataDog can do in their RUM implementation for NextJS:

https://github.com/DataDog/browser-sdk/blob/10ce053327f6d9c8f09736af06bd5f25a810bd69/packages/browser-rum-nextjs/src/domain/nextJSRouter/datadogAppRouter.tsx#L9-L17

https://github.com/DataDog/browser-sdk/blob/10ce053327f6d9c8f09736af06bd5f25a810bd69/packages/browser-rum-nextjs/src/domain/nextJSRouter/computeViewNameFromParams.ts#L5

I could copypaste their code and use it instead of <PostHogView/>, but I think it should be included in @posthog/next as an option

## Related sub-libraries

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [x] @posthog/next

marandaneto:
## Problem

`@posthog/next` pageview tracking only captures concrete paths such as `/posts/123`. Applications with dynamic routes cannot group pageviews by a stable route template such as `/posts/[id]`.

Closes #3785.

## Changes

- Add an opt-in `captureRouteTemplate` prop to `PostHogPageView` for both the App Router and Pages Router.
- Preserve concrete values in `$current_url` and `$pathname`, and capture the route template separately in `$route`.
- Use `router.pathname` for exact Pages Router templates and normalize optional catch-alls to `[...param]`.
- Infer App Router templates from `useParams()`. Ambiguous paths omit `$route` rather than emitting an incorrect template.
- Require Next.js 13.3 or later for App Router template inference at runtime. Earlier versions continue capturing ordinary pageviews and warn once when the option is enabled.
- Add coverage for dynamic, multiple, catch-all, encoded, ambiguous, optional catch-all, and unsupported-version behavior.

### Libraries affected

- [ ] All of them
- [ ] posthog-js (web)
- [ ] posthog-node
- [ ] posthog-react-native
- [ ] @posthog/react
- [ ] @posthog/ai
- [ ] @posthog/convex
- [x] @posthog/next
- [ ] @posthog/nextjs-config
- [ ] @posthog/nuxt
- [ ] @posthog/rollup-plugin
- [ ] @posthog/webpack-plugin
- [ ] @posthog/types
- [ ] @posthog/browser-common

## Checklist

- [x] Tests for new code

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. The App Router does not expose its source route template through a public API, so the implementation uses best-effort inference based on the DataDog approach referenced in the issue. Ambiguous static and dynamic segment values omit `$route` instead of emitting an incorrect template, while `$pathname` remains URL-derived.


marandaneto:
Browser validation passed with a temporary Next.js 15 app installed from the local `@posthog/next` tarball.

| Router | Concrete URL | Captured `$pathname` |
| --- | --- | --- |
| App Router | `/products/123?source=tarball-app-test` | `/products/[id]` |
| App Router | `/teams/acme/projects/42` | `/teams/[team]/projects/[projectId]` |
| App Router | `/docs/guides/setup` | `/docs/[...slug]` |
| App Router, ambiguous match | `/blog/blog` | `/blog/blog` fallback |
| Pages Router | `/legacy/987?source=tarball-pages-test` | `/legacy/[id]` |

I also verified client-side navigation for both routers. Each navigation captured a new pageview with the concrete URL in `$current_url` and the stable template in `$pathname`. The test ingestion requests returned HTTP 200.

A caller-provided route-template resolver could support exact templates when App Router inference is insufficient, including distinguishing `[...slug]` from `[[...slug]]`. That can be considered as a follow-up. This PR focuses on opt-in automatic inference with a safe concrete-path fallback for ambiguous matches.

packages/next/src/client/PostHogPageView.tsx
```diff
@@ -33,18 +34,97 @@ import { getCurrentUrl } from '../shared/browser.js'
...
+    }
+
+    for (const [paramName, paramValue] of Object.entries(params)) {
+        if (typeof paramValue !== 'string' || !paramValue) {
+            continue
+        }
+
+        const matches = findMatchingIndexes(segments, [paramValue])
+        if (matches.length !== 1) {
+            return undefined
```
turnipdabeets, line 107: When two params happen to share a value `computeRouteTemplate` returns undefined and we send no `$pathname` at all, so posthog-js backfills the raw path. Confirmed with a repro:

```
/orgs/1/projects/2  ->  /orgs/[orgId]/projects/[projectId]
/orgs/1/projects/1  ->  /orgs/1/projects/1   (no template)
```

One route, two rows in the Paths tile, and nothing on the event marking which is which — so the grouping the flag exists to provide is silently incomplete for anyone who turns it on.

Bailing is the right call, to be clear — I tried hard to make this emit a *wrong* template and couldn't. I think we just need the fallback to be visible: always set `$pathname`, and add a flag like `$pathname_is_template` so people can filter out the ones that fell back.

packages/next/src/client/PostHogPageView.tsx
```diff
@@ -53,8 +133,11 @@ function PageViewTracker() {
             return
         }
 
-        posthog.capture('$pageview', { $current_url: currentUrl })
-    }, [pathname, searchParams, posthog])
+        posthog.capture('$pageview', {
+            $current_url: currentUrl,
+            ...(routeTemplate ? { $pathname: routeTemplate } : {}),
```
turnipdabeets: [question] Web Analytics reads `$pathname` for the Paths tile, but Entry Path and End Path come from `$entry_pathname`, which is derived from `$current_url` (`sessions_v2.py`). So after opting in you get templates in one tile and raw paths in the tile next to it. `$prev_pageview_pathname` and every non-pageview event stay raw too.

Has web analytics agreed that `$pathname` can hold a template? If they'd rather keep it URL-derived, a separate `$route` property (like RN's `$screen_name`) avoids the mismatch.

packages/next/src/client/PostHogPageView.tsx
```diff
@@ -33,18 +34,97 @@ import { getCurrentUrl } from '../shared/browser.js'
  * }
  * ```
  */
-export function PostHogPageView() {
+export interface PostHogPageViewProps {
```
turnipdabeets: [suggestion] The doc block above (with the `app/layout.tsx` example) now attaches to `PostHogPageViewProps` instead of `PostHogPageView`, so hovering the component in an editor shows nothing. Declaring the interface above the doc block puts it back.

packages/next/src/pages/PostHogPageView.tsx
```diff
@@ -27,7 +27,17 @@ import { getCurrentUrl } from '../shared/browser.js'
  * }
  * ```
  */
-export function PostHogPageView() {
+export interface PostHogPageViewProps {
```
turnipdabeets: [suggestion] Same here — the doc block moved onto `PostHogPageViewProps` and `PostHogPageView` is left undocumented.

packages/next/src/pages/PostHogPageView.tsx
```diff
@@ -39,8 +49,11 @@ export function PostHogPageView() {
             return
         }
 
-        posthog.capture('$pageview', { $current_url: currentUrlWithoutHash })
-    }, [router.asPath, router.isReady, posthog])
+        posthog.capture('$pageview', {
+            $current_url: currentUrlWithoutHash,
+            ...(captureRouteTemplate ? { $pathname: router.pathname } : {}),
```
turnipdabeets: [nit] For optional catch-alls `router.pathname` gives `/docs/[[...slug]]`, but the App Router side normalizes to `/docs/[...slug]` — same route, two different strings. Worth normalizing here too, or noting it in this JSDoc the way the App Router one does.

arnohillen:
`useParams` doesn't exist in `next/navigation` before 13.3.0 (`next@13.2.4/dist/client/components/navigation.js` exports only `useSearchParams`, `usePathname`, `useRouter`, `useSelectedLayoutSegment(s)`, `redirect`, `notFound`), while `packages/next/package.json` still declares `next: ">=13.0.0"`, so on 13.0-13.2 opting into `captureRouteTemplate` calls an undefined import at line 115 and throws during render. Bumping the peer floor to `>=13.3.0` covers it.

packages/next/src/pages/PostHogPageView.tsx
marandaneto: Fixed in aeee7b21c. The Pages Router props interface is now declared before the component doc block. The rebuilt declaration confirms the JSDoc is attached to `PostHogPageView`.

packages/next/src/client/PostHogPageView.tsx
marandaneto: Fixed in aeee7b21c. The props interface is now declared before the component doc block. After rebuilding, the generated declaration has the JSDoc attached to the `PostHogPageView` function rather than `PostHogPageViewProps`.

packages/next/src/client/PostHogPageView.tsx
marandaneto: Fixed in aeee7b21c. Route templates now use the separate `$route` property while `$pathname` remains concrete. When App Router inference is ambiguous, `$route` is omitted, so the fallback is visible as a missing route value rather than being mixed into the template property. The equal-parameter regression test and browser validation both confirm this behavior.

packages/next/src/client/PostHogPageView.tsx
marandaneto: Fixed in aeee7b21c. I confirmed the mismatch: Web Analytics Paths reads event `$pathname`, while entry/end paths remain URL-derived, and `$prev_pageview_pathname` remains concrete. The implementation now preserves URL-derived `$pathname` and captures the template separately in `$route`.

packages/next/src/pages/PostHogPageView.tsx
marandaneto: Fixed in aeee7b21c. Pages Router optional catch-alls are normalized from `[[...slug]]` to `[...slug]`, matching App Router output. Added a focused regression test and browser-validated both routers.

marandaneto:
Addressed the Next.js compatibility feedback in aeee7b21c without raising the package-wide peer floor. `captureRouteTemplate` now feature-detects `useParams`; on Next.js versions before 13.3 it falls back to ordinary pageview capture and warns once instead of throwing. The prop JSDoc documents the 13.3 minimum for App Router template inference. Verified with a focused unavailable-`useParams` regression test and an isolated Next.js 13.2 build/runtime reproduction.

marandaneto:
@dustinbyrne opinions here (see $route)?

packages/next/src/client/PostHogPageView.tsx
```diff
@@ -33,18 +48,101 @@ import { getCurrentUrl } from '../shared/browser.js'
...
+        segments[matches[0]] = `[${paramName}]`
+    }
+
+    return segments.join('/')
+}
+
+let hasWarnedAboutUnsupportedRouteTemplates = false
+
+function RouteTemplatePageViewTracker() {
+    if (typeof useParams !== 'function') {
```
turnipdabeets: [question] Can this branch actually fire? `useParams` is a named import at the top of the file, so on a bundler that treats `next/navigation` as ESM the missing export is a build error before any warn runs. I haven't tried it on 13.2 — have you? If it can't fire, documenting 13.3+ on the prop would let us drop the guard and the module-level warn flag.

dustinbyrne:
i didn't review for correctness here because the bots already have

one consideration i might add is that named parameters can be subject to change on a single static route. e.g., using the example from @turnipdabeets above:

```txt
/orgs/1/projects/2 -> /orgs/[orgId]/projects/[projectId]
```

if one day someone changes the parameter from `orgId` -> `organizationId`, it might silently break analytics reporting for that route. this is somewhat minor, but might be worth weighing in on before we solidify this. other options would be to use a bash-like substitution which does not include the parameter name, like:

```txt
/orgs/1/projects/2 -> /orgs/$1/projects/$2
/blog/hello-world -> /blog/$1
```

or a glob-style routes:
```txt
/orgs/1/projects/2 -> /orgs/*/projects/*
/docs/guides/getting-started -> /docs/**
```

or we leave it as it's currently implemented which also seems reasonable to me despite the potential for drift. after all, as noted in the original issue, this is the pattern that others are using as well.

dustinbyrne:
other than that, i think this looks great

marandaneto:
@dustinbyrne Good point. I’d keep the Next-native named template for now because the original request explicitly asks for named parameters, `router.pathname` already provides this format for Pages Router, and the referenced Datadog implementation emits the same `[paramName]` form. The names also make catch-all routes and breakdowns easier to interpret. A parameter rename would therefore be treated as a route-schema change. We can revisit anonymous normalization if this causes reporting drift in practice.

dustinbyrne (approved):

